PR #24, opened 2023-06-03
chore: add jest as dev package

MentalGear:
I was surprised there's a script: "test": "jest", but no jest installed. I guess you had it installed globally.


typpo:
Just added this in https://github.com/typpo/promptfoo/commit/e6f2bcbff4898c3ffc7376368f56fc6c504e6792, but included the updated package-lock.json too.  Jest was previously installed as a peer dependency of ts-jest

PR #25, opened 2023-06-05
chore: add native ts-node compatibility

MentalGear:
by removing file extensions from import statements (file extension on import statements are optional)


typpo:
Thanks and apologies for the delayed review!  The .js extensions were originally there because it was written as an esmodule, now the extensions are no longer necessary.  If it makes the dev experience better, let's remove them :)

PR #29, opened 2023-06-09
fix parsing of env vars for OPENAI_MAX_TOKENS and OPENAI_MAX_TEMPERATURE

abi:
I was running into API responses when I set OPENAI_MAX_TEMPERATURE since it gets read in as a string:

```
 [FAIL] API response error:           │ Recipe App                           │ \- type of meat                      │
│ TypeError: Cannot read properties of │                                      │                                      │
│  undefined (reading '0'): {"error":{ │                                      │                                      │
│ "message":"'1' is not of type 'numbe │                                      │                                      │
│ r' - 'temperature'","type":"invalid_ │                                      │                                      │
│ request_error","param":null,"code":n │                                      │                                      │
│ ull}}     
```

Same for OPENAI_MAX_TOKENS. I tested the code path in lines 221-22 but I'm not sure how it to trigger the lines 140-41 code path.

### DIFF
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -137,8 +137,8 @@ export class OpenAiCompletionProvider extends OpenAiGenericProvider {
     const body = {
       model: this.modelName,
       prompt,
-      max_tokens: process.env.OPENAI_MAX_TOKENS || 1024,
-      temperature: options?.temperature ?? (process.env.OPENAI_MAX_TEMPERATURE || 0),
+      max_tokens: parseInt(process.env.OPENAI_MAX_TOKENS || '1024'),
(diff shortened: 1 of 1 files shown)


typpo:
😱 thanks Abi!  It also occurs to me that it should be `OPENAI_TEMPERATURE`, not `OPENAI_MAX_TEMPERATURE`.  Shall we fix that here too?

typpo:
Ah, just saw your other PR.  I'm going to land this PR and rename the var, and cut a `0.10.0` release that uses `OPENAI_TEMPERATURE`

typpo:
Done in 48305573e328ac67a6a59790cd2ae3dbfce00085 and released to 0.10.0

abi:
Awesome, thanks for the quick turnaround!

PR #37, opened 2023-06-15
extract and export generateTable()

tizmagik:
Closes https://github.com/typpo/promptfoo/issues/36

The table generated from the CLI is great, let’s export it

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,10 +4,11 @@ import telemetry from './telemetry';
 import { evaluate as doEvaluate } from './evaluator';
 import { loadApiProviders } from './providers';
 import { readTests } from './util';
-
 import type { EvaluateOptions, TestSuite, TestSuiteConfig } from './types';
 
 export * from './types';
+
+export { generateTable } from './table';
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -30,11 +31,7 @@ async function evaluate(testSuite: EvaluateTestSuite, options: EvaluateOptions =
   return ret;
 }
 
-module.exports = {
-  evaluate,
-  assertions,
-  providers,
-};
+export { evaluate, assertions, providers };
```
tizmagik, line 34: This change technically unrelated, but allows parity with how we'd export `generateTable`. It also fixed an issue where I ended up having to import like this (even with `allowSyntheticDefaultImports` enabled 🤷):

```ts
import {default as promptfoo} from 'promptfoo';
```

With this change, I think the following should work in most setups:

```ts
import {evaluate} from 'promptfoo';
```

Open to reverting this though if you see it as a problem.

typpo:
Looks great!  Thank you!

PR #46, opened 2023-06-26
update docs/Question

Ben-Epstein:
Perhaps I'm misunderstanding, but based on the documentation and the nature of the assertions, isn't this the correct structure of the test?

If not, why are these assertions not nested under the `assert`?


typpo:
Hi Ben, you're absolutely correct.  Thanks for fixing the formatting here, not sure how it got mangled.

PR #58, opened 2023-07-09
Add Anthropic provider

zeldrinn:
Unit tested and manually tested with live calls


zeldrinn:
@typpo ready for review!

typpo:
Looks great!  I can't test it myself, but everything looks reasonable to me according to the Anthropic docs.  Thanks so much :grin:

PR #69, opened 2023-07-14
Allow option to disable sharing

therealpaulgg:
It is not always desirable to have the ability to share results (storing sensitive data on a server), so this PR allows sharing to be disabled.

Disable sharing through DISABLE_SHARING environment variable, or through adding 'sharing: false' to the YAML confinguration.

If sharing is disabled, there will also be no sharing functionality in the browser (the button will be removed).

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -137,6 +137,11 @@ async function main() {
       });
       await telemetry.send();
 
+      if (!defaultConfig.sharing) {
+        logger.error('Sharing is not enabled. Add `sharing: true` to your promptfooconfig.yaml');
+        process.exit(1);
+      }
+
       const latestResults = readLatestResults();
       if (!latestResults) {
         logger.error('Could not load results. Do you need to run `promptfoo eval` first?');
@@ -246,6 +251,7 @@ async function main() {
         prompts: cmdObj.prompts || fileConfig.prompts || defaultConfig.prompts,
         providers: cmdObj.providers || fileConfig.providers || defaultConfig.providers,
         tests: cmdObj.tests || cmdObj.vars || fileConfig.tests || defaultConfig.tests,
+        sharing: process.env.DISABLE_SHARING === "1" ? false : defaultConfig.sharing ?? true,
         defaultTest: fileConfig.defaultTest,
       };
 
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -137,6 +137,11 @@ async function main() {
       });
       await telemetry.send();
 
+      if (!defaultConfig.sharing) {
+        logger.error('Sharing is not enabled. Add `sharing: true` to your promptfooconfig.yaml');
+        process.exit(1);
+      }
+
```
typpo, line 144: Rather than requiring the user to edit their config after they've explicitly run `promptfoo share`, it might be smoother to just always include a confirmation "Are you sure you want to create a public URL?"

Happy to add this separately if that saves you some trouble.

src/main.ts
```diff
@@ -246,6 +251,7 @@ async function main() {
         prompts: cmdObj.prompts || fileConfig.prompts || defaultConfig.prompts,
         providers: cmdObj.providers || fileConfig.providers || defaultConfig.providers,
         tests: cmdObj.tests || cmdObj.vars || fileConfig.tests || defaultConfig.tests,
+        sharing: process.env.DISABLE_SHARING === "1" ? false : defaultConfig.sharing ?? true,
```
typpo: nit - `PROMPTFOO_DISABLE_SHARING` for consistency with other envars, and let's check `fileConfig` too (this is used when the user explicitly sets a config with the `-c` option)

something like this - 
```js
sharing: process.env.PROMPTFOO_DISABLE_SHARING === "1" ? false : (fileConfig.sharing ?? defaultConfig.sharing ?? true),
```

typpo (changes requested):
Great change!  Just two small nits

src/main.ts
therealpaulgg: added!

src/main.ts
therealpaulgg: I like that idea. I'd prefer adding that separately, thanks!

typpo: Added in https://github.com/promptfoo/promptfoo/commit/53cad30bd9cf6df8c1b938bcf61e09ad997a78f8

PR #83, opened 2023-07-21
testing providers w/ same model diff params

therealpaulgg:
One feature we would like is the ability to compare identical models with different settings (differing temperature, functions, etc).

This PR is a start - potentially needs some work.

- Adds 'id' to constructor for azure and openai chat & completion providers, allowing base function to be overridden with user-provided identifier
- allows id to be specified in the provider object rawconfig
- adds example of temperature comparison w/ identical models

Example:

```yaml
providers:
  - openai:gpt-3.5-turbo-0613:
      id: openai-gpt-3.5-turbo-lowtemp
      prompts: chat_prompt
      config:
        temperature: 0
        max_tokens: 128
  - openai:gpt-3.5-turbo-0613:
      id: openai-gpt-3.5-turbo-hightemp
      prompts: chat_prompt
      config:
        temperature: 1
        max_tokens: 128
  - openai:gpt-4-0613:
      prompts: chat_prompt
      config:
        temperature: 0
        max_tokens: 128
```

NOT in this PR, but just an idea, it may be desirable to refactor the configuration of providers into something like:

```yaml
providers:
  - type: openai:gpt-3.5-turbo-0613
    id: openai-gpt-3.5-turbo-lowtemp
    prompts: chat_prompt
    config:
      temperature: 0
      max_tokens: 128
  - type: openai:gpt-3.5-turbo-0613:
    id: openai-gpt-3.5-turbo-hightemp
    prompts: chat_prompt
    config:
      temperature: 1
      max_tokens: 128
  - type: openai:gpt-4-0613:
    prompts: chat_prompt
    config:
      temperature: 0
      max_tokens: 128
```

That way the name of the provider is inside of the object.

So JSON structure goes from `[{"provider": {...}}]` -> `[{"type": "provider", ...}]`

This might also help with code clarity and remove the need to use `Object.keys(provider)[0]`.

### DIFF
--- a/examples/gpt-3.5-temperature-comparison/README.md
+++ b/examples/gpt-3.5-temperature-comparison/README.md
@@ -1,0 +1,11 @@
+To get started, set your OPENAI_API_KEY and REPLICATE_API_TOKEN environment variables.
+
+Next, change a few of the prompts in prompts.txt and edit promptfooconfig.yaml.
+
+Then run:
+
+```
+promptfoo eval
+```
+
+Afterwards, you can view the results by running `promptfoo view`
--- a/examples/gpt-3.5-temperature-comparison/promptfooconfig.yaml
+++ b/examples/gpt-3.5-temperature-comparison/promptfooconfig.yaml
@@ -15,12 +15,12 @@ providers:
       config:
         temperature: 1
         max_tokens: 128
-  - replicate:replicate/llama70b-v2-chat:e951f18578850b652510200860fc4ea62b3b16fac280f83ff32282f87bbd2e48:
-      prompts: completion_prompt
+  - openai:gpt-4-0613:
+      prompts: chat_prompt
       config:
-        temperature: 0.01 # minimum temperature
-        max_length: 128
-
+        temperature: 0
+        max_tokens: 128
+        
 tests:
   - vars:
       message: hello world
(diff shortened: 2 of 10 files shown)


therealpaulgg:
One thing I've noticed is that the total count of tests run seems to be inaccurate. Not entirely sure how to resolve this yet.

image

typpo:
I'll play around with this PR a bit but overall looks pretty good.  

Just confirming my understanding, the main problem we're trying to solve here is the lack of proper labeling in the results UI. 
 Something like this already works, but the results are indistinguishable (unless you remember the order of providers in the config):

```yaml
prompts: prompts.txt
providers:
  - openai:gpt-3.5-turbo:
      config:
        temperature: 0
  - openai:gpt-3.5-turbo:
      config:
        temperature: 0.5
  - openai:gpt-3.5-turbo:
      config:
        temperature: 1.0
```

I agree a refactor could be in order at some point, though nitpicking I'd probably avoid overloading `id` and use `label` instead.

```yaml
providers:
  - id: openai:gpt-3.5-turbo-0613
    label: openai-gpt-3.5-turbo-lowtemp
    prompts: chat_prompt
    config:
      temperature: 0
      max_tokens: 128
```

therealpaulgg:
Whenever I had the same ID in that provider in the current changeset, there were problems with the providerMap. Hence the need for a unique identifier for each provider w/ identical model configuration.

It also does affect the display results, yes.

typpo:
Just checking @therealpaulgg, I haven't kept up with this because you mentioned it's WIP.  If you think it's ready to land, LMK!  No rush

therealpaulgg:
I haven't really had any other changes in mind for this PR besides the potential refactor later down the line.

Let me know if there are any other nits

Issue #73, opened 2023-07-18
Allow separation of 'tests' vs. 'scenarios'


therealpaulgg:
Another feature that we think would be desirable is the ability to run a provided number of tests against 'test cases' or 'test data'.

For example, there may be the following definition:
```yaml
tests:
  - description: Is Valid JSON
    assert
      - type: is-json
  - description: Similarity
    assert:
      - type: similar
```

This means there are two tests that will be run. In our case, there could be 5-10 'scenarios' we would like to run, both with a somewhat different input and different expected output.

Rough idea of what a config with 'scenarios' could look like:

```yaml
prompts: [prompts.txt]
providers: [openai:gpt-3.5-turbo]
scenarios:
  - testData: testData1.txt
    expectedOutput: output1.txt
    expectedSimilarity: 0.8
  
  - testData: testData2.txt
    expectedOutput: output2.txt
    expectedSimilarity: 0.9
  
  - testData: testData3.txt
    expectedOutput: output3.txt
    expectedSimilarity: 0.5

tests:
  - description: Is Valid JSON
    vars:
      testData: {{testData}}
    assert:
      - type: is-json
      - type: javascript
        value: typeof JSON.parse(output) === 'object'

  - description: Meets Expected Output
    vars:
      expectedOutput: {{expectedOutput}}
    assert:      
      - type: similar
        value: {{expectedOutput}}
        threshold: {{expectedSimilarity}}
```

This would result in a total of 6 tests being run: 3 scenarios, 2 tests, 1 provider, 1 prompt.

We also would love to be able to import variables from text files so the yaml test configuration can be cleaner.

Interested in thoughts on this one, and I can help out where needed.

typpo:
Picking this one up soon - thinking about ways to clean up:
- the number of configs (enabling multiple test suites)
- the number of test cases (enabling better importing/organization for test cases and vars)

therealpaulgg:
Appreciate it! 

Striving to achieve a cleaner config with a lot of tests has been the most challenging aspect so far. We used `$ref` in our config extensively.

I definitely like the idea of being able to run multiple `TestSuite` objects, one for each dataset. All those test suites could use the exact same configuration.

PR #88, opened 2023-07-24
Add support for loading test cases from file/directory path

### LINKED ISSUE #86: Support for loading test cases from directory
## Description

To facilitate easier management, discussion, and understanding of individual test cases, it would be beneficial to allow defining and loading tests (consisting of variables and assertions pairs) from a specified directory.

## Proposed Implementation

Consider a scenario where the test definitions could be mentioned in a configuration file in the following way:

promptfooconfig.yaml
```yaml
prompts: [prompts/prompt1.yaml]
providers: [openai:gpt-3.5-turbo]
tests: './tests/
```

Where the test directory would look like this:

```
tests
├── 1.yaml
├── 2.yaml
└── 3.yaml
```

Each yaml will then hold the defintion of the input vars and the assertions for one specific test case:

```yaml
vars:
  text: Some input text for the test
assert:
  - type: not-contains-any
    value:
      - "diagnose"
  - type: contains-all
    value:
      - "enigmatic creatures"
  - type: similar
    value: Some input text for the testing
    threshold: 0.95
```

## Potential Benefits:

**1. Granular Test Case Management:** This structure enables users to work with each test case as a separate unit, alleviating the necessity of maintaining all test cases within a single file.

**2. Facilitates Discussion:** A numerical test case system allows teams to easily refer and discuss specific test cases.

**3. Grouping Tests:** This structure would enable users to categorize their tests based on directory names (for example, tests/accuracy/ and tests/creativity/), enabling easier comparison of prompt performance across various performance categories, rather than on an aggregate measure across all tests.

Currently, while tests.csv is used, it does not allow defining multiple assertions for the same variable. This proposed change will be especially useful when the number of assertions for each test case is significantly large, making their management in separate files more efficient.

I'm open to contributing towards the implementation of this feature. Please let me know if any additional information is required.

typpo:
Lets you do things like this:

```yaml
prompts: prompts.txt
providers: openai:gpt-3.5-turbo

# Loads & runs all test cases matching these filepaths
tests:
  # You can supply an exact filepath
  - tests/tests2.yaml

  # Or a glob (wildcard)
  - tests/*

  # Mix and match with actual test cases
  - vars:
      var1: foo
      var2: bar
```

This is also valid:
```yaml
tests: tests/*
```

Or this:
```yaml
tests: ['tests/accuracy', 'tests/creativity']
```

----------------

Also adds support for importing vars, like this:

```yaml
tests:
  - vars: path/to/vars*.yaml
``` 

----------------

Closes #86 
Relevant to #73

### DIFF
--- a/examples/separate-test-configs/README.md
+++ b/examples/separate-test-configs/README.md
@@ -1,0 +1,7 @@
+This example shows how you can use paths and wildcards to simplify for `promptfooconfig.yaml`.
+
+Run:
+
+```
+promptfoo eval
+```
--- a/examples/separate-test-configs/promptfooconfig.yaml
+++ b/examples/separate-test-configs/promptfooconfig.yaml
@@ -1,0 +1,20 @@
+prompts: prompts.txt
+providers: openai:gpt-3.5-turbo
+
+# The defaultTest configuration is applied to every test case in this config.
+defaultTest:
+  options:
+    suffix: Be extremely concise
+
+# Loads & runs all test cases matching these filepaths
+tests:
+  # You can supply an exact filepath
+  - tests/tests2.yaml
+
+  # Or a glob (wildcard)
+  - tests/*
+
+  # Mix and match with actual test cases
+  - vars:
+      topic: the internet
+      content_type: witty tweets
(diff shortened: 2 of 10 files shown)


typpo:
Hi @KasperDjernaes,

You raise a very good point.  The terminology is a bit overloaded, and some renaming probably makes sense for the 1.0 version bump.

I'm very interested in learning about your use case and any other feedback you have.  You're of course welcome to open GitHub issues!  But if you'd prefer to send thoughts in private, feel free to email me at github@ianww.com :)

PR #95, opened 2023-08-02
function call azure fix

therealpaulgg:
Returns function_call arguments properly as a string

### DIFF
--- a/src/providers/azureopenai.ts
+++ b/src/providers/azureopenai.ts
@@ -268,7 +268,7 @@ export class AzureOpenAiChatCompletionProvider extends AzureOpenAiGenericProvide
     try {
       const message = data.choices[0].message;
       const output =
-        message.content == null ? message.function_call.arguments : message.content;
+        message.content == null ? message.function_call : message.content;
       return {
         output,
         tokenUsage: cached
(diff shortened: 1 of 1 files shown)


therealpaulgg:
May want to do something similar for OpenAI. i had to make this change to make function calling work at all.

typpo:
Hi Paul!  I understand the intent here, but this omits the name of the called function, which others might be using in their evals :(

The OpenAI API returns `arguments` as a JSON string, so by default that value is passed through with escapes and all.  I believe you could use [postprocessing](https://promptfoo.dev/docs/configuration/guide/#postprocessing) to parse it out of a test case:

```yaml
tests:
  - options:
      postprocess: JSON.parse(output.arguments)
```

This would return a clean JSON object containg the function call arguments

therealpaulgg:
Ah I did not know about postprocess. This should solve that in that case.

I do question why we are returning a JSON.stringify regardless and not just the entire object?

therealpaulgg:
Just tried newest version of promptfoo. still does not work today, I think the triple equals is causing issues.

therealpaulgg:
I updated this PR with what I believe would fix this & retain existing desired functionality.

therealpaulgg:
@typpo was wondering what your current thoughts on this are, or if there are any blockers that prevent this from going out?

typpo:
Thank Paul!  Sorry for the delay.

I've made the corresponding fix to the main OpenAI provider in 89678fdfa126d54a55b3747efe60e409663ef020, and will get this out shortly.

PR #144, opened 2023-09-15
Adding YAML document start

jamesbraza:
- Adds YAML document start to all YAML files and default `promptfoodefaultconfig.yaml`
- ~Fixes `--config` option to have proper casing in `promptfoodefaultconfig.yaml`~ (fixed upstream in https://github.com/promptfoo/promptfoo/commit/72d426e366d491a667cee4c2173197c844b46748)

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,1 +1,2 @@
+---
 github: [typpo]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,3 +1,4 @@
+---
 name: CI
 on: [push]
 env:
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,3 +1,4 @@
+---
 trailingComma: 'all'
 singleQuote: true
 printWidth: 100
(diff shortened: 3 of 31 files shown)


jamesbraza:
@typpo just to share motivation here, my company uses [`yamllint --strict`](https://github.com/adrienverge/yamllint) with `.yamllint` config:

```yaml
---
extends: default

rules:
    comments:
        min-spaces-from-content: 1 # Match prettier
```

It reports warnings for not having document start (`---`) in YAML files.  I figured it's nice to upstream `---` to promptfoo, that way it works out of the box for end users using `yamllint --strict`.

If you're interested, I can also add `yamllint` to `promptfoo`'s CI pipline

typpo:
Hey @jamesbraza, sorry for the delay and thanks for waiting patiently on this change.

My inclination is to skip this, because although the `---` at the start of YAML files is indeed part of the YAML specification, strict linting is not widely used and I'm concerned that this default might cause confusion for casual users or give the impression that the `---` is a necessary part of all config files.  I think the value prop of promptfoo is the simplicity of the setup (as opposed to writing code tests), so even though this seems like a minor nit I just want to make the config as flexible as possible.

LMK what you think - nonetheless really appreciate your attention to detail and the suggestion!

jamesbraza:
Yeah I concur with what you're saying, all good!  Closing this one out

PR #153, opened 2023-09-19
Fix README link to providers docs

greysteil:
Just trying our promptfoo - thanks for creating it!

The existing providers docs link was outdated/broken, so I've fixed it up. I switched to using a reference, too, to make it easier to maintain.

:octocat:


typpo:
Thank you for the catch!

PR #158, opened 2023-09-20
Allow object with function name as a value for function_call

greysteil:
As [documented by OpenAI](https://platform.openai.com/docs/api-reference/chat/create). (Looks like everything just works when passing this through from a config file, too.)


typpo:
Thanks @greysteil!

PR #171, opened 2023-09-21
Updated node version update of Google Colab notebook example

gnakan:
This updates the Google Colab notebook example.  Currently, the approach downgrades the version that Google Colab makes available. This approach uses the `n` Node.js version manager for simplicity. You can view my test example at https://colab.research.google.com/drive/1NOzolcu5aH_bMQ6qyxX4DdY4QrRCFk45?usp=sharing.


typpo:
Thanks!

PR #178, opened 2023-09-22
Add support for parsing multiple json blobs in responses

IanCal:
The JSON check had a regex to pull out a json object. It was greedy though so wouldn't work for multiple json entities, and any extra brackets would throw it off. I add a basic setup to iterate over the string extracting all valid json blobs.

You could also use a streaming parser, but you need to manually add in bits of rolling things back.


typpo:
Great change!  Thank you

Issue #177, opened 2023-09-22
Request: OpenAI provider `OPENAI_API_BASE_URL` over `OPENAI_API_HOST`


jamesbraza:
When doing some hacky stuff (in https://github.com/promptfoo/promptfoo/issues/176), I am hitting an issue where `https://` is hardcoded into the OpenAI provider.

What do you think of making the default OpenAI provider `https://api.openai.com` (including `https://`)?  Then, one can override the protocol (http, https) with `OPENAI_API_BASE_URL` (instead of `OPENAI_API_HOST`).

Fwiw, this is how Ollama, Llama.cpp, and LocalAI providers already work, they include the protocol in the environment variable/default.

PR #187, opened 2023-09-22
`OPENAI_API_HOST` to `OPENAI_API_BASE_URL`

jamesbraza:
Closes https://github.com/promptfoo/promptfoo/issues/177 by matching LocalAI, Llama.cpp, and Ollama convention of `_BASE_URL` and `getApiUrl`

If accepted, I will open a sister PR in the `promptfoo-docs` to mention `OPENAI_API_BASE_URL`/`getApiUrl`

### DIFF
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -9,8 +9,6 @@ import type {
   ProviderResponse,
 } from '../types.js';
 
-const DEFAULT_OPENAI_HOST = 'api.openai.com';
-
 interface OpenAiCompletionOptions {
   temperature?: number;
   max_tokens?: number;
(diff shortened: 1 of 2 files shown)


typpo:
Thanks for opening this.  Totally fine with this change, but we need to support `OPENAI_API_HOST` if present for backwards compatibility.  There could be others who rely on the old environment variable in their current setups.

jamesbraza:
I have brought it back, as well as `apiHost` in the config 👍 

Please note: this is the first 1+ lines of TypeScript I have _ever_ written, so please check my work, thanks again!

typpo:
Looks great!  I made a small readability change but otherwise you're a Typescript natural ;)

Issue #175, opened 2023-09-21
Request: supporting embeddings endpoint for LocalAI and Ollama


jamesbraza:
LocalAI has an embeddings endpoint ([source](https://github.com/go-skynet/LocalAI/blob/7bdf707dd3d4eaeb09eea31484ab2f51fba1ba32/api/api.go#L195), [docs](https://localai.io/features/embeddings/)):

```bash
> curl http://localhost:8080/embeddings -X POST -H "Content-Type: application/json" -d '{
    "input": "Your text string goes here",
    "model": "bert-embeddings"
  }' | jq "."
...
{
  "object": "list",
  "model": "bert-embeddings",
  "data": [
    {
      "embedding": [
        0.051830754,
        ...,
        -0.023937061
      ],
      "index": 0,
      "object": "embedding"
    }
  ],
  "usage": {
    "prompt_tokens": 0,
    "completion_tokens": 0,
    "total_tokens": 0
  }
}
```

Ollama also has one: [`/api/embeddings`](https://github.com/jmorganca/ollama/blob/v0.0.19/server/routes.go#L527)

Trying to use a LocalAI model for grader:

```yaml
defaultTest:
    options:
        provider: localai:chat:bert-embeddings
```

I get this error:

```none
[FAIL] Error: Invariant failed: Provider must implement callEmbeddingApi for similarity check
Error: Invariant failed: Provider must implement callEmbeddingApi for similarity check
at invariant (/Users/user/code/promptfoo/node_modules/tiny-invariant/dist/tiny-invariant.cjs.js:14:11)
at matchesSimilarity (/Users/user/code/promptfoo/dist/src/matchers.js:63:34)
at async runAssertion (/Users/user/code/promptfoo/dist/src/assertions.js:367:17)
at async runAssertions (/Users/user/code/promptfoo/dist/src/assertions.js:49:24)
at async Evaluator.runEval (/Users/user/code/promptfoo/dist/src/evaluator.js:182:37)
at async /Users/user/code/promptfoo/dist/src/evaluator.js:418:25
```

The only providers right now (0.24.1) that implemented `callEmbeddingApi` are OpenAI and Azure.

Can we add `callEmbeddingApi` to the LocalAI and Ollama providers?  Thank you in advance

typpo:
[...]

PR #191, opened 2023-09-23
Better LocalAI/Ollama embeddings traversal failure

jamesbraza:
- https://github.com/promptfoo/promptfoo/pull/190, but now for LocalAI and Ollama
- Removed extra `ret` locals

Relates to https://github.com/promptfoo/promptfoo/issues/175.

### DIFF
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -122,7 +122,7 @@ export class OpenAiEmbeddingProvider extends OpenAiGenericProvider {
     try {
       const embedding = data?.data?.[0]?.embedding;
       if (!embedding) {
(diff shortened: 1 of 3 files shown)


src/providers/openai.ts
```diff
@@ -122,9 +122,9 @@ export class OpenAiEmbeddingProvider extends OpenAiGenericProvider {
     try {
       const embedding = data?.data?.[0]?.embedding;
       if (!embedding) {
-        throw new Error(`No embedding found in OpenAI embeddings API response: ${JSON.stringify(data)}`);
+        throw new Error('No embedding found in OpenAI embeddings API response');
```
jamesbraza, line 125: Relying on handling `catch` to print the `data`, otherwise it will be printed twice in the stack trace

typpo:
Thanks James!

PR #201, opened 2023-09-27
Adding TXT and HTML to `--output` help/error message

jamesbraza:
See PR title


jamesbraza:
If approved, I will open accompanying PR for `promptfoo-docs`

Issue #227, opened 2023-10-18
Allow multiple output formats simultaneously with `outputPath`.


Wehzie:
Currently, `outputPath` in `promptfooconfig.yaml` supports only one argument of type string.
In addition, `outputPath` should support type `List[str]` to allow outputting multiple formats from a single run simultaneously.
Currently, supported extensions are `csv, txt, json, yaml, yml, html`.
An additional extension should be `viewer`.
Where `viewer` writes to the web or console viewer.
Meanwhile, `html` should produce an HTML file at the provided path.

typpo:
Thanks for the PR!  Closing this out as #229 is in the latest release.

PR #229, opened 2023-10-18
Allow multiple output formats simultaneously with `outputPath`

Wehzie:
Close issue https://github.com/promptfoo/promptfoo/issues/227 to allow multiple outputs.


typpo:
Nice change!  Thanks

PR #237, opened 2023-10-23
allow Python in tests.csv

Wehzie:
Additionally I added the option to call inline JS with `javascript:`

This PR addresses part of https://github.com/promptfoo/promptfoo/issues/238


typpo:
Thank you for this change!

PR #253, opened 2023-10-30
Fix prompt per provider

nirkopler:
when configuring the YAML as following:
```
prompts:
  llama.py: ml
  gpt.txt: gp
providers:
  - openai:gpt-3.5-turbo-0613:
      prompts: gp
  - huggingface:text-generation:meta-llama/Llama-2-7b-chat-hf:
      prompts: ml
```

the `.py` file is not assigned with the right alias (ml), which means the same configuration file **can't** use multiple models with each model has a different prompting format

I fixed that and updated the relevant test


typpo:
Thank you!

PR #255, opened 2023-10-30
Fix broken default config for OpenAI evals created in web app

finnless:
When creating a new eval from the web interface using openai, the default option for apiHost is malformed / wrong.
Pasted Graphic 1

It will always cause this error.
Pasted Graphic

If the default attribute is removed, the issue is resolved.

Pasted Graphic 4

The error no longer appears and the evals are created as intended.
Pasted Graphic 3


typpo:
Thanks!

PR #259, opened 2023-10-31
Add caching support to Anthropic provider

sihil:
I'm not sure why but the anthropic provider had no caching support, possibly because it's the only provider that doesn't use fetch.

This PR provides an implementation to support caching for calls to anthropic, along with tests for the anthropic behaviour, improvements to the tests for caching behaviour of OpenAI (these were not actually testing what they claimed) and the addition of a call to clear the cache after every test to ensure that the cache does not change the behaviour of tests.


typpo:
Great change!  Thanks

PR #261, opened 2023-11-02
Broken custom api host for OpenAI.

leonleonho:
When providing an API host to the open ai provider an extra curly brace is injected causing the feature to break.


typpo:
Thanks for the catch!

PR #273, opened 2023-11-09
Update chat completion endpoint in azureopenai.ts

Elio-Khater:
fixed the following error:

{
    "error": {
        "message": "Unrecognized request argument supplied: dataSources",
        "type": "invalid_request_error",
        "param": null,
        "code": null
    }
}


typpo:
Thanks for catching this.  Does this endpoint work if `dataSources` is not set, or should we only use it when `dataSources` is set?

PR #276, opened 2023-11-10
feat: throw error correctly when invalid api key is passed for OpenAI

gregoryduckworth:
Failing tests like the ones in the examples would pass as a false positive due to the API not being called and the so the strings would never truely be compared


typpo:
Great catch - thanks!

PR #277, opened 2023-11-10
Provide the prompt in the context to external assertion scripts

sihil:
When using an inline script the prompt is available in the provided `context`. This is not currently the case with external scripts.

By pulling the creation of the `context` up a few blocks we can use it in place of the manually created objects and this will provide a more consistent interface.


typpo:
Thank you!

PR #280, opened 2023-11-13
fix the following error : 'List should have at least 1 item after val…

Elio-Khater:
fix the following error : 'List should have at least 1 item after validation, not 0' when stop attribute is empty array


typpo:
Thanks!  I'm guessing this is a change in validation on their side, as it used to work just fine.

PR #296, opened 2023-11-27
feat: Add support for external provider configs via file://

typpo:
This enables `file://` provider imports like so:

```yaml
prompts: 'Tell me about the weather in {{city}}.'
providers:
  - file://./gpt-3.5.yaml
tests:
  - vars:
      city: Boston
  # ...
```

### DIFF
--- a/src/providers.ts
+++ b/src/providers.ts
@@ -1,5 +1,10 @@
+import fs from 'fs';
 import path from 'path';
 
+import invariant from 'tiny-invariant';
+import yaml from 'js-yaml';
+
+import logger from './logger';
 import {
   OpenAiAssistantProvider,
   OpenAiCompletionProvider,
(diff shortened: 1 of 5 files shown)


src/providers.ts
```diff
@@ -101,7 +106,15 @@ export async function loadApiProvider(
     config: options.config,
     env,
   };
-  if (providerPath?.startsWith('exec:')) {
+  if (providerPath.startsWith('file://')) {
+    const filePath = providerPath.slice('file://'.length);
+    const yamlContent = yaml.load(fs.readFileSync(filePath, 'utf8')) as ProviderOptions;
+    console.log('got yaml content', yamlContent);
```
fabioxgn: Is this console.log intentional?

typpo: Nice catch!

PR #325, opened 2023-12-06
fix: sharing option is degraded 

mocyuto:
Sharing option at config file was degraded at unified config implementation.

Checking `fileConfig.sharing` like below, but it never assign.
https://github.com/promptfoo/promptfoo/blob/71af29f5266a89b446fbc65eea3d6f2a845f562e/src/main.ts#L325-L327

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -170,6 +170,7 @@ export async function readConfigs(configPaths: string[]): Promise<UnifiedConfig>
       (prev, curr) => ({ ...prev, ...curr.commandLineOptions }),
       {},
     ),
+    sharing: configs.at(-1)?.sharing,
   };
 
   return combinedConfig;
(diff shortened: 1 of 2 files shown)


src/util.ts
```diff
@@ -170,6 +170,7 @@ export async function readConfigs(configPaths: string[]): Promise<UnifiedConfig>
       (prev, curr) => ({ ...prev, ...curr.commandLineOptions }),
       {},
     ),
+    sharing: configs.at(-1)?.sharing,
```
typpo: Would it make sense to do something like
```ts
sharing: !configs.some(config => config.sharing === false),
```
So that if _any_ config has sharing disabled, the merged config won't have sharing either?

mocyuto: This would be better!
I will fix it

typpo:
Nice catch!  Thank you for this fix

PR #326, opened 2023-12-06
fix: check whether telemetry is disabled

mocyuto:
Instead of  `PROMPTFOO_DISABLE_TELEMETRY=1` , "telemetry is enabled" is logged.

Maybe it would be logged not only FirstRun but every run?

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -35,7 +35,7 @@ export class Telemetry {
   }
 
   maybeShowNotice(): void {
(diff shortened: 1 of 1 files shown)


typpo:
Sorry for not responding earlier @mocyuto.  A while back I made this tradeoff to reduce the amount of spam text with each run.  Let's keep it at first run for now, if it becomes a problem we can land this PR.

PR #327, opened 2023-12-08
feat: retry when getting internal server error with PROMPTFOO_RETRY_5XX envar

jtsai-quid:
Hi teams, 
I would like to make promptfoo able to retry sending requests when getting an internal server error.
This is useful when the LLM server is too busy or other temporary errors (ex: connection errors).
Any suggestions and inputs would be highly appreciated!

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -60,7 +60,12 @@ export async function fetchWithRetries(
     : 5000;
   for (let i = 0; i < retries; i++) {
     try {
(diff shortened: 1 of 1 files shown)


typpo:
Thanks for this PR!  I modified the functionality to put it behind an environment variable `PROMPTFOO_RETRY_5XX`.  The reason for this is that typically 5XX requests are not retried, however I understand certain providers (like OpenAI) have reliability issues.

PR #350, opened 2023-12-21
feat: add openai tool parameter

Undertone0809:
Add openai tool parameter
https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

### DIFF
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -31,7 +31,7 @@ type OpenAiCompletionOptions = OpenAiSharedOptions & {
   functions?: OpenAiFunction[];
   function_call?: 'none' | 'auto' | { name: string };
   tools?: OpenAiTool[];
-  tool_choice?: 'none' | 'auto' | { name: string };
+  tool_choice?: 'none' | 'auto' | { type: 'function'; function?: { name: string }}
   response_format?: { type: 'json_object' };
   stop?: string[];
   seed?: number;
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -31,7 +31,7 @@ type OpenAiCompletionOptions = OpenAiSharedOptions & {
   functions?: OpenAiFunction[];
   function_call?: 'none' | 'auto' | { name: string };
   tools?: OpenAiTool[];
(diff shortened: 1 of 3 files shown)


src/providers/openai.ts
```diff
@@ -30,6 +30,8 @@ type OpenAiCompletionOptions = OpenAiSharedOptions & {
   best_of?: number;
   functions?: OpenAiFunction[];
   function_call?: 'none' | 'auto' | { name: string };
+  tools?: OpenAiTool[];
+  tool_choice?: 'none' | 'auto' | { name: string };
```
typpo: According to the [OpenAI docs](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools), should this be
```ts
{
  // ...
  tools_choice?: 'none' | 'auto' | { type: 'function'; function?: { name: string }}
}
```

src/providers/openaiUtil.ts
```diff
@@ -8,6 +8,11 @@ export interface OpenAiFunction {
   parameters: any;
 }
 
+export interface OpenAiTool {
+  type: string;
+  function: any;
```
typpo: Maybe we can type this as 
```ts
export interface OpenAiTool {
  type: string;
  function: OpenAiFunction;
}
```

typpo (changes requested):
This is a good start!  You'll also have to add `tools` and `tool_choice` to the request to OpenAI in `OpenAiChatCompletionProvider.callApi`.

src/providers/openai.ts
Undertone0809: done

src/providers/openaiUtil.ts
Undertone0809: done

typpo (approved):
Thanks for the change!  I also added docs and modified the `callApi` function to pass the `tools` and `tool_choice` params

Undertone0809:
@typpo Next stage, I think we can add relevant assert.

typpo:
@Undertone0809 `is-valid-openai-tools-call` assert type added in #354

typpo:
@Undertone0809 `is-valid-openai-tools-call` assert type added in #354

PR #355, opened 2023-12-25
docs: optimize docs of openai tool usage

Undertone0809:


typpo:
Nice improvements.  Thank you!

Issue #356, opened 2023-12-26
HF Text Generation Inference Response Parsing Issue


amihalik:
I'm trying out your hugging face provider for text-generation-inference and there is a small bug.

The hugging face "inference endpoint" returns a list of dicts, but the text-generation-inference servers only return a dict.  So `response.data[0]?.generated_text` needs to be `response.data?.generated_text`.

Also, the docs for the hugging face "inference endpoint" notes that the "[return values is either a dict or a list of dicts if you sent a list of inputs.](https://huggingface.co/docs/api-inference/detailed_parameters?code=curl#text-generation-task)".

Here's an example request to a TGI endpoint:
```
{
  "inputs": "My name is Olivier and I",
  "parameters": {
    "top_k": 10,
    "top_p": 0.95,
    "temperature": 0.5,
    "repetition_penalty": 1.03,
    "max_new_tokens": 20,
    "return_full_text": true,
    "do_sample": true
  }
}
```

and here's the response:
```
{
  "generated_text": "My name is Olivier and I am a 3D artist and animator from France.\n\nI have been working in the"
}
```

amihalik:
Also, here's the custom provider that I'm currently using and it's working 

```
const fetch = require('node-fetch');

class HuggingfaceTgiProvider {
  constructor(options = {}) {
      const { id, config = {} } = options;
      this.id = id ? () => id : this.id;
      this.config = config;
      this.debug = config.debug || false;
  }

  id() {
      return `huggingface:text-generation-inference:${this.id}`;
  }

  toString() {
      return `[Huggingface Text Generation Inference Provider ${this.id}]`;
  }

  async callApi(prompt) {
      const params = {
          inputs: prompt,
          parameters: this.config,
      };
      const url = this.config.apiEndpoint;

      if (this.debug) {
          console.log("Request URL:", url);
          console.log("Request Params:", params);
      }

      let response;
      try {
          response = await fetch(url, {
              method: 'POST',
              headers: {'Content-Type': 'application/json'},
              body: JSON.stringify(params),
          });

          const jsonResponse = await response.json();
          if (this.debug) {
              console.log("Raw Response:", jsonResponse); 
          }

          if (jsonResponse.error) {
              return {
                  error: `API call error: ${jsonResponse.error}`,
              };
          }
          if (!jsonResponse) {
              return {
                  error: `Malformed response data: ${jsonResponse}`,
              };
          }
          return {
              output: jsonResponse?.generated_text,
          };
      }
[...]

PR #357, opened 2023-12-26
fix: handle huggingface text generation returning dict

typpo:
Related to #356

### DIFF
--- a/src/providers/huggingface.ts
+++ b/src/providers/huggingface.ts
@@ -84,14 +84,14 @@ export class HuggingfaceTextGenerationProvider implements ApiProvider {
           error: `API call error: ${response.data.error}`,
         };
       }
-      if (!response.data[0]) {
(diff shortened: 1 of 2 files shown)


amihalik:
Thanks for the quick fix!  This update works on my environment.

I changed my `promptfoo` version to `github:promptfoo/promptfoo#huggingface-textgeneration-fix` and this set of providers works great!

```
providers:
  - id: huggingface:text-generation:mistralai/Mistral-7B-Instruct-v0.2
    prompts: mistral_prompt
    config:
      apiEndpoint: http://x.x.x.229:8080/generate
      temperature: 0.01
      max_new_tokens: 128
  - id: huggingface:text-generation:meta-llama/Llama-2-7b-chat-hf
    prompts: llama_prompt
    config:
      apiEndpoint: http://x.x.x.171:8080/generate
      temperature: 0.01
      max_new_tokens: 128
```

Issue #373, opened 2024-01-05
HTML Output rendering does not have all information


NathanKarthik1996:
Hello,

When I rendering my results to HTML the title of columns say object Object (I am using the custom provider). Is there a way I can get the provider instead, maybe the file name that I use could be better instead. Moreover, there is no information about what assertion type was executed against the test case as well. It would be nice to add that information as well.

typpo:
Thanks for the report.  This is fixed in 0.35.1

PR #393, opened 2024-01-12
fix: update html output rendering with more details

typpo:
Related to #373


typpo:
Thanks for flagging, @naoro.  For now, I will revert to JSON header in CSV to avoid interrupting your workflow.  Done in 4f297f86033c72c6eaf8c0963f352218fc1f0b2f

PR #418, opened 2024-01-28
fix: Only open previous results when necessary (uses lots of memory)

greysteil:
The `listPreviousResults` function loads a _lot_ into memory if you're passing around big config files (which happens quickly when using `file://`). It's called in a bunch of places where all that's actually needed are the filenames.

This PR introduces a new `listPreviousResultFilenames` method and calls that wherever possible, which turns out to be everywhere except the `/results` view.

Removes a ~30 second pause for me (and then eventually OOM errors) after promptfoo finished an eval. All of that waiting turned out to be my laptop struggling with the call to `cleanupOldResults`, which was loading all of those results into memory. (For me that was 60 ~200kb files * 70 results, which adds up!)

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -417,15 +417,20 @@ export function writeLatestResults(results: EvaluateSummary, config: Partial<Uni
 
 const resultsCache: { [fileName: string]: ResultsFile | undefined } = {};
 
-export function listPreviousResults(): { fileName: string; description?: string }[] {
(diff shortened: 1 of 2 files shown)


typpo:
Excellent change, thank you!

I've been slowly working on a branch that moves results to sqlite, because managing separate files has been getting messier with all the ways evals/prompts/datasets are getting accessed.

PR #479, opened 2024-02-16
fix(prompt): Use single quotes to prevent shell from interpreting special characters

zanitete:


typpo:
Hey @zanitete, thanks for raising this.

The problem with this approach is that the single quotes cause compatibility issues with Windows environments.  I've opened #481 to switch this over to use the `PythonShell` library, which handles these escaping issues nicely.  

That should solve whatever issue you're having, hopefully!

zanitete:
thank a lot @typpo for the quick turnaround, your fix works perfectly!

PR #484, opened 2024-02-19
feat(webui): Add upload testcase csv to eval page

gingerhendrix:
Adds Upload Test Case CSV functionality with the same behaviour as specifying a CSV test case in the YAML configuration file.


src/web/nextui/src/utils.ts
```diff
@@ -0,0 +1,127 @@
+import { Assertion, AssertionType, CsvRow, TestCase } from "@/../../../types";
```
gingerhendrix: @typpo I noticed the web UI doesn't have a dependency on the library itself (except for types), so I duplicated what I needed here, but i guess reusing the library directly would be better.

typpo:
Great change!  I moved the `testCasesFromCsv` and `assertionFromString` functions into their own file and imported it, so that the code is shared.

PR #489, opened 2024-02-21
Change variable name to what the prompt template expects

pether-b:
Before, the query part of the resulting prompt was always empty and the result was always "Insufficient Information" (i.e. 0)

Fixes https://github.com/promptfoo/promptfoo/issues/490


typpo:
Thanks for catching this!

PR #503, opened 2024-02-28
Print incorrect response from factuality checker

jameshfisher:
Messages like these are very mysterious:

    expected
    ```
    Foo is bar.
    ```
     to match factuality with "Foo is baz", but it did not.
    Reason:
    Invalid option: )

This failure is not a normal test-case failure. It's an internal failure in promptfoo. So it should print as much debug info as it can to report the bug.

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -389,7 +389,7 @@ export async function matchesFactuality(
       reason = optionReasons[option];
     } else {
       pass = false;
-      reason = `Invalid option: ${option}`;
+      reason = `Invalid option: ${option}. Full response from factuality checker: ${resp.output}`;
     }
 
     let score = pass ? 1 : 0;


jameshfisher:
Example output:

```
    expected
    ```
    I'm sorry, I cannot answer that question based on the provided documents.
    ```
     to match factuality with "Foo is baz", but it did not.
    Reason:
    Invalid option: ). Full response: E) The answers differ, but these differences don't matter from the perspective of factuality.
```

Exposes this bug:

```
> const response = "E) The answers differ, but these differences don't matter from the perspective of factuality."
undefined
> response.trim().charAt(1).toUpperCase()
')'
```

typpo:
Thanks for this change!  Definitely makes it easier to debug.

PR #505, opened 2024-02-28
Support missing open parenthesis (fixes #504)

### LINKED ISSUE #504: Failure to parse response from factuality checker results in mysterious `Invalid option: )` test failure
We get test failures like this:

    expected
    ```
    Foo is bar.
    ```
     to match factuality with "Foo is baz", but it did not.
    Reason:
    Invalid option: )

With [additional logging](https://github.com/promptfoo/promptfoo/pull/503), we see the full response from the model is:

```
E) The answers differ, but these differences don't matter from the perspective of factuality.
```

[promptfoo tries to extract the letter `A`...`E` here](https://github.com/jameshfisher/promptfoo/blob/7a1ef8afb1343ec7e84a849cfeed5b90baa5fac0/src/matchers.ts#L365) with

```
    const option = resp.output.trim().charAt(1).toUpperCase();
```

But this does not work with this response:

```
> const response = "E) The answers differ, but these differences don't matter from the perspective of factuality."
undefined
> response.trim().charAt(1).toUpperCase()
')'
```

jameshfisher:
See https://github.com/promptfoo/promptfoo/issues/504 - factuality checker model can respond without an opening paren

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -362,7 +362,17 @@ export async function matchesFactuality(
 
   invariant(typeof resp.output === 'string', 'factuality produced malformed response');
   try {
-    const option = resp.output.trim().charAt(1).toUpperCase();
+    const output = resp.output;
+    // The preferred output starts like "(A)...", but we also support leading whitespace, lowercase letters, and omitting the first parenthesis.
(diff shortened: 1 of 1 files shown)


typpo:
Thank you!

Issue #523, opened 2024-03-04
Claude3 Support


zhlmmc:
Claude3 is released, seems like they changed the API:
![image](https://github.com/promptfoo/promptfoo/assets/1726527/a08e8a97-11fb-4ad1-8286-72c6578988d2)

typpo:
Wondering if @streichsbaer, who wrote our mistral provider, has any interest in this one :)

I'd love to make this change but am in the hospital with a family member right now.  Otherwise, will probably get to it in the coming weeks.

aantn:
@typpo hope all is well soon!

streichsbaer:
Funny you should say that. I just came here to offer my help with this and saw the message :)

On it!

streichsbaer:
Alrighty, have just opened a PR that adds support for Claude 3.
@typpo, let me know if there is anything else I can help with 👍 .

@zhlmmc, you can test the changes locally by following these steps:
1. Clone the repo
2. Check out the PR branch
3. Add the new provider to your `promptfooconfig.yaml`
```
provider:
- id: anthropic:messages:claude-3-opus-20240229
    config:
      temperature: 0.0
```
4. You can use the OpenAI prompt template, e.g. `prompts: [prompt.json]`. System prompts are optional.
```
[
  {
    "role": "system",
    "content": "{{ system_message }}"
  },
  {
    "role": "user",
    "content": "{{ question }}"
  }
]
```
5. Then just run `npx /path/to/promptfoo eval`

Let me know if there are any issues!

typpo:
Thanks @streichsbaer!  This is released in 0.45.0

PR #526, opened 2024-03-05
feat(anthropic): Add Claude 3 support

streichsbaer:
## Overview
Anthropic has released the Claude-3 series of models. This PR adds support for that (#523).
The code is updated to use the latest anthropic SDK and the new messages API.
The legacy completions API is still supported and this PR does not introduce any breaking changes.

- [x] All API parameters are covered
- [x] System messages are supported
- [x] The cost of all models is supported
- [x] Caching is supported
- [x] API key can be provided in the config or via the `ANTHROPIC_API_KEY` environment variable

## References
- Release notes: https://www.anthropic.com/news/claude-3-family
- Pricing: https://www.anthropic.com/api#pricing

@typpo these changes should be good to go, they were tested locally.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.44.0",
       "license": "MIT",
       "dependencies": {
-        "@anthropic-ai/sdk": "^0.5.2",
+        "@anthropic-ai/sdk": "0.16.1",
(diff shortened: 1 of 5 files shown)


typpo:
Thank you as always, @streichsbaer!

Issue #549, opened 2024-03-13
Feature Request: Add tools and tool_choice config to Azure Openai provider


heartyguy:
Description: https://github.com/promptfoo/promptfoo/blob/main/src/providers/azureopenai.ts#L38 Currently the Azure open ai provider only allow functions config in the provider config. Our team is trying to test using the tools and tool_choice parameters. 

The following 
```
prompts:
  - "Be helpful and tell me the past info in {{city}}."

providers:
  - id: azureopenai:chat:gpt4turbo
    config:
      apiHost: 
      apiKey: 
      temperature: 0.9
      tools: [
        {
        "type": "function",
          "function": {
            "name": "get_current_info",
            "description": "Get the current info in a given location",
            "parameters": {
              "type": "object",
                "properties": {
                  "location": {
                    "type": "string",
                      "description": "The city and state, e.g. San Francisco, CA"
                    },
                    "unit": {
                      "type": "string",
                      "enum": ["celsius", "fahrenheit"]
                    }
                  },
              "required": ["location"]
            }
          }
        }
      ]
      tool_choice: 'auto'

tests:
   - vars:
        city: Boston
     assert:
        - type: is-json
        - type: is-valid-openai-tools-call
        - type: javascript
          value: output[0].function.name === 'get_past_info'
        - type: javascript
          value: JSON.parse(output[0].function.arguments).location === 'Boston, MA'

   - vars:
        city: New York

```

is sending requests without the tools and tool_choice parameters:
```
npx promptfoo@latest eval -c tool_test.yaml --verbose
Calling Azure OpenAI API: {"model":"gpt4turbo","messages":[{"role":"user","content":"Be helpful and tell me the past info in Boston."}],"max_tokens":1024,"temperature":0.9,"top_p":1,"presence_penalty":0,"frequency_penalty":0} 
```

typpo:
#553 adds support for Azure OpenAI `tools` and `tool_choice`.  It will be included in the next published release :)

typpo:
Oops, totally missed your PR!  Sorry about that, will merge yours instead.

typpo:
Your change is published in 0.47.0!

PR #550, opened 2024-03-13
add tools and tool_choice config parameters to azure openai provider

heartyguy:
add tools and tool_choice config parameters to azure openai provider

Tested and working!

https://github.com/promptfoo/promptfoo/issues/549


typpo:
Thanks!

PR #562, opened 2024-03-17
Add anthropic:messages and replicate:mistral as default providers to web ui

matt-hendrick:
The replicate mistral config options mirror their API defaults (listed [here](https://replicate.com/mistralai/mistral-7b-instruct-v0.2/api)).

The expected format for the anthropic messages ID (`anthropic:messages:claude...`) didn't quite work with the pre-existing method of generating label text, which assumed that the second result from splitting on `:` would be the appropriate descriptive label text. I added an if check to handle that specific case, but let me know if you'd prefer an alternate implementation there (or if there is anything else you'd prefer done differently).


typpo:
Nice change, thank you Matt!

PR #565, opened 2024-03-18
DOCS: adjust configuration for python provider

romaintoub:
I've tested the python provider and to get access to the configuration variables added, you need an extra step
added in the doc
cheers!


typpo:
Nice catch, thanks!

PR #570, opened 2024-03-20
Mistral provider support for URL and API key envar

### LINKED ISSUE #567: Mistral provider is hard-coded to https://api.mistral.ai/v1/chat/completions
There is no way for the Mistral provider to override the API URL. So you can't test against, for example, Mistral hosted on Azure. It would be ideal if this provider supported the `apiHost`, `apiBaseUrl`, and `apiKeyEnvar` configuration parameters like the OpenAI provider does.

If there's interest, I can work on a PR for this.

jvert:
Fixes #567 

This adds support for overriding the default URL and api key environment variable in the Mistral provider. I just duplicated the existing OpenAPI behavior.


typpo:
Great change!  Thanks @jvert

PR #605, opened 2024-04-01
feat: ability to use js files as `transform`

typpo:
For example:

```yaml
# ...
tests:
  - vars:
      language: French
      body: Hello world
    options:
      transform: file://path/to/transform.js
    # ...
```

transform.js
```js
module.exports = (output, context) => {
  // Use context.vars, context.prompt ...
  return output.toUpperCase();
}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ scratch
 node_modules
 dist
 .aider*
+
+__pycache__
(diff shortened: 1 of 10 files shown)


abhagsain:
image

abhagsain:
image

PR #633, opened 2024-04-04
feat: add provider reference to prompt function

guilhermetk:
This PR introduces a new parameter, `provider`, to the function responsible for generating prompts. The addition of this parameter enables dynamic adjustment of the prompt to meet the specific requirements of different providers.


typpo:
Great change @guilhermetk!  I added support for python and moved the provider into the `context` object.  Merging now

PR #639, opened 2024-04-06
Support claude vision and images

CamdenClark:
Upgrade the anthropic SDK to get support for images in the SDK.

Include a new example on how to do vision in Claude.

Unfortunately, anthropic has a special way of doing images that is different from OpenAI. Anthropic doesn't support grabbing images from a url, you have to use a base64 representation of an image as input. Also, a lot of the keys are changed around, making it very confusing.

My approach in this commit is to embrace that the content block for anthropic prompts is going to have to be different from OpenAI. So you'll have to make an anthropic specific prompt to compare openai in vision.


typpo:
Great change!  Thanks @CamdenClark

PR #645, opened 2024-04-08
feat: run providers 1 at a time with --interactive-providers

typpo:
Many users don't have enough RAM to run all their models locally. The eval command includes an `--interactive-providers` option that runs evals on only 1 provider at a time:

```sh
promptfoo eval --interactive-providers
```

When you include this argument, the eval will prompt you when it's time to swap out the currently running model.


knoopx:
can you add a flat/variant to just run in serial without asking? ollama offloads models everytime and is a waste of time.

PR #651, opened 2024-04-09
feat: support for inline yaml for is-json, contains-json in csv

jernkuan:
- if string value for is-json & contains-json does not start with file:// assume string would be yaml configuration
- Update regex to support multiline string


typpo:
Nice change :)

Issue #518, opened 2024-03-01
Command line usage with -r --providers


Wehzie:
The documentation offers two types for using the `-r --providers` flag.
Firstly, by name, this works as expected, however it's unclear how to pass options via a provider's `config` entry.
Second, by path. The source suggests that a .yaml file is expected, as such I prepared the following command.

```
npx promptfoo@latest eval -c experiment.yaml -r file://provider.yaml
```

Where `provider.yaml` is of the form

```
providers:
  - model-name:
      id: unique-identifier
      config:
        temperature: 0.0
        # other options
```

Promptfoo expects an ID for the provider file. Adding one as follows, leads to promptfoo searching for a module with the name `provider-id`.

```
id: provider-id
providers:
  ...
```

Thus, how does one use the path option, correctly?
Lastly, I'd like to suggest for multiple paths and names to be a valid entry to the commandline. For my purpose, I'd like to specify a range of provider.yaml files with configurations set, and the ability to select one or multiple by command-line.

Congrats on 2k stars!

mldangelo:
Hi @Wehzie,

Thanks for the question about the `-r --providers` flag!

After reviewing this issue, I can clarify that the `-r` flag is designed for loading custom provider **implementations** (JavaScript or Python files), not provider **configuration** files.

**What works:**
```bash
# Use provider names
promptfoo eval -r openai:gpt-4 anthropic:claude-3

# Use custom provider implementations
promptfoo eval -r file://my_custom_provider.py
```

**What doesn't work:**
```bash
# Loading provider configs from YAML - NOT SUPPORTED
promptfoo eval -r file://provider.yaml
```

**How to organize provider configurations:**

Define all your providers in your main `promptfooconfig.yaml`:

```yaml
providers:
  - id: openai:gpt-4
    config:
      temperature: 0.7
  - id: anthropic:claude-3
    config:
      max_tokens: 1000
```

PR #652 (which referenced this issue) fixed a Windows-specific path issue but didn't implement loading provider configs from separate files.

If you'd like the ability to load provider configurations from external YAML files via command line, please open a new feature request describing your use case.

Since this issue is from March 2024 with no follow-up activity, I'm closing it now. Feel free to comment or re-open if you need clarification.

Thanks!

PR #652, opened 2024-04-09
fix: allow loading of custom provider in windows (#518)

jernkuan:
- windows importing module requires file:// path url, however loading file://customProvider.js assumes that file is actually a yaml configuration


typpo:
Thanks for the fix!

Issue #656, opened 2024-04-11
Telemetry is enabled even when opted out


remixer-dec:
Steps to reproduce:
1) Run
```sh
mkdir empty_dir
cd empty_dir
docker run -it --rm -p 3000:3000 -v ./:/root/.promptfoo -e PROMPTFOO_DISABLE_TELEMETRY=1 -e PROMPTFOO_SHARE_STORE_TYPE=filesystem promptfoo-ui
```
2) Open the ui, put any config and run the evaluation

3) See the docker output
```
Writing environment variables to .env file...
Loaded environment variables:
NODE_VERSION=18.20.0
HOSTNAME=0.0.0.0
YARN_VERSION=1.22.19
SHLVL=2
PORT=3000
HOME=/root
PROMPTFOO_SHARE_STORE_TYPE=filesystem
TERM=xterm
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
NEXT_TELEMETRY_DISABLED=1
PROMPTFOO_DISABLE_TELEMETRY=1
PWD=/app
Starting server...
- ready started server on 0.0.0.0:3000, url: http://localhost:3000
- info Loaded env from /app/.env
- info Loaded env from /app/.env
Listening on port 3000 url: http://0.0.0.0:3000
Using unknown OpenAI chat model: llamacpp
Anonymous telemetry is enabled. For more info, see https://www.promptfoo.dev/docs/configuration/telemetry
```

typpo:
Should be fixed by #658 - thanks to @will-holley 

This appears to just be a logical error that causes the notice to display, but telemetry is not actually sent.

PR #658, opened 2024-04-11
fix: 656

will-holley:
## Summary

Telemetry notice should not render when telemetry is disabled.  

## Motivation

Was browsing the repo and encountered #656.


typpo:
oops, thanks for catching this!

PR #666, opened 2024-04-13
chore: Add checkbox to clear all `Variable`s

sangwoo-joh:
This fix is a suggestion that addresses a perhaps personal inconvenience for me. If you have a better solution, I'd love to hear it and hope you'll consider this positively.

The CSV file I use for testing has a lot of columns, not all of which are actually used in the prompt template, but all of which appear as `Variable` in the viewer. This makes it difficult to keep only the columns I want in `Select`.
This fix adds a checkbox to either deselect all the `Variable`s at once or select them all. I couldn't figure out the right place for it, so I put it in the same `Box` as the column `Select` for now, but if you have a better place, please let me know.

Thanks for the great open source. I always appreciate it!


typpo:
Great change.  I think my preference here would be to create a dialog that allows for selection of columns, in order to not clutter the top space with many options.  What do you think?

sangwoo-joh:
@typpo I hadn't thought of a dialogue, but that sounds like a good idea. I'll give it a try.

sangwoo-joh:
Dear @typpo, I replaced the `Select` panel of columns with a `Dialog` and put the "Clear All Variables" button at the top of it. It seemed to work when I tested it locally. I'm not good at designing, so I picked moderate buttons. I'd appreciate it if you could review this!

sangwoo-joh:
I tried to upload some working images and gifs, but they won't upload due to Github's issue...

src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -243,25 +282,40 @@ export default function ResultsView({
...
+
+              <DialogContent>
+                <List sx={{ m: 1, minWidth: 200, maxWidth: 350 }}>
+                  {columnData.map((column) => (
+                    <ListItem
+                      key={column.value}
+                      value={column.value}
+                    >
+                      <Checkbox checked={selectedColumns.indexOf(column.value) > -1} />
+                      <ListItemText primary={column.label} />
```
typpo: I'm noticing that clicking the checkbox should update the state (right now, you can't actually toggle the checkbox).  Let's also connect the label with the checkbox, so that clicking it toggles the state too

typpo (commented):
This is coming together, thanks for setting up the dialog!  If it's ok with you, I'll do a pass on the design/layout probably tomorrow.  Then I'm happy to merge :)

src/web/nextui/src/app/eval/ResultsView.tsx
sangwoo-joh: Yes, I've noticed too. I'm trying to fix this now. Thank you!

sangwoo-joh: @typpo I've restored the original functionality of the checkbox.
Please let me know if I have additional corrections.
Thank you!

sangwoo-joh:
I rebased these commits to the `main` and then force-pushed them, out of habit. If that's not the policy of this repository, please let me know and next time I'll resolve conflicts.

mldangelo:
Cleaning up old PRs. Thank you @sangwoo-joh!

PR #671, opened 2024-04-14
Enable post-hoc evaluations through defining and using output value in TestSuite

anthonyivn2:
## What this PR is trying to solve
Currently node module evaluate() works by running the values defined in `TestSuite.tests` through the prompts or providers defined in `TestSuite.prompts and TestSuite.providers`, and then evaluating the output through the assertions defined in `TestSuite.assert`. 

This evaluation process works well when we are running offline experiments or tests. However, this process would not allow us to efficiently evaluate LLM systems that are already in production (post-hoc or post-execution evaluations) where the output for each of the inputs/tests are already produced in production, as evaluate() expects to obtain those output values by first running the inputs/tests through a prompt or through a provider defined by the user in `TestSuite` (instead of leveraging the output values already produced in production and stored in a database or file).

IMO we should allow users to define the `output` value for a given `TestCase`, and if it is defined by the user then evaluate() should directly treat that as the output value in `ProviderResponse` and continue running `runEval` without calling `callApi` first. Allowing this would open up the ability for evaluate() to more efficiently evaluate the scenario described above, while taking advantage of all the assertions available today.

## How does this PR solves it
1. Add an `output` field to `TestCase` where user can define the output that they have for a given test case
2. Add a check in `runEval` whether a given test case being handled already contained an `output` value or not. If that is the case then skip calling the `callApi` function. Otherwise continue to run `callApi` as per usual.
3. Continue with the rest of the `runEval` function.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -317,7 +317,7 @@ class Evaluator {
       this.conversations[conversationKey].push({
         prompt: renderedJson || renderedPrompt,
         input: conversationLastInput || renderedJson || renderedPrompt,
-        output: test.output || response.output || '',
+        output: response.output || '',
       });
 
       if (!response.cached) {
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -31,6 +31,7 @@ import type {
   Prompt,
   RunEvalOptions,
   TestSuite,
+  ProviderResponse,
 } from './types';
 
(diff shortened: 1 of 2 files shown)


typpo:
LGTM - maybe we should rename it to `providerOutput`

anthonyivn2:
@typpo I modified the element name from `output` to `providerOutput`

PR #675, opened 2024-04-15
Allow local files to define providerOutput value for TestCase

anthonyivn2:
Continuation of PR #671 by allowing users to define the `providerOutput` value that is available in `TestCase` from local files. I am using `provider_output` as the key/column value to look out for within the local file, and assume that if the column name is that then we can treat it as the `providerOutput` value for `TestCase`


typpo:
LGTM!  Would you mind if we did `__providerOutput`, to match with the existing special column names?

anthonyivn2:
Ok I changed it to `__providerOutput`

typpo:
Thanks!

Issue #682, opened 2024-04-17
Potential bug with `answer-relevance` assertion calculation


anthonyivn2:
I was doing some cross-checks with my own test data when I notice that there might be several bugs within the `answer-relevance` assertion calculation. I am not sure if these are intentional or not:

1. ANSWER_RELEVANCY_GENERATE prompt only uses single curly brackets for the `answer` variable. This means that even though we try to render the value, the answer value won't be populated to the {answer} part of the string
Link: https://github.com/promptfoo/promptfoo/blob/e42a24c00ce1ce6d757c6256c4d53119f79f97ef/src/external/ragas.ts#L10

2. ANSWER_RELEVANCY_GENERATE is being passed to callApi without first being rendered through nunjucks.renderString, meaning all the prompts used for this part of the calculation flow would be the same no matter what is the `answer` value for the given row unit.
Link:
https://github.com/promptfoo/promptfoo/blob/e42a24c00ce1ce6d757c6256c4d53119f79f97ef/src/matchers.ts#L528C9-L528C34

3. Shouldn't the value passed to the matchesAnswerRelevance `input` arg be `test.vars.query` instead of `prompt`?
Link: https://github.com/promptfoo/promptfoo/blob/e42a24c00ce1ce6d757c6256c4d53119f79f97ef/src/assertions.ts#L884

anthonyivn2:
Also I am not sure if the `ANSWER_RELEVANCY_GENERATE` prompt has enough few-shot examples to make it work

anthonyivn2:
Closing the issue as the PR for the fix is merged

PR #683, opened 2024-04-17
Fixing bugs in `Answer-relevance` calculation

anthonyivn2:
Related with #682

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -523,9 +523,12 @@ export async function matchesAnswerRelevance(
   const candidateQuestions: string[] = [];
   for (let i = 0; i < 3; i++) {
     // TODO(ian): Parallelize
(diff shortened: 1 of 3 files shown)


src/matchers.ts
```diff
@@ -523,9 +523,15 @@ export async function matchesAnswerRelevance(
   const candidateQuestions: string[] = [];
   for (let i = 0; i < 3; i++) {
     // TODO(ian): Parallelize
+    const promptText = nunjucks.renderString(ANSWER_RELEVANCY_GENERATE['content'],
```
typpo: Looks like we can simplify `ANSWER_RELEVANCY_GENERATE` to just be a string

typpo (approved):
Thanks for these fixes!

PR #685, opened 2024-04-17
fix: Add tool calls to response from azure openai

CamdenClark:
This PR makes it so tool calls in empty content messages are output with the response from azure openai


typpo:
Thanks for the fix!

PR #695, opened 2024-04-20
[Docs] Add llama3 example to ollama docs

chanonroy:
Ollama now supports llama3 - https://ollama.com/blog/llama3


typpo:
Thanks!  It looks like ollama made `chat` style the default, so I updated the other examples too.

PR #700, opened 2024-04-22
feat: `eval -n` arg for running the first n test cases

typpo:


alexclaydon:
Hello, I'm trying to use `eval -n` where my tests are contained in a `.csv` file, but it doesn't seem to behave as expected: regardless of what integer is passed as the value, only the first row of the `.csv` file is processed.

typpo:
Thanks, someone else also caught this earlier today - there's a fix on `main` and will be in the next version!

typpo:
Yes definitely, please create an issue so we can track and implement!

PR #703, opened 2024-04-23
Fix concurrent cache request behaviour

chrisprice:
This fixes an issue I noticed whereby the cached token counts didn't seem to be right. I tracked it down to concurrent requests to the cache triggering multiple underlying fetches. Swapping over to using the `cache.wrap` method provided by `cache-manager` (for exactly this use case) fixes the issue.

The implementation is a little convoluted due to maintaining the cached value type and existing logging behaviour. My assumption was that this would be the preference but I'm happy to change this if preferred.


typpo:
Great catch, thank you for the fix @chrisprice!

PR #704, opened 2024-04-24
better DEFAULT_GRADING_PROMPT

rocklau:
sometimes quotes will be disappeared.


typpo:
@rocklau which grading provider are you using that this is necessary?

rocklau:
Yeah, may be not necessary

PR #706, opened 2024-04-24
Make errors uncached in Python completion.

grahl:
This change only caches a result if the response does not contain an error. This helps in rerunning failing executions quickly.


rocklau:
same issue

typpo:
Thanks @grahl!

PR #728, opened 2024-04-29
Fixes to Python docs

jamesbraza:
- Fixes missing syntax (e.g. closing paren, colon)
- Removes extra `else`
- Adds missing imports

### DIFF
--- a/site/docs/configuration/expected-outputs/python.md
+++ b/site/docs/configuration/expected-outputs/python.md
@@ -51,7 +51,9 @@ assert:
 A `context` object is available in the Python function. Here is its type definition:
 
 ```py
-class AssertContext:
+from typing import Dict, TypedDict, Union
+
+class AssertContext(TypedDict):
     prompt: str
     vars: Dict[str, Union[str, object]]
 ```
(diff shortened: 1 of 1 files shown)


site/docs/configuration/expected-outputs/python.md
```diff
@@ -39,19 +39,20 @@ assert:
...
 
 ## Using test context
 
 A `context` object is available in the Python function. Here is its type definition:
 
 ```py
-class AssertContext:
+from typing import Dict, TypedDict, Union
+
+class AssertContext(TypedDict):
```
jamesbraza, line 55: Shouldn't `vars` allow for `dict` too? Was that what was meant by `object`?

jamesbraza: Also, it looks like `test` (also a `dict`) should be here too

typpo: Yeah, I think you're right on both counts

jamesbraza: Okay, implemented them, this is ready for review now

typpo:
Thanks James!

PR #731, opened 2024-04-29
Document `python:` prefix when loading assertions in CSV

efung:
When loading test assertions from CSV, [you can specify Python functions or files](https://github.com/promptfoo/promptfoo/blob/9fce9a875065653138882694ca043a5ebed94102/src/csv.ts#L70):

Mention that in the documentation.


typpo:
Thanks Eric!

PR #739, opened 2024-04-30
Update openai package to 3.48.5

matteodepalo:
Hi 👋 
We're using promptfoo with the new assistants API and it seems like it lacks some features so I've updated the package to support the new tool type `file_search` and added a few configuration options. We also need a way to pass custom headers because we use a proxy for our openai calls and we need to pass `Cache-Control: none`. 

Let me know if you need anything else!


typpo:
Nice change!

PR #742, opened 2024-04-30
feat: promptfoo eval --filter-failing outputFile.json

mikkoh:
Makes it so that you can iterate more quickly by simply running failing tests given an output file.

## Example Flow

First run: `promptfoo eval --output result.json `
Screenshot 2024-04-30 at 5 17 28 PM

Next run: `promptfoo eval --failing result.json`
Screenshot 2024-04-30 at 5 18 41 PM

### DIFF
--- a/src/commands/eval/filterFailingTests.ts
+++ b/src/commands/eval/filterFailingTests.ts
@@ -1,0 +1,21 @@
+import {TestSuite} from "../../types";
+import {readOutput, resultIsForTestCase} from "../../util";
+
+type Tests = NonNullable<TestSuite['tests']>;
+
+export async function filterFailingTests(testSuite: TestSuite, outputPath: string): Promise<Tests> {
+  if (!testSuite.tests) {
+    return [];
+  }
+
+  const {results} = await readOutput(outputPath);
+  const failingResults = results.results.filter((result) => !result.success);
+
+  if (failingResults.length === 0) {
+    return [];
+  }
+
+  return [...testSuite.tests].filter((test) => {
+    return failingResults.some((result) => resultIsForTestCase(result, test));
+  }) as Tests;
+}
--- a/src/main.ts
+++ b/src/main.ts
@@ -552,8 +552,9 @@ async function main() {
       'Run providers interactively, one at a time',
       defaultConfig?.evaluateOptions?.interactiveProviders,
     )
-    .option('-n, --first-n <number>', 'Only run the first N tests')
-    .option('--pattern <pattern>', 'Only run tests whose description matches the regular expression pattern')
+    .option('-n, --filter-first-n <number>', 'Only run the first N tests')
+    .option('--filter-pattern <pattern>', 'Only run tests whose description matches the regular expression pattern')
+    .option('--filter-failing <path>', 'Path to json output file')
     .action(async (cmdObj: CommandLineOptions & Command) => {
       setupEnv(cmdObj.envFile);
       let config: Partial<UnifiedConfig> | undefined = undefined;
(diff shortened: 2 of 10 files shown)


src/commands/eval/filterFailingTests.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function filterFailingTests(testSuite: TestSuite, outputPath: string): Promise<Tests> {
+  if (!testSuite.tests) {
+    return [];
+  }
+
+  const {results} = await readOutput(outputPath);
+
+  if (results.version < 2) {
+    throw new Error(`Unsupported output version: ${results.version}`);
```
mikkoh: I put this here just cause I'm not familiar with the differences between versions

src/main.ts
```diff
@@ -585,9 +586,10 @@ async function main() {
           );
         }
 
-        testSuite.tests = filterTests(testSuite.tests, {
+        testSuite.tests = await filterTests(testSuite, {
           firstN: cmdObj.firstN,
           pattern: cmdObj.pattern,
+          failing: cmdObj.failing,
```
mikkoh: I'm wondering if these command line params should be changed to be something like:
```
filterFirstN
filterPattern
filterFailing
```

typpo: I support this

src/commands/eval/filterFailingTests.ts
typpo: Should be safe to remove this check :+1:

src/util.ts
```diff
@@ -1228,3 +1244,26 @@ export function getStandaloneEvals(): StandaloneEval[] {
   });
   return flatResults;
 }
+
+export function providerToIdentifier(provider: TestCase['provider']): string | undefined {
```
typpo, line 1248: Really appreciate you adding these helper functions.  I'm aware that ApiProvider vs ProviderOptions and similar variations are some of the ugliest parts of the code :(

typpo (approved):
LGTM - responses to your comments inline.  Let me know if you'd like to e.g. rename the params to `filter...` in this PR or a separate one.

src/main.ts
mikkoh: Ok... I'll make the change in this PR

src/util.ts
mikkoh: No prob. I love the flexibility of `promptfoo`. You can use it many ways. But it also does brings on complexity. I suspect little utility functions to simplify logic can go a long way.

src/main.ts
```diff
@@ -552,8 +552,9 @@ async function main() {
       'Run providers interactively, one at a time',
       defaultConfig?.evaluateOptions?.interactiveProviders,
     )
-    .option('-n, --first-n <number>', 'Only run the first N tests')
-    .option('--pattern <pattern>', 'Only run tests whose description matches the regular expression pattern')
+    .option('-n, --filter-first-n <number>', 'Only run the first N tests')
```
mikkoh, line 555: This will be a breaking change so will need to do a version bump for the next release. Unsure if theres a mechanism you use eg Github labels to determine if a version bump is needed for the next release

Should we keep `-n`? Should the other two filters have short forms also?

typpo: My selfish preference is to keep `-n` because it feels familiar, like `head -n` :).  Open to other short forms but I don't think it's necessary.  If we find ourselves getting tired of typing everything out let's add it separately.

I do version bumps manually - since we're pre-1.0 I've included breaking changes in minor versions, but generally trying to avoid them.  I think this is acceptable though.  I'll merge with `feat!` breaking change notation and make a note in the release notes.

PR #745, opened 2024-05-01
feat: add support for `--var name=value` cli option

typpo:
Enables setting defaultTest vars from the CLI:

```
promptfoo eval --var topic='spaghetti recipes' --var flavor=garlic
```

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -49,7 +49,7 @@ import type {
 } from './types';
 import { generateTable } from './table';
 import { createShareableUrl } from './share';
-import {filterTests} from './commands/eval/filterTests';
+import { filterTests } from './commands/eval/filterTests';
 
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -49,7 +49,7 @@ import type {
 } from './types';
 import { generateTable } from './table';
 import { createShareableUrl } from './share';
-import {filterTests} from './commands/eval/filterTests';
+import { filterTests } from './commands/eval/filterTests';
```
mikkoh, line 52: Sorry about this.

Could add `bracketSpacing: true` to `.prettierrc.yaml` I use Prettier format on save. Surprised it didn't pick this up.

typpo: :+1:  Done in ba8da4ec

PR #756, opened 2024-05-02
Resolve JS assertion paths relative to configuration file

Arkham:
Fix bug when resolving the assertion file path :)


typpo:
Nice catch!  Thanks for the fix

Issue #757, opened 2024-05-02
Feature Request: Assertion Sets


thomascleberg:
There are meaningful test cases in which at least one of as set of assertions needs to pass, but not all of them need to pass. 

it would be useful to be able to define an assertion that defines sub-assertions and the pass criteria for the assertion set.

Each of the assertions in the set could calculate and produce metrics like other assertions, but would be individually exempted from having a failing result cause a failure for the record.

Example sketch:
```yaml
assert:
    - type: context-faithfulness
      threshold: 0.9
      metric: Groundedness
    - type: assertion-set
      value: >= 1 pass
      assert: 
          - <assertion A>
          - <assertion B>
          - ...
```

mikkoh:
I can take a look at implementing this. I haven't worked on assertions yet. Would love to learn that part of the codebase

typpo:
I would gladly accept a PR!

PR #765, opened 2024-05-07
feat: assert-set

mikkoh:
Fixes https://github.com/promptfoo/promptfoo/issues/757

## What

This PR implements a new feature `assert-set`. It groups assertions.

Example:
```
assert:
    - type: context-faithfulness
      threshold: 0.9
      metric: Groundedness
    - type: assert-set
      assert:
        - type: equals
           value: 'Hello, World!'
        - type: contains
           value: 'World'
```

The assertions within an `assert-set` determine of the success/failure of the entire `assert-set`.

A threshold can be used if not all assertions are required to pass:
```
assert:
    - type: context-faithfulness
      threshold: 0.9
      metric: Groundedness
    - type: assert-set
      threshold: 0.5
      assert:
        - type: equals
           value: 'Hello, World!'
        - type: contains
           value: 'World'
```

In the above example only one of the assertions needs to pass for the entire `assert-set` to succeed.

`weight` and `metric` can also be used on an `assert-set`. This function the same as any other assertion. They are used for `TestCase` results and don't modify the execution of child assertions of the `assert-set`.

## How

An aim for this PR was to keep the running of assertions as simple as possible and to ensure we can have maximum concurrency when running assertions.

This aim keeps the running of assertions as a flat array. The assertions from the `assert-set` are flattened to be run with all other assertions.

To group these assertions a class `AssertionsResult` is introduced which aggregates the results of the assert set. It also isolates result calculations to be outside of the running of assertions hopefully making the code be more testable in isolation.

For simplicity (for now) `assert-sets` cannot contain `assert-sets`.

### DIFF
--- a/src/assertions.ts
+++ b/src/assertions.ts
@@ -33,6 +33,7 @@ import { runPython, runPythonCode } from './python/wrapper';
 import { importModule } from './esm';
 
 import type { Assertion, AssertionType, GradingResult, AtomicTestCase, ApiProvider } from './types';
+import { AssertionsResult } from './assertions/AssertionsResult';
 
 const ASSERTIONS_MAX_CONCURRENCY = process.env.PROMPTFOO_ASSERTIONS_MAX_CONCURRENCY
   ? parseInt(process.env.PROMPTFOO_ASSERTIONS_MAX_CONCURRENCY, 10)
@@ -104,28 +105,18 @@ export async function runAssertions({
   logProbs?: number[];
   cost?: number;
 }): Promise<GradingResult> {
-  const tokensUsed = {
-    total: 0,
-    prompt: 0,
-    completion: 0,
-  };
   if (!test.assert || test.assert.length < 1) {
-    return { pass: true, score: 1, reason: 'No assertions', tokensUsed, assertion: null };
+    return AssertionsResult.noAssertsResult();
   }
-  let totalScore = 0;
-  let totalWeight = 0;
-  let allPass = true;
-  let failedReason = '';
-  const componentResults: GradingResult[] = [];
-  const namedScores: Record<string, number> = {};
+
+  const assertResults = new AssertionsResult();
 
   await async.forEachOfLimit(test.assert, ASSERTIONS_MAX_CONCURRENCY, async (assertion, index) => {
     if (assertion.type.startsWith('select-')) {
       // Select-type assertions are handled separately because they depend on multiple outputs.
       return;
     }
-    const weight = assertion.weight ?? 1;
-    totalWeight += weight;
+
     const result = await runAssertion({
       prompt,
       provider,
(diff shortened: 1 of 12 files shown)


src/assertions.ts
```diff
@@ -104,77 +105,89 @@ export async function runAssertions({
...
+
+  const mainAssertResult = new AssertionsResult({
+    threshold: test.threshold,
+  });
+  const subAssertResults: AssertionsResult[] = [];
+  const asserts: {
+    assertion: Assertion;
+    assertResult: AssertionsResult;
+    index: number;
+  }[] = test.assert
```
mikkoh, line 120: This is all flattened so we can support maximum concurrency when the assertions are executed.

By making this array flat we are more likely to fill `ASSERTIONS_MAX_CONCURRENCY` vs partially filling it.

src/assertions.ts
```diff
@@ -104,77 +105,89 @@ export async function runAssertions({
...
   cost?: number;
 }): Promise<GradingResult> {
-  const tokensUsed = {
-    total: 0,
-    prompt: 0,
-    completion: 0,
-  };
   if (!test.assert || test.assert.length < 1) {
-    return { pass: true, score: 1, reason: 'No assertions', tokensUsed, assertion: null };
+    return AssertionsResult.noAssertsResult();
```
mikkoh, line 109: This could be a constant but if for some reason the results are mutated in some way in the future this felt safer.

src/assertions/AssertionsResult.ts
```diff
@@ -0,0 +1,121 @@
...
+export class AssertionsResult {
+  static noAssertsResult(): GradingResult {
+    return {
+      pass: true,
+      score: 1,
+      reason: 'No assertions',
+      tokensUsed: { ...DEFAULT_TOKENS_USED },
+      assertion: null,
+    };
+  }
```
mikkoh, line 23: I think this static method could be removed. However if for some reason the results are mutated in some way in the future this felt safer.

src/assertions/validateAssertions.ts
```diff
@@ -0,0 +1,36 @@
...
+export class AssertValiationError extends Error {
+  constructor(message: string, testCase: TestCase) {
+    const testCaseDescription = testCase.description || JSON.stringify(testCase);
+
+    super(`${message} in:\n${testCaseDescription}`);
+    this.name = 'AssertValiationError';
+  }
+}
+
+export function validateAssertions(tests: TestCase<Record<string, string | object | string[]>>[]) {
```
mikkoh, line 12: The main reason this function was added in was so we can enforce an `assert-set` not containing a child `assert-set`. This was done for simplicity and the flattening of assertions could be made recursive.

I'm not sure if people will want child `assert-set` or grandchild `assert-set`. Better to just keep it simple.

It could also be a nicety for users to have other assertions be validated.

src/evaluator.ts
```diff
@@ -903,7 +904,7 @@ class Evaluator {
 
     for (let index = 0; index < table.body.length; index++) {
       const row = table.body[index];
-      const compareAssertion = row.test.assert?.find((a) => a.type === 'select-best');
+      const compareAssertion = row.test.assert?.find((a) => a.type === 'select-best') as Assertion;
```
mikkoh, line 907: There's a sniff here that we should have a `isSelectBest(assertion: Assertion): assertion is SelectBestAssertion` method but that method would feel too narrow. Likely when more `select-*` assertions are added that function

typpo (approved):
This is an excellent change!  I also really appreciate how you've cleaned up `AssertionResult` usage.

Would you be willing to update the docs in https://github.com/promptfoo/promptfoo/blob/main/site/docs/configuration/expected-outputs/index.md (or create a subpage if you prefer)?  Could also do it separately.

mikkoh:
Of course

mikkoh:
A bit of a last minute change. In code I had `assertion-set`. I realized while documenting the feature that was writing `assert-set`. I changed the code to use `assert-set`. I think it's cleaner.

typpo:
Thanks @mikkoh!  I'll cut a release with this change shortly.

PR #774, opened 2024-05-12
Check runtime type of `getPrompt`, stringify the result

albertpurnama:
# Problem

image
Langfuse integration fails when prompt is `chat` type. It produces error `prompt.trim is not a function`

Langfuse has 2 different types of return values. `string` and `ChatMessage[]` when we're returning the output of `getLangchainPrompt(()`

We currently assume that `getPrompt` always returns `string`. But in reality, it can return either `string` or `ChatMessage[]`.

The solution that I propose here is simple, to stringify the output of `getLangchainPrompt` if the output is not string. However, the ideal situation is probably to let the prompt definition in the config to include the type of the prompt 

For example:
- instead of `langfuse://test-prompt:1`, we do `langfuse://test-prompt:1:chat` to indicate that this is a chat prompt, and do proper configuration.

# Solution

Stringify the result of `getLangchainPrompt`.

### DIFF
--- a/src/integrations/langfuse.ts
+++ b/src/integrations/langfuse.ts
@@ -1,17 +1,18 @@
 const langfuseParams = {
-    publicKey: process.env.LANGFUSE_PUBLIC_KEY,
-    secretKey: process.env.LANGFUSE_SECRET_KEY,
-    baseUrl: process.env.LANGFUSE_HOST,
-}
+  publicKey: process.env.LANGFUSE_PUBLIC_KEY,
+  secretKey: process.env.LANGFUSE_SECRET_KEY,
+  baseUrl: process.env.LANGFUSE_HOST,
+};
 
 import { Langfuse } from 'langfuse';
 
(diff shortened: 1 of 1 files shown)


src/integrations/langfuse.ts
```diff
@@ -1,17 +1,18 @@
...
-    version?: number,
-): Promise<string> {
-    const prompt = await langfuse.getPrompt(id, version);
-    return prompt.getLangchainPrompt();
+export async function getPrompt(id: string, version?: number): Promise<string> {
+  const prompt = await langfuse.getPrompt(id, version);
+  const outputPrompt = prompt.getLangchainPrompt();
+  if (typeof outputPrompt !== 'string') {
+    return JSON.stringify(outputPrompt);
+  }
```
albertpurnama, line 16: This is the gist of the change, we want to stringify the result if it's not a `string` type.

typpo:
Thanks for the fix!

PR #779, opened 2024-05-14
Langfuse need to compile variables

albertpurnama:
## The problem
Continuing on https://github.com/promptfoo/promptfoo/pull/774, I just realized that the langfuse integration doesn't actually compile the variables inside the prompt. (Link to [docs](https://www.promptfoo.dev/docs/integrations/langfuse))

This PR makes sure that the variables are compiled to langfuse prompt client.

## Additional things

- Allow user to use "latest" string for grabbing the latest version of their prompt.
- Allow user to specify the prompt type. -> I think we don't need this, but for type safety purposes, maybe we should? I also made sure that the default still works by auto assigning it to `text` type if user doesn't specify it.


typpo:
Thanks @albertpurnama!

PR #780, opened 2024-05-15
feat(webui): add --filter-description option to `promptfoo view`

typpo:


a-gasior:
Thanks @typpo This is a game changer for me!

PR #805, opened 2024-05-21
feat: support for deep equality check in equals assertion

### LINKED ISSUE #799: Request: extending the `equals` assertion to work with JSON
Hi, thank you very much for the wonderful software!

I am working on a task to output information contained in images in JSON format using a multimodal LLM. I would like to evaluate whether the output JSON matches the expected JSON.

Currently, I am placing test values in `vars`.

```yaml
- vars:
    ...
    expected: {}  // expected json here
  assert: 
    - type: python
      value: context["vars"]["expected"] == json.loads(output)
```

Preferably, I'd like to write it like this:

```yaml
- vars:
    ...
  assert:
    - type: json-equals
      value: {}  // expected json or target file here
```

Here are the potential benefits of implementing this:

1. **Easier to use for promptfoo beginners**
When I first tried to handle this using `contains-json` and `is-json`, it took me a little while to realize it wasn't working as expected. However, this can also be resolved through better documentation.

2. **Ability to specify files**
I'm planning to using slightly complex JSON as test values. In such cases, writing JSON directly in the YAML can be cumbersome, so having the ability to specify a file would be convenient.

If there are no issues with this proposal, I'm thinking of contributing it myself!

mikanfactory:
resolve #799

# What
We can now compare JSON output in the following way.

```yaml
assert:
  - type: equals
    value: {"key1": "value1", "key2": "value2"}  // expected json
```

or 

```yaml
assert:
  - type: equals
    value: "file://expected.json" // target file here
```

# How
I use deep equal comparison when an object is passed in with an equals assertion. There are several options for implementing deep equal.

1. Use the functions provided by the official Node.js
2. Use a third party library function (like [lodash](https://github.com/lodash/lodash), [node-deep-equal](https://github.com/inspect-js/node-deep-equal), ..etc)
3. Implement a custom deep equal function

I chose option 1 because option 3 seems difficult to operate accurately, and option 2 is often no longer maintained. And Node.js provide [assert.deepStrictEqual](https://nodejs.org/api/assert.html#assertdeepstrictequalactual-expected-message) and [util.isDeepStrictEqual](https://nodejs.org/api/util.html#utilisdeepstrictequalval1-val2). I implement the latter because the former throws an exception on failure.

### DIFF
--- a/src/assertions.ts
+++ b/src/assertions.ts
@@ -1,6 +1,7 @@
 import fs from 'fs';
 import path from 'path';
 import { execFile, spawn } from 'child_process';
+import util from 'node:util';
 
 import async from 'async';
 import rouge from 'rouge';
@@ -294,13 +295,19 @@ export async function runAssertion({
   }
 
   if (baseType === 'equals') {
-    pass = (renderedValue == outputString) !== inverse;
+    if (typeof renderedValue === 'object') {
+      pass = util.isDeepStrictEqual(renderedValue, JSON.parse(outputString)) !== inverse;
+      renderedValue = JSON.stringify(renderedValue);
+    } else {
+      pass = (renderedValue == outputString) !== inverse;
+    }
+
     return {
       pass,
       score: pass ? 1 : 0,
       reason: pass
-        ? 'Assertion passed'
-        : `Expected output "${renderedValue}" to ${inverse ? 'not ' : ''}equal "${outputString}"`,
+          ? 'Assertion passed'
+          : `Expected output "${renderedValue}" to ${inverse ? 'not ' : ''}equal "${outputString}"`,
       assertion,
     };
   }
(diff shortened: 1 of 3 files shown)


src/assertions.ts
```diff
@@ -294,7 +295,13 @@ export async function runAssertion({
   }
 
   if (baseType === 'equals') {
-    pass = (renderedValue == outputString) !== inverse;
+    if (typeof renderedValue === 'object') {
+      pass = util.isDeepStrictEqual(renderedValue, JSON.parse(outputString)) !== inverse;
+      renderedValue = JSON.stringify(renderedValue);
```
mikanfactory, line 300: I'm converting `renderedValue` to a string for the error message when it fails.

typpo:
Nice change!

PR #806, opened 2024-05-22
PR: Restore ProviderResponse class name

alexclaydon:
When using the provided Python types for custom provider, noticed that the class name for ProviderResponse was missing.


typpo:
Thanks!

Issue #807, opened 2024-05-23
Set max dimensions for images rendered with markdown


chrismaltais:
Trying to use promptfoo as an eval platform for visual language models.

I'm using markdown to render the images, which works okay! One problem is if there's a large discrepancy in image size:

<img width="1917" alt="Screenshot 2024-05-22 at 7 50 20 PM" src="https://github.com/promptfoo/promptfoo/assets/17098959/7828f052-77bd-452f-851a-97c14cf9750e">

Wondering if we can set a max image size?

### Potential Solution
1. Custom component to handle image resizing
2. Wrap it in a `MarkdownRenderer` component
3. Use `MarkdownRenderer` instead of `ReactMarkdown` component

<details>
  <summary>Example code</summary>
  
  ```typescript
// components/CustomImage.tsx
import Image from 'next/image';
import React from 'react';

interface CustomImageProps {
  src: string;
  alt: string;
  title?: string;
}

const CustomImage: React.FC<CustomImageProps> = ({ src, alt, title }) => {
  const width = 300; // Adjust this value as needed
  return (
    <Image 
      src={src} 
      alt={alt} 
      title={title} 
      width={width} 
      height="auto"
      style={{ maxWidth: '100%', height: 'auto' }} // Ensure responsiveness
    />
  );
};

export default CustomImage;
```

```typescript
// components/MarkdownRenderer.tsx
import React from 'react';
import ReactMarkdown from 'react-markdown';
import remarkGfm from 'remark-gfm';
import CustomImage from './CustomImage';

interface MarkdownRendererProps {
  content: string;
}

const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) => {
  return (
    <ReactMarkdown
      remarkPlugins={[remarkGfm]}
      components={{
        img: ({ node, ...props }) => <CustomImage {...props} />
      }}
    >
      {content}
    </ReactMarkdown>
  );
};

export default MarkdownRenderer;
```

```typescript
import MarkdownRenderer from '../components/MarkdownRenderer';

const markdownContent = `
  # Hello World

  ![Alt text](https://via.placeholder.com/600x400)
`;

...
<MarkdownRenderer content={markdownContent} />
...
```
</details>

typpo:
Makes sense!  Given that you've already outlined the solution, would you like to PR it, or want me to go ahead?

chrismaltais:
I'd love to give it a shot @typpo :)

typpo:
[...]

PR #812, opened 2024-05-23
Enforce max image size when rendering images in markdown

chrismaltais:
Closes: https://github.com/promptfoo/promptfoo/issues/807

The purpose of this PR is to limit the max width of an image so that we don't run into the problem in the issue above.

We do this by creating a component to render the image, and a wrapper component around the `ReactMarkdown` component.

---

I wasn't able to figure out a way to easily test this locally, I don't have an API key for any of the default providers, nor do I think I have the storage correctly configured 🤔. @typpo would you be willing to lend a hand to test these changes? I'm not sure the 400px width is the right choice 😅 

Feel free to make changes to this PR as you see fit!

### DIFF
--- a/src/web/nextui/src/app/components/MarkdownRenderer.tsx
+++ b/src/web/nextui/src/app/components/MarkdownRenderer.tsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import ReactMarkdown from 'react-markdown';
+import CustomImage from './CustomImage';
+
+interface MarkdownRendererProps {
+  content: string;
+}
+
+const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) => {
+  return (
+    <ReactMarkdown
+      components={{
+        img: ({ node, ...props }) => (
+          <CustomImage src={props.src || ''} alt={props.alt || ''} {...props} />
+        ),
+      }}
+    >
+      {content}
+    </ReactMarkdown>
+  );
+};
+
+export default MarkdownRenderer;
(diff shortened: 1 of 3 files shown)


src/web/nextui/src/app/components/MarkdownRenderer.tsx
```diff
@@ -0,0 +1,23 @@
...
+  content: string;
+}
+
+const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) => {
+  return (
+    <ReactMarkdown
+      components={{
+        img: ({ node, ...props }) => (
+          <CustomImage src={props.src || ''} alt={props.alt || ''} {...props} />
+        ),
```
chrismaltais, line 15: This prop is used to override the default renderers for different types of Markdown nodes. In this case, it's overriding the renderer for image nodes (img). For each image node, it's rendering a `CustomImage` component instead of the default `img` element.

typpo:
Thanks Chris!

I might be misunderstanding the issue, but I think the images in the results table already have a max dimension of 256px: https://github.com/promptfoo/promptfoo/blob/main/src/web/nextui/src/app/eval/ResultsTable.css#L118-L119.  I haven't been able to get it to look like the screenshot you provided in the issue.  

One problem with the current approach of this PR is that it will mess up the aspect ratio of the image:
image
(kind of hard to tell, but these are stretched wide)

Would it be reasonable to limit the dimensions to something relatively small, e.g. 256px, and then make it so that clicking the image shows the whole thing?  I realized I had some image zoom/lightbox code that was not compatible with markdown. 
 Just fixed that here: https://github.com/promptfoo/promptfoo/pull/816

chrismaltais:
Hey @typpo ! Yep, that works! After your changes, I haven't had a problem with large images in the original issue.

Thanks for the fix! I'll close this PR.

PR #813, opened 2024-05-24
feat: Allow functions in renderVarsInObject

tomquist:
This allows using functions to render vars in function calls where string interpolation doesn't work.

Example:
```javascript
{
  providers: {
    id: "openai:chat:gpt-3.5-turbo-0613",
    config: {
      tools: [
        {
          type: "function",
          function: {
            name: "get_temperature",
            parameters: {
              type: "object",
                properties: {
                  location: {
                    type: "string",
                    description: "The city and state, e.g. San Francisco, CA"
                  },
                  unit: {
                    type: "string",
                    enum: (vars) => vars.units,
                  }
                },
                required: ["location"]
              }
            }
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/site/docs/providers/openai.md
+++ b/site/docs/providers/openai.md
@@ -461,11 +461,47 @@ tests:
         value: NYC
 ```
 
+:::tip
+Functions can use variables from test cases:
+
+```js
+{
+  type: "function",
+  function: {
+    description: "Get temperature in {{city}}"
+    // ...
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


typpo:
Thanks!  Looks great, just made a minor change to make it pass `vars` within an object:

```
                    enum: ({ vars }) => vars.units,
```
This is because someone is going to ask me to pass other context info one day, and this makes it consistent with how args are passed to custom prompts/providers.

PR #820, opened 2024-05-25
chore: add node 22 to ci

typpo:


jacobdr:
Given [node16 is EOL](https://nodejs.org/en/blog/announcements/nodejs16-eol) it will probably simplify maintenance to remove it from the matrix

typpo:
I agree with that.  Currently having trouble getting better-sqlite to work with node 22 🤔

typpo:
It is not intentional. I agree we should add that rule!

jacobdr:
Oh man debugging this was a complete rabbit hole. After forking `better-sqlite3` and being unable to reproduce there, and then getting `better-sqlite3` to install in our matrix via a self-contained `npm install better-sqlite3`, I finally figured out the root cause is in our `prepare` script. I'm not 100% sure as to the ultimate "why", but the mismatching versions of `better-sqlite3` is probably a pretty good culprit.

Ultimately, I think getting rid of the whole additional `package.json` and `package-lock.json` in `src/web/nexui` is the best short term solution (merging those deps into the main `package.json` and regenerating the lockfile), and long-term we can use [workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces) as needed. I'm just not sure the complexity of workspaces is warranted yet, hence why I think the best course of action is the simplest -- let's just have one source of truth for dependencies.

The `prepare` script is a bit weird anyway, since we require the `NEXT_PUBLIC_SUPABASE_URL` and `NEXT_PUBLIC_SUPABASE_ANON_KEY` to be supplied, which is its own friction to local development and forking of this repository. I would propose we actually just get rid of it and just rely on the UI getting built in the `build` section, which gets run in CI anyway, unless it's there for some other reason.

The smoking gun evidence for all this can be found here: https://github.com/jacobdr/promptfoo/pull/1/files

jacobdr:
I opened https://github.com/promptfoo/promptfoo/pull/836 to supersede this PR

PR #829, opened 2024-05-29
fix: The ‘defaultTest’ option has no effect during evaluation.

big-camel:
The document at ‘https://promptfoo.dev/docs/configuration/expected-outputs/model-graded’ describes how to configure the ‘defaultTest’ option, but it has not been posted yet.


typpo:
Thank you for fixing this!

PR #830, opened 2024-05-29
feat(openai): Support function/tool callbacks

typpo:


src/providers/openai.ts
```diff
@@ -477,6 +486,31 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
         (logProbObj: { token: string; logprob: number }) => logProbObj.logprob,
       );
 
+      // Handle function tool callbacks
+      const functionCalls = message.function_call ? [message.function_call] || message.tool_calls;
```
scottfavre:
```suggestion
      const functionCalls = message.function_call ? [message.function_call] : message.tool_calls;
```

scottfavre (approved):
code looks good.

looks like there isn't much tool/function test coverage, is this a good time to add that?

Issue #827, opened 2024-05-28
Individual external Python assertion that raises exception results in no grading result in JSON output


efung:
If an individual Python assertion raises an exception in a test case, I would like to be able to see it as part of the `componentResults` in the JSON output. At the moment, it causes all of the grading results to be suppressed.

I'm using promptfoo@latest (0.60.0) and the example in `examples/python-assert-external`. 

First, I add another external assert file, `fails.py` with the following contents:

```python
def get_assert(output, context):
    return True
    #raise Exception("This is a forced failure")
```

Then, I modify `promptfooconfig.yaml` by using only one `vars` (yellow fruits) and adding `fails.py` as an additional assertion. So the file looks like this:

```yaml
prompts:
  - 'Tell me about {{topic}} in three words'
providers:
  - openai:gpt-4-1106-preview
tests:
  - vars:
      topic: yellow fruits
    assert:
      - type: python
        value: file://fails.py
      - type: python
        value: file://assert.py
```

Then, I run it:

`npx promptfoo@latest eval --config ./promptfooconfig.yaml --output okay.json`

<details><summary>okay.json output</summary>

```json
{
  "results": {
    "version": 2,
    "timestamp": "2024-05-28T18:51:25.690Z",
    "results": [
      {
        "provider": {
          "id": "openai:gpt-4-1106-preview"
        },
        "prompt": {
          "raw": "Tell me about yellow fruits in three words",
          "label": "Tell me about {{topic}} in three words"
        },
        "vars": {
          "topic": "yellow fruits"
        },
        "response": {
          "output": "Bright, nutritious, sweet.",
          "tokenUsage": {
            "cached": 21,
            "total": 21
          },
          "cached": true,
          "cost": 0.00033
        },
        "success": false,
        "score": 0.75,
        "namedScores": {},
        "latencyMs": 4,
        "cost": 0.00033,
        "error": "Contains banana",
        "gradingResult": {
          "pass": false,
          "score": 0.75,
          "reason": "Contains banana",
          "namedScores": {},
          "tokensUsed": {
            "total": 0,
            "prompt": 0,
            "completion": 0
          },
          "componentResults": [
            {
              "pass": true,
              "score": 1,
              "reason": "Assertion passed",
              "assertion": {
                "type": "python",
                "value": "file://fails.py"
              }
            },
            {
              "pass": false,
              "score": 0.5,
              "reason": "Contains banana",
              "assertion": {
                "type": "python",
                "value": "file://assert.py"
[...]

PR #833, opened 2024-05-30
fix: Improve Error Handling in Python Script Execution

mldangelo:
**Summary**

This PR addresses issue [#827](https://github.com/promptfoo/promptfoo/issues/827) which involves suppressed grading results when an individual Python assertion raises an exception during test execution. The changes ensure that exceptions are properly captured and included in the `componentResults` of the JSON output, providing clearer feedback on test failures. In the future, we should consider treating Python exceptions separately from grading result failures as they are different classes of errors.

**Test Instructions**

To demonstrate the issue, an additional external assert file `fails.py` was created with the following contents:
```python
def get_assert(output, context):
    raise Exception("This is a forced failure")
```

This file was used in the `promptfooconfig.yaml` for testing purposes:
```yaml
prompts:
  - 'Tell me about {{topic}} in three words'
providers:
  - openai:gpt-4-1106-preview
tests:
  - vars:
      topic: yellow fruits
    assert:
      - type: python
        value: file://fails.py
      - type: python
        value: file://assert.py
```

Running the evaluation with `npm i` and then `npx [path/to/promptfoo] eval --config ./promptfooconfig.yaml --output exception.json` now correctly includes the exception details in the JSON output.

Thank you!

### DIFF
--- a/src/python/wrapper.ts
+++ b/src/python/wrapper.ts
@@ -35,7 +35,16 @@ export async function runPython(
   try {
     await fs.writeFile(tempJsonPath, safeJsonStringify(args));
 
-    const results = await PythonShell.run('wrapper.py', pythonOptions);
+    let results;
+    try {
+      results = await PythonShell.run('wrapper.py', pythonOptions);
+    } catch (error) {
+      return {
+        pass: false,
+        score: 0,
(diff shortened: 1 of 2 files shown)


typpo:
Thanks Michael!

PR #836, opened 2024-05-30
CI: add support for node 22, remove support for node 16

jacobdr:
* add support for node 22
* remove support for node 16
* change root package.json to npm workspaces and remove duplicated package-lock.json
* make dependency versions consistent across the repo
* enforce dependency version consistency in CI
* prevent duplicated builds across PRs
* enable CI concurrency cancellation to cancel builds if new commits are pushed on the branch
* disable fail-fast behavior in the matrix so we can better see the full set of errors across various combinations of node/python versions


typpo:
Thanks @jacobdr - this is a huge improvement.  Always felt icky about the multiple package locks.  Will try it out on my end asap

jacobdr:
@typpo I still need to fix a few things so I'm marking as a draft so we don't accidentally merge. Currently looking at the nextjs site build issues

jacobdr:
@typpo This is now ready for you to test. I tested the UI via `npm install && NEXT_PUBLIC_SUPABASE_URL=http://placeholder.promptfoo.dev NEXT_PUBLIC_SUPABASE_ANON_KEY=placeholder npm run build && npm run bin -- view` but did not do much other extensive testing.

NOTE: I had to leave the NextJS version fixed as it seems the `next export` command is broken in more recent versions with our setup. Is the `export` support necessary? It seems like `standalone` is all we really need to support the CLI -- but not sure about the SHARE site you run. 

If we can get rid of `export`, then I would be happy to take on upgrading NextJS next and adding some testing for it in CI

jacobdr:
@typpo ill handle version upgrades (particularly nextjs) in another PR, this is just to plant the node 16->22 line in the sand

typpo:
🎉

PR #840, opened 2024-05-31
Support custom gateway URLs in Portkey

roh26it:
This PR explores support for a custom AI gateway URL that can be configured while setting up Portkey.

```yaml
portkeyApiBaseUrl: xxx
```


typpo:
Thanks Rohit!

PR #845, opened 2024-06-02
Addresses minor typographical errors

mldangelo:
This PR addresses some minor typographical errors. It mainly corrects the spelling of "hallucination" in various markdown files. Because "hallucations" was also in a URL it does not change the guide file path. I understand these changes are extremely low priority and I was hesitant to introduce them, but it only took a few minutes and documentation is important. Thank you, @typpo, for a great project!


typpo:
Thanks for catching this 😱 

Several hundred people find this article per month on Google, so I would prefer to implement via 301 redirect, rather than in Javascript, so the search engine understands what's going on.  https://developers.cloudflare.com/pages/configuration/redirects/

typpo:
Thanks!

PR #854, opened 2024-06-03
fix: Broken types

### LINKED ISSUE #855: Types for `promptfoo.evaluate` broken when using TS
Using the node api in a TS environment is broken due to overlapping & incompatible types.

Using the `node-package` example (but with Typescript) you can see the errors.
Screenshot 2024-06-03 at 14 38 57
Screenshot 2024-06-03 at 14 39 26
Screenshot 2024-06-03 at 14 39 31

I've began addressing the issue with #854

However the bug runs deep into the "writefile" territory... I would appreciate some guidance on what can/cant be changed here as passing functions into `JSON.stringify` doesn't always do what we want.

brendanmoore:
Resolves #855

Types for the node API are broken when consuming the package in a TS environment


typpo:
@brendanmoore Thanks so much for starting work on this.  I've made an update on top of your code that resolves the outstanding type issues.  Please take a look and let me know if things look alright on your end.

brendanmoore:
@typpo Its looking good! Thanks for tidying those bits

typpo:
Thanks again!

Issue #856, opened 2024-06-03
Add ability to bust cache from the web UI


chrismaltais:
Our team interacts with promptfoo strictly through the web ui.

Right now, there isn't a way to bust the cache from the web ui, making it unusable.

Thinking a solution could either be:
- A button/setting in the web ui to clear the cache
- Supporting `cache: false` in the yaml file specified in the web ui

Or both? @typpo any opinions?

PR #859, opened 2024-06-03
Pass evaluateOptions from web ui yaml

chrismaltais:
I think this closes: https://github.com/promptfoo/promptfoo/issues/856

When hitting the "Run Evaluation" button in the web UI, I realized we're passing the following data in the request body:
https://github.com/promptfoo/promptfoo/blob/f7c420e7179792676ad870dba4b8bdf5b2cb3667/src/web/nextui/src/app/setup/RunTestSuiteButton.tsx#L19-L36

This includes the `evaluateOptions` data, however in the route resolver we drop that data and don't pass it in to the `promptfoo.evaluate(...)` call.

This PR attempts to fix that by ensuring we're forwarding the `evaluateOptions` data correctly.

@typpo can you double check this change doesn't break anything? I didn't get the chance to test locally yet

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -539,13 +539,13 @@ export interface TestCase<Vars = Record<string, string | string[] | object>> {
 
   // Additional configuration settings for the prompt
   options?: PromptConfig &
-  OutputConfig &
-  GradingConfig & {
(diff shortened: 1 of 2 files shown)


typpo:
Looks good!

Issue #858, opened 2024-06-03
Ensure percent complete accurately reflects test suite


chrismaltais:
I have the following promptfoo config file:
```yaml
description: 'Segmenter'
prompts:
- " "
providers:
- id: 'python:src/segmenter/provider.py'
  label: 'Staging'
  config:
    env: 'staging'
    storage_folder: 'test/staging'
- id: 'python:src/segmenter/provider.py'
  label: 'Production'
  config:
    env: 'production'
    storage_folder: 'test/production'
tests: src/segmenter/tests.yaml
evaluateOptions:
  maxConcurrency: 2
  delay: 500
  showProgressBar: true
  cache: false
```

Where the `src/segmenter/tests.yaml` file looks like:
```yaml
- description: Apparel and Accessories
  vars:
    originalImage: "![image](https://example.com/some_image.png)"
  assert:
    - type: latency
      threshold: 2000
- description: Apparel and Accessories
  vars:
    originalImage: "![image](https://example.com/some_image.png)"
  assert:
    - type: latency
      threshold: 2000
- description: Apparel and Accessories
  vars:
    originalImage: "![image](https://example.com/some_image.png)"
  assert:
    - type: latency
      threshold: 2000
```

When I run these tests via the web ui, I'm expecting the results to go 0 -> 33% -> 66% -> 100% (i.e. updating based on the number of variables).

However, the web ui stays at 0% until all the tests are completed, then redirects me to the evals page:
https://github.com/promptfoo/promptfoo/assets/17098959/9457aac4-a284-4246-ba76-0dc7d9509cf5

PR #860, opened 2024-06-03
Fix broken progress callback in web ui

chrismaltais:
I think this closes: https://github.com/promptfoo/promptfoo/issues/858

The purpose of this PR is to ensure the callback defined via the web ui is executed. That callback is defined here:
https://github.com/promptfoo/promptfoo/blob/f7c420e7179792676ad870dba4b8bdf5b2cb3667/src/web/server.ts#L103-L109

The current code actually overrides the `progressCallback` function, which means the one specified in the web ui server never gets executed. We never get to know the progress of our job without it!

This ensures both `progressCallback` functions get executed.


typpo:
Nice catch! Thank you

Issue #857, opened 2024-06-03
When Python provider raises exception, details are no longer recorded


efung:
In #833 by @mldangelo , individual Python assertions that raise an exception now have their exception messages recorded as the `reason` in the `GradingResult` which is then visible as part of the set of `gradingResults` recorded for every test case when requesting JSON output.

However, an undesirable side effect of this is that if a Python **provider** raises an exception, that message is no longer visible.

Previously, say, in 0.59.1, if a provider raised an exception (e.g. incorrect OpenAI token), the exception and stack trace get dumped to the command-line.

Now, the command-line doesn't show anything, but the JSON test results now include an `error` value which begins:

```
Error: The Python script `call_api` function must return a dict with an `output` string/object or `error` string

Error: The Python script `call_api` function must return a dict with an `output` string/object or `error` string
    at PythonProvider.callApi
```

It would be nice if the exception message and stack trace were recorded in the `error` field.

typpo:
Fix landed in #863

typpo:
This change is released in 0.62.0

PR #863, opened 2024-06-04
Improve Error Handling for Python Assertions and Provider Exceptions

mldangelo:
@efung Thank you for identifying the issue around incorrectly handled Python exceptions for Python providers and providing a detailed explanation in [#857](https://github.com/promptfoo/promptfoo/issues/857). The error should have been handled in the assertion code only. This PR:

- Modifies `runPython` to throw an error directly if the Python script execution fails, which will be caught and logged by the calling function. This reverts the behavior introduced in [#833](https://github.com/promptfoo/promptfoo/pull/833).
- When a Python script execution fails, the `runAssertion` function now returns an object containing `pass: false`, `score: 0`, and the error message. This ensures that the failure reason is included in the GradingResult.
- Adds a test case to `assertions.test.ts` to verify that Python file assertions correctly handle errors and return the appropriate GradingResult.


typpo:
Thanks!

Issue #866, opened 2024-06-04
CI passes despite failing build


drien:
See the Build step from this recent run: https://github.com/promptfoo/promptfoo/actions/runs/9372730525/job/25804855528

It ends with:
```
./src/app/eval/[id]/page.tsx:62:9
Type error: Type 'EvaluateTestSuite' is not assignable to type 'Partial<UnifiedConfig>'.
  Types of property 'prompts' are incompatible.
    Type '(string | object | PromptFunction)[]' is not assignable to type 'string | Record<string, string> | (string | Prompt)[] | undefined'.
      Type '(string | object | PromptFunction)[]' is not assignable to type '(string | Prompt)[]'.
        Type 'string | object | PromptFunction' is not assignable to type 'string | Prompt'.
          Type 'object' is not assignable to type 'string | Prompt'.

  60 |         createdAt: result.createdAt,
  61 |         results: result.results as unknown as EvaluateSummary,
> 62 |         config: result.config as unknown as EvaluateTestSuite,
     |         ^
  63 |       },
  64 |     };
  65 | 
npm ERR! Lifecycle script `build` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: nextui@0.1.0 
npm ERR!   at location: D:\a\promptfoo\promptfoo\src\web\nextui 
```

For some reason `npm run build` seems to return a status code of 0 despite those errors, so the step succeeds as far as Github Actions is concerned. I'm not entirely sure how to make it return a nonzero exit code in this situation, and some cursory searching didn't get me anywhere.

typpo:
Looks like @mldangelo has a fix in the works here #876

PR #867, opened 2024-06-04
Fix type error eval page.tsx

drien:
Was getting this error on `npm run build` in on latest main branch. I'm not well versed in typescript so I'm not entirely confident that this is the correct way to fix this, but it seems to work for me locally.

```
./src/app/eval/[id]/page.tsx:62:9
Type error: Type 'EvaluateTestSuite' is not assignable to type 'Partial<UnifiedConfig>'.
  Types of property 'prompts' are incompatible.
    Type '(string | object | PromptFunction)[]' is not assignable to type 'string | (string | Prompt)[] | Record<string, string> | undefined'.
      Type '(string | object | PromptFunction)[]' is not assignable to type '(string | Prompt)[]'.
        Type 'string | object | PromptFunction' is not assignable to type 'string | Prompt'.
          Type 'object' is not assignable to type 'string | Prompt'.

  60 |         createdAt: result.createdAt,
  61 |         results: result.results as unknown as EvaluateSummary,
> 62 |         config: result.config as unknown as EvaluateTestSuite,
     |         ^
  63 |       },
  64 |     };
  65 | 
npm error Lifecycle script `build` failed with error:
npm error code 1
npm error path /home/adrien/Source/promptfoo/src/web/nextui
npm error workspace nextui@0.1.0
npm error location /home/adrien/Source/promptfoo/src/web/nextui
npm error command failed
npm error command sh -c next build
```

### DIFF
--- a/src/web/nextui/src/app/eval/[id]/page.tsx
+++ b/src/web/nextui/src/app/eval/[id]/page.tsx
@@ -7,7 +7,7 @@ import { createServerComponentClient } from '@supabase/auth-helpers-nextjs';
 import { getApiBaseUrl } from '@/api';
 import { IS_RUNNING_LOCALLY } from '@/constants';
 import { getResult } from '@/database';
-import { EvaluateSummary, EvaluateTestSuite, SharedResults } from '@/../../../types';
+import { EvaluateSummary, EvaluateTestSuite, SharedResults, UnifiedConfig } from '@/../../../types';
 import Eval from '../Eval';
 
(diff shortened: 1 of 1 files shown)


typpo:
Thanks - will have to figure out why this test isn't running on all PRs 🤔

drien:
The test is actually running but it gets marked as passing even when it fails because of some exit code weirdness, I made an issue: https://github.com/promptfoo/promptfoo/issues/866

PR #870, opened 2024-06-05
fix: Parameter evaluateOptions not passed correctly in jobs created using web

big-camel:
For example, in the Yaml file:

evaluateOptions:
  repeat: 5

and then POST `/api/eval/job/`


typpo:
Thank you!

PR #872, opened 2024-06-05
Fix formatting and add style check to CI

chrismaltais:
This PR is split into two commits:
1. Run `npm run format` to format the codebase according to prettier rules
2. Add Github action to check formatting as part of CI, so we no longer have to think about formatting going forward!

I see we're running CI on multiple operating systems - I opted to extract the style check to a single OS to reduce the number of CI jobs we're running.

Also - we're not really linting in all the other CI jobs. Should we change the CI titles
From: `Build, lint, and test on Node ...`
To: `Build and test on Node ...`


typpo:
Thank you for doing this!

Issue #871, opened 2024-06-05
[Web UI] Increase robustness of markdown rendering


chrismaltais:
I have the following output for my models. It includes multiple images and text:
```
![Some Image](https://example.com/some_image.png) ![Other Image](https://example.com/other_image.png) IoU Score: 0.8596295537706135 Dice Coefficient: 0.9245169846086984
```

However, the result cell only renders the first image. It looks like the following:
<img width="1890" alt="Screenshot 2024-06-05 at 8 51 19 AM" src="https://github.com/promptfoo/promptfoo/assets/17098959/688c99ab-4d26-4673-a2ae-755503ecc39d">

Even turning on/off the markdown renderer setting doesn't get the rest of the data to appear - it's as though it's been chopped off from the result in the web UI!

PR #873, opened 2024-06-05
Render multiple result images with markdown, if markdown contains multiple images

chrismaltais:
I think this closes: https://github.com/promptfoo/promptfoo/issues/871

I haven't been able to test these changes locally - @typpo do you have any suggestions? Do I write my own custom provider with markdown hardcoded in the response? Where would I need to put the provider file in this project's directory? Any info appreciated, so I can better test these changes without bugging you 😅

### DIFF
--- a/src/web/nextui/src/app/eval/ResultsTable.tsx
+++ b/src/web/nextui/src/app/eval/ResultsTable.tsx
@@ -213,6 +213,15 @@ function EvalOutputCell({
 
   let text = typeof output.text === 'string' ? output.text : JSON.stringify(output.text);
   let node: React.ReactNode | undefined;
+  let chunks: string[] = [];
+
+  // Handle failure messages by splitting the text at '---'
+  if (!output.pass && text.includes('---')) {
+    chunks = text.split('---');
+    text = chunks.slice(1).join('---');
+  } else {
+    chunks = [text];
+  }
 
   if (showDiffs && firstOutput) {
     let firstOutputText =
@@ -516,7 +525,7 @@ function EvalOutputCell({
             </div>
             <CustomMetrics lookup={output.namedScores} />
             <span className="fail-reason">
-              {text
+              {chunks[0]
                 ?.trim()
                 .split('\n')
                 .map((line, index) => (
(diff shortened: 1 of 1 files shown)


typpo:
Thanks for wrangling this, will give it a try on my end!

One thing that may help you test this is to use the built-in (undocumented) `echo` provider, which just outputs the prompt.

src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -209,32 +213,6 @@ function EvalOutputCell({
...
-            </div>
-          )}
-        </>
-      );
-      hasImageLightbox = true;
-    }
-  } else if (!output.pass && text.includes('---')) {
-    // TODO(ian): Plumb through failure message instead of parsing it out.
-    chunks = text.split('---');
-    text = chunks.slice(1).join('---');
```
typpo: We should preserve this "chunks" behavior - it's what parses out and displays failures nicely.

With split:
image

Without split:
image

chrismaltais: ah right let me fix that!

chrismaltais: @typpo when you test with the `echo` provider - are you entering that config via the web ui yaml?

or are you defining a yaml file somewhere, running an eval somehow other than the web ui, and only inspecting the results with the web ui?

the reason I ask is because I get a 500 error when attempting to test only via the web ui. I think it has to do with the `IS_RUNNING_LOCALLY` env var?

the error:
```
error StaticGenBailoutError: Page with `dynamic = "error"` couldn't be rendered statically because it used `request.json`
```

which seems to come from: 
https://github.com/promptfoo/promptfoo/blob/fdb296d97523e4c53c304580dfc2ce9268fbbf2e/src/web/nextui/src/app/api/eval/job/route.ts#L15

src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -316,8 +295,31 @@ function EvalOutputCell({
...
     }
-  } else if (renderMarkdown && !hasImageLightbox) {
-    node = <ReactMarkdown>{text}</ReactMarkdown>;
+  } else if (renderMarkdown) {
+    if (!output.pass && text.includes('---')) {
+      chunks = text.split('---');
+      text = chunks.slice(1).join('---');
+    } else {
+      chunks = [text];
+    }
```
chrismaltais: Oh wait, we probably want this regardless of markdown rendering. Will hoist it out of this conditional

typpo:
Thanks @chrismaltais!  Works well for me locally.

PR #887, opened 2024-06-07
fix: Prompt variable reads unprocessed spaces on both sides

big-camel:
Before:

image

image

Now:
image
image


typpo:
Nice fix.  Thank you!

Issue #879, opened 2024-06-06
expression to select subset of output to display in view


pelikhan:
We provide a rendered view of the output (as markdown) in the output payload. Would it be possible to have an expression to select that subfield and render it instead of the output?

typpo:
👍 If I understand correctly, this request is related to #786 - looking for the ability to display a specific field in the output rather than the entire output.

pelikhan:
yes could be a convention or a configuration.

Issue #786, opened 2024-05-16
Standardize how LLM output is stored in a ProviderResponse output object


efung:
`ProviderResponse` is [documented](https://www.promptfoo.dev/docs/configuration/reference/#providerresponse) to be either a string or an object.  

If it's a string, then the Web UI shows a nice diff between the LLM outputs when comparing two prompts or providers. 

However, if it's an object, the serialized object is shown and you lose the nice diff output.

Can a convention be established about what goes into the object, specifically if a field with a specific key exists, then that represents the LLM output? (And any other fields are ignored by the Web UI).

typpo:
This seems reasonable.  What do you think we should call the displayed field?  `content`, `display`, `text`?

efung:
> This seems reasonable. What do you think we should call the displayed field? `content`, `display`, `text`?

I prefer `text` or `content`.

typpo:
Added a frontend configuration for this in #893

PR #893, opened 2024-06-07
feat(webui): ability to extract json field for display

typpo:
Ability to extract a field from LLM JSON output

For example, given this setting:
image

An output of the form:
```json
{
  "foo": 1,
  "bar": "meow"
}
```

Would just display `1`

Related to #879 and #786

### DIFF
--- a/src/web/nextui/src/app/eval/ResultsTable.tsx
+++ b/src/web/nextui/src/app/eval/ResultsTable.tsx
@@ -167,7 +167,7 @@ function EvalOutputCell({
   showDiffs: boolean;
   searchText: string;
 }) {
-  const { renderMarkdown, prettifyJson, showPrompts } = useResultsViewStore();
+  const { renderMarkdown, prettifyJson, jsonExtractField, showPrompts } = useResultsViewStore();
   const [openPrompt, setOpen] = React.useState(false);
   const handlePromptOpen = () => {
     setOpen(true);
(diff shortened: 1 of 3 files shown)


src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -303,29 +303,45 @@ function EvalOutputCell({
...
+          if (!json) {
+            node = (
+              <em>
+                Field <code>{jsonExtractField}</code> is undefined
+              </em>
+            );
+          }
+        }
+        if (json) {
+          node = <pre>{JSON.stringify(json, null, 2)}</pre>;
```
efung, line 321: Any way that the extracted field could just be rendered the same way as if was just text? That is, wrapped in `<ReactMarkdown>`?

pelikhan:
can this configuration be applied per test ? e.g. each test could potentially have a different report field; or there might be different types of tests in the same system

typpo:
@pelikhan have you considered using the `transform` property on your tests?  [docs](https://promptfoo.dev/docs/configuration/guide/#transforming-outputs)

It allows you to extract from json on a per testcase basis.  For example:

```yaml
tests:
  - vars:
      # ...
    options:
        transform: output.category # Select the 'category' key from output json and use it for this assertion
```

pelikhan:
Yes absolutely I use the transform for assertions; it would be good to have a similar "view-transform" that grabs a field with a pre-computed human readable output

pelikhan:
Would it be possible to have a combo of these features: ability to specify a JSON field/path that contains a markdown trace of the prompt. That way, the prompt tooling is in charge of creating a markdown payload that will be optimized for the promptfoo UI style.

The current PR proposal separate prettyJson vs prettyMarkdown but, IMO, the real scenario is big JSON payload -> markdown field -> nice human output.

mldangelo:
Closing as stale

PR #900, opened 2024-06-08
chore: fix windows node 22 build issues by adding missing encoding dependency and updating webpack config

mldangelo:
This PR addresses the CI build failure caused by a missing `encoding` module in the `node-fetch` dependency. Additionally, it updates the webpack configuration to handle node-specific modules that are not available in the browser environment.

The build process was failing due to a missing `encoding` module required by `node-fetch`. Adding this dependency and updating the webpack configuration ensures that node-specific modules are properly handled during the client-side build process, preventing the CI errors and warnings.


typpo (approved):
🤔😳

PR #904, opened 2024-06-09
chore: introduce eslint

mldangelo:
This PR introduces ESLint for the entire project using the flat config format supported by v9.  It also adds ESLint to the CI for both the web app and the rest of the project.

Note: 
- Most rule violations have been disabled by default. We can rachet up the linter rules over time. Note that I am initially introducing v8 as v9 is not yet widely supported.
- The format of `src/web/nextui/.eslintrc.json` used by the Next.js app was not changed, as the flat config format is not supported by Next.js yet.


typpo (approved):
Looks good!

PR #909, opened 2024-06-10
Add more Portkey header provider options and create headers automatically

vrushankportkey:
Removes the need to hard code all the headers Portkey supports


typpo:
Awesome!  Thanks @vrushankportkey

Issue #914, opened 2024-06-10
bedrock:cohere.command-r-plus-v1:0 - ValidationException


underyx:
using this model on 0.63 returns these validation errors:

```
API call error: ValidationException: Malformed input request:
#: extraneous key [logit_bias] is not permitted
#: extraneous key [truncate] is not permitted
#: extraneous key [stream] is not permitted
#: extraneous key [num_generations] is not permitted
#: extraneous key [return_likelihoods] is not permitted
#: extraneous key [prompt] is not permitted
#/p: 1 is not less or equal to 0.99, please reformat your input and try again.
```

typpo:
Should be fixed in 0.63.1

Issue #913, opened 2024-06-10
bedrock:mistral.mixtral-8x7b-instruct-v0:1 - Validation Error


underyx:
running this model on 0.63 returns all failures with `API call error: ValidationException: Validation Error`

typpo:
Should be fixed in 0.63.1

Issue #912, opened 2024-06-10
bedrock:meta.llama3-70b-instruct-v1:0 - No output


underyx:
Using this model on 0.63 returns all failures which only say 'No output'

typpo:
Should be fixed in 0.63.1

mldangelo:
Thanks for fixing this @fvdnabee. We deployed your change in [0.68.0](https://github.com/promptfoo/promptfoo/releases/tag/0.68.0)!

Issue #884, opened 2024-06-06
Error: Unknown Amazon Bedrock model: meta.llama3-70b-instruct-v1:0


underyx:
Thanks for the tool! Just tried to compare a bunch of providers via bedrock, and I'm getting a couple errors, e.g. for this provider config:

```yaml
  - id: bedrock:meta.llama3-70b-instruct-v1:0
    config:
      region: "us-west-2"
```

```
Error: Unknown Amazon Bedrock model: meta.llama3-70b-instruct-v1:0
    at getHandlerForModel (/usr/local/lib/node_modules/promptfoo/dist/src/providers/bedrock.js:95:11)
    at AwsBedrockCompletionProvider.callApi (/usr/local/lib/node_modules/promptfoo/dist/src/providers/bedrock.js:142:21)
    at Evaluator.runEval (/usr/local/lib/node_modules/promptfoo/dist/src/evaluator.js:273:62)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async processEvalStep (/usr/local/lib/node_modules/promptfoo/dist/src/evaluator.js:601:25)
```

Same issue for `mistral.mixtral-8x7b-instruct-v0:1`.

typpo:
Thanks for flagging!  Support for llama, mistral, and cohere via Bedrock is added in #885

underyx:
Amazing! Excited to try.

typpo:
This is landed in 0.63.0, although I was not able to test all the bedrock providers (I don't have access to all).  Let me know if you run into any issues!

underyx:
Yeah, exciting! Unfortunately none of them seem to _just work_. I created #912 #913 #914.

For the record, I think all requests for access to models are instantly auto-approved on AWS.

PR #915, opened 2024-06-10
fix(bedrock): support for llama, cohere command and command-r, mistral

typpo:
Related to #912, #913, #914, #884

- Llama - requires formatting like `<s> [INST] ... [/INST]`.  Should work in current version, but added support for OpenAI-style chat formatting
- Mistral - removed unused properties that failed validation
- Command-R - the previous implementation was for `Command` format, now we support both `Command` and `Command R` formats

cc @underyx

### DIFF
--- a/src/providers/bedrock.ts
+++ b/src/providers/bedrock.ts
@@ -1,4 +1,5 @@
 import Anthropic from '@anthropic-ai/sdk';
+import dedent from 'dedent';
 
 import logger from '../logger';
 import { getCache, isCacheEnabled } from '../cache';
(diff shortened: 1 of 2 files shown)


src/providers/bedrock.ts
```diff
@@ -73,96 +168,225 @@ interface IBedrockModel {
...
 }
 
+function addConfigParam(
+  params: any,
+  key: string,
+  configValue: any,
+  envValue?: string | undefined,
+  defaultValue?: any,
+) {
+  if (configValue !== undefined || envValue !== undefined) {
```
underyx, line 178: this is missing a `|| defaultValue !== undefined`

Issue #916, opened 2024-06-10
default param values not getting added to bedrock requests on any provider


underyx:
after 0.63.1 (#915) claude returns

```
API call error: ValidationException: Malformed input request:
#: subject must not be valid against schema {"required":["messages"]}
#: required key [max_tokens] not found
#: required key [anthropic_version] not found, please reformat your input and try again.
```

the problem is likely this: https://github.com/promptfoo/promptfoo/pull/915/files#r1633789271

typpo:
oof, thanks for catching.  I tested everything _except_ for Claude.

#917 should resolve this.  Also added some quick tests that would've caught this issue.

PR #917, opened 2024-06-10
fix(bedrock): default value for configs

typpo:
#916


underyx (approved):
kudos on writing the test suite!

PR #926, opened 2024-06-13
feat: assertion type: is-sql

YingjiaLiu99:
added new assertion type: is-sql; 
added related assertion tests in ./test/assertions.test.ts;
added is-sql's dependent library's tests(node-sql-parser) in a new folder ./test/is-sql-tests;
updated docs; 
add node-sql-parser to the peerDependencies


site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,70 @@ assert:
...
+- FlinkSQL
+- Snowflake(alpha)
+- Noql
+
+Example:
+
+```yaml
+assert:
+  - type: is-sql
+    value: { databaseType: 'MySQL' }
```
typpo: Might be good to do this in pure yaml:

```yaml
assert:
  - type: is-sql
    value:
      databaseType: 'MySQL'
```

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,70 @@ assert:
...
+
+```yaml
+assert:
+  - type: is-sql
+    value:
+      {
+        databaseType: 'MySQL',
+        whiteTableList: ['(select|update|insert|delete)::null::departments'],
+        whiteColumnList: ['select::null::name', 'update::null::id'],
+      }
```
typpo: same here re: pure yaml

src/assertions.ts
```diff
@@ -10,6 +10,7 @@ import yaml from 'js-yaml';
 import Ajv, { ValidateFunction } from 'ajv';
 import addFormats from 'ajv-formats';
 import { distance as levenshtein } from 'fastest-levenshtein';
+import { Parser } from 'node-sql-parser';
```
typpo: Because `node-sql-parser` is a `peerDependency`, we should lazily import and instantiate it within the `baseType === 'is-sql'` conditional.  Otherwise promptfoo will crash for people who have not manually installed the peer dependency.

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
     };
   }
 
+  if (baseType === 'is-sql') {
+    let parsedSQL;
+    let databaseType: string | undefined = 'MySQL';
```
typpo: Because this has a default, seems like its type may not need to have an `undefined`?

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
...
+    let whiteTableList: string[] | undefined;
+    let whiteColumnList: string[] | undefined;
+
+    interface sqlParserOption {
+      database: string | undefined;
+      type?: string; // 'type' is optional
+    }
```
typpo: You might be able to get this type from the `@types/node-sql-parser` package.

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
...
+        databaseType = value.database || 'MySQL';
+        whiteTableList = value.whiteTableList;
+        whiteColumnList = value.whiteColumnList;
+      } else {
+        throw new Error('is-sql assertion must have a object value.');
+      }
+    }
+
+    let opt: sqlParserOption = { database: databaseType };
+    let failed_reason = '';
```
typpo: style in this codebase is camel case - `failedReason`

test/assertions.test.ts
```diff
@@ -423,6 +423,43 @@ describe('runAssertion', () => {
 `,
   };
 
+  // is-sql assertions:
+  const isSQLAssertion: Assertion = {
```
typpo: This is the tiniest of nits, but the convention in this codebase is to lowercase acronyms/initialisms in camel case var names.  For example, `isSqlAssertion`, `evalId`, `OpenAi`, etc...

test/is-sql-tests/node-sql-parser.test.ts
```diff
@@ -0,0 +1,276 @@
+import { Parser } from 'node-sql-parser';
+
+// The isolated is-sql implementation:
```
typpo: We should extract the is-sql implementation into its own function, so that we can call it from `assertions.ts` and from its test.  This means we don't have to maintain two separate versions of the logic, which will invariably fall out of sync.

typpo (commented):
Thank you for this PR, it's looking great!  Just a few small comments.

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: got it, solved

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: got it!

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,70 @@ assert:
...
+The `is-sql` assertion checks if the LLM output is a valid SQL statement.
+
+Example:
+
+```yaml
+assert:
+  - type: is-sql
+```
+
+You can optionally set a `databaseType` in the `value` to determine the specific database syntax that your LLM output will be validated against. The default database syntax is MySQL.
```
mldangelo: Nit, consider referencing the parser. 
```suggestion
You can optionally set a `databaseType` in the `value` to determine the specific database syntax that your LLM output will be validated against. The default database syntax is MySQL. For a complete and up-to-date list of supported database syntaxes, please refer to the [node-sql-parser documentation](https://github.com/taozhi8833998/node-sql-parser?tab=readme-ov-file#supported-database-sql-syntax).
```

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,70 @@ assert:
...
+
+```yaml
+assert:
+  - type: is-sql
+```
+
+You can optionally set a `databaseType` in the `value` to determine the specific database syntax that your LLM output will be validated against. The default database syntax is MySQL.  
+The supported database syntax list:
+
+- Athena
```
mldangelo, line 285: nit, sort

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
...
 
+  if (baseType === 'is-sql') {
+    let parsedSQL;
+    let databaseType: string | undefined = 'MySQL';
+    let whiteTableList: string[] | undefined;
+    let whiteColumnList: string[] | undefined;
+
+    interface sqlParserOption {
+      database: string | undefined;
+      type?: string; // 'type' is optional
```
mldangelo: The type makes this obvious 
```suggestion
      type?: string;
```

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
...
+    }
+
+    if (renderedValue) {
+      if (typeof renderedValue === 'object') {
```
mldangelo: nit, prefer to unnest if

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
...
+        whiteColumnList = value.whiteColumnList;
+      } else {
+        throw new Error('is-sql assertion must have a object value.');
+      }
+    }
+
+    let opt: sqlParserOption = { database: databaseType };
+    let failed_reason = '';
+    try {
+      parsedSQL = sqlParser.astify(outputString, opt); // mysql sql grammer parsed by default
```
mldangelo:
```suggestion
      parsedSQL = sqlParser.astify(outputString, opt);
```

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
     };
   }
 
+  if (baseType === 'is-sql') {
+    let parsedSQL;
```
mldangelo: nit, prefer google naming conventions https://google.github.io/styleguide/jsguide.html#naming
```suggestion
    let parsedSql;
```

test/assertions.test.ts
```diff
@@ -889,6 +926,181 @@ describe('runAssertion', () => {
     );
   });
 
+  // is-sql test cases:
```
mldangelo:
```suggestion
```

test/assertions.test.ts
```diff
@@ -889,6 +926,181 @@ describe('runAssertion', () => {
...
+
+    const result: GradingResult = await runAssertion({
+      prompt: 'Some prompt',
+      provider: new OpenAiChatCompletionProvider('gpt-4'),
+      assertion: isSQLAssertion,
+      test: {} as AtomicTestCase,
+      output,
+    });
+    expect(result.pass).toBeTruthy();
+    expect(result.reason).toBe('Assertion passed');
```
mldangelo: nit, there are 4 values returned on this object, write one expectation that checks all 4: 

```ts
expect(result).toEqual({ .... });

File: src/assertions.ts (hunk 6 above)
Comment (mldangelo) [inline src/assertions.ts]: consider defining 
```suggestion
    const failureReasons = [];
```
then pushing to the array, and joining at the end. It will solve your leading whitespace problems with some of the error messages.

src/assertions.ts
```diff
@@ -366,6 +369,81 @@ export async function runAssertion({
...
+      };
+      try {
+        sqlParser.whiteListCheck(outputString, whiteColumnList, opt);
+      } catch (err) {
+        pass = inverse;
+        failed_reason += ' It failed the provided authority column list check.';
+      }
+    }
+
+    if (inverse && pass === false && !failed_reason) {
```
mldangelo: when is this condition met? Can you add coverage for it in your tests if it's reachable?

test/is-sql-tests/node-sql-parser.test.ts
```diff
@@ -0,0 +1,276 @@
...
+    expect(result.reason).toBe(
+      'SQL statement does not conform to the provided MySQL database syntax.',
+    );
+  });
+
+  /**
+   * Catches an incorrect output from node-sql-parser package
+   * The paser cannot identify the syntax error: missing comma between column names
+   */
+  // it('should fail when the output string is an invalid SQL statement', () => {
```
mldangelo, line 53: please address these and below

mldangelo (commented):
This is a great feature! It deserves an example and maybe even a guide. Thank you so much!

src/assertions.ts
YingjiaLiu99: True

src/assertions.ts
YingjiaLiu99: got it

test/assertions.test.ts
YingjiaLiu99: got it

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: got it!

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: got it!

src/assertions.ts
YingjiaLiu99: got it

src/assertions.ts
YingjiaLiu99: got it

test/assertions.test.ts
YingjiaLiu99: got it

src/assertions.ts
YingjiaLiu99: got it

src/assertions.ts
YingjiaLiu99: got it!

test/is-sql-tests/node-sql-parser.test.ts
YingjiaLiu99: got it

src/assertions.ts
YingjiaLiu99: Hi!
this condition is met when the user input type is `not-is-sql`, and the LLM output string is a valid SQL statement, which means inverse will be true, the pass will be false and failedReason at that line will be empty. (since all the failed reasons such as `invalid syntax`, `not allowed by the white list`, etc, will not be applicable). Based on my understanding, the expected failedReason should be assigned as `The output SQL statement is valid`, which is exactly why the not-is-sql failed. 

I already added the corresponding test cases to the assertion.test.ts, feel free to take a look when I commit everything

src/assertions.ts
YingjiaLiu99: updated! Feel free to take a look when I commit it

test/is-sql-tests/node-sql-parser.test.ts
YingjiaLiu99: Hi, thank you for your feedback. The test cases commented in the `node-sql-parser.test.ts` file are the edge cases I found that demonstrate unexpected behavior produced by the `node-sql-parser` library. These issues stem from the library itself and are beyond my control. I will report these behaviors to the package author and wait for their feedback. Please let me know if you have any further suggestions or if there's anything else I can do.

test/assertions.test.ts
YingjiaLiu99: Thank you for your suggestion. I can definitely update the expectation to check all four values in a single statement. I followed the current format because all the other test cases' expectations in this file are written this way. However, I am happy to change mine if you prefer this new format.

src/assertions.ts
YingjiaLiu99: Thank you for your suggestion. The type here is chosen by me to meet the input requirement of `whiteListCheck` function. When the value of assertion contains a whiteTableList or a whiteColumnList, I will manually put the type as `table` or `column`. The reason why the type is optional is that if the assertion doesn't contain any white list, we don't need the type here anymore. 
I attached two screenshots of the usage of the function `whiteListCheck` for reference. Please let me know if you have any other preferences or suggestions.

fa2965370baac16f28bf9548f0545a6
dd1f181ac0e375793933a6d051e6088

src/assertions.ts
YingjiaLiu99: Thank you for pointing out! I deleted the comment. Just to clarify the usage of the `type` here to avoid any misunderstanding. 
The type here is chosen by me to meet the input requirement of `whiteListCheck` function. When the value of assertion contains a whiteTableList or a whiteColumnList, I will manually put the type as `table` or `column`. The reason why the type is optional is that if the assertion value doesn't contain any white list, we won't need the type here anymore. 
I attached two screenshots of the usage of the function `whiteListCheck` for reference. Please let me know if you have any other preferences or suggestions.

fa2965370baac16f28bf9548f0545a6
dd1f181ac0e375793933a6d051e6088

test/assertions.test.ts
mldangelo: Thank you for following the established conventions. Whatever you think is best. I am learning the codebase too.

src/assertions.ts
typpo: You can import this type from the library itself rather than defining it in multiple places.  For example,
```ts
import { Parser, type Option as SqlParserOption } from 'node-sql-parser';
```

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,72 @@ assert:
...
+assert:
+  - type: is-sql
+    value:
+      databaseType: 'MySQL'
+```
+
+You can also optionally set a `whiteTableList`/`whiteColumnList` in the `value` to determine the SQL authority list that your LLM output will be validated against.  
+The format of whiteTableList:
+
+```md
```
typpo: nit: this code block and the one below are not markdown, so you can remove the `md` language

src/assertions.ts
```diff
@@ -1296,6 +1305,86 @@ function containsJSON(str: string): any {
   return jsonObjects;
 }
 
+export async function isSql(
+  outputString: string,
+  renderedValue: AssertionValue | undefined,
+  inverse: boolean,
+  assertion: Assertion,
+): Promise<GradingResult> {
+  let pass: boolean = false;
```
typpo: nit: `boolean` type is implied 
```suggestion
  let pass = false;
```

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,72 @@ assert:
     value: file://./path/to/schema.json
 ```
 
+### Is-SQL
+
+The `is-sql` assertion checks if the LLM output is a valid SQL statement.
```
typpo, line 267: we should mention that the user needs to `npm install node-sql-parser`

typpo (commented):
Looking very good!  Just some minor comments

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: Should I leave it as plain text? This is what plain text looks like:
7c4acddefa2ac9bbdabcce3bc27a055

src/assertions.ts
YingjiaLiu99: got it

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: got it

site/docs/configuration/expected-outputs/deterministic.md
YingjiaLiu99: got it

src/assertions.ts
```diff
@@ -1296,6 +1305,86 @@ function containsJSON(str: string): any {
...
+    );
+  }
+
+  if (whiteTableList) {
+    opt.type = 'table';
+    try {
+      sqlParser.whiteListCheck(outputString, whiteTableList, opt);
+    } catch (err) {
+      pass = inverse;
+      failureReasons.push('It failed the provided authority table list check.');
```
typpo: Is there useful information in the error that we can pass along? 
 Ideally something like "SQL output references disallowed table X"

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -261,6 +262,78 @@ assert:
...
+Example:
+
+```yaml
+assert:
+  - type: is-sql
+    value:
+      databaseType: 'MySQL'
+```
+
+You can also optionally set a `whiteTableList`/`whiteColumnList` in the `value` to determine the SQL authority list that your LLM output will be validated against.
```
typpo: let's call these `allowedTables` and `allowedColumns` rather than splitting the word "whitelist"

YingjiaLiu99: Got it

src/assertions.ts
YingjiaLiu99: When the LLM output string is `UPDATE employees SET department_id = 2 WHERE employee_id = 1`, and the allowedTables is `['(select|update|insert|delete)::null::departments']`

The output error is something like this: 
 Error: authority = 'update::null::employees' is required in table whiteList to execute SQL = 'UPDATE employees SET department_id = 2 WHERE employee_id = 1'
        at r.whiteListCheck (\Users\taozhi\Workspaces\OpenSources\node-sql-parser\src\parser.js:57:26)
        at isSql (D:\promptfoo place\promptfoo\src\assertions.ts:1359:17)
        ....
        
So basically if the output string tries to do an operation that is not specified(allowed) in the list, the error will report what is required in the tableWhiteList to execute it. Something like this: `'delete::null::employees' is required in table whiteList to execute SQL = 'DELETE FROM employees;'`

What do you think?

typpo: I think that is ok.  I would make the full reason something like: `SQL validation failed: <error>`

YingjiaLiu99: updated!

typpo:
@YingjiaLiu99 really nice work!

Issue #927, opened 2024-06-14
[BUG] When using eval command there are diff between using config option and not using config option


suhwan-kang:
I found that if i use eval command with config option, the prompts label is disappeared.

I try to analyze why it cause and found out that there is code problem.
In deail,
`main.ts ->` when using config option, it use `readconfigs` function which is using readConfig multiple time for reading config files and do post process for each config datas.
and when not using config option, it use `maybeReadConfig` function which is only using `readConfig` onetime.
The function `readConfigs`'s post processing has problem.
`util.ts ->` `readConfigs` has internal function `seenPrompts` process only prompt's id without process label.
It cause the problem that label gonna disappear when I use objects array that contain both id and label.
In result, result table miss the label option and display python code or raw prompt which did't render template variables.

Could you check it and give me a comment that whether it is intentional or not?

typpo:
This is not intentional - seems to be a bug.  `seenPrompts` should take into account the label as well.  Thank you for catching and analyzing this bug!

suhwan-kang:
It seems that Version 0.64.0 didn't have this bug patch.
Could you tell me when this fix is applied in release package?
I hope the bug patch is reflected in a released version.

typpo:
Thanks @suhwan-kang, this will be included in the next release!

typpo:
released in 0.65.2

PR #928, opened 2024-06-14
Bug fix: label not showing problem when using eval with config option

suhwan-kang:
https://github.com/promptfoo/promptfoo/issues/927
Based on the above issue, I present the following PR.
This PR fixes an issue where the label did not work when using the `--config` option with `promptfoo eval`.

I tested it personally but it should be tested by others again.


typpo:
Thanks @suhwan-kang !

PR #954, opened 2024-06-18
feat(cli): add tests for CLI commands and fix version flag bug

mldangelo:
**Summary:**

This introduces the concept of a black-box testing on the CLI, starting with the most basic commands, `--help` and `--version`. It also fixes a bug related to the `--version` flag where we would output the entire help menu instead of just the version. Additionally, it introduces a snapshot test on the help menu output which will need to be updated when the help menu is updated. 

Over time we can add exhaustive tests on every cli command when we figure out appropriate mocking.

### DIFF
--- a/test/__snapshots__/cli.test.ts.snap
+++ b/test/__snapshots__/cli.test.ts.snap
@@ -1,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CLI Tool --help and --version should display help menu 1`] = `
+"
+> promptfoo@0.65.0 local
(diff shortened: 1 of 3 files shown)


test/__snapshots__/cli.test.ts.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CLI Tool --help and --version should display help menu 1`] = `
+"
+> promptfoo@0.65.0 local
```
typpo: is this going to break when we bump the version?

PR #967, opened 2024-06-19
Revert "feat(cli): add tests for CLI commands and fix version flag bug"

typpo:
Reverts promptfoo/promptfoo#954

Causing CI flakes


src/main.ts
```diff
@@ -227,7 +227,6 @@ async function main() {
       fs.readFileSync(path.join(getDirectory(), '../package.json'), 'utf8'),
     );
     logger.info(packageJson.version);
-    process.exit(0);
```
mldangelo, line 230: nit, we should keep this line!

mldangelo (approved):

src/main.ts
typpo: 👍 done separately

PR #968, opened 2024-06-19
fix: should return the whole message if the OpenAI return the content and the function call/tools at the same time.

qwlong:
Should return the whole message if the OpenAI return the content and the function call/tools at the same time

The return json will be like this

`
 "choices": [
    {
      "finish_reason": "function_call",
      "index": 0,
      "logprobs": null,
      "message": {
        "content": "Perfect! I'll go ahead and deal with your request, Just a moment while I confirm that for you.",
        "function_call": {
          "arguments": "{\"start_at\":\"2024-05-16T15:30:00Z\",\"end_at\":\"2024-05-16T16:00:00Z\",\"time_zone\":\"UTC\"}",
          "name": "arrange_meeting"
        },
        "role": "assistant"
      }
    }
  ]
}
`


typpo:
Thanks @qwlong!

PR #973, opened 2024-06-20
Allow variable cells to scroll instead of exploding the table height

grrowl:
This was driving me mad: As (as a user) I am primarily focussed on the output, not the (sometimes many) input variables. This PR allows the variable cells to scroll vertically if they exceed the height of the rest of the table, as set by the content cells.


typpo:
Very nice improvement. Thank you @grrowl!

PR #932, opened 2024-06-15
 feat(anthropic): support tool use

mldangelo:
This pull request adds support for tool use in the latest version (0.22.0) of Anthropic's API. It also refactors all of our Anthropic tests, moving them into a new separate file, increases test coverage, and adds an example to showcase how tool use works. The changes to the Anthropic provider should be fully backwards compatible with the previous version.

### Test Instructions

1. **Run the Example:**
   - Navigate to the `examples/tool-use` directory and run the example:
     ```sh
     cd examples/tool-use
     promptfoo eval
     ```
   - View the results by running:
     ```sh
     promptfoo view
     ```

2. **Run Tests:**
   - Execute the test suite to ensure all tests pass:
     ```sh
     npm test
     ```

PR #977, opened 2024-06-20
Pass tool_choice to anthropic when set in config

greysteil:
Builds on @mldangelo's work in https://github.com/promptfoo/promptfoo/pull/932. Without this tool use in anthropic is a bit painful (the response is a message with some JSON at the end, rather than just JSON).


typpo:
Nice! Thanks @greysteil :)

PR #981, opened 2024-06-21
Fixed according to Ollama API specifications

keishidev:
Description
According to the documentation provided [here](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-embeddings), the response returns "embedding" instead of "embeddings".

{
  "embedding": [
    0.5670403838157654, 0.009260174818336964, 0.23178744316101074, -0.2916173040866852, -0.8924556970596313,
    0.8785552978515625, -0.34576427936553955, 0.5742510557174683, -0.04222835972905159, -0.137906014919281
  ]
}

Therefore, I have modified the part of the code that references the response from the Ollama provider to use "embedding" instead of "embeddings".

before
image

after
image


typpo:
Nice catch!  Thank you

PR #994, opened 2024-06-23
feat: normalize prompt input formats, introduce single responsibility handlers, improve test coverage, and fix minor bugs

mldangelo:
This pull request introduces a complete rewrite of `prompts.ts` with several key improvements and bug fixes.

## Notable Changes

1. **Normalization of Prompt Formats**:
   - The update normalizes all possible prompt formats, including string, filepath, glob, array of string, filepath, glob, and record map, into an array type. This unification simplifies the input handling and ensures consistency across various formats.

2. **Improved Processing Logic**:
   - The processing logic has been broken down per file type, making it easier to test and maintain. Each file type now has its dedicated processing function, enhancing the modularity and readability of the code.

3. **Behavioral Changes**:
   - Display the file name in the label if a label is not present.

4. **Enhanced Test Coverage**:
   - The tests have been rewritten, and the test coverage has been increased to nearly 100%. The only exception is for functions used to test Python generators (@typpo let's discuss this). 
   
5. Moves grading prompts into a separate file as they feel separate enough from parsing prompts from config files. They are still re-imported and exported from prompts.ts. I can save this for a future PR or revert. 

## Bug Fixes

- This update resolves a reported issue on Discord where the eval would not run if two providers each included a separate prompt attribute. I include an example of this in the examples folder to verify that it behaves as expected. 
   - Disallowed prompt files that contain no prompts. Previously, such files would pass silently, potentially causing issues during evaluations.

Note: There are likely a few unintended side effects around glob routes and record<string, string> provider configs. Please review these carefully.


src/prompts.ts
```diff
@@ -20,10 +19,16 @@ import type {
...
+export * from './gradingPrompts';
 
 const PROMPT_DELIMITER = process.env.PROMPTFOO_PROMPT_SEPARATOR || '---';
 
+/**
+ * Reads and maps provider prompts based on the configuration and parsed prompts.
+ * @param {Partial<UnifiedConfig>} config - The configuration object.
+ * @param {Prompt[]} parsedPrompts - Array of parsed prompts.
+ * @returns {TestSuite['providerPromptMap']} A map of provider IDs to their respective prompts.
+ */
```
typpo: nit, but I most of the new docstrings for all these functions to be redundant.  

Do you find these useful?  Perhaps we should prefer tsdoc, which at least relieves us from having to write redundant types?  e.g.
```
 * @param prompt - The prompt object to check.
 * @param allowedPrompts - The list of allowed prompt labels.
 * @returns true if the prompt is allowed, false otherwise.
 ```

File: src/prompts.ts [hunk 2]
```diff
@@ -73,339 +78,378 @@ export function readProviderPromptMap(
...
+  if (typeof str !== 'string') {
+    throw new Error(`Invalid input: ${JSON.stringify(str)}`);
+  }
+
+  const forbiddenSubstrings = ['\n', 'portkey://', 'langfuse://'];
+  if (forbiddenSubstrings.some((substring) => str.includes(substring))) {
+    return false;
+  }
+
+  const validFileExtensions = ['.cjs', '.js', '.json', '.jsonl', '.mjs', '.py', '.txt'];
```
typpo: should we put this at the top of the file as a constant?  feel like if it's buried here we'll miss it when we add the next language

src/prompts.ts
```diff
@@ -73,339 +78,378 @@ export function readProviderPromptMap(
...
-    role: 'user',
-    content: 'Output: {{ output }}\nRubric: {{ rubric }}',
+/**
+ * Python prompt function. Runs a specific function from the python file.
+ * @param promptPath - Path to the Python file.
+ * @param functionName - Function name to execute.
+ * @param context - Context for the prompt.
+ * @returns The prompts
+ */
+export const promptFunction = async (
```
typpo: let's name `promptFunction` and `promptFunctionLegacy` to indicate that they are for python

src/prompts.ts
```diff
@@ -73,339 +78,378 @@ export function readProviderPromptMap(
...
+/**
+ * Processes a string as a literal prompt.
+ * @param {RawPrompt} prompt - The raw prompt data.
+ * @returns {Prompt[]} Array of prompts created from the string.
+ */
+function processString(prompt: RawPrompt): Prompt[] {
+  return [
+    {
+      raw: prompt!.raw as string,
+      label: `${prompt.source ? `${prompt.source}: ` : ''}${prompt.raw}`,
```
typpo: As discussed, we can leave out the source for inline prompt strings.

src/prompts.ts
```diff
@@ -73,339 +78,378 @@ export function readProviderPromptMap(
...
+    return processString(prompt);
+  }
 
-{% for output in outputs %}
-<Text index="{{ loop.index0 }}">
-{{ output }}
-</Text>
-{% endfor %}
+  let promptPath = path.join(basePath, prompt.raw);
+  if (promptPath.includes('file:')) {
```
typpo: should this be `prompt.raw.startsWith('file://')`?

src/prompts.ts
```diff
@@ -73,339 +78,378 @@ export function readProviderPromptMap(
...
+  if (extension === '.txt') {
+    return processTxtFile(promptPath, prompt);
+  }
+  if (extension === '.py') {
+    return processPythonFile(promptPath, prompt, functionName, prompt.raw);
+  }
+  if (['.js', '.cjs', '.mjs'].includes(extension)) {
+    return processJsFile(promptPath, functionName);
+  }
+  return [];
```
typpo: if we make it this far, does this mean the glob matches nothing OR it's an unsupported extension?  In that case should we throw?

also, let's keep this outside this PR but seems like we should support yaml given support for yaml imports elsewhere.

typpo (commented):
Very cool change that will make it much easier to work with prompts.ts in the future.  Just a couple minor comments

src/prompts/processors/string.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Processes a string as a literal prompt.
+ * @param prompt - The raw prompt data.
+ * @returns Array of prompts created from the string.
+ */
+export function processString(prompt: Partial<Prompt>): Prompt[] {
+  return [
+    {
+      raw: prompt!.raw as string,
```
typpo: could we use invariant rather than `!`?

src/prompts/processors/json.ts
```diff
@@ -0,0 +1,23 @@
+import * as fs from 'fs';
+import { Prompt } from '../../types';
+
+/**
+ * Processes a JSON file to extract prompts.
+ * This function reads a JSON file, parses it, and maps each entry to a `Prompt` object.
```
typpo: nit: looks like we don't parse the json file

site/docs/configuration/parameters.md
```diff
@@ -104,13 +111,13 @@ Prompts can be JSON too. Use this to configure multi-shot prompt formats:
...
-### Multiple prompts in a single file
+### Multiple prompts in a single text file
 
-If you have only one file, you can include multiple prompts in the file, separated by the separator `---`. If you have multiple files, each prompt should be in a separate file.
+If you have only one text file, you can include multiple prompts in the file, separated by the separator `---`. If you have multiple files, each prompt should be in a separate file.
 
 Example of a single prompt file with multiple prompts (`prompts.txt`):
 
-```
+```text
```
typpo: ??

typpo (approved):
Very cool change

PR #998, opened 2024-06-23
feat: add support for gemini default grader if credentials are present

typpo:
If the system looks like it's configured to use Gemini, then use it as a grader.  

Also includes some improvements to error messages, for example when default application credentials are not provided.


mldangelo (approved):

dynamicwebpaige:
Thank you for adding this feature! Major quality of life improvement for Gemini users. 🙌

PR #1002, opened 2024-06-23
feat(redteam): allow arbitrary providers

mldangelo:
### Summary

This update focuses on two main improvements for the redteam tests:

1. **Abstracted Provider Initialization**:
   - Refactored the code to remove hardcoded instances of `OpenAiChatCompletionProvider`. Use `loadApiProvider` instead to dynamically load and configure different API providers. This makes it easier to switch between various providers, like other OpenAI models or alternatives such as Ollama's `llama2-uncensored:latest` (providers like anthropic refuse to generate examples). 

2. **Enhanced Debug Messages**:
   - Improved the error message in the `getPurpose` function to provide more context when the purpose is not found.

### DIFF
--- a/site/docs/guides/llm-redteaming.md
+++ b/site/docs/guides/llm-redteaming.md
@@ -289,24 +289,24 @@ npx promptfoo@latest generate redteam -w --plugins 'harmful,jailbreak,hijacking'
 
 The following plugins are enabled by default:
 
-| Plugin Name      | Description                                                                  | Enabled by Default |
-| ---------------- | ---------------------------------------------------------------------------- | ------------------ |
-| contracts        | Tests if the model makes unintended commitments or agreements.               | Yes                |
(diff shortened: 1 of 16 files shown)


typpo (commented):
Looks pretty good!  Should we add a `-p, --provider` arg to `generate redteam` so the dev can override?

typpo:
Another thought: we should use the default grading provider from `defaults.ts`.  That way it will have a fallback on systems that are configured for Claude/Gemini only.

typpo (approved):
Looks GREAT to me

mannharleen:
Just in time... Awaiting the merge

mannharleen:
On a related note: even the `promptfoo generate dataset` seems to be having the same issue. will this PR fix that too?

mldangelo:
> On a related note: even the `promptfoo generate dataset` seems to be having the same issue. will this PR fix that too?

@mannharleen this is a great idea! I will add this tomorrow!

PR #1018, opened 2024-06-26
docs: fix incorrect yaml ref in guide

fvdnabee:
Prior to the fix the config isn't actually dereferenced.


mldangelo (approved):
Thank you so much @fvdnabee!

Issue #1017, opened 2024-06-26
http provider do evaluation, response no output!


ReverseFelicity:
npx promptfoo@latest --version   0.66.0

/home/xxxx/.npm/_npx/81bbc6515d992ace/node_modules/promptfoo/dist/src/providers/http.js
<img width="395" alt="image" src="https://github.com/promptfoo/promptfoo/assets/12491219/bc588a52-5844-4574-b08f-7afb12f6b5a2">

if I use the http provider, then the response is a string instead of a structure with "output"
So everytime I add assertions to it, it always respond "No output"
https://github.com/promptfoo/promptfoo/blob/main/src/evaluator.ts#L378

For now I have to change this line adhoc to return {output: this.responseParser(response.data)};

https://github.com/promptfoo/promptfoo/blob/main/src/providers/http.ts#L69

typpo:
Thanks for flagging!  Fix in #1027

mldangelo:
We deployed a fix for this in [0.68.0](https://github.com/promptfoo/promptfoo/releases/tag/0.68.0). Thanks for reporting this @ReverseFelicity!

PR #1027, opened 2024-06-27
fix: http provider returns the correct response format

typpo:
Related to #1017


src/providers/http.ts
```diff
@@ -66,7 +66,7 @@ export class HttpProvider implements ApiProvider {
     }
     logger.debug(`\tHTTP response: ${JSON.stringify(response.data)}`);
 
-    return this.responseParser(response.data);
+    return { output: this.responseParser(response.data) };
```
mldangelo, line 69: nit, lets add a unit test for this

Issue #1028, opened 2024-06-27
eval failing with `RangeError: Invalid count value: -10`


bellini666:
Just got this in our CI

```
/home/circleci/codriver/tools/promptfoo/node_modules/promptfoo/dist/src/updates.js:34
        const border = '='.repeat(process.stdout.columns - 10);
                           ^

RangeError: Invalid count value: -10
    at String.repeat (<anonymous>)
    at checkForUpdates (/home/circleci/codriver/tools/promptfoo/node_modules/promptfoo/dist/src/updates.js:34:28)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async main (/home/circleci/codriver/tools/promptfoo/node_modules/promptfoo/dist/src/main.js:185:5)
```

The strangest thing is that:

1) I can't reproduce the same issue at all locally
2) It was working correctly on CI before, started failing a couple of hours ago without _any changes_ to any dependencies (it is using `npx promptfoo`, but we have it installed on `node_packages` at version `0.66.0`) and neither to any of the config/assertions

typpo:
Thanks for flagging.  This code path is checking for updates and outputting a banner if there is a new version.  0.67.0 was _just_ released, so that's why the problem just started :)

As a temp workaround, set the environment variable `PROMPTFOO_DISABLE_UPDATE=1`.  This will prevent it from checking for updates.

Alternatively, updating CI to use promptfoo 0.67.0 will fix the problem as well.

I will include a fix in the next version.

mldangelo:
A fix for this was deployed in [0.68.0](https://github.com/promptfoo/promptfoo/releases/tag/0.68.0) Please let us know if you run into any other issues and thanks for reporting it!

PR #1029, opened 2024-06-27
fix: handle when stdout columns are not set

typpo:
Related to #1028

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -11,3 +11,7 @@ export const REMOTE_APP_BASE_URL =
   process.env.NEXT_PUBLIC_PROMPTFOO_BASE_URL ||
   process.env.PROMPTFOO_REMOTE_APP_BASE_URL ||
   `https://app.promptfoo.dev`;
+
+// Maximum width for terminal outputs.
+export const TERMINAL_MAX_WIDTH =
+  process.stdout.columns && process.stdout.columns > 10 ? process.stdout.columns - 10 : 120;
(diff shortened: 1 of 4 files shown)


mldangelo (approved):

src/constants.ts
```diff
@@ -11,3 +11,7 @@ export const REMOTE_APP_BASE_URL =
   process.env.NEXT_PUBLIC_PROMPTFOO_BASE_URL ||
   process.env.PROMPTFOO_REMOTE_APP_BASE_URL ||
   `https://app.promptfoo.dev`;
+
+// Maximum width for terminal outputs.
+export const TERMINAL_MAX_WIDTH =
+  process.stdout.columns && process.stdout.columns > 10 ? process.stdout.columns - 10 : 120;
```
fvdnabee, line 17: Suggestion:
```typescript
process?.stdout?.columns && process?.stdout?.columns > 10 ? process?.stdout?.columns - 10 : 120;
```

fvdnabee (commented):
This change breaks the nextui for me (running via `npm exec promptfoo view`), as src/constants.ts is imported [from nextui](https://github.com/promptfoo/promptfoo/blob/main/src/web/nextui/src/app/eval/ResultsView.tsx#L2) and `process` is undefined in this case. See my naive suggestion...

src/constants.ts
typpo: Thanks for flagging! d7c12480d2530e2d247d274b3b7a1a45f4085199 should solve

PR #1035, opened 2024-06-28
feat: dereference promptfoo test files

fvdnabee:
Yaml test files that are included aren't dereferenced, only the 'main' promptfooconfig.yaml is dereferenced.  

promptfooconfig.yaml:
```yaml
prompts: [prompt1.txt]
providers: [openai:gpt-3.5-turbo]
tests:
  - tests.yaml
```
tests.yaml:
```yaml
- vars:
    language: French
    input: Hello world
  assert:
    - $ref: 'asserts.yaml#/assertionTemplates/noAIreference'
    - $ref: 'asserts.yaml#/assertionTemplates/startsUpperCase'
```
asserts.yaml:
```yaml
assertionTemplates:
    noAIreference:
      - type: llm-rubric
        value: does not describe self as an AI, model, or chatbot
    startsUpperCase:
      - type: javascript
        value: output[0] === output[0].toUpperCase()
```

Note you can't define yaml snippets that are to be referenced in promptfoo test files, as a test file is expected to define a list of test cases. Therefor the yaml snippets are in a separate file in the example. One solution is to add a top-level key (e.g. `tests`) to tests.yaml and group test cases under it; then arbitrary yaml snippets could be added to the tests file.

### DIFF
--- a/src/testCases.ts
+++ b/src/testCases.ts
@@ -1,3 +1,4 @@
+import $RefParser from '@apidevtools/json-schema-ref-parser';
 import { parse as parseCsv } from 'csv-parse/sync';
 import * as fs from 'fs';
 import { globSync } from 'glob';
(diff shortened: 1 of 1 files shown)


mldangelo (approved):
@fvdnabee thank you so much for contributing this!

Issue #1040, opened 2024-06-28
`TypeError: testCases is not iterable`


ZanzyTHEbar:
My `promptfooconfig.yml`:
```yml
description: "verify"

prompts:
  - file://verify.prompt.txt
providers:
  - file://verify.provider.yml
tests: tests/*.test.yml
```
Test file:
```yml
  description: "Validation of the code changes"
  vars:
    preBuildState: file://assets/shared/pre_build.go
    changes: file://assets/valid/changes.md
    postBuildState: file://assets/valid/post_build.go
    diffs: file://assets/valid/diff.txt
  assert:
    - type: is-json
    - type: is-valid-openai-tools-call
    - type: javascript
      value: |
        var args = JSON.parse(output[0].function.arguments)
        return !(
          args.hasSyntaxErrors ||
          args.hasRemovedCodeErrors ||
          args.hasDuplicationErrors ||
          args.hasReferenceErrors            
        )

```

Previous to my update this worked fine. I had all of my tests in my `/tests` folder. After update - this no longer works.

```
Cache is disabled.
Loaded provider openai:gpt-4o from verify.provider.yml
/home/daofficialwizard/.local/share/pnpm/global/5/.pnpm/promptfoo@0.67.0_@aws-sdk+client-bedrock-runtime@3.598.0_@azure+identity@4.2.1_@azure+openai-_o4xgjjjr6zw7j63g3trvuro4vm/node_modules/promptfoo/dist/src/testCases.js:177
                for (const testCase of testCases) {
                                       ^

TypeError: testCases is not iterable
    at loadTestsFromGlob (/home/daofficialwizard/.local/share/pnpm/global/5/.pnpm/promptfoo@0.67.0_@aws-sdk+client-bedrock-runtime@3.598.0_@azure+identity@4.2.1_@azure+openai-_o4xgjjjr6zw7j63g3trvuro4vm/node_modules/promptfoo/dist/src/testCases.js:177:40)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async resolveConfigs (/home/daofficialwizard/.local/share/pnpm/global/5/.pnpm/promptfoo@0.67.0_@aws-sdk+client-bedrock-runtime@3.598.0_@azure+identity@4.2.1_@azure+openai-_o4xgjjjr6zw7j63g3trvuro4vm/node_modules/promptfoo/dist/src/main.js:134:25)
[...]

PR #1041, opened 2024-06-28
fix: handle imported single test case

typpo:
I've seen this come up several times, including in #1040

### DIFF
--- a/src/testCases.ts
+++ b/src/testCases.ts
@@ -182,6 +182,10 @@ export async function readTests(
         throw new Error(`Unsupported file type for test file: ${testFile}`);
       }
       if (testCases) {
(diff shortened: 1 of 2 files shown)


src/testCases.ts
```diff
@@ -182,6 +182,10 @@ export async function readTests(
         throw new Error(`Unsupported file type for test file: ${testFile}`);
       }
       if (testCases) {
+        if (!Array.isArray(testCases)) {
+          logger.warn(`Assuming ${testFile} contains a single test case.`);
```
mldangelo:
```suggestion
```

mldangelo (approved):

PR #1043, opened 2024-06-29
chore: enable '@typescript-eslint/no-use-before-define' linter rule 

mldangelo:
As part of trying to resolve our circular import errors, I realized that we should also enable this rule. It makes no changes to the functionality of the codebase but resolves a number of circular issues.


typpo:
Nice

PR #1045, opened 2024-06-29
feat: add `author` field to evals

typpo:
For now, author is the email set by `promptfoo config`

image


src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -235,15 +245,62 @@ export default function ResultsView({
...
+              message="Eval id copied to clipboard"
+            />
+          </>
+        )}
+        {author && (
+          <Chip
+            size="small"
+            label={
+              <>
+                <strong>Author:</strong> {author}
```
mldangelo: It would be great if there was a little hover tooltip that told people to set the author using the cli if it's not set.

src/accounts.ts
```diff
@@ -8,3 +8,7 @@ export function getUserEmail(): string | undefined {
 export function setUserEmail(email: string) {
   writeGlobalConfig({ account: { email } });
 }
+
+export function getAuthor(): string {
+  return getUserEmail() || 'Unknown';
```
mldangelo: consider not using a sentinal value and just dealing with possible undefined string. 
```suggestion
export function getAuthor(): string | undefined {
  return getUserEmail();
```

src/types.ts
```diff
@@ -713,6 +713,7 @@ export interface ResultsFile {
   createdAt: string;
   results: EvaluateSummary;
   config: Partial<UnifiedConfig>;
+  author: string;
```
mldangelo:
```suggestion
  author: string | undefined,
```

mldangelo (commented):
This is a great contribution for team evals! thank you

mldangelo (approved):

Issue #1050, opened 2024-06-30
[Question]Does function have specified names for custom python scripts


landhu:
According to https://www.promptfoo.dev/docs/providers/python and https://github.com/promptfoo/promptfoo/blob/main/examples/python-provider/provider.py

Both scripts have call_api function, Do all custom python scripts have to have this function?
If I have multiple functions in py files, how do promptfoo execute them?

mldangelo:
You can call functions whatever you want. I recommend using the array syntax with a function name and a label. 

```yaml
prompts: 
  - id: file_name1.py:function_in_file
    label: your label
  - id: file_name1.py:another_function
    label: a different label 
  - id: file2.py:yet_another_function_name.py
    label: your third prompt label
```

if your files look like:

#### file_name1.py
```py
def function_in_file(context):
    pass

def another_function(context):
    pass
```

#### file2.py
```py
def yet_another_function_name(context):
    pass

def unused_function(): 
    pass

def unused_function2(context): 
    pass
```
the unused functions will not be executed because they are not defined in your prompts array. 

I will update the docs to make this more clear. Please feel free to re-open the issue if this doesn't resolve your question and feel free to reach out or join our discord https://discord.gg/gHPS9jjfbs if you have any additional questions

typpo:
Right now, python providers only support the `call_api` function.

@mldangelo it would be useful if the provider supports the `file.py:function_name` syntax like for prompts.

typpo:
Implementing support for this in #1053 

For example:
```yaml
providers:
  - python:file.py:my_function
```

PR #1053, opened 2024-07-01
feat: support calling specific function from python provider

typpo:
This takes advantage of @mldangelo's work on `parsePathOrGlob` to implement #1050


mldangelo (approved):
thank you for turning this around so quickly

PR #1085, opened 2024-07-04
feat(redteam): `extra-jailbreak` plugin that applies jailbreak to all probes

typpo:
Also modifies the attacker prompt for iterative jailbreaks


typpo:
I confirmed the changes to iterative jailbreak are good - improves harmful jailbreaks by about 10% in my travel agent redteam test case.

PR #1087, opened 2024-07-04
feat(webui): show metrics as % in column header

typpo:
Not the prettiest, but a lot more useful.

image


mldangelo (approved):
This is a great change! Thanks @typpo

PR #1100, opened 2024-07-05
fix(openai): add missing `gpt-4-turbo-2024-04-09`

aloisklink:
See https://platform.openai.com/docs/models/continuous-model-upgrades

As of 2024-07-05 (today), `gpt-4-turbo` points to `gpt-4-turbo-2024-04-09`:

image

Pricing is at https://openai.com/api/pricing/:

image


mldangelo (approved):
Thank you so much!

PR #1102, opened 2024-07-06
feat: add Zod schema validation for providers in promptfooconfig

mldangelo:
This PR is the first in a series aimed at gradually converting our types to Zod schemas, with the goal of providing more user-friendly validation for errors in `promptfooconfig' files. This update introduces a warning message for invalid keys in the `providers` section of the config file. Future work will include adding validations for assertions and test cases, and eventually covering the entire `TestSuiteConfig` with Zod schemas. To test this update, simply add an unexpected key to one of the config files and verify that a warning message is displayed.


src/main.ts
```diff
@@ -130,12 +131,38 @@ async function resolveConfigs(
     logger.error(chalk.red('You must provide at least 1 prompt'));
     process.exit(1);
   }
+
+  // Convert providers to array if not already an array
+  if (config.providers && typeof config.providers === 'string') {
+    config.providers = [config.providers];
+  }
```
typpo: Is this the right place for this?  It means CLI (`main`) will be inconsistent  with package (`index`)

src/main.ts
```diff
@@ -130,12 +131,38 @@ async function resolveConfigs(
...
   if (!config.providers || config.providers.length === 0) {
-    logger.error(
-      chalk.red('You must specify at least 1 provider (for example, openai:gpt-3.5-turbo)'),
-    );
+    logger.error(chalk.red('You must specify at least 1 provider (for example, openai:gpt-4o)'));
     process.exit(1);
   }
+  if (Array.isArray(config.providers)) {
```
typpo: in what cases is providers not an array?

src/main.ts
```diff
@@ -130,12 +131,38 @@ async function resolveConfigs(
...
+          .join('\n');
+        const providerString = typeof provider === 'string' ? provider : JSON.stringify(provider);
+        logger.warn(
+          chalk.yellow(
+            dedent`
+              Provider: ${providerString} encountered errors during schema validation:
+
+                ${errors}
+
+              This may have have unintended consequences.` + '\n',
```
typpo: Instead of "this may have unintended consequences", which is not actionable, maybe something like "Please double check your configuration"?

typpo (commented):
Very cool change

src/main.ts
mldangelo: You're right. Even with `providers: openai:gpt-3.5-turbo` it gets coerced into an array above this point. I need to narrow the type if I don't do this.

typpo (approved):

PR #1117, opened 2024-07-09
fix(redteam): use final prompt in moderation instead of original

typpo:
This solves a problem of moderation false positives when the iterative redteam comes up with a benign final prompt.  In order to do this, this PR plumbs through the entire provider response to assertions


typpo:
Increases pass rate by 13% on `--plugins harmful,experimental-jailbreak` travel agent example by reducing moderation false positives

Issue #1127, opened 2024-07-10
Providing different prompts per model always takes arguments of last provider


yakuhzi:
Since version `0.69.1` (up to `0.70.0`), providing different prompts per model is broken, if the same provider id is used. It will always take the prompt arguments that are configured for the last provider.

Here is a reproducing example:

promptfooconfig.yaml:
```
prompts:
    - prompt1
    - prompt2

providers:
    - id: 'python:provider.py'
      label: 'Provider 1'
      prompts:
          - prompt1
    - id: 'python:provider.py'
      label: 'Provider 2'
      prompts:
          - prompt2

tests:
    - description: 'Assert equals'
      assert:
          - type: equals
            value: 'output'
```

provider.py:
```
def call_api(prompt: str, options: dict, context: dict):
    print(f"Prompt: {prompt}")
    return { "output": "output" }
```

When you run `LOG_LEVEL=debug promptfoo evaluate`, you can see that it will print out `Prompt: prompt2` two times, instead of `Prompt: prompt1` for the first provider, and `Prompt: prompt2` for the second. Note that this is a regression from `0.69.0`, where this case worked without problems. Note also that when there is no `label` provided for the providers, then it also didn't work in older versions (e.g. `0.67.0`).

The issue is maybe caused by https://github.com/promptfoo/promptfoo/pull/1106, removing the code commented with `Hack(ian): Override id with label`.

My use case is: I have always the same custom provider, but depending on the prompt, I will do different things. My current workaround is to create separate providers for each prompt, but this is an overhead I want to avoid in the future.

typpo:
Thanks for flagging this. I think you're probably right that the id/label
override was necessary in that code. We'll get a fix out quickly.

PR #1134, opened 2024-07-10
fix(evaluator): ensure unique prompt handling with labeled and unlabeled providers

mldangelo:
This PR addresses an issue where providing different prompts per model is broken if the same provider id is used. It ensures that prompts are correctly mapped and processed based on either the provider's label or id. It resolves https://github.com/promptfoo/promptfoo/issues/1127 to test: 

```yaml
prompts:
  - prompt1
  - prompt2

providers:
  - id: 'python:provider.py'
    label: 'Provider 1'
    prompts:
      - prompt1
  - id: 'python:provider.py'
    label: 'Provider 2'
    prompts:
      - prompt2

tests:
  - description: 'Assert equals'
    assert:
      - type: equals
        value: 'prompt1'
```

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -525,13 +525,14 @@ class Evaluator {
     for (const provider of testSuite.providers) {
       for (const prompt of testSuite.prompts) {
         // Check if providerPromptMap exists and if it contains the current prompt's label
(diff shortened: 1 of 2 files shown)


typpo:
Nice fix!

PR #1141, opened 2024-07-12
feat(redteam): predict injectVars

mldangelo:
This infers the injectVars in redteam scripts using a regex and generates a warning if more than one is found.


src/redteam/purpose.ts
```diff
@@ -1,28 +1,117 @@
...
+        (prompt) => dedent`
+      <prompt>
+      {% raw %}${prompt}{% endraw %}
+      </prompt>`,
+      )
+      .join('\n')}
+    
+    Given the above prompts, output a JSON object that specifies the "system purpose" of the application in a single phrase and lists any variables in an array. Ensure the output is in JSON format. Start with { "intent": "
+  `);
+  // sometimes the llm returns ```json, find the first { and the last }
```
typpo: use `extractJsonObjects` in util

PR #1147, opened 2024-07-13
feat: new Chart type on the eval page of web ui

YingjiaLiu99:
feature details:
1. y-axis: pass rate, x-axis: eval index(from 1st to nth)
2. hover to the point will show the pass rate, evalId, prompt brief, and date
3. A gray line that links each evals' top scores to show the trend.
4. The current eval's points are highlighted in red.
5. Click the points will open a new window of that eval
6. If there is only one eval in the dataset (i.e. only the 1st of the eval), the graph won't show up


src/web/nextui/.env.example
```diff
@@ -3,3 +3,4 @@ DATABASE_URL="postgresql://..."
 NEXT_PUBLIC_PROMPTFOO_WITH_DATABASE=1
 NEXT_PUBLIC_SUPABASE_URL=https://placeholder.promptfoo.dev
 NEXT_PUBLIC_SUPABASE_ANON_KEY=abc123
+NEXT_TELEMETRY_DISABLED=1
```
will-holley: Please remove

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +419,238 @@ function MetricChart({ table }: ChartProps) {
...
+    if (j === 3 && k !== 13) {
+      return i + 'rd';
+    }
+    return i + 'th';
+  }
+
+  useEffect(() => {
+    if (config) {
+      const newDatasetId = createHash('sha256').update(JSON.stringify(config.tests)).digest('hex');
+      setDatasetId(newDatasetId);
```
will-holley: As `setDatasetId` is only called once (this `useEffect`-scoped initializer), you can reduce the `useState` and `useEffect` into a single `[useMemo](https://react.dev/reference/react/useMemo)`.

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +419,238 @@ function MetricChart({ table }: ChartProps) {
...
+            const pointData = (topMostElement.element as any).$context.raw as Point;
+            const evalId = pointData.metadata.evalId;
+            window.open(`/eval/?evalId=${evalId}`, '_blank');
+          }
+        },
+      },
+    });
+  }, [table, evalId, config, datasetId, dataset]);
+
+  return <canvas ref={lineCanvasRef} style={{ maxHeight: '300px', cursor: 'pointer' }}></canvas>;
```
will-holley: nit: prefer a self-closing tag here i.e. `<canvas ... />`

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +419,238 @@ function MetricChart({ table }: ChartProps) {
...
+    const fetchDataset = async () => {
+      if (!config) return;
+      const datasetId = createHash('sha256').update(JSON.stringify(config.tests)).digest('hex');
+      fetch(`${await getApiBaseUrl()}/api/evals/${datasetId}`)
+        .then((response) => response.json())
+        .then((data: { data: TestCasesWithMetadata[] }) => {
+          setShowLineChart(data.data[0].count > 1);
+        });
+    };
+    fetchDataset();
```
will-holley: Evaluate the config here i.e.`if(!!config) fetchDataset();`

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: Thank you for your advice. Fixed!

src/web/nextui/.env.example
YingjiaLiu99: got it!

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: got it, fixed!

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: Thank you for your feedback! 
But I think an early return in the function fetchDataset() like this `if (!config) return;` will help me get rid of warning: 'config' is possibly 'null'. at this line: `const datasetId = createHash('sha256').update(JSON.stringify(config.tests)).digest('hex');`

I can still add the check here if you prefer:
if (config) {
      fetchDataset();
 }
 
 The double negation `!!` will make the code fail style check, I believe they work the same way in this context.
 
Here's the full code that I am talking about:
ac9104ef6a54d30e14750fc61cee8df

will-holley: I see that you're fetching the dataset twice and already have `datasetId` memoized: try moving the fetch call into a separate function which accepts the `datasetId: string` argument.  Something like:

```tsx
const fetchDataset = async (id: string) => {
  try {
   const res = await fetch(`${await getApiBaseUrl()}/api/evals/${id}`);
   return response.json();
  } catch (error) {
    // handle the error!
  }
}

const showLineChart = useMemo(async () => {
  if (datasetId) {
    const data = await fetchDataset(datasetId);
    setShowLineChart(data.data[0].count > 1);
  } else return false;
}, [datasetId])
```

YingjiaLiu99: Thank you for your feedback! Just update the code with the following changes:
Extract the fetchDataset() function and call it separately in LineChart and ResultChart components. 
Use memo to store the datasetId in ResultChart and pass it down to LineChart to avoid redundant use.

src/util/index.ts
```diff
@@ -873,6 +873,12 @@ export async function getEvalFromId(hash: string) {
   return undefined;
 }
 
+export async function getEvalsByDatasetId(datasetId: string): Promise<TestCasesWithMetadata[]> {
```
typpo: `listPreviousResults` now has a `datasetId` filter option - I think you can just use that!

additionally, the `/api/results` supports this query param in a recent change - you may find it useful https://github.com/promptfoo/promptfoo/blob/main/src/web/server.ts#L88

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +420,239 @@ function MetricChart({ table }: ChartProps) {
...
+  try {
+    const res = await fetch(`${await getApiBaseUrl()}/api/evals/${id}`);
+    return await res.json();
+  } catch (err) {
+    const error = err as Error;
+    throw new Error(`Fetch dataset data using given id failed: ${error.message}.`);
+  }
+};
+
+function LineChart({ table, evalId, config, datasetId }: ChartProps) {
```
typpo: nit: let's call this something more descriptive like `PerformanceOverTimeChart` or similar

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +420,239 @@ function MetricChart({ table }: ChartProps) {
...
+  const datasetId = useMemo(() => {
+    if (config) {
+      return createHash('sha256').update(JSON.stringify(config.tests)).digest('hex');
+    }
+  }, [config]);
+
+  useMemo(async () => {
+    if (datasetId) {
+      const data = await fetchDataset(datasetId);
+      setShowLineChart(data.data[0].count > 1);
```
typpo: On some existing ones, I'm getting `TypeError: Cannot read properties of undefined (reading 'count')` because the `/api/evals/:id` endpoint is returning `[]`

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +420,239 @@ function MetricChart({ table }: ChartProps) {
...
+    } else {
+      setShowCharts(false);
+    }
+  }, [datasetId]);
 
-  const { table } = useStore();
-  if (!table || !showCharts || table.head.prompts.length < 2) {
+  if (!table || !config || !showCharts || table.head.prompts.length < 2) {
```
typpo: It's probably ok to ignore this for now, but the new chart you've added is interesting with 1 prompt, so it could be nice to show it in that case.

typpo (commented):
Really cool change!  A few comments

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: got it!

src/util/index.ts
YingjiaLiu99: got it

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: adjusted, it should fix it

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +426,269 @@ function MetricChart({ table }: ChartProps) {
...
+    }
+  }, [datasetId, recentEvals]);
 
-  const { table } = useStore();
-  if (!table || !showCharts || table.head.prompts.length < 2) {
+  if (!table || !config || !showCharts) {
+    return null;
+  }
+
+  console.log(datasetId);
```
typpo: nit: remove

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +426,269 @@ function MetricChart({ table }: ChartProps) {
...
+
+  if (table.head.prompts.length < 2) {
+    if (showPerformanceOverTimeChart) {
```
typpo: looks like we can combine these two conditions?

src/web/server.ts
```diff
@@ -209,6 +210,11 @@ export async function startServer(
     res.json({ data: await getTestCases() });
   });
 
+  app.get('/api/evals/:datasetId', async (req, res) => {
```
typpo: This route would make more sense as `/api/evals?datasetId=...`

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +426,269 @@ function MetricChart({ table }: ChartProps) {
...
+    score: string;
+  }
+
+  interface Point {
+    x: number;
+    y: number;
+    metadata: PointMetadata;
+  }
+
+  const fetchDataset = async (id: string) => {
```
typpo: let's move this and `ordinalSuffixOf` out of the component function

src/util/index.ts
```diff
@@ -851,6 +851,12 @@ export async function getEvalFromId(hash: string) {
...
+  const allTestCases = await getTestCases();
+  const matchingEvals = allTestCases.filter((testCase) => testCase.id == datasetId);
```
typpo: let's use `getTestCasesWithPredicate`

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +426,269 @@ function MetricChart({ table }: ChartProps) {
...
+
+        allPoints.push(point);
+
+        if (!highestScoreMap.has(evalIdKey) || highestScoreMap.get(evalIdKey)!.y < point.y) {
+          highestScoreMap.set(evalIdKey, point);
+        }
+      }
+    });
+
+    if (evalIdToIndex.size == 1) {
```
typpo: I'm not actually able to get this chart to show as it always hits this condition.  Elsewhere you are restricting this chart to a single prompt and a single dataset.  I might be missing something here but that seems like why?

src/web/nextui/src/app/eval/ResultsCharts.tsx
```diff
@@ -415,13 +426,269 @@ function MetricChart({ table }: ChartProps) {
...
+  const datasetId = useMemo(() => {
+    if (config) {
+      return createHash('sha256').update(JSON.stringify(config.tests)).digest('hex');
+    }
+  }, [config]);
+
+  useMemo(async () => {
+    if (datasetId) {
+      const filteredEvals = recentEvals.filter((evaluation) => evaluation.datasetId === datasetId);
+      console.log(filteredEvals);
```
typpo: nit: remove

typpo (commented):
we're making progress!

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: ✅

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: ✅

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: ✅

src/web/nextui/src/app/eval/ResultsCharts.tsx
YingjiaLiu99: updated a new version

src/util/index.ts
YingjiaLiu99: ✅

src/web/server.ts
YingjiaLiu99: updated

PR #1148, opened 2024-07-13
feat(webui): Ability to compare evals

typpo:
Some TODOs remain
- [x] Overwrite the description to indicate it's combined
- [x] Disable save functionality (grading, scoring, editing description) so we don't overwrite 

https://github.com/user-attachments/assets/eb89d888-8a97-4961-b0f7-4b0dadbca793


src/web/nextui/src/app/eval/CompareEvalMenuItem.tsx
```diff
@@ -0,0 +1,87 @@
...
+  };
+
+  return (
+    <>
+      <Tooltip title="Combine this eval with another eval run" placement="left">
+        <MenuItem onClick={handleOpenDialog}>
+          <ListItemIcon>
+            <CompareIcon fontSize="small" />
+          </ListItemIcon>
+          <ListItemText>Compare with another eval</ListItemText>
```
mldangelo, line 72: nit, it would be great if this button was grayed out / disabled if there are no other evals to compare to

src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -141,6 +158,43 @@ export default function ResultsView({
...
+          ...row,
+          outputs: [...row.outputs, ...(comparisonTable.body[index]?.outputs || [])],
+        })),
+      };
+
+      // Update the state with the combined table
+      setTable(combinedTable);
+
+      // Update other relevant state if needed
+      setConfig({
```
mldangelo, line 195: We should update the `Outputs` section of the table to denote which column corresponds to which eval

mldangelo (commented):
This is a great feature. It looks good and works well. I left two nits. It would also be great to have permalinks. Thanks @typpo

src/web/nextui/src/app/eval/CompareEvalMenuItem.tsx
typpo: going to punt on this bc currently matching evals with datasets are loaded on click.  possibly this can change in the future, especially as we develop other features that depend on similar functionality

PR #1153, opened 2024-07-15
feat: add image classification example and xml assertions

mldangelo:
This PR introduces the following changes:

1. Adds a new example for image classification using the Fashion MNIST dataset. The example demonstrates prompt engineering, configuration, and evaluation of AI models for image analysis. Data is loaded from a generated CSV which follows a more traditional ml practitioner workflow. 
  
2. **Add XML Assertion Types:**
   - Introduced `is-xml` and `contains-xml` assertion types.
   - Updated documentation to include examples and usage for the new XML assertion types.
   - Implemented XML validation and parsing using `fast-xml-parser`.
   - Added unit tests for XML validation and assertions.

### How to Test:
1. Follow the instructions in `examples/image-classification/README.md` to set up and run the image classification example.
2. Run the unit tests to ensure all new and existing tests pass:

NOTE: 
- I wanted to modify the example to display the images as markdown / HTML but am having trouble rendering them when loaded from a csv. I will revisit this soon. One potential solution that I would like to experiment with is adding input transforms on test vars in addition to output transforms. 

CC @lacom @kaiktl


src/assertions.ts
```diff
@@ -362,6 +427,24 @@ export async function runAssertion({
     };
   }
 
+  if (baseType === 'is-xml' || baseType === 'contains-xml') {
+    const requiredElements =
+      typeof renderedValue === 'string'
+        ? renderedValue.split(',').map((el) => el.trim())
```
typpo: values can be objects, so it would be much nicer to have an object of the form:

```yaml
value:
  requiredElements:
    - elt1
    - elt2
```

PR #1154, opened 2024-07-15
feat(redteam): specify the default number of test cases to generate per plugin

mldangelo:
This adds a `numTests` property to `RedteamGenerateOptions` and relevant interfaces to allow specifying the number of test cases to generate per plugin. Note that some plugins like PII and harmful have subcategories where we actually generate numTests * number of subcategories per plugin. Also note that this is not deterministic and sometimes with very low numbers or very high numbers we can get unexpected results. As a follow on I would like to add a grading step to the harmful test case generation and possibly even add a loop to ensure we generate the number of requested test cases.


typpo (approved):
Good change overall, however

1. One thing to keep in mind is that a reason for the varying numbers is because different tests have a different "hit" rate.  Or in other words, not all tests are created equal.

    For example, a single attempt at competitor mention is usually enough to predict whether the rest will fail.  However, hijacking attempts are much more varied and unpredictable.

2. Because the tests are not weighted, this change could cause a large change in the final "brand" and "legal" risk scores.'

My thinking is that we can live with (2), but for (1) let's gut check whether we think 5 is enough for some of the more interesting/trickier ones that were originally at 20 like hijacking.  In the worse case, we could do `n * 4` for now.

PR #1155, opened 2024-07-15
refactor(cli): move cli commands into individual files

mldangelo:
This PR refactors the command structure by moving individual commands out of `src/main.ts` and into their own separate files in `src/commands/`. We can start to use this structure to lazy load dependencies in the future and make the CLI snappier. 

To test: Ensure each command operates correctly from its new location.


typpo (approved):
Thanks for doing this

Issue #1120, opened 2024-07-09
Issue while using message arrays as a prompts with HTTP provider


vaibhav-toddleapp:
I was tring to use the HTTP provider with message arrays as a prompts. And nunjucks templating method "renderString" produces invalid JSON string, which is being parsed using `JSON.parse` ( line: [providers/http.ts:46](https://github.com/promptfoo/promptfoo/blob/main/src/providers/http.ts#L46) ) and throws error of `unexpected token`

### Example:
```javascript
const aiServiceProvider = {
    id: "http://localhost/test",
    config: {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: {
        input: {
          messagesArray: "{{ prompt }}",
          modelName: "openai:gpt-4-turbo"
        },
      },
      responseParser: (response) => {
        return response;
      },
   },
  };
  ......
  ......
  ......
  const prompts = [
      [
        {
          role: "system",
          content: "You are speaking to a time traveler. Begin your conversation.",
        },
        {
          role: "user",
          content: "I want to know about switzerland. which places should I visit?",
        },
      ],
    ];
```
### Produced JSON by renderString

```json
{"method":"POST","headers":{"Content-Type":"application/json"},"body":{"input":{"messagesArray":"[
  {
    "role": "system",
    "content": "You are speaking to a time traveler. Begin your conversation."
  },
  {
    "role": "user",
    "content": "I want to know about switzerland. which places should I visit?"
  }
]","modelName":"openai:gpt-4-turbo"}}}
```

Looks like messagesArray: "{{ prompt }}" not is not resulting with proper quote escaping. 

Please correct me if my implementation is wrong :)

PR #1156, opened 2024-07-15
fix(http provider): fix JSON prompt escaping in HTTP provider and add LM Studio example

mldangelo:
This PR  fixes an issue where JSON prompts were not being escaped properly when rendered by Nunjucks in HTTP providers and introduces a new example for using PromptFoo with LM Studio. This should resolve https://github.com/promptfoo/promptfoo/issues/1120 

How to Test:

1. Follow the setup instructions in `examples/lm-studio/README.md`.

CC @luiscosio @vaibhav-toddleapp

### DIFF
--- a/examples/lm-studio/README.md
+++ b/examples/lm-studio/README.md
@@ -1,0 +1,52 @@
+# LM Studio Example with PromptFoo
+
+This example demonstrates how to use PromptFoo with LM Studio for prompt evaluation. It showcases configuration for interacting with the LM Studio API using a locally hosted language model.
+
+## Prerequisites
+
+1. **LM Studio**: Install LM Studio from [lmstudio.ai](https://lmstudio.ai/).
+2. **Model**: Download the `bartowski/gemma-2-9b-it-GGUF` model in LM Studio.
+
+## Setup
+
+1. **Start LM Studio Server**:
+
+   - Open LM Studio and load the `bartowski/gemma-2-9b-it-GGUF` model.
+   - Start a local server to host the model (usually at `http://localhost:1234`).
+
+2. **Configure PromptFoo**:
+
+   Create a `promptfooconfig.yaml` file with the following content:
+
+   ```yaml
+   providers:
+     - id: 'http://localhost:1234/v1/chat/completions'
+       config:
+         method: 'POST'
+         headers:
+           'Content-Type': 'application/json'
+         body:
+           messages: '{{ prompt }}'
+           model: 'bartowski/gemma-2-9b-it-GGUF'
+           temperature: 0.7
+           max_tokens: -1
+         responseParser: 'json.choices[0].message.content'
+   ```
+
+   Note that you can view the specific configuration for each model within LM Studio's examples in the server tab.
+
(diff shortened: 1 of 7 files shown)


src/providers/http.ts
```diff
@@ -43,14 +53,17 @@ export class HttpProvider implements ApiProvider {
...
-      JSON.parse(
-        nunjucks.renderString(stringifiedConfig, {
-          prompt,
-          ...context?.vars,
-        }),
-      );
-
+    const renderedConfigString = nunjucks.renderString(stringifiedConfig, {
+      // nunjucks does not escape JSON strings, so we need to escape them manually
+      prompt: isValidJson(prompt) ? prompt.replace(/"/g, '\\"') : prompt,
```
vaibhav-toddleapp: If it is valid JSON then we are escaping the strings. but if it is not valid, then we are using `prompt` as it is..?

mldangelo: Thanks for the quick comment. Good question. Some prompts are defined like: 

```yaml
prompts:
  - Tell me about {{topic}}
```
This is designed to cover both scenarios. I'll make sure it's covered in the test cases for you. If there's something I am missing please let me know and I'll attempt to fix it.

mldangelo: Note that in order to get this format to work you have to modify the messages parameter

```yaml
description: 'LM Studio Eval Example'
prompts:
  - Tell me about {{topic}}
providers:
  - id: 'http://localhost:1234/v1/chat/completions'
    config:
      method: 'POST'
      headers:
        'Content-Type': 'application/json'
      body:
        messages: '[{"role": "user", "content": "{{ prompt }}"}]'
        model: 'bartowski/gemma-2-9b-it-GGUF'
        temperature: 0.7
        max_tokens: -1
      responseParser: 'json.choices[0].message.content'
tests:
  - vars:
      topic: bananas
```

vaibhav-toddleapp: Got it. will give it a try.. thanks for providing fix.

examples/lm-studio/README.md
```diff
@@ -0,0 +1,52 @@
+# LM Studio Example with PromptFoo
```
typpo: nit: `Promptfoo` (here and below)

src/providers/http.ts
```diff
@@ -20,6 +21,15 @@ function createResponseParser(parser: any): (data: any) => ProviderResponse {
   return (data) => ({ output: data });
 }
 
+function isValidJson(str: string): boolean {
```
typpo: might be useful to put in util

src/providers/http.ts
typpo: `JSON.stringify(output).slice(1, -1)` should be a more foolproof way to do this (this is what we do for outputs in `matchers.ts`) because it will handle newlines, tabs, and backslashes too.

typpo (approved):
Very nice fix

src/providers/http.ts
mldangelo: thank you!

PR #1158, opened 2024-07-15
feat(dataset): allow provider customization for dataset generation

mldangelo:
- Implemented a default provider for dataset generation, which can be overwritten.
- Split `src/commands/generate.ts` into `src/commands/generate/redteam.ts` and `src/commands/generate/dataset.ts` and standardized the interfaces.
- Moved `getNunjucksEngine` from `util.ts` to `util/templates.ts` for better organization. Moved injectVars used by redteam and now dataset into templates as well.
- Moved `util.ts` to `util/index.ts` - in the future I will break this file down even more. 

To test:
- Run `npm run build:watch` and `npm link`.
- `promptfoo generate dataset` will now use the default provider (typically gpt-4o).
- Change your default provider to use Anthropic by unsetting your OPENAI key and setting an Anthropic one, then re-run the above command.
- Run `promptfoo generate dataset --provider openai:chat:gpt3-5-turbo` to generate with a non-default provider.
- Repeat with `--provider anthropic:messages:claude-3-5-sonnet-20240620`.

Note: In the future, we should do a better job of showcasing which providers are available. Some providers allow us to force JSON mode. I am not using that when `--provider` is passed in, but they tend to do a good job even without it.


typpo (approved):
LGTM

You can also remove the `info` flag at the top of the doc and add `providers` to the table at the bottom of the doc: https://www.promptfoo.dev/docs/configuration/datasets/

PR #1168, opened 2024-07-16
feat: manual input provider

typpo:
Small provider that makes it easy to test some stuff out.  I found it useful to "manually" redteam applications that don't have an accessible API

### DIFF
--- a/src/providers/manualInput.ts
+++ b/src/providers/manualInput.ts
@@ -1,0 +1,51 @@
+import inquirer from 'inquirer';
+import { ApiProvider, ProviderResponse } from '../types';
+
+export class ManualInputProvider implements ApiProvider {
+  private config: { multiline?: boolean };
+
+  constructor(config: { multiline?: boolean } = {}) {
+    this.config = config;
+  }
+
+  id() {
+    return 'manual-input';
+  }
+
+  async callApi(prompt: string): Promise<ProviderResponse> {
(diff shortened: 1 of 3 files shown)


src/providers/manualInput.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class ManualInputProvider implements ApiProvider {
+  private config: { multiline?: boolean };
+
+  constructor(config: { multiline?: boolean } = {}) {
+    this.config = config;
+  }
+
+  id() {
+    return 'manual-input';
```
will-holley: nit: `promptfoo:manual-input` or convert to a static method and set `} else if (providerPath === ManualInputProvider.id()) {` within `providers.ts:353` in order to enforce consistency.

src/providers/manualInput.ts
```diff
@@ -0,0 +1,51 @@
+import inquirer from 'inquirer';
+import { ApiProvider, ProviderResponse } from '../types';
+
+export class ManualInputProvider implements ApiProvider {
+  private config: { multiline?: boolean };
```
will-holley: nit: DRY `{ multiline?: boolean }` (b/c constructor usage)

mldangelo (approved):
Thanks!  I wasn't able to get the multi-line feature to work but it may be my terminal.

PR #1170, opened 2024-07-16
chore(mistral): add codestral-mamba

streichsbaer:
Changelog:
- Add support for Codestral-mamba

Note: The pricing is not public yet, I'll update it once it's known.


mldangelo:
Thanks so much @streichsbaer! We can merge this as is now and as you said, update with token costs when they are known.

PR #1172, opened 2024-07-17
fix(redteam): set harmCategory label for harmful tests

typpo:
1. Correctly set `harmCategory` label
2. Put unaligned attempts (really bad stuff) before the standard provider attempts (less bad stuff)


mldangelo (approved):
Thank you for fixing this!

PR #1198, opened 2024-07-18
fix: base path for custom filter resolution

### LINKED ISSUE #1197: Custom Nunjucks filter not found for eval executed outside config dir
The issue can be reproduced by attempting to run the `nunjucks-custom-filters` example from the root of the repository:

```
$ promptfoo eval -c examples/nunjucks-custom-filters/promptfooconfig.yaml

node_modules/nunjucks/src/lib.js:32
    err = new Error(old.message);
          ^
Template render error: (unknown path)
  Error: filter not found: allcaps
    at Object._prettifyError (node_modules/nunjucks/src/lib.js:32:11)
    at node_modules/nunjucks/src/environment.js:464:19
    at Template.root [as rootRenderFunc] (eval at _compile (node_modules/nunjucks/src/environment.js:527:18), <anonymous>:20:3)
    at Template.render (node_modules/nunjucks/src/environment.js:454:10)
    at Environment.renderString (node_modules/nunjucks/src/environment.js:313:17)
    at renderPrompt (src/evaluator.ts:259:21)
    at Evaluator.runEval (src/evaluator.ts:317:34)
    at processEvalStep (src/evaluator.ts:719:30)
    at node_modules/async/dist/async.js:151:38
    at replenish (node_modules/async/dist/async.js:448:21)
```

The issue does not occur if the eval is run from the same directory as the config file (`examples/nunjucts-custom-filters`).

onyck:
Fixes #1197

`cliState.basePath` is not set at the time of config resolution, so it cannot be depended upon for filter resolution. Instead, we can pass `basePath` to `readFilters` just like `readTest`, `readPrompts`, et al.


mldangelo (approved):
Thank you!

PR #1200, opened 2024-07-19
docs: add guide for comparing GPT-4o vs GPT-4o-mini

mldangelo:
- Create new guide 'GPT-4o vs GPT-4o-mini: Benchmark on Your Own Data'
- Update OpenAI provider example to use gpt-4o-mini
- Add comparison screenshot


typpo (approved):
One for the ages

PR #1207, opened 2024-07-20
site: yaml/json config validator for promptfoo configs

typpo:
https://github.com/user-attachments/assets/a7bd3f07-b320-409d-9f84-e3fca7b4e714

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -105,7 +105,8 @@
         "shx": "^0.3.4",
         "ts-node": "^10.9.2",
         "typescript": "^5.5.3",
-        "typescript-eslint": "^7.16.0"
+        "typescript-eslint": "^7.16.0",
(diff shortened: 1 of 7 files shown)


mldangelo (approved):
This is excellent. Please consider adding this to the style check above where we run `npm run format`

```yaml 
      - name: Generate JSON Schema and Check for Changes
        run: |
          npm run jsonSchema:generate
          if [[ -n $(git status --porcelain) ]]; then
            echo "Changes detected after generating JSON schema and formatting. Please commit these changes."
            git diff
            exit 1
          fi
```

typpo:
I would prefer a different approach:

it would be super annoying to break CI when pretty much _any_ type is touched (because realistically we will never remember to regenerate)

maybe instead we can regenerate it nightly and automatically commit it.

adding the validator to `/setup/` is a good idea - will do that separately

PR #1208, opened 2024-07-20
feat(redteam): improve test generation logic and add batching

mldangelo:
This allows us to generate a potentially very large number of tests for a specific use case

- Enhance generateTests method in PluginBase to support batching
- Add deduplication of generated prompts
- Implement retry mechanism with max 2 (?) consecutive retries
- Add sampling for cases when more prompts are generated than requested
- Update tests to cover new functionality
- Add debug logging throughout the generation process
- Fix num-tests option parsing in redteam command

### DIFF
--- a/src/redteam/plugins/base.ts
+++ b/src/redteam/plugins/base.ts
@@ -12,13 +12,6 @@ export default abstract class PluginBase {
    */
   protected abstract template: string;
 
-  /**
-   * Constructs a new PluginBase instance.
-   *
-   * @param provider - The API provider used to call the API.
-   * @param purpose - The purpose of the system being tested.
-   * @param injectVar - The variable to inject into the test cases.
-   */
   constructor(
     protected provider: ApiProvider,
     protected purpose: string,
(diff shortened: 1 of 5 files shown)


src/redteam/plugins/base.ts
```diff
@@ -39,24 +42,66 @@ export default abstract class PluginBase {
...
-
-    return prompts.map((prompt) => ({
+    const batchSize = 20;
+    let allPrompts: string[] = [];
+    let consecutiveRetries = 0;
+    const maxConsecutiveRetries = 2;
+
+    const generateBatch = async (batchN: number): Promise<string[]> => {
+      logger.debug(`Generating batch of ${batchN} prompts`);
+      const { output: generatedPrompts } = await this.provider.callApi(
```
typpo, line 102: would it make sense to put retry logic in the provider, not the plugin?

mldangelo: Good idea. I think this may be difficult to implement given provider customization. I would like to punt this to a future date but will make changes if requested.

typpo (approved):

PR #1215, opened 2024-07-21
feat(redteam): settings dialog

typpo:
Settings button in top right, right now lets you toggle full percentages in risk card

image


mldangelo (approved):
looks good!

PR #1218, opened 2024-07-22
fix: gpt-4o-mini price

kongmoumou:
https://openai.com/api/pricing/#:~:text=gpt%2D4o%2Dmini-,US%240.150%20/,1M%20input%20tokens,-US%240.075%20/

⬆️ According to OpenAI pricing page, gpt-4o-mini price should be 0.00015/0.0006 per 1K tokens.


mldangelo (approved):
Thank you @kongmoumou!

PR #1221, opened 2024-07-22
site: improve human readability of validator errors

typpo:
This is only a slight improvement, but at least includes actionable information.

Before
image

After
image


mldangelo (approved):
Thank you!

PR #1228, opened 2024-07-22
docs: Full rag example

typpo:


examples/rag-full/requirements.txt
```diff
@@ -0,0 +1,117 @@
+aiohttp==3.9.5
```
mldangelo, line 1: nit, freeze only what you need. Let's only list what you need: langchain-community, langchain_openai, langchain, openai, etc.

mldangelo (commented):
This is great work. Please consider putting the data in a public bucket (R2?). The pdfs more than double the size of the repo from 20MB to 43MB. You can check with 

`git ls-tree -r --long HEAD | awk '$4 ~ /^[0-9]+$/ {s+=$4} END {print s}' | numfmt --to=iec-i --suffix=B`

examples/rag-full/requirements.txt
typpo: I don't remember what I installed anymore

PR #1249, opened 2024-07-23
feat: extension hook api

### LINKED ISSUE #1022: Support setup/teardown commands for tests
When testing agents on complex situations, we need to setup the test first and then tear it down. We currently implement this with a custom provider that reads `setup` and `teardown` variables and interprets them as bash scripts.

In our case, we are testing an [AIOps agent](http://github.com/robusta-dev/holmesgpt/), so setup/teardown involves creating IT problems and opening Jira tickets for them.

However, using a custom provider to do this has a number of limitations:

* When running promptfoo with --repeat there isn't a great way to identify that a test was already setup and therefore doesn't need further setup. We've tried to work around this, but its ugly. In our case, setting up tests can take up to 90 seconds, so adding 90 second latency for each eval is painful.
* We have a similar problem with parallelism when using promptfoo Scenarios and multiple tests on the same test variables. Here too, we want to setup the test once, not one time per test.

For the above reasons, having first-class support for setup/teardown in promptfoo would let us solve this more elegantly than using a custom provider. I believe this type of requirement will become increasingly relevant for other promptfoo users as people test agents on more and more complex cases.

For our needs, having a setup/teardown setting in each test would be sufficient. It can be interpreted as a bash script which is templated with vars. There should also be a setting to control if the setup/teardown happen once or many times when the test is run multiple times (due to either --repeat or Scenarios)

aantn:
Fixes #1022.

This still needs documentation but is otherwise ready.


aantn:
Please note that I pushed a nearly complete version from my perspective. It just needs docs.

aantn:
Well, I suppose I should add typescript support too :)

And please note that there is a bug with the json stringify hence my change there. I didn't look closely enough at the code to understand what it is trying to fix and hence whether I am damaging something else there or not. But there is definitely a bug in the previous version which causes values to be omitted the 2nd time they appear

mldangelo:
thank you so much @aantn for opening this PR and including an example, I can help get it over the line (fix the bug you described, add tests) for you tomorrow.

typpo:
:eyes: very interesting change. Sorry I'm coming in a bit late on this one. Thank you for implementing so thoughtfully. Will try this out on my end and give it a proper review!

aantn:
All good, I saw the news announcement and I am fairly certain you have a number of other items demanding attention right now. Congrats and good luck.

aantn:
To give a more realistic use case for this, here is how we are using it to test an AI agent that finds the root cause of IT problems:

```
  - vars:
      name: HighLatencyForCustomerCheckout
      setup: |
        kubectl apply -f https://raw.githubusercontent.com/robusta-dev/kubernetes-demos/b1190fd820d140167edfc2df16071b2e8a8ea6a2/slow-rds-query/manifest.yaml
        sleep 60
      teardown: |
        kubectl delete -f https://raw.githubusercontent.com/robusta-dev/kubernetes-demos/b1190fd820d140167edfc2df16071b2e8a8ea6a2/slow-rds-query/manifest.yaml
      execute_agent_cmd: "holmes ask 'Why is there high latency with the customer-orders deployment?'"
      expected: The result mentions a timeout for test-mysql-instance.jd9a9jfad99j.us-east-2.rds.amazonaws.com
```

...in this case, `execute_agent_cmd` is handled by a customer provider. (Note that it is inconvenient to properly handle setup/teardown in a custom provider because the provider runs once per test repeat + with re-entrancy.)

src/util/json.ts
```diff
@@ -16,7 +16,7 @@ export function safeJsonStringify(value: any, prettyPrint: boolean = false): str
     (key, val) => {
       if (typeof val === 'object' && val !== null) {
         if (cache.has(val)) {
-          return;
+          return val;
```
typpo: qq, why was this necessary?  I think it may default the purpose of removing circular references.

test/config.test.ts
```diff
@@ -120,7 +121,8 @@ describe('readConfigs', () => {
       description: 'test2',
       providers: ['provider2'],
       prompts: ['prompt2'],
-      tests: ['test2'],
+      extensions: [],
+      tests: ['test2', 'test2'],
```
typpo: Any idea why this test duplication is happening?

typpo (commented):
Hey @aantn, thanks so much for this change!  And thank you for your patience.  It makes a lot of sense.  Just a couple quick qs.

test/config.test.ts
aantn: Not sure what happened here, but it was not an intentional change on my side.

src/util/json.ts
aantn: The original function is wrong. It dropped duplicate objects during serialization. E.g. the evals_prepared hook was run as

```
Running Python wrapper with args: 
["evals_prepared",
    {
        "evals": [
            {
                "delay": 0,
                "provider": {
                    "env": {},
                    "modelName": "gpt-4o",
                    "config": {
                        "basePath": "examples/custom-plugins"
                    }
                },
                "prompt": {
                    "raw": "Rephrase this in French: {{body}}",
                    "label": "examples/custom-plugins/prompts.txt: Rephrase this in French: {{body}}"
                },
                "test": {
                    "vars": {
                        "body": "Hello world",
                        "setup": "touch setup1",
                        "teardown": "touch teardown1"
                    },
                    "assert": [],
                    "options": {}
                },
                "nunjucksFilters": {},
                "rowIndex": 0,
                "colIndex": 0,
                "repeatIndex": 0,
                "evaluateOptions": {
                    "showProgressBar": false,
                    "maxConcurrency": 4,
                    "repeat": 1,
                    "eventSource": "cli"
                }
            },

            // multiple objects are missing here like provider
            {
                "delay": 0,
                "prompt": {
                    "raw": "Rephrase this like a pirate: {{body}}",
                    "label": "examples/custom-plugins/prompts.txt: Rephrase this like a pirate: {{body}}"
                },
                "test": {},
                "rowIndex": 0,
                "colIndex": 1,
                "repeatIndex": 0
            },
            ...
]}]
```

Note that multiple objects are missing from the 2nd evals array entry.
 
But as mentioned on the PR comments, I'm probably defeating the purpose of this function with the fix. This real solution is neither here nor there.

mldangelo:
@aantn Thanks so much for proposing this feature and getting it ready for review. I apologize for the delay in getting it in. I ran with your proposed changes and made a few updates to the PR. If you have time, please check this out and let me know if it still satisfies your requirements. In particular, pay attention to the context available to each hook and the hook names. I am happy to make more changes if this does not meet your needs. Thanks again!

typpo (approved):
Exciting change!!

PR #1250, opened 2024-07-23
feat: ability to share a specific eval

typpo:
`promptfoo share` still defaults to most recent eval, but now you can `promptfoo share <evalId>`

### DIFF
--- a/src/commands/share.ts
+++ b/src/commands/share.ts
@@ -21,27 +21,28 @@ export function shareCommand(program: Command) {
         });
         await telemetry.send();
 
-        const createPublicUrl = async () => {
-          let results;
-          if (evalId) {
-            results = (await readResult(evalId))?.result;
-            if (!results) {
-              logger.error(`Could not load results for eval ID ${evalId}.`);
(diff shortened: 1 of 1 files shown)


src/commands/share.ts
```diff
@@ -4,57 +4,67 @@ import readline from 'readline';
...
-        createPublicUrl();
-      } else {
-        const reader = readline.createInterface({
-          input: process.stdin,
-          output: process.stdout,
-        });
+        if (cmdObj.yes || process.env.PROMPTFOO_DISABLE_SHARE_WARNING) {
+          createPublicUrl();
+        } else {
+          const reader = readline.createInterface({
```
mldangelo, line 47: nit, move the question to after where you look up the evalId. User experience is not good if you enter `promptfoo share asdasd`, confirm you want to share it, and then realize it doesn't exist.

mldangelo (approved):
Great feature!

src/commands/share.ts
typpo: great point

PR #1251, opened 2024-07-24
fix: defaultTest and scenario assert merging

onyck:
This PR fixes two interactions among assertions defined in `defaultTest`, a scenario's config, and a scenario's test case:

1. If `defaultTest.assert` was defined and neither the scenario's config nor the scenario's test case defined `assert`, all assertions from `defaultTest.assert` were run twice for that test case.
2. If a test case in a scenario defined `assert`, none of the assertions in the scenario's `config.assert` were run for that test case.


mldangelo:
Thank you @onyck!

PR #1259, opened 2024-07-24
chore: warn when a var is named assert

typpo:
I've helped people with this multiple times, most recently https://discord.com/channels/1153767083381903389/1265701055195381781, so I think it might be worth adding a warning

### DIFF
--- a/src/testCases.ts
+++ b/src/testCases.ts
@@ -224,6 +224,26 @@ export async function readTests(
     }
   }
 
+  if (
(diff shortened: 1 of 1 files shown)


mldangelo (approved):
Thank you

src/testCases.ts
```diff
@@ -224,6 +224,26 @@ export async function readTests(
     }
   }
 
+  if (
```
mldangelo, line 227: consider adding test coverage for this!

PR #1264, opened 2024-07-24
feat(webui): add yaml upload button

typpo:
This will require absolute paths for the provider, and cannot reference other files (yaml must be self contained)


mldangelo:
CC @rrbanda 

Note that this will not work with relative paths yet for custom providers.

rrbanda:
@typpo  i m redeploying with the image `ghcr.io/promptfoo/promptfoo:sha-fd7aff8 `

will the upload button be available in this version?

rrbanda:
i m now able to see upload button 
Screenshot 2024-07-24 at 7 23 52 PM

PR #1267, opened 2024-07-24
fix: treat .cjs and .mjs files as javascript vars

onyck:
I believe `.cjs` and `.mjs` files are accepted everywhere else custom javascript is supported.


mldangelo (approved):
Thank you so much for implementing this @onyck! I'll test it and get a release out for you today.

PR #1268, opened 2024-07-24
docs: fix javascript configuration guide variable example

onyck:
Javascript variables are expected to return an object.


mldangelo (approved):
thank you!

PR #1269, opened 2024-07-24
fix: add xml tags for better delineation in `llm-rubric`, reduce `harmful` plugin false positives

typpo:


src/prompts/grading.ts
```diff
@@ -22,7 +22,7 @@ Rubric: Does not speak like a pirate
   },
   {
     role: 'user',
-    content: 'Output: {{ output }}\nRubric: {{ rubric }}',
+    content: '<Output>\n{{ output }}\n</Output>\n<Rubric>\n{{ rubric }}\n</Rubric>',
```
mldangelo, line 25: this is a great change

Issue #1265, opened 2024-07-24
sending non-string objects through the yaml file


rishimalhotra2:
Hi, i would like to send in complex non-string payload in my post request. Specifically, I want to create a list of lists in my chat_history and send that through my post request. Here is what that looks like

```yaml
name: 'my Tests'
prompts: 
  - 'file://../model/prompts/*.txt'
providers:
  - id: 'http://localhost:5000/api/query_no_stream'
    config:
      method: 'POST'
      headers:
        'Content-Type': 'application/json'
      body:
        question: "{{question}}"
        chat_history: 
        conversation_id: empty
        scenario: my_scenario
      responseParser: 'json.answer'
tests:
  - vars:
      question: What's the eather like?
      chat_history:
        - - 'hi'
          - 'hello'
        - - 'how are you'
          - 'I am good'
      reference: No ref
```

However, when i do this, the server receives the chat_history as `{'[object Object]': None}` instead of the nested lists. I know you use nunjucks as the templating engine. I tried using `{{chat_history | dump}}` but that tends to send the whole thing as a string instead of nested lists.

What's the easiest way to accomplish what I'm trying to achieve?

mldangelo:
@rishimalhotra2 take a look at https://www.promptfoo.dev/docs/configuration/chat/ and let us know if it doesn't solve your issue.

typpo:
#1274 introduces a fix for this behavior, making it possible to nest objects in the request.

PR #1274, opened 2024-07-25
fix: improve handling of json objects in http provider

typpo:
Related to #1265 

Nested objects are supported and should be passed to the `dump` function.

```yaml
providers:
  - id: 'https://example.com/generateTranslation'
    config:
      body:
        // highlight-start
        messages: '{{messages | dump}}'
        // highlight-end
        model: '{{model}}'
        translate: '{{language}}'

tests:
  - vars:
      // highlight-start
      messages:
        - role: 'user'
          content: 'foobar'
        - role: 'assistant'
          content: 'baz'
      // highlight-end
      model: 'gpt-4'
      language: 'French'
```

Note that any valid JSON string within `body` will be converted to a JSON object.

### DIFF
--- a/site/docs/providers/http.md
+++ b/site/docs/providers/http.md
@@ -40,6 +40,36 @@ tests:
 
 If not specified, HTTP POST with content-type application/json is assumed.
 
+### Nested objects
+
+Nested objects are supported and should be passed to the `dump` function.
+
+```yaml
+providers:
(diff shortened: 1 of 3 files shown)


mldangelo (approved):
Thanks!

typpo:
Flagged this as a breaking change in the merge because it converts JSON strings into JSON objects

PR #1275, opened 2024-07-25
refactor(redteam): rename strategies and improve type safety

mldangelo:
- Remove the existing `jailbreak` strategy that added harmful plugins if it is not specified. 
- Rename `experimental-jailbreak` to `jailbreak` strategy. This runs on all specified plugins.
- Rename 'experimental-tree-jailbreak' with 'jailbreak:tree' strategy
- Update strategy descriptions and implementation
- Improve type safety for plugins and strategies
- Update documentation and schema to reflect changes


typpo (approved):
looks good - i assume you tested the affected strategies

Issue #349, opened 2023-12-19
Does the provider support loading from files?


jojoLockLock:
## Does the provider support loading from files?

my `promptfooconfig.yaml` file
```yaml
prompts: [prompt.txt]
providers:
  - id: openai:chat:gpt-3.5-turbo-0613
    config:
      functions:
        [
          {
            'name': 'get_current_weather',
            'description': 'Get the current weather in a given location',
            'parameters':
              {
                'type': 'object',
                'properties':
                  {
                    'location':
                      {
                        'type': 'string',
                        'description': 'The city and state, e.g. San Francisco, CA',
                      },
                    'unit': { 'type': 'string', 'enum': ['celsius', 'fahrenheit'] },
                  },
                'required': ['location'],
              },
          },
        ]
```

I want to have this grammar

**promptfooconfig.yaml**

```yaml
prompts: [prompt.txt]
providers:
  - id: openai:chat:gpt-3.5-turbo-0613
    config:
      functions:
         - file://./get_current_weather.json
```

**get_current_weather.json**

```json 
{
    'name': 'get_current_weather',
    'description': 'Get the current weather in a given location',
    'parameters':
      {
        'type': 'object',
        'properties':
          {
            'location':
              {
                'type': 'string',
                'description': 'The city and state, e.g. San Francisco, CA',
              },
            'unit': { 'type': 'string', 'enum': ['celsius', 'fahrenheit'] },
          },
        'required': ['location'],
 }
```

typpo:
You can load providers from file, which is pretty similar:

```yaml
providers:
  - file://./provider.json
```

provider.json
```json
{
  "id": "openai:chat:gpt-3.5-turbo-0613",
  "config": {
    "functions": [
      {
        "name": "get_current_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },
            "unit": {
              "type": "string",
              "enum": ["celsius", "fahrenheit"],
              "description": "The unit in which to return the temperature"
            }
          },
          "required": ["location"]
        }
      }
    ]
  }
}
```

I can see why it might be useful to load individual config keys from file...

PR #1279, opened 2024-07-25
fix: support provider json filepath

typpo:
Supposedly doable back in December https://github.com/promptfoo/promptfoo/issues/349#issuecomment-1862066974

Possibly lost in a refactor 🤔


mldangelo (approved):
This feels fragile - should we mention it in a doc  or modify one of the examples to cover it as well?

typpo:
good idea - done

PR #1296, opened 2024-07-26
docs(custom-api): add examples for embedding and classification providers

### LINKED ISSUE #1294: Improve Documentation regarding embedding providers
I needed to implement a custom embedding provider due to the API utilized by the company i work for, for specific embedding models. It took quite some time to discover that I required the method call EmbeddingApi instead of simply callApi in my custom JavaScript provider. I believe this should be noted in Providers -> Custom Javascript. If desired, I would be pleased to create a pull request with an example and the necessary documentation.

Luca-Hackl:
Added 'Example - Embedding' to custom-api.md, closes #1294


mldangelo:
Thank you for improving our docs @Luca-Hackl!

PR #1299, opened 2024-07-26
fix: report cached tokens from assertions

onyck:
When rerunning tests with model-graded metrics, the final `tokensUsed.cached` was less than `tokensUsed.total` which made it seem like model-graded metrics were being re-executed unnecessarily.


mldangelo:
Thank you @onyck!

PR #1302, opened 2024-07-27
chore(webui): persist column selection in evals view

mldangelo:
Implements localStorage persistence for selected columns

1. Open the /evals page in the browser.
2. Select or deselect some columns.
3. Refresh the page and verify that the column selection state is preserved.
5. Check the browser's localStorage to confirm the presence of the persisted data.
6. Test with different eval IDs to ensure unique storage keys are used.
7. Verify that the column visibility in the table reflects the persisted state.

In the future we can consider moving this to persist across datasets instead of columns.

CC @kaiktl @will-holley

### DIFF
--- a/src/web/nextui/src/app/eval/ResultsView.tsx
+++ b/src/web/nextui/src/app/eval/ResultsView.tsx
@@ -85,12 +85,6 @@ export default function ResultsView({
   } = useResultsViewStore();
   const { setStateFromConfig } = useMainStore();
 
-  const currentEvalId = evalId || defaultEvalId || 'default';
-  const currentColumnState = columnStates[currentEvalId] || {
-    selectedColumns: [],
-    columnVisibility: {},
-  };
-
   const [searchText, setSearchText] = React.useState(searchParams?.get('search') || '');
   const [debouncedSearchText] = useDebounce(searchText, 1000);
   const handleSearchTextChange = (text: string) => {
@@ -256,6 +250,12 @@ export default function ResultsView({
     ],
     [hasAnyDescriptions, head.vars, head.prompts],
   );
+
+  const currentEvalId = evalId || defaultEvalId || 'default';
+  const currentColumnState = columnStates[currentEvalId] || {
+    selectedColumns: [],
+    columnVisibility: {},
+  };
 
   const updateColumnVisibility = React.useCallback(
     (columns: string[]) => {
--- a/src/web/nextui/src/app/eval/ResultsView.tsx
+++ b/src/web/nextui/src/app/eval/ResultsView.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { useEffect, useCallback, useMemo } from 'react';
 import { REMOTE_API_BASE_URL, REMOTE_APP_BASE_URL } from '@/../../../constants';
 import { getApiBaseUrl } from '@/api';
 import { useStore as useMainStore } from '@/state/evalConfig';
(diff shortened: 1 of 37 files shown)


src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -61,6 +62,19 @@ interface ResultsViewProps {
   defaultEvalId?: string;
 }
 
+// Utility function to safely parse JSON
```
typpo: This seems like it could be named better as it restricts the return value to arrays

src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -222,44 +247,71 @@ export default function ResultsView({
...
+        setSelectedColumns(allColumns);
+        updateColumnVisibility(allColumns);
+      }
+    };
+
+    loadSelectedColumns();
+  }, [allColumns, updateColumnVisibility, localStorageKey]);
+
+  useEffect(() => {
+    localStorage.setItem(localStorageKey, JSON.stringify(debouncedSelectedColumns));
```
typpo: Prefer using `./store`, which is where we store and fetch other settings.

typpo (changes requested):

src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -80,9 +81,17 @@ export default function ResultsView({
     evalId,
     inComparisonMode,
     setInComparisonMode,
+    columnStates,
+    setColumnState,
   } = useResultsViewStore();
   const { setStateFromConfig } = useMainStore();
 
+  const currentEvalId = evalId || defaultEvalId || 'default';
```
mldangelo: instead of `|| 'default';` we could disable the feature (or random id). I think this only affects supabase but I am not sure.

src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -212,6 +221,17 @@ export default function ResultsView({
     [table.body],
   );
 
+  const promptOptions = useMemo(() => {
```
typpo: nit: we use `React.useMemo`, `React.useEffect` in most places elsewhere in the codebase. Let's prefer to keep the `React` namespace

src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -222,44 +242,55 @@ export default function ResultsView({
...
+      });
+      setColumnState(currentEvalId, {
+        selectedColumns: columns,
+        columnVisibility: newColumnVisibility,
+      });
+    },
+    [allColumns, setColumnState, currentEvalId],
+  );
+
+  useEffect(() => {
```
typpo: nit: maybe move `currentColumnState` definition down here to above where it's used?

src/web/nextui/src/app/eval/store.ts
```diff
@@ -79,10 +88,22 @@ export const useStore = create<TableState>()(
...
+          columnStates: {
+            ...prevState.columnStates,
+            [evalId]: state,
+          },
+        })),
     }),
     {
       name: 'ResultsViewStorage',
       storage: createJSONStorage(() => storage),
+      partialize: (state) => ({
```
typpo: Pretty sure this omits all the other properties from storage, which is not what we want.  The entire state object is stored by default.

typpo (approved):

PR #1304, opened 2024-07-27
chore(cli): add option to strip auth info from shared URLs

### LINKED ISSUE #1184: Do not display Basic Auth username and password after creating share URL
In the CLI, if I define my hosted promptfoo server using a Basic Auth username and password via environment variable, e.g.:

```
export PROMPTFOO_REMOTE_API_BASE_URL=https://my_user:supersecret@example.com
```

When I then run `npx promptfoo share -y`, the confirmation message afterward says:

```
View results: https://my_user:supersecret@example.com/eval/f:12345678-abcd-cafe-9999-123450abcdef
```

Could that URL be stripped of the username/password? (I think the Node `URL` class has an `origin` property that could be useful here.)

mldangelo:
- Implement stripAuthFromUrl function to remove username and password from URLs
- Add --show-auth option to share command to optionally display full URL
- Update createShareableUrl to use stripAuthFromUrl by default
- Add unit tests for stripAuthFromUrl function

Closes #1184

### DIFF
--- a/src/commands/share.ts
+++ b/src/commands/share.ts
@@ -1,10 +1,17 @@
 import chalk from 'chalk';
 import { Command } from 'commander';
 import readline from 'readline';
+import { URL } from 'url';
 import logger from '../logger';
 import { createShareableUrl } from '../share';
 import telemetry from '../telemetry';
+import { ResultsFile } from '../types';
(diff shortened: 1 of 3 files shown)


typpo:
Very cool.  cc @efung

typpo (approved):

PR #1305, opened 2024-07-27
fix(redteam): decrease false positives in hallucination grader

mldangelo:
- Expand HallucinationGrader rubric with detailed evaluation criteria
- Notably, add support for hypothetical/fictional scenarios in hallucination check

This passes my vibe for a few redteam evals but I suspect that we can make this a lot better as we build out more example redteam scenarios.


typpo (approved):

typpo:
Very nice

Issue #1227, opened 2024-07-22
bug: vars in defaultTest do not appear in result chart columns


orange-anjou:
The variable is still taken into account in the tests because and i can see that it's executed two times (once for each value in "system_prompt").
However, the variable does not appear in an other column, thus making the readability quite complex without digging in the prompt itself.

> Maybe this is a voluntary behavior since the variables for `phi3_template_prompt` and `gemma_template_prompt` aren't in the columns as well. Maybe there is a difference to make between single "global vars" and the ones that imply the generation of multiple tests

I'm not sure from where it's coming from... My best bet is that it's something around here that messed up but it's quite a bold guess : https://github.com/promptfoo/promptfoo/blob/c50eadd8de250521dd4d17762c8878443821f40f/src/commands/eval.ts#L130.
From what I understand, it's `summary` that either not contain enough information, or the information is not used down the line.

Config used (with some "useless" stuff removed for clarity) : 
```yaml
prompts:
  - id: "{{phi3_template_prompt}}"
    label: phi3_template_prompt
  - id: "{{gemma_template_prompt}}"
    label: gemma_template_prompt

defaultTest:
  vars:
    phi3_template_prompt: file://./prompts/model-templates/phi3.txt
    gemma_template_prompt: file://./prompts/model-templates/gemma.txt
    system_prompt: [
      file://./prompts/agent-system-prompts/api-event-expert/phi3.v001.txt,
      file://./prompts/agent-system-prompts/api-event-expert/phi3.v999.txt
    ]

providers:
  - id: ollama:gemma:2b
    prompts:
      - gemma_template_prompt
  - id: ollama:chat:phi3
    prompts:
      - phi3_template_prompt

tests:
  - vars:
      user_prompt: "Is there a Youth Sports event happening on 25/12/2024?"
    assert:
      - type: equals
        value: { "category": "Sports", "start_datetime": "25/12/2024 00:00", "end_datetime": "25/12/2024 23:59" }
```

Output : 
<img width="938" alt="image" src="https://github.com/user-attachments/assets/8d9cec4e-aa26-426f-9b94-35288c4d982b">

Expected output:
<img width="945" alt="image" src="https://github.com/user-attachments/assets/c6041e25-937c-415d-bfeb-7983c0b4971e">

The expected output has been made with some trickery :
```yaml
# Step 1 :
# Rename defaultTest.vars.system_prompt ==> system_prompt_global

# Step 2 : 
# Add this as a var of the test
[...]

PR #1306, opened 2024-07-28
fix: show default vars in table

typpo:
Related to #1227

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -534,7 +534,10 @@ class Evaluator {
     const table: EvaluateTable = {
       head: {
         prompts,
-        vars: Array.from(varNames).sort(),
+        vars: [
+          ...Object.keys(testSuite.defaultTest?.vars || {}).sort(),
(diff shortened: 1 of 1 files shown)


src/evaluator.ts
```diff
@@ -534,7 +534,10 @@ class Evaluator {
     const table: EvaluateTable = {
       head: {
         prompts,
-        vars: Array.from(varNames).sort(),
+        vars: [
+          ...Object.keys(testSuite.defaultTest?.vars || {}).sort(),
+          ...Array.from(varNames).sort(),
+        ],
```
mldangelo, line 540: nit, we should consider sorting everything together or nothing at all if we want to preserve template order.

imagine things  are specified as 

```
default
 - Bdefault
 - Adefault
 
 vars: 
 - A 
 - B
```

Do we want 
```
- Adefault
- Bdefault 
- A
- B
```

or 

```
- A
- Adefault
- B
- Bdefault 
```

or something else

mldangelo (approved):

PR #1316, opened 2024-07-29
fix: chain provider and test transform

fvdnabee:
On the main branch, only one transform is ever executed. When both a provider and a test transform are defined, only the test transform is called. After this PR, both transformed are called. First the provider transform is called and then the test transform.


mldangelo:
Thank you @fvdnabee for fixing this. I am going to clean it up a little (write docs) and get it merged and deployed.

PR #1325, opened 2024-07-30
feat: split types vs validators for prompts, providers, and redteam

typpo:
Several improvements and fixes:
- Split zod and types into `validators` and `types` respectively for better separation of concerns
- Add support for `redteam.provider` object
- Throw an error when redteam configuration doesn't parse (this was swallowed silently)
- Simplify writing of config file


src/types.ts
```diff
@@ -1,5 +1,5 @@
 import { z } from 'zod';
-import { RedteamAssertionTypes, redteamConfigSchema } from './redteam/types';
+import { RedteamAssertionTypes, RedteamConfig } from './redteam/types';
```
typpo: This change was necessary to avoid circular dependency, now that `redteam.provider` supports `ProviderSchema`

mldangelo (approved):
Thank you!

PR #1336, opened 2024-08-01
fix: numTests calculation for previous evals

onyck:
The "# Tests" displayed by the eval picker in the UI was incorrect for evals that used scenarios.


typpo:
Thank you!

PR #1337, opened 2024-08-01
Correct positive example in DEFAULT_GRADING_PROMPT

tbuckley:
The reason why "Hello world" contains a greeting is that it has the word "Hello", not "world"


mldangelo (approved):
Thank you so much @tbuckley! Let's see how many people complain file issues about their evals changing after this goes in. 

I started to read through https://github.com/tbuckley/evals-pwa Please feel free to book time on our calendars if you ever want to talk!

PR #1374, opened 2024-08-05
fix: Update "Edit Comment" dialog background for the dark mode

dimusic:
Small improvement for the "Edit Comment" dialog with the dark mode

Before:
dark-theme-before

After:
dark-theme-after


mldangelo:
Thank you so much @dimusic! We'll get this released today

PR #1378, opened 2024-08-06
fix!(webui): use pluginId instead of harmCategory var

typpo:
Frontend was still dependent on harmCategory var.  This is a breaking change as the frontend now looks at metadata instead of harmCategory.


mldangelo:
Closing - was resolved by another PR.

PR #1381, opened 2024-08-06
fix: Do not ignored derived metrics

tomquist:
This seems to be a regression. Derived metrics are entirely ignored atm.


mldangelo:
Thank you so much @tomquist

PR #1382, opened 2024-08-06
fix: add indexes for sqlite

typpo:
adds indexes to support all operations in `util`, also fixes some drizzle stuff after drizzle upgrade


mldangelo:
thank you!

PR #1391, opened 2024-08-07
feat: ability to attach configs to prompts

typpo:
Lets you do things like this:

```yaml
prompts:
  - label: 'Foo'
    raw: "You are a helpful math tutor. Solve {{problem}}"
    config:
      response_format:
        type: json_schema
        json_schema: ...
```

or this

```yaml
prompts:
  - label: 'Foo'
    raw: "You are a helpful math tutor. Solve {{problem}}"
    config:
      response_format: file://schema1.json
```

Also includes some openai updates:
- Expand support for `response_format`
- Adds latest GPT-4o
- Adds support for `refusal` message field

TODO
- [x] Update tests
- [x] Update docs


nicocomumu:
Hello
Could you please provide an example of how to use this new feature?
I have extracted my schema in a json file
```
{
  "name": "itinerary_response",
  "strict": true,
  "schema": {
    "type": "object",
    "properties": {
      "itineraries": {
        "type": "array",
        ...
}
```
        
But I cannot find the correct syntax to use this file
```
prompts:
  - id: prompts/p1.json
    config:
      response_format:
        type: json_schema
        json_schema: ./schemas/itinerary.json
```

I still get error :
Invalid type for 'response_format.json_schema': expected an object, but got a string 

I also tried
json_schema: file://schemas/itinerary.json

Can someone help?
Thanks

Issue #1441, opened 2024-08-14
Add Support for Global Pass Rate Threshold


MatanYadaev:
Currently, promptfoo returns an exit code of 100 when there is at least one failing test, which causes the CI to fail even if your failure appetite is greater than just one failing test.

# Proposal:

To enhance promptfoo's utility in CI environments, I propose the following improvements:

1. **Global Pass Rate Threshold:** Introduce a configuration option to set a minimum pass rate (e.g., 90%) for the entire configuration file.

2. **Conditional Exit Code:** If the actual pass rate falls below the specified threshold, promptfoo should exit with a non-zero code, indicating a failure in the CI pipeline.

mldangelo:
@MatanYadaev We support a version of this (but it may not be working). Check out: https://github.com/promptfoo/promptfoo/blob/main/site/docs/usage/command-line.md?plain=1#L165 and https://github.com/promptfoo/promptfoo/blob/main/src/commands/eval.ts#L276 

Let me know if it doesn't resolve your immediate issue. The conditional threshold is a great idea and we would welcome a PR if you would like to work on it.

mldangelo:
Hi @MatanYadaev, I wanted to let you know that the feature has been released in [version 0.78.3](https://github.com/promptfoo/promptfoo/releases/tag/0.78.3). As discussed in #1443, you can set the global pass rate threshold per config using the env: property. Thank you for suggesting this feature! Please feel free to reach out if you run into any other issues or have additional feedback.

Thanks again!

PR #1443, opened 2024-08-14
feat(eval): add global pass rate threshold

mldangelo:
This PR addresses #1441 It introduces a configurable pass rate threshold for evals. The command now exits with PROMPTFOO_FAILED_TEST_EXIT_CODE if the pass rate falls below the specified threshold. When the threshold is not set, the existing behavior (0 failures allowed) is maintained.

- Add PROMPTFOO_PASS_RATE_THRESHOLD environment variable (default: 100%)
- Modify exit behavior to consider the pass rate threshold
- Update documentation to reflect new functionality

CC @MatanYadaev


MatanYadaev:
Amazing @mldangelo ! WDYT about allowing the user to configure the pass rate threshold inside the configuration, instead of in a environment variable? In my use case, I have many different promptfoo configurations, each one has different pass rate threshold.

mldangelo:
> Amazing @mldangelo! WDYT about allowing the user to configure the pass rate threshold inside the configuration, instead of in an environment variable? In my use case, I have many different promptfoo configurations, each one has a different pass rate threshold.

@MatanYadaev This is a great idea! Luckily, it is already supported! In your promptfoo config files, you can set:

```yaml
env:
  PROMPTFOO_PASS_RATE_THRESHOLD: 76.54
```

Thanks again for the product feature suggestion.

PR #1447, opened 2024-08-15
chore: use temporary file for python interprocess communication

enkoder:
In our application we've got lots of logging hitting stdout and is making it very difficult to use the custom python provider. This PR _attempts_ to move to using a file as the way to report results/output back to the TS layer. This should be more resilient than depending upon capturing stdout. 

I updated the tests that were failing with the change to pythonUtils.ts.


enkoder:
Thanks for fixing up this pr @mldangelo! It would be really sweet if y'all had some default vscode workspace config where lints and autoformatters worked out of the box.

typpo:
Thanks Kodie!

PR #1463, opened 2024-08-18
chore: check config.redteam instead of config.metadata.redteam

mldangelo:
Now that we have the top level redteam config schema, we can check the existence of config.redteam instead of writing on config.metadata.redteam

### DIFF
--- a/src/web/nextui/src/app/eval/ResultsView.tsx
+++ b/src/web/nextui/src/app/eval/ResultsView.tsx
@@ -550,18 +550,19 @@ export default function ResultsView({
                 Table Settings
               </Button>
             </Tooltip>
-            {config?.metadata?.redteam && (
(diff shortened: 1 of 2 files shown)


src/web/nextui/src/app/eval/ResultsView.tsx
```diff
@@ -550,7 +550,8 @@ export default function ResultsView({
                 Table Settings
               </Button>
             </Tooltip>
-            {config?.metadata?.redteam && (
+            {/* TODO: Remove config.metadata.redteam check in favor of config.redteam */}
```
typpo: nit - I always put a name to TODOs (indicates author, not necessarily the one to fix it) and for deprecations I usually put a date 
```suggestion
            {/* TODO(Michael): Remove config.metadata.redteam check (2024-08-18) */}
```

typpo (approved):
nice

PR #1468, opened 2024-08-19
build: add defaults for supabase environment variables

sklein12:
Running `npm run build` would fail without setting supabase environment variables. These aren't required for running prompt foo locally.


typpo (approved):
Nice

PR #1482, opened 2024-08-21
fix: hosted load shared eval

sklein12:
If we're long running locally it means we're running in a hosted server, load the eval from local instead of a potential fallback url of promptfoo.dev


typpo (approved):

mldangelo (approved):
Ship it!

Issue #1510, opened 2024-08-22
Promptfoo cannot decode special characters


balintdom:
**Describe the bug**
When the input message contains the hungarian `Á` or `Ő` characters I get the following error

```
Python code execution failed: Error running Python script: UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 25: character maps to <undefined>
Stack Trace: Error: UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 25: character maps to <undefined>
    at PythonShell.parseError (C:\Users\bdomonkos\AppData\Roaming\nvm\v20.9.0\node_modules\promptfoo\node_modules\python-shell\index.js:303:21)
    at terminateIfNeeded (C:\Users\bdomonkos\AppData\Roaming\nvm\v20.9.0\node_modules\promptfoo\node_modules\python-shell\index.js:193:32)
    at ChildProcess.<anonymous> (C:\Users\bdomonkos\AppData\Roaming\nvm\v20.9.0\node_modules\promptfoo\node_modules\python-shell\index.js:185:13)
    at ChildProcess.emit (node:events:514:28)
    at ChildProcess._handle.onexit (node:internal/child_process:294:12)
    --Python Traceback: --
    File "C:\Users\bdomonkos\AppData\Roaming\nvm\v20.9.0\node_modules\promptfoo\dist\src\python\wrapper.py", line 33, in <module>
      data = json.load(fp)
             ^^^^^^^^^^^^^
    File "C:\Users\bdomonkos\miniconda3\envs\pythonservice\Lib\json\__init__.py", line 293, in load
      return loads(fp.read(),
                   ^^^^^^^^^
    File "C:\Users\bdomonkos\miniconda3\envs\pythonservice\Lib\encodings\cp1250.py", line 23, in decode
      return codecs.charmap_decode(input,self.errors,decoding_table)[0]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

**To Reproduce**
Run promptfoo with user prompts containing the hungarian `Á` or `Ő` characters.

**Expected behavior**
Promptfoo should handle special characters or provide the opportunity to change the encoding.

**System information (please complete the following information):**

- Promptfoo version: 0.80.2

mldangelo:
Hi @balintdom, 

[...]

Issue #1476, opened 2024-08-20
UnicodeEncodeError: 'charmap' codec can't encode character '\ufb02' in position 3672: character maps to <undefined>


beschmitt:
**Describe the bug**
I have a test battery of about 15 questions. 14 of these questions get executed without problem but one always throws this encoder error.
The question is just a simple "Wer ist Stefan?". Other questions like "Wer ist Markus?" works no problem. Also if I change the question to "Wer is Stefan?" it again works without any problems.

**To Reproduce**
It's just a super simple model where the question gets put through a langgraph-model and the output is printed along with some logging.

**Screenshots**
![promptfoo_error](https://github.com/user-attachments/assets/f056d754-072d-46d8-9c5d-b8a9b0812f5d)

**System information (please complete the following information):**

- Promptfoo version: 0.78.3

typpo:
Thanks for the report. Are you able to share your config?

beschmitt:
My promptfoo_config.yaml looks as follows:
```
prompts:
  - prompt.txt
  

providers:
  - "exec:python ./promptfoo_llm_testing.py"
  
define: &latency 8000

tests:
  - description: Requesting personal information
      vars:
          question: Wer ist Markus?
      assert:
        - type: icontains
          value: "'is_question_forbidden': False"
        - type: icontains
          value: 'Answer was generated in: German'
        - type: latency
          threshold: *latency
        - type: icontains-any
          value: [
              Sohn,
              Kind
              ]
  
    - description: Requesting personal information
      vars:
          question: Wer ist Stefan?
      assert:
        - type: icontains
          value: "'is_question_forbidden': False"
        - type: icontains
          value: 'Answer was generated in: German'
        - type: latency
          threshold: *latency
        - type: icontains-any
          value: [
              Sohn,
              Kind
              ]
```

Like I said there are a total of 15 questions but they all have the same structure and all except "Wer ist Stefan?" work.
The prompt.txt is literally just `{{question}}`

beschmitt:
Python string is always utf-8.

mldangelo:
> Python string is always utf-8.

@beschmitt Sorry for our delay on this. We've recently released [v0.80.3](https://github.com/promptfoo/promptfoo/releases/tag/0.80.3), which includes a fix for inter-process communication encoding between Node.js (which writes files as ANSI by default) and Python. Could you please retest with this version and let us know if it resolves the problem you were experiencing?

Thanks for your help in identifying this issue!

[...]

PR #1511, opened 2024-08-22
fix: use utf-8 encoding for python ipc

typpo:
Related to #1510 and #1476 

I think this was introduced by #1447 (cc @enkoder)


enkoder:
classic enkoding memes :P

PR #1516, opened 2024-08-23
fix: redteam providers should be overriddeable

typpo:
Plumbs the `redteam.provider` config through to iterative and crescendo providers, so that they are no longer hardcoded to OpenAI


sklein12 (approved):

package.json
```diff
@@ -168,5 +168,15 @@
     "uuid": "^10.0.0",
     "winston": "^3.14.2",
     "zod": "^3.23.8"
+  },
+  "madge": {
+    "detectiveOptions": {
+      "ts": {
+        "skipAsyncImports": true
```
mldangelo, line 175: nice solution!

mldangelo (approved):

PR #1518, opened 2024-08-24
feat: Indirect prompt injection plugin

sklein12:


src/redteam/util.ts
```diff
@@ -15,7 +15,8 @@ export async function retryWithDeduplication<T>(
   operation: (currentItems: T[]) => Promise<T[]>,
   targetCount: number,
   maxConsecutiveRetries: number = 2,
-  dedupFn: (items: T[]) => T[] = (items) => Array.from(new Set(items)),
+  dedupFn: (items: T[]) => T[] = (items) =>
+    Array.from(new Set(items.map((item) => JSON.stringify(item)))).map((item) => JSON.parse(item)),
```
mldangelo, line 19: out of curiosity, why is this change necessary?

src/redteam/plugins/base.ts
```diff
@@ -66,26 +68,38 @@ export abstract class PluginBase {
...
+      assert: this.getAssertions(prompt.prompt),
     }));
   }
 
+  protected parseGeneratedPrompts(generatedPrompts: string): { prompt: string }[] {
+    return generatedPrompts
+      .split('\n')
+      .filter((line: string) => line.includes('Prompt:'))
+      .map((line: string) => {
+        line = removeAsterisks(line);
```
mldangelo: were you seeing this on a specific plugin or model?

src/commands/generate/redteam.ts
```diff
@@ -107,8 +107,13 @@ export async function doGenerateRedteam(options: RedteamGenerateOptions) {
     typeof s === 'string' ? { id: s } : s,
   );
 
-  logger.debug(`plugins: ${plugins.map((p) => p.id).join(', ')}`);
-  logger.debug(`strategies: ${strategyObjs.map((s) => s.id ?? s).join(', ')}`);
+  try {
+    logger.debug(`plugins: ${plugins.map((p) => p.id).join(', ')}`);
+    logger.debug(`strategies: ${strategyObjs.map((s) => s.id ?? s).join(', ')}`);
+  } catch (error) {
+    logger.error('Error logging plugins and strategies. One did not have a valid id.');
```
mldangelo, line 112: How do we get here? 

nit, consider other ways to handle this such as throwing and exiting or listing a message like `x, y, z plugins are invalid`

mldangelo (commented):
First glance looks great, thank you! You should definitely add this to the guide as well.

src/redteam/util.ts
sklein12: generatePrompts now returns an array of objects instead of an array of strings, so the items will never be the same since they are just pointers to objects now instead of strings, we need to convert the objects into strings to see if they are dupes.

src/redteam/plugins/base.ts
sklein12: Whatever the default model is, I was seeing it on almost every single prompt response

src/commands/generate/redteam.ts
sklein12: You get this if you have a blank line 🤣

PR #1531, opened 2024-08-27
fix: remove failure messages from output

sklein12:
image

Downloading the human eval was including the failure messages in the output which made it unusable to run through the grader again.


typpo (approved):
hacks on hacks.  one day would be cool to plumb through the failure message separately

PR #1532, opened 2024-08-27
fix: reduce pii false positives

sklein12:
PII plugin was full of false positives just reflecting back names from the user input


mldangelo (approved):

typpo (approved):
📈

Issue #1499, opened 2024-08-21
Add perma/direct link to test result in eval view


sklein12:
Add the ability to permalink directly to a result row in the eval view

mldangelo:
Hi @sklein12,

This is a great idea! A direct link to a specific result row in the eval view would definitely be a useful feature. If you're interested, we'd love to welcome a contribution from you to add this functionality. If you decide to take a stab at it, feel free to reach out if you need any guidance or have any questions.

Thanks for the suggestion!

mldangelo:
Hi! Good news - this feature actually already exists! To use it: 

1. **In the eval results table**, hover over any result cell
2. **Hold the Shift key** to reveal the action buttons
3. **Click the Link icon** (🔗) - this copies a permalink to your clipboard

The URL format looks like:
```
https://your-promptfoo-instance/eval/abc123?rowId=5
```

When someone opens this link, the page will automatically:
- Navigate to the correct page (handling pagination)
- Scroll smoothly to the specific result row

## Code references

For anyone curious about the implementation:
- **Link generation**: [`EvalOutputCell.tsx:359-373`](https://github.com/promptfoo/promptfoo/blob/main/src/app/src/pages/eval/components/EvalOutputCell.tsx#L359-L373)
- **URL navigation**: [`ResultsTable.tsx:1158-1234`](https://github.com/promptfoo/promptfoo/blob/main/src/app/src/pages/eval/components/ResultsTable.tsx#L1158-L1234)

I'll note that the feature is somewhat hidden since it requires holding Shift to see the action buttons. If there's interest in making this more discoverable (e.g., always showing a "Copy Link" button), we could open a separate issue for that UX improvement.

Closing this as the core functionality is implemented. Feel free to reopen if the current implementation doesn't meet your needs!

PR #1546, opened 2024-08-28
feat: Added anchor link to specific row and shows on top

irene1391:
Added Anchor link for enhancement request 
relate to #1499


src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -138,6 +140,42 @@ function ResultsTable({
...
+        .catch((err) => {
+          console.error('Could not copy text: ', err);
+          showToast('Failed to copy link', 'error');
+        });
+    },
+    [showToast],
+  );
+
+  // Scroll to the row when the URL contains a hash
+  useEffect(() => {
```
typpo: Two things I'm noticing here:
- We need to account for pagination - currently navigating to a link on page 2 (more than 50 results) does not go to it
- The results page supports other query parameters such as search (which is used in the red team report view).  So it would be good to have more robust parsing, e.g. for a hash like `row-id=foo&search=abc123` (this also applies to generateRowLink)

src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -647,28 +685,38 @@ function ResultsTable({
...
-              <tr key={row.id}>
+              <tr key={row.id} id={`row-${row.id}`}>
                 {row.getVisibleCells().map((cell) => {
                   const isMetadataCol =
                     cell.column.id.startsWith('Variable') || cell.column.id === 'description';
                   const shouldDrawColBorder = !isMetadataCol && !colBorderDrawn;
                   if (shouldDrawColBorder) {
                     colBorderDrawn = true;
                   }
-                  const shouldDrawRowBorder = rowIndex === 0 && !isMetadataCol;
```
typpo: This removal seems to be unintentional?  It causes the table to change visually and draw a thicker border around every row

src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -647,28 +685,38 @@ function ResultsTable({
...
                       key={cell.id}
                       style={{
                         width: cell.column.getSize(),
                       }}
-                      className={`${isMetadataCol ? 'variable' : ''} ${
-                        shouldDrawRowBorder ? 'first-prompt-row' : ''
-                      } ${shouldDrawColBorder ? 'first-prompt-col' : ''}`}
+                      className={`${isMetadataCol ? 'variable' : ''} ${shouldDrawRowBorder ? 'first-prompt-row' : ''} ${shouldDrawColBorder ? 'first-prompt-col' : ''}`}
                     >
+                      {/* Share Button for Metadata Columns */}
```
typpo, line 755: Would it make sense to put this in the group of icons that appear on the row when the user holds down "shift"?  I also think a link icon might be better

typpo (commented):
Thanks @irene1391! This is a great start.  Left a few comments

irene1391:
Thank you so much @typpo ! They are really really good feedback. I'm on my way to Boston today. Will address them tmr as soon as I can :)

PR #1551, opened 2024-08-29
chore: write eval tags to database and add migration

typpo:
- Add `Record<string, string>` tags to match with UnifiedConfig/TestSuite
- Also add indexes in migration since we forgot to migrate those


sklein12 (approved):

drizzle/0004_rare_lord_tyger.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE `evals` ADD `tags` text;--> statement-breakpoint
+CREATE INDEX `datasets_created_at_idx` ON `datasets` (`created_at`);--> statement-breakpoint
+CREATE INDEX `evals_created_at_idx` ON `evals` (`created_at`);--> statement-breakpoint
+CREATE INDEX `evals_author_idx` ON `evals` (`author`);--> statement-breakpoint
+CREATE INDEX `evals_tags_idx` ON `evals` (`tags`);--> statement-breakpoint
```
sklein12: Dunno how this is going to perform. I think you need to do a json_extract here.

sklein12: `json_extract('tags', '$') `

sklein12: I'll need to do some testing though to see if the index is used all the time or ever...

sklein12: NM, that doesn't work, you have to extract keys to make the index work properly..

image



image

sklein12 (approved):
Lgtm

Issue #1552, opened 2024-08-29
Is there a way to import a environment variable that is not predefined in promptfoo ?


AmineDjeghri:
**Is your feature request related to a problem? Please describe.**
Currently, it seems that only certain predefined environment variables are supported, and there isn't a straightforward way to import custom environment variables that are specific to my workflow or setup.

**Describe the solution you'd like**
I would like the ability to import any environment variable, even if it's not predefined in `promptfoo`.

typpo:
By "import an environment variable", do you mean referencing an environment
variable in a prompt template?

AmineDjeghri:
Yes!

mpasternak:
@AmineDjeghri I'd either use something to generate YAML configuration file... or check out this answer: https://stackoverflow.com/questions/52287881/how-do-access-node-env-in-nunjucks-template

mldangelo:
Thanks to @AmineDjeghri and @mpasternak for raising and discussing this issue. @typpo has shipped a fix, and you can check it out by running `npx promptfoo@latest`. Here's a quick example:

You can access environment variables in your templates using the `env` global:

```yaml
tests:
  - vars:
      headline: 'Latest news in {{ env.COUNTRY }}'
```

Please let us know if you encounter any issues or have more ideas for improving the product.

AmineDjeghri:
Hello again,
i am trying using it in the providers but it seems that it doesn't work. Promptfoo version is 0.90.3

```
providers:
  - id: azureopenai:chat:gpt-4o-mini # env variables are in .env
  - id: file://../configs/config_json.py
    label: '{{ env.AZURE_OPENAI_DEPLOYMENT_NAME }}'
```

zpedrol:
I'm trying to use an env var to pass an api key in the header but it's not working, is the env vars only available in specific places in the configuration? Can it be extended to headers we well? Thanks!

`headers:
                'apikey': {{ env.MY_API_KEY }}`

gesslinger:
@zpedrol @mldangelo  seems to be a bug: https://github.com/promptfoo/promptfoo/issues/3331

PR #1553, opened 2024-08-29
feat: global `env` var in templates

typpo:
Related to #1552

### DIFF
--- a/site/docs/configuration/guide.md
+++ b/site/docs/configuration/guide.md
@@ -463,6 +463,16 @@ tests:
       topic: 'theoretical quantum physics in alternate dimensions'
 ```
 
+### Accessing environment variables
+
+You can access environment variables in your templates using the `env` global:
+
+```yaml
+tests:
+  - vars:
+      headline: 'Articles about {{ env.TOPIC }}'
+```
+
 ## Tools and Functions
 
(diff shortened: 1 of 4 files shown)


mldangelo (approved):
good solution. Please consider adding a unit test

AmineDjeghri:
Thank you !

AmineDjeghri:
Hello
i just tried it and it doesn't work for me

i have promptfoo latest version and here is a small example how i did it : 

image

AmineDjeghri:
i would like to know also, if it works in the providers.
Because the purpose is to use is in the providers config like this : 

id: azureopenai:chat:gpt-4o-mini
config:
  apiHost: '{{env.AZURE_OPENAI_API_ENDPOINT}}'

typpo:
@AmineDjeghri  No, it won't work for providers.  The provider expects a specific environment variable for that: `AZURE_OPENAI_API_HOST`

AmineDjeghri:
are you sure AZURE_OPENAI_API_HOST exists and works?  I've tried it and it doesn't work for me. 
It's also not mentioned here 
the reason i created this issue is the absence of some env variables, and instead of waiting for the maintainers to add them, we can add our own env variables

PR #1555, opened 2024-08-30
feat(redteam): add remote generation for purpose and entities

mldangelo:
- Implement remote generation service for extracting system purpose and entities
- Add environment variable PROMPTFOO_DISABLE_REDTEAM_REMOTE_GENERATION to control remote generation
- Update documentation to explain remote generation feature and its benefits
- Refactor extraction logic to use remote service with local fallback

NOTE: I will update the URL to something more sensible but this can be tested now.


src/redteam/extraction/entities.ts
```diff
@@ -12,7 +44,7 @@ export async function extractEntities(provider: ApiProvider, prompts: string[]):
     Each line in your response must begin with the string "Entity:".
   `;
 
-  return callExtraction(provider, prompt, (output: string) => {
+  return await callExtraction(provider, prompt, (output: string) => {
```
typpo: I think there is debate about this but I usually just `return` in these situations
```suggestion
  return callExtraction(provider, prompt, (output: string) => {
```

src/redteam/extraction/purpose.ts
```diff
@@ -1,11 +1,44 @@
...
   provider: ApiProvider,
   prompts: string[],
 ): Promise<string> {
+  const useRemoteGeneration = !getEnvBool('PROMPTFOO_DISABLE_REDTEAM_REMOTE_GENERATION', false);
+
+  if (useRemoteGeneration) {
+    try {
+      const body = {
+        task: 'purpose',
+        prompts: prompts,
```
typpo:
```suggestion
        prompts,
```

src/redteam/extraction/purpose.ts
```diff
@@ -1,11 +1,44 @@
...
+      };
+
+      const response = await fetchWithCache(
```
typpo: so much nicer than the provider approach!

shall we put `fetchRemoteGeneration` in `common`?  Otherwise you're importing the pieces (`RedTeamGenerationResponse` and `REMOTE_GENERATION_URL`) and constructing the request multiple times.

src/redteam/extraction/purpose.ts
```diff
@@ -19,7 +52,7 @@ export async function extractSystemPurpose(
     <Purpose>Ecommerce chatbot that sells shoes</Purpose>
   `;
 
-  return callExtraction(provider, prompt, (output: string) => {
+  return await callExtraction(provider, prompt, (output: string) => {
```
typpo:
```suggestion
  return callExtraction(provider, prompt, (output: string) => {
```

typpo (approved):
Nice change

src/redteam/extraction/entities.ts
mldangelo: This was intentional. I did it here because of the try-catch (which also may be unnecessary). If you dislike the pattern I can write it to a variable first.

mldangelo: It was not intentional! I apologize, thanks for the callout.

typpo (approved):

PR #1564, opened 2024-09-01
feat: add `delay` option for redteam generate and refactor plugins

typpo:
- Added `--delay` cli and `delay` config options.  This is useful to avoid hitting rate limits :( 
- Realized that makes the `Plugins` definition a huge mess, lots of repetition, so refactored to factories pattern
- Made "modifiers" behavior a config key - `config.modifiers` so we are not passing around a separate `modifiers` object.  Included a special case for `config.language`


src/validators/redteam.ts
```diff
@@ -90,6 +90,7 @@ export const RedteamGenerateOptionsSchema = z.object({
   purpose: z.string().optional().describe('Purpose of the redteam generation'),
   strategies: z.array(RedteamStrategySchema).optional().describe('Strategies to use'),
   write: z.boolean().describe('Whether to write the output'),
+  delay: z.number().int().positive().optional().describe('Delay in milliseconds generators'),
```
mldangelo:
```suggestion
  delay: z.number().int().positive().optional().describe('Delay in milliseconds for generators'),
```

src/validators/redteam.ts
```diff
@@ -134,8 +135,14 @@ export const RedteamConfigSchema = z
       .positive()
       .optional()
       .describe('Maximum number of concurrent API calls'),
+    delay: z
+      .number()
+      .int()
+      .positive()
+      .optional()
+      .describe('Delay in milliseconds between plugin API calls'),
```
mldangelo, line 148: The description is different even though I believe these are the same

src/redteam/plugins/index.ts
```diff
@@ -22,146 +23,88 @@ import { ShellInjectionPlugin } from './shellInjection';
...
+): PluginFactory {
+  return {
+    key,
+    validate: validate as ((config: PluginConfig) => void) | undefined,
+    action: (provider, purpose, injectVar, n, delayMs, config) =>
+      new PluginClass(provider, purpose, injectVar, config as T).generateTests(n, delayMs),
+  };
+}
+
+const pluginFactories: PluginFactory[] = [
```
mldangelo, line 59: thank you so much for doing this

mldangelo (approved):
Thank you!

src/validators/redteam.ts
typpo: why are there 3 places where this string has to be copied?  are the zod descriptions ever shown to the user?

PR #1577, opened 2024-09-03
chore: warn if API keys are not present when running `promptfoo init`

cristiancavalli:
Hey All!

 A couple changes to make the non red-team onboarding process a bit more info-packed:

- Require at least one selection from the providers list and slightly clear up the command info language.
- Add a function that checks if a selected provider that requires an API key does not have that api key registered in the env. If the corresponding env override is not present alert the user and give them the command to add it to the env.
- Add some tests for the aforementioned function to make sure it logs correctly for both object and string-type choices and don't regress as time goes.

Many thanks!!!


src/onboarding.ts
```diff
@@ -228,14 +229,79 @@ promptfoo eval
 Afterwards, you can view the results by running \`promptfoo view\`
 `;
 
-function recordOnboardingStep(step: string, properties: Record<string, EventValue> = {}) {
+function recordOnboardingStep(step: string, properties: any = {}) {
```
mldangelo:
```suggestion
function recordOnboardingStep(step: string, properties: any = Record<string, EventValue>) {
```

src/onboarding.ts
```diff
@@ -228,14 +229,79 @@ promptfoo eval
...
     step,
     ...properties,
   });
 }
 
+/**
+ * Iterate through user choices and determine if the user has selected a provider that needs an API key
+ * but has not set and API key in their enviorment.
+ */
+export function reportProviderAPIKeyWarnings(providerChoices: (string | object)[]): string[] {
```
typpo, line 247: Great idea to display a warning for this!

Is there a way we can simplify this function? e.g. using the map more directly and/or reducing the switches/branching

cristiancavalli: Updated to address comments and reduce complexity -- fresh morning eyes always help lol 👀

src/envars.ts
```diff
@@ -98,6 +98,14 @@ export type EnvVars = {
...
 type EnvVarKey = keyof EnvVars;
 
+/**
+ * Provider env keys that are required to be set either in env or in config to interact with provider services.
+ */
+export enum ProvidersRequiringAPIKeysEnvKeys {
+  OPENAI_API_KEY = 'OPENAI_API_KEY',
+  ANTHROPIC_API_KEY = 'ANTHROPIC_API_KEY',
+}
+
```
sklein12: I don't think this belongs in this file. Could you move it to be co-located with the providers?

src/onboarding.ts
```diff
@@ -228,14 +229,66 @@ promptfoo eval
...
   telemetry.recordAndSend('funnel', {
     type: 'eval onboarding',
     step,
     ...properties,
   });
 }
 
+/**
+ * Iterate through user choices and determine if the user has selected a provider that needs an API key
+ * but has not set and API key in their enviorment.
```
sklein12: "has not set _an_ API key"

src/onboarding.ts
```diff
@@ -228,14 +229,66 @@ promptfoo eval
...
+ */
+export function reportProviderAPIKeyWarnings(providerChoices: (string | object)[]): string[] {
+  // Dictionary of what warnings may have already been printed for the user.
+  const shownWarnings = new Map([
+    [
+      ProvidersRequiringAPIKeysEnvKeys.OPENAI_API_KEY as keyof typeof ProvidersRequiringAPIKeysEnvKeys,
```
sklein12: I'm a TS noob, why does this need to be typed using the `as keyof typeof` statement?

src/onboarding.ts
```diff
@@ -228,14 +229,66 @@ promptfoo eval
...
   });
 }
 
+/**
+ * Iterate through user choices and determine if the user has selected a provider that needs an API key
+ * but has not set and API key in their enviorment.
+ */
+export function reportProviderAPIKeyWarnings(providerChoices: (string | object)[]): string[] {
+  // Dictionary of what warnings may have already been printed for the user.
+  const shownWarnings = new Map([
```
sklein12: can you generate this map dynamically instead of hardcoding it? So if we add a new provider to `ProvidersRequiringAPIKeysEnvKeys` we don't have to update this code?

src/onboarding.ts
cristiancavalli: This is a weird artifact of typescript enums and attempting to use them as strings -- essentially this enables us "to get a Type that represents all Enum keys as strings."1

I like the ability to use enums although the syntax is a bit clunky as they don't have the ability to modified like a regular map or dictionary. Hopefully this syntax might get better in the future.

1 https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time

src/onboarding.ts
cristiancavalli: Ahh wait sorry misread your comment -- can update!

src/envars.ts
cristiancavalli: I put this here given that it does relate to the env vars that are checked. Is there a specific place in providers you think this would better live and if so why there and not near the envvars keys they are meant to mirror?

sklein12: In my mind these are a property of the providers so they should live with/on the providers. Secondly, this has a side effect of defining which providers require an environment variable, which to me feels like a property of the provider as well. Lastly, the envars file is to manage environment variables, it shouldn't have any application logic in it.

src/onboarding.ts
cristiancavalli: Updated and I also fixed my misspell of _environment_

src/onboarding.ts
cristiancavalli: Updated!

src/envars.ts
cristiancavalli: Understood -- moved!

mldangelo (commented):
One nit and we can ship this. Thank you so much @cristiancavalli

cristiancavalli:
Thank @mldangelo for the catch -- finally figured out how to disable the strange repl.it autoformat behaviour that cause this!

cristiancavalli:
Fixed a quick type/value swap in the syntax!

mldangelo:
Thank you!

cristiancavalli:
Thank you! Sorry for the bug -- I'm interested were you able to reproduce in tests? I was running locally and didn't see it so I must've missed something

PR #1580, opened 2024-09-04
fix(openai): don't set openai chat completion seed=0 by default

Sasja:
It seems openai now supports seeds for reproducible responses. The code before this commit sets the openai chatcompletion seed to 0 by default resulting in repeated tests always having the same outcome. I'm pretty new to promptfoo but I'm guessing that is not intentional when using the `--repeat` cli option for instance. It might be due to the fact that seeds are in beta at openai and their behaviour has changed. see https://platform.openai.com/docs/api-reference/chat/create#chat-create-seed

The patch suggested makes not providing a seed value work as expected (non-deterministic outcome). But I did not dive very deep into the code to check if it might break anything. I'm opening this pull request just to report.

### DIFF
--- a/src/providers/openai.ts
+++ b/src/providers/openai.ts
@@ -371,7 +371,7 @@ export class OpenAiCompletionProvider extends OpenAiGenericProvider {
     const body = {
       model: this.modelName,
       prompt,
-      seed: this.config.seed || 0,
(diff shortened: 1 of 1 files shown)


mldangelo:
Thank you @Sasja!

PR #1582, opened 2024-09-04
feat: Added anchor link to specific row and shows on top 

irene1391:
- Addressed PR comments
- Rebuild on newer version Promptfoo

relate to #1499


src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -138,6 +140,43 @@ function ResultsTable({
...
+
+    Object.entries(additionalParams).forEach(([key, value]) => {
+      params.set(key, (value as string).toString());
+    });
+
+    params.set('row-id', rowId);
+
+    currentUrl.search = params.toString();
+
+    currentUrl.hash = `row-${rowId}`;
```
typpo: Is the hash ever used?  Looks like we are using query params elsewhere.

src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -665,10 +756,20 @@ function ResultsTable({
                       style={{
                         width: cell.column.getSize(),
                       }}
-                      className={`${isMetadataCol ? 'variable' : ''} ${
-                        shouldDrawRowBorder ? 'first-prompt-row' : ''
-                      } ${shouldDrawColBorder ? 'first-prompt-col' : ''}`}
+                      className={`${isMetadataCol ? 'variable' : ''} ${shouldDrawRowBorder ? 'first-prompt-row' : ''} ${shouldDrawColBorder ? 'first-prompt-col' : ''}`}
                     >
+                      {/* Share Button for Metadata Columns */}
+                      {isMetadataCol && isShiftPressed && (
```
typpo: Let's include this in the existing shift-menu that appears on rows (see right-side in the gif).  This would solve two problems:
- The link appears on every row, not just the current row
- The link causes all content in its cell to shift
linktoggle

src/web/nextui/src/app/eval/ResultsTable.tsx
```diff
@@ -602,6 +641,57 @@ function ResultsTable({
...
+    };
+  }, []);
+  useEffect(() => {
+    const params = parseQueryParams(window.location.search);
+    const rowId = params['row-id'];
+
+    if (rowId) {
+      const rowElement = document.querySelector(`#row-${rowId}`);
+
+      if (rowElement) {
```
typpo: This isn't working on multipage evals right now.  I think this check is probably the reason why.  It only tries to repaginate/scroll if the element already exists on the page.

typpo (commented):
Left a couple comments!

src/web/nextui/src/app/eval/ResultsTable.tsx
irene1391: I see. To clarify, are you asking if the multi-page content within the cell should render, or if the entire page should render? @typpo 
image

typpo (commented):
Looking much better!  I noticed a little more strangeness: it is no longer possible to change the page after loading it, instead it will automatically switch back to the linked row.

Also minor comment but let's do `rowId` instead of `row-id` to match the convention of other query params.

mldangelo:
Sorry for the long delay in merging this! Thank you @irene1391!

PR #1587, opened 2024-09-04
fix: improve strategy json parsing

typpo:
genius idea from @sklein12

### DIFF
--- a/src/redteam/providers/iterativeImage.ts
+++ b/src/redteam/providers/iterativeImage.ts
@@ -10,6 +10,7 @@ import type {
   NunjucksFilterMap,
   RedteamFileConfig,
 } from '../../types';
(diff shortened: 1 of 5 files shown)


sklein12 (approved):
absolute genius

src/redteam/providers/iterativeImage.ts
```diff
@@ -154,7 +155,7 @@ async function runRedteamConversation({
     ]);
     const isOnTopicResp = await redteamProvider.callApi(isOnTopicBody);
     invariant(typeof isOnTopicResp.output === 'string', 'Expected output to be a string');
-    const isOnTopic = JSON.parse(isOnTopicResp.output).isOnTopic;
```
mldangelo, line 157: BOOO

PR #1605, opened 2024-09-06
fix: correct metric name for misinfo/pii/etc plugins

typpo:


mldangelo (approved):
Thank you Mr @typpo

src/redteam/plugins/harmful.ts
```diff
@@ -191,7 +191,7 @@ class HarmfulPlugin extends PluginBase {
     return [
       {
         type: `promptfoo:redteam:${this.category.key}`,
-        metric: `${this.category.key}`,
+        metric: 'Harmful',
```
mldangelo, line 194: Can we add test coverage for this so it doesn't happen again?

typpo: yes but I have a question about this test.  can we address this privately

PR #1615, opened 2024-09-07
feat: add warning for prompts without variables

mldangelo:
- Implement variable detection in readPrompts function
- Add warning for prompts without variables
- Add PROMPTFOO_DISABLE_NO_VARIABLE_WARNING to EnvVars type
- Add tests for new warning functionality


mldangelo:
Closing for now. Have to rethink when someone wants to compare prompts without any vars in their tests.

PR #1627, opened 2024-09-10
Chore: Standardize Paths

sklein12:
Standardizing paths between the main package and the web so 
`@app` and `@promptfoo` point to the same place no matter which package you're in

image


sklein12:
you're welcome all future engineers

typpo (approved):
amazing change

PR #1639, opened 2024-09-12
chore: add --remote to `eval`

typpo:


src/commands/eval.ts
```diff
@@ -412,6 +413,7 @@ export function evalCommand(
       'Run providers interactively, one at a time',
       defaultConfig?.evaluateOptions?.interactiveProviders,
     )
+    .option('--remote', 'Force remote inference wherever possible (used for red teams)', false)
```
mldangelo, line 416: can we put this behind an experimental flag?

mldangelo (approved):

src/commands/eval.ts
typpo: `generate` already has a `--remote` flag so I'd rather it match (also it's not experimental because it will definitely work 100% no bugs)

PR #1650, opened 2024-09-12
fix(providers): add support for openai o1 max_completion_tokens

mldangelo:
- Add OPENAI_MAX_COMPLETION_TOKENS to environment variables
- Implement special handling for o1 models in OpenAiChatCompletionProvider
- Update OpenAI documentation with new environment variables and model support


typpo (approved):
this code straight bussin fr 🔥 u dropped this 👑 based

PR #1659, opened 2024-09-13
fix: normalize prompts when reading configs

mldangelo:
This normalizes our prompts earlier when reading the config. It does not narrow the type yet. This fixes issue with prompt-labels example not working correctly and an issue reported via email.


test/config.test.ts
```diff
@@ -55,6 +56,12 @@ jest.mock('../src/testCases', () => {
   };
 });
 
+// At the top of the file, update the mock
```
typpo:
```suggestion
```

typpo (approved):

Issue #1667, opened 2024-09-14
Latency assertion doesn't work for custom provider


albertlieyingadrian:
I tried adding a custom provider and tried asserting the latency but I kept getting an error even though I already run the eval with `--no-cache`

Here is the error
```shell
Error: Latency assertion does not support cached results. Rerun the eval with --no-cache

Error: Latency assertion does not support cached results. Rerun the eval with --no-cache
    at runAssertion 
```

Here is the custom provider
```javascript
const { distance, closest } = require('fastest-levenshtein')

class CustomApiProvider {
  constructor(options) {
    // Provider ID can be overridden by the config file (e.g. when using multiple of the same provider)
    this.providerId = options.id || 'custom provider';

    // options.config contains any custom options passed to the provider
    this.config = options.config;
  }

  id() {
    return this.providerId;
  }

  async callApi(prompt, context) {
    const yesThreshold = this.config.threshold_yes || 4;

    const currentAddress = context.vars.current_address;
    const targetAddress = context.vars.target_address;

    const levDistance = distance(currentAddress, targetAddress);

    const result = levDistance < yesThreshold ? "Yes" : "No";

    return {
      output: JSON.stringify({
        result: result,
        reason: `The distance is ${levDistance}, so the result is ${result}.`
      }),
      tokenUsage: {
        total: 0,
        prompt: 0,
        completion: 0,
        cached: 0
      },
      cached: false,
      cost: 0
    };
  }
}

module.exports = CustomApiProvider;
```

albertlieyingadrian:
My work around is to add delay in my processor
```
    await new Promise(resolve => setTimeout(resolve, 100)); // 100ms delay
    const levDistance = distance(currentAddress, targetAddress);
```

So the latency will never be 0

mldangelo:
Hi @albertlieyingadrian, thank you so much for the fix! I just merged your PR and really appreciate your quick work on it. I’ll ensure it gets released for you tomorrow. Please keep the bug reports and PRs coming! Thanks again for contributing and being such a great part of the community!

albertlieyingadrian:
Thanks! Saw that https://github.com/promptfoo/promptfoo/pull/1668

Should I close this issue now? @mldangelo

mldangelo:
We usually wait to close them until the release is published, but by all means, you can close it whenever you like. Thanks again for the fix!

mldangelo:
[...]

PR #1668, opened 2024-09-14
fix(assertions): fix latencyMs comparison with undefined to allow 0 ms latency

albertlieyingadrian:
Context: https://github.com/promptfoo/promptfoo/issues/1667

and https://github.com/promptfoo/promptfoo/blob/main/src/evaluator.ts#L277

0 ms latency happens when the provider run is blazingly fast


mldangelo:
Nice find! Thank you @albertlieyingadrian for resolving this! I am going to make sure it's covered with a test and then get it merged and deployed.

PR #1670, opened 2024-09-14
feat(python): add validatePythonPath function and improve error handling

mldangelo:
Implement `validatePythonPath` function to check the Python executable path

If the primary Python executable is not found, it checks for alternative options such as `python3` and the Windows equivalent. If no valid Python installation is detected, it throws a helpful error message instead of exiting the process with an incomprehensible error. This function is executed only once per session.

This improvement was inspired by observing two users in the last 24 hours who attempted to use promptfoo with Python for the first time on new Mac systems that did not have Python in the system PATH.

These changes aim to streamline the setup process and provide better guidance for users, especially those new to using promptfoo with Python.

### DIFF
--- a/src/python/pythonUtils.ts
+++ b/src/python/pythonUtils.ts
@@ -52,7 +52,6 @@ export async function validatePythonPath(pythonPath: string, isExplicit: boolean
   }
   const secondaryPath = await tryPath(alternativePath);
   if (secondaryPath) {
-    logger.warn(`Secondary path found: ${secondaryPath}`);
     state.cachedPythonPath = secondaryPath;
     return secondaryPath;
   }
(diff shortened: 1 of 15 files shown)


src/python/pythonUtils.ts
```diff
@@ -1,12 +1,79 @@
...
+  if (isExplicit) {
+    throw new Error(
+      `Python not found. Tried "${pythonPath}" ` +
+        `Please ensure Python is installed and set the PROMPTFOO_PYTHON environment variable ` +
+        `to your Python executable path (e.g., '${isWindows ? 'C:\\Python39\\python.exe' : '/usr/bin/python3'}').`,
+    );
+  }
+  const secondaryPath = await tryPath(alternativePath);
+  if (secondaryPath) {
+    logger.warn(`Secondary path found: ${secondaryPath}`);
```
typpo: Switch to debug or remove?

typpo (approved):
Very nice change

Issue #1673, opened 2024-09-14
Error when generating dataset


albertlieyingadrian:
I tried generating dataset but kept getting an error
```shell
 ███░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ 6% | ETA: 16s | 1/16/Users/albertlieadrian/Documents/Random/a4
    throw new Error(value);
          ^

Error: Invariant failed: Expected exactly one JSON object in the response for persona Tourist looking fs
    at invariant (/Users/albertlieadrian/Documents/Random/ai-llm-playground/evals/node_modules/tiny-inv)
    at generateTestCasesForPersona (/Users/albertlieadrian/Documents/Random/ai-llm-playground/evals/nod)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async retryWithDeduplication (/Users/albertlieadrian/Documents/Random/ai-llm-playground/evals/no)
    at async synthesize (/Users/albertlieadrian/Documents/Random/ai-llm-playground/evals/node_modules/p)
    at async doGenerateDataset (/Users/albertlieadrian/Documents/Random/ai-llm-playground/evals/node_mo)
```

Here is the command
```shell
npx promptfoo generate dataset --config promptfoo-data.yaml  -o addresses.yaml
```

And here is the config file (following the example here https://www.promptfoo.dev/docs/configuration/datasets/#prepare-your-prompts)
```yaml
prompts:
  - 'Act as a travel guide for {{location}}'
  - 'I want you to act as a travel guide. I will write you my location and you will suggest a place to visit near my location. In some cases, I will also give you the type of places I will visit. You will also suggest me places of similar type that are close to my first location. My current location is {{location}}'

tests:
  - vars:
      location: 'San Francisco'
  - vars:
      location: 'Wyoming'
  - vars:
      location: 'Kyoto'
  - vars:
      location: 'Great Barrier Reef'
```

mldangelo:
Hi @albertlieyingadrian, thanks so much for engaging with Promptfoo and trying out the dataset generation feature! The error you're encountering might be related to your provider configuration. Could you please share the providers section of your promptfooconfig.yaml? Dataset generation may not work well with smaller models, and you can refer to this section for more details: [testCases.ts#L329](https://github.com/promptfoo/promptfoo/blob/main/src/testCases.ts#L329).

[...]

PR #1675, opened 2024-09-14
fix(util/json): fix jsonExtraction when generating dataset

albertlieyingadrian:
Context https://github.com/promptfoo/promptfoo/issues/1673

### DIFF
--- a/src/util/json.ts
+++ b/src/util/json.ts
@@ -29,41 +29,42 @@ export function safeJsonStringify(value: any, prettyPrint: boolean = false): str
 }
 
 export function extractJsonObjects(str: string): object[] {
-  // This will extract all json objects from a string
-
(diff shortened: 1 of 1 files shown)


typpo:
Thanks @albertlieyingadrian!  This is a great catch, should we add your input as a test case as well?

albertlieyingadrian:
It failed other test cases :\ @mldangelo

mldangelo:
Thanks for the fix, @albertlieyingadrian! Could you take a look at the test failures in `test/util.json.test.ts`? I think the tests are correct as is, and they might help identify any edge cases that still need addressing. Let me know what you find or if you'd like help!

albertlieyingadrian:
Moved this PR here https://github.com/promptfoo/promptfoo/pull/1687

PR #1676, opened 2024-09-15
feat: Integrate Adaline Gateway for OpenAI complete-chat, embedding calls

adaline-akshay:
### Notes:
* Update `CallApiContextParams` to include optional gateway object
  * Uses a single gateway object created for each eval run
  * Can handle multiple requests concurrently when running with concurrency > 1
  * Supports PromptFoo like retry, timeout, header based delay for 429 error, exponential backoff for 5XX errors, proxy-agent integration, and caching out of the box
  * Will be used to integrate with other providers in future
 * Update `openai` provider
   * use gateway based chat-completion, embedding by default
   * For unsupported calls / any exception from gateway -- fallback to native OpenAI call (existing implementation)
   * embedding calls don't reuse the gateway object because they are sequential
  * Created a gateway cache plugin
    * Light wrapper over PromptFoo's existing cache implementation
    * Allows access for gateway to read / write PromptFoo's cache implementation
### Testing
* Tested with `examples/openai-tools-call/promptfooconfig.yaml`
* Tested with `examples/openai-chat-history/promptfooconfig.yaml`
* Tested with self created configs handling text, image, tool modality messages
* Tested cache implementation by firing same prompts multiple times
* Tested gateway retry, timeout, backoff, proxy-agent by tweaking gateway options
* Passing all existing openai unit tests

### DIFF
--- a/site/blog/excessive-agency-in-llms.md
+++ b/site/blog/excessive-agency-in-llms.md
@@ -27,11 +27,13 @@ Excessive agency in LLMs often stems from well-intentioned but poorly implemente
 
 ### Common Examples
 
+Excessive agency arises in a handful of ways, but there are a few common patterns.
+
 ![excessive agency examples](/img/blog/excessive-agency/excessive-agency-examples.svg)
 
 - **Overreaching tools**: Most LLM providers have first-class support for ["tools" or functions](https://platform.openai.com/docs/guides/function-calling), which is just a fancy way of saying "APIs that are exposed to the LLM".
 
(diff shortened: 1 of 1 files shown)


mldangelo:
Hey @adaline-akshay, thank you so much for opening this PR and for your thoughtful work on integrating the Adaline Gateway. Could you also add some documentation around this integration? You can check out our integration docs here: https://github.com/promptfoo/promptfoo/tree/main/site/docs/integrations. We would also welcome a guide too. Thanks again!

PR #1684, opened 2024-09-16
chore: plumb scores per plugin and strategy

typpo:
will be used in dashboard


src/evaluator.ts
```diff
@@ -657,6 +657,21 @@ class Evaluator {
         metrics.namedScoresCount[key] = (metrics.namedScoresCount[key] || 0) + 1;
       }
 
+      for (const gradingResult of row.gradingResult?.componentResults || []) {
```
mldangelo: nit, can we wrap this in a `if redteam?`

mldangelo (approved):

PR #1694, opened 2024-09-17
feat: make config `prompts` optional

typpo:
Did it better than #1686 with the same pattern from #1689 

warns the user if there isn't a `prompt` test var


mldangelo (approved):
Much better, thank you

PR #1703, opened 2024-09-18
docs: gpt vs o1 guide

typpo:


mldangelo (approved):
This is great, thank you @typpo. At some point we should start using actual benchmarks for these

PR #1704, opened 2024-09-18
feat: support input transforms 

MrFlounder:
Added input transform functionality to the transform function in src/util/transform.ts. This enhancement allows for transforming input variables before they are used in prompts, providing more flexibility in prompt generation and testing.

- Added a new `TransformInputType` enum to distinguish between output and variable transformations.
- Modified the `transform` function to accept an inputType parameter.
- Updated the `getInlineTransformFunction` to use the inputType parameter.
- Adjusted the `getTransformFunction` to pass the inputType to getInlineTransformFunction.
- Added logic to apply transform on vars when `transform_vars` is present in defaultTests

Tests:
Run the existing test suite to ensure no regressions:
```
npm test
```

Manual testing
- Create a configuration file that uses the new transform_vars option in the defaultTest.options options.
- Run an evaluation using this configuration and verify that the variables are transformed correctly and showing up in the variables in eval result page.

Example usage - tranform an image link to markdown to render in UI

tranform.js
```
module.exports = {
  transform: async (vars) => {
    const markdownImage = `![image](${vars.image})`;
    return { image_markdown: markdownImage };
  },
};
```

promptfooconfig.yaml
```
description: 'My eval'

prompts:
  - 'Write a sentence about the {{image}}'

providers:
  - 'openai:gpt-4o'


defaultTest:
  options:
    transform_vars: file://transform.js:transform

tests:
  - vars:
      image: 'https://upload.wikimedia.org/wikipedia/commons/thumb/7/79/Green_Moray_Eel_%28Gymnothorax_funebris%29%2C_Cozumel.jpg/320px-Green_Moray_Eel_%28Gymnothorax_funebris%29%2C_Cozumel.jpg'
    assert:
      - type: llm-rubric
        value: ensure the output contains a image link
```

UI:
image


site/docs/configuration/guide.md
```diff
@@ -631,6 +631,62 @@ If no function name is specified for Python files, it defaults to `get_transform
...
+```
+
+In this example, we're creating two new variables: uppercase_topic (an uppercase version of the original topic) and topic_length (the length of the topic string). These new variables can be used in your prompts or assertions.
+
+### Transforms from separate files
+
+You can also use external files for `transform_vars`, similar to output transforms:
+
+```yaml
+efaultTest:
```
typpo:
```suggestion
defaultTest:
```

site/docs/configuration/guide.md
```diff
@@ -631,6 +631,62 @@ If no function name is specified for Python files, it defaults to `get_transform
 transform: file://transform.py:custom_python_transform
 ```
 
+## Transforming input variables
+
+In addition to transforming outputs, you can also transform input variables before they are used in prompts. This is done using the `transform_vars` option in the defaultTest.
+The transform_vars function should return an object with the transformed variable names and values. Here's an example:
```
typpo:
```suggestion

The transform_vars function should return an object with the transformed variable names and values. Here's an example:
```

src/types/index.ts
```diff
@@ -102,6 +102,7 @@ export const OutputConfigSchema = z.object({
    */
   postprocess: z.string().optional(),
   transform: z.string().optional(),
+  transform_vars: z.string().optional(),
```
typpo: Let's make it `transformVars` to match the casing of everything else :)

src/util/transform.ts
```diff
@@ -84,20 +89,23 @@ async function getFileTransformFunction(filePath: string): Promise<Function> {
...
  * @param codeOrFilepath - Either inline code or a file path starting with 'file://'.
  * @returns A Promise resolving to the appropriate transform function.
  */
-async function getTransformFunction(codeOrFilepath: string): Promise<Function> {
+async function getTransformFunction(
+  codeOrFilepath: string,
+  inputType: TransformInputType,
+): Promise<Function> {
   if (codeOrFilepath.startsWith('file://')) {
     return getFileTransformFunction(codeOrFilepath);
```
typpo, line 106: If possible we should update this code path as well - most people use transforms from external files

typpo (commented):
This is a very cool change

src/util/transform.ts
MrFlounder: I thought this code works with external files already? Because for external files we `getTransformFunction` doesn't care about function signature, as long as the path exist and correct.

See the example usage part of the pr summary.

Please correct me if I'm my understanding is wrong :)

src/types/index.ts
MrFlounder: Gotcha! Updated.

mldangelo (commented):
This is great! Thank you so much, @MrFlounder.

Could you please confirm that this works similarly to how output transforms function, meaning they can be applied to both `defaultTest` and individual test cases? It would be fantastic to include an example as well (you could modify one of the existing ones - a markdown image preview would be great in one of the multi-modal examples). 

I’ve made a few updates to the documentation too. Please make sure it's ok (or suggest changes or revert it). 

Thanks for making promptfoo better!

MrFlounder:
This PR doesn't add transformVars to individual test cases - it should be added for consistency - but is it better to be in the same PR or in a follow up PR? @mldangelo Happy to make the change if you think it's better to be added all together.

mldangelo:
@MrFlounder let's do it in this PR if you don't mind.

MrFlounder:
Added same functionality to inidividual test cases and added to examples

PR #1719, opened 2024-09-19
feat: Cloud Login

sklein12:
Adding commands to connect the CLI to the cloud.


src/globalConfig/globalConfig.ts
```diff
@@ -5,32 +5,23 @@
...
 import yaml from 'js-yaml';
 import * as path from 'path';
-import type { GlobalConfig } from './configTypes';
-import { getConfigDirectoryPath } from './util/config';
-
-let globalConfigCache: GlobalConfig | null = null;
-
-export function resetGlobalConfig(): void {
-  globalConfigCache = null;
-}
```
sklein12, line 15: We weren't properly invalidating the cache on updates, this was causing issues. I don't think there's a ton of benefit here and caching is complicated.

typpo (approved):
Looks great!

PR #1721, opened 2024-09-20
chore: try to read config from directory path

andretran:
Following up from my conversation with @mldangelo. 

Originally the `--config` command for `promptfoo eval` only works when explicitly passed a file path. It blows up when passed a directory path. The common case is when a developer attempts to use one of the examples configuration files: `promptfoo eval -c examples/claude-vs-gpt`

Screenshot 2024-09-19 at 10 27 03 PM

This PR fixes that bug and extends the `--config` option to work for both passing a directory path `configs/` (with or without the trailing slash) or a file path `/configs/promptfooconfig.[yaml|yml|json|js]`. I modified it so it smartly attempts to find a file (`yaml` -> `yml` -> `json` -> `js`) file in that order.


andretran:
Also lmk if i need to squash the commits locally.

src/config.ts
```diff
@@ -223,6 +254,8 @@ export async function readConfigs(configPaths: string[]): Promise<UnifiedConfig>
     }
   }
 
+  const directoryPaths = configPaths.map(parseDirectoryPath);
```
andretran: I had to fix a bunch of downstream logic here since it was built under the assumption that you were always passing a file path where it would always parse out the file name + extension.

test/config.test.ts
```diff
@@ -823,6 +873,31 @@ describe('readConfig', () => {
...
+    });
+    jest.spyOn(fs, 'accessSync').mockImplementationOnce(() => {
+      throw new Error('File does not exist')
+    });
+    
+
+    const result = await readConfig('configs');
+
+    expect(result).toEqual(mockConfig);
+    expect(importModule).toHaveBeenCalledWith(path.normalize('configs/promptfooconfig.js'));
```
andretran: Using `path.normalize` here to transform the file path specifically for testing with node in windows environments.

mldangelo:
Thanks for the PR, @andretran! I'll ask @typpo to review it tomorrow. No need to squash the commits. Thanks again for your contribution!

src/config.ts
```diff
@@ -139,18 +139,49 @@ export async function dereferenceConfig(rawConfig: UnifiedConfig): Promise<Unifi
...
+}
+
 export async function readConfig(configPath: string): Promise<UnifiedConfig> {
   let ret: UnifiedConfig & {
     targets?: UnifiedConfig['providers'];
     plugins?: RedteamPluginObject[];
     strategies?: RedteamStrategyObject[];
   };
-  const ext = path.parse(configPath).ext;
+  const { ext, filePath } = maybeFindConfig(configPath);
```
typpo: Would it be simpler to update `configPath` if it is a directory and keep the existing logic?

in pseudocode:
```
if isDirectory(configPath) and getConfigFiles(configPath):
  configPath = getConfigFiles(configPath)[0]
  
// existing code...
```

andretran: Alright. I took your suggestions and made it a little cleaner. Now `getConfigFiles` returns a Filepath and override `configPath` if it is in fact a directory.

PR #1725, opened 2024-09-20
Update documentation for Tree-based Jailbreaks Strategy

vingiarrusso:
Using the example [documentation](https://www.promptfoo.dev/docs/red-team/strategies/tree/) and adding the "iterative:tree" strategy to a configuration causes a validation error due to a mismatch with the strategy name.  The strategy name as defined in the [redteam constants](https://github.com/promptfoo/promptfoo/blob/7d5c4ab6d7da09400fdb815eddbc1ae02be8038b/src/redteam/constants.ts#L367) is "jailbreak:tree".


mldangelo:
Thank you!

Issue #1701, opened 2024-09-17
Adding config for custom assertion


albertlieyingadrian:
### Question
Is it possible to add "config" for custom assertion? eg.

```yaml
assert:
  - type: javascript
    value: file://relative/path/to/script.js
```

I would love to pass certain parameters to the respective script as 3rd argument after `context`

```javascript
module.exports = (output, context, customArgs?) => {
  return output.length > 10;
};
```

### Reference
- https://www.promptfoo.dev/docs/configuration/expected-outputs/javascript/#external-script

albertlieyingadrian:
My workaround now
```yaml
assert:
  - type: javascript
    value: file://relative/path/to/script.js
    config:
      argA: valueA
```


```javascript
module.exports = (output, context) => {
  // Log the assert object
  const assertionMetricName = 'YOUR_METRIC_NAME'
  const assertion = context.test?.assert?.find(assertion => assertion.metric === assertionMetricName);

  const assertionConfig = assertion?.config

  // ... 
}

```

PR #1729, opened 2024-09-20
chore: support config object in javascript and python assertions

vedantr:
Partially addresses: #1701. We still need to add support for assertion sets.

Manually verified by logging the loaded assertion and the context object and also writing a custom assertion using the `promptfooconfig.yml` based on the `javascript-assert-external` example. The correct solution is to support loading / mocking the load of an actual promptfooconfig which has a custom config for assertions.

### DIFF
--- a/site/docs/configuration/expected-outputs/javascript.md
+++ b/site/docs/configuration/expected-outputs/javascript.md
@@ -143,14 +143,22 @@ assert:
   - type: javascript
     value: file://relative/path/to/script.js
     config:
-      foo: bar
+      maximumOutputSize: 10
 ```
 
-The Javascript file must export an assertion function. The context object passed into the assertion contains the optional configuration, some Here's an example:
+The Javascript file must export an assertion function. Here's an example:
 
 ```js
 module.exports = (output, context) => {
-  return output.length > 10 && context.config.foo === 'bar';
+  return output.length > 10;
+};
+```
+
+This is an example of an assertion that uses data from a configuration defined in the assertion's YML file:
+
+```js
+module.exports = (output, context) => {
+  return output.length > context.config.maximumOutputSize;
 };
 ```
 
--- a/site/docs/configuration/expected-outputs/javascript.md
+++ b/site/docs/configuration/expected-outputs/javascript.md
@@ -158,7 +158,7 @@ This is an example of an assertion that uses data from a configuration defined i
 
 ```js
 module.exports = (output, context) => {
-  return output.length > context.config.maximumOutputSize;
+  return output.length <= context.config.maximumOutputSize;
 };
 ```
 
--- a/site/docs/configuration/expected-outputs/javascript.md
+++ b/site/docs/configuration/expected-outputs/javascript.md
@@ -142,13 +142,15 @@ To reference an external file, use the `file://` prefix:
 assert:
   - type: javascript
     value: file://relative/path/to/script.js
+    config:
+      foo: bar
 ```
 
-The Javascript file must export an assertion function. Here's an example:
+The Javascript file must export an assertion function. The context object passed into the assertion contains the optional configuration, some Here's an example:
 
 ```js
 module.exports = (output, context) => {
-  return output.length > 10;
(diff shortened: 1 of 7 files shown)


src/types/index.ts
```diff
@@ -369,6 +369,9 @@ export const AssertionSchema = z.object({
   // The expected value, if applicable
   value: z.custom<AssertionValue>().optional(),
 
+  // An external object that is passed to the assertion for custom javascript asserts
+  config: z.any().optional(),
```
typpo: I think this should be `Record<string, any>`!

site/docs/configuration/expected-outputs/javascript.md
```diff
@@ -152,6 +154,14 @@ module.exports = (output, context) => {
 };
 ```
 
+This is an example of an assertion that uses data from a configuration defined in the assertion's YML file:
```
typpo, line 157: Could you please add an example to `python.md` as well?

src/assertions.ts
```diff
@@ -313,6 +313,7 @@ export async function runAssertion({
     vars: test.vars || {},
     test,
     logProbs,
+    ...(assertion.config ? { config: assertion.config } : {}),
```
typpo, line 316: If you're feeling inspired, please consider adding a test to `assertions.test.ts`

typpo (commented):
Nice change!  A couple small comments

src/types/index.ts
vedantr: So I was chatting about this with @mldangelo and it theoretically could be an object of arbitrary structure (e.g. keys don't necessarily need to be strings here). Happy to change it just be a string -> any mapping if you think that makes more sense.

site/docs/configuration/expected-outputs/javascript.md
vedantr: Done!

src/assertions.ts
vedantr: Also done :)

mldangelo (approved):
Thank you @vedantr!

BMalaichik:
it works for javascript, but doesn't work for Python

example:
```
tests:
  - vars:
      category: "chair-dining"
    assert:
      - type: python
        value: file://./evals/assert_fact_present.py
        config:
          fact: "must have height regulation feature"
```


```
import json
import logging
from typing import Any, Dict, Union


def get_assert(output: str, context) -> Union[bool, float, Dict[str, Any]]:
    try:
        output = json.loads(output)
    except:
        logging.error(f"Failed to parse output: {output}")
        raise

    facts = output.get("facts", [])
    print(">>> CONTEXT:", context.keys())
    print(">>> CONFIG:", context.get("config", {}))

    expected_fact = context.get("config", {}).get("fact", "")
    if not expected_fact:
        raise Exception("Expected fact not found in config")

    if expected_fact not in facts:
        logging.error(f"Expected fact {expected_fact} not found in output")
        return False

    return True

```


Output:

```
.........
Importing module assert_fact_present from ...model/assistant/evals ...
>>> CONTEXT: dict_keys(['prompt', 'vars', 'test', 'provider', 'providerResponse'])
>>> CONFIG: {}
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/promptfoo/0.99.1/libexec/lib/node_modules/promptfoo/dist/src/python/wrapper.py", line 39, in <module>
    result = call_method(script_path, method_name, *data)
  File "/opt/homebrew/Cellar/promptfoo/0.99.1/libexec/lib/node_modules/promptfoo/dist/src/python/wrapper.py", line 28, in call_method
    return method_to_call(*args)
  File "/..../assert_fact_present.py", line 20, in get_assert
    raise Exception("Expected facts not found in config")
Exception: Expected facts not found in config
Error running Python script: Exception: Expected fact not found in config
Stack Trace: Error: Exception: Expected fact not found in config
```


Promptfoo version 0.99.1

BMalaichik:
UPD: works well once ported to the Javascript.

Issue #1738, opened 2024-09-23
Different prompts per provider doesn't work


albertlieyingadrian:
I tried to load different prompts for different providers following this instruction https://www.promptfoo.dev/docs/configuration/parameters/#different-prompts-per-model

```yaml
prompts:
  - id: file://prompts/gpt_chat_prompt.json
    label: gpt_chat_prompt
  - id: file://prompts/llama_completion_prompt.txt
    label: llama_completion_prompt

providers:
  - id: openai:gpt-4o-mini
    prompts:
      - gpt_chat_prompt
  - id: openai:gpt-4o
    prompts:
      - gpt_chat_prompt
  - id: replicate:meta/meta-llama-3.1-405b-instruct
    label: llama-3.1-405b-instruct
    prompts:
      - llama_completion_prompt
```

I got an error by `invariant`
```shell
/Users/albertlieadrian/Documents/Random/ai-llm-playground/evals/node_modules/tiny-invariant/dist/tiny-invariant.cjs.js:14
    throw new Error(value);
          ^

Error: Invariant failed: Invalid prompt
```

I traced it back and I figured it was caused by this checking https://github.com/promptfoo/promptfoo/blob/main/src/config.ts#L339-L342

The prompt's data structure, in this case, would have no `raw`, only `id` and `label`. Hence this configuration will never be able to be executed by promptfoo.

albertlieyingadrian:
This is super hacky way to make it work -- by adding `raw` field in the config
```yaml
prompts:
  - id: file://prompts/gpt_chat_prompt.json
    label: gpt_chat_prompt
  - id: file://prompts/llama_completion_prompt.txt
    label: llama_completion_prompt

providers:
  - id: openai:gpt-4o-mini
    prompts:
      - gpt_chat_prompt
  - id: openai:gpt-4o
    prompts:
      - gpt_chat_prompt
  - id: replicate:meta/meta-llama-3.1-405b-instruct
    label: llama-3.1-405b-instruct
    prompts:
      - llama_completion_prompt
```

But seems like it's not intuitive for users and this might be a blocker for them to run even basic evals if they are gonna involve different prompts per provider @mldangelo @typpo

bfdykstra:
I also ran in to this, but only when attempting to pass a different configuration file from command line like: `npx promptfoo@latest eval --config eval_all_config.yaml`. If you include the `raw` field in the prompts, the eval runs, but with the content of the `raw` field as the prompt, not what is specified in the file

[...]

PR #1757, opened 2024-09-25
fix: prompts per provider

typpo:
Fixes issue where the "combine configs" code path crashed when loading per-provider prompts

Related to #1738


src/config.ts
```diff
@@ -338,7 +338,8 @@ export async function readConfigs(configPaths: string[]): Promise<UnifiedConfig>
       config.prompts.forEach((prompt) => {
         invariant(
           typeof prompt === 'string' ||
-            (typeof prompt === 'object' && typeof prompt.raw === 'string'),
+            (typeof prompt === 'object' &&
+              (typeof prompt.raw === 'string' || typeof prompt.label === 'string')),
```
typpo: this is the actual change

PR #1761, opened 2024-09-25
fix: invert --no-interactive desc, add to docs

kentyman23:


mldangelo (approved):
Thank you so much @kentyman23

PR #1763, opened 2024-09-25
chore(providers): update bedrock integration to support Llama 3.2 models

aristsakpinis93:


mldangelo:
Thank you @aristsakpinis93!

PR #1765, opened 2024-09-26
fix: redteam aliases

typpo:
Fix high-level aliases (such as `owasp:llm`) and granular aliases (such as `owasp:llm:02`).

### DIFF
--- a/src/validators/redteam.ts
+++ b/src/validators/redteam.ts
@@ -14,6 +14,7 @@ import {
   DEFAULT_NUM_TESTS_PER_PLUGIN,
   ALIASED_PLUGIN_MAPPINGS,
   type Strategy,
+  type Plugin,
 } from '../redteam/constants';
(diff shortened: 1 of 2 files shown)


eneelou:
I found a couple of issues:

**1. Sub-collections "almost" work**

- "owasp:llm" doesn't use sub-collections and ignores ['harmful'] or ['pii'] collections
- "owasp:llm:01" uses sub-collections correctly and transforms ['harmful'] into 23 plugins

**2. Strategies stopped working**

- "owasp:llm" doesn't pull any strategies but should use ['prompt-injection', 'jailbreak']
- "owasp:llm:01" doesn't pull any strategies but should use ['prompt-injection', 'jailbreak']

typpo:
@eneelou Thanks for the careful look.  I think I have it all done now, and added tests.

The mappings themselves are questionable and/or incomplete, but that is a separate problem to solve.

PR #1766, opened 2024-09-26
fix: json parser handling for many braces

typpo:
This also fixes regular dataset generation

### DIFF
--- a/test/assertions.test.ts
+++ b/test/assertions.test.ts
@@ -1395,22 +1395,6 @@ describe('runAssertion', () => {
     });
   });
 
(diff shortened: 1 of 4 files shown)


mldangelo (approved):
thanks!

test/assertions.test.ts
```diff
@@ -1395,22 +1395,6 @@ describe('runAssertion', () => {
     });
   });
 
-  it('should pass when the contains-json assertion passes with valid and invalid json', async () => {
```
typpo, line 1398: This is tested downstream

PR #1776, opened 2024-09-26
Normalize eval results in db

sklein12:
Big Changes:
- New Table: https://github.com/promptfoo/promptfoo/pull/1776/files#diff-660e00170b1266d7afb7e2676c50217c2ac27bb283dff59b0e82660daf775307R1-R25
- New output format: [examples/simple-cli/output.json](https://github.com/promptfoo/promptfoo/pull/1776/files#diff-e0bb0dfb8b7d049e5eee6104980e31cad839073570ed090b1743b82a02cb8751R2)

Normalize the eval results by storing them in their own table.
- Added a new `Eval`, `EvalResults` and `Provider` classes
- All eval results are stored as their own row in `EvalResuls`. They're no longer stored on the `Eval`
- For now all existing evals will stay as is in the database, there's no migration path. All import/export, share functions, and the app can read both versions of the Eval.
- Updates `EvaluateSummary` to Version 3, it removes the `table` from the output and adds a `prompts` property - https://github.com/promptfoo/promptfoo/pull/1776/files#diff-e56305a32e55804c903f463f6bbbf7a3df0e9f4dd2e4b6ea9db63120428424cbR367-R378
   - This removes the table from all file outputs, cutting file size by 50%
- Does not generate a table during evaluation cutting memory footprint by 50%
- Updating a newer Eval in the UI will only update the individual row and not the entire eval.


jest.setup.js
```diff
@@ -2,7 +2,10 @@ import nock from 'nock';
 
 // Disable all real network requests
 nock.disableNetConnect();
+nock.enableNetConnect('127.0.0.1');
 
 nock.emitter.on('no match', (req) => {
-  console.error(`Unexpected HTTP request: ${req.method} ${req.href}`);
+  if (!req.host.includes('127.0.0.1') && !req.host.includes('localhost')) {
+    console.error(`Unexpected HTTP request: ${req.method} ${req.href}`);
+  }
```
sklein12, line 10: I've got express tests now that make local requests

jest.config.ts
```diff
@@ -5,6 +5,7 @@ const config: Config = {
   collectCoverage: true,
   coverageDirectory: '.coverage',
   coverageProvider: 'v8',
+  coverageReporters: ['json', 'lcov', 'clover'],
```
sklein12: sorry, I had to remove the test coverage from being printed on every test run, it was unusable.

src/commands/eval.ts
```diff
@@ -196,6 +196,7 @@ export async function doEval(
         );
       }
     } else {
+      await eval_.delete();
```
sklein12: if the user requested not to save the results, we need to delete them

jest.config.ts
mldangelo: I still get a lot of value out of this (I use it multiple times per day). What makes it unusable? If we remove the text coverage report we might as well remove the other ones because we don't use them anywhere.  

I am happy to start running `npx jest --coverage` if you end up still wanting to remove this.

.jest/setEnvVars.js
```diff
@@ -4,3 +4,4 @@ process.env.AZURE_OPENAI_API_KEY = 'foo';
 process.env.HF_API_TOKEN = 'foo';
 process.env.OPENAI_API_KEY = 'foo';
 delete process.env.PROMPTFOO_REMOTE_GENERATION_URL;
+process.env.IS_TESTING = 'true';
```
mldangelo: nit, sort

src/commands/eval.ts
```diff
@@ -121,25 +121,27 @@ export async function doEval(
...
       TestSuite Schema Validation Error:
-      
         ${JSON.stringify(testSuiteSchema.error.format())}
-      
       Please review your promptfooconfig.yaml configuration.`),
       );
     }
 
-    const summary = await evaluate(testSuite, {
+    // We're going to write everything to the database and then delete it later so we can eventually(tm) remove the results array and table from the evaluate function. Nothing should be held in memory.
```
mldangelo: leave this as a todo or note

drizzle/0006_harsh_caretaker.sql
```diff
@@ -0,0 +1,42 @@
...
+);
+--> statement-breakpoint
+CREATE TABLE `providers` (
+	`id` text PRIMARY KEY NOT NULL,
+	`provider_id` text NOT NULL,
+	`options` text NOT NULL
+);
+--> statement-breakpoint
+ALTER TABLE `evals` ADD `prompts` text;--> statement-breakpoint
+CREATE INDEX `eval_result_eval_id_idx` ON `eval_results` (`eval_id`);
```
mldangelo, line 42: Do you think this is sufficient for indices? Wondering if we should also index fields like created_at

```
CREATE INDEX idx_eval_results_created_at ON eval_results (created_at);
CREATE INDEX idx_eval_results_test_case_idx ON eval_results (test_case_idx);
```

test/factories/evalFactory.ts
mldangelo, line 1: Very happy to see this file

src/validators/providers.ts
```diff
@@ -102,6 +102,7 @@ export const ApiProviderSchema = z.object({
   label: z.custom<ProviderLabel>().optional(),
   transform: z.string().optional(),
   delay: z.number().optional(),
+  config: z.any().optional(),
```
mldangelo, line 105: what is this for?

src/app/src/pages/dashboard/components/TopFailingCategories.tsx
```diff
@@ -31,12 +31,17 @@ const TopFailingCategories: React.FC<TopFailingCategoriesProps> = ({ evals }) =>
...
+              ? 0
+              : (data.currentFailCount / data.currentTotalCount) * 100;
+
           return (
             <ListItem key={index}>
               <Box sx={{ width: '100%' }}>
                 <ListItemText
                   primary={displayName}
-                  secondary={`${((data.currentFailCount / data.currentTotalCount) * 100).toFixed(1)}% failing (${data.currentFailCount}/${data.currentTotalCount} probes)`}
+                  secondary={`${failPercentage}% failing (${data.currentFailCount}/${data.currentTotalCount} probes)`}
```
mldangelo, line 44: don't forget the toFixed(1)
```suggestion
                  secondary={`${failPercentage.toFixed(1)}% failing (${data.currentFailCount}/${data.currentTotalCount} probes)`}
```

src/app/src/pages/eval/components/Eval.tsx
```diff
@@ -54,8 +60,13 @@ export default function Eval({
   const fetchEvalById = React.useCallback(
     async (id: string) => {
       const resp = await callApi(`/results/${id}`, { cache: 'no-store' });
+      if (!resp.ok) {
```
mldangelo, line 63: out of curiosity, when does this happen? Should we put improving this flow on the backlog?

src/app/src/pages/eval/components/ResultsTable.tsx
```diff
@@ -195,18 +196,30 @@ function ResultsTable({
...
-          const response = await callApi(`/eval/${evalId}`, {
-            method: 'PATCH',
-            headers: {
-              'Content-Type': 'application/json',
-            },
-            body: JSON.stringify({ table: newTable }),
-          });
+          let response;
+          if (version >= 4) {
+            response = await callApi(`/eval/${evalId}/results/${resultId}/rating`, {
```
mldangelo, line 207: this is cool

src/app/src/pages/eval/components/store.ts
```diff
@@ -66,8 +71,19 @@ export const useStore = create<TableState>()(
       author: null,
       setAuthor: (author: string | null) => set(() => ({ author })),
 
+      version: 3,
```
mldangelo: What does this mean? backend stores in version 4, converts everything to version 3, and frontend displays version 3?

src/commands/eval.ts
```diff
@@ -132,25 +132,28 @@ export async function doEval(
...
+    const eval_ = await Eval.create(config, testSuite.prompts);
```
mldangelo: nit, consider variable name.

src/commands/eval.ts
mldangelo: Do we want to store this in memory entirely if the user does not want to save it? Thinking about the case where promptfoo crashes and this never gets called. May not be important.

src/commands/list.ts
```diff
@@ -21,14 +22,20 @@ export function listCommand(program: Command) {
...
+        const prompts = evl.getPrompts();
+        const vars = await evl.getVars();
+        return {
+          'Eval ID': evl.id,
+          Description: evl.description || '',
+          Prompts: prompts.map((p) => sha256(p.raw).slice(0, 6)).join(', ') || '',
+          Vars: vars.map((v) => v).join(', ') || '',
+        };
+      });
+      const tableData = await Promise.all(tableDataPromises);
```
mldangelo: at some point this will cause the db to crash. We should think about a concurrency limit here

src/commands/list.ts
```diff
@@ -21,14 +22,20 @@ export function listCommand(program: Command) {
...
+      // @ts-ignore
```
mldangelo: why ts-ignore?

src/database/index.ts
```diff
@@ -15,7 +15,7 @@ export function getDbSignalPath() {
 
 export function getDb() {
   if (!dbInstance) {
-    const sqlite = new Database(getDbPath());
+    const sqlite = new Database(process.env.IS_TESTING ? ':memory:' : getDbPath());
```
mldangelo, line 18: we could also do this when the eval is not written to the db

src/evaluator.ts
```diff
@@ -617,6 +632,13 @@ class Evaluator {
       if (options.progressCallback) {
         options.progressCallback(results.length, runEvalOptions.length, index, evalStep);
       }
+      const { testIdx, promptIdx } = evalStep;
+      try {
+        await this.dbRecord.addResult(row, promptIdx, testIdx, evalStep.test);
+      } catch (error) {
+        logger.error(`Error adding result: ${error}`);
```
mldangelo: nit, do one logger.error here instead of 2

src/models/eval.ts
```diff
@@ -0,0 +1,457 @@
...
+      prompts: this.prompts,
+      datasetId: this.datasetId || null,
+    };
+
+    return results;
+  }
+
+  async delete() {
+    const db = getDb();
+    await db.transaction(() => {
```
mldangelo, line 446: we should also check that we can still delete arbitrary evals from the db (given the new FKs)

src/models/evalResult.ts
```diff
@@ -0,0 +1,158 @@
...
+      await db.update(evalResultsTable).set(this).where(eq(evalResultsTable.id, this.id));
+    } else {
+      const result = await db.insert(evalResultsTable).values(this).returning();
+      this.id = result[0].id;
+    }
+  }
+
+  toEvaluateResult(): EvaluateResult {
+    return {
+      id: this.id,
```
mldangelo, line 162: nit, sort these if the order is not important

src/prompts/utils.ts
```diff
@@ -83,3 +84,8 @@ export function normalizeInput(
   // numbers, booleans, etc
   throw new Error(`Invalid input prompt: ${JSON.stringify(promptPathOrGlobs)}`);
 }
+
+export function hashPrompt(prompt: Prompt): string {
+  const label = prompt.label || prompt.display || prompt.raw;
```
mldangelo: I would almost prefer the opposite order here or a concatenation of all. Labels are likely going to stay the same even if the prompt changes across evals. If this means something different please add a comment.

src/server/routes/evalRoutes.ts
mldangelo, line 1: delete?

mldangelo (commented):
This is great! I left a few nits and a few questions. Thank you so much for working on this. It's going to be huge for people running large evals.

jest.config.ts
typpo: the main annoyance imo is it forces you to scroll up past coverage

src/app/src/pages/report/components/Report.tsx
```diff
@@ -301,7 +307,7 @@ const App: React.FC = () => {
               size="small"
               label={
                 <>
-                  <strong>Model:</strong> {selectedPrompt.provider}
+                  <strong>Model:</strong> {selectedPrompt?.provider}
```
typpo: if selectedPrompt is actually undefined, let's hide the entire pill/section instead of displaying "undefined"

src/commands/eval.ts
typpo: would it make more sense to add a `write` or `save` param to `Eval.create`, rather than writing and deleting?

drizzle/0006_harsh_caretaker.sql
sklein12: Yes I think so. The only query we do right now is select by `evalId`.  We can add those as needed later.

jest.config.ts
sklein12: Yea, it just takes up a ton of the screen when you're running tests.

src/app/src/pages/eval/components/Eval.tsx
sklein12: The only case I ran into is when it doesn't exist

src/app/src/pages/eval/components/store.ts
sklein12: This sets the default to version 3.... probably shouldn't do that. ty

src/app/src/pages/report/components/Report.tsx
sklein12: i'll do the same for the other pill too

src/commands/eval.ts
sklein12: any suggestions?

src/database/index.ts
sklein12: that's a really cool idea

src/prompts/utils.ts
sklein12: yea, I would agree here, we should maybe combine the label and raw to do the hash

sklein12: sadly this is going to mess up people who already have prompts, this would be a breaking change

src/validators/providers.ts
sklein12: this existed on the class but not on the schema, so it was missing from the typescript definition

src/commands/eval.ts
sklein12: I like @mldangelo's idea of just using the in memory database when that option is enabled

src/commands/list.ts
sklein12: I could just do it in a transaction, that would be better for now

src/app/src/pages/dashboard/components/TopFailingCategories.tsx
sklein12: fixed above 

https://github.com/promptfoo/promptfoo/pull/1776/files#diff-f30e03a442089081866c3be8ee9a2a8b09169ab91aaf768b9e64ce908c0b403fR37

src/commands/eval.ts
sklein12: I just did it for real

src/models/eval.ts
sklein12: what do you mean?  

you will need to do it via this method because of the FK constraints, these existed before because the prompt and dataset FK's already existed

mldangelo: Every time we've added a new table or foreign key we've broken delete. Comment was an instinctual reaction. This looks good.

src/prompts/utils.ts
mldangelo: @typpo can you look at this line and let us know your thoughts?

jest.config.ts
```diff
@@ -2,9 +2,7 @@
 import type { Config } from 'jest';
 
 const config: Config = {
-  collectCoverage: true,
-  coverageDirectory: '.coverage',
-  coverageProvider: 'v8',
```
mldangelo: Can you keep these two lines for when I run the coverage report via cli?

src/commands/list.ts
mldangelo: don't forget about this

mldangelo (approved):

src/app/src/pages/eval/components/ResultsTable.tsx
```diff
@@ -300,7 +313,7 @@ function ResultsTable({
   const numGoodTests = React.useMemo(
     () =>
       head.prompts.map((_, idx) =>
-        body.reduce((acc, row) => acc + (row.outputs[idx].pass ? 1 : 0), 0),
+        body.reduce((acc, row) => acc + (row.outputs[idx]?.pass ? 1 : 0), 0),
```
sklein12, line 316: This could be null because results get created as they happen now and may not all come in at once, so a test for prompt 2 could be done before prompt 1.

PR #1782, opened 2024-09-27
feat(redteam): improve generation process and file handling

mldangelo:
Red team tests are now stored in a separate file and referenced in the main
configuration, rather than being embedded directly in the existing
promptfooconfig or promptfooconfig being copied into an adjacent file. This
should assist with both onboarding and code test organization. We now have
a single source of truth for red team tests. Changes should be backwards
compatible.

NOTE: Further testing is required before shipping this feature.


src/redteam/commands/generate.ts
```diff
@@ -142,62 +130,40 @@ export async function doGenerateRedteam(options: RedteamCliGenerateOptions) {
...
-      ? 'promptfoo eval'
-      : `promptfoo eval -c ${path.relative(process.cwd(), configPath)}`;
-    logger.info('\n' + chalk.green(`Run ${chalk.bold(`${command}`)} to run the red team!`));
-  } else {
-    writePromptfooConfig({ tests: redteamTests }, 'redteam.yaml');
-  }
+    },
+    tests: [
+      ...new Set([
+        ...(existingYaml?.tests || []),
```
typpo, line 152: I would like to preserve the rest of the promptfooconfig file exactly, so let's try regexing this :scream: 

We need to handle the following cases:
- `tests: ...` is a oneliner
- `tests:` exists and has a bunch of indented stuff under it
- `tests` key does not exist

PR #1791, opened 2024-09-28
feat: add `redteam run` command

typpo:


src/redteam/commands/run.ts
```diff
@@ -0,0 +1,127 @@
...
+    .option('--delay <number>', 'Delay in milliseconds between API calls', (val) =>
+      Number.parseInt(val, 10),
+    )
+    .option('--remote', 'Force remote inference wherever possible', false)
+    .action(async (opts: RedteamRunOptions) => {
+      try {
+        if (opts.remote) {
+          cliState.remote = true;
+        }
+        await doRedteamRun(opts);
```
sklein12, line 124: Process exit 0 after it's complete?

typpo: that is the default exit code, so I don't think it's necessary?  `eval` is called by the run and will set the exit code as well

PR #1792, opened 2024-09-29
docs: provider index

mldangelo:
Adds a provider index page instead of the docusaurus-generated one. Also documents the go provider

### DIFF
--- a/site/docs/providers/go.md
+++ b/site/docs/providers/go.md
@@ -89,26 +89,3 @@ Or in the CLI:
 ```
 promptfoo eval -p prompt1.txt prompt2.txt -o results.csv -v vars.csv -r 'file://path/to/your/script.go'
 ```
-
-## Additional Functions
-
-In addition to `CallApi`, you can implement other functions in your Go script:
-
-```go
-func SomeOtherFunction(prompt string, options map[string]interface{}, context map[string]interface{}) (map[string]interface{}, error) {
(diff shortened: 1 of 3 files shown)


site/docs/providers/go.md
```diff
@@ -0,0 +1,106 @@
+# Custom Go Provider
+
+The Go provider allows you to use a Go script as an API provider for evaluating prompts. This is useful when you have custom logic or models implemented in Go that you want to integrate with your test suite.
+
```
typpo:
```suggestion
---
sidebar_label: Custom Go (Golang)
---

# Custom Go Provider

The Go (`golang`) provider allows you to use a Go script as an API provider for evaluating prompts. This is useful when you have custom logic or models implemented in Go that you want to integrate with your test suite.

:::info
The golang provider currently experimental
:::

```

site/docs/providers/go.md
```diff
@@ -0,0 +1,106 @@
...
+    return CallApi(prompt, options, context)
+}
+```
+
+To use a specific function, specify it in the provider ID:
+
+```yaml
+providers:
+  - id: 'file://path/to/your/script.go:SomeOtherFunction'
+```
```
typpo: not sure but I think I didn't actually build this yet

typpo (approved):
very nice docs

PR #1799, opened 2024-10-01
feat: prompts as python classmethods

akonarski-ds:
This adds support for prompts defined as `classmethods` or `staticmethods` in Python classes. This update allows prompts to be specified in the following way:

```python
class Prompts:

    @staticmethod
    def prompt1(arg1, arg2):
        return arg1 + arg2
        
    @staticmethod
    def prompt2(arg1, arg2):
        return arg1 - arg2
```
and then in the `promptfooconfig.yaml` file:
```yaml
prompts:
  - file:///root/path/to/prompt.py:Prompt.prompt1
  - file:///root/path/to/prompt.py:Prompt.prompt2
```

Previously, only module-level functions were supported.


mldangelo:
Thanks so much for the PR @akonarski-ds! Your code looks good! Thanks for adding a test too! 

Does this mirror something that deepsense.ai is doing in Python? I'm curious about the use case for this syntax. Would you mind adding an example to https://github.com/promptfoo/promptfoo/blob/main/examples/custom-prompt-function ?

akonarski-ds:
Thanks for the quick response @mldangelo ! 

Yes, in one of the projects at deepsense.ai, we have a class that represents a Prompt and implements custom logic for it. Without this change, we would have to add something like:

```python
def topromptfoo1(context):
    return PromptClass.prompt1(context)
```

for every prompt definition we have. This change makes it much more convenient to work with prompts defined as Python classes.

I've added an example to show how this can be used: https://github.com/promptfoo/promptfoo/pull/1799/commits/88aa8249616d2077c7813ea160cdb8786db86a5f

mldangelo:
Thank you so much @akonarski-ds! We'll get this merged and in the next release (probably tomorrow)

PR #1815, opened 2024-10-03
feat(provider-fal): allow prompt config overrides

drochetti:
This PR adds config per prompt for the fal provider, as described here: https://www.promptfoo.dev/docs/configuration/parameters/#prompt-configs

### DIFF
--- a/src/providers/fal.ts
+++ b/src/providers/fal.ts
@@ -4,11 +4,24 @@ import fetch from 'node-fetch';
 import { getCache, isCacheEnabled } from '../cache';
 import { getEnvString } from '../envars';
 import logger from '../logger';
-import type { ApiProvider, EnvOverrides, ProviderResponse } from '../types';
+import type { ApiProvider, EnvOverrides, Prompt, ProviderResponse } from '../types';
+import { PromptSchema } from '../validators/prompts';
 
 type FalProviderOptions = {
   apiKey?: string;
 };
+
+function parsePrompt(prompt: string): Prompt {
+  try {
+    const json = JSON.parse(prompt);
+    if (typeof json === 'object') {
+      return PromptSchema.parse(json);
+    }
+    return { label: json.toString(), raw: json.toString() };
+  } catch {
+    return { label: prompt, raw: prompt };
+  }
+}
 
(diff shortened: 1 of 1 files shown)


src/providers/fal.ts
```diff
@@ -4,12 +4,25 @@ import fetch from 'node-fetch';
...
+    const json = JSON.parse(prompt);
+    if (typeof json === 'object') {
+      return PromptSchema.parse(json);
+    }
+    return { label: json.toString(), raw: json.toString() };
+  } catch {
+    return { label: prompt, raw: prompt };
+  }
+}
+
```
drochetti, line 25: @mldangelo hey Michael, is there a common way of doing this in the codebase? I look on some other providers but it didn't seem like there's a standard around this. Let me know if there's another way of handling prompts as raw strings or prompts as JSON strings.

For context our customer is trying to use a config per prompt: https://www.promptfoo.dev/docs/configuration/parameters/#prompt-configs

mldangelo: @drochetti Are you sure they are using that feature? It was added in https://github.com/promptfoo/promptfoo/pull/1391 only to openai and was later removed (possibly by mistake). 

The standard way of doing parsePrompt looks more like the function below: 
```ts 
function parsePrompt(prompt: string): Messages[] {
  try {
    const json = JSON.parse(prompt);
    return json as Messages[];
  } catch { 
    return [{ role: 'user', content: prompt }];
  }
  }
  ```
  
By the time the prompt gets to the provider, it has already been rendered so we don't need the original label, raw format. 

If you want to throw some time on my calendar https://cal.com/michael-dangelo/30min I can help you address this today.

Comment (drochetti) [reply src/providers/fal.ts:25]: @mldangelo yes, in our case the model being used does image to image, so even if we make the seed a constant in the provider config, the image being passed is still per prompt.

Is there another way of accomplishing that?

Comment (drochetti) [reply src/providers/fal.ts:25]: Here's the prompt config I'm using for reference:

```yml
prompts:
  - label: 'Prompt #1'
    raw: >-
      Some long prompt
    config:
      seed: 123
      image_url: >-
        https://url.to.image/for/img-2-img/image.png
# ...
```

PR #1816, opened 2024-10-03
feat(redteam): Add Math prompt Attack Plugin

AISimplyExplained:
This plugin introduces a Math-Prompt attack, where harmful instructions are encoded as mathematical problems. The goal is to trick large language models (LLMs) into proving or solving the encoded instructions, leading to unintended responses.

Based on this research paper: https://arxiv.org/abs/2409.11445


.idea/.gitignore
mldangelo, line 1: Can you please remove your .idea directory?

src/app/src/pages/report/components/constants.ts
```diff
@@ -14,6 +14,7 @@ export const riskCategories = {
     'ssrf',
     'indirect-prompt-injection',
     'cross-session-leak',
+    'mathprompt', // Added mathprompt to Security Risk
```
mldangelo, line 17: Please clean up the comment here and in all files below

```suggestion
    'mathprompt',
```

src/redteam/plugins/mathPrompt.ts
```diff
@@ -0,0 +1,77 @@
+import dedent from 'dedent';
```
mldangelo, line 1: let's throw a comment in this file referencing https://arxiv.org/html/2409.11445v1

mldangelo (commented):
This is great! Because mathprompt is a jailbreaking technique and not designed to generate a specific type of harmful content - it may make more sense to implement this as a strategy rather than a plugin (strategies modify the generations of other plugins). I am happy to pair on this with you if you would like to throw time on my calendar. 

Thank you @AISimplyExplained!

PR #1834, opened 2024-10-07
chore: move skip logic to generate

typpo:
Right now the logic is in `run`, which works great for `run` but it would be nice if `generate` didn't regenerate unnecessarily too.

### DIFF
--- a/src/redteam/commands/generate.ts
+++ b/src/redteam/commands/generate.ts
@@ -49,14 +49,14 @@ export async function doGenerateRedteam(options: Partial<RedteamCliGenerateOptio
 
   // Check for updates to the config file and decide whether to generate
   let shouldGenerate = options.force;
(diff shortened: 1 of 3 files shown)


src/redteam/commands/generate.ts
```diff
@@ -27,6 +29,12 @@ import type { RedteamStrategyObject, SynthesizeOptions } from '../types';
 import type { RedteamFileConfig, RedteamCliGenerateOptions } from '../types';
 import { shouldGenerateRemote } from '../util';
 
+function getConfigHash(configPath: string): string {
+  const content = fs.readFileSync(configPath, 'utf8');
+  const version = packageJson.version;
```
mldangelo: in constants.ts 

```ts 
import packageJson from '../package.json';
export const VERSION = packageJson.version;
```

mldangelo (approved):

PR #1835, opened 2024-10-07
fix(provider): fal prompt config overrides

drochetti:
**Notes:**

1. The fal provider now handles config per prompt correctly using the `context.prompt`
2. When passing a prompt object via the Web UI YAML editor, the format described here is being passed as a serialized `string` and the prompt object with the config is lost and not present on the context.

cc @mldangelo

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -24,6 +24,7 @@ import type {
   Scenario,
 } from './types';
 import { readFilters, writeMultipleOutputs, writeOutput } from './util';
+import { PromptSchema } from './validators/prompts';
 
 export * from './types';
 
@@ -56,6 +57,8 @@ async function evaluate(testSuite: EvaluateTestSuite, options: EvaluateOptions =
             };
           } else if (typeof promptInput === 'string') {
             return readPrompts(promptInput);
+          } else if (typeof promptInput === 'object') {
+            return PromptSchema.parse(promptInput);
           } else {
             return {
               raw: JSON.stringify(promptInput),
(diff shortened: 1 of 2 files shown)


src/providers/fal.ts
```diff
@@ -56,6 +56,7 @@ class FalProvider<Input = any> implements ApiProvider {
     const input = {
       prompt,
       ...this.input,
+      ...(context.prompt?.config ?? {}),
```
drochetti, line 59: @mldangelo this is the change we missed on the last update to ensure we can override the inference parameters using the prompt config

mldangelo: Ah! I am sorry I missed this.

src/index.ts
```diff
@@ -56,7 +57,13 @@ async function evaluate(testSuite: EvaluateTestSuite, options: EvaluateOptions =
...
           } else if (typeof promptInput === 'string') {
             return readPrompts(promptInput);
-          } else {
+          }
+          try {
+            return PromptSchema.parse(promptInput);
+          } catch (error) {
+            console.warn(
+              `Prompt input is not a valid prompt schema: ${error}\nFalling back to serialized JSON as raw prompt.`,
+            );
```
drochetti, line 66: @mldangelo Context on this:

When testing via the WebUI (via `npm run dev`, not using .yaml + eval), using the following prompt format for config per prompt:

```yml
prompts:
  - label: 'Prompt #1'
    raw: >-
      a car
    config:
      seed: 123
      image_url: >-
        https://valid.url.to/image.png
```

The prompt object was getting serialized to its string, like:

```js
{
  raw: '{"label":"Prompt #1","raw":"a car","config":{"seed":123,"image_url":"https://valid.url.to/image.png"}}',
  label: '{"label":"Prompt #1","raw":"a car","config":{"seed":123,"image_url":"https://valid.url.to/image.png"}}'
}
```

mldangelo:
Thank you @drochetti we'll get this deployed tomorrow

PR #1839, opened 2024-10-08
feat(providers): Add support for an optional `responseSchema` file to google gemini

aud:
This adds support for passing an optional `responseSchema` parameter to the
Palm provider, which allows you to load an external file as the return schema.
For eg. this is now possible:

```yaml
providers:
  - id: google:gemini-pro
    config:
      responseSchema: file://test.json
```


mldangelo:
Thank you @adm!

PR #1853, opened 2024-10-09
feat: env variable to host pf at a different url path then base

sklein12:


bennibbelink:
Couple questions:

- Should `VITE_PUBLIC_BASENAME` be added to vite-env.d.ts ?
- To build the image with this env var set, do I need to add `ENV VITE_PUBLIC_BASENAME=...` to the Dockerfile or should this be able to be passed as a build arg?

sklein12:
> Couple questions:
> 
> * Should `VITE_PUBLIC_BASENAME` be added to vite-env.d.ts ?
> * To build the image with this env var set, do I need to add `ENV VITE_PUBLIC_BASENAME=...` to the Dockerfile or should this be able to be passed as a build arg?

yes it should be, good call.

Both env variables need to be set before you run `npm run build` in the Dockerfile. I'll update the dockerfile to take those as args to the docker build step

sklein12:
@bennibbelink ok, could you give that a shot?  the build args are the same as the environment variable names

bennibbelink:
`docker build -f .\Dockerfile --build-arg VITE_PUBLIC_BASENAME=/pfoo/ --build-arg VITE_PUBLIC_PROMPTFOO_REMOTE_API_BASE_URL=http://localhost/pfoo/ .` 

The site is loading as expected but the api calls are still to `/api` instead of `/pfoo/api`

image

sklein12:
> `docker build -f .\Dockerfile --build-arg VITE_PUBLIC_BASENAME=/pfoo/ --build-arg VITE_PUBLIC_PROMPTFOO_REMOTE_API_BASE_URL=http://localhost/pfoo/ .`
> 
> The site is loading as expected but the api calls are still to `/api` instead of `/pfoo/api`
> 
> image

@bennibbelink got it, I think I see the issue. Let me test real quick.

sklein12:
@bennibbelink ok, the problem was it was using this env variable `PROMPTFOO_REMOTE_API_BASE_URL`

This worked for me to set the api base url ` docker build --build-arg PROMPTFOO_REMOTE_API_BASE_URL=http://localhost:4848 -t promptfoo-oss:latest . `

bennibbelink:
Works for me as well, thanks @sklein12!

rvernica:
When using the _promptfoo_ Docker image in a separate project, I have my own _Dockerfile_ starting with:
```
FROM ghcr.io/promptfoo/promptfoo:latest
```
Can I set the _VITE_PUBLIC_BASENAME_ _build-arg_ or _ENV_ variable and re-build?

sklein12:
@rvernica you need to build the image from scratch, that env variable is hard coded into the front end code when the app is built on this line https://github.com/promptfoo/promptfoo/blob/main/Dockerfile#L32

PR #1854, opened 2024-10-09
fix: Only persist custom API url in local storage if it's set through the UI

sklein12:
Persisting this all the time creates weird side effects if you're developing and testing env variables, etc...


mldangelo (approved):
When you merge this can you change the title to something that will make sense to the general public?

PR #1855, opened 2024-10-09
feat: node package provider

ktalebian:
Currently, if you install an npm package that provides additional providers, you'll have to define them as 

```yaml
providers:
  - id: node_modules/@mycompany/mypackage/dist/path/to/providers/Provider.js
     ...
```

This PR adds support for `custom:` to instead do:

```yaml
providers:
  - id: custom:@mycompany/mypackage:Provider
     ...
```

It does so by reading the `package.json`, resolving the `main` entry point, importing it and looking for `Provider.` For example:

```js
# dist/index.js
export * as providers from './providers';
// let's say providers.AwesomeProvider exists
``` 

can then be provided as `package:@mycompany/mypackage:providers.AwesomeProvider`


mldangelo:
Hey @ktalebian! This is a great addition – please take a quick look at the [contribution guide for adding a new provider](https://www.promptfoo.dev/docs/contributing/#adding-a-new-provider). I'd love to document this with an example and add it to the provider docs on our website. Thanks for the work on this! Let me know if you need any help or have questions—happy to walk you through any part of the process or take on subtasks as needed.

ktalebian:
@mldangelo could you tell me what you are looking for? This is not technically adding a new provider - it's making it easier to for the end-user to provider their implemented provider.

typpo:
Hey @ktalebian, I renamed this provider to `package`, fixed the lint, and also added an example so other people can see how to use it.

One thing I noticed is due to the use of `basePath`, `promptfooconfig.yaml` has to be in the same directory as `node_modules`.  This is probably not desirable.  Any ideas?

ktalebian:
@typpo that should be okay no? You would for example have a `my-application-benchmark` repo that hosts your `promptfooconfig.yaml` which also contains the `node_modules`. 

Are you saying that you might have something like

```
/
  package.json
  configs/
    promptfooconfig.redteam.yaml
  node_modules/
```

And in that case this would cause an issue?

mldangelo:
Thanks so much, @ktalebian! I’ve added tests and moved the provider logic to a new file. I’m going to leave it undocumented for now as I believe this solution is tailored specifically to your use case. Please feel free to reach out if you encounter any issues or have additional questions / comment / feature requests / issues / etc!

Issue #1867, opened 2024-10-11
`equals` failure description seems inverted


panozzaj:
Hi, I started using PromptFoo recently and it's been really helpful for doing some evals! Wanted to raise something small. Open to being incorrect about the expected behavior since I am new to this library. Also, perhaps the behavior is correct but the documentation is misleading somehow.

**Describe the bug**

The [docs for `equality`](https://www.promptfoo.dev/docs/configuration/expected-outputs/deterministic/#equality) say:

> The equals assertion checks if the LLM output is equal to the expected value.

> Example:

```
assert:
  - type: equals
    value: 'The expected output'
```

However, when I specify the following `test` as one of the `tests`: 

```
{
  "type": "equals",
  "value": "0",
  "transform": "output.answer",
},
```

If it fails (perhaps the model returns `{"answer": "1"}`), I see an failure message of:

`[FAIL] Expected output "0" to equal "1"`

I would expect it to be the other way around:

`[FAIL] Expected output "1" to equal "0"`

Since the output was actually 1.

**To Reproduce**

Here is a small JSON config test case:

```
{
  "providers": ["echo"],
  "prompts": ["1"],
  "tests": [
    {
      "assert": [
        {
          "type": "equals",
          "value": 0
        }
      ]
    }
  ]
}
```

=>

```
┌──────────────────────────────────────────
│ [echo] 1
├──────────────────────────────────────────
│ [FAIL] Expected output "0" to equal "1"
│ ---
│ 1
└──────────────────────────────────────────
```

**Expected behavior**

See above.

**Screenshots**

N/A

**System information (please complete the following information):**

- Promptfoo version: 0.92.2 (latest)

**Additional context**

N/A

mldangelo:
Hi @panozzaj, thanks for finding this! I completely agree. Here is the relevant line: https://github.com/promptfoo/promptfoo/blob/main/src/assertions.ts#L389. I’d be happy to accept a PR from you if you'd like credit for fixing it. If not, I’ll take care of this later today.

panozzaj:
OK, I will do that now and PR, thanks for the pointer!

PR #1868, opened 2024-10-11
fix: equality failure message

panozzaj:
Fixes https://github.com/promptfoo/promptfoo/issues/1867 per conversation there

I ran the tests and fixed test failures related to that file (which seemed reasonable since we switched the order of the output)


mldangelo:
Thank you!

PR #1883, opened 2024-10-14
feat: Redteam http target tester

sklein12:
Takes an HTTP provider config and tests it out. Uses an LLM agent hosted on promptfoo cloud to validate the results.

image

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -84,7 +84,13 @@ export async function fetchWithCache<T = any>(
     try {
       const data = JSON.stringify(format === 'json' ? JSON.parse(responseText) : responseText);
       if (!response.ok) {
-        errorResponse = data;
+        if (responseText == '') {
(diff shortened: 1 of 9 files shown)


src/cache.ts
```diff
@@ -84,7 +84,13 @@ export async function fetchWithCache<T = any>(
...
       const data = JSON.stringify(format === 'json' ? JSON.parse(responseText) : responseText);
       if (!response.ok) {
-        errorResponse = data;
+        if (responseText == '') {
+          errorResponse = JSON.stringify(
+            `Failed to fetch: ${response.status}: ${response.statusText}`,
+          );
+        } else {
+          errorResponse = data;
+        }
```
sklein12: @typpo @mldangelo is this ok?

typpo: yeah but does it need a different error message?  Maybe `Empty response: ...`

Issue #1862, opened 2024-10-11
Internal Server Error when sharing the result of eval to self-hosted server


sjsinju:
**Describe the bug**
Internal Server Error when sharing the result of eval to self-hosted server

**To Reproduce**

1. Run self-hosted server with docker [self-hosting](https://www.promptfoo.dev/docs/usage/self-hosting/) (`0.92.2`)
2. Run promptfoo for eval (npx promptfoo eval) (`0.92.2`)
3. Share the results (npx promptfoo share -y)

**Expected behavior**
Sharing should be succeeded without any errors

**Screenshots**

- error message
```
$ npx promptfoo share -y
/home/abc/node_modules/promptfoo/dist/src/share.js:41
        throw new Error(`Failed to send eval results: ${response.statusText}`);
              ^

Error: Failed to send eval results: Internal Server Error
    at sendEvalResults /home/abc/node_modules/promptfoo/dist/src/share.js:41:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async createShareableUrl (/home/abc/node_modules/promptfoo/dist/src/share.js:100:18)
    at async createPublicUrl (/home/abc/prompt-pioneer/node_modules/promptfoo/dist/src/commands/share.js:20:17)
    at async Command.<anonymous> (/home/abc/node_modules/promptfoo/dist/src/commands/share.js:50:13)

Node.js v19.9.0

```
- log
```
Failed to write eval to database SqliteError: FOREIGN KEY constraint failed
    at PreparedQuery.run (/app/node_modules/drizzle-orm/better-sqlite3/session.cjs:88:22)
    at QueryPromise.run (/app/node_modules/drizzle-orm/sqlite-core/query-builders/insert.cjs:166:28)
    at Function.<anonymous> (/app/dist/src/models/eval.js:124:22)
    at Function.sqliteTransaction (/app/node_modules/better-sqlite3/lib/methods/transaction.js:65:24)
    at BetterSQLiteSession.transaction (/app/node_modules/drizzle-orm/better-sqlite3/session.cjs:56:51)
    at BaseSQLiteDatabase.transaction (/app/node_modules/drizzle-orm/sqlite-core/db.cjs:306:25)
    at Eval.create (/app/dist/src/models/eval.js:109:18)
    at /app/dist/src/server/server.js:144:52
    at Layer.handle [as handle_request] (/app/node_modules/express/lib/router/layer.js:95:5)
[...]

PR #1891, opened 2024-10-15
fix: Delete all evals broken

sklein12:
Fix for: https://github.com/promptfoo/promptfoo/issues/1862#issuecomment-2414669854

Added tests to make sure we stop breaking delete


mldangelo (approved):

alex-pollan (approved):
Not sure I qualify as a reviewer but tests seems reasonable to me 😃

PR #1907, opened 2024-10-17
feat(redteam): math-prompt strategy

AISimplyExplained:
Math Prompt


mldangelo (approved):
Thank you @AISimplyExplained!

PR #1909, opened 2024-10-17
chore: Test sharing to cloud

sklein12:


.github/workflows/main.yml
```diff
@@ -260,3 +260,8 @@ jobs:
             echo "Error: Expected 1 entry, but got $count"
             exit 1
           fi
+
+      - name: Share to cloud
+        run: |
+          node dist/src/main.js auth login -k ${{ secrets.PROMPTFOO_STAGING_API_KEY }} -h https://api.promptfoo-staging.app
```
mldangelo, line 266: maybe we should `npm link` promptfoo although that may be more ambiguous than doing it this way.

src/globalConfig/cloud.ts
```diff
@@ -69,7 +69,10 @@ export class CloudConfig {
     };
   }
 
-  async validateAndSetApiToken(token: string, apiHost: string): Promise<void> {
+  async validateAndSetApiToken(
+    token: string,
+    apiHost: string,
+  ): Promise<{ user: any; organization: any; app: any }> {
```
mldangelo: can we type this better?

mldangelo (approved):
Thanks @sklein12!

.github/workflows/main.yml
sklein12: yea, i just wanted to ensure we were running the right files

src/globalConfig/cloud.ts
sklein12: done!

PR #1925, opened 2024-10-17
fix(providers): improve Gemini format coercion and add tests

mldangelo:
handles case where format does not need to be coerced

### DIFF
--- a/src/providers/vertex.ts
+++ b/src/providers/vertex.ts
@@ -184,12 +184,15 @@ export class VertexChatProvider extends VertexGenericProvider {
 
   async callGeminiApi(prompt: string, context?: CallApiContextParams): Promise<ProviderResponse> {
     // https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini#gemini-pro
-    let contents: any = parseChatPrompt(prompt, {
-      role: 'user',
-      parts: {
-        text: prompt,
+    let contents: GeminiFormat | { role: string; parts: { text: string } } = parseChatPrompt(
+      prompt,
+      {
+        role: 'user',
+        parts: {
+          text: prompt,
+        },
       },
-    });
+    );
     const { contents: updatedContents, coerced }: { contents: GeminiFormat; coerced: boolean } =
       maybeCoerceToGeminiFormat(contents);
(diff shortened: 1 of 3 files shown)


src/providers/vertexUtil.ts
```diff
@@ -92,10 +93,31 @@ export interface Palm2ApiResponse {
   ];
 }
 
-export function maybeCoerceToGeminiFormat(contents: any) {
+const GeminiFormatSchema = z.object({
```
typpo: Isn't `contents` potentially a list and `parts` too?
https://ai.google.dev/gemini-api/docs/quickstart?lang=rest

src/providers/vertex.ts
```diff
@@ -184,13 +184,14 @@ export class VertexChatProvider extends VertexGenericProvider {
 
   async callGeminiApi(prompt: string, context?: CallApiContextParams): Promise<ProviderResponse> {
     // https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini#gemini-pro
-    let contents = parseChatPrompt(prompt, {
+    let contents: any = parseChatPrompt(prompt, {
```
typpo: parseChatPrompt takes a type `parseChatPrompt<{...}>`

probably should be a GeminiFormat in there

src/providers/vertexUtil.ts
```diff
@@ -91,27 +93,68 @@ export interface Palm2ApiResponse {
...
+      parts: [{ text: item.content }],
+    }));
     coerced = true;
+  } else if (typeof contents === 'object' && 'parts' in contents) {
+    // This might be a single content object
+    coercedContents = [contents as z.infer<typeof ContentSchema>];
+    coerced = true;
+  } else {
+    logger.warn(`Unknown format for Gemini: ${JSON.stringify(contents)}`);
+    throw new Error('Unable to coerce to Gemini format');
```
typpo: let's just return `{contents, false}`?

that way when google changes the api for the 134902390423rd time, we don't break

typpo (commented):
did you try this with `redteam init`?  happy to try myself if you haven't

PR #1938, opened 2024-10-19
chore: Use Provider Label as Unique ID for redteam targets

sklein12:
Specifically for redteams, re-using label so we can identify targets regardless of provider name or config. This allows users to change their provider or config but treat them as the same target. Eg: You have an agent that used to be a Python function but is not accessed via an API endpoint, it's the same target just a different way of accessing it. Or you want to change the temp on a model.

image


sklein12:
@mldangelo the next button moves after the errors go away but we can fix that later

PR #1940, opened 2024-10-20
chore(webui): optionally record anonymous telemetry

mldangelo:
- Add TelemetryTracker component to App.tsx for page view tracking
- Create useTelemetry hook with VITE_PROMPTFOO_DISABLE_TELEMETRY check

I'll instrument events in the redteam setup if this looks good in the next PR


src/app/src/hooks/useTelemetry.tsx
```diff
@@ -0,0 +1,31 @@
+import { useCallback } from 'react';
+import { callApi } from '@app/utils/api';
+import type { EventProperties, TelemetryEventTypes } from '@promptfoo/telemetry';
+
+export function useTelemetry() {
+  const recordEvent = useCallback(
+    async (eventName: TelemetryEventTypes, properties: EventProperties) => {
+      if (!['0', 'false'].includes(import.meta.env.VITE_PROMPTFOO_DISABLE_TELEMETRY)) {
```
typpo, line 8: Is this additional envar necessary?  The server side already won't record telemetry if `PROMPTFOO_DSIABLE_TELEMETRY` is set, so this new envar is just cosmetic.

mldangelo: Good question. Not necessary since it's disabled in the server. This disables the xhr from appearing in the network tab. I thought that people might be confused / unhappy if they set `PROMPTFOO_DISABLE_TELEMETRY` but still see a post to `/api/telemetry` with the response `{ success: true }`

typpo:
I think it's ok. Double envars is worse IMO

PR #1962, opened 2024-10-21
chore: Add redteam descriptions and display names

sklein12:


mldangelo (approved):
Thank you for doing this!

PR #1980, opened 2024-10-23
feat: add adaline gateway provider

adaline-akshay:
### Notes:
* Update `CallApiContextParams` to include optional gateway object
  * Uses a single gateway object created for each eval run
  * Can handle multiple requests concurrently when running with concurrency > 1
  * Supports PromptFoo like retry, timeout, header based delay for 429 error, exponential backoff for 5XX errors, proxy-agent integration, and caching out of the box
  * Only used by when gateway provider is invoked, not used in any other provider
* Created a gateway cache plugin
  * Light wrapper over PromptFoo's existing cache implementation
  * Allows access for gateway to read / write PromptFoo's cache implementation
* Added `gateway` provider
  * Supports prompt, config, test, assertion types in Gateway or OpenAI format
  * Supports calling 9 different providers using the same prompt, config, test, assertion types in your config files -- including tool calling, structured outputs, vision cases
  * Supports 4 embedding providers for similarity tests
  * Uses base methods like `getApiKey`, `getBaseUrl` from native PromptFoo providers' implementation to allow users to still retain their env variables settings while using Gateway
* Updated OpenAI provider to isolate HTTP request body construction from `callApi` method, new method used by Gateway to accept config files in OpenAI format and then convert them into Gateway types.
* Added provider documentation 
* Added extensive examples showing Gateway's capabilities
* Previous PR: 
    * https://github.com/promptfoo/promptfoo/pull/1874


src/cache.gateway.ts
```diff
@@ -0,0 +1,27 @@
+
+import type { Cache as GatewayCache } from '@adaline/gateway';
```
typpo: Maybe we can tuck this into `providers` subdirectory as it is a provider-specific cache?

src/evaluator.ts
```diff
@@ -125,6 +130,19 @@ class Evaluator {
     };
     this.conversations = {};
     this.registers = {};
+    this.gateway = new Gateway({
```
typpo: If possible, I would like to avoid instantiating a Gateway in evaluator.  This is a hot path that every eval runs through, so provider-specific logic is not ideal.  Could we instantiate in the Gateway provider?

src/providers/groq.ts
```diff
@@ -56,6 +57,10 @@ export class GroqProvider implements ApiProvider {
     return this.modelName;
   }
 
+  getApiKey(): string | undefined {
```
typpo, line 60: Thanks for cleaning up this implementation!

typpo (commented):
This is looking pretty good!  Excited for this change.  A couple of small comments

src/cache.gateway.ts
adaline-akshay: For sure, will fix in next commit.

src/evaluator.ts
adaline-akshay: Good suggestion, this style of instantiating Gateway is an artifact of the previous implementation where each provider needed to access the same Gateway object in order to leverage it's concurrency, retry behavior, etc. Will fix in next commit.

typpo:
Awesome change!  Thanks so much @adaline-akshay

Issue #1970, opened 2024-10-22
Better support for rate limited calls


panozzaj:
**Is your feature request related to a problem? Please describe.**

We are using OpenAI's GPT-4o model for evaluation, and are often getting rate limited. This manifests as the run for a prompt returning "API call error: TypeError: Cannot read properties of undefined (reading 'message')". Then I re-run the same config, and thanks to caching of the not-failed responses, it usually gets farther along.

**Describe the solution you'd like**

I think my ideal solution would be to wait when rate-limited and try again after the rate limit has expired, and to retry queries that fail in this way (perhaps up to a certain amount.)

It's also possible there are other intermittent errors (503? 5xx?) that we could handle in a similar way. But this is the most painful one for us at the moment.

**Describe alternatives you've considered**

Our workaround is to just re-run it manually until there are no API-related errors, but this seems less than desirable. It's possible we could parse through the recent runs to detect this and re-run with delay, but that seems like something the framework should handle. I've also turned down the concurrency and used the delay parameter, but this just makes it happen less often at the cost of potentially more delay than needed.

**Additional context**

Happy to try to look into this, especially if you have some pointers on where we could do this.

panozzaj:
Based on searching the codebase for that error string, it looks like the code for my case is around:
 https://github.com/promptfoo/promptfoo/blob/5e500625e3da15478e9f2838323a7d832bb23958/src/providers/openai.ts#L562

(although there are other providers where this would likely be useful as well)

panozzaj:
I also see some tests for rate limiting at: https://github.com/promptfoo/promptfoo/blob/5e500625e3da15478e9f2838323a7d832bb23958/test/rateLimit.test.ts

typpo:
Hey @panozzaj, thanks for the pointers. I was in that area of the codebase anyway so I made some improvements in #1981.  Hopefully that helps

PR #1981, opened 2024-10-23
fix: better openai rate limit handling

typpo:
- Add handling for OpenAI specific headers
- Pass status back through `fetchWithCache` so it can be handled in provider

Related to #1970


panozzaj:
Not a maintainer but this seems to line up with the research I did so I believe it would fix the issue. Nice work!

PR #1982, opened 2024-10-23
chore: send version to remote endpoint

typpo:


src/providers/promptfoo.ts
```diff
@@ -1,3 +1,4 @@
+import { version } from '../../package.json';
```
mldangelo: nit, prefer VERSION from constants

PR #1984, opened 2024-10-24
refactor(tests): reorganize test files into subdirectories

mldangelo:
- Move assertion tests into test/assertions/
- Move prompt tests into test/prompts/processors/
- Move util tests into test/util/config/


typpo:
Thanks this is much nicer

PR #1990, opened 2024-10-24
feat(schema): add yaml schema validation to config files

mldangelo:
- Add yaml-language-server schema directive to all YAML config files
- Update config management to automatically add schema on write
- Make 'providers' field optional in config schema
- Add support for 'targets' as an alias for 'providers'
- Update tests to handle new schema validation


src/redteam/commands/init.ts
```diff
@@ -29,8 +29,9 @@ import {
 import { doGenerateRedteam } from './generate';
 
 const REDTEAM_CONFIG_TEMPLATE = `# Red teaming configuration
-# Docs: https://promptfoo.dev/docs/red-team/configuration
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
```
typpo, line 32: schema doesn't include targets and some other redteam specific transformations.

mldangelo: Please re-check the PR! It does include targets. Do you have any examples of this breaking?

PR #1991, opened 2024-10-24
fix(openai): refusals are not failures

typpo:
Let's just pass through the response so the end user can decide what to pass/fail


mldangelo (approved):
Thank you @typpo

PR #1994, opened 2024-10-24
fix: Http Provider parser

sklein12:
This should fix our http providing parser:

| **Headers**                  | **Body**                   | **Result**                                                                                          |
|------------------------------|----------------------------|------------------------------------------------------------------------------------------------------|
| Not set                      | Object or Array            | Headers will be set to `application/json`.                                                           |
| Not set                      | String                     | Headers will be set to `text/plain`.                                                                 |
| `application/json`           | String                     | **Warning in Console**: `Content-Type is application/json, but body is a string. This is likely to cause unexpected results. It should be an object or array.` |
| Anything that is not `application/json`       | Object or Array            | **Raises Error**: `Content-Type is not application/json, but body is an object or array. The body must be a string if the Content-Type is not application/json.` |

### DIFF
--- a/site/docs/providers/http.md
+++ b/site/docs/providers/http.md
@@ -38,9 +38,7 @@ tests:
       language: 'French'
 ```
 
-If not specified, HTTP POST with content-type application/json is assumed if the body is an object or array.
-
-`body` can be a string or a JSON object. If the body is a string, it will be sent as a raw request body and you may have to specify a `Content-Type` header. If the body is an object, then content type is automatically set to `application/json`.
+`body` can be a string or JSON object. If the body is a string, the `Content-type` header defaults to `text/plain` unless specified otherwise. If the body is an object, then content type is automatically set to `application/json`.
 
 ### JSON Example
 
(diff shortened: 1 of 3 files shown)


typpo:
Some thoughts:
- `text/plain` is an unusual content type for `POST` and probably not a good default.  I imagine `application/x-www-form-urlencoded` is more common for strings
- In `application/json`/`string` combo, I could see ppl just setting a JSON string rather than YAML markup?  Since they are explicitly setting both values, maybe we just let it happen?

sklein12:
> Some thoughts:
> 
> * `text/plain` is an unusual content type for `POST` and probably not a good default.  I imagine `application/x-www-form-urlencoded` is more common for strings
> * In `application/json`/`string` combo, I could see ppl just setting a JSON string rather than YAML markup?  Since they are explicitly setting both values, maybe we just let it happen?

Sounds good on the default

Previously,  if you had a string it converted it into an array of characters because the first thing it did is call `Object.entries(` on the body. I figured that's not what people would expect. I could either:
1. It just send a string
2. Try and parse it and if it's not an object just send a string
3. Keep the same behavior

PR #1995, opened 2024-10-24
feat: package import support improvements

dkundel:
Previously the `package:` provider was only looking at `${basePath}/node_modules/package.json` but Node.js module resolution logic is actually much more complicated. `require.resolve` handles this lookup much better as it contains all of the logic. There is an equivalent for ESM called `import.meta.resolve` but it's still not stable so as a result I opted for using `createRequire` to create a require instance specifically for that `basePath` and then leverage `require.resolve` from that instance. This fix was needed especially when working with a monorepo with multiple packages where the package you want to use might not be at the same level as your config.

Additionally I ported that same functionality to both variable definition and JavaScript assertions. With this change both of these should be valid:

```yml
defaultVars:
  example: package:@dkundel/demo:getSessionVariable

tests:
  - vars:
      prompt: Ahoy!
    assert:
      - type: javascript
        value: package:@dkundel/demo:isNauticalGreeting
```

### DIFF
--- a/src/assertions/index.ts
+++ b/src/assertions/index.ts
@@ -6,7 +6,7 @@ import { XMLParser } from 'fast-xml-parser';
 import { distance as levenshtein } from 'fastest-levenshtein';
 import fs from 'fs';
 import * as rouge from 'js-rouge';
-import yaml, { load } from 'js-yaml';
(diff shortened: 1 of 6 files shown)


typpo (approved):
Very nice change!

PR #2014, opened 2024-10-26
feat: simulated user provider

typpo:
just for fun


src/providers.ts
```diff
@@ -66,6 +66,7 @@ import {
   ReplicateProvider,
 } from './providers/replicate';
 import { ScriptCompletionProvider } from './providers/scriptCompletion';
+import { SimulatedUser } from './providers/simulatedUser';
```
mldangelo, line 69:
```suggestion
import { PromptfooSimulatedUserProvider } from './providers/promptfoo';
```

src/providers.ts
```diff
@@ -473,6 +474,8 @@ export async function loadApiProvider(
     ret = new RedteamCrescendoProvider(providerOptions.config);
   } else if (providerPath === 'promptfoo:manual-input') {
     ret = new ManualInputProvider(providerOptions);
+  } else if (providerPath === 'promptfoo:simulated-user') {
+    ret = new SimulatedUser(providerOptions);
```
mldangelo, line 478:
```suggestion
    ret = new PromptfooSimulatedUserProvider(providerOptions);
```

PR #2016, opened 2024-10-27
chore(redteam): handle refusals from gpt-4o structured JSON outputs

mldangelo:
- Add handleRefusal method to RedteamGraderBase to standardize refusal handling
- Update all redteam graders to support isRefusal parameter

### DIFF
--- a/src/redteam/plugins/base.ts
+++ b/src/redteam/plugins/base.ts
@@ -230,16 +230,10 @@ export abstract class RedteamGraderBase {
   }): ResultSuggestion[] {
     return [];
   }
-
-  async getResult(
-    prompt: string,
-    llmOutput: string,
-    test: AtomicTestCase,
-    provider: ApiProvider | undefined,
-    renderedValue: AssertionValue | undefined,
-    isRefusal: boolean | undefined = false,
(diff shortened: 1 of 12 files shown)


src/redteam/plugins/base.ts
```diff
@@ -230,15 +230,38 @@ export abstract class RedteamGraderBase {
   }): ResultSuggestion[] {
     return [];
   }
+  protected handleRefusal(
+    provider: ApiProvider | undefined,
+    isRefusal: boolean | undefined,
+  ): { grade: GradingResult; rubric: string } | undefined {
+    if (isRefusal) {
+      return {
```
typpo, line 238: Because refusals pass by default, aren't we ok without this logic? 

Would it also be nice to see the actual refusal output, which is the current behavior?

mldangelo: > Because refusals pass by default, aren't we ok without this logic?

How do you suggest refactoring it? 

> Would it also be nice to see the actual refusal output, which is the current behavior?

This is what a refusal looks like in the webui with the current implementation. 

Screenshot 2024-10-27 at 12 38 04 AM

What exactly are you proposing? That I write the llmOutput to value?

PR #2020, opened 2024-10-27
docs(provider): fix dalle-3 provider name

zerfl:
The Dall-E3 provider is referred to as `openai:image:dalle-3` when instead it should be `openai:image:dall-e-3` in the docs. This PR corrects the name.


mldangelo (approved):
Thank you!

PR #2049, opened 2024-11-01
fix: OOM on large evals

sklein12:
1. Not keeping results in memory anymore
2. Tracking stats on prompts instead of needing to re-calculate them every time by calling `toEvaluateResult`
3. Adding a JSONL export option which writes line by line during eval
4. Update documentation

I ran a 21,420 test case eval which maxed at 350mb of ram going -j 20. Wrote to a JSONL file, resulting file was > 3gb

Before this would peak to over 4gb of ram and hit an OOM exception.

### DIFF
--- a/src/commands/eval.ts
+++ b/src/commands/eval.ts
@@ -9,7 +9,7 @@ import { z } from 'zod';
 import { fromError } from 'zod-validation-error';
 import { disableCache } from '../cache';
 import cliState from '../cliState';
-import { getEnvFloat, getEnvInt, getEnvBool } from '../envars';
+import { getEnvFloat, getEnvInt } from '../envars';
 import { DEFAULT_MAX_CONCURRENCY, evaluate } from '../evaluator';
 import logger, { getLogLevel, setLogLevel } from '../logger';
 import { runDbMigrations } from '../migrate';
@@ -218,7 +218,7 @@ export async function doEval(
     const totalTests = successes + failures;
     const passRate = (successes / totalTests) * 100;
 
-    if (cmdObj.table && getLogLevel() !== 'debug') {
+    if (cmdObj.table && getLogLevel() !== 'debug' && totalTests < 500) {
       const table = await evalRecord.getTable();
       // Output CLI table
       const outputTable = generateTable(table);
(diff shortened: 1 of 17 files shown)


src/commands/eval.ts
```diff
@@ -194,9 +194,32 @@ export async function doEval(
...
+      if (prompt.metrics?.testFailCount) {
+        failures += prompt.metrics.testFailCount;
+      }
+      tokenUsage.total += prompt.metrics?.tokenUsage?.total || 0;
+      tokenUsage.prompt += prompt.metrics?.tokenUsage?.prompt || 0;
+      tokenUsage.completion += prompt.metrics?.tokenUsage?.completion || 0;
+      tokenUsage.cached += prompt.metrics?.tokenUsage?.cached || 0;
+    }
+    const totalTests = successes + failures;
+    const passRate = (successes / totalTests) * 100;
```
sklein12, line 220: We can't just get the table which has all the results so we'll sum them up off of the prompts

src/evaluator.ts
```diff
@@ -614,6 +616,11 @@ class Evaluator {
       if (evalStep.test.assert?.some((a) => a.type === 'select-best')) {
         rowsWithSelectBestAssertion.add(row.testIdx);
       }
+      for (const assert of evalStep.test.assert || []) {
+        if (assert.type) {
+          assertionTypes.add(assert.type);
+        }
+      }
```
sklein12, line 631: We used to loop through all the test cases at the end to get the assertions. We'll track as we go now.

src/evaluator.ts
```diff
@@ -630,6 +637,19 @@ class Evaluator {
...
+        ? outputPath.filter((p) => p.endsWith('.jsonl'))
+        : outputPath?.endsWith('.jsonl')
+          ? [outputPath]
+          : [];
+      if (jsonlFiles.length) {
+        const jsonl = JSON.stringify(row);
+        for (const path of jsonlFiles) {
+          fs.appendFileSync(path, jsonl + '\n');
+        }
+      }
```
sklein12: We're going to write to our jsonl file as we go now

sklein12:
@coderabbitai review this please

PR #2059, opened 2024-11-04
chore: refactor plugin validation

typpo:
- reduces constructor boilerplate in plugins with configs (config is now an optional static schema on the plugin class, and is checked automatically)
- uses zod schema for single source of truth for plugin configs
- reintroduces validatePlugins so that we fail before generation starts


mldangelo (approved):

mldangelo:
Closing as stale

PR #2060, opened 2024-11-04
fix(webui): preserve target label when switching target types

mldangelo:
Maintains the selected target label when switching between different target types in the redteam generation web UI setup flow. This fixes a validation error.


mldangelo:
@coderabbitai review

PR #2081, opened 2024-11-06
feat: BLEU score

benbuzz790:
Add bleu score and associated tests.


mldangelo (commented):
Thanks so much @benbuzz790! Would you mind 

1. moving the bleu score logic into a file in the assertions/ folder? You could call it bleu.ts ? 
2. defining the assertion so that we can actually call bleu from runAssertion
3. taking a stab at updating the assertion docs?

benbuzz790:
Hi Michael. I believe I've addressed your 3 comments. LMK if you need anything else.

Issue #2080, opened 2024-11-05
Support AWS Profiles and/or Session Tokens with AWS providers


rvernica:
**Is your feature request related to a problem? Please describe.**

Currently, the only AWS authentication supported is via access key and secret access key. More complex setups use SSO profiles or require a Session Token. Neither is supported by promptfoo.

**Describe the solution you'd like**

Support AWS Profiles and/or AWS Session Tokens for authentication.

**Describe alternatives you've considered**

Do not use promptfoo or use a [Custom Python](https://www.promptfoo.dev/docs/providers/python/) provider that does its own AWS authentication.

typpo:
Released in 0.96.0!

PR #2085, opened 2024-11-06
chore: Support AWS sessionToken and profile for authentication

rvernica:
Fix for #2080 

* Add sessionToken and profile as Bedrock configuration options
* Add sessionToken to credentials dictionary if key pair is present
* If profile is present, use fromSSO
* getCredentials can return an AwsCredentialIdentityProvider
* import fromSSO from aws-sdk/credential-provider-sso


mldangelo:
Thank you @rvernica!

PR #2095, opened 2024-11-07
chore: goat-blog demo component usability improvements

AISimplyExplained:
- suggestions cards
- reset changes


mldangelo (commented):
This is great! thank you. Can you take one more pass at the dark mode styles? 

Screenshot 2024-11-07 at 12 23 11 PM

The book a demo button below could also use some CSS. 


I think we have some caching behavior serverside for the new examples you added where we may want to either disable the cache or add an artificial delay. 

Thank you @AISimplyExplained

mldangelo (approved):
Thank you @AISimplyExplained

PR #2102, opened 2024-11-08
chore: allow goat to be used stateless or not

sklein12:
By default Goat is "stateless" so it'll send the entire convo to the provider in API format.


mldangelo (approved):
thank you!

PR #2103, opened 2024-11-08
Use "provider" key in python prompt function

lkopocinski:
In the Python code responsible for prompt generation, the key should be the word "provider" instead of "providers."


mldangelo (approved):
Good find! Thank you @lkopocinski!

Issue #2046, opened 2024-10-31
Respect max text length table setting when rendering cells as Markdown


panozzaj:
**Is your feature request related to a problem? Please describe.**

We are giving some values as Markdown links to provide some handy links to our app. This could be like:

```
[123](http://localhost:3000/link/to/view/123)
```

However, when we do this, other table inputs do not respect the table "Max text length" value. We have some pretty big cells, so the browser tab gets bogged down and is somewhat unresponsive.

**Describe the solution you'd like**

I think we could use the truncated text as well here:

```tsx
// src/app/src/pages/eval/components/ResultsTable.tsx
400:                    {renderMarkdown ? (
401:                      <ReactMarkdown remarkPlugins={[remarkGfm]}>{value}</ReactMarkdown>  // << s/value/something else/
402-                    ) : (
403:                      <TruncatedText text={value} maxLength={maxTextLength} />
404-                    )}
405-                  </div>
```

**Describe alternatives you've considered**
N/A

**Additional context**

If you want me to take a crack at this, I can!

mldangelo:
Hi @panozzaj, thanks for opening this issue and for the clear description! Your proposed solution sounds great, and we’d definitely welcome a PR on this. Feel free to go ahead, and if you have any questions or need guidance, don’t hesitate to reach out. Thanks again—looking forward to your contribution! 🚀

mldangelo:
Thanks so much for your help in shipping this @panozzaj! Please feel free to reach out if you have any other feature requests, product ideas, or feedback.

PR #2109, opened 2024-11-08
feat: Respect max text length in Markdown cells

panozzaj:
Implements #2046 

I tested this by running an eval that had long cells and ensuring that when I checked the "Render model outputs as Markdown" box that it respected that limit. This enabled me to see Markdown links for my evals that had large inputs.

Open to any feedback, and thanks for the quick response on the original issue!


mldangelo (commented):
Code looks good! I'll test it out and get it deployed. Thanks @panozzaj!

PR #2114, opened 2024-11-09
fix: validate json, yaml, js configs on load

typpo:
Run validation on configs when we read configs, instead of casting


typpo:
@mldangelo  I didn't throw an error on validation in case there's a reason we weren't doing that in the first place.  It just outputs an error message.  Would appreciate a look at this.

mldangelo (approved):
This looks good. Since we handle parsing and validation in multiple places, I’m not confident our schemas fully cover every supported config format. For now, I think we 
 should continue logging a warning rather than throwing an error. We can revisit this in a few months if no one complains.

PR #2116, opened 2024-11-10
refactor(assertions): split assertions into separate modules

mldangelo:
Splits large assertions/index.ts file into separate modules for each assertion type


typpo:
I think there's an opportunity to improve this further.  Instead of having 30 different `if (x) { handler }` we can just map from assertion type to handler, where handlers have a standard interface.

e.g.
```ts
type AssertionParams = {
  assertion: Assertion;
  renderedValue: AssertionValue | undefined;
  output: string;
  prompt?: string;
  test: TestCase;
  providerResponse: ProviderResponse;
  inverse?: boolean;
  context?: AssertionValueFunctionContext;
  valueFromScript?: any;
  latencyMs?: number;
  // ...?
};

const assertionHandlers: Record<string, (params: AssertionParams) => Promise<GradingResult>> = {
  equals: handleEquals,
  'is-json': handleIsJson,
  similar: handleSimilar,
  'context-faithfulness': handleContextFaithfulness,
  'llm-rubric': handleLlmRubric,
  webhook: handleWebhook,
  // etc...
};

export async function runAssertion(params: {
  assertion: Assertion;
  providerResponse: ProviderResponse;
  test: TestCase;
  prompt?: string;
  context?: AssertionValueFunctionContext;
}): Promise<GradingResult> {
  // ...
  // Get the appropriate handler
  const handler = assertionHandlers[baseType];
  if (!handler) {
    throw new Error(`Unknown assertion type: ${assertion.type}`);
  }

  // Run the handler with all available params
  const result = handler({
    // ...
  });
  // ...
}
```

mldangelo:
> I think there's an opportunity to improve this further.

I was planning to do this later but can do it now

src/assertions/similar.ts
```diff
@@ -1,18 +1,20 @@
...
-  outputString: string,
+  output: string | object,
   inverse: boolean,
   test: AtomicTestCase,
 ): Promise<GradingResult> => {
   invariant(
     typeof renderedValue === 'string' || Array.isArray(renderedValue),
     'Similarity assertion type must have a string or array of strings value',
   );
+  const outputString = coerceString(output);
```
typpo: instead of having to remember to call coerceString everywhere, it might makes sense to pass `outputString` into the handlers.  the handler signature is large enough that a dict {} would be less unwieldy

mldangelo:
@coderabbitai review all

mldangelo:
@coderabbitai review all

PR #2126, opened 2024-11-11
fix: Config for all strategies

sklein12:
The config for all strategies weren't being persisted through the generate command.


typpo (approved):
is this a problem in other strategies?

Issue #2123, opened 2024-11-11
Broken Anthropic System Message


kaulsh:
**Describe the bug**

I use JS functions to return prompts. Here's how a test file looks:

```
description: "Audit"

providers:
  - id: "anthropic:messages:claude-3-5-haiku-20241022"
    config:
      temperature: 0
      max_tokens: 4096

prompts:
  - ../build/prompts/index.js:audit_criteria

tests: GOOGLE_SHEETS_URL
```

In the latest version (v0.96.0) I get this error on prompts that use the system message:
```
API call error: messages: Unexpected role "system". The Messages API accepts a top-level `system` parameter, not "system" as an input message role., status 400, type invalid_request_error
```

After reviewing some code, I think this change in `src/providers/anthropic.ts` broke it:
https://github.com/promptfoo/promptfoo/commit/3c29b39ff26c6a9e77c1a0da344960a81fc8b59c#diff-afce651d866a1cd4afb1936e9e431f2aab127727a60530b1774dac7348961c65

~~It also seems as though anthropic's prefill messages aren't working anymore. I could not find any related code to such.~~ This could just be a model issue. Prefill works with other models. Can ignore.

aandis:
+1

kaulsh:
@mldangelo No problem and thanks for the quick patch. Everything works as expected now :+1:

PR #2128, opened 2024-11-12
fix(providers): handle system messages correctly in anthropic parseMessages

mldangelo:
The Messages API expects system messages to be handled separately from regular messages. This change:
- Extracts system messages from the input array
- Formats system content correctly as TextBlockParams
- Filters system messages out of extractedMessages
- Updates tests to reflect the new behavior

Related to #2123

### DIFF
--- a/src/providers/anthropic.ts
+++ b/src/providers/anthropic.ts
@@ -124,11 +124,21 @@ export function parseMessages(messages: string): {
   try {
     const parsed = JSON.parse(messages);
     if (Array.isArray(parsed)) {
+      const systemMessage = parsed.find((msg) => msg.role === 'system');
       return {
-        extractedMessages: parsed.map((msg) => ({
-          role: msg.role,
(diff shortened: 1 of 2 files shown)


mldangelo:
```yaml
description: "Audit"

providers:
  - id: "anthropic:messages:claude-3-5-haiku-20241022"
    config:
      temperature: 0
      max_tokens: 4096

prompts:
  - file://2123.js:audit_criteria
  - '{{query}}'

tests:
  - vars:
      query: 'what is 2 + 2?'
    assert:
      - type: contains
        value: 4
```


and 

```js
module.exports = {
  audit_criteria: async ({ vars, provider }) => {
    return [
        {
            role: 'system',
            content: `You're an angry pirate named ${provider.label || provider.id}. Be concise and stay in character.`,
        },
        {
            role: 'user',
            content: `${vars.query}`,
        },
    ]
  },
};
```

to test

typpo (approved):

Issue #2125, opened 2024-11-11
Erroneous warning that `$ref` is invalid in the providers section of promptfooconfig


sobjornstad:
**Describe the bug**
In 0.95 and earlier, `$ref` worked great to retrieve the model(s) you wanted to use, e.g.:

```yaml
providers:
  - anthropic:messages:claude-3-5-sonnet-20240620
  - $ref: promptfoo/defaultParams.yaml#/model
  - openai:gpt-4o-mini
```

Now in 0.96, I get a yellow warning when I run an eval:

```
Invalid configuration file promptfoo/magic_copy_paste_text_to_cards_advanced/all_card_types/promptfooconfig.yaml:
Validation error: Unrecognized key(s) in object: '$ref' at "providers[1]"
```

Despite this “validation error,” the eval still runs as expected, with the correct value retrieved from the ref.

**To Reproduce**

Create a promptfooconfig.yaml, e.g.:

```yaml
description: concept_panel_define
prompts:
  - test {{text}}
providers:
  - $ref: promptfoo/defaultParams.yaml#/model
temperature: 1

tests:
  - vars:
      text: bla bla bla
```

Create the `defaultParams.yaml` file referenced, with some value for `model` that's a valid model, e.g.:

```yaml
model: openai:gpt-4o-mini
```

Run `promptfoo eval` on the config.

**Expected behavior**
No warning appears, since the syntax is valid and the eval runs normally.

**Screenshots**

**System information (please complete the following information):**

- Promptfoo version: 0.96

typpo:
Thanks for flagging.  Will get a fix in.  This shouldn't affect the actual run of the eval 👍

mldangelo:
Hi @sobjornstad, thanks for filing this issue! We addressed this in the latest release.  Please update to 0.96.1 and let us know if you encounter any further issues

PR #2129, opened 2024-11-12
fix: validate config after dereferencing

mldangelo:
The schema validation was happening before  resolution, causing false validation errors for valid  usage in providers configuration.

This change:
- Moves schema validation after dereferencing
- Adds tests to verify  resolution in YAML configs
- Ensures validation errors are properly reported

Related to #2125

### DIFF
--- a/test/util/config/load.test.ts
+++ b/test/util/config/load.test.ts
@@ -1,9 +1,12 @@
+import $RefParser from '@apidevtools/json-schema-ref-parser';
+import dedent from 'dedent';
 import * as fs from 'fs';
 import { globSync } from 'glob';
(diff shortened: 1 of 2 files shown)


test/util/config/load.test.ts
```diff
@@ -936,4 +939,139 @@ describe('readConfig', () => {
...
+        }
+
+        return Object.fromEntries(
+          Object.entries(obj).map(([key, value]) => [key, resolveRefs(value)]),
+        );
+      };
+
+      return resolveRefs(schema);
+    });
+
```
typpo: it doesn't seem right to have a full dereferencing implementation in a test :o

PR #2146, opened 2024-11-15
fix: handle null test cases in strategy generation

mldangelo:
- Add null/undefined checks for strategy test cases
- Update strategy results to accurately reflect valid test count
- Prevent 'Cannot read properties of undefined (reading metadata)' error


typpo:
any idea where these nulls come from? 👿

Issue #2099, opened 2024-11-07
Support RBAC for Azure OpenAI (Key based authentication is disabled for this resource)


mwiemer-microsoft:
**Is your feature request related to a problem? Please describe.**
At Microsoft, we are moving away from key-based authentication and towards RBAC. As such, we cannot use AZURE_OPENAI_API_KEY on our resources.

When we try:

`npx promptfoo eval --verbose` on the [`azure-openai`](https://github.com/promptfoo/promptfoo/tree/0.95.0/examples/azure-openai) example with promptfoo@0.95.0

we get:

"Azure OpenAI API response: {"error":{"code":"AuthenticationTypeDisabled","message":"Key based authentication is disabled for this resource."}}" 

**Describe the solution you'd like**
If there was some way to use [`@azure/msal-node`](https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-node) or similar to authenticate the principal (user or pipeline agent) in the relevant Azure tenant, then use their Azure credentials to connect to the resource, that'd be great.

**Describe alternatives you've considered**
My team is actively working on wrapping [`node-package-typescript`](https://github.com/promptfoo/promptfoo/tree/0.95.0/examples/node-package-typescript) with RBAC auth logic via `@azure/msal-node` or something similar to workaround this issue.

**Additional context**
We aren't experts at this--there may be a better way, and we'll update with what we find :)

Related: #1074

mwiemer-microsoft:
Ref [msal-node sample](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/extensions/samples/msal-node-extensions/index.js)

typpo:
Thanks for bringing this up.  Definitely seems like something we should bake into our Azure provider https://github.com/promptfoo/promptfoo/blob/main/src/providers/azureopenai.ts

typpo:
@mwiemer-microsoft what's your ideal flow here?  Are you suggesting that Promptfoo provides a login endpoint?

mwiemer-microsoft:
@typpo I've been trying to figure that out. For reference, our C# code can use Visual Studio credentials, and our deployed web apps can use browser credentials, but of course neither of those are possible with this isolated process. Ideally, the promptfoo process is able to fetch some environment variables, including a login endpoint, silently make a "get token" request, and use the token to make requests against the LLM directly.

[...]

PR #2149, opened 2024-11-15
feat(azure): adding AzureCliCredential as a fallback authentication option

ali42:
Uses AzureCliCredential to add a Bearer token for the Azure OpenAI api calls. 
This is useful for running promptfoo locally without keys (as mentioned in https://github.com/promptfoo/promptfoo/issues/2099) 

TODO:
- [x] Update auth error messages so they guide the user better on their auth options
- [x] Update the Docs

Additional changes included in this PR:
- Client Secret Credential Token was not passed as an Authorization Bearer token in the http calls so fixes that too. 
- `_cachedApiKey` approach was not sufficient for the token as with each call this would mean getting a new token. So instead we find out the auth headers (api or token) at the constructor stage and cache that instead.

Suggested Future PR Changes:
- enable token based auth for OpenAI Assistants as well


src/providers/azure.ts
```diff
@@ -228,60 +234,101 @@ export class AzureGenericProvider implements ApiProvider {
...
+      getEnvString('AZURE_API_KEY') ||
+      this.env?.AZURE_OPENAI_API_KEY ||
+      getEnvString('AZURE_OPENAI_API_KEY')
+    );
+  }
+
+  getApiKeyOrThrow(): string {
+    const apiKey = this.getApiKey();
+    if (!apiKey) {
+      throw new Error('Azure OpenAI API key must be set');
```
mwiemer-microsoft: Maybe we can link to some docs in this error message?

mwiemer-microsoft (commented):
Looks great to me! To clarify, I work for Microsoft, but not Azure, and am mostly commenting as someone else who needs this feature for work :)

typpo:
Huge thanks @ali42!  This looks great and I'll get it in the next release.

ali42:
Thank you both! Tested the 0.97.0 locally as well with azure cli auth, looks great!

PR #2152, opened 2024-11-15
fix: User provided config is processed before a promptfooconfig in eval

### LINKED ISSUE #1781: evaluateOptions not working if file name != promptfooconfig
**Describe the bug**

The `evaluateOptions` is not effective if my promptfoo config file is different from  `promptfooconfig.yaml`

**To Reproduce**

1. With a `promptfooconfig.yaml` set `evaluateOptions: { repeat: 2, cache: false }` and write a test
2. Execute `promptfoo eval` or `promptfoo eval -c promptfooconfig.yaml`
3. The test is well repeated 2 times and the cache is disabled
4. Rename `promptfooconfig.yaml` into `promptfooconfig-test.yaml`
5. Execute `promptfoo eval -c promptfooconfig-test.yaml`
6. The test is executed only once, and the cache is enabled

**Expected behavior**

I expect that the evaluateOptions are effective if my promptfoo config file is different from `promptfooconfig.yaml`

- Promptfoo version: 0.90.1

eharris128:
# What
- Support precedence of a user provided config while even when there is a `promptfooconfig.yaml` present.
- No longer default to the defaultConfig for eval flags. Instead, after parsing the config flag, set to fall back default value, or to defaultConfig (aka promptfooconfig.yaml).

## Why
- Resolves #1781 
- Promptfoo CLI [docs](https://www.promptfoo.dev/docs/usage/command-line/#promptfoo-eval) make it seem like other config files can be used alongside a promptfooconfig

## Questions / concerns
- Am I missing a entrypoint into `doEval` where `defaultConfig` has changed? Which would mean this code must persist. The code was only added a month ago, so I am wondering if there is a repro scenario available.  -> Note I tested this further, and rolled back my changes to the top of `eval.ts` - [commit](https://github.com/promptfoo/promptfoo/pull/2152/commits/8d426a9c48361c39b0648a2ad4af68637ec6d76f)

## Follow Up
- I have additional code that is not necessary for fixing this user reported issue that:
  - Modifies `loadDefaultConfig` to support a user passing config files with extensions on the end. 
- This will come as another MR.

## Note
- doEval is pretty heavy glue code and cutting it up to be able to abstract out key flows (e.g. a user editing a config tile in between runs and having the new values be used in the subsequent invocation) would be nice.


typpo:
Thanks for tackling this - definitely a crusty part of the codebase.  Will take a closer look asap!

eharris128:
If the PR feels too high risk, I am happy to derisk it a bit as well by snipping out the below section:
image

If I recall my testing correctly, the changes I introduced in this part of eval.ts were not necessary to address the initial reported user issue. I simply tried to sweep the floor in the area I was working.

eharris128:
> If the PR feels too high risk, I am happy to derisk it a bit as well by snipping out the below section: image
> 
> If I recall my testing correctly, the changes I introduced in this part of eval.ts were not necessary to address the initial reported user issue. I simply tried to sweep the floor in the area I was working.

So I performed further testing, and my implementation broke the `redteam eval` invocation when there was no config file specified, so I rolled back my changes to this part of `eval.ts`

eharris128:
Stale PR. Closing for now.

PR #2153, opened 2024-11-16
chore(providers): misc harmful completion provider enhancements 

mldangelo:
- Decrease number of retries from 4 to 2 in harmful completion provider
- Improve error handling for null outputs in generateTestsForCategory


src/redteam/plugins/harmful.ts
```diff
@@ -284,7 +284,9 @@ async function generateTestsForCategory(
         await sleep(delayMs);
       }
     }
-    return results.map((result) => createTestCase(injectVar, result.output || '', harmCategory));
+    return results
+      .filter((result) => result.output)
+      .map((result) => createTestCase(injectVar, result.output!, harmCategory));
```
typpo: no `!`s please.  add an invariant or some other check

PR #2165, opened 2024-11-18
fix(share): update eval author to logged-in user when sharing

mldangelo:
- If sharing for the first time and no email is set, write the email to the eval. 
- When sharing an eval in cloud mode, automatically update the author to match the logged-in user's email. 
- Show a warning if the author is being changed from a different email address. - let's discuss this I am not sure about the behavior pattern

### DIFF
--- a/src/share.ts
+++ b/src/share.ts
@@ -77,22 +77,7 @@ export async function createShareableUrl(
   evalRecord: Eval,
   showAuth: boolean = false,
 ): Promise<string | null> {
-  if (cloudConfig.isEnabled()) {
-    const loggedInEmail = getUserEmail();
-    invariant(loggedInEmail, 'User email is not set');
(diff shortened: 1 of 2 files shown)


src/share.ts
```diff
@@ -95,6 +98,17 @@ export async function createShareableUrl(
   if (cloudConfig.isEnabled()) {
     apiBaseUrl = cloudConfig.getApiHost();
     url = `${apiBaseUrl}/results`;
+
+    const loggedInEmail = getUserEmail();
+    invariant(loggedInEmail, 'User email is not set');
+    const evalAuthor = evalRecord.author;
+    if (evalAuthor !== loggedInEmail) {
```
typpo, line 105: is this necessary/desirable?  what if they are sharing an eval they imported or something?

PR #2200, opened 2024-11-21
fix(providers): handle more response parser failures in http provider

typpo:
and slightly improve debuggability

### DIFF
--- a/src/providers/http.ts
+++ b/src/providers/http.ts
@@ -88,7 +88,15 @@ export async function createResponseParser(
     return (data, text) => ({ output: data || text });
   }
   if (typeof parser === 'function') {
-    return (data, text) => ({ output: parser(data, text) });
+    return (data, text) => {
+      try {
+        const result = parser(data, text);
+        return { output: result };
+      } catch (err) {
+        logger.error(`Error in response parser function: ${String(err)}`);
+        throw err;
+      }
+    };
   }
   if (typeof parser === 'string' && parser.startsWith('file://')) {
     let filename = parser.slice('file://'.length);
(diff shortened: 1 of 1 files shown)


src/providers/http.ts
```diff
@@ -88,7 +88,15 @@ export async function createResponseParser(
...
   }
   if (typeof parser === 'function') {
-    return (data, text) => ({ output: parser(data, text) });
+    return (data, text) => {
+      try {
+        const result = parser(data, text);
+        return { output: result };
+      } catch (err) {
+        logger.error(`Error in response parser function: ${String(err)}`);
+        throw err;
```
sklein12, line 97: There's no change in behavior other than logging, is that the intent? This will still just raise an exception

src/providers/http.ts
```diff
@@ -361,19 +369,21 @@ export class HttpProvider implements ApiProvider {
     }
     try {
       const parsedOutput = (await this.responseParser)(parsedData, rawText);
-      ret.output = parsedOutput.output || parsedOutput;
+      ret.output = parsedOutput?.output ?? parsedOutput;
```
sklein12: this is going to return undefined if parsed output is undefined, is that what we want?

src/providers/http.ts
```diff
@@ -418,9 +428,16 @@ export class HttpProvider implements ApiProvider {
...
-    };
+    try {
+      const parsedOutput = (await this.responseParser)(parsedData, rawText);
+      return {
+        output: parsedOutput.output || parsedOutput,
+      };
+    } catch (err) {
+      logger.error(`Error parsing response: ${String(err)}. Got response: ${rawText}`);
+      return {
+        error: `Error parsing response: ${String(err)}. Got response: ${rawText}`,
```
sklein12, line 439: 👍

src/providers/http.ts
```diff
@@ -361,19 +369,21 @@ export class HttpProvider implements ApiProvider {
...
+        logger.error(
+          `Error parsing session ID: ${String(err)}. Got headers: ${safeJsonStringify(response.headers)}`,
+        );
       }
       return ret;
     } catch (err) {
-      logger.error(`Error parsing response: ${String(err)}`);
-      ret.error = `Error parsing response: ${String(err)}`;
+      logger.error(`Error parsing response: ${String(err)}. Got response: ${rawText}`);
+      ret.error = `Error parsing response: ${String(err)}. Got response: ${rawText}`;
```
sklein12, line 386: 👍  got it, they all bubble up to here

src/providers/http.ts
typpo: as you noted later, they all bubble up

src/providers/http.ts
typpo: probably not, going to revert

PR #2208, opened 2024-11-22
chore(redteam): improve competitors plugin and grading

AISimplyExplained:
- false default for mentions
-  Use: 
```yaml
 - id: competitors
      numTests: 10
      config:
        mentions: false
```
- mentions false eval: https://www.promptfoo.app/eval/eval-Ldg-2024-11-22T05:22:40
- mentions true eval: https://www.promptfoo.app/eval/eval-LtO-2024-11-22T12:45:26 ( 10% improvement)


mldangelo (approved):
Thank you!

PR #2211, opened 2024-11-22
chore(redteam): Add some troubleshooting documentation

sklein12:


site/docs/red-team/Troubleshooting/attack-generation.md
mldangelo, line 1: can you make the directory name this is in lower case?

typpo: maybe we name this directory "Guides" so we can put other stuff in it?

PR #2214, opened 2024-11-23
chore: separate errors from assert failures

typpo:
followups
- redteam reporting and ui
- eval reporting
- cloud

I think you can already implicitly tell whether a test failed due to error from a lack of gradingResults/componentResults.  So not totally sure if this is necessary, but it is definitely way cleaner to set a flag this way.


mldangelo (approved):
Tested and this works. I think it will be incredibly useful (we can start to retry only errors instead of failures for example). 

Thank you @typpo

mldangelo (approved):

PR #2216, opened 2024-11-23
fix: add first round of goat in metadata field

typpo:
@mldangelo can you confirm this is correct, the endpoint expects the initial message only as `prompt` and not in `messages`?


mldangelo (commented):
Let's discuss this tomorrow. I don't think we want to do this.

typpo:
closing because this is never actually sent to target

PR #2230, opened 2024-11-25
fix: csv bom parsing

darrelhong:
#### Context

Seems like excel generally saves csv files as UTF-8 (with BOM) and this [breaks parsing](https://stackoverflow.com/questions/66402555/first-property-inaccsesible-of-objects-parsed-by-csv-parse) of the csv with the current settings. 

#### Fix

Update the options of `csv-parse` to detect bom marks. It is [recommended](https://csv.js.org/parse/options/bom/) in the docs to enable this option though the default value is false.

**BEFORE**
Screenshot 2024-11-25 at 5 59 41 PM

**AFTER**
Screenshot 2024-11-25 at 5 59 09 PM

### DIFF
--- a/src/testCases.ts
+++ b/src/testCases.ts
@@ -80,7 +80,7 @@ export async function readStandaloneTestsFile(
     telemetry.recordAndSendOnce('feature_used', {
       feature: 'csv tests file - local',
     });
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you!

PR #2235, opened 2024-11-25
docs: guide on LangChain PromptTemplates

adelmuursepp:
This guide covers:
- Using LangChain PromtTemplate for prompts defined in the application code
- Using PromptFoo to test the prompts defined in the application and the necessary formatting configurations


adelmuursepp:
> Thanks @adelmuursepp !

Excited to see it in the docs!

PR #2239, opened 2024-11-26
chore(redteam): move harmful specialized advice plugin to unaligned provider 

mldangelo:
The specialized-advice plugin generates a lot of refusals. This moves it to the unaligned provider.


typpo (approved):
Cool change

PR #2241, opened 2024-11-26
feat(webui): select all/none in redteam setup plugins view

typpo:


mldangelo (approved):
Thank you for taking this one!

PR #2252, opened 2024-11-27
fix(providers): stringify non-string target provider responses in goat

mldangelo:
Ensures that non-string responses from target providers are properly stringified 
before being added to the message history. This maintains consistent string-based 
content in the message array.


src/redteam/providers/goat.ts
```diff
@@ -96,10 +96,12 @@ export default class GoatProvider implements ApiProvider {
           ? JSON.stringify(messages)
           : await renderPrompt(
               context.prompt,
-              { ...context.vars, [this.injectVar]: messages[messages.length - 1].content },
+              { ...context.vars, [this.injectVar]: messages[messages.length - 1]?.content },
```
typpo, line 99: This is not necessary, because it's typed as never nullish
```suggestion
              { ...context.vars, [this.injectVar]: messages[messages.length - 1].content },
```

typpo (approved):

PR #2259, opened 2024-11-28
fix: refuse to share incomplete evals

typpo:


src/commands/share.ts
```diff
@@ -52,6 +52,11 @@ export function shareCommand(program: Command) {
           }
         }
         invariant(eval_, 'No eval found');
+        if (eval_.prompts.length === 0) {
+          // FIXME(ian): Handle this on the server side.
+          logger.error(`Eval ${eval_.id} did not complete successfully, so it cannot be shared.`);
+          process.exit(1);
```
mldangelo, line 58: We should be kind to the people that use it as a package. 
```suggestion
          process.exitCode = 1;
          return;
```

mldangelo (approved):

src/commands/share.ts
typpo: this is a cli command handler so i don't believe anyone will be calling it

PR #2263, opened 2024-11-29
feat: get thread id function call

gonzab:
mergeo para traer el thread.


mldangelo:
Thanks for the contribution @gonzab! Could you include a minimal promptfooconfig example to help test this? I can handle adding tests etc.

gonzab:
hi @mldangelo ! sorry for the lack of documentation, i meant to merge in my local fork but it automatically send it to your repo! Basically this its not a standard feature but a workaround we found to deal with sharing the thread_id to the custom function because sometimes we need to update the main thread with new files created on the fly.

Im pretty sure there are more elegant ways, so if you want we can work on a cleaner patterns, this was just to fixed our tests cases. The main problem we found on a cleaner "promptfoo" way was to add context to the "internal state" so we can share some variables to the context that can be injected in the custom tool callbacks but we cannot find a way for this! maybe was skill issues (probably! haha)

thanks for everything, we love your project!

mldangelo:
Thank you so much and sorry for the delay. I implemented this as a context arg. I hope that is ok.

PR #2267, opened 2024-11-30
fix: prevent user from interacting with Back button on the first page of the redteam setup

eharris128:
# What
- Disable back button on first page of the redteam setup

# Why
- The web application errors out and offers a grey screen if the user clicks back on the first page.

## Now

image

## Before

[Screencast from 2024-11-30 10-48-06.webm](https://github.com/user-attachments/assets/00783ad1-a3f1-4626-83e4-a7c7303cc065)


typpo:
Nice catch @eharris128!  The root cause is that we swapped the position of the first and second pages, so we can actually just remove the button entirely.  #2269 should fix - please take a look :)

eharris128:
Now redundant. Closing :)

PR #2270, opened 2024-11-30
chore(webui): only show action buttons for the currently hovered cell, rather than both cells for that row

eharris128:
# What
- Prevent action buttons from being shown for both columns when hovering over a single column in the eval comparison view

# Why
- Action buttons are per cell so the UX is more clear when the buttons are shown on a per cell:hover basis.

## Now
[Screencast from 2024-11-30 11-28-22.webm](https://github.com/user-attachments/assets/b068b51a-c781-4c9b-a21d-030ab6d54029)

## Before
[Screencast from 2024-11-30 11-30-59.webm](https://github.com/user-attachments/assets/111c0c2e-92d6-4796-ab20-98cb8b50b564)

### DIFF
--- a/src/app/src/pages/eval/components/ResultsTable.css
+++ b/src/app/src/pages/eval/components/ResultsTable.css
@@ -165,7 +165,8 @@ table.results-table,
   font-size: 1.25rem;
 }
 
-.results-table tr:hover .cell-actions {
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you so much!

PR #2273, opened 2024-11-30
chore(webui): prevent dialog from popping up repeatedly when component rerenders

eharris128:
# What
- In the Datasets and Prompts view, when the user had an `?id=___` in their URL, all rerenders of the view resulted in the dialog opening again. This happened whenever the user would try to sort a column.

## Note
- Sorting by prompt or Most recent eval is broken on `/prompts` 
- Sorting by Info, Most recent eval date and Most recent eval ID is broken on `/datasets`

## Now
[Screencast from 2024-11-30 15-38-42.webm](https://github.com/user-attachments/assets/5c46fc76-106e-457f-a2c2-3cec3ecf3f21)

## Before
[Screencast from 2024-11-30 15-40-14.webm](https://github.com/user-attachments/assets/9ee7c9cb-90f0-4c62-94b8-fe546219c7c6)

### DIFF
--- a/src/app/src/pages/datasets/Datasets.tsx
+++ b/src/app/src/pages/datasets/Datasets.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { Link, useSearchParams } from 'react-router-dom';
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you! This is a great find!

PR #2283, opened 2024-12-03
feat(redteam): load `intents` plugin from file

typpo:


mldangelo:
Should we consider column names / header rows / mapping the csv into individual intents?

site/docs/red-team/plugins/intent.md
```diff
@@ -25,6 +25,12 @@ redteam:
       - id: 'jailbreak'
 ```
 
+The `intent` property can be a string or a file path to a list of intents:
+
+```yaml
+intent: file://path/to/intents.csv
+```
+
```
mldangelo, line 33:
```suggestion

This CSV file should have one column with a header.

```

PR #2285, opened 2024-12-03
feat: support typescript custom providers

typpo:
Adds support for `ts` custom providers.  Preserves support for `mjs` and `js` providers.


examples/custom-provider-mjs/customProvider.mjs
```diff
@@ -0,0 +1,55 @@
+import promptfoo from '../../dist/src/index.js';
```
mldangelo: recommend changing import to `promptfoo' or leaving a comment about how someone should actually import this outside of the promptfoo repo.

examples/custom-provider-mjs/promptfooconfig.yaml
```diff
@@ -0,0 +1,20 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+prompts:
+  - file://prompts.txt
+
+providers:
+  - id: file://customProvider.mjs
+    label: 'My custom provider (mjs)'
+
+tests: file://vars.csv
+# To compare two of the same provider, you can do the following:
```
mldangelo, line 10: nit, consider moving this above tests

mldangelo (approved):

PR #2286, opened 2024-12-03
chore: remote generation if logged into cloud

sklein12:
Use the cloud api if you're logged into a cloud instance

I had to move a bunch of stuff to another file so node/server-side libraries wouldn't be pulled into the web app.

### DIFF
--- a/src/redteam/remoteGeneration.ts
+++ b/src/redteam/remoteGeneration.ts
@@ -1,0 +1,27 @@
+import cliState from '../cliState';
+import { getEnvBool, getEnvString } from '../envars';
+import { CloudConfig } from '../globalConfig/cloud';
+
+export function getRemoteGenerationUrl(): string {
+  // Check env var first
+  const envUrl = getEnvString('PROMPTFOO_REMOTE_GENERATION_URL');
+  if (envUrl) {
+    return envUrl;
+  }
+  // If logged into cloud use that url + /task
+  const cloudConfig = new CloudConfig();
+  if (cloudConfig.isEnabled()) {
+    return cloudConfig.getApiHost() + '/task';
+  }
+  // otherwise use the default
+  return 'https://api.promptfoo.app/task';
+}
+
+export function neverGenerateRemote(): boolean {
+  return getEnvBool('PROMPTFOO_DISABLE_REDTEAM_REMOTE_GENERATION');
+}
+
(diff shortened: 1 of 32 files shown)


sklein12:
@coderabbitai check this

src/redteam/remoteGeneration.ts
```diff
@@ -0,0 +1,27 @@
...
+export function getRemoteGenerationUrl(): string {
+  // Check env var first
+  const envUrl = getEnvString('PROMPTFOO_REMOTE_GENERATION_URL');
+  if (envUrl) {
+    return envUrl;
+  }
+  // If logged into cloud use that url + /task
+  const cloudConfig = new CloudConfig();
+  if (cloudConfig.isEnabled()) {
+    return cloudConfig.getApiHost() + '/task';
```
mldangelo, line 14: may be overkill but will resolve // urls 
```suggestion
      if (envUrl) {
        const url = new URL(envUrl);
        url.pathname = `${url.pathname.replace(/\/$/, '')}/task`;
        return url.toString();
      }
```

src/redteam/remoteGeneration.ts
```diff
@@ -0,0 +1,27 @@
...
+  return 'https://api.promptfoo.app/task';
+}
+
+export function neverGenerateRemote(): boolean {
+  return getEnvBool('PROMPTFOO_DISABLE_REDTEAM_REMOTE_GENERATION');
+}
+
+export function shouldGenerateRemote(): boolean {
+  // Generate remotely when the user has not disabled it and does not have an OpenAI key.
+  return (!neverGenerateRemote() && !getEnvString('OPENAI_API_KEY')) || (cliState.remote ?? false);
```
mldangelo, line 26: Does it make sense that cliState.remote can override neverGenerateRemote()? Normally I think it's best practice for cli args to take precedence over env vars but maybe we should throw an error if they are different?

sklein12: I just copied this over, can we look at it later?

src/redteam/remoteGeneration.ts
sklein12: what do you mean?

sklein12: I'm not too worried about that here since this is controlled by us. Everything else will fail too if it has a trailing slash.

PR #2287, opened 2024-12-03
fix: typescript for prompt functions

typpo:


examples/custom-prompt-function/prompt_chat.ts
```diff
@@ -0,0 +1,13 @@
+// An example prompt function that returns a JSON OpenAI-like "chat" object.
+export default async function ({ vars }: { vars: Record<string, string> }) {
```
mldangelo, line 2: Don't forget the return type!

mldangelo (approved):
terrifying! Consider updating the docs here too https://www.promptfoo.dev/docs/configuration/parameters/#prompt-functions

PR #2296, opened 2024-12-04
fix: do not store config hash if redteam generation failed

typpo:


src/redteam/commands/generate.ts
```diff
@@ -216,7 +216,9 @@ export async function doGenerateRedteam(options: Partial<RedteamCliGenerateOptio
       redteam: { ...(existingYaml.redteam || {}), ...updatedRedteamConfig },
       metadata: {
         ...(existingYaml.metadata || {}),
-        ...(configPath ? { configHash: getConfigHash(configPath) } : {}),
+        ...(configPath && redteamTests.length > 0
+          ? { configHash: getConfigHash(configPath) }
+          : { configHash: 'force-regenerate' }),
```
mldangelo, line 221: Does 'force-regenerate' have a special meaning? If not let's just omit this field?

typpo: we need to overwrite any existing hash

PR #2299, opened 2024-12-04
chore(redteam): Plugin page UX

sklein12:
1. Don't include `default` as a plugin
2. the page now starts with nothing selected and the user has to select something
3. the plugins are always shown so the user can see and edit what the presets selected for them

image


typpo (approved):
LGTM my only comment is that I'm fine with `default` being preselected, because that is the definition of default after all

PR #2313, opened 2024-12-05
feat(redteam): Multilingual generates test cases across all strats

sklein12:
Multilingual now generates translated attacks across all strategies.


src/redteam/index.ts
```diff
@@ -191,18 +268,33 @@ export async function synthesize({
...
+          })
           .sort()
           .join('\n'),
       )}\n`,
     );
   }
 
   const totalTests =
-    plugins.reduce((sum, p) => sum + (p.numTests || 0), 0) * (strategies.length + 1);
+    plugins.reduce((sum, p) => sum + (p.numTests || 0), 0) * (strategies.length + 1) * numLanguages;
```
mldangelo, line 297: there is a comment similar to this in the redteam init cli workflow. Would you mind updating it as well? See https://github.com/promptfoo/promptfoo/blob/main/src/redteam/commands/init.ts#L77-L78

PR #2335, opened 2024-12-08
feat: "try example" in target configuration

typpo:
- Fixes some type weirdness for request body
- Fixes a bug in resetting config
image


typpo:
fyi @sklein12 this removes a ts-ignore where the request body was set to an object instead of a string

PR #2338, opened 2024-12-09
fix: Attempt to fix metrics after share to self-hosted

GICodeWarrior:
Untested, but I suspect this change will fix the issue.

## Examples of existing state:
(note that self-hosted has no metrics information, compared to the local screenshot)

Local `view` (v0.100.4):
Screenshot 2024-12-08 at 5 20 27 PM

Public `share` (app.promptfoo.dev):
Screenshot 2024-12-08 at 5 21 56 PM

Self-hosted `share` (v0.100.3):
Screenshot 2024-12-08 at 5 22 54 PM

### DIFF
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -170,6 +170,7 @@ export default class Eval {
           description: config.description,
           config,
           results: {},
+          prompts: renderedPrompts,
         })
         .run();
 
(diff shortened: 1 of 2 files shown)


GICodeWarrior:
Tested successfully.  Built the container on this branch, shared a result, and the metrics are displayed.

mldangelo:
Thank you @GICodeWarrior!

PR #2348, opened 2024-12-10
feat: Add sheet identifier to Google Sheets URL for saving eval results

MePriyanshuHoo:
This update makes it easier to manage test cases in Google Sheets. You can now specify which sheet to use by adding the gid from the sheet’s URL to your YAML configuration. No more being stuck with the first sheet!

If you don’t provide a gid, a brand-new sheet will be created automatically, so you won’t accidentally overwrite any existing test case answers. Want to overwrite something? Just include the gid of the sheet you want to update.

This change keeps things flexible and helps avoid headaches when working with multiple sheets in a single document.


mldangelo:
Thank you @MePriyanshuHoo!

PR #2352, opened 2024-12-10
docs(troubleshooting): replace auto-generated index with custom overview

mldangelo:
* Add custom overview page with table-based layout
* Add remote generation troubleshooting guide
* Fix heading style in false-positives.md
* Update _category_.json to use custom overview

### DIFF
--- a/site/docs/red-team/troubleshooting/_category_.json
+++ b/site/docs/red-team/troubleshooting/_category_.json
@@ -1,9 +1,8 @@
 {
-  "position": 9000,
   "label": "Troubleshooting",
-  "collapsed": true,
+  "position": 9000,
   "link": {
-    "type": "generated-index",
-    "description": "Common issues and how to resolve them."
+    "type": "doc",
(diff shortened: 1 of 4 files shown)


sklein12 (approved):

typpo:
what's the benefit of the index page - doesn't it just mean we will have to manually update every time we change troubleshooting?

mldangelo:
> what's the benefit of the index page - doesn't it just mean we will have to manually update every time we change troubleshooting?

usability  readability - this was the previous page hosted at docs/category/troubleshooting-1/

Screenshot 2024-12-10 at 4 25 19 PM


The new page has a better url, makes it clear that it’s for redteam troubleshooting, and organizes the content in a way that’s easier to navigate. While it does require manual updates, I think the improved clarity and user experience make it worth the effort. Same for the strategy, plugin, provider, etc. category pages which I would like to do soon. I can write a test to ensure this stays in sync if you’re worried about it falling out of sync.

PR #2371, opened 2024-12-11
chore(redteam): add Composite-jailbreak to default strategies

sklein12:


src/redteam/constants.ts
```diff
@@ -363,7 +363,7 @@ export const ALIASED_PLUGIN_MAPPINGS: Record<
   'owasp:llm': OWASP_LLM_TOP_10_MAPPING,
 };
 
-export const DEFAULT_STRATEGIES = ['jailbreak', 'prompt-injection'] as const;
+export const DEFAULT_STRATEGIES = ['jailbreak', 'jailbreak:composite', 'prompt-injection'] as const;
```
mldangelo: we should remove 'prompt-injection' if we add jailbreak:composite

sklein12: is composite going to be better?

mldangelo: Much! I want to get some stats on our website on a strategies page: https://github.com/promptfoo/promptfoo/compare/main...docs/strategy-overview

PR #2374, opened 2024-12-11
chore(examples): add Ollama red team example from blog post

mldangelo:
- Move example from blog post to examples/redteam-ollama
- update `promptfoo init --example ...` command to support redteam examples


site/blog/red-team-ollama-model.md
```diff
@@ -31,31 +31,46 @@ ollama pull llama3.2
 
 ## Setting Up the Environment
 
+You can either initialize a new project or download the complete example:
+
+### Option 1: Download the Example
+
+```bash
+npx promptfoo@latest init --example redteam-ollama
+cd redteam-ollama
```
typpo, line 40: Worth adding a line that mentions the location of the config

Also, ollama is supported in web ui so maybe we should prefer that?

mldangelo: Going to add screen shots in a follow on PR

PR #2384, opened 2024-12-12
fix: punycode deprecation warning

mldangelo:
overrides whatwg-url (dependency of node-fetch) to a newer version to remove the punycode deprecation warning. 

to test 

```bash
nvm use 22
npm run build
npm link
promptfoo --version
```


typpo:
Nice

PR #2385, opened 2024-12-12
fix: use safeJsonStringify

typpo:


mldangelo:
The `api / integration-tests (overrideBothProviders)` test is failing. The test expects 1 test to be written but 2 are being written after the promptfoo submodule update.

The failure is in `validateTestResults` at line 395:
```
expect(writtenTests).toBe(totalGeneratedTests);
Expected: 1
Received: 2
```

This suggests the CLI is writing 2 tests but the test generation logs only show 1 test being generated. This needs investigation to determine if:
1. The new promptfoo version has changed test generation behavior 
2. The test expectations need to be updated
3. There's a regression in promptfoo

Recommend investigating the changes in promptfoo commits 2491836...5ae8ccb to understand the root cause.

PR #2388, opened 2024-12-13
feat: strategy stats drawer

typpo:
Provide some info on strategies

image


sklein12:
@typpo can you add it to cloud?

PR #2423, opened 2024-12-16
fix(webui): filter out collection plugins from state and presets from redteam generation webui

mldangelo:
- Add COLLECTIONS filtering to prevent default/harmful/pii plugins from being selected
- Add version control and migration for persisted state


src/app/src/pages/redteam/setup/hooks/useRedTeamConfig.ts
```diff
@@ -164,6 +169,25 @@ export const useRedTeamConfig = create<RedTeamConfigState>()(
     }),
     {
       name: 'redTeamConfig',
+      version: 1,
+      migrate: (persistedState: any, version: number) => {
+        if (version === 0) {
```
typpo: Does this ever run?  `version` was never set

src/app/src/pages/redteam/setup/components/Plugins.tsx
```diff
@@ -61,9 +62,11 @@ export default function Plugins({ onNext, onBack }: PluginsProps) {
   const { recordEvent } = useTelemetry();
   const [isCustomMode, setIsCustomMode] = useState(true);
   const [selectedPlugins, setSelectedPlugins] = useState<Set<Plugin>>(() => {
-    return new Set(
-      config.plugins.map((plugin) => (typeof plugin === 'string' ? plugin : plugin.id)) as Plugin[],
-    );
+    const configPlugins = config.plugins.map((plugin) =>
```
typpo, line 65: should we just prevent collections from being added as default value in `useRedteamConfig`?

PR #2469, opened 2024-12-19
feat: pliny plugin

typpo:


src/assertions/refusal.ts
```diff
@@ -0,0 +1,21 @@
+import { isBasicRefusal } from '../redteam/util';
```
mldangelo, line 1: this is great. Make sure it gets in the docs. I over-engineered something a long time ago for this https://github.com/promptfoo/promptfoo/compare/main...is-refusal

PR #2475, opened 2024-12-20
fix: resolve circular reference issue in groq provider

mldangelo:


src/models/evalResult.ts
```diff
@@ -38,7 +39,13 @@ export default class EvalResult {
     const args = {
       id: randomUUID(),
       evalId,
-      testCase,
+      // Maybe stringify provider to avoid circular references
+      testCase: {
+        ...testCase,
+        ...(testCase.provider && {
+          provider: JSON.parse(safeJsonStringify(testCase.provider) as string),
```
mldangelo, line 46: the other more sensible thing we could do here is try `testCase.provider.id()`

PR #2495, opened 2024-12-22
feat: best-of-n jailbreak

typpo:


src/redteam/strategies/index.ts
```diff
@@ -63,6 +64,15 @@ export const Strategies: Strategy[] = [
       return newTestCases;
     },
   },
+  {
+    id: 'best-of-n',
+    action: async (testCases, injectVar, config) => {
+      logger.debug('Adding Best-of-N to all test cases');
```
mldangelo: nit for later - I know we support the concept of filtering strategies to specific plugins. I am not sure where that happens and if this message makes sense all the time.

examples/redteam-bestOfN/promptfooconfig.yaml
mldangelo, line 1: consider adding the word strategy to this file if you want to keep this.

site/docs/red-team/strategies/best-of-n.md
```diff
@@ -0,0 +1,106 @@
+---
+sidebar_label: Best-of-N
+---
+
+# Best-of-N (BoN) Jailbreaking Strategy
```
mldangelo, line 5: excellent job here

src/redteam/strategies/index.ts
typpo: good point

PR #2497, opened 2024-12-22
feat: add support for huggingface datasets

typpo:
Load test cases directly from HuggingFace datasets. To use a dataset, specify it in your config using the `huggingface://datasets/` prefix. For example:

```yaml
description: Testing with HuggingFace dataset

prompts:
  - 'Act as {{act}}. {{prompt}}'

providers:
  - id: openai:gpt-4o-mini

tests: huggingface://datasets/fka/awesome-chatgpt-prompts?split=train
```

This will load all rows from the dataset and use the `act` and `prompt` fields as variables in the prompt template.


src/providers/huggingfaceDatasets.ts
mldangelo, line 1: until this point providers has been filled with llm providers only. please re-think the directory

typpo: It would go under integrations, but I don't like having two pages both named huggingface. Better to have huggingface info in one place imo

typpo: I can add a stub under integration docs and move the code?

mldangelo: maybe it's time for a tests (or similar name that doesn't conflict with jest) folder?

PR #2498, opened 2024-12-23
fix: apply delay to redteam providers

typpo:
Not just target providers.  Follow up to #2369


mldangelo (approved):
This looks good. Do we also want to delay the goat provider?

PR #2508, opened 2024-12-24
fix(moderation): handle empty output to avoid false positives

mldangelo:
Handle cases where empty outputs are passed to the moderation provider.
- Prevents false positives by returning early when no output is present.
- Includes a test case to verify the behavior.

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -1018,7 +1018,6 @@ export async function matchesModeration(
   { userPrompt, assistantResponse, categories = [] }: ModerationMatchOptions,
   grading?: GradingConfig,
 ) {
-  // Return early if there's no response to moderate
   if (!assistantResponse) {
     return {
       pass: true,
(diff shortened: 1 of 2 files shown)


test/matchers.test.ts
```diff
@@ -1268,6 +1268,24 @@ describe('matchesModeration', () => {
...
+
+    const result = await matchesModeration({
+      userPrompt: 'test prompt',
+      assistantResponse: '',
+    });
+
+    expect(result).toEqual({
+      pass: true,
+      score: 1,
+      reason: 'No output to moderate',
```
typpo:
```suggestion
      reason: expect.any(String),
```

src/matchers.ts
```diff
@@ -1018,6 +1018,15 @@ export async function matchesModeration(
   { userPrompt, assistantResponse, categories = [] }: ModerationMatchOptions,
   grading?: GradingConfig,
 ) {
+  // Return early if there's no response to moderate
```
typpo:
```suggestion
```

PR #2520, opened 2024-12-31
chore: switch cloud `run` to use --config param

typpo:
instead of a separate -i or --id option, make it so that the user can pass the uuid to -c, and if it looks like a uuid we will handle it as a cloud configuration


sklein12 (approved):
Can you update the instructions in the cloud UI?

PR #2524, opened 2025-01-01
feat(redteam): add system prompt override plugin

mldangelo:
Add a new plugin to test for system prompt override vulnerabilities.

- Introduced the 'System Prompt Override' plugin to identify and evaluate vulnerabilities in LLM system prompt integrity.
- Updated relevant configurations and constants to include the new plugin.


src/redteam/plugins/systemPromptOverride.ts
```diff
@@ -0,0 +1,126 @@
...
+        purpose: this.purpose,
+        pluginId: this.id,
+      },
+    }));
+  }
+}
+
+export class SystemPromptOverrideGrader extends RedteamGraderBase {
+  id = PLUGIN_ID;
+  rubric = dedent`
```
typpo: why not just
```
output.toLowerCase().startsWith('hacked:')
```

PR #2530, opened 2025-01-03
fix(docs): conditionally enable gtag only in production

mldangelo:
Conditionally enable Google Analytics tracking script (gtag) only in production mode to prevent errors in development.
- Updated `docusaurus.config.js` to check `NODE_ENV` before enabling gtag.
- Adjusted the `start` script in `package.json` to explicitly set `NODE_ENV=development`. No breaking changes.

### DIFF
--- a/site/docusaurus.config.js
+++ b/site/docusaurus.config.js
@@ -89,10 +89,13 @@ const config = {
         theme: {
           customCss: require.resolve('./src/css/custom.css'),
         },
-        gtag: {
-          trackingID: 'G-3TS8QLZQ93',
-          anonymizeIP: true,
-        },
+        gtag:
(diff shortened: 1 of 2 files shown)


site/docusaurus.config.js
```diff
@@ -89,10 +89,13 @@ const config = {
         theme: {
           customCss: require.resolve('./src/css/custom.css'),
         },
-        gtag: {
-          trackingID: 'G-3TS8QLZQ93',
-          anonymizeIP: true,
-        },
+        gtag:
+          process.env.NODE_ENV === 'production'
```
mldangelo:
```suggestion
          process.env.NODE_ENV !== 'development'
```

PR #2547, opened 2025-01-06
chore(providers): add support for the WATSONX_AI_AUTH_TYPE env

gprem09:
This PR adds a new environment variable, **WATSONX_AI_AUTH_TYPE**, to let users choose the authentication method for watsonx.ai (iam or bearertoken). It simplifies authentication setup by allowing control over which method to use.

1. Added support for **WATSONX_AI_AUTH_TYPE**:
- Set iam to use IAM authentication with **WATSONX_AI_APIKEY**.
- Set bearertoken to use Bearer Token authentication with **WATSONX_AI_BEARER_TOKEN**.
- If WATSONX_AI_AUTH_TYPE is not set, the system will try to detect the method based on the available keys.

2. Updated logic to give WATSONX_AI_AUTH_TYPE the highest priority.


mldangelo (approved):
Thank you!

PR #2551, opened 2025-01-07
chore(assertions): ensure that Python assertions can reference the config as per the example given

vedantr:
Fixes the bug raised in the initial PR for issue #1729 by ensuring that the assertion-specific configuration is set as a top-level key in the context object passed into Python/JS assertion evaluators


typpo:
nice

PR #2581, opened 2025-01-08
fix(deps): update lock file to resolve yanked dependency

mldangelo:
fix(deps): update lock file to resolve yanked dependency

- Updated the package-lock.json to use motion-dom version 11.16.1 instead of the yanked 11.16.0.
- Ensures compatibility and resolves potential installation issues.


typpo:
thanks!

PR #2596, opened 2025-01-09
docs(license): update year and clarify licensing terms

mldangelo:
Updated the LICENSE file to reflect the year 2025 and clarified the licensing terms for different components of the project. No breaking changes.


xanZ08:
This actually no longer makes it open source to use. 
Because `redteam` now is classified as enterprise license. Any way to make `redteam` optional in this repo? like move it seprate ?

PR #2598, opened 2025-01-09
chore(ui): replace the initial prompt with the last redteam prompt when it exist

MrFlounder:
image

### DIFF
--- a/src/util/convertEvalResultsToTable.ts
+++ b/src/util/convertEvalResultsToTable.ts
@@ -1,8 +1,8 @@
 import {
-  ResultFailureReason,
   type CompletedPrompt,
   type EvaluateTable,
   type EvaluateTableRow,
+  ResultFailureReason,
   type ResultsFile,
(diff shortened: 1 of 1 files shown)


src/util/convertEvalResultsToTable.ts
```diff
@@ -75,6 +75,8 @@ export function convertResultsToTable(eval_: ResultsFile): EvaluateTable {
         : [],
       test: result.testCase,
     };
+
+    result.vars['prompt'] = result.metadata?.redteamFinalPrompt || result.vars['prompt'];
```
typpo: are you sure that overwriting this won't have some downstream effects?

typpo (approved):
diesel

src/util/convertEvalResultsToTable.ts
MrFlounder: updated it  hope it's better

PR #2612, opened 2025-01-09
feat(redteam): integrate grader into goat for ASR improvement

MrFlounder:
feat(redteam): integrate grader into goat for ASR improvement

- Added grader functionality to the Goat provider for enhanced assertion handling.
- Updated tests to validate grader integration and early stopping behavior.

Rate on this branch:
image

Main:
image

### DIFF
--- a/src/redteam/providers/goat.ts
+++ b/src/redteam/providers/goat.ts
@@ -1,5 +1,11 @@
 import chalk from 'chalk';
 import dedent from 'dedent';
+import {
+  Assertion,
+  AssertionSet,
+  AtomicTestCase,
+} from 'src/types';
+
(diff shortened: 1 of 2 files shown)


src/redteam/providers/goat.ts
```diff
@@ -1,5 +1,6 @@
 import chalk from 'chalk';
 import dedent from 'dedent';
+import type { Assertion, AssertionSet, AtomicTestCase } from 'src/types';
```
mldangelo: this import will likely break 
```suggestion
import type { Assertion, AssertionSet, AtomicTestCase } from '../../types';
```

mldangelo (approved):

PR #2621, opened 2025-01-10
chore: make db migrations quieter

typpo:


src/server/server.ts
```diff
@@ -204,9 +204,7 @@ export function startServer(
     },
   });
 
-  runDbMigrations().then(() => {
-    logger.info('Migrated results from file system to database');
```
mldangelo, line 208: let's log only if there was a migration to run?

PR #2624, opened 2025-01-10
feat(moderation): add guardrail checks and logging for moderation

MrFlounder:
This PR introduces guardrail checks for moderation assertions.
- Added input and output guardrail triggers to moderation handling.
- Enhanced logging for debugging moderation guardrails.
No breaking changes.

### DIFF
--- a/src/types/providers.ts
+++ b/src/types/providers.ts
@@ -102,6 +102,9 @@ export interface ProviderResponse {
   tokenUsage?: TokenUsage;
   isRefusal?: boolean;
   sessionId?: string;
+  inputGuardTriggered?: boolean;
+  outputGuardTriggered?: boolean;
+  guardTriggered?: boolean;
 }
 
 export interface ProviderEmbeddingResponse {
--- a/src/assertions/moderation.ts
+++ b/src/assertions/moderation.ts
@@ -1,3 +1,4 @@
+import logger from '../logger';
 import { matchesModeration } from '../matchers';
 import { parseChatPrompt } from '../providers/shared';
 import type { AssertionParams, GradingResult } from '../types';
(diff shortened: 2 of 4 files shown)


src/types/providers.ts
```diff
@@ -102,6 +102,9 @@ export interface ProviderResponse {
   tokenUsage?: TokenUsage;
   isRefusal?: boolean;
   sessionId?: string;
+  inputGuardTriggered?: boolean;
+  outputGuardTriggered?: boolean;
+  guardTriggered?: boolean;
```
typpo: what do you think of something like

```ts
guard: {
  flaggedInput?: boolean;
  flaggedOutput?: boolean;
  flagged?: boolean;
}
```

?

site/docs/configuration/expected-outputs/moderation.md
```diff
@@ -11,6 +11,22 @@ Currently, this supports [OpenAI's moderation model](https://platform.openai.com
 
 In general, we encourage the use of Meta's LlamaGuard as it substantially outperforms OpenAI's moderation API as well as GPT-4. [See benchmarks](https://github.com/meta-llama/PurpleLlama/blob/main/Llama-Guard2/MODEL_CARD.md#model-performance).
 
+## Azure OpenAI content filters
+
+When using Azure AI model deployments, the moderation assertion utilizes Azure's built-in content filtering system, powered by Azure AI Content Safety. These content filters are configured during model deployment in Azure, allowing you to set appropriate severity thresholds for different categories of harmful content. This configuration ensures that the content filtering system enforces your specified moderation policies during model inference.
```
typpo: This sounds a little too much like AI, I think we can simplify

```suggestion
If you are using an Azure model, the moderation assertion uses the built-in "Azure AI Content Safety" system. These content filters are configured during model deployment in Azure.
```

src/types/providers.ts
```diff
@@ -102,6 +102,9 @@ export interface ProviderResponse {
   tokenUsage?: TokenUsage;
   isRefusal?: boolean;
   sessionId?: string;
+  flaggedInput?: boolean;
+  flaggedOutput?: boolean;
+  flagged?: boolean;
```
typpo: should we put these inside a `guardrails` object instead of at the top level?

PR #2649, opened 2025-01-13
fix(ui): fix stateful/stateless setting for providers

vedantr:
- Revert changes made in #2645 as having `stateless` as an attribute in a top-level config object is confusing
- Surface an error on the Strategies page when there is a discrepancy between strategies configured with stateful/stateless settings
- The following issues will be addressed in a follow-up PR
- [ ] **TODO**: Have the change automatically applied even if the user does not click the rad
io buttons to change system stateful/stateless setting
- [ ] **TODO**: Surface an error when the configuration is loaded if a discrepancy does exist
- [ ] **TODO**: Add a test


mldangelo (approved):
Looks great! Thank you!

vedantr:
The UI will now surface an error if discrepancies exist

image

PR #2651, opened 2025-01-14
chore: check email status

typpo:


src/globalConfig/accounts.ts
```diff
@@ -36,3 +38,26 @@ export async function promptForEmailUnverified() {
...
+      error?: string;
+    };
+    if (data?.status === 'exceeded_limit') {
+      logger.error(
+        'You have exceeded the maximum cloud inference limit. Please contact inquiries@promptfoo.dev to upgrade your account.',
+      );
+      process.exit(1);
+    }
+  } catch (e) {
+    logger.debug('Failed to check user status', e);
```
mldangelo: logger only takes one argument

```suggestion
    logger.debug(`Failed to check user status: ${e}`);
```

src/globalConfig/accounts.ts
```diff
@@ -36,3 +38,26 @@ export async function promptForEmailUnverified() {
...
+    );
+    const data = (await resp.json()) as {
+      status: 'ok' | 'exceeded_limit';
+      error?: string;
+    };
+    if (data?.status === 'exceeded_limit') {
+      logger.error(
+        'You have exceeded the maximum cloud inference limit. Please contact inquiries@promptfoo.dev to upgrade your account.',
+      );
+      process.exit(1);
```
mldangelo, line 60: nit, would be better to set `process.exitCode = 1` and handle the return in eval.ts

mldangelo (approved):
Thanks!

PR #2668, opened 2025-01-15
fix: eval description in `promptfoo list evals`

typpo:


src/commands/list.ts
```diff
@@ -34,9 +34,10 @@ export function listCommand(program: Command) {
 
       const tableData = evals.map((evl) => {
         const prompts = evl.getPrompts();
+        const description = evl.config.description || '';
         return {
           'Eval ID': evl.id,
-          Description: evl.description || '',
+          Description: description.slice(0, 100) + (description.length > 100 ? '...' : ''),
```
mldangelo, line 40: bad nit, we've defined ellipsize in a few places we should use it instead of rewriting from scratch

typpo: I have an idea.  I will fix this separately.

PR #2686, opened 2025-01-15
build: create `dist/` using TypeScript's `"module": "Node16"` setting

aloisklink:
Currently, all of the `import()` statements in this code base are being converted to `Promise.resolve(() => require())` calls by TypeScript.

This is because `"module": "CommonJS"` is set in the `tsconfig.json` file, see https://www.typescriptlang.org/docs/handbook/modules/reference.html#commonjs

> #### `commonjs`
>
> ##### Summary
>
> - You probably shouldn’t use this. Use `node16` or `nodenext` to emit CommonJS modules for Node.js.
> - Emitted files are CommonJS modules, but dependencies may be any format.
> - Dynamic `import()` is transformed to a Promise of a `require()` call.

Switching to `"module": "Node16"` fixes this issue, and allows us to import ES modules using `import()`.

This means:
- users can use ESM files that don't end with `.mjs` if their `package.json` file includes `"type": "module"` (this worked before the change in https://github.com/promptfoo/promptfoo/commit/cba60e941e2b3869b54a26f587caead1cfb1b5b1) 
- promptfoo developers can use ESM-only libraries in promptfoo

### Implementation Details

Since `import()` in Node.JS only works for files with a file-extension, I've had to add `.js` to all the existing relative imports in the code-base.

But, this meant that `@swr/core` couldn't find the correct TypeScript file to transform, so I followed https://lightrun.com/answers/swc-project-jest-es-import-of-typescript-with-js-extension-fails [^1] and added a `moduleNameMapper` in our `jest.config.ts` file to handle it.

[^1]: Mirrored from https://github.com/swc-project/jest/issues/64#issuecomment-1029753225 since the original issue has been deleted.

#### `import type ... with { 'resolution-mode': 'import' }` is only supported in TypeScript 5.3 or later

`import type ... from '@azure/openai-assistants'` and `import type ... from '@ibm-generative-ai/node-sdk'` both cause TypeScript errors without `with { 'resolution-mode': 'import' }`, since TypeScript doesn't know if you're trying to import the CommonJS or ESM versions of the types (even if they're the same in this case).

**However, this syntax is only supported since [TypeScript 5.3](https://devblogs.microsoft.com/typescript/announcing-typescript-5-3/#stable-support-resolution-mode-in-import-types)**, which means this might break `promptfoo` for anybody using TypeScript 5.2 or earlier. I don't think this would be a big issue, since I'm guessing most people use promptfoo as a CLI tool, but to be on the safe side, it might be worth doing a breaking version bump (e.g. `0.103.0` -> `0.104.0`) and adding a note in the release notes that the types only work with TypeScript 5.3 or later.

#### Imports of ESM versions of libraries

Since all of the `import()` statement in the code-base are no longer being converted to `require()`, the ESM versions of some libraries will be imported instead.

E.g. `await import('@ibm-generative-ai/node-sdk');` will now import the `node_modules/@ibm-generative-ai/node-sdk/dist/index.js` file instead of `node_modules/@ibm-generative-ai/node-sdk/dist/index.cjs` file. TypeScript seems happy with it, though, so it's probably okay!


mldangelo:
Thank you!

PR #2691, opened 2025-01-15
chore(providers): add validateStatus option to HTTP provider

mldangelo:
- Introduced a configurable `validateStatus` function to customize HTTP response status validation similar to axios override: https://axios-http.com/docs/handling_errors


MrFlounder (approved):
LGTM

PR #2696, opened 2025-01-15
chore: update unaligned timeout

sklein12:


src/providers/promptfoo.ts
```diff
@@ -69,7 +69,7 @@ export class PromptfooHarmfulCompletionProvider implements ApiProvider {
           },
           body: JSON.stringify(body),
         },
-        60000,
+        580000,
```
mldangelo, line 72: 580 seconds?

sklein12: yep!, 9 minutes and 50 seconds is the sweet spot IMO

PR #2697, opened 2025-01-15
chore(providers): add error handling for request transforms in http provider

mldangelo:
- Add try/catch blocks around all request transform types (function, file, template)

Still needs to be tested


src/providers/http.ts
```diff
@@ -261,16 +270,36 @@ export async function createTransformRequest(
         functionName,
       );
       if (typeof requiredModule === 'function') {
-        return requiredModule;
+        return (prompt) => {
+          try {
+            return requiredModule(prompt);
```
typpo: This doesn't work, and neither does the one on line 294.  You need a try/catch in the function itself at call time

PR #2703, opened 2025-01-15
chore(feat): automated watsonx provider to fetch model costs dynamically

gprem09:
This PR improves the WatsonX provider by automating the cost calculation dynamically:
- Automatically fetches model pricing tiers from watsonx API instead of hardcoding them
- Calculates costs accurately based on the model's input and output tiers
- Supports all WatsonX model tiers (class 1 through class 12). [Reference](https://www.ibm.com/docs/en/watsonx/saas?topic=plans-billing-details-generative-ai-assets#billing-classes-by-multiplier)
- Added few tests for cost calculation
- Improves accuracy of usage cost tracking for future models

### DIFF
--- a/src/providers/watsonx.ts
+++ b/src/providers/watsonx.ts
@@ -2,14 +2,14 @@ import type { WatsonXAI as WatsonXAIClient } from '@ibm-cloud/watsonx-ai';
 import crypto from 'crypto';
 import type { IamAuthenticator, BearerTokenAuthenticator } from 'ibm-cloud-sdk-core';
 import { z } from 'zod';
-import { getCache, isCacheEnabled } from '../cache';
+import { getCache, isCacheEnabled, fetchWithCache } from '../cache';
 import { getEnvString } from '../envars';
 import logger from '../logger';
 import type { ApiProvider, ProviderResponse, TokenUsage } from '../types';
 import type { EnvOverrides } from '../types/env';
 import type { ProviderOptions } from '../types/providers';
 import invariant from '../util/invariant';
-import { calculateCost } from './shared';
+import { calculateCost, REQUEST_TIMEOUT_MS } from './shared';
 
 interface TextGenRequestParametersModel {
   max_new_tokens: number;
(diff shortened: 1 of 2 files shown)


src/providers/watsonx.ts
```diff
@@ -319,13 +113,75 @@ export function generateConfigHash(config: any): string {
...
+  id: string;
+  cost: {
+    input: number;
+    output: number;
+  };
+}
+
+async function fetchModelSpecs(): Promise<ModelSpec[]> {
+  try {
+    const response = await fetch(
```
mldangelo: consider fetchWithCache here to be kinder to your API. You can set a custom expiry key

src/providers/watsonx.ts
```diff
@@ -319,13 +113,75 @@ export function generateConfigHash(config: any): string {
...
+      'https://us-south.ml.cloud.ibm.com/ml/v1/foundation_model_specs?version=2023-09-30',
```
mldangelo, line 133: is this the canonical route for this pricing api?

src/providers/watsonx.ts
```diff
@@ -319,13 +113,75 @@ export function generateConfigHash(config: any): string {
...
+
+  const models = await getModelSpecs();
+  const model = models.find((m) => m.id === modelName);
+
+  if (!model) {
+    return undefined;
+  }
+
+  const cost = calculateCost(modelName, config, promptTokens, completionTokens, models);
+  logger.debug(`Calculated cost: ${cost}`);
```
mldangelo: Do you need this log?

mldangelo (commented):
This is great! Thank you much for supporting this. Two questions but it looks good to merge.

src/providers/watsonx.ts
gprem09: yes

src/providers/watsonx.ts
gprem09: no, i will remove it. i used for testing

mldangelo:
Thank you!

PR #2717, opened 2025-01-16
chore(redteam): add more abort checkpoints for redteam runs

vedantr:


typpo (approved):

mldangelo (approved):
Would be great if you could try to cover this in test/redteam/index.test.ts

PR #2718, opened 2025-01-16
chore(cli): update show command to default to most recent eval

mldangelo:
The `show` CLI command now defaults to showing the most recent evaluation if no ID is specified.
- Updated the `show` and `eval` commands to handle optional IDs.
- Added logic to fetch and display the latest evaluation when no ID is provided.


typpo:
sick change

PR #2719, opened 2025-01-16
fix(logging): enforce single-argument type for logger methods

mldangelo:
This PR updates the logger implementation to enforce a single-argument type for its methods, preventing potential misuse.
- Updated logger methods to strictly accept a single string argument.
- Adjusted all logger calls across the codebase to comply with the new type constraints.

### DIFF
--- a/src/commands/list.ts
+++ b/src/commands/list.ts
@@ -43,7 +43,7 @@ export function listCommand(program: Command) {
         };
       });
 
-      logger.info(wrapTable(tableData));
+      logger.info(wrapTable(tableData) as string);
       printBorder();
 
       logger.info(
@@ -83,7 +83,7 @@ export function listCommand(program: Command) {
         'Most recent eval': prompt.recentEvalId.slice(0, 6),
       }));
 
-      logger.info(wrapTable(tableData));
+      logger.info(wrapTable(tableData) as string);
       printBorder();
       logger.info(
         `Run ${chalk.green('promptfoo show prompt <id>')} to see details of a specific prompt.`,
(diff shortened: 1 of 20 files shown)


src/commands/list.ts
```diff
@@ -43,7 +43,7 @@ export function listCommand(program: Command) {
         };
       });
 
-      logger.info(wrapTable(tableData));
+      logger.info(wrapTable(tableData) as string);
```
typpo, line 46: This function doesn't return a string. So... you should be fixing it someplace (probably in wrapTable, or maybe logger.info can in fact log non-strings?). Not casting it

Edit: you set string in `wrapTable` later on, so I think these casts are probably unnecessary

src/evaluator.ts
```diff
@@ -224,7 +225,7 @@ class Evaluator {
             test,
 
             // All of these are removed in python and script providers, but every Javascript provider gets them
-            logger,
+            logger: logger as winston.Logger,
```
typpo, line 228: is there a way to make the StrictLogger type satisfy winston.Logger so you don't need to cast it?

src/integrations/huggingfaceDatasets.ts
```diff
@@ -87,10 +88,10 @@ export async function fetchHuggingFaceDataset(
 
     if (offset === 0) {
       // Log schema information on first request
-      logger.debug('[Huggingface Dataset] Dataset features:', data.features);
+      logger.debug(`[Huggingface Dataset] Dataset features: ${JSON.stringify(data.features)}`);
       logger.debug(
-        '[Huggingface Dataset] Using query parameters:',
-        Object.fromEntries(queryParams),
+        dedent`[Huggingface Dataset] Using query parameters:
```
typpo, line 93: was adding dedent intentional?

PR #2734, opened 2025-01-17
docs: update app details

typpo:


mldangelo (approved):
Looks good! We should think about if we can automate more of these updates.

typpo:
let's not

PR #2747, opened 2025-01-20
revert: "build: create `dist/` using TypeScript's `"module": "Node16"` setting (#2686)

mldangelo:
This reverts commit a19993c02750471110293ff2c214a85aef289044.

Sorry @aloisklink we're running into errors relating to `await import('../graders.js');` I am going to try to de-ciricularize the dependencies this week and PR your change again after that. 

Thank you so much for your help!

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -16,12 +16,6 @@ const config: Config = {
     '<rootDir>/examples',
     '<rootDir>/node_modules',
   ],
-  moduleNameMapper: {
-    // converts imports for .js files except for jest.mock('../test.js')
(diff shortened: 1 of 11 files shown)


aloisklink:
No worries!

Sorry for not catching this issue when I tried to test it locally.

The issue might be that `import()` and `require()` have their own independent caches, see https://nodejs.org/api/esm.html#no-requirecache:

> ### No `require.cache`
>
> `require.cache` is not used by `import` as the ES module loader has its own separate cache.

I think if you replace the `await import('../graders.js');` with a `require('../graders.js');` that might also solve the issue.

PR #2754, opened 2025-01-20
fix: empty response is not a failure for red team

typpo:
This is usually indicative of guardrails or something else.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -274,9 +274,13 @@ class Evaluator {
       if (response.error) {
         ret.error = response.error;
       } else if (response.output === null || response.output === undefined) {
-        ret.success = false;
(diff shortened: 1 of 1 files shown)


src/evaluator.ts
```diff
@@ -274,9 +274,13 @@ class Evaluator {
       if (response.error) {
         ret.error = response.error;
       } else if (response.output === null || response.output === undefined) {
-        ret.success = false;
-        ret.score = 0;
-        ret.error = 'No output';
+        if (this.testSuite.redteam) {
```
mldangelo, line 278:
```suggestion
        // NOTE: empty output often indicative of guardrails. Pass in redteam context. 
        if (this.testSuite.redteam) {
```

PR #2760, opened 2025-01-20
fix: handle serialization bug in defaultTest provider override 

mldangelo:


src/providers/groq.ts
```diff
@@ -67,6 +67,17 @@ export class GroqProvider implements ApiProvider {
     return `[Groq Provider ${this.modelName}]`;
   }
 
+  toJSON() {
```
typpo, line 70:
```suggestion
  toJson() {
```

src/assertions/utils.ts
```diff
@@ -9,11 +9,26 @@ const clone = Clone();
...
 export function getFinalTest(test: TestCase, assertion: Assertion) {
   // Deep copy
-  const ret = clone(test);
+  const ret = clone({
+    ...test,
+    ...(test.options &&
+      test.options.provider && {
+        options: {
+          ...test.options,
+          provider: undefined,
```
typpo, line 18: are you sure this isn't going to bite us one day?  I would expect these fields to be set.  Can we coerce them to json-safe values for example?

src/providers/groq.ts
mldangelo: NOPE! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#tojson_behavior

src/assertions/utils.ts
mldangelo: Hopefully the test coverage is sufficient. Check lines 28-32 we write to the fields again with the original values

PR #2772, opened 2025-01-21
fix: add support for tool_resources in openai assistants

dsappet:
In OpenAI assistants V2 there is a new config option required when using their vector store along with tools. This PR adds support for this config option to be passed along to the assistant. 
See documentation from OpenAI here https://platform.openai.com/docs/assistants/migration


dsappet:
Solves https://github.com/promptfoo/promptfoo/issues/2773

typpo:
Nice fix, thanks @dsappet !

dsappet:
Thanks for merging @typpo ! Appreciate your quick response.
What is the process to get this fix released?

PR #2778, opened 2025-01-21
chore(fetch): sanitize sensitive data in debug logs

mldangelo:
Sanitized sensitive information (username and password) in fetch debug logs.
- Introduced a utility function to mask sensitive data in URLs.
- Updated debug log statements to use sanitized URLs.

### DIFF
--- a/src/fetch.ts
+++ b/src/fetch.ts
@@ -98,6 +98,7 @@ export async function fetchWithProxy(
   }
 
   if (proxyUrl) {
+    logger.debug(`Using proxy: ${sanitizeUrl(proxyUrl)}`);
     const agent = new ProxyAgent({
       uri: proxyUrl,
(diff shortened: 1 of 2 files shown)


src/fetch.ts
```diff
@@ -9,13 +9,26 @@ import logger from './logger';
 import invariant from './util/invariant';
 import { sleep } from './util/time';
 
+function sanitizeUrl(url: string): string {
+  try {
+    const parsedUrl = new URL(url);
+    if (parsedUrl.username || parsedUrl.password) {
+      parsedUrl.username = '***';
+      parsedUrl.password = '***';
```
vedantr, line 17: Quick thought, should we at least show the first/last character of the username and password? 

Could be helpful for debugging.

PR #2784, opened 2025-01-22
feat: pull and set sessionIds in the request and response body

sklein12:


src/providers/http.ts
```diff
@@ -139,6 +147,16 @@ export async function createSessionParser(
       `Response transform malformed: ${filename} must export a function or have a default export as a function`,
     );
   } else if (typeof parser === 'string') {
+    if (parser.startsWith('body.') || parser.startsWith('headers.')) {
+      return ({ headers, body }) => {
+        const [obj, ...keyParts] = parser.trim().split('.');
```
typpo: This approach is a little scary IMO.  I wonder if it would make sense to use [JSONPath](https://en.wikipedia.org/wiki/JSONPath) or just go with Javascript eval like the others?

sklein12: yea, I was debating that too. We could just drop this and just tell them to write JS code.

typpo: I think that might be best.  The downside here is JS is going to bite us one day from a security perspective.  JSONPath might be a compromise because it is more expressive but doesn't execute live code.

sklein12: I'll just do JS, it's the most flexible and we'll need to figure a bunch out in the future when the security bug bites us because we use it in the other parsers too.

PR #2839, opened 2025-01-27
chore: switch from stateless to stateful

sklein12:
Having stateless as a config name is really confusing and just bad manners. Switching to use Stateful as the config name.

1. Move stateful configuration onto the target so it's persisted and global instead of inferring it from strategy configs
2. Updated stateless -> stateful while maintaining backwards compat for now


sklein12:
@coderabbitai can you review this?

PR #2841, opened 2025-01-27
chore(providers): support chatgpt-4o-latest as a provider

vedantr:
- As the title says, also ensure that we do not explicitly fire a warning that this is an unknown OpenAI model.


test/providers.test.ts
```diff
@@ -61,6 +62,18 @@ describe('loadApiProvider', () => {
     expect(provider).toBeDefined();
   });
 
+  it('should load ChatGPT-4o-latest provider', async () => {
```
mldangelo: Let's skip this test. I don't think we need it.

vedantr: Fair point, I just had this in there to make sure it wasn't barfing.

mldangelo (approved):
Thanks!

PR #2842, opened 2025-01-27
chore(tests): add support for loading tests from JSONL files

mldangelo:
This update introduces support for JSONL files in test configurations.

- Updated documentation to reflect JSONL usage.
- Modified test case loader to handle JSONL format.


mldangelo:
@gru-agent add tests

PR #2848, opened 2025-01-27
chore: strategy probe calculations

sklein12:


typpo (dismissed):
extremely cool change

PR #2855, opened 2025-01-28
fix(sharing): combine sharing configuration from multiple promptfooconfigs

carl09:
When i tried using self hosted and sharing to my instance it still posted to the public server.

Follow the example here: https://www.promptfoo.dev/docs/usage/self-hosting/#pointing-promptfoo-to-your-hosted-instance

```yaml
sharing:
  apiBaseUrl: http://localhost:3000
  appBaseUrl: http://localhost:3000
```

I have made a change to allow the sharing property to be an object or a boolean


mldangelo (commented):
Thank you @carl09 for the find! We'll get this deployed tomorrow!

mldangelo (approved):

PR #2870, opened 2025-01-28
chore: update sqlInjection and shellInjection graders

MrFlounder:
They get better score in grader evals now.


mldangelo (approved):
Looks good!

PR #2884, opened 2025-01-29
chore(deps): move tsx to dev dependencies

mldangelo:
This PR updates the package-lock.json to correctly mark 'tsx' as a dev dependency.
- Ensures consistency between package.json and package-lock.json.
- No breaking changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,7 +56,6 @@
         "rfdc": "^1.4.1",
         "semver": "^7.6.3",
         "socket.io": "^4.8.1",
-        "tsx": "^4.19.2",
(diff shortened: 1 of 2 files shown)


aloisklink:
Hi @mldangelo, I'm not super aware of Promptfoo's codebase, but I think `tsx` might be necessary in `dependencies` for promptfoo to import configs/assertions/providers/metrics etc ending with a `.ts` or `.mjs` file extension: https://github.com/promptfoo/promptfoo/blob/d9909bc657ac1c3eb83fdb7300cf0f3a7cc419a1/src/esm.ts#L28

PR #2885, opened 2025-01-29
docs: misc cleanup

mldangelo:


site/docs/configuration/guide.md
```diff
@@ -61,11 +65,13 @@ tests:
...
 ```
 
-We can create additional tests. Let's add a couple other [types of assertions](/docs/configuration/expected-outputs). Use an array of assertions for a single test case to ensure all conditions are met.
+We can create additional tests. Let's add a couple other **[types of assertions](/docs/configuration/expected-outputs)**. Use an array of assertions for a single test case to ensure all conditions are met.
 
 In this example, the `javascript` assertion runs Javascript against the LLM output. The `similar` assertion checks for semantic similarity using embeddings:
 
-```yaml
```
typpo, line 73: I think this change just pollutes the examples.  Not sure we should do this

examples/redteam-deepseek/redteamconfig.yaml
```diff
@@ -1,6 +1,9 @@
 # yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
 
-description: 'Deepseek Censorship Redteam'
+description: 'DeepSeek Sensitive Topics Red Team'
+
+prompts:
+  - '{{prompt}}'
```
mldangelo, line 6:
```suggestion
```

PR #2888, opened 2025-01-29
fix(golang-provider): support internal package imports by preserving module structure

mldangelo:
This PR fixes the Golang provider to support packages imported from internal folders.
- Ensures module structure is preserved during script execution.
- Copies internal directories and module files to temporary directories.

### DIFF
--- a/src/providers/golangCompletion.ts
+++ b/src/providers/golangCompletion.ts
@@ -55,6 +55,7 @@ export class GolangProvider implements ApiProvider {
     apiType: 'call_api' | 'call_embedding_api' | 'call_classification_api',
   ): Promise<any> {
     const absPath = path.resolve(path.join(this.options?.config.basePath || '', this.scriptPath));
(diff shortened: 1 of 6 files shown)


src/providers/golangCompletion.ts
```diff
@@ -55,6 +55,7 @@ export class GolangProvider implements ApiProvider {
     apiType: 'call_api' | 'call_embedding_api' | 'call_classification_api',
   ): Promise<any> {
     const absPath = path.resolve(path.join(this.options?.config.basePath || '', this.scriptPath));
+    const moduleRoot = path.dirname(absPath); // Assuming the go.mod is in the same directory as the script
```
gregnuttall, line 58: hi @mldangelo , I believe this line has broken our evaluation script since our go.mod file is in a parent directory of the main.go file used to run promptfoo evaluations. I have raised an issue here: https://github.com/promptfoo/promptfoo/issues/3057

Thanks!

PR #2892, opened 2025-01-29
fix(webui): Make navigation bar sticky and fix content overlap

AISimplyExplained:
This PR fixes the navigation bar behavior by making it sticky at the top of the page and properly handling content overlap.

Changes:
- Changed AppBar position to fixed
- Added proper z-index and width to StyledAppBar
- Added padding to Root component to prevent content overlap
- Removed redundant margin from navigation bar

Before this change, the navigation bar would scroll out of view and content would overlap with it. Now it stays fixed at the top and all content is properly positioned below it.
Problem
image

Fix:
image


src/app/src/components/Navigation.css
```diff
@@ -6,6 +6,11 @@
   box-shadow: 0 4px 6px rgba(0, 0, 0, 0.1);
 }
 
+/* Add spacing for fixed navigation */
+main {
+  padding-top: 64px; /* Height of the navigation bar */
```
mldangelo: this may fix the redteam ui but removes too much margin on the eval view 
Screenshot 2025-01-29 at 9 22 51 AM

Can you please try to fix it a different way?

AISimplyExplained: Yes, have fixed and pushed the changes

src/app/src/components/PageShell.tsx
```diff
@@ -67,7 +68,20 @@ const lightTheme = createAppTheme(false);
 const darkTheme = createAppTheme(true);
 
 function Layout({ children }: { children: React.ReactNode }) {
-  return <div>{children}</div>;
+  const location = useLocation();
+  const isConfigPage = location.pathname.includes('/setup');
```
mldangelo: Would prefer a more general solution than this. Can we change the styles on the redteam setup ui instead?

sklein12: agreed

mldangelo (commented):
This is a great improvement to the redteam ui but makes the stick header on the eval view worse. Can you please take a look at that? You can compare it with https://local.promptfoo.app/eval - make sure you select "Sticky header" under table settings.

mldangelo:
Closing as stale.

PR #2897, opened 2025-01-29
chore: Move callEval outside of the function so we can re-use it 

sklein12:
We want to be able to call this function inside other providers and strategies.


mldangelo (commented):
for the love of god please add some tests

src/types/index.ts
```diff
@@ -123,18 +123,29 @@ export const OutputConfigSchema = z.object({
...
   test: AtomicTestCase;
   nunjucksFilters?: NunjucksFilterMap;
-  evaluateOptions: EvaluateOptions;
+  evaluateOptions?: EvaluateOptions;
 
   testIdx: number;
   promptIdx: number;
   repeatIndex: number;
+
+  conversations?: EvalConversations;
```
mldangelo, line 146: nit, may be worth documenting these types for open source (and my sanity)

sklein12:
@mldangelo your wish is my command

PR #2898, opened 2025-01-29
fix(webui): update Next button styling to support dark mode

AISimplyExplained:
Changes
- Update Next button styling to use theme colors instead of hardcoded values
- Fix button background and hover states in dark mode
- Use MUI's theme system for consistent styling

image

Before: 
The Next button in the config setup page had hardcoded colors that didn't adapt to dark mode.
image


mldangelo (approved):
Thank you!

PR #2902, opened 2025-01-30
chore: Save test case from EvalResult

sklein12:
Not sure why we were passing in the test case separately. The test case on the evaluateResult is just a reference to the test case we passed in.

We passed in the `evalStep` here:
https://github.com/promptfoo/promptfoo/blob/2caf61f043eb6d7a4851f1121947bd74b1c90e96/src/evaluator.ts#L698

Deconstructed in `callApi` arguments:
https://github.com/promptfoo/promptfoo/blob/2caf61f043eb6d7a4851f1121947bd74b1c90e96/src/evaluator.ts#L104

Pass it to `callApi` here: 
https://github.com/promptfoo/promptfoo/blob/2caf61f043eb6d7a4851f1121947bd74b1c90e96/src/evaluator.ts#L181C1-L182C1

Set it to `testCase` on the `EvaluateResult` Here:
https://github.com/promptfoo/promptfoo/blob/2caf61f043eb6d7a4851f1121947bd74b1c90e96/src/evaluator.ts#L228

But then pass it in to `addResult` from the `evalStep`?
https://github.com/promptfoo/promptfoo/blob/2caf61f043eb6d7a4851f1121947bd74b1c90e96/src/evaluator.ts#L745

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -742,7 +742,7 @@ class Evaluator {
       }
 
       try {
-        await this.evalRecord.addResult(row, evalStep.test);
+        await this.evalRecord.addResult(row);
       } catch (error) {
         logger.error(`Error saving result: ${error} ${safeJsonStringify(row)}`);
       }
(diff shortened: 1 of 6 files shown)


mldangelo (approved):
Looks good to me! Please fix the lint

mldangelo (approved):

PR #2903, opened 2025-01-30
chore: improvements to refusal detection

typpo:


src/redteam/util.ts
```diff
@@ -68,9 +138,6 @@ export function isEmptyResponse(response: string): boolean {
 
 export function isBasicRefusal(response: string): boolean {
   const lowerResponse = normalizeApostrophes(response.trim().toLowerCase());
-  if (lowerResponse.includes('\n')) {
-    return false;
-  }
```
mldangelo, line 73: slightly nervous about this one but I will trust you evaled it

mldangelo (approved):

src/redteam/util.ts
typpo: Anthropic in particular returns multi-paragraph refusals

PR #2908, opened 2025-01-30
feat(providers): add Alibaba Model Studio provider

mldangelo:
feat(providers): add Alibaba Model Studio provider

- Integrated Alibaba Cloud's DashScope API for Qwen language models.
- Added documentation and tests for the new provider.

No breaking changes.


src/providers/alibaba.ts
```diff
@@ -0,0 +1,77 @@
...
+  'qwen1.5-32b-chat',
+  'qwen1.5-14b-chat',
+  'qwen1.5-7b-chat',
+  // Embedding models
+  'text-embedding-v3',
+]);
+
+const API_BASE_URL = 'https://dashscope-intl.aliyuncs.com/compatible-mode/v1';
+
+export function createAlibabaProvider(
```
typpo: Not a fan of this `createXXXProvider` pattern.  It means there's an extra step when importing the providers to use them (e.g. in cloud), and it's not better in any way compared to making a class.  

Can we just implement the shared interface instead?

typpo (changes requested):

Issue #2901, opened 2025-01-30
[BUG] crescendo redteam provider doesn't transform query correctly


AetherPrior:
**Describe the bug**
I'm trying to configure my red-teaming config to send requests to an openai-compatible vLLM server hosting a llamaguard model, based on the [red-teaming configuration](https://www.promptfoo.dev/docs/red-team/configuration/) docs, and the [HTTP provider documents](https://www.promptfoo.dev/docs/providers/http/). I'm sending in a JSON request via my request body - but it appears that it isn't expected by promptfoo.  
Here's my red-teaming configuration: 

```
# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
description: My Red Team Configuration
targets:
  - id: http
    config:
      url: http://0.0.0.0:6969/v1/chat/completions
      method: post
      Content-Type: application/json
      body:
        model: /checkpoints/postsft_v2_beavertails-rlhf-dpo-filtered_8_epoch/
        messages: '{{prompt}}'
        max_tokens: 50
        temperature: 0.7
        top_p: 0.9
        'n': 1
        stream: false
        stop:
          - |+

      transformResponse: json.choices[0].message.content
    label: '''amzn-prize'''
redteam:
  provider:
    id: http
    config:
      url: http://0.0.0.0:1420/v1/chat/completions
      method: POST
      headers:
        Content-Type: application/json
      body:
        model: meta-llama/Llama-Guard-3-8b
        messages:
          - role: user
            content: '{{prompt}}'
        max_tokens: 50
        temperature: 0.7
        top_p: 0.9
        'n': 1
        stream: false
        stop:
          - |+

      transformRequest: 'file://request.js'
      transformResponse: (json, text) => json.choices[0].message.content

  plugins:
    - id: beavertails
  purpose: Evaluate and test language model responses to various prompts
  strategies:
    - id: crescendo
      config:
        maxTurns: 2
```

However, promptfoo throws me this error: 
```
Invalid options:
  Invalid output from extraction. Got: (json, text) => json.choices[0].message.content
```
Any guidance on what I'm doing wrong/what can I do would be appreciated!  

**To Reproduce**
Steps to reproduce the behavior, including example Promptfoo configurations if applicable:

1. create a config as above and save it as `redteam.yaml` in the CWD.
2. run `promptfoo redteam run -c redteam.yaml --verbose`

**Expected behavior**
[...]

PR #2917, opened 2025-01-30
fix(providers): handle function expressions in transform response

MrFlounder:
fix(http): handle function expressions in transform response

- Added support for parsing and executing function expressions in the transformResponse logic.
- Extended test coverage to validate various function expression scenarios.

This fixes the issuse reported in: https://github.com/promptfoo/promptfoo/issues/2901


mldangelo (approved):
Looks good!

Issue #2922, opened 2025-01-31
After upgrading to latest version 0.103.17 from 0.103.17, Unable to config defined in typescript


carl09:
**Describe the bug**
Unable to run Promptfoo configs written in Typescript since update to version 0.103.17

**To Reproduce**

1. define config in config.ts file:

```typescript
import type { UnifiedConfig } from 'promptfoo';

const config: UnifiedConfig = {
  description: 'Simple Test',
  prompts: ['Say hello to {{person}}'],
  providers: [
    {
      id: 'openai:gpt-4o-mini',
    },
  ],
  tests: [
    {
      vars: {
        person: 'carl',
      },
      assert: [],
    },
  ],
};

export default config;

```

2 run `npx promptfoo@latest eval -c ./config.ts`

**System information:**

```
Attempting to import module: {"resolvedPath":"/home/carl09/dev/testing/src/test/promptfoo-config.ts","moduleId":"test/promptfoo-config.ts"}
TypeScript/ESM module detected, importing tsx/cjs
ESM import failed: Error: Cannot find module 'tsx/cjs'
Require stack:
- /home/carl09/.npm/_npx/81bbc6515d992ace/node_modules/promptfoo/dist/src/esm.js
- /home/carl09/.npm/_npx/81bbc6515d992ace/node_modules/promptfoo/dist/src/util/index.js
- /home/carl09/.npm/_npx/81bbc6515d992ace/node_modules/promptfoo/dist/src/commands/cache.js
- /home/carl09/.npm/_npx/81bbc6515d992ace/node_modules/promptfoo/dist/src/main.js
Attempting CommonJS require fallback...
CommonJS require also failed: SyntaxError: Unexpected token '{'
file:///home/carl09/dev/testing/src/test/promptfoo-config.ts:1
import type { UnifiedConfig } from 'promptfoo';
            ^

SyntaxError: Unexpected token '{'
```

- Promptfoo version: 0.103.17

**Additional context**

I think this happened because the package `tsx` was moved to a devDependancy in this PR #2884

typpo:
Thanks for the fix!

mldangelo:
The fix has been released in [0.103.18](https://github.com/promptfoo/promptfoo/releases/tag/0.103.18) Please pull the latest version and let us know if it resolves your issue.

PR #2923, opened 2025-01-31
fix: added tsx back to dependencies

carl09:
Adding `tsx` package back to been a dependency for typescript execution ref: #2922


mldangelo (approved):
Thank you so much @carl09! I'll cut a release for you.

carl09:
Thank you @mldangelo, there is no rush on a release i am happy to work around for now, I also created an issue, i was not sure if it was part of a larger refactor

PR #2930, opened 2025-01-31
chore: make testcase generation more robust to malformed generations

typpo:


mldangelo:
What if we add structured outputs on the generation model?

mldangelo:
Closing as stale! Thanks @typpo!

PR #2935, opened 2025-01-31
fix(grader): reduce false positives in politics plugin

AISimplyExplained:
### Description
Modified the passing criteria in the politics plugin to better align with its core purpose of identifying political bias while reducing false positives.

### Change
Changed the evaluation logic from requiring all criteria to be met to requiring any one or more of the following criteria for passing.

### Why
The current prompt was introducing false positives of the nature where the AI provides a neutral/vague response but it doesn't redirect the question to authority (may not be necessary to redirect).


mldangelo (approved):
Thanks!

PR #2938, opened 2025-01-31
fix(webui): align settings icon to top right in strategy cards

AISimplyExplained:
### Problem
The settings icon in strategy cards (particularly in the "Single-shot Optimization" card) was not properly aligned and would overflow when the card content was long.

### Changes
Improved the layout of strategy cards by:
- Moving the settings icon to an absolute position in the top right corner of each card
- Adding proper positioning context to the card container
- Removing the space-between layout that was causing alignment issues

## Visual Changes
Before: The settings icon would be pushed out of alignment when the card title or content was long.
image

After: The settings icon is consistently positioned in the top right corner of each card, matching the design of the "Basic" card.
image


mldangelo:
@AISimplyExplained would you still like to land this?

PR #2940, opened 2025-01-31
chore(providers): add o3 mini as an option to OpenAI provider

vedantr:


src/providers/openai.ts
```diff
@@ -46,6 +46,13 @@ export const OPENAI_CHAT_MODELS = [
       output: 12 / 1e6,
     },
   })),
+  ...['o3-mini', 'o3-mini-2024-09-12'].map((model) => ({
```
joakimriedel: o3-mini-2025-01-31 ?

PR #2964, opened 2025-02-03
chore(docs): change docusaurus default port

vedantr:
Change the default port of the docusaurus server to be 3100 to not conflict w/ the PF dev server and obviating the need for the user to manually agree to a different port number


site/package.json
```diff
@@ -9,7 +9,7 @@
     "swizzle": "docusaurus swizzle",
     "deploy": "docusaurus deploy",
     "clear": "docusaurus clear",
-    "serve": "docusaurus serve",
+    "serve": "docusaurus serve --port 3100",
```
mldangelo, line 12: Did you mean to do this on start or serve or both? I recommend using `npm start` in development

vedantr: > Did you mean to do this on start or serve or both? I recommend using `npm start` in development

That's a good idea, I'll update it for both.

mldangelo (approved):

PR #2996, opened 2025-02-06
feat(providers): add support for multi-turn websockets with an example

vedantr:
- [ ] **TODO** Fix breaking tests + bug with `multi-turn conversation` feature, which causes tests to fail and the evaluator to hang


examples/websockets-conversation/promptfooconfig.js
```diff
@@ -0,0 +1,16 @@
...
+      // This provider will maintain WebSocket connection state between prompts
+      maintainConnectionBetweenCalls: true,
+      transformResponse: (response) => {
+        const parsed = JSON.parse(response);
+        return parsed.output || parsed.error || JSON.stringify(parsed);
+      },
+    },
+  ],
+  defaultProvider: 'websocket-chat',
+};
```
vedantr, line 16: NOTE: Need to update/remove this example, it is no longer relevant.

PR #3002, opened 2025-02-06
chore: making json parsing less strict for matchers

mshavliuk:
This is a proof of concept, which might be refined based on the discussion

## Problem statement

Quite often the language model used for `llm-rubric` returns a string which does not fully comply with the JSON syntax. Although this theoretically might be mitigated by the tool use / function calling configured for the prompt, to my knowledge this is not supported by promptfoo at the moment, at least for AWS Bedrock.

Some real examples:
```
{
  "reason": "``The summary accurately reflects...",
  "score": 1.0 // The summary is fully accurate and faithful
}
```

```
{
  "reason": "The summary provided is mostly accurate but contains some omissions and slight inaccuracies that do not significantly distort the main meaning. Here are the details:

- The summary correctly captures the essence ...
- It mentions the need to reveal system messages...
...

  ",
  "score": 0.75
}
```

## Proposed solution
I've evaluated several fail-tolerant JSON parsers:

| Library | Weekly Downloads | Features | GitHub Stars | Latest Release |
|---------|-----------------|----------|--------------|----------------|
| [jsonrepair](https://github.com/josdejong/jsonrepair) | 330k+ | missing quotes around keys, escape characters, commas, support single quotes,  JSONL, streaming, strip comments, JSONP notation, MongoDB data types | 615 | minutes ago |
| [dirty-json](https://github.com/RyanMarcus/dirty-json) | 60k+ | unquoted keys, single quotes, trailing commas, comments, line/column error messages in output, hexadecimal/octal numbers, parse JavaScript-style objects | 315 | 4 years ago |

I've decided to use `jsonrepair` due to its robust feature set, active maintenance, and widespread adoption.

The benefits include:
- More reliable llm-rubrics with virtually eliminated `Could not extract JSON from llm-rubric response` errors
- No need for custom JSON parsing logic or sanitization
- Comprehensive handling of common LLM output issues like unescaped quotes and comments

## Discussion Points
1. **Backward Compatibility**
Some unit tests fail with these changes because previously omitted object properties now become part of a "repaired" valid JSON object. For example, the string `no json here` becomes a valid `"no json here"` JSON string. These changes might break existing evals and code that relies on non-JSON outputs.

2. **`extractJsonObjects` Return Type**
The function currently always returns an array regardless of whether the LLM output is a JSON array/`jsonl` or not. With `jsonrepair`, we can correctly infer the underlying value type (array/object/string/number). Consider refactoring `extractJsonObjects` to:
- Return `any` instead of an array
- Delegate type checking responsibility to return value consumers

3. **Additional Considerations**
- Performance impact of additional JSON repair step?
- Documentation updates for new behavior?


mshavliuk:
> the tool use / function calling configured for the prompt, to my knowledge this is not supported by promptfoo at the moment

What I found is that while activating a "tool use" is possible with correct provider configuration, the generated output is not supported by the [`matchesLlmRubric`](https://github.com/promptfoo/promptfoo/blob/a8c145beffbe5f0c022ca8ae2d1122cf0122fa0b/src/matchers.ts#L404) function. For AWS Bedrock Nova model the returned output has the following form:
```json
{"name":"grade","toolUseId":"512ff780-0ae6-429f-bf8e-97c0d959379f","input":{"grade":0.5, "reason": "Some justification"}}
```
while the `matchesLlmRubric` expects `{"grade":0.5, "reason": "Some justification"}` 

But this is a topic for another PR 😏

mldangelo:
Thank you so much @mshavliuk. I switched this to yaml for parsing and verified that it passes over your test cases.

PR #3009, opened 2025-02-06
chore: expose `deleteFromCache` to evict cache keys after fetch by providers

typpo:
I am adding this because some openai-compatible apis do not return HTTP status codes that reflect the actual outcome (openrouter). This means we cache things like rate limit errors.

In theory this should be handled at the evaluator level, because we never want to cache an error for any provider.  But that would be a heavy refactor

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -123,8 +123,7 @@ export async function fetchWithCache<T = any>(
       // Don't cache if the parsed data contains an error
       if (format === 'json' && parsedData?.error) {
         logger.debug(`Not caching ${url} because it contains an 'error' key: ${parsedData.error}`);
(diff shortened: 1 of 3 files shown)


src/cache.ts
```diff
@@ -116,6 +120,12 @@ export async function fetchWithCache<T = any>(
         // Don't cache empty responses
         return;
       }
+      // Don't cache if the parsed data contains an error
+      if (format === 'json' && parsedData?.error) {
+        logger.debug(`Not caching ${url} because it contains an 'error' key: ${parsedData.error}`);
+        errorResponse = data;
+        return;
+      }
```
typpo: This is potentially controversial magic

PR #3012, opened 2025-02-06
ci: check for keys

typpo:


vedantr (approved):

mldangelo:
Github advanced security already does this for us on push!

typpo:
not worth it then

PR #3015, opened 2025-02-06
chore: more refusal detection

typpo:


mldangelo (approved):
may be time for a model graded refusal detection metric

PR #3026, opened 2025-02-07
fix(providers): correct golang behavior for prompts with quotes

mldangelo:


mldangelo:
@gregnuttall can you please try this implementation out?

PR #3033, opened 2025-02-09
chore(providers): handle edge case in openai chat completion provider where some models return tool calls with undefined instead of null content

stonyz:
When return tool_calls, some LLM may not return the message.content at all, so it's undefined, and can't be checked by message.content === null.


mldangelo (approved):
Thank you so much @stonyz!

PR #3054, opened 2025-02-11
fix(harmful): refine grader logic for specific categories

MrFlounder:
Refines the logic for determining harmful graders based on specific categories.
- Simplifies the assignment of assertion types.
- No breaking changes introduced.


src/redteam/plugins/harmful/common.ts
```diff
@@ -7,10 +7,7 @@ import {
 } from '../../constants';
 
 export function getHarmfulAssertions(harmCategory: keyof typeof HARM_PLUGINS): Assertion[] {
-  const assertionType: AssertionType =
-    harmCategory === 'harmful:privacy'
-      ? 'promptfoo:redteam:harmful:privacy'
-      : 'promptfoo:redteam:harmful';
+  const assertionType: AssertionType = `promptfoo:redteam:${harmCategory}`;
```
mldangelo, line 10: I can't believe we missed this. This is a good change but I don't think it will work for every grader given many harm categories don't have a specific grader. Can you rewrite it in such a way that it falls back to the generic `promptfoo:redteam:harmful` if we don't have the specific harm category?

MrFlounder: They will get mapped to default harmful I believe
I mean all the keyof typeof HARM_PLUGINS they all exist in graders mapping
The ones that doesn't have a specific grader they still existing in graders mapping but having the default harmful()

PR #3060, opened 2025-02-11
fix: broken handling of absolute paths in hook/code import

vedantr:
- Module imports were breaking on absolute paths due to an issue in how we were calling `path.resolve`. We didn't check whether or not the right-most path was already an absolute path
- Implement a new `safeJoin` and `safeResolve` method that does just this.

### DIFF
--- a/src/util/file.ts
+++ b/src/util/file.ts
@@ -1,7 +1,19 @@
 import path from 'path';
+import { fileURLToPath } from 'url';
 
 function isAbsolute(filePath: string): boolean {
-  return filePath.startsWith('file:///') || path.isAbsolute(filePath);
+  try {
+    // Handle both Windows and POSIX file URL formats
+    // Windows: file://C:/path or file:///C:/path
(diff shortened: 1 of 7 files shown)


src/util/file.ts
```diff
@@ -7,3 +9,33 @@
...
+ * If the last path is absolute or an absolute URL, returns it directly
+ *
+ * @param paths - The path segments to join
+ * @returns The joined path if last path is relative, or the last path if it's absolute
+ */
+export function safeJoin(...paths: string[]): string {
+  const lastPath = paths[paths.length - 1] || '';
+  if (lastPath.startsWith('file:///') || path.isAbsolute(lastPath)) {
+    return lastPath;
+  }
```
sklein12: This code is duplicated above, can it be combined into a function?

sklein12 (approved):
One nit comment. Otherwise looks good. Did you test it with some other use cases too?

PR #3062, opened 2025-02-12
chore(auth): synchronize email between `promptfoo config set email` and `promptfoo auth login`

mldangelo:
- Add email sync during login to handle existing email updates
- Add email unset on logout for consistent state management
- Add validation to prevent email updates while logged in

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -86,18 +86,28 @@ export function authCommand(program: Command) {
         }
         const { user } = await cloudConfig.validateAndSetApiToken(token, apiHost);
 
-        // Store token in global config
+        // Store token in global config and handle email sync
+        const existingEmail = getUserEmail();
+        if (existingEmail && existingEmail !== user.email) {
+          logger.info(
+            chalk.yellow(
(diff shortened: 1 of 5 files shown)


src/commands/auth.ts
```diff
@@ -106,8 +116,10 @@ export function authCommand(program: Command) {
     .description('Logout')
     .action(async () => {
       await cloudConfig.delete();
+      // Always unset email on logout
```
mldangelo, line 119: @sklein12 would appreciate your take on this

src/commands/config.ts
```diff
@@ -34,6 +35,14 @@ export function configCommand(program: Command) {
     .command('email <email>')
     .description('Set user email')
     .action(async (email: string) => {
+      if (cloudConfig.getApiKey()) {
+        logger.error(
+          "Cannot update email while logged in. Email is managed through 'promptfoo auth login'. Please use 'promptfoo auth logout' first if you want to use a different email.",
```
mldangelo, line 40: error message may be too verbose 
```suggestion
          "Cannot update email while logged in. Email is managed through 'promptfoo auth login'",
```

typpo:
nice

PR #3077, opened 2025-02-13
feat: add support for for tools in Vertex provider

kieranmilan:
feat: add support for for tools in Vertex provider

Add Tool[] block to the completion options + provide an example.
Also minor fix to avoid cache key collisions when calling the same prompt against two different Vertex models.


mldangelo:
Thank you so much @kieranmilan!

PR #3105, opened 2025-02-15
refactor(env): centralize environment variable schema

mldangelo:
Centralized the environment variable schema into a single module for consistency and maintainability.
- Removed duplicate schema definitions.
- Updated references to use the centralized schema.


typpo (changes requested):
This seems like the most useless zod schema ever?  should we just get rid of it?  because envars are always strings

mldangelo:
> This seems like the most useless zod schema ever? should we just get rid of it? because envars are always strings

I believe this is an improvement over the current code but am happy to iterate on it. Right now, ProviderEnvOverridesSchema is used in our (zod) TestSuiteSchema to limit which env variables can be specified / overridden in `promptfooconfig.yaml` Would be happy to consider allowing any env var to be overridden. It's (mildly) useful for things like autocomplete in the schema so I would also like to consider doing the opposite and adding all of EnvVars (src/envars.ts) as a Zod schema to TestSuiteSchema. Happy to defer to your direction.

mldangelo:
Spoke with typpo about this and merging.

PR #3108, opened 2025-02-15
fix: Clearer error message in testCaseReader, test cleanup

dhodun:
- fixing error message to both indicate this code path can be reached with .js and .ts files, as well as resolution hint
- removing unused mocks

Open to feedback on the error message text. But the current test is pretty explicitly checking for extra colons and would be nice to provide a clearer error.

I'd like to next add full support for .ts and .js to take function names with the same default function name, to be consistent with dynamic prompts. Should be simple fix, same implementation as provider, and adding tests.


typpo (approved):

typpo:
Thank you!

PR #3117, opened 2025-02-16
feat: extra information plugin

typpo:


AISimplyExplained:
Sample Run of this plugin:  https://www.promptfoo.app/eval/eval-HP2-2025-07-31T16:07:33?metric=ExtraInformation

src/redteam/plugins/base.ts
```diff
@@ -393,6 +394,7 @@ export abstract class RedteamGraderBase {
     test: AtomicTestCase,
     provider: ApiProvider | undefined,
     renderedValue: AssertionValue | undefined,
+    providerResponse: ProviderResponse | undefined,
```
sklein12, line 397: Why did you need to add this for this plugin?

AISimplyExplained: It was in Ian's original commit and I think he's using it to check the raw response data.

sklein12: ok, if we need it, we should do a couple things

1. The `llmOutput` argument is included in the provider response so we're passing the same data twice. We can remove the `llmoutput`
2. If we're going to change every call to this we should change this a single typed object argument instead of positional arguments.

mldangelo:
Closing as stale

PR #3126, opened 2025-02-18
chore: Enable SWC for ts-node for faster dev server

faizanminhas:
Running the local dev server was pretty slow, but enabling SWC seems to have sped it up by maybe 10x on my machine.


mldangelo (approved):
This is a great improvement! Thank you so much @faizanminhas!

PR #3127, opened 2025-02-18
refactor(providers): extract provider registry to dedicated module

mldangelo:
- Implement factory pattern for provider creation using ProviderFactory interface
- Improve error handling by replacing process.exit with thrown errors


src/providers/index.ts
```diff
@@ -0,0 +1,237 @@
...
+  HuggingfaceTokenExtractionProvider,
+} from './huggingface';
+import { LlamaProvider } from './llama';
+import {
+  LocalAiCompletionProvider,
+  LocalAiChatProvider,
+  LocalAiEmbeddingProvider,
+} from './localai';
+import { ManualInputProvider } from './manualInput';
+import { MistralChatCompletionProvider, MistralEmbeddingProvider } from './mistral';
```
typpo: Nice change. Instead of importing everything here _and_ in the registry, is there a one or two-line approach where the registry does the exports and we just pass them through?

PR #3133, opened 2025-02-18
fix: issue where script.py:myFunc fails fs stat check when PROMPTFOO_STRICT_FILES=true

dhodun:
- current behavior: `PROMPTFOO_STRICT_FILES=true` not compatible with python.py:MyFunc format.
- changing the parsing order so that `fs.statSync` hits the file without the function name
- this was largely masked before because `PROMPTFOO_STRICT_FILES` defaults to `false`

General comment: I ran an eval with PROMPTFOO_STRICT_FILES=false (default) and while I passed the `fs.statSync` check, the script still tried to load my python file and errored out, so I'm unclear on what the default behavior is intended to be.

### DIFF
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -1018,15 +1018,6 @@ export function parsePathOrGlob(
   }
   const filePath = path.resolve(basePath, promptPath);
 
-  let stats;
-  try {
-    stats = fs.statSync(filePath);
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Nice find! Thank you @dhodun!

PR #3141, opened 2025-02-19
fix: properly classify API errors with ResultFailureReason.ERROR

AISimplyExplained:
## Issue
When examining the code, I identified that API errors (like 400 Bad Request) were not being properly classified. The root cause was in the error handling logic:

1. In `src/types/index.ts`, we have the `ResultFailureReason` enum that defines different types of failures:
```typescript
export enum ResultFailureReason {
  NONE = 0,      // The test passed, or we don't know exactly why the test case failed
  ASSERT = 1,    // The test case failed because an assertion rejected it
  ERROR = 2,     // Test case failed due to some other error
}
```

2. In `src/evaluator.ts`, API errors (like 400 Bad Request) were being handled in the `if (response.error)` block, but this block was only setting the error message without setting `failureReason` to `ResultFailureReason.ERROR`:
```typescript
if (response.error) {
  ret.error = response.error;
  // Missing: ret.failureReason = ResultFailureReason.ERROR
}
```

3. The `failureReason` was only being set to `ResultFailureReason.ERROR` in the catch block at the end, which only handles exceptions thrown during evaluation, not API errors returned in the response.

## Changes
- Modified the error handling in `src/evaluator.ts` to properly set `failureReason = ResultFailureReason.ERROR` when an API error occurs in the response
- This ensures API errors are correctly classified and show up appropriately in both the "Errors" and "Failures" views

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -248,6 +248,7 @@ export async function runEval({
 
     if (response.error) {
       ret.error = response.error;
+      ret.failureReason = ResultFailureReason.ERROR;
     } else if (response.output === null || response.output === undefined) {
       // NOTE: empty output often indicative of guardrails, so behavior differs for red teams.
       if (isRedteam) {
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you @AISimplyExplained!

PR #3147, opened 2025-02-19
fix: Remove content length header

sklein12:
Remove the content length header before sending raw http request


src/providers/http.ts
```diff
@@ -756,6 +756,8 @@ export class HttpProvider implements ApiProvider {
       `${protocol}://${parsedRequest.headers['host']}`,
     ).toString();
 
+    delete parsedRequest.headers['content-length'];
```
mldangelo, line 760: Can you leave a note for why?  Is the request modified after the header is set?

mldangelo (approved):
Don't understand but LGTM

PR #3166, opened 2025-02-20
fix: handle openai guardrail responses to be a PASS

AISimplyExplained:
### Problem
When OpenAI flags content for potential policy violations (via `invalid_prompt` errors), these are treated as errors, instead it should be a fail for a redteam since the model technically refused to say anything harmful


src/evaluator.ts
```diff
@@ -246,68 +246,73 @@ export async function runEval({
 
     invariant(ret.tokenUsage, 'This is always defined, just doing this to shut TS up');
 
+    // First check if this is a guardrail response
+    if (response.guardrails?.flagged) {
```
typpo, line 260: I think we're missing a check for redteam here?

src/evaluator.ts
```diff
@@ -246,68 +246,73 @@ export async function runEval({
...
+    }
+
+    // Then check for errors
     if (response.error) {
+      ret.success = false;
+      ret.score = 0;
+      ret.failureReason = ResultFailureReason.ERROR;
       ret.error = response.error;
-    } else if (response.output === null || response.output === undefined) {
-      // NOTE: empty output often indicative of guardrails, so behavior differs for red teams.
```
typpo, line 264: do we definitely want to get rid of this behavior?

will-holley:
Hey @AISimplyExplained, I'm going to take over this PR!

PR #3174, opened 2025-02-21
chore(eval): pass eval-id to --filter-failing and --filter-errors-only eval flags

mldangelo:
Enhance the `filter-failing` and `filter-errors-only` commands to accept evaluation IDs or file paths for filtering. This improves flexibility and usability in test filtering workflows.


typpo (dismissed):
seems fine

PR #3233, opened 2025-02-26
fix(providers): Support inject vars in query params for raw requests for http provider

sklein12:
Spaces in query params need to be encoded properly. We weren't doing that since we are just injecting the prompts directly into the url.


mldangelo (approved):
Neat!

sklein12:
@coderabbitai review this please

PR #3250, opened 2025-02-28
chore: Add Foundational Model Reports links to Resources menu and footer

AISimplyExplained:


typpo (dismissed):
Nit: change to "foundation" model reports

PR #3255, opened 2025-03-01
chore: Fix an error in Google SpreadSheet(Authenticated) with a header without a value

kaibadash:
We got error:
```
TypeError: Cannot read properties of undefined (reading 'trim')
    at testCaseFromCsvRow (/path/to/node_modules/promptfoo/dist/src/csv.js:140:23)
    at /path/to/node_modules/promptfoo/dist/src/util/testCaseReader.js:203:51
    at Array.map (<anonymous>)
    at readStandaloneTestsFile (/path/to/node_modules/promptfoo/dist/src/util/testCaseReader.js:202:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async combineConfigs (/path/to/node_modules/promptfoo/dist/src/util/config/load.js:264:30)
    at async resolveConfigs (/path/to/node_modules/promptfoo/dist/src/util/config/load.js:423:22)
    at async runEvaluation (/path/to/node_modules/promptfoo/dist/src/commands/eval.js:136:55)
    at async doEval (/path/to/node_modules/promptfoo/dist/src/commands/eval.js:388:12)
```

This problem seems to occur in the following case.

- There is a header
- No value generated by function( `=IF('sheet'!A1, brabra, "")` )

`interface CsvRow { [key: string]: string; }` is not null.  But it has null because of `sheets_v4.Schema$ValueRange.values?: any[][] | null | undefined`.

### DIFF
--- a/src/googleSheets.ts
+++ b/src/googleSheets.ts
@@ -68,7 +68,7 @@ export async function fetchCsvFromGoogleSheetAuthenticated(url: string): Promise
   return dataRows.map((row) => {
     const csvRow: CsvRow = {};
     headers.forEach((header, index) => {
-      csvRow[header] = row[index];
+      csvRow[header] = row[index] ?? '';
     });
     return csvRow;
   });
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you!

PR #3268, opened 2025-03-03
docs: more bedrock multimodal docs

typpo:


site/docusaurus.config.js
```diff
@@ -175,7 +175,7 @@ const config = {
                 label: 'Docs',
               },
               {
-                href: 'https://promptfoo.dev/models/',
+                href: '/models/',
```
mldangelo, line 178: will break docusaurus

mldangelo (changes requested):

PR #3277, opened 2025-03-04
fix(providers): catch Vertex finish_reason errors correctly

kieranmilan:
Multi-part responses with a finish_reason error (e.g. MALFORMED_FUNCTION_CALL) on parts > 1 resulted in the errors being ignored. This fix ensures that they're caught correctly


mldangelo (approved):
Thank you!

PR #3285, opened 2025-03-05
chore(providers): integrate DeepSeek reasoning context into output

mldangelo:
Relates to https://github.com/promptfoo/promptfoo/issues/3280


MrFlounder:
I wonder if we are not grading on the reasoning + output? I think we should make sure only grade the final output

maybe reasoning is a good thing for guiding updating jailbreak prompts

typpo (changes requested):
Let's support the `showThinking` config option

PR #3288, opened 2025-03-05
fix: support @smithy/node-http-handler ^4.0.0

aloisklink:
Change the `peerDependencies` on `@smithy/node-http-handler` to support `^3.1.1 || ^4.0.0`, as that's the version used by the new AWS SDKs since v3.723.0.

The only breaking change in v4.0.0 seems to be dropping Node.JS v16 support, see [`@smithy/node-http-handler` CHANGELOG.md][1].

[1]: https://github.com/smithy-lang/smithy-typescript/blob/9ffe82d60dd45787a4279eb9723290edb39d0e10/packages/node-http-handler/CHANGELOG.md

---

`npm test` seems to work fine, but if anybody uses AWS Bedrock, it might be worth retesting the AWS Bedrock provider just to sanity check that the [`@smithy/node-http-handler` CHANGELOG.md][1] is correct and there are no other changes!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -141,7 +141,7 @@
         "@ibm-cloud/watsonx-ai": "^1.1.0",
         "@ibm-generative-ai/node-sdk": "^2.0.6",
         "@playwright/browser-chromium": "^1.47.2",
-        "@smithy/node-http-handler": "^3.1.1",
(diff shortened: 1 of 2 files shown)


mldangelo:
thanks @aloisklink! Really appreciate this + other recent PRs. Send me an email (michael@promptfoo.dev) and I'll send you some swag!

PR #3309, opened 2025-03-06
chore: update azure moderation doc

MrFlounder:


mldangelo (approved):
Don’t forget to document the magic number!

PR #3323, opened 2025-03-08
feat: add model-scan command

typpo:


src/main.ts
```diff
@@ -60,11 +61,12 @@ async function main() {
   deleteCommand(program);
   exportCommand(program);
   feedbackCommand(program);
-  const generateCommand = program.command('generate').description('Generate synthetic data');
```
mldangelo, line 63: just an fyi - this changes the help text display alphabetic order

mldangelo (dismissed):

PR #3337, opened 2025-03-11
fix: load unknown prompt files

typpo:


typpo:
closing because this may affect glob behavior

PR #3342, opened 2025-03-11
feat: convert to image strategy

typpo:


src/redteam/strategies/simpleImage.ts
```diff
@@ -0,0 +1,106 @@
...
+    // We're using SVG as an intermediate format as it's easy to generate without canvas
+    const svgImage = `
+      <svg width="800" height="400" xmlns="http://www.w3.org/2000/svg">
+        <rect width="100%" height="100%" fill="white"/>
+        <text x="50" y="50" font-family="Arial" font-size="16" fill="black">${escapeXml(text)}</text>
+      </svg>
+    `;
+
+    // Convert SVG to PNG using sharp
+    const pngBuffer = await sharp(Buffer.from(svgImage)).png().toBuffer();
```
mldangelo: instead of requiring sharp we could put this into our backend or even use a service like quickchart

PR #3345, opened 2025-03-11
feat: multimodal live sequential function calls

abrayne:
Enable continuation after a non-inbuilt function call, such that multiple functions can be run one after the other. To start with the functions will all have null return values, but it can then be extended with the mocked response functionality.

### DIFF
--- a/examples/google-multimodal-live-tools/promptfooconfig.yaml
+++ b/examples/google-multimodal-live-tools/promptfooconfig.yaml
@@ -28,19 +28,19 @@ tests:
         value: '{{location}}'
         threshold: 0.9
         transform: JSON.parse(output).toolCall.functionCalls[0].args.city
-  # Example of built-in google search tool that is executed by mm-live
-  - vars:
-      query: google search for why the sea is salty
-    assert:
-      - type: icontains
-        value: salt
-      - type: icontains-any
-        value:
-          - 'weathering'
-          - 'rainwater'
-  # Example of built-in tool code execution that is executed by mm-live
-  - vars:
-      query: calculate the result of multiplying 1341 by 23
-    assert:
-      - type: icontains
-        value: 30843
+  # # Example of built-in google search tool that is executed by mm-live
+  # - vars:
+  #     query: google search for why the sea is salty
+  #   assert:
+  #     - type: icontains
+  #       value: salt
+  #     - type: icontains-any
+  #       value:
+  #         - 'weathering'
+  #         - 'rainwater'
+  # # Example of built-in tool code execution that is executed by mm-live
+  # - vars:
+  #     query: calculate the result of multiplying 1341 by 23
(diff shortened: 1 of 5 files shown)


abrayne:
@mldangelo the only thing I'm unsure about is separating the output into text and toolCall. It would be nicer to have a separate way to assert on tool calls that relied on a different variable to output? Anything you'd recommend?

mldangelo:
This looks great! We'll get it in the next release (later today)

mldangelo:
> @mldangelo the only thing I'm unsure about is separating the output into text and toolCall. It would be nicer to have a separate way to assert on tool calls that relied on a different variable to output? Anything you'd recommend?

In other providers like openai/chat.ts we adjust the output dynimcally depending on whether or not there's a tool call. For example:

``` ts 
      let output = '';
      if (message.reasoning) {
        output = message.reasoning;
      } else if (message.content && (message.function_call || message.tool_calls)) {
        if (Array.isArray(message.tool_calls) && message.tool_calls.length === 0) {
          output = message.content;
        } else {
          output = message;
        }
      } else if (
        message.content === null ||
        message.content === undefined ||
        (message.content === '' && message.tool_calls)
      ) {
        output = message.function_call || message.tool_calls;
      } else {
        output = message.content;
      }
```

So far I think you are the only person using this provider so you can break it if you'd like. IMHO it would probably be better if the format was consistent.

abrayne:
> In other providers like openai/chat.ts we adjust the output dynimcally depending on whether or not there's a tool call. For example:
> 
> ```ts
>       let output = '';
>       if (message.reasoning) {
>         output = message.reasoning;
>       } else if (message.content && (message.function_call || message.tool_calls)) {
>         if (Array.isArray(message.tool_calls) && message.tool_calls.length === 0) {
>           output = message.content;
>         } else {
>           output = message;
>         }
>       } else if (
>         message.content === null ||
>         message.content === undefined ||
>         (message.content === '' && message.tool_calls)
>       ) {
>         output = message.function_call || message.tool_calls;
>       } else {
>         output = message.content;
>       }
> ```
> 
> So far I think you are the only person using this provider so you can break it if you'd like. IMHO it would probably be better if the format was consistent.

Thank you, I spotted that, but wanted to check there wasn't anything else.

The reason that this format isn't ideal is that this provider should enable multiple function calls (sequentially or in parallel) and the output of them should be used by the LLM to produce a final textual response. It would be nice to both place assertions on the function calls that should have happened, as well as the textual result produced.

I think this paradigm of testing will be useful universally across providers and should maybe be a higher level addition to /Users/brayne/Documents/customer_success/promptfoo/src/types/providers.ts:ProviderResponse that is reachable from inside the assertions? Some providers already enable (mocked) function execution, so it would be useful for them too if they plan to test chain function calls etc.

PR #3356, opened 2025-03-12
fix(webui): fix missing plugins in report view 

mldangelo:
Uses metadata.strategyId now that it's been in the testcase configs for months


mldangelo:
@coderabbitai review

Issue #3360, opened 2025-03-12
Not all models from AWS Bedrock work as providers


PavelPolyakov:
**Describe the bug**
Not all models from `bedrock` work.
For example:
```
us.deepseek.r1-v1:0
meta.llama3-2-3b-instruct-v1:0
```
fail

**To Reproduce**
This is my config, I just tried to run against number of random models.
All models were enabled on my AWS account.
```yaml
description: 'Getting started'
prompts:
   - "What is 1+1?"

providers:
  - id: bedrock:us.anthropic.claude-3-5-sonnet-20241022-v2:0
    config:
      profile: 'bedrock'
      region: 'us-east-1'
  - id: bedrock:us.anthropic.claude-3-7-sonnet-20250219-v1:0
    config:
      profile: 'bedrock'
      region: 'us-east-1'
  - id: bedrock:us.deepseek.r1-v1:0
    config:
      profile: 'bedrock'
      region: 'us-east-1'
  - id: bedrock:meta.llama3-2-3b-instruct-v1:0
    config:
      prfile: 'bedrock'
  - id: bedrock:amazon.nova-pro-v1:0
    config:
      profile: 'bedrock'

tests:
  - assert:
      - type: contains
        value: '2'
```

**Expected behavior**
All models work

**Screenshots**

<img width="1272" alt="Image" src="https://github.com/user-attachments/assets/c7b46be6-24a5-48bc-ad8c-48a6c700d655" />

**System information:**

```
{
  "version": "0.106.3",
  "platform": {
    "os": "darwin",
    "release": "24.3.0",
    "arch": "arm64",
    "nodeVersion": "v18.19.0"
  },
  "env": {},
  "configInfo": {
    "defaultConfigPath": "/%REDACTED%/promptfoo-test/promptfooconfig.yaml",
    "configExists": true,
    "configContent": {
      "config": {
        "tags": {},
        "description": "Getting started",
        "prompts": [
          "What is 1+1?"
        ],
        "providers": [
          {
            "id": "bedrock:us.anthropic.claude-3-5-sonnet-20241022-v2:0",
            "config": {
              "profile": "bedrock",
              "region": "us-east-1"
            }
          },
          {
            "id": "bedrock:us.anthropic.claude-3-7-sonnet-20250219-v1:0",
            "config": {
              "profile": "bedrock",
              "region": "us-east-1"
            }
          },
          {
            "id": "bedrock:us.deepseek.r1-v1:0",
            "config": {
[...]

PR #3362, opened 2025-03-12
chore(providers): add missing bedrock models

typpo:
#3360


src/providers/bedrock.ts
```diff
@@ -765,6 +765,7 @@ export const AWS_BEDROCK_MODELS: Record<string, IBedrockModel> = {
   'meta.llama3-1-8b-instruct-v1:0': BEDROCK_MODEL.LLAMA3_1,
   'meta.llama3-70b-instruct-v1:0': BEDROCK_MODEL.LLAMA3,
   'meta.llama3-8b-instruct-v1:0': BEDROCK_MODEL.LLAMA3,
+  'meta.llama3-2-3b-instruct-v1:0': BEDROCK_MODEL.LLAMA3_2,
```
mldangelo: Are you sure this model is available as is? If so we should probably add all of the other llama 3.2s and 3.3s as well. 

Related: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html

PR #3386, opened 2025-03-13
refactor(providers): centralize and streamline default provider configurations

mldangelo:
This refactor centralizes the logic for loading default providers, aligning it with their respective implementations.

- Moved default provider configuration logic to individual provider files.
- Updated provider selection logic to dynamically load configurations.

No breaking changes introduced.


src/providers/defaults.ts
```diff
@@ -60,114 +36,135 @@ export async function setDefaultEmbeddingProviders(provider: ApiProvider) {
   defaultEmbeddingProvider = provider;
 }
 
+/**
+ * Helper to check if a key is set in environment or env overrides
+ * Exported for testing
+ */
+export function isKeySet(key: keyof EnvOverrides, env?: EnvOverrides): boolean {
+  return !!getEnvString(key) || !!env?.[key];
```
typpo: consider moving this to env.ts

src/providers/defaults.ts
```diff
@@ -60,114 +36,135 @@ export async function setDefaultEmbeddingProviders(provider: ApiProvider) {
...
+}
+
+/**
+ * Simplified provider configuration with credential checking
```
typpo: bad comment

src/providers/bedrock/defaults.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// Use Claude model through Bedrock for LLM rubric evaluation
+export const DefaultLlmRubricProvider = new AwsBedrockCompletionProvider(
+  'anthropic.claude-3-7-sonnet-20250219-v1:0',
+);
+
+/**
+ * AWS Bedrock provider configuration
+ */
+export const BedrockProviderConfig: ProviderConfiguration = (env) => {
```
typpo: This change was hard to understand in part because `XYZProviderConfig` is not a descriptive name.  Can we change to `getDefaultProviders` or similar?

src/providers/defaults.ts
```diff
@@ -60,114 +36,135 @@ export async function setDefaultEmbeddingProviders(provider: ApiProvider) {
...
+
+  /** Provider configuration function */
+  config: ProviderConfiguration;
+
+  /** Check if this provider's credentials are available */
+  hasCredentials(env?: EnvOverrides): boolean | Promise<boolean>;
+}
+
+// Exported for testing
+export const PROVIDERS: ProviderEntry[] = [
```
typpo, line 57: add a note that order matters.  so you don't come back in 1 year and make it alphabetical

typpo (commented):
are you sure this approach in better?  i found it harder to understand

PR #3397, opened 2025-03-14
chore: PROMPTFOO_INSECURE_SSL true by default

typpo:


mldangelo (commented):
looks good if ci passes

sklein12 (approved):

PR #3402, opened 2025-03-15
chore(providers): migrate OpenAI assistant API to use fetch instead of SDK

mldangelo:
This PR updates the OpenAI assistant provider to use fetch for API calls instead of the OpenAI SDK.

- Removed OpenAI SDK dependency.
- Implemented fetch-based API calls for assistant functionalities.

No breaking changes.


typpo (approved):
cool change assuming it works!

mldangelo:
Closing as I think this is no longer needed.

PR #3408, opened 2025-03-16
feat(assertions): enhance factuality assertion handling and examples

mldangelo:
Enhance the factuality assertion mechanism to support JSON responses and improve documentation.

- Added support for JSON response parsing in factuality assertions.
- Updated examples and documentation to reflect new capabilities.
- Improved test coverage for factuality assertion scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,5 +30,5 @@ __pycache__
 
 # Examples-related
 examples/**/package-lock.json
+examples/huggingface-dataset-factuality/.cache
 examples/redteam-dalle/images
-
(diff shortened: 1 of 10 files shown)


mldangelo:
> This assertion type is specifically an implementation of https://github.com/openai/evals/blob/main/evals/registry/modelgraded/fact.yaml
> 
> I'm not sure we should change it unless we think it's really not working well.

I cover this many times in the docs in this PR. It regularly breaks for providers like anthropic and meta. I included a hotpot QA example that demonstrates how the new factuality prompt works better (run the example on main and in this branch). We can go over it together.

PR #3410, opened 2025-03-16
fix(workflow): temporarily disable redteam-custom-enterprise-server job

mldangelo:


.github/workflows/main.yml
```diff
@@ -295,8 +295,8 @@ jobs:
           npm run test:redteam:integration
 
   redteam-custom-enterprise-server:
-    name: Redteam Custom Enterprise Server
-    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    # FIXME(mldangelo): 2025-03-16: Skip until we replace staging auth
```
mldangelo, line 298:
```suggestion
    # FIXME(mldangelo): 2025-03-16: Skip until we replace staging auth
    name: Redteam Custom Enterprise Server
```

Issue #3358, opened 2025-03-12
if using prompt function enable skipping random nunjacks rendering


billybonks:
**Is your feature request related to a problem? Please describe.**
i am populating variables myself in my prompt function so i have no need nunjucks to get involved

currently it throws an error and adds no value as far i am concerned in terms of features used

```
//nodejs/22.14.0/lib/node_modules/promptfoo/node_modules/nunjucks/src/lib.js:32
    err = new Error(old.message);
          ^

Template render error: (unknown path) [Line 14, Column 50]
  expected name as lookup value, got *
    at Object._prettifyError (//nodejs/22.14.0/lib/node_modules/promptfoo/node_modules/nunjucks/src/lib.js:32:11)
    at Template.render (//nodejs/22.14.0/lib/node_modules/promptfoo/node_modules/nunjucks/src/environment.js:442:21)
    at Environment.renderString (//nodejs/22.14.0/lib/node_modules/promptfoo/node_modules/nunjucks/src/environment.js:313:17)
    at renderPrompt (//nodejs/22.14.0/lib/node_modules/promptfoo/dist/src/evaluatorHelpers.js:252:56)
    at async runEval (//nodejs/22.14.0/lib/node_modules/promptfoo/dist/src/evaluator.js:143:28)
    at async processEvalStep (//nodejs/22.14.0/lib/node_modules/promptfoo/dist/src/evaluator.js:648:26)
    at async //nodejs/22.14.0/lib/node_modules/promptfoo/dist/src/evaluator.js:859:11
```

this is the error thrown, i solved it by putting  an early return in the code

<img width="843" alt="Image" src="https://github.com/user-attachments/assets/109af59b-dbbf-454d-95b6-cd7941dfd485" />

https://github.com/promptfoo/promptfoo/blob/main/src/evaluatorHelpers.ts#L177

happy to make a pr with the correct guidance as how you would do it? i feel like i might be missing some complexity as well given that a few lines lower i see a function called resolveVars, but i already used them in the function so 🤷🏿

billybonks:
https://gist.github.com/billybonks/4c35b734c7bf4b7b41402694faf67138

PR #3429, opened 2025-03-19
feat: enable skipping nunjucks rendering

billybonks:
Ideally prompts are tested with the actual files that are used in production, for that reason I render the prompt in my prompt function using my own templating tools

Originally I just let it flow through nunjucks afterwards because it didn't break anything, but my prompt embeds user code and that causes nunjucks to explode. This change enables users to easily skip the rendering process if required.

more context and example are in the issue https://github.com/promptfoo/promptfoo/issues/3358


billybonks:
closing since there is already an env var defined

PR #3460, opened 2025-03-21
chore(content): update user statistics

mldangelo:
Update user statistics across the website content to reflect the current number of users as 70,000+.
- Adjusted text in multiple pages to reflect the updated user count.


typpo (approved):
nice

PR #3477, opened 2025-03-24
docs: Ensure consistent redteam flag usage in guides

will-holley:
This is a slightly opinionated PR.

Preferred `--no-gui` over GUI as the GUI's "Load Example" exposes the user to more complexity in a single pass (more form fields / visible steps) and the outputted config takes longer to run.

Conversely, `--no-gui` provides suitable defaults which are hidden from the user (until the user is interested in _how_ a redteam is configured and navigates to the generated `promptfooconfig.yaml`) and the generated configuration has a shorter runtime i.e. faster time to results.

Also removes previously deprecated `--no-interactive` flag which will break invocation.


site/docs/guides/llm-redteaming.md
```diff
@@ -54,23 +54,25 @@ First, install [Node 18 or later](https://nodejs.org/en/download/package-manager
...
 Run the eval:
 
-```
+```sh
+cd my-redteam-project
 npx promptfoo@latest redteam run
 ```
 
-This will create a file `redteam.yaml` with adversarialtest cases and run them through your application.
+This will create a file `redteam.yaml` with adversarial test cases and run them through your application.
```
mldangelo, line 75: NICE!

mldangelo (approved):

site/docs/guides/llm-redteaming.md
typpo: **NICE**

PR #3478, opened 2025-03-25
chore(redteam): Add a PlinyGrader to more accurately grade Pliny results

faizanminhas:
Previously, Pliny test cases would fail if the LLM didn't understand or acknowledge the prompt, which should probably be a pass, as it's a soft refusal. This change adds a new Pliny grader which should more accurately address those cases as pass rather than fail. 

Also fixes a minor bug in which cybersecval didn't have a metric name and would show up as undefined in the eval table.

### DIFF
--- a/src/redteam/plugins/pliny.ts
+++ b/src/redteam/plugins/pliny.ts
@@ -75,8 +75,6 @@ export class PlinyPlugin extends RedteamPluginBase {
       },
       assert: this.getAssertions(text),
       metadata: {
-        // Store the original text separately in metadata
-        jailbreakContent: text,
         pluginId: 'pliny',
       },
     }));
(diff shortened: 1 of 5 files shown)


src/redteam/plugins/pliny.ts
```diff
@@ -73,6 +74,59 @@ export class PlinyPlugin extends RedteamPluginBase {
         [this.injectVar]: ' ' + text,
       },
       assert: this.getAssertions(text),
+      metadata: {
+        // Store the original text separately in metadata
+        jailbreakContent: text,
```
typpo: Is this actually used anywhere?

faizanminhas: Good catch, I think this was leftover from another bugfix attempt

mldangelo (approved):
Thank you!

PR #3500, opened 2025-03-26
feat: google live mock stateful api

abrayne:
This enables stateful apis to be tested in promptfoo.

In future, it may be simply extended and tested with external APIs, providing purely the url and no python file.


mldangelo:
@abrayne I'll spend some time over the weekend QA'ing this! Thank you so much.

mldangelo:
Thanks so much @abrayne!

PR #3522, opened 2025-03-28
Log all reasons in geval in case of multiply assert values

schipiga:
It fixes the issue, that if you have a test case with array of values in `g-eval` assertion, and the test is passed, then in web-server test history you don't see reasons and can't understand, why llm decided, that test is passed. It fixes this problem

Screenshot 2025-03-28 at 11 58 31


mldangelo:
Thank you!

PR #3549, opened 2025-03-31
refactor: google types

abrayne:
Share types between ai studio, live and vertex apis.

The only non-trivial change is "stopSequence" -> "stopSequences" in vertex and live. The live code was copied across from vertex and as far as I can see from the docs that should have originally been "stopSequences", so it was a typo that this change fixes.

See:
https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/content-generation-parameters#stop-sequences


mldangelo (approved):
Thank you so much!

PR #3554, opened 2025-03-31
feat(evals): Adds evals index page

will-holley:
## Features

image

## Changes

Use the shared data grid component for the eval selector.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -167,6 +167,7 @@
     "@anthropic-ai/sdk": "^0.39.0",
     "@apidevtools/json-schema-ref-parser": "^11.9.3",
     "@googleapis/sheets": "^9.6.0",
+    "@mui/x-data-grid": "^7.28.2",
     "@types/ws": "^8.18.0",
     "ajv": "^8.17.1",
     "ajv-formats": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -167,7 +167,6 @@
     "@anthropic-ai/sdk": "^0.39.0",
     "@apidevtools/json-schema-ref-parser": "^11.9.3",
     "@googleapis/sheets": "^9.6.0",
-    "@mui/x-data-grid": "^7.28.2",
     "@types/ws": "^8.18.0",
     "ajv": "^8.17.1",
     "ajv-formats": "^3.0.1",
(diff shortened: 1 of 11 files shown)


src/app/src/components/Navigation.tsx
```diff
@@ -129,7 +129,7 @@ export default function Navigation({
           <NavSection>
             <Logo />
             <Dropdown />
-            <NavLink href="/eval" label="Evals" />
+            <NavLink href="/evals" label="Evals" />
```
mldangelo: Should think through the default behavior when we open promptfoo view and whether or want we want /eval to remain in the top nav. This may confuse people

package.json
```diff
@@ -167,6 +167,7 @@
     "@anthropic-ai/sdk": "^0.39.0",
     "@apidevtools/json-schema-ref-parser": "^11.9.3",
     "@googleapis/sheets": "^9.6.0",
+    "@mui/x-data-grid": "^7.28.2",
```
mldangelo: should be in src/app/package.json instead

mldangelo (commented):
This looks really good! Consider asking the llm nicely to make it prettier. 

One note on changing the navbar link. 

Are we also going to update prompts, history, and datasets pages (doesn't have to be in this PR)

src/app/src/components/Navigation.tsx
will-holley: I think it makes sense that `promptfoo view` opens `http://localhost:15500/eval` which loads/renders the most recent eval.

Are you thinking that clicking "Evals" in the top nav and then clicking the first eval in the data grid in order to access the most recent eval isn't intuitive? The current design – where the user sees a detailed view of what the eval entails and how it performed – is definitely more information rich than the data grid.

mldangelo: I think we should make this a drop down that has the latest eval and the evals page if not adding a second button. 

Thinking through use cases - most people that run promptfoo locally will sit on /eval and we have the magic websocket auto refresh - I want to continue to have a link to that page. With the list everything is a hard link meaning selecting the most recent from the top of the list will not use the auto refresh behavior

PR #3557, opened 2025-03-31
fix: correct tool count

typpo:


mldangelo (approved):
How do we test this and make sure it wont break again?

PR #3566, opened 2025-04-01
feat: add skip generation

MrFlounder:
This pr adds skip generation option when running redteam, this is needed when re-run failed tests from on-prem.


mldangelo (approved):
NICE

typpo (commented):
nice

PR #3567, opened 2025-04-01
refactor: rename vertexUitl to util and google provider to ais provider

abrayne:


mldangelo (approved):
Thanks again! Really appreciate all of your hard work in cleaning this up.

abrayne:
Thanks @mldangelo! I'm hoping if I unify the 2/3 providers it will be easier to maintain all of their overlapping features in parallel. AIS and Vertex in particular should be quite similar. Live has some added complexity.

mldangelo:
> Thanks @mldangelo! I'm hoping if I unify the 2/3 providers it will be easier to maintain all of their overlapping features in parallel. AIS and Vertex in particular should be quite similar. Live has some added complexity.

This is a huge help! Thank you

PR #3569, opened 2025-04-01
chore: link URLs in metadata

### LINKED ISSUE #3568: Hyperlinks for full URLs in metadata drop down
I have several hyperlinks as metadata fields in my TestCases. I'd like them to be clickable

Recently got easy copy+paste support which is great: https://github.com/promptfoo/promptfoo/issues/3376

The ability to click hyperlinks in these metadata fields which then opens in a new tab (think jumping to a historical in-app conversation that was created by my private provider API call).

Currently copying + pasting in a new tab URL which is enabled by the above feature, but this would be easier.

I understand this may be counter to providing raw metadata fields, as you're adding some additional formatting. But does improve end-user experience, hopping from promptfoo to an application.

typpo:
Closes #3568


mldangelo (approved):
Thanks!

PR #3578, opened 2025-04-02
docs: update docs around google tools and rename multimodal live

abrayne:
Multimodal live api has been rebranded to live.


mldangelo (approved):
Thank you so much @abrayne! Should we rename `examples/google-multimodal-live`  to `examples/google-live` as well?

abrayne:
Yeah any code mentioning multimodal or mm should be changed too. Thought it might be better to do that separately in a non-docs PR.

PR #3580, opened 2025-04-02
feat(evals table): Pass rate column

will-holley:
- Removes 100 result limit
- Fixes issue where data grid height would exceed page height


src/models/eval.ts
```diff
@@ -502,48 +503,72 @@ export default class Eval {
...
+   *
+   * - Test statistics are derived from the prompt metrics as this is the only reliable source of truth
+   * that's written to the evals table.
+   */
+  return results.map((result) => {
+    const passCount =
+      result.prompts?.reduce((memo, prompt) => {
+        // TODO(will): Verify whether prompt.metrics *should* be required.
+        invariant(!!prompt.metrics, 'Prompt metrics are required');
+        return memo + prompt.metrics!.testPassCount;
```
mldangelo, line 540: don't use the non-null operation, invariant should be able to catch this

src/models/eval.ts
```diff
@@ -502,48 +503,72 @@ export default class Eval {
...
+  /**
+   * Deserialize the evals. A few things to note:
+   *
+   * - Test statistics are derived from the prompt metrics as this is the only reliable source of truth
+   * that's written to the evals table.
+   */
+  return results.map((result) => {
+    const passCount =
+      result.prompts?.reduce((memo, prompt) => {
+        // TODO(will): Verify whether prompt.metrics *should* be required.
```
mldangelo, line 538: can you do this and remove the comment or remove the invariant?

PR #3597, opened 2025-04-05
chore(webui): update provider selector in create eval page 

mldangelo:
Removes old models, adds new ones, and makes misc formatting and style improvements to the page. 

I tested at least one model from every provider


typpo:
Thanks for doing this

PR #3603, opened 2025-04-07
fix: Generating a dataset without a `providers` key in the configuration

will-holley:
Fixes a bug where the `providers` key is required when generating datasets.

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -895,6 +895,10 @@ export const UnifiedConfigSchema = TestSuiteConfigSchema.extend({
 
 export type UnifiedConfig = z.infer<typeof UnifiedConfigSchema>;
 
+export enum ConfigType {
+  DatasetGeneration,
+}
+
 export interface EvalWithMetadata {
   id: string;
   date: Date;
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -895,10 +895,6 @@ export const UnifiedConfigSchema = TestSuiteConfigSchema.extend({
 
 export type UnifiedConfig = z.infer<typeof UnifiedConfigSchema>;
 
-export enum ConfigType {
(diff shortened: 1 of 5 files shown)


src/util/config/load.ts
```diff
@@ -536,7 +542,11 @@ export async function resolveConfigs(
     process.exit(1);
   }
 
-  if (!hasProviders) {
+  if (
```
mldangelo, line 544: please add test coverage for this! How are we not throwing during  `invariant(Array.isArray(config.providers), 'providers must be an array');` below?

src/types/index.ts
```diff
@@ -895,6 +895,10 @@ export const UnifiedConfigSchema = TestSuiteConfigSchema.extend({
 
 export type UnifiedConfig = z.infer<typeof UnifiedConfigSchema>;
 
+export enum ConfigType {
```
mldangelo: prefer string union or as const to an enum

src/util/config/load.ts
will-holley: https://github.com/promptfoo/promptfoo/blob/a08fc4e4da6f64e529e9d6350f307a7e59c82f25/src/util/config/load.ts#L498

whereas `hasProviders` is true when the `.length > 0`.

src/types/index.ts
mldangelo: sorry after thinking about this, I would prefer if we did not define a type for it and just made the resolveConfigs new argument of type const or string union. There's no good reason to expose this to the package and if we were to define config type it should probably also include redteam

src/util/config/load.ts
mldangelo: Ahah, ` || []` is bad

PR #3608, opened 2025-04-07
feat: Eval sharing idempotence

will-holley:


src/share.ts
```diff
@@ -403,17 +434,32 @@ export async function createShareableUrl(
...
+    if (cloudConfig.isEnabled()) {
+      headers['Authorization'] = `Bearer ${cloudConfig.getApiKey()}`;
+    }
+
+    const response = await fetchWithProxy(`${url}/${eval_.id}`, {
+      method: 'GET',
+      headers,
+    });
+
+    return response.status !== 404;
```
mldangelo: Would like Chief @sklein12 to take a look at this. 

I think I would prefer a new endpoint over checking the status. 

Sharing should probably become an update operation if it's already been shared instead of blocking it.

will-holley: What update is necessary?

mldangelo: For when people grade evals and re-share which is a not uncommon use case

sklein12:
@will-holley why did you decide to do it on the client side and not server side?

src/share.ts
```diff
@@ -403,17 +434,67 @@ export async function createShareableUrl(
...
 
-  return showAuth ? fullUrl : stripAuthFromUrl(fullUrl);
+    if (cloudConfig.isEnabled()) {
+      headers['Authorization'] = `Bearer ${cloudConfig.getApiKey()}`;
+    }
+
+    const res = await fetchWithProxy(`${url}/${eval_.id}`, {
+      method: 'GET',
+      headers,
+    });
```
sklein12, line 458: we have a utility to call the cloud API, can you please use that?

will-holley: What function is that? I copied `fetchWithProxy` from `sendEvalResults` defined above.

sklein12:
It seems like you needed to do that anyways. I'm just wondering if there's
a simpler way than needing to maintain the shared state locally. Like if
you push an existing eval it just overwrites what's on the cloud. Maybe
give a warning in the CLI "This already exists in the cloud, do you want to
overwrite it?"

PR #3627, opened 2025-04-08
chore(evals): If results have >5 variable columns, default to displaying first 5

will-holley:
- If results have >5 variable columns, default to displaying the first 5
- Cleans up the Eval state initialization logic
  - Separation of concerns (data fetching, data fetching effects, state population)
  - Simpler, reactive state updating
- Removes IndexedDB backing from eval results store
  - State is currently re-initialized from the server on mount
  - Can contain stale state (i.e. constructed by deprecated logic)
  - This combination results in difficult-to-trace state reconciliation edge cases
- Removes deprecated props


src/app/src/pages/eval/components/Eval.tsx
```diff
@@ -1,190 +1,188 @@
...
+      };
 
-      setTableFromResultsFile(body.data);
-      setConfig(body.data.config);
-      setAuthor(body.data.author);
-      setEvalId(id);
+      socket.on('init', handleMessage);
+      socket.on('update', handleMessage);
+      socket.on('error', (error: Error) => next(error));
+      return () => socket.disconnect();
```
sklein12, line 100: I don't believe this even works in cloud, it's just an artifact from copying it over from open source

src/app/src/pages/eval/components/ResultsView.tsx
```diff
@@ -174,6 +169,11 @@ export default function ResultsView({
     setColumnState,
     setAuthor,
   } = useResultsViewStore();
+
+  // ResultsView can only render when there is an eval selected.
+  invariant(typeof evalId === 'string', 'evalId must be a string');
+  evalId! as string;
```
sklein12, line 175: Does this case happen? Is this component rendered when `evalId` is `null`?

sklein12:
do we need all these changes to fix the bug? I don't want to get hung up on fixing this bug for the sake of a refactor right now.

package.json
```diff
@@ -207,6 +207,7 @@
     "rfdc": "^1.4.1",
     "semver": "^7.7.1",
     "socket.io": "^4.8.1",
+    "swr": "^2.3.3",
```
sklein12, line 210: how is this package different than reactQuery?

sklein12:
> I cleaned up eval state management / store state population in order to enforce an invariant that column state visibility is always populated during eval store initialization. As such, I want to make sure that I didn't miss any edge cases e.g. I'm not confident that there will always be an eval id when results are rendered, so please advise.

If this is a problem then maybe we should simplify it by passing the results down as props instead of fetching them from a store. That way we can always ensure there's an eval and/or evalId when rendering that component.

That being said, if you're not confident in the change, I'd see if we can fix the bug with a smaller blast radius.  This is an isolated issue for a non paying customer. If we can't fix it quickly, let's just tell them not to run evals with 20 columns.

src/app/src/pages/eval/components/ResultsView.tsx
will-holley: That's one of the things I'm unsure about but currently it seems to be an allowed state.

package.json
will-holley: They're very similar.  I typically use SWR so I defaulted to it but can swap if React Query is preferable. FWIW, I'm not sure that there's an equivalent functionality to [`useSWRSubscription`](https://swr.vercel.app/docs/subscription) in RQ so a switch may take additional work.

sklein12: Let's remove this. Having two libraries to fetch data is going to cause confusion. I can show you how to do this with react query and the built-in socket library.

src/app/src/pages/eval/components/ResultsView.tsx
sklein12: In that case you don't want to use an invariant. You should handle that state gracefully.

src/app/src/pages/eval/components/store.ts
```diff
@@ -60,78 +49,71 @@ interface TableState {
...
+      set(() => ({ table: results.table, version: resultsFile.version }));
+    }
+  },
+  config: null,
+  setConfig: (config: Partial<UnifiedConfig> | null) => set(() => ({ config })),
+
+  maxTextLength: 250,
+  setMaxTextLength: (maxTextLength: number) => set(() => ({ maxTextLength })),
+  wordBreak: 'break-word',
+  setWordBreak: (wordBreak: 'break-word' | 'break-all') => set(() => ({ wordBreak })),
```
sklein12, line 87: I realized after thinking about this a little bit that I'm not sure we want to remove this all from persistent storage. We persist the state of the table like options they selected and people may be benefiting from that as a feature.

sklein12: I would just check with Ian before removing.

src/app/src/pages/eval/components/ResultsView.tsx
will-holley: That's what I did with

```tsx
evalId && (
          <ResultsView
            recentEvals={recentEvalsData ?? []}
            onRecentEvalSelected={handleRecentEvalSelection}
          />
        )
```

in `Eval.tsx`.

However this `evalId` gets loaded from the store so the invariant (attempts) to keep these values in sync.

PR #3638, opened 2025-04-09
fix(logger): fix `[Object object]` empty string error

mldangelo:
Also improves test coverage


src/logger.ts
```diff
@@ -74,9 +74,22 @@ type StrictLogger = Omit<winston.Logger, keyof typeof LOG_LEVELS> & {
   [K in keyof typeof LOG_LEVELS]: StrictLogMethod;
 };
 
+/**
+ * Extracts the actual message string from potentially nested info objects
+ */
+function extractMessage(info: any): string {
```
will-holley, line 80: Why `any` instead of something like `info: { message: { message?: string } | null }`?

src/logger.ts
```diff
@@ -156,25 +171,28 @@ export function isDebugEnabled(): boolean {
   return getLogLevel() === 'debug';
 }
 
+/**
+ * Creates a logger method for the specified log level
+ */
+function createLogMethod(level: keyof typeof LOG_LEVELS): StrictLogMethod {
```
will-holley, line 177: IMO the previous format is preferable OR don't use this function for `debug`.

src/logger.ts
mldangelo: I implemented 

```ts 
type LogInfo = {
  message: string | { message?: any } | null | any;
  [key: string]: any;
};
```

but I think type safety is good but the any type is actually reasonable here given Winston's flexible log object structure.

src/logger.ts
mldangelo: I do see your point. I was trying to DRY off. This is handling a complex requirement - debug level should always include location, and other levels should include location only when debug is enabled.

PR #3650, opened 2025-04-09
fix: remove deprecated login flow

sklein12:
Login via email address is deprecated with the new cloud auth, pointing towards the website to login and get the token.

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -1,37 +1,11 @@
 import chalk from 'chalk';
 import type { Command } from 'commander';
 import dedent from 'dedent';
-import readline from 'readline';
 import { fetchWithProxy } from '../fetch';
 import { getUserEmail, setUserEmail } from '../globalConfig/accounts';
 import { cloudConfig } from '../globalConfig/cloud';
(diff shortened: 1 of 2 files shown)


src/commands/auth.ts
```diff
@@ -48,59 +22,36 @@ export function authCommand(program: Command) {
...
-        const { user } = await cloudConfig.validateAndSetApiToken(token, apiHost);
-
-        // Store token in global config and handle email sync
-        const existingEmail = getUserEmail();
-        if (existingEmail && existingEmail !== user.email) {
           logger.info(
-            chalk.yellow(
-              `Updating local email configuration from ${existingEmail} to ${user.email}`,
-            ),
+            `After logging in, you can get your api token at ${chalk.green('https://promptfoo.app/welcome')}`,
```
faizanminhas, line 51: Thought: `/welcome` is a weird path for this.

faizanminhas (approved):

src/commands/auth.ts
sklein12: agreed, but it's what we have

PR #3651, opened 2025-04-09
fix(testCaseReader): make JSON test file parsing preserve test case structure

### LINKED ISSUE #3629: Import from JSON does not work correctly
**Describe the bug**
[The documentation states](https://www.promptfoo.dev/docs/configuration/parameters/#import-from-jsonjsonl) that tests can be defined as json in a format like:

```
[
  { "vars": { "var1": "value1" }, "assert": [], "description": "Test #1" },
  { "vars": { "var1": "value2" }, "assert": [], "description": "Test #2" }
]
```

But the code seems to parse those contents as vars directly. This means any assertions defined in "assert" do not run, resulting in what look like passed tests!

**To Reproduce**

promptfooconfig.yaml
```
prompts:
  - 'Translate the following text to {{target_language}}: {{text}}'

providers:
  # Just repeats the markdown back to us
  - echo

tests: file://cases.json
```

cases.json
```
[
  { "vars": { "text": "Hello", "target_language": "Spanish" }, "assert": [{ "type": "equals", "value": "Hola" }], "description": "Test #1" }
]
```

Running promptfoo eval with the above configuration gives:

```
┌────────────┬────────────┬────────────┬────────────┐
│ assert     │ descripti… │ vars       │ [echo]     │
│            │            │            │ Translate  │
│            │            │            │ the        │
│            │            │            │ following  │
│            │            │            │ text to    │
│            │            │            │ {{target_… │
│            │            │            │ {{text}}   │
├────────────┼────────────┼────────────┼────────────┤
│            │ Test #1    │            │ [PASS]     │
│            │            │            │ Translate  │
│            │            │            │ the        │
│            │            │            │ following  │
│            │            │            │ text to :  │
└────────────┴────────────┴────────────┴────────────┘
==================================================

```

You can see that the vars were not parsed, the asserts did not come through and the prompt variables were not substituted. The test passes!

**Expected behavior**
This test should fail. The assert and vars should be properly parsed.

**Screenshots**

**System information:**

- Promptfoo version:

**Additional context**

mldangelo:
Fixes #3629
We should have returned earlier instead of passing into the csv row function.


kevinschaul:
Thanks!

PR #3653, opened 2025-04-09
feat(evals): Sharing Idempotence

will-holley:


src/share.ts
```diff
@@ -403,17 +434,32 @@ export async function createShareableUrl(
   }
   logger.debug(`New eval ID on remote instance: ${evalId}`);
 
-  const { domain } = determineShareDomain(evalRecord);
+  return getShareableUrl(evalRecord, showAuth);
```
sklein12, line 437: nice touch moving this logic into it's own function

src/share.ts
```diff
@@ -403,17 +434,32 @@ export async function createShareableUrl(
...
+    if (cloudConfig.isEnabled()) {
+      headers['Authorization'] = `Bearer ${cloudConfig.getApiKey()}`;
+    }
+
+    const res = await fetchWithProxy(`${url}/${eval_.id}`, {
+      method: 'GET',
+      headers,
+    });
+
+    return res.status !== 404;
```
sklein12: We have all of our cloud queries in here: https://github.com/promptfoo/promptfoo/blob/main/src/util/cloud.ts

could you move this query to this file and re-use the makeRequest helper?

PR #3660, opened 2025-04-10
fix: google vertex ais output format

abrayne:
This is a breaking change in the Vertex and AIStudio providers. This will have no impact on standard text responses. It will only impact the format when functions are called. This will break tests that involve function calling. It will require the transform to be updated.

At the moment, in AIS and Vertex, textual responses are concatenated into one large string. On top of that function calls are converted from json to string and concatenated with the text. This is very messy.

This change instead leaves the output in a structured format if a function call is present in the output.

This looks a bit more like the OpenAI approach.

I don't have a good feeling for how many tests this will impact, but I am keen to make this change as it is the right long term decision. AIStudio function calling was implemented by me recently, so I don't think many people will be using it. Vertex function calling has been around a little bit longer (a month or so), but I also implemented that.

@mldangelo if you are happy with this change please merge it. Otherwise we can discuss.


mldangelo:
Looks great! This is definitely a cleaner approach. Thank you!

abrayne:
Thanks Michael! Would be great if you could get a new version out with the last few days changes.

PR #3668, opened 2025-04-11
fix: promptfoo share respects sharing config from promptfooconfig.yaml

### LINKED ISSUE #3667: `promptfoo share` ignores `sharing.apiBaseUrl` and `sharing.appBaseUrl` in `promptfooconfig.yaml`
Given the following in `promptfooconfig.yaml` (as described in [the docs](https://www.promptfoo.dev/docs/usage/sharing/#advanced-self-hosted-sharing)):

```yaml
sharing:
  apiBaseUrl: https://doesnotexist.google.com
  appBaseUrl: https://doesnotexisteither.google.com
```

and running

```bash
promptfoo share
```

I get

```
» You need to have a cloud account to securely share your results.

1. Please go to https://promptfoo.app to sign up or log in.
2. Follow the instructions at https://promptfoo.app/welcome to login to the command line.
3. Run promptfoo share
```

So it looks like promptfoo is ignoring `sharing.apiBaseUrl` and `sharing.appBaseUrl` from `promptfooconfig.yaml`.

I would expect to get the same error as when doing this:

```bash
PROMPTFOO_REMOTE_API_BASE_URL=https://doesnotexist.google.com PROMPTFOO_REMOTE_APP_BASE_URL=https://doesnotexisteither.google.com 
```

which is:

```
[...]
    errno: -3008,
    code: 'ENOTFOUND',
    syscall: 'getaddrinfo',
    hostname: 'doesnotexist.google.com'
[...]
```

---

I've confirmed that my `promptfooconfig.yaml` is really parsed by adding some nonsense to it:

```yaml
sharing:
  apiBaseUrl: https://doesnotexist.google.com
  appBaseUrl: https://doesnotexisteither.google.com
asdasd
```

and running

```bash
promptfoo share
```

again:

```
YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key (5:1)
```

---

Version used:

```bash
promptfoo --version
```

```
0.109.1
```

mldangelo:
Fixes #3667

## Issue
When using `promptfoo share`, the command ignores the `sharing.apiBaseUrl` and `sharing.appBaseUrl` settings from `promptfooconfig.yaml`. Instead of using the custom API and app URLs specified in the config file, it displays a message prompting users to create a cloud account.

## Changes
- Modified `shareCommand` to load current configuration from `promptfooconfig.yaml` 
- The sharing settings from the config file are now applied to the eval record before checking if sharing is enabled
- Added a test to verify that the sharing configuration is properly applied


will-holley:
@mldangelo, this fix introduces a new bug. To reproduce:

1. Create two promptfoo configs with share blocks e.g. `example-a/promptfooconfig.yaml` and `example-b/promptfooconfig.yaml`.
2. Run `promptfoo eval -c example-a/promptfooconfig.yaml`.
3. Now run `cd example-b && promptfoo share`.
4. The sharing settings for Example B will overwrite the correct settings defined on Example A.

It's not clear to me looking at https://github.com/promptfoo/promptfoo/issues/3667 whether the example was ever evaluated.  If so, the sharing values scoped to `eval_.config.sharing` (which you're overwriting here) _should_ have been correct.  If it was evaluated and those values were missing, the solution is to fix _that_ problem.

PR #3669, opened 2025-04-11
feat: GLEU score

adityabharadwaj198:
* Added code to calculate gleu score.
* Added corresponding tests 
* Moved the getNGrams() method to a common file so it can be re-used by both gleu & bleu

### DIFF
--- a/src/assertions/gleu.ts
+++ b/src/assertions/gleu.ts
@@ -1,6 +1,3 @@
-/**
- * 
- */
 import type { AssertionParams, GradingResult } from '../types';
 import invariant from '../util/invariant';
 import { getNGrams } from './ngrams';
--- a/src/assertions/gleu.ts
+++ b/src/assertions/gleu.ts
@@ -1,0 +1,131 @@
+/**
+ * 
+ */
+import type { AssertionParams, GradingResult } from '../types';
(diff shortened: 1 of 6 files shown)


src/assertions/gleu.ts
```diff
@@ -0,0 +1,128 @@
...
+  test,     // Access to test case data
+}: AssertionParams): GradingResult {
+  // Validate inputs
+  invariant(
+    typeof renderedValue === 'string' ||
+    (Array.isArray(renderedValue) && renderedValue.every((v) => typeof v === 'string')),
+    '"gleu" assertion must have a string or array of strings value'
+  );
+
+  const threshold = assertion.threshold ?? 0.5;
```
adityabharadwaj198, line 115: @mldangelo I'm hoping the threshold value is correct. Let me know if not, and I can change it.

mldangelo:
hey @adityabharadwaj198! We can get this in if you would like to re-open it.

src/assertions/gleu.ts
mldangelo: 0.5 is a good choice

adityabharadwaj198:
@mldangelo  I'll reopen a new PR in some time.

PR #3671, opened 2025-04-11
fix(webui): fix eval comparison mode filter

will-holley:
- Right click eval id to open it in a new tab
- Only allow comparisons between evals w/ the same dataset


mldangelo (approved):
Thanks!

MrFlounder (approved):

PR #3678, opened 2025-04-11
perf(webui): Reduce memory usage of eval results

will-holley:
Serializing large result sets as JSON to IndexedDB adds significant overhead, which this PR reduces.  Eval result settings are now saved to local storage.

image


src/app/src/pages/eval/components/store.ts
```diff
@@ -69,31 +57,33 @@ interface TableState {
...
+  setEvalId: (evalId: string) => set(() => ({ evalId })),
+
+  author: null,
+  setAuthor: (author: string | null) => set(() => ({ author })),
+
+  version: null,
+  setVersion: (version: number) => set(() => ({ version })),
+
+  table: null,
+  setTable: (table: EvaluateTable | null) => set(() => ({ table })),
```
sklein12, line 71: What do you think about moving this out into its own store that's not persisted in local storage? would that help memory usage/processing time too?

mldangelo:
Note for Mr. JIT, in my unscientific testing, memory consumption went up from ~180MB to 270MB on a 400 row eval.

src/app/src/pages/eval/components/store.ts
will-holley: This won't be; this store has its persistence removed.

sklein12: ohhh right right sorry, i misread the file.

sklein12 (approved):
Amazing work!! Very simple change with massive impact!

sklein12:
> Note for Mr. JIT, in my unscientific testing, memory consumption went up from ~180MB to 270MB on a 400 row eval.

memory consumption went up?

will-holley:
@sklein12 @mldangelo 

This branch:

image

Main (before it crashed):

image

sklein12:
image
I'm still getting this

will-holley:
Sharing some findings here.  High memory usage is attributable to a handful of issues which arise when working with large strings:

- Data duplication in the store
- Operations on strings
- UI Interactions (e.g. scrolling, searching)

A patch proposed by @sklein12 was to truncate variable strings using [`String.prototype.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice); unfortunately this broke base64 encoded images as the `end` parameter (e.g. 5000 chars) is set to an arbitrary length and we lack an efficient way to check whether the length of a base64 encoded image string exceeds `end`.

PR #3703, opened 2025-04-15
fix(providers): output json rather than string from google live provider

abrayne:
Another breaking change to avoid having to transform every live output with JSON.parse.

Hopefully I've got this in before too many people are building tests with Live. If there are people building tests with live, we can probably leave this as is. It is a nice to have rather than a must have.


mldangelo (approved):
I think this is ok! Thank you so much @abrayne

PR #3709, opened 2025-04-15
chore(cli): improves robustness of hasEvalBeenShared util

will-holley:
Witnessed cases where `500` status codes will cause `hasEvalBeenShared` to return `true`.

### DIFF
--- a/src/share.ts
+++ b/src/share.ts
@@ -447,10 +447,22 @@ export async function createShareableUrl(
  */
 export async function hasEvalBeenShared(eval_: Eval): Promise<boolean> {
   try {
-    const { status } = await makeCloudRequest(`/${eval_.id}`, 'GET');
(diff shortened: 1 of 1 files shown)


src/share.ts
```diff
@@ -447,10 +447,22 @@ export async function createShareableUrl(
...
+    switch (res.status) {
+      // 200: Eval already exists i.e. it has been shared before.
+      case 200:
+        return true;
+      // 404: Eval not found i.e. it has not been shared before.
+      case 404:
+        return false;
+      default:
+        throw new Error(
+          `[hasEvalBeenShared]: unexpected API error: ${res.status}\n${res.statusText}`,
```
mldangelo, line 461: You are double logging `[hasEvalBeenShared]: `

PR #3713, opened 2025-04-16
feat(providers): Add support for Amazon nova-sonic

faizanminhas:
# Amazon Nova-Sonic

This PR adds support for Nova-sonic as a provider in promptfoo. Nova sonic is a speech to speech model that makes use of a bidirectional streaming api that can propagate events in both directions at any time. The protocol is very precise and the steps are documented here https://docs.aws.amazon.com/nova/latest/userguide/input-events.html. 

## Protocol
In short, the protocol:

1. Send a sessionStart event
2. Send a promptStart event
3. Send a system prompt
4. (optional) Send conversation history.
5. Stream audio- each piece of data is sent with a contentStart, content, and then a contentEnd event.
6. Wait for responses (There are event handlers set up to handle most cases)
7. end session when the api sends a contentEnd with a `END_TURN` stop reason

That said, the API itself is very error prone based on the configurations sent, and error messages are very opaque and non descriptive. It's recommended to tweak your configuration one at a time. The only output currently supported is in raw `audio/lpcm`, which most browsers can't natively play. 

## Multi-turn

Multi turn was implemented for this api. You can either directly pass in audio files in text form, or use the provided example to format the messages accordingly in `examples/amazon-bedrock/nova_sonic_prompt.js`

## Next steps
- Formatting audio files and variables correctly in results table.
- Token usage counting (info doesn't seem to be in the API)
- Tool handling local callbacks
- Using nova sonic as a grader


site/docs/providers/aws-bedrock.md
```diff
@@ -188,6 +188,42 @@ Nova models use a slightly different configuration structure compared to other B
...
+        topP: 0.95 # Nucleus sampling parameter
+      textOutputConfiguration:
+        mediaType: text/plain
+      toolConfiguration: # Optional tool configuration
+        tools:
+          - toolSpec:
+              name: 'getDateTool'
+              description: 'Get information about the current date'
+              inputSchema:
+                json: '{"$schema":"http://json-schema.org/draft-07/schema#","type":"object","properties":{},"required":[]}'
```
will-holley, line 211: Is `inputSchema` required in this example since (assumedly) the date getter would have no args?

src/providers/bedrock.ts
```diff
@@ -235,6 +235,72 @@ export interface BedrockAmazonNovaGenerationOptions extends BedrockOptions {
...
+                type: string;
+              };
+            };
+            required: string[];
+          };
+        };
+      };
+    }[];
+    toolChoice?: {
+      any?: any;
```
will-holley: Should `any` and `auto` be booleans?

examples/amazon-bedrock/promptfooconfig.nova-sonic.yaml
```diff
@@ -0,0 +1,56 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Amazon Bedrock Nova Sonic with tool use and variables
```
will-holley: Isn't this missing the tool definitions?

examples/amazon-bedrock/promptfooconfig.nova-sonic.yaml
```diff
@@ -0,0 +1,56 @@
...
+      - type: llm-rubric
+        value: the date and time is today
+  - vars:
+      audio_file: file://assets/weather.wav
+    metadata:
+      type: audio
+      conversationId: date
+    assert:
+      - type: llm-rubric
+        value: the weather is sunny
```
will-holley: In practice I found this will fail, so I changed it to `contains a weather forecast`.

examples/amazon-bedrock/nova_sonic_prompt.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * This function generates a properly formatted conversation for the OpenAI Realtime API.
```
will-holley: OpenAI Realtime API?

examples/amazon-bedrock/nova_sonic_prompt.js
```diff
@@ -0,0 +1,58 @@
...
+ * It handles the _conversation variable to maintain conversation history.
+ */
+module.exports = async function ({ vars, provider }) {
```
will-holley, line 16: It would be helpful if `vars` had a JSDoc type def.

examples/amazon-bedrock/promptfooconfig.nova-sonic.yaml
```diff
@@ -0,0 +1,56 @@
...
+      type: audio
+      conversationId: date
+    assert:
+      - type: llm-rubric
+        value: the date and time is today
+  - vars:
+      audio_file: file://assets/weather.wav
+    metadata:
+      type: audio
+      conversationId: date
```
will-holley, line 53: weather?

examples/amazon-bedrock/promptfooconfig.nova-sonic.yaml
```diff
@@ -0,0 +1,56 @@
...
+  - file://nova_sonic_prompt.js
+
+defaultTest:
+  vars:
+    system_message: You are a helpful assistant. Answer the question based on the audio input.
+  metadata:
+    type: text
+    conversationId: date
+
+tests:
```
will-holley, line 44: Please add a comment that each test case is part of a multi-turn conversation i.e. why the `conversationId` is used

site/docs/providers/aws-bedrock.md
faizanminhas: I believe inputschema is required, even when there are no args- https://docs.aws.amazon.com/nova/latest/userguide/speech-tools-use.html (See the example with the provided schema with no args). 

That said, the error given with a malformed tool config is very opaque so I'd recommend following the example where possible.

src/providers/bedrock.ts
faizanminhas: this was wrong, I fixed it

mldangelo:
Thanks!

PR #3715, opened 2025-04-16
refactor: colocate fetching evalID

billybonks:
Easier when reading the code if you do not need to flip between different places to understand how evalId is resolved.


src/app/src/pages/eval/components/Eval.tsx
```diff
@@ -13,7 +13,7 @@ import { useResultsViewSettingsStore, useStore } from './store';
 import './Eval.css';
 
 interface EvalOptions {
-  fetchId?: string;
+  fetchId?: string | null;
```
will-holley: This should be `fetchId: string | null` because the value will always be defined.

will-holley (dismissed):

faizanminhas (dismissed):

src/app/src/pages/eval/components/Eval.tsx
billybonks: that makes sense 👍🏿  and done :)

will-holley (approved):

mldangelo:
@billybonks can you please either make this editable by us or resolve merge conflicts and we'll merge it? Thanks!

billybonks:
@mldangelo done.

billybonks:
@mldangelo @will-holley @faizanminhas bumping this pr for merge.

mldangelo (dismissed):
Thank you!

mldangelo:
Thanks @billybonks!  Really appreciate the refactor!

billybonks:
@mldangelo would love to know if there is a particular reason why its supported as URL part and as query parameter, if it is legacy behavior would you be open to me removing the query parameter version entirely?

PR #3716, opened 2025-04-16
fix: use ajv with formats everywhere

schipiga:
Hi,
What do you think about to use `ajv` instance from `is-json` everywhere, which includes different formats together with `strict` mode managing?


schipiga:
Hi @mldangelo, I suppose it could be useful to have same `ajv` settings in every assertion. For instance, right now I have failed `is-valid-openai-tools-call` assert, due to used format `time` in my schema:
```js
properties: {
	dateFromValue: {
		type: 'string',
		description: 'Value of the type DATE....',
		format: 'date',
	},
```
And it's because `ajv-formats` is used only in `is-json`

mldangelo (approved):
Thank you! This is a great idea.

PR #3717, opened 2025-04-16
feat: persist search in url

billybonks:
# Problem
Currently, search state is maintained during navigation but is not reflected in the URL. This creates two issues:

We can't share links with specific search parameters
LLMs cannot easily extract search information from the page URL

# Solution
Implement URL-based search state persistence by adding search parameters to the URL.

While the search state is not lost during navigation, if i want to for example link the page to someone else with the correct search state, or get a LLM to extract data from the page.

# Implementation Details

- Modified navigation to use URL fragments instead of query parameters for a cleaner URL structure and requires less work merging the evalId qp and search qp
- When updating search parameters, we include previous parameters with Object.fromEntries(searchParams.entries()) this is only so that we do not loose existing evalId qp

# Questions
Why does page support evalId in both query parameter format and URL fragment format? is there a desired approach, i see most of the code relies on the qp format. personally i would prefer deprecating the qp format

somewhat related to this https://github.com/promptfoo/promptfoo/pull/3715.

### DIFF
--- a/src/app/src/pages/eval/components/Eval.tsx
+++ b/src/app/src/pages/eval/components/Eval.tsx
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React, { useCallback } from 'react';
 import { useSearchParams, useNavigate } from 'react-router-dom';
 import { IS_RUNNING_LOCALLY } from '@app/constants';
 import { ShiftKeyProvider } from '@app/contexts/ShiftKeyContext';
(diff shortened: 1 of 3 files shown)


src/app/src/pages/eval/components/ResultsView.tsx
```diff
@@ -174,9 +174,14 @@ export default function ResultsView({
   const { showToast } = useToast();
   const [searchText, setSearchText] = React.useState(searchParams.get('search') || '');
   const [debouncedSearchText] = useDebounce(searchText, 200);
-  const handleSearchTextChange = React.useCallback((text: string) => {
-    setSearchText(text);
-  }, []);
+  const handleSearchTextChange = React.useCallback(
+    (text: string) => {
+      const searchEntries = Object.fromEntries(searchParams.entries());
+      setSearchParams({ ...searchEntries, search: text });
```
will-holley: You can do `setSearchParams((prev) => ({ ...prev, search: text }));` and get rid of `searchEntities`.

src/app/src/pages/eval/components/Eval.tsx
```diff
@@ -76,7 +76,10 @@ export default function Eval({
   );
 
   const handleRecentEvalSelection = async (id: string) => {
```
will-holley:
```ts
const [searchParams] = useSearchParams();
const handleRecentEvalSelection = useCallback(
    async (id: string) => {
      navigate({
        pathname: `/eval/${encodeURIComponent(id)}`,
        search: searchParams.toString(),
      });
    },
    [searchParams, navigate],
  );
```

will-holley (changes requested):
Hi @billybonks, thanks for the contribution! There are a few minor improvements but otherwise this LGTM.

billybonks:
thanks for the feedback @will-holley, it has been a while since i have used react :).

I made the updates and i also added some tests.

will-holley (approved):
Thanks for making those changes! LGTM.

PR #3721, opened 2025-04-16
chore(cli/redteam/run): Default to outputting generated Redteam config in same dir as input config

will-holley:
If a value is not provided for `--output` when running `promptfoo redteam run` the default is to create a `redteam.yaml` in the current working directory.  This PR changes the default to create the `redteam.yaml` in the same directory as the given config file (i.e. `promptfooconfig.yaml`) if the latter is provided  i.e. `promptfoo redteam run -c ~/attack/config.yaml` will generate `~/attack/redteam.yaml`.

This change fixes an downstream bug that I encountered: a `promptfooconfig.yaml` that contains relative filepaths e.g.

```yaml
targets:
  - file://./targets/chat.yaml
```

will result in a `redteam.yaml` that breaks at runtime unless `.` is the current working directory.

### DIFF
--- a/src/redteam/commands/run.ts
+++ b/src/redteam/commands/run.ts
@@ -21,7 +21,7 @@ export function redteamRunCommand(program: Command) {
     )
     .option(
       '-o, --output [path]',
-      'Path to output file for generated tests. Defaults to redteam.yaml',
+      'Path to output file for generated tests. Defaults to redteam.yaml in the same directory as the configuration file.',
     )
     .option('--no-cache', 'Do not read or write results to disk cache', false)
     .option('--env-file, --env-path <path>', 'Path to .env file')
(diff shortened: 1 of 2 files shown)


mldangelo (commented):
this is a nice catch and has been annoying me for months. Can you please add a test to cover your change to src/redteam/shared.ts? 

We will need to make sure this doesn't break the cloud runner

test/redteam/shared.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    await doRedteamRun({ output: outputPath });
+    expect(doGenerateRedteam).toHaveBeenCalledWith(
+      expect.objectContaining({
+        output: outputPath,
+      }),
+    );
+  });
+
+  it('should locate the out file in the same directory as the config file if output is not specified', async () => {
+    const dirPath = faker.system.directoryPath();
```
mldangelo: Please don't add a dependency for this one line.

mldangelo (commented):
Change is good! Please remove faker and feel free to merge after. This is technically breaking and should result in a minor version bump.

PR #3732, opened 2025-04-16
chore: better parsing

MrFlounder:
using gpt 4.1 the judge response gives much more json string with no closing }, so this pr update the json parsing logic and also handle parsing error better.


MrFlounder:
@MrFlounder no i didn't   this is what I see from a few redteam runs though
We can also enforce json return - I can make it after switching to 4.1 - and I think our json return for gpt 4o might not be working

PR #3741, opened 2025-04-17
feat: update openai grading model to gpt 4.1

MrFlounder:


mldangelo:
Thank you @MrFlounder! Next release needs to be a minor because of this.

PR #3743, opened 2025-04-17
chore: Auth for remote generation

sklein12:
Add auth token in with remote generation


mldangelo (commented):
Let’s discuss before you land this

PR #3746, opened 2025-04-18
feat: Modify LLM Rubric rubricPrompt rendering to support arbitrary objects

GICodeWarrior:
Example config:
```yaml
prompts:
  - ' '

providers:
  - echo

tests:
  - vars:
      root: root_attribute_value
      nested:
        attribute: nested_attribute_value

defaultTest:
  assert:
    - type: llm-rubric
      provider: openai:gpt-4o-mini
      rubricPrompt:
        - role: user
          content: |
            Root value -- {{ root }}
            Nested value -- {{ nested.attribute }}
```

Before this change, the `nested.attribute` value is blank.  After this change, it has the value `nested_attribute_value`.


mldangelo (approved):
Thank you!

PR #3753, opened 2025-04-18
chore: Increase max outuput tokens for `google:gemini-2.5-pro-exp-03-25` to 2048 in Gemini example

will-holley:
Current ceiling is insufficient for thinking and causes eval to fail.


mldangelo (approved):
Thanks!

PR #3756, opened 2025-04-20
chore: update default to azure

MrFlounder:
https://github.com/promptfoo/promptfoo-cloud/pull/1111 depend on this

In cloud, we will be using azure open ai as default provider when we can


mldangelo (approved):
Thanks!

MrFlounder:
@mldangelo sorry I meant to close this pr, this pr may cause side effect. What I really wanted to do is when doing remote grading, etc in cloud, set the default provider to azure
Another pr in cloud is merged earlier today. I think this pr doesn't have value anymore. We can chat tomorrow

PR #3785, opened 2025-04-22
feat(redteam): mp plugin

will-holley:


src/evaluator.ts
```diff
@@ -246,24 +247,73 @@ export async function runEval({
...
-        filters,
-        originalProvider: provider,
-        test,
+      // The memory poisoning plugin requires the following additional steps:
+      // 1. Generate a scenario containing memories and follow up questions/commands which are dependent on the memories.
+      // 2. Send the memory message to the provider.
+      // 3. Send the test case to the provider; the test case should poison the memory created in the previous step.
+      // 4. Send the follow up question to the provider.
+      // A memory is successfully poisoned if the follow up question is answered incorrectly.
+      if (test.metadata?.pluginId === 'agentic:memory-poisoning') {
```
will-holley: I preferred this over a custom provider so it isn't the user's responsibility to choose the provider that corresponds with this plugin.

src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -0,0 +1,83 @@
...
+  ): Promise<ProviderResponse> {
+    const targetProvider: ApiProvider | undefined = context?.originalProvider;
+    const purpose = context?.test?.metadata?.purpose;
+    invariant(targetProvider, 'Expected originalProvider to be set');
+    invariant(context?.vars, 'Expected vars to be set');
+    invariant(purpose, 'Expected purpose to be set');
+
+    // Generate a scenario containing memories and follow up questions/commands which are dependent on the memories.
+    const scenarioRes = await fetch(getRemoteGenerationUrl(), {
+      body: JSON.stringify({ task: 'agentic:memory-poisoning-scenario', purpose }),
```
sklein12: please include the user email and package version on this call

src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -0,0 +1,83 @@
...
+      headers: { 'Content-Type': 'application/json' },
+      method: 'POST',
+    });
+    const scenario = await scenarioRes.json();
```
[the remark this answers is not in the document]
sklein12: yes, please do this and add error handing for this function in the event it fails. We will need detailed logs about the failure.

src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -0,0 +1,119 @@
...
+    invariant(targetProvider, 'Expected originalProvider to be set');
+    invariant(context?.vars, 'Expected vars to be set');
+    invariant(purpose, 'Expected purpose to be set');
+
+    // Generate a scenario containing memories and follow up questions/commands which are dependent on the memories.
+    const scenario = await generateMemoryPoisoningScenario(purpose);
+    context.vars.scenario = scenario;
+
+    // Send the memory message to the provider.
+    const memoryResponse = await targetProvider!.callApi(scenario.memory, context);
```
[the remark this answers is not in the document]
sklein12: ellipsis is right here

src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    // Send the test case to the provider; the test case should poison the memory created in the previous step.
+    const testResponse = await targetProvider!.callApi(prompt, context);
+
+    // Send the follow up question to the provider.
+    const response = await targetProvider!.callApi(scenario.followUp, context);
+
+    // Log the conversation onto the vars object; all references to vars will have access
+    // to the conversation object.
```
sklein12: 👍 

Will the user be able to see this in the UI? I think we have some treatment for this for multi-turn things like GOAT already.

will-holley: Yeah, as a variable.

src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -0,0 +1,103 @@
...
+        {
+          input: scenario.followUp,
+          output: response.output,
+          type: 'followup',
+        },
+      ];
+
+      return { output: response.output };
+    } catch (error) {
+      const _logger = context?.logger ?? logger;
```
sklein12: I've never seen this before. What is a logger on the context?

will-holley: It's of the same class as `logger` but otherwise I couldn't tell you why it's scoped to `context`.  I inherited in case there is some additional context that is scoped to `context.logger`.

sklein12: I have no idea, I don't think it's used, I think you can remove this and just use the `logger` you imported

sklein12 (dismissed):
Great Work!!

PR #3795, opened 2025-04-24
chore: add target option to cli redteam run

faizanminhas:
Not entirely sure that this is the best way to do this


mldangelo (approved):
Would be nice to capture this with a test so we don't accidentally break it later. Might also be nice to update the integration tests

mldangelo (approved):
Thank you!

PR #3801, opened 2025-04-25
refactor: remove if string check

billybonks:
I was exploring the HTTP Provider to see points of extensibility specifically propagating metadata / usage tokens, i noticed that the if statement was not required (probably was used in the past)

Thought i would clean it up in the mean time.


mldangelo (approved):
Thanks!

PR #3805, opened 2025-04-25
chore: resolve relative provider paths from cloud configs

faizanminhas:
This specifically handles the case where `promptfoo redteam run` is called with a cloud config (i.e. `-c <config_uuid>`). If the cloud config was specified to be a custom provider for golang, python, etc, then if the path provided isn't absolute, it will resolve it based on the current working directory.

### DIFF
--- a/src/providers/registry.ts
+++ b/src/providers/registry.ts
@@ -12,7 +12,7 @@ import RedteamPandamoniumProvider from '../redteam/providers/pandamonium';
 import { ServerToolDiscoveryMultiProvider } from '../redteam/providers/toolDiscoveryMulti';
 import type { LoadApiProviderContext } from '../types';
 import type { ApiProvider, ProviderOptions } from '../types/providers';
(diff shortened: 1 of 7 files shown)


src/providers/registry.ts
```diff
@@ -449,7 +449,11 @@ export const providerMap: ProviderFactory[] = [
     ) => {
       // Load script module
       const scriptPath = providerPath.split(':')[1];
-      return new ScriptCompletionProvider(scriptPath, providerOptions);
+      // Resolve relative paths against context.basePath or current working directory
+      // If using a cloud config, always use process.cwd() instead of context.basePath
+      const isCloudConfig = providerOptions.config?.isCloudConfig === true;
```
mldangelo: Can you abstract this into a helper? Could even start with `const scriptPath = providerPath.startsWith('file://')`

typpo (approved):
lgtm!

PR #3823, opened 2025-04-28
fix(matchers): score results correctly with trailing newlines.

philnash:
In trying to use the context-recall matcher I was getting unexpected results. It turned out that the response included a newline at the end, causing the matcher to count one extra failed sentence. (This was using Gemini 2.0 Flash for grading.)

To correct, I added a function that splits on new lines and filters empty lines and it worked as expected for me after.

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -380,6 +380,10 @@ function tryParse(content: string) {
   return content;
 }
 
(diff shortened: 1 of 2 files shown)


typpo (approved):
Great catch!  Thanks so much

PR #3825, opened 2025-04-28
fix: incorrect evalId socket mode /evals

billybonks:
eval data is fetched separately then the latest evalId,

evalId is fetch using /results and evalData uses the socket init. /results is not sorted by created at, this means that the ui can have a random evalId based on sqllite default ordering while the data is correct.

this commits update the source of truth to be the same for the data and the eval by returning the evalId along side the data.

but the latest eval for non socket instances will also be wrong since it relies on the same ordering

eval grid is unaffected as the table does local ordering but would break if it was server paginated

### DIFF
--- a/src/app/src/pages/eval/components/Eval.tsx
+++ b/src/app/src/pages/eval/components/Eval.tsx
@@ -122,13 +122,10 @@ export default function Eval({
         setTableFromResultsFile(data);
         setConfig(data?.config);
         setAuthor(data?.author || null);
-        fetchRecentFileEvals().then((newRecentEvals) => {
-          if (newRecentEvals && newRecentEvals.length > 0) {
-            setDefaultEvalId(newRecentEvals[0]?.evalId);
-            setEvalId(newRecentEvals[0]?.evalId);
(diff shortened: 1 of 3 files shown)


will-holley:
Hi @billybonks,

Thanks for identifying and fixing this issue.  I am concerned, however, that the fix may have unintended side-effects due to the removal of `fetchRecentFileEvals`.  State management within this component is already complex and could use refactoring; therefore, in consideration of this risk and the lack of a felt pain-point (since we're not server-side paginating), I am going to close this PR.

Thanks for your contribution & understanding,

Will

billybonks:
@will-holley i 100% agree that state loading code is overcomplicated. and removing fetch recent evals does cause a bug in hindsight, because it populates the first dropdown ` setRecentEvals(body.data);`

part of the refactor would be for the socket to return the eval id instead of us trying to merge the data from two sources 

Would you accept a pr that improves the socket api to return all required data instead of partial data, as a start?

image

billybonks:
Just did a quick stab at refactoring and ResultsFile definitely needs id to improve the code

PR #3832, opened 2025-04-28
[Draft] feat(providers): mcp support for all models that support function calling

AISimplyExplained:
Currently
MCP Support for Anthropic Messages
OpenAI Chat
Cleanup for mcp servers and associated code.


typpo (dismissed):
nice!

PR #3838, opened 2025-04-28
chore: set telemetry key

sklein12:
There will never be an env var, so this accidentally disabled telemetry. I set the env var so it wouldn't run in local dev, didn't think about it properly.


MrFlounder (approved):

src/telemetry.ts
```diff
@@ -24,18 +25,16 @@ export type TelemetryEvent = z.infer<typeof TelemetryEventSchema>;
...
 try {
-  const posthogKey = getEnvString('POSTHOG_KEY');
-  if (posthogKey) {
-    posthogClient = new PostHog(posthogKey, {
-      host: EVENTS_ENDPOINT,
-    });
-  }
+  posthogClient = new PostHog(POSTHOG_KEY, {
+    host: EVENTS_ENDPOINT,
+  });
```
faizanminhas: smart

src/telemetry.ts
```diff
@@ -24,18 +25,16 @@ export type TelemetryEvent = z.infer<typeof TelemetryEventSchema>;
...
 } catch {
   posthogClient = null;
 }
```
faizanminhas: can the posthog constructor even throw?

src/util/isLocalDev.ts
```diff
@@ -0,0 +1,28 @@
...
+      path.join(projectRoot, '.git'),
+    ];
+
+    // If any of these dev files exist, we're in the development environment
+    return devFileChecks.some((file) => fs.existsSync(file));
+  } catch {
+    // If there's any error in checking (e.g., permission issues), assume we're not in dev
+    return false;
+  }
+}
```
faizanminhas: What would `NODE_ENV` be for the `node_modules` or when running locally? I'm surprised you have to check for a git repo

src/telemetry.ts
```diff
@@ -96,7 +95,7 @@ export class Telemetry {
   }
 
   private sendEvent(eventName: TelemetryEventTypes, properties: EventProperties): void {
-    if (getEnvString('NODE_ENV') === 'test') {
```
mldangelo, line 99: please revert this

src/telemetry.ts
```diff
@@ -24,18 +25,16 @@ export type TelemetryEvent = z.infer<typeof TelemetryEventSchema>;
 export type TelemetryEventTypes = TelemetryEvent['event'];
 export type EventProperties = TelemetryEvent['properties'];
 
+export const POSTHOG_KEY = 'phc_E5n5uHnDo2eREJL1uqX1cIlbkoRby4yFWt3V94HqRRg';
```
mldangelo: do we definitely want to hardcode this or can we drop it in from an env variable at build time?

src/util/isLocalDev.ts
```diff
@@ -0,0 +1,28 @@
+import fs from 'fs';
+import path from 'path';
+
+export function isLocalDev(): boolean {
+  if (process.env.NODE_ENV === 'test') {
```
mldangelo: use getEnvString here

mldangelo (commented):
Please consider removing isLocalDev and passing in POSTHOG_KEY at build time. Would much rather do it that way.

sklein12:
@mldangelo how do you write a constant when doing a tsc build? Is there any secrets here or do I have to code gen a file?

sklein12:
@coderabbitai take a look plz

PR #3848, opened 2025-04-29
docs: updates to grading documentation

vsauter:


site/docs/red-team/troubleshooting/false-positives.md
```diff
@@ -1,31 +0,0 @@
----
```
mldangelo, line 1: Are you sure you want to get rid of this file?

mldangelo (commented):
Looks great, thank you!

site/docs/red-team/troubleshooting/false-positives.md
vsauter: It's replaced by the configuring grader FAQ, per Ian's rec. we can create a separate file if you want. WDYT?

vsauter: Stays within the troubleshooting section

mldangelo (dismissed):
Thanks!

mldangelo (approved):

PR #3859, opened 2025-05-01
chore(webui): mark dataset plugins in setup UI

mldangelo:


src/redteam/constants.ts
```diff
@@ -1082,14 +1093,9 @@ export const riskCategories: Record<string, Plugin[]> = {
     'harmful:unsafe-practices',
     'harmful:violent-crime',
     'harmful:weapons:ied',
-    'unsafebench',
```
typpo, line 1117: Do we need to create a new category so they show on the report?

PR #3862, opened 2025-05-01
chore: Respect Max text length for variable cells in results table

faizanminhas:
Previously, the entire value was shown, which wasn't using the maxtextlength property from the table.

## Summary by Sourcery

Chores:
- Updated rendering of table cell values to use TruncatedText component


faizanminhas:
@mldangelo Just verified that images are unaffected: 
Screenshot 2025-05-01 at 11 07 58 AM

audio is actually more bearable since currently we don't handle audio input correctly, hopefully fixed by my other PR #3864

PR #3864, opened 2025-05-01
feat: Display audio file variables in result table

faizanminhas:
When reading in `file://` for vars, it will also store the extension, and use that to set the mime type in order to play back audio and video vars correctly. 

Adds metadata to a new key that should be internal to promptfoo (`_promptfooFileMetadata`). I've tested this with nova sonic (fixes applied), as well as the openai audio tests and it seems to work well:

Screenshot 2025-05-19 at 7 46 58 AM

## Summary by Sourcery

Add support for displaying audio, video, and image files in the results table with file metadata

New Features:
- Enable rendering of media files (audio, video, images) in the results table
- Add file metadata collection for different media types

Enhancements:
- Improve results table to show media files with their original file path and type
- Add metadata extraction for file variables


adelmuursepp:
I am testing the UI for audio and have a lower priority question. Is the reason for Nova Sonic showing no conversation for the first and then {{ interrupted }} for the second one related to the audio processing in the provider?

image

adelmuursepp:
> When reading in `file://` for vars, it will also store the extension, and use that to set the mime type in order to play back audio and video vars correctly.
> 
> Adds metadata to a new key that should be internal to promptfoo (`_promptfooFileMetadata`). I've tested this with nova sonic (fixes applied), as well as the openai audio tests and it seems to work well:
> 
> Screenshot 2025-05-19 at 7 46 58 AM
> ## Summary by Sourcery
> Add support for displaying audio, video, and image files in the results table with file metadata
> 
> New Features:
> 
> * Enable rendering of media files (audio, video, images) in the results table
> * Add file metadata collection for different media types
> 
> Enhancements:
> 
> * Improve results table to show media files with their original file path and type
> * Add metadata extraction for file variables

For this I'm confused why in the output there is also a audio bar and transcript? When I test I don't see it. I assume you did some extra config. Asking so I can also do that and test that with adding similarly audio in output for Gemini I don't break anything you worked on.

adelmuursepp:
@faizanminhas Thanks for the comments. Yes then that makes sense. If you update branch with  the base branch I will review also and can approve!

mldangelo (approved):
Let's get this merged! Thanks so much @faizanminhas!

PR #3892, opened 2025-05-05
fix(cli): Read `.env` file prior to calling env var getters

will-holley:
I encountered an issue where values in my `.env` were not being read e.g. `LOG_LEVEL=debug`.  I traced this case to `src/logger.ts` and logged `getEnvString('LOG_LEVEL')` which returned `undefined` because my `.env` file had not been read prior to `getEnvString`'s call.  This change ensures the `.env` file is _always_ read prior to getter calls.

## Summary by Sourcery

Bug Fixes:
- Fixed issue where environment variables from `.env` file were not being read before being accessed

### DIFF
--- a/src/envars.ts
+++ b/src/envars.ts
@@ -1,3 +1,4 @@
+import 'dotenv/config';
 import cliState from './cliState';
 import type { EnvOverrides } from './types/env';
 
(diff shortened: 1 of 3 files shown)


src/envars.ts
```diff
@@ -1,3 +1,4 @@
+import 'dotenv/config';
```
mldangelo, line 1: many of the cli commands support overriding the env var path. I think this change is ok but make sure it makes sense in that context too

will-holley: That'll be safe because `override: true` is passed to that call, ensuring any conflicting values will prefer the user override.

PR #3907, opened 2025-05-06
feat(redteam): Target Purpose Discovery

will-holley:
## Summary by Sourcery

Add a new 'discover' command to the redteam module for automatically discovering a target application's purpose

New Features:
- Introduce a new 'discover' command to enhance attack probe efficacy by identifying target application purpose

Enhancements:
- Add command-line options for configuration, target, environment path, and verbosity


src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,179 @@
...
+          purposes[targetIdx] = purpose;
+        }
+
+        targetIdx++;
+      }
+
+      // Persist the purposes:
+      if (args.target) {
+        // TODO(Will): Save to the database
+        throw new Error('Saving purpose to database is not yet implemented');
```
MrFlounder: what's the consideration of saving it to db?

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,179 @@
...
+          invariant(false, 'Config is missing both "targets" and "providers" keys');
+        }
+      }
+
+      // Handle target: load it from Cloud.
+      if (args.target) {
+        targets.push(await getProviderFromCloud(args.target));
+      }
+
+      invariant(targets.length === 0, 'An error occurred loading the target config');
```
MrFlounder: should this be checking the opposite?

examples/redteam-purpose-generation/promptfooconfig.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - id: http
+    config:
+      url: https://redpanda-internal-rag-example.promptfoo.app/chat
+      method: POST
+      headers:
+        Content-Type: application/json
+      body:
+        input: '{{prompt}}'
+      transformResponse: json.response
+    purpose: >-
```
MrFlounder: Should this exist in the example? or it should be empty or commented out?

MrFlounder (commented):
Looks pretty clean!

examples/redteam-purpose-generation/promptfooconfig.yaml
will-holley: Good question. I'm leaning towards removing the `purpose` property to better demonstrate that it's generated.

examples/redteam-purpose-generation/README.md
```diff
@@ -0,0 +1,37 @@
...
+# Note that the same output as the previous step is re-used in order to provide
+# access to the purpose:
+promptfoo redteam generate \
+    -c examples/redteam-purpose-generation/promptfooconfig.yaml \
+    -o examples/redteam-purpose-generation/redteam.yaml
+```
+
+## Full Redteam Scan
+
+Run discover within the context of a full Redteam scan:
```
faizanminhas: nit: `red team` is 2 words

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,285 @@
...
+export async function doTargetPurposeDiscovery(
+  target: ApiProvider,
+  maxTurns?: number,
+): Promise<string> {
+  const conversationHistory: { type: 'promptfoo' | 'target'; content: string }[] = [];
+
+  let turnCounter = 1;
+
+  const pbar = new cliProgress.SingleBar({
+    format: `Discovering purpose {bar} {percentage}% | {value}/${maxTurns ? '{total}' : '∞'} turns`,
```
faizanminhas: Would the infinity freak people out / cause confusion? Maybe omit the fraction part if maxTurns isn't supplied.

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,285 @@
...
+    barCompleteChar: '\u2588',
+    barIncompleteChar: '\u2591',
+    hideCursor: true,
+  });
+
+  pbar.start(
+    // fallback: estimate of 25 turns
+    maxTurns ?? 25,
```
faizanminhas: oh, it's 25- maybe the infinity should be 25 then?

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,285 @@
...
+            error instanceof Error ? error.stack : ''
+          }`,
+        );
+        process.exitCode = 1;
+        return;
+      }
+
+      // Then, handle the response:
+      // If preview is enabled, print the purpose to the console:
+      if (!args.preview) {
```
faizanminhas: This comment doesn't match I think- there's no else to preview / logger.info the purpose? Unless it's printed as part of the discovery process

src/redteam/commands/discover.ts
will-holley: The progress bar requires a total so `25` is a estimation to satisfy it.

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,287 @@
...
+  });
+
+  pbar.start(
+    // fallback: estimate of 25 turns
+    maxTurns ?? 25,
+    0,
+  );
+
+  while (true) {
+    const res = await fetch(getRemoteGenerationUrl(), {
```
sklein12: `fetchWithProxy` plz

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,287 @@
...
+/**
+ * Saves the purpose to the database.
+ *
+ * @param targetId - The target ID.
+ * @param purpose - The purpose.
+ * @returns The response from the database.
+ */
+async function savePurpose(targetId: string, purpose: string) {
+  // First, attempt to load from the environment (self-hosted)
+  let apiDomain: string | undefined = getEnvString('PROMPTFOO_REMOTE_API_BASE_URL');
```
sklein12: self hosted doesn't have providers

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,287 @@
...
+    apiDomain = cloudConfig.getApiHost();
+  }
+
+  invariant(apiDomain, 'Error determining API domain');
+
+  const url = `${apiDomain!}/api/v1/providers/${targetId}`;
+
+  logger.debug(`Saving purpose to ${url}`);
+
+  const res = await fetch(url, {
```
sklein12: `fetchWithProxy` please

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,287 @@
...
+      // If user provides a config, read the target from it:
+      if (args.config) {
+        config = await readConfig(args.config);
+        invariant(config, 'An error occurred loading the config');
+        invariant(config.providers, 'Config must contain targets or providers');
+        const providers = await loadApiProviders(config.providers);
+        target = providers[0];
+      }
+      // If the target flag is provided, load it from Cloud:
+      else if (args.target) {
```
sklein12: should you check and let the user know these both can't be set?

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,287 @@
...
+        const providerOptions = await getProviderFromCloud(args.target);
+        target = await loadApiProvider(providerOptions.id, { options: providerOptions });
+      }
+      // A config or a target must be provided:
+      else {
+        logger.error(
+          'A config (-c, --config <path>) or a target (-t, --target <id>) must be provided!\n',
+        );
+        process.exitCode = 1;
+        return;
```
sklein12: should we support the default `promptfooconfig.yaml` in the current directory like our other commands?

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,287 @@
...
+        // Persist the purposes:
+        if (args.target) {
+          await savePurpose(args.target, purpose);
+        } else {
+          invariant(config, 'Config is required');
+
+          // Write to a redteam.yaml file:
+          if (!args.output) {
+            args.output = path.relative(process.cwd(), 'redteam.yaml');
+          }
```
sklein12: do we want to write it to a redteam.yaml or write it to the existing config file?

src/redteam/commands/generate.ts
```diff
@@ -85,22 +102,33 @@ export async function doGenerateRedteam(
...
+    if (resolved.config.providers && Array.isArray(resolved.config.providers)) {
+      // Validate that at least one provider is provided:
+      invariant(
+        resolved.config.providers.length > 0,
+        'At least one provider must be provided in the config file',
+      );
+      const providers = await loadApiProviders(resolved.config.providers);
+      targetPurpose = await doTargetPurposeDiscovery(providers[0]);
+    }
+  } else if (options.purpose && targetPurpose) {
```
sklein12: when would we ever hit this logic branch?

src/redteam/commands/discover.ts
will-holley: My original implementation updated the `promptfooconfig.yaml` file in place but this was vetoed (yesterday during a review) because we don't otherwise mutate configs.

src/redteam/commands/discover.ts
will-holley: Ok so this only needs to work for targets that were defined in Cloud or an on-prem deployment?

sklein12: It should also work for providers defined in a config file.

src/redteam/commands/generate.ts
will-holley: The user passed in `--purpose`

src/redteam/commands/discover.ts
will-holley: This function is used to save the generated purpose when the `--target <id>` flag is used.

sklein12: oh got it, then this wouldn't be applicable to self-hosted open source

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,309 @@
...
+      let config: UnifiedConfig | null = null;
+      // Although the providers/targets property supports multiple values, Redteaming only supports
+      // a single target at a time.
+      let target: ApiProvider | undefined = undefined;
+
+      // If user provides a config, read the target from it:
+      if (args.config) {
+        config = await readConfig(args.config);
+        invariant(config, 'An error occurred loading the config');
+        invariant(config.providers, 'Config must contain targets or providers');
```
sklein12: can you throw errors  and log information to the user about how they should resolve them?  Invariants should be used when something should never ever be true where it indicates something has gone terribly wrong.

src/redteam/commands/discover.ts
will-holley: Do you know if `cloudConfig.isEnabled()` will be truthy for on-prem or do I need a separate flag?

src/redteam/commands/generate.ts
```diff
@@ -85,22 +102,28 @@ export async function doGenerateRedteam(
     );
     testSuite = resolved.testSuite;
     redteamConfig = resolved.config.redteam;
-  } else if (options.purpose) {
-    // There is a purpose, so we can just have a dummy test suite for standalone invocation
+
+    // Discover the purpose from the provider:
+    if (resolved.config.providers && Array.isArray(resolved.config.providers)) {
+      const providers = await loadApiProviders(resolved.config.providers);
+      targetPurpose = await doTargetPurposeDiscovery(providers[0]);
```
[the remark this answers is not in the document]
sklein12: i had the same question, did you intend to overwrite the existing purpose if it exits

will-holley:
Redteam integration tests are failing because the corresponding Cloud task has not yet been deployed

examples/redteam-purpose-generation/README.md
```diff
@@ -0,0 +1,37 @@
+# Red Team: Target Application Purpose Auto-Discovery Example
```
mldangelo: You should remove this example and merge this into one or more  of the existing examples

site/docs/usage/command-line.md
```diff
@@ -7,27 +7,24 @@ sidebar_label: Command line
...
+- `auth` - Manage authentication
+- `cache` - Manage cache
+- `config` - Edit configuration settings
+- `debug` - Display debug information for troubleshooting
+- `delete` - Delete various resources
+- `export` - Export an eval record to a JSON file
+- `generate` - Generate synthetic data
+- `feedback` - Send feedback to the promptfoo developers
+- `import` - Import an eval record from a JSON file
+- `list` - List various resources
```
mldangelo: Can you keep the detailed information here?

mldangelo (commented):
Nice, requires documentation!

site/docs/usage/command-line.md
will-holley: Sure. For context, I copy and pasted this from the CLI.

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+
+/**
+ * Queries Cloud for the purpose-discovery logic, sends each logic to the target,
+ * and summarizes the results.
+ *
+ * @param target - The target API provider.
+ * @param maxTurns - The maximum number of turns to run the discovery process.
+ * @returns The purpose of the target.
+ */
+export async function doTargetPurposeDiscovery(
```
mldangelo, line 72: We need a nice behavior for if remote generation is disabled. You should talk about this in.the docs too

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+    .option('-t, --target <id>', 'UUID of a Cloud-defined target to run the discovery on')
+    .option('--preview', 'Preview discovery results without writing to an output file', false)
+    .addOption(
+      new Option(
+        '--turns <turns>',
+        'A maximum number of turns to run the discovery process. Lower is faster but less accurate.',
+      ).argParser(Number.parseInt),
+    )
+
+    .option('--env-path <path>', 'Path to a custom .env file')
```
mldangelo: You can remove this because we add it to every command by default

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+    .option('-v, --verbose', 'Show debug logs')
```
mldangelo:
```suggestion
```

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+        return;
+      }
+
+      // Set up the environment:
+      setupEnv(args.envPath);
+
+      // Set up logging:
+      if (args.verbose) {
+        setLogLevel('debug');
+      }
```
mldangelo:
```suggestion
```

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+
+      // Record telemetry:
+      telemetry.record('command_used', {
+        name: 'redteam discover',
+      });
+
+      // Always use remote for discovery
+      cliState.remote = true;
```
mldangelo: Don't mutate the global state

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+        if (!config) {
+          throw new Error(`Config is invalid at ${fallbackConfigPath}`);
+        }
+
+        if (!config.providers) {
+          throw new Error('Config must contain at least one target or provider');
+        }
+
+        const providers = await loadApiProviders(config.providers);
+        target = providers[0];
```
mldangelo, line 280: Versions of this exist in other parts of the code base. Would be better if you could re-use

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,318 @@
...
+
+          if (args.output === DEFAULT_OUTPUT_PATH) {
+            args.output = path.relative(process.cwd(), DEFAULT_OUTPUT_PATH);
+          }
+
+          logger.debug(`Writing purpose to ${args.output}`);
+
+          // TODO: Burp support:
+          if (args.output.endsWith('.burp')) {
+            throw new Error('Burp support is not yet implemented');
```
mldangelo: Why bother to do this at all?

src/redteam/commands/discover.ts
will-holley: Did you already add the global env flag?

src/redteam/commands/discover.ts
will-holley: What do you suggest alternatively?

src/redteam/commands/discover.ts
will-holley: Because remote generation is a necessary prerequisite, I would think the command should log a warning and return immediately

src/redteam/commands/discover.ts
mldangelo: Just don't check for .burp extensions?

examples/redteam-purpose-generation/README.md
mldangelo: You should remove this example. It's not useful on it's own given that purpose generation is now a default behavior. Please review the other redteam examples instead and see if they still work with is feature or if they or the documentation around them need to be updated.

src/redteam/commands/generate.ts
```diff
@@ -182,7 +195,7 @@ export async function doGenerateRedteam(
     entities: redteamConfig?.entities,
     plugins,
     provider: redteamConfig?.provider || options.provider,
-    purpose: redteamConfig?.purpose || options.purpose,
+    purpose: generatedPurpose ?? redteamConfig?.purpose ?? options.purpose,
```
sklein12, line 203: @will-holley i thought we were going to merge them?

sklein12: oh nm I see it in the discover code

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+                  : purpose,
+            };
+            writePromptfooConfig(existingYaml as UnifiedConfig, args.output!);
+          } else {
+            // Create a new config file with the purpose.
+            writePromptfooConfig({ redteam: { purpose } } as UnifiedConfig, args.output!);
+          }
+
+          logger.info(`\nPurpose written to ${chalk.italic(args.output)}`);
+        }
```
sklein12, line 353: I think we should output what we discovered to the log terminal at the `info` level. We should show the user what we found

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+          if (args.output === DEFAULT_OUTPUT_PATH) {
+            args.output = path.relative(process.cwd(), DEFAULT_OUTPUT_PATH);
+          }
+
+          logger.debug(`Writing purpose to ${args.output}`);
+
+          // TODO: Burp support:
+          if (args.output!.endsWith('.burp')) {
+            throw new Error('Burp support is not yet implemented');
+          }
```
sklein12: i would just remove this as a supported format

sklein12 (approved):
looks great, let's get it out

examples/redteam-purpose-generation/README.md
sklein12: agreed!

site/docs/usage/command-line.md
```diff
@@ -7,27 +7,27 @@ sidebar_label: Command line
 
 The `promptfoo` command line utility supports the following subcommands:
 
-- `init [directory]` - Initialize a new project with dummy files.
```
mldangelo: Please revert or move these changes to another PR.

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+export const ArgsSchema = z
+  .object({
+    config: z.string().optional(),
+    output: z.string().optional(),
+    target: z.string().optional(),
+    preview: z.boolean(),
+    turns: z.number().optional(),
+    overwrite: z.boolean(),
+  })
+  // Config and target are mutually exclusive:
```
mldangelo, line 34: You should explain why in the comments instead of restating the logical condition

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+  })
+  // if `preview` is false, `output` must be provided:
+  .refine((data) => !(data.preview === false && !data.output), {
+    message: 'If preview is false, output must be provided!',
+    path: ['preview', 'output'],
+  });
+
+type Args = z.infer<typeof ArgsSchema>;
+
+const DEFAULT_OUTPUT_PATH = 'redteam.yaml';
```
mldangelo: This would be a great thing to put into our redteam constants file.

src/redteam/commands/discover.ts
mldangelo: Set 25 as the default in your zod schema! How is this nullable?

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+import invariant from '../../util/invariant';
+import { getRemoteGenerationUrl } from '../remoteGeneration';
+
+export const ArgsSchema = z
+  .object({
+    config: z.string().optional(),
+    output: z.string().optional(),
+    target: z.string().optional(),
+    preview: z.boolean(),
+    turns: z.number().optional(),
```
mldangelo, line 31:
```suggestion
    turns: z.number().optional().default(25),
```

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+ * Queries Cloud for the purpose-discovery logic, sends each logic to the target,
+ * and summarizes the results.
+ *
+ * @param target - The target API provider.
+ * @param maxTurns - The maximum number of turns to run the discovery process.
+ * @returns The purpose of the target.
+ */
+export async function doTargetPurposeDiscovery(
+  target: ApiProvider,
+  maxTurns: number = 100,
```
mldangelo: Should this be the same as 25 below?

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+    hideCursor: true,
+  });
+
+  pbar.start(
+    // fallback: estimate of 25 turns
+    maxTurns ?? 25,
+    0,
+  );
+
+  while (true) {
```
mldangelo, line 89: Infinite loops are bad. Consider while turns < max trurns or even a for loop with an early break

src/redteam/commands/discover.ts
```diff
@@ -0,0 +1,355 @@
...
+      if (pbar) {
+        pbar.increment();
+        pbar.stop();
+      }
+      logger.info(`\nPurpose:\n\n${chalk.green(purpose)}\n`);
+      return purpose as string;
+    } else {
+      if (!question) {
+        logger.error(`Failed to discover purpose: ${res.statusText}`);
+        process.exit(1);
```
mldangelo, line 116: prefer proces.exitCode = 1;

will-holley: That's functionally different; on error the process should be terminated.

PR #3914, opened 2025-05-07
fix: Increase timeout for python path validation

hansmaaherra:
The python path validation is causing flakiness on my computer, depending on level of concurrency.  The default concurrency (4) gives med about 90% success rate, the rest gets "Error: Python 3 not found". If I turn up the concurrency to 10 I get 0% success.

This is mostly for redteam plugins, regular "evals" are fine.

I'm using a MacBook Pro from 2021 with a M1 Pro CPU, and I use asdf and pyenv.

## Summary by Sourcery

Increase timeout for Python path validation to improve reliability

Bug Fixes:
- Increased timeout for Python path detection to reduce flaky validation errors

Enhancements:
- Extended timeout from 250ms to 2500ms to accommodate slower systems and environments

### DIFF
--- a/src/python/pythonUtils.ts
+++ b/src/python/pythonUtils.ts
@@ -20,7 +20,7 @@ export async function tryPath(path: string): Promise<string | null> {
     const result = await Promise.race([
       execAsync(`${path} --version`),
       new Promise<never>((_, reject) =>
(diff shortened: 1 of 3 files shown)


mldangelo:
Thank you @hansmaaherra!

PR #3920, opened 2025-05-07
refactor: improve self hosting environment variable handling

mldangelo:
## Summary by Sourcery

Refactor environment variable handling and improve modularity of constants and sharing-related functions

New Features:
- Added a new function to check idempotency support for sharing

Enhancements:
- Converted static constants to dynamic getter functions for environment-based configuration
- Improved flexibility of environment variable retrieval for sharing and API base URLs

Chores:
- Updated test cases to work with new function-based approach
- Simplified environment variable access patterns


src/constants.ts
```diff
@@ -6,25 +6,32 @@ export const VERSION = packageJson.version;
...
-  getEnvString('NEXT_PUBLIC_PROMPTFOO_REMOTE_API_BASE_URL') ||
-  getEnvString('NEXT_PUBLIC_PROMPTFOO_BASE_URL') ||
-  getEnvString('PROMPTFOO_REMOTE_API_BASE_URL') ||
-  `https://api.promptfoo.app`;
-
-export const DEFAULT_SHARE_VIEW_BASE_URL = getEnvString(
-  'PROMPTFOO_SHARING_APP_BASE_URL',
-  `https://promptfoo.app`,
-);
+export function getShareApiBaseUrl(): string {
```
will-holley, line 12: What's the purpose of switching these to getter functions?

src/constants.ts
```diff
@@ -6,25 +6,32 @@ export const VERSION = packageJson.version;
...
+  return (
+    // TODO(ian): Backwards compatibility, 2024-04-01
+    getEnvString('NEXT_PUBLIC_PROMPTFOO_REMOTE_API_BASE_URL') ||
+    getEnvString('NEXT_PUBLIC_PROMPTFOO_BASE_URL') ||
+    getEnvString('PROMPTFOO_REMOTE_API_BASE_URL') ||
+    `https://api.promptfoo.app`
```
will-holley: This should be hardcoded as a constant and the constant should be used here and in the idempotence check.

src/util/server.ts
```diff
@@ -57,8 +57,8 @@ export async function checkServerRunning(port = DEFAULT_PORT): Promise<boolean>
 }
 
 export async function openBrowser(
-  browserBehavior: BrowserBehavior,
-  port = DEFAULT_PORT,
+  browserBehavior: BrowserBehavior = BrowserBehavior.ASK,
```
will-holley, line 60: This change is unrelated

test/integrations/huggingfaceDatasets.test.ts
```diff
@@ -9,20 +9,20 @@ jest.mock('../../src/fetch', () => ({
 }));
 
 describe('huggingfaceDatasets', () => {
-  let originalHfApiToken: string | undefined;
+  // Save original environment variables
+  const originalEnv = { ...process.env };
```
will-holley: How is this change related?

src/constants.ts
mldangelo: The getter guarantees that the .env file will have been red before these are invoked. With the way a lot of these are defined now, either it doesn’t happen at all or they don’t work if we override the default .env path with —env-file

test/integrations/huggingfaceDatasets.test.ts
mldangelo: This test started failing when I changed how the env variables were read

will-holley (dismissed):
please switch `https://api.promptfoo.app` to a const; then lgtm.

sklein12 (dismissed):

src/share.ts
```diff
@@ -482,3 +482,19 @@ export async function hasEvalBeenShared(eval_: Eval): Promise<boolean> {
...
+ * Idempotency is supported when:
+ * 1. Using cloud mode (cloudConfig.isEnabled)
+ * 2. Using the default promptfoo.app service
+ *
+ * Idempotency is NOT supported when using custom self-hosted URLs.
+ */
+export function isIdempotencySupported(): boolean {
+  return (
+    cloudConfig.isEnabled() ||
+    (getShareViewBaseUrl() === getDefaultShareViewBaseUrl() &&
```
will-holley: You can remove this line.

PR #3941, opened 2025-05-08
build: add Node.js 24 support

mldangelo:


mldangelo:
Node 24 tests will continue to fail until there's a new version of google-auth-library. Going to convert to a draft

PR #3945, opened 2025-05-09
chore: Revert "feat(cli): Adds global `--verbose` option"

sklein12:
Reverts promptfoo/promptfoo#3931 

We discussed and don't want to add a flag that goes before the sub commands


sklein12:
@will-holley @mldangelo reverting as we talked about

PR #3950, opened 2025-05-09
chore: add global verbose option to all commands recursively

mldangelo:


raiderss:
Very clean code! This reminds me of our development standards at Eyes Store FiveM. Would love to collaborate!

PR #3954, opened 2025-05-09
chore: Clean up providers for redteam generate

faizanminhas:
Fix cleanup race condition with MCP server initialization

Resolve issue where cleanup calls fail because providers haven't completed MCP server initialization. This initialization happens during provider loading for redteam generate, even when providers aren't explicitly called.

This minimal fix addresses the race condition. An alternative approach could be preventing MCP server instantiation during generate-only calls, but the current solution is sufficient and safe.


mldangelo (approved):
Looks good! I would love to see a test

PR #3956, opened 2025-05-09
chore: better error handling for remote generation

sklein12:


faizanminhas (approved):
omg thank you, this confused me this morning

sklein12:
@faizanminhas since it's throwing errors instead of failing silently 😮‍💨

test/redteam/plugins/index.test.ts
```diff
@@ -185,23 +185,23 @@ describe('Plugins', () => {
...
+      await expect(
+        plugin?.action({
+          provider: mockProvider,
+          purpose: 'test',
+          injectVar: 'testVar',
+          n: 1,
+          config: {},
+          delayMs: 0,
+        }),
+      ).rejects.toThrow('No test cases generated for contracts');
```
faizanminhas, line 204: my opinion: I like clear error messages and exceptions when things fail.

sklein12: i made a poor decision and reverted. We have existing behavior that allows a redteam to continue and just fails the current plugin, we can't override that

faizanminhas: Wondering if that's desirable? If I configure a plugin, I'd expect it to work, or to inform me if it doesn't work. If I have 30 plugins and 2 failed, theoretically I'd wanna know without having to comb through logs?

PR #3962, opened 2025-05-09
fix: no proxy values should take priority in fetch

MrFlounder:


sklein12:
@MrFlounder I didn't realize it would take all of this logic to do it. Should we use that library instead?

MrFlounder:
@sklein12 that would be a much bigger refactor though - the types are different and I think we need to touch many files

MrFlounder:
updated the logic and tested manually - this should work, will fix the tests after lunch

I don't love it but it would take much longer if we switch package

src/fetch.ts
```diff
@@ -48,7 +48,77 @@ export function sanitizeUrl(url: string): string {
...
+      return false;
+    }
+
+    for (const pattern of noProxyList) {
+      if (hostname === pattern) {
+        return false;
+      }
+
+      // Check for wildcard domain matches (*.example.com)
+      if (pattern.startsWith('*.') && hostname.endsWith(pattern.substring(1))) {
```
sklein12: I don't think this works properly

sklein12:
```
> "*.asdf.com".substring(1)
'.asdf.com'
> 
```

src/fetch.ts
```diff
@@ -48,7 +48,77 @@ export function sanitizeUrl(url: string): string {
...
+        return false;
+      }
+
+      // Check for domain suffix matches (.example.com matches subdomain.example.com)
+      if (pattern.startsWith('.') && hostname.endsWith(pattern)) {
+        return false;
+      }
```
sklein12: should `.example.com`  match `example.com`? it doesn't in this case

src/fetch.ts
```diff
@@ -48,7 +48,77 @@ export function sanitizeUrl(url: string): string {
...
+    }
+
+    return true;
+  } catch (e) {
+    logger.debug(`URL parsing failed in shouldProxy: ${e}`);
```
sklein12: should this be a warn? is this actionable if this errors?

src/fetch.ts
sklein12: does that work if the hostname doesn't have a subdomain? or does it always start with a `.`?

mldangelo (approved):
This looks good! I really hope it works. Thank you @MrFlounder

sklein12 (approved):
works great!

mldangelo:
Merging it so we can cut a release this weekend!

PR #3965, opened 2025-05-09
chore: Better error handling and logging for remote generation

sklein12:
We need better error handling when remote generation fails and we need to ensure we are always returning an array

### DIFF
--- a/src/redteam/plugins/index.ts
+++ b/src/redteam/plugins/index.ts
@@ -96,8 +96,7 @@ async function fetchRemoteTestCases(
       REQUEST_TIMEOUT_MS,
     );
     if (status !== 200 || !data || !data.result || !Array.isArray(data.result)) {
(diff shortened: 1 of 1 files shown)


src/redteam/plugins/index.ts
```diff
@@ -95,6 +95,11 @@ async function fetchRemoteTestCases(
       },
       REQUEST_TIMEOUT_MS,
     );
+    if (status !== 200 || !data || !data.result || !Array.isArray(data.result)) {
+      const errorMessage = `Error generating test cases for ${key}: ${statusText} ${JSON.stringify(data)}`;
+      logger.error(errorMessage);
```
faizanminhas:
```suggestion
      logger.error(`Error generating test cases for ${key}: ${statusText} ${JSON.stringify(data)}`);
```

faizanminhas (dismissed):

PR #3968, opened 2025-05-10
fix(evaluator): correctly count named scores based on contributing assertions

timothyasp:
Previously, `namedScoresCount` was incremented by 1 for each test case that included a named score, regardless of how many assertions contributed to that score, which led to inaccurate `namedScoreCounts` in result exports if multiple assertions with the same metric name existed.


typpo (approved):
Great catch!

mldangelo (approved):
Thank you! Send me an email at michael @ promptfoo.dev for some swag!

PR #3974, opened 2025-05-11
docs: how to create inline assertions for pacakge users

billybonks:
As a user i would prefere to define my assertions inline versus in a seperate file imported using a path.

This makes the code cleaner and allows me to do things like currying the assertion fuction

in order to do this i had to read through the code, so to simplify for the next user i added documentation

coersion happens here:
https://github.com/promptfoo/promptfoo/blob/46e784ec001d29920bac5bfeb25903265c4d6d65/src/assertions/index.ts#L220


mldangelo:
Thank you!

PR #3975, opened 2025-05-11
fix: divergent behviour http

billybonks:
This pr is more about having a conversation with examples depending on the outcome it can be closed or merged, i have also added descriptions to the commits explaining some specifics.

I have split the work into two commits.
   1. indicating what the current behavior is.
   2. Changing the behavior to the recommendation of the pr.

When i was using prompt foo i put a lot of effort to return correct status codes so that the behavior was as expected. and was surprised when 500 was not reflecting correctly, i figured it was because of `validateStatus`.

But while exploring the code i had a hunch that if i mess with `PROMPTFOO_RETRY_5XX` it will have a divergent behavior.

I do think that http provider should respect HTTP status codes as they are intended, if a developer wants do other stuff with 500 they can. 

i found the pr that makes the change https://github.com/promptfoo/promptfoo/pull/2712, but it does not explain why this change was made?

### DIFF
--- a/src/providers/http.ts
+++ b/src/providers/http.ts
@@ -511,12 +511,8 @@ export function determineRequestBody(
 }
 
 export async function createValidateStatus(
-  validator: string | ((status: number) => boolean) | undefined,
+  validator: string | ((status: number) => boolean),
 ): Promise<(status: number) => boolean> {
-  if (!validator) {
-    return (status: number) => true;
-  }
-
   if (typeof validator === 'function') {
     return validator;
   }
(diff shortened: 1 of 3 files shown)


mldangelo (commented):
This is a great change, thank you @billybonks. I am going to sync with the rest of the team tomorrow and try to find out if we'll break anything if we merge this.

mldangelo:
I've addressed all the review feedback:

1. ✅ Changed `||` to `??` (nullish coalescing) for validateStatus default to prevent empty strings or false from being treated as missing
2. ✅ Fixed typos in test descriptions:
   - "rety" → "retry" 
   - "configued" → "configured"
3. ✅ Added comprehensive test coverage for 2xx status code range:
   - Tests 200, 201, 299 (success cases)
   - Tests 199, 300 (boundary failure cases)
   - Tests 400, 500 (other failure cases)

All tests are passing. Ready for review\!

mldangelo:
This is a great feature. We intentionally made the change because many APIs break spec, especially with guardrail responses. In a follow up PR I will clearly document this. Thank you @billybonks

Issue #3991, opened 2025-05-13
Relative path resolution fails for file providers


pablosjv:
**Describe the bug**

Promptfoo eval is not able to resolve the path for the providers relatively to the `promptfooconfig.yaml` file location in the latest versions

**To Reproduce**

Promptfoo version:

```console
❯ promptfoo --version
(node:53198) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
0.112.5
```

Project structure:

```console
❯ tree -L 2
.
├── evaluation
│   └── promptfooconfig.yaml
└── providers
    └── example.py
```

`evaluation/promptfooconfig.yaml`:

```yaml
providers:
  - id: 'file://providers/example.py'
    label: 'Debug 0 level'
  - id: 'file://../providers/example.py'
    label: 'Debug 1 level'
  - id: 'file://../../providers/example.py'
    label: 'Debug 2 levels'
  - id: 'file://../../../providers/example.py'
    label: 'Debug 3 levels'

prompts:
  - "Who are you?"

tests:
  - assert:
    - type: icontains
      value: "?"
```

`providers/example.py`:

```python
from typing import Any


def call_api(prompt: str, options: dict[str, Any], context: dict[str, Any]):
    def call_llm(prompt):
        return f"Stub response for prompt: {prompt}"

    output = call_llm(prompt)

    result = {
        "output": output,
    }

    return result
```

Running `promptfoo eval` command gives several errors for like: `ERROR] Error: ENOENT: no such file or directory`. Interestingly, these are the files resolved for each level:

- Level 0: `file://providers/example.py`
    - `/tmp/promptfoo-debug/evaluation/providers/example.py`
    - ✅ This the expected
- Level 1: `file://../providers/example.py`
    - `/tmp/promptfoo-debug/evaluation/providers/example.py`
    - ❌ The same resolution as the first one
- Level 2: `file://../../providers/example.py`
    - `/tmp/promptfoo-debug/providers/example.py`
    - ❓ This passes the test and can be resolved, but it is not expected
- Level 3: `file://../../../providers/example.py`
    - `/tmp/providers/example.py'`
    - ❓ The behavior is kept consistent from the previous one, but again it is not expected

**Expected behavior**

**System information:**

```console
❯ promptfoo debug --config evaluation
...
{
  "version": "0.112.5",
  "platform": {
    "os": "darwin",
[...]

PR #4001, opened 2025-05-14
fix: Resolve paths relative to promptfooconfig when not cloud config

faizanminhas:
Should resolve https://github.com/promptfoo/promptfoo/issues/3991

## Summary
The changes in https://github.com/promptfoo/promptfoo/pull/3805 resolved file paths to cwd for cloud configs, but also for relative paths when not a cloud config, if a config.basePath was not set. This should've only applied for cloud configs, since path resolution is done already in https://github.com/promptfoo/promptfoo/blob/6bf829bb5f4214909c96e9dba94793ea8ba98b0e/src/util/index.ts#L415 for scripts. As a result, we would now be applying the base path twice, causing some weird effects, as seen in the issue. 

<details>
<summary>Test Process</summary>

input:
```
providers:
  - id: 'file://providers/example.py'
  - id: 'file://../providers/example.py'
  - id: 'file://../../providers/example.py'
  - id: 'file://../../../providers/example.py'
```

file structure:
```
├── evaluation
│   └── promptfooconfig.yaml
└── providers
    └── example.py
```

When evaluating from `/workspace/promptfoo`,

Previous resolution behaviour:
```
*current*
filePath /workspace/promptfoo/evaluation/evaluation/providers/example.py
filePath /workspace/promptfoo/evaluation/providers/example.py
filePath /workspace/promptfoo/providers/example.py
filePath /workspace/providers/example.py
```

With this PR: (i.e. now relative to promptfooconfig.yaml, as previously)
```
filePath /workspace/promptfoo/evaluation/providers/example.py
filePath /workspace/promptfoo/providers/example.py
filePath /workspace/providers/example.py
filePath /providers/example.py
```
</details>

### DIFF
--- a/src/providers/scriptBasedProvider.ts
+++ b/src/providers/scriptBasedProvider.ts
@@ -46,7 +46,7 @@ export function createScriptBasedProviderFactory(
 
       // If using a cloud config, always use process.cwd() instead of context.basePath
       const isCloudConfig = providerOptions.config?.isCloudConfig === true;
-      const resolvedPath = getResolvedRelativePath(scriptPath, context.basePath, isCloudConfig);
+      const resolvedPath = getResolvedRelativePath(scriptPath, isCloudConfig);
 
       return new providerConstructor(resolvedPath, providerOptions);
     },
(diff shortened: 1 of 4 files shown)


mldangelo (approved):
Thank you for fixing this!

PR #4018, opened 2025-05-15
chore: add headers support for url remote mcp servers

faizanminhas:
Adding an option to specify headers, and passing those into the http & sse MCP clients


mldangelo (approved):
Thanks!

PR #4024, opened 2025-05-15
feat(redteam): Off-Topic Plugin

AISimplyExplained:
The plugin specifically tries to make models perform tasks outside their domain


faizanminhas:
I think the way this is written, it won't work with strategies like `jailbreak` or `crescendo`, right? I guess that's not a big deal, since it's already multi-turn.

src/redteam/plugins/offTopic.ts
```diff
@@ -0,0 +1,178 @@
...
+      injectVar,
+      ...config,
+    };
+
+    logger.debug(`[OffTopic] Initialized with config: ${JSON.stringify(this.config)}`);
+  }
+
+  // This is required by the abstract base class but isn't used in our implementation
+  protected async getTemplate(): Promise<string> {
+    return ''; // Empty template since we're not using it
```
will-holley: Should this function be called? If not, raise an invariant.

src/redteam/plugins/offTopic.ts
```diff
@@ -0,0 +1,178 @@
...
+  protected getAssertions(): Assertion[] {
+    return [
+      {
+        type: PLUGIN_ID,
+        metric: 'OffTopic',
+      },
+    ];
+  }
+
+  async callApi(
```
will-holley: There aren't any references to this method outside of the unit tests. Was it left here in error? If so, please remove.

src/redteam/plugins/offTopic.ts
```diff
@@ -0,0 +1,178 @@
...
+        metadata: {
+          error: String(error),
+        },
+      };
+    }
+  }
+
+  async generateTests(n: number): Promise<TestCase[]> {
+    return Array.from({ length: n }, () => {
+      const template = this.getRandomTemplate();
```
will-holley: A single use class method isn't necessary.

```ts
const randomTemplate = INITIAL_PROMPT_TEMPLATES[Math.floor(Math.random() * INITIAL_PROMPT_TEMPLATES.length)]

File: src/redteam/plugins/offTopic.ts [hunk 4]
```diff
@@ -0,0 +1,178 @@
...
+        },
+        metadata: {
+          pluginId: this.id,
+          description: template.description,
+        },
+      };
+    });
+  }
+
+  private getRandomTemplate(): InitialPromptTemplate {
```
will-holley: See above comment on necessity and remove.

src/redteam/providers/offTopic.ts
```diff
@@ -0,0 +1,192 @@
+import { randomBytes } from 'crypto';
+import logger from '../../logger';
+import type { ApiProvider, CallApiContextParams, CallApiOptionsParams } from '../../types';
+import { getRemoteGenerationUrl } from '../remoteGeneration';
+import type { Message } from './shared';
+import { messagesToRedteamHistory } from './shared';
+
+// Match server's MAX_ROUNDS constant
```
will-holley: This value is configurable, not constant on the server; preferably pass this value to the server within the call payload to ensure it stays in sync across the service boundary.

src/redteam/providers/offTopic.ts
```diff
@@ -0,0 +1,192 @@
...
+      return content;
+    }
+    try {
+      return JSON.stringify(content);
+    } catch {
+      return String(content);
+    }
+  }
+
+  async callApi(
```
will-holley, line 53: please add a docstring

src/redteam/providers/offTopic.ts
```diff
@@ -0,0 +1,192 @@
...
+          messages: this.messages,
+          sessionId: this.sessionId,
+          roundsCompleted: this.roundNum,
+          redteamHistory: messagesToRedteamHistory(this.messages),
+        };
+
+        // Check if conversation is complete
+        isComplete =
+          isComplete ||
+          this.state.currentPhase >= 3 || // End after phase 3
```
will-holley: What is "phase 3"?

src/redteam/providers/offTopic.ts
```diff
@@ -0,0 +1,192 @@
...
+  deviationAttempts: number;
+  currentPhase: number;
+  lastDeviationSuccess: boolean;
+}
+
+export class ServerOffTopicProvider implements ApiProvider {
+  private sessionId: string;
+  private messages: Message[] = [];
+  private roundNum: number = 0;
+  private state: ServerState = {
```
will-holley: Why are `messages`, `roundNum`, `state`, and `originalPurpose` scoped?

src/redteam/providers/offTopic.ts
```diff
@@ -0,0 +1,192 @@
...
+
+interface ServerState {
+  originalTopic: string;
+  deviationAttempts: number;
+  currentPhase: number;
+  lastDeviationSuccess: boolean;
+}
+
+export class ServerOffTopicProvider implements ApiProvider {
+  private sessionId: string;
```
will-holley: How is `sessionId` used?

sklein12:
@AISimplyExplained I don't understand the pattern of `callApi` in side of a plugin, can you explain it?  How does that work with our other strategies?

AISimplyExplained:
Wanted to make it a strategy exempt multi turn plugin like tool discovery multi turn

src/redteam/providers/offTopic.ts
AISimplyExplained: where it verifies deviation

PR #4035, opened 2025-05-16
chore(cli/redteam/discover): Sets default turn count to 5

will-holley:


sklein12 (approved):

mldangelo (approved):

src/redteam/commands/discover.ts
```diff
@@ -61,7 +61,8 @@ type Args = z.infer<typeof ArgsSchema>;
 
 // A larger turn count is more accurate (b/c more probes) but slower.
 // TODO: Optimize this default to balance quality/runtime using the Discover eval.
-const DEFAULT_TURN_COUNT = 50;
+// NOTE: Set to 5 because UI lacks ability to set the count.
```
mldangelo, line 64: why not add the ability to the ui?

will-holley: I didn't make the call.

PR #4038, opened 2025-05-16
chore(cli/redteam/discover): Miscellaneous improvements

will-holley:
- fixes redteam-rag example
- migrates from turns to question/followup

### DIFF
--- a/src/redteam/commands/discover.ts
+++ b/src/redteam/commands/discover.ts
@@ -212,7 +212,7 @@ export function discoverCommand(program: Command) {
     .option(
       '--turns <turns>',
       'A maximum number of turns to run the discovery process. Lower is faster but less accurate.',
-      (value) => Number.parseInt(value),
+      (value) => Number.parseInt(value, 10),
       DEFAULT_TURN_COUNT,
     )
     .action(async (rawArgs: Args) => {
--- a/src/redteam/commands/discover.ts
+++ b/src/redteam/commands/discover.ts
@@ -87,7 +87,14 @@ export async function doTargetPurposeDiscovery(
     hideCursor: true,
   });
 
-  pbar.start(maxTurns, turnCounter);
+  pbar.start(maxTurns, 0);
+
+  function onCompletion(purpose: string) {
+    pbar.increment();
+    pbar.stop();
+    logger.info(`\nPurpose:\n\n${chalk.green(purpose)}\n`);
+    return purpose;
+  }
 
   while (true) {
     const res = await fetchWithProxy(getRemoteGenerationUrl(), {
@@ -109,10 +116,7 @@ export async function doTargetPurposeDiscovery(
     };
 
     if (done) {
-      pbar.increment();
-      pbar.stop();
-      logger.info(`\nPurpose:\n\n${chalk.green(purpose)}\n`);
-      return purpose as string;
+      return onCompletion(purpose as string);
     } else {
       if (!question) {
         throw new Error(`Failed to discover purpose: ${res.statusText}`);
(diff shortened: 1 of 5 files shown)


src/redteam/commands/discover.ts
```diff
@@ -144,19 +104,50 @@ export async function doTargetPurposeDiscovery(
...
 
-    if (purpose) {
-      return purpose;
-    }
+      // Handle the edge case where the server did not call `done` when max turns are reached:
+      if (purpose) {
+        logger.info(`\nPurpose:\n\n${chalk.green(purpose)}\n`);
+        return { purpose: purpose as string, conversationHistory };
+      } else {
+        throw new Error(
```
mldangelo: This error message will not make sense to our users

src/redteam/commands/discover.ts
```diff
@@ -144,19 +104,50 @@ export async function doTargetPurposeDiscovery(
...
+    } else {
+      if (!question) {
+        throw new Error(`Failed to discover purpose: ${res.statusText}`);
+      }
 
-    throw new Error('Failed to discover purpose after exhausting maximum turns');
-  } finally {
-    // Ensure progress bar is always stopped, even if an exception occurs
-    if (pbar) {
-      pbar.stop();
```
mldangelo, line 157: don't remove this it's causing the tests to stay open

src/redteam/commands/discover.ts
```diff
@@ -144,19 +104,50 @@ export async function doTargetPurposeDiscovery(
...
+      // Call the target with the question:
+      const response = await target.callApi(question as string);
+      logger.debug(JSON.stringify({ question, output: response.output }, null, 2));
+      conversationHistory.push({ type: 'target', content: response.output });
+
+      pbar.increment();
     }
   }
+
+  throw new Error('Failed to discover purpose: max turns reached');
```
mldangelo: Can you remove this error?

mldangelo (commented):
Please run `npx jest discover --detectOpenHandles`

Running `promptfoo redteam discover` should pick up the default config similar to redteam generate (see main.ts) 

Would love to see a few updates to the discover tests with this. 

Thanks @will-holley!

src/redteam/commands/discover.ts
will-holley: It's deleting the outputted purpose.

src/redteam/commands/discover.ts
will-holley: This is an us error, not a user error – there's no action the user can take to correct for it – so it needs to make sense to us if reported by the users.

src/redteam/commands/discover.ts
will-holley: Why?

will-holley:
> Running promptfoo redteam discover should pick up the default config similar to redteam generate (see main.ts)

By default it _does_ pick up `promptfooconfig.yaml` in the current working directory and writes to `redteam.yaml` if no output path is specified. Are you referring to some other behavior?

mldangelo:
> > Running promptfoo redteam discover should pick up the default config similar to redteam generate (see main.ts)
> 
> By default it _does_ pick up `promptfooconfig.yaml` in the current working directory and writes to `redteam.yaml` if no output path is specified. Are you referring to some other behavior?

create a promptfooconfig.yaml in your working directory. Run `npm run local -- (or promptfoo) redteam discover`. It will log 

```
Invalid options:
  config.target: Either config or target must be provided!
```

change the command to `npm run local -- redteam generate` or `npm run local -- redteam run`. They each pick up the default config

PR #4039, opened 2025-05-16
feat: add copy as markdown button to doc pages

faizanminhas:
Screenshot 2025-05-16 at 10 03 19 AM

https://github.com/user-attachments/assets/5300665e-c16a-4c84-89b9-c09738dbcb88

dark mode:
Screenshot 2025-05-18 at 11 56 29 PM

### DIFF
--- a/site/package.json
+++ b/site/package.json
@@ -3,10 +3,9 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
-    "prebuild": "rm -rf static/docs && cp -r docs static/docs && sed -i '' '/---/,/---/d' static/docs/**/*.md",
     "docusaurus": "docusaurus",
-    "start": "npm run prebuild && NODE_ENV=development docusaurus start --port ${PORT:-3100}",
-    "build": "npm run prebuild && docusaurus build",
+    "start": "NODE_ENV=development docusaurus start --port ${PORT:-3100}",
+    "build": "docusaurus build",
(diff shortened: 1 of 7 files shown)


site/package.json
```diff
@@ -3,9 +3,10 @@
   "version": "0.0.1",
   "license": "MIT",
   "scripts": {
+    "prebuild": "rm -rf static/docs && cp -r docs static/docs && sed -i '' '/---/,/---/d' static/docs/**/*.md",
```
faizanminhas: maybe not the most elegant solution, but I just copy the contents of each markdown file and statically host it, which is then fetched during copy

mldangelo (commented):
This is sweet!

site/package.json
mldangelo: Look at how I implemented llms.txt!

faizanminhas: this is just what I was looking for, awesome

faizanminhas:
@mldangelo Updated with dark styles, and also configured docusaurus's dev server to serve the md files directly, so it should work in development now too. Also tested the build & serve to ensure that that still works.

mldangelo (approved):
Thanks!

will-holley:
Is the motivation making it easier to include pages within context windows?

faizanminhas:
@will-holley Yeah, I noticed a few other doc sites like openai & model context protocol both had these buttons, and they made it easy to interact with an LLM regarding a page.

PR #4041, opened 2025-05-16
chore(providers): Adds support for openai `codex-mini-latest`

will-holley:
https://platform.openai.com/docs/models/codex-mini-latest


src/providers/openai/util.ts
```diff
@@ -206,6 +206,20 @@ export const OPENAI_CHAT_MODELS = [
       output: 2 / 1e6,
     },
   })),
+  ...['o4-mini', 'o4-mini-2025-04-16'].map((model) => ({
```
will-holley, line 209: Note: Cursor Agent added this; can remove if desired.

mldangelo (dismissed):
Thanks!

PR #4044, opened 2025-05-16
feat(cli): track per provider token usage statistics

mldangelo:
This PR introduces tracking of token usage statistics per provider.

- Adds detailed token and character usage tracking.
- Implements CLI table display for usage summary.


src/commands/eval.ts
```diff
@@ -65,6 +67,116 @@ function showRedteamProviderLabelMissingWarning(testSuite: TestSuite) {
   }
 }
 
+// Add this function to display the token usage summary using cli-table3
```
typpo:
```suggestion
```

src/types/shared.ts
```diff
@@ -22,8 +22,18 @@ export const BaseTokenUsageSchema = z.object({
   // Request metadata
   numRequests: z.number().optional(),
 
-  // Detailed completion information
+  // Detailed completion token breakdown (o1 models)
```
mldangelo:
```suggestion
  // Detailed completion information
```

PR #4050, opened 2025-05-17
chore: Improve memory usage when sharing results

sklein12:
## Summary
- add batched sample and counting helpers on `Eval`
- stream results in `sendChunkedResults`

### DIFF
--- a/src/commands/eval.ts
+++ b/src/commands/eval.ts
@@ -268,6 +268,8 @@ export async function doEval(
       abortSignal: evaluateOptions.abortSignal,
     });
 
+    evalRecord.clearResults();
+
     const wantsToShare = cmdObj.share && config.sharing;
 
     const shareableUrl =
--- a/src/commands/eval.ts
+++ b/src/commands/eval.ts
@@ -268,6 +268,7 @@ export async function doEval(
       abortSignal: evaluateOptions.abortSignal,
     });
 
+    // Clear results from memory to avoid memory issues
     evalRecord.clearResults();
 
     const wantsToShare = cmdObj.share && config.sharing;
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -360,6 +360,39 @@ export default class Eval {
     }
   }
 
+  async fetchSampleResults(sampleSize: number = 100): Promise<EvalResult[]> {
+    const results: EvalResult[] = [];
+    let remaining = sampleSize;
+    for await (const batch of this.fetchResultsBatched(sampleSize)) {
+      if (batch.length >= remaining) {
+        results.push(...batch.slice(0, remaining));
+        break;
+      } else {
+        results.push(...batch);
+        remaining -= batch.length;
+      }
+      if (remaining <= 0) {
+        break;
+      }
+    }
+    return results;
+  }
+
+  async getResultsCount(): Promise<number> {
+    if (this.resultsCount > 0) {
+      return this.resultsCount;
+    }
+    const db = getDb();
+    const result = await db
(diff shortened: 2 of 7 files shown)


src/commands/eval.ts
```diff
@@ -268,6 +268,8 @@ export async function doEval(
       abortSignal: evaluateOptions.abortSignal,
     });
 
+    evalRecord.clearResults();
```
will-holley, line 272: Please add a comment here explaining _why_ results are being cleared.

src/models/eval.ts
```diff
@@ -69,6 +69,7 @@ export default class Eval {
   prompts: CompletedPrompt[];
   oldResults?: EvaluateSummaryV2;
   persisted: boolean;
+  _resultsLoaded: boolean;
```
will-holley:
```ts
private _resultsLoaded: boolean = false;
```

src/models/eval.ts
```diff
@@ -360,6 +362,35 @@ export default class Eval {
     }
   }
 
+  async fetchSampleResults(sampleSize: number = 100): Promise<EvalResult[]> {
```
will-holley: Why is this method preferable to sampling via

```ts
 const sampleResults = (await evalRecord.fetchResultsBatched(100).next()).value ?? [];
```

within `sendChunkedResults`?

src/models/eval.ts
```diff
@@ -360,6 +362,35 @@ export default class Eval {
...
+      if (remaining <= 0) {
+        break;
+      }
+    }
+    return results;
+  }
+
+  async getResultsCount(): Promise<number> {
+    const db = getDb();
+    const result = await db
```
will-holley: Are you not getting this linter error?

image

src/models/eval.ts
```diff
@@ -392,9 +425,15 @@ export default class Eval {
       return this.oldResults.results;
     }
     await this.loadResults();
+    this._resultsLoaded = true;
     return this.results;
   }
 
+  clearResults() {
+    this.results = [];
```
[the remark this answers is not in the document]
will-holley: This

src/share.ts
```diff
@@ -107,32 +108,11 @@ function findLargestResultSize(results: any[], sampleSize: number = 1000): numbe
...
-  const chunks: any[][] = [];
-  for (let i = 0; i < results.length; i += estimatedResultsPerChunk) {
-    chunks.push(results.slice(i, i + estimatedResultsPerChunk));
-  }
-
-  return chunks;
-}
-
-async function sendInitialEvalData(evalRecord: Eval, url: string, headers: Record<string, string>) {
+async function sendInitialEvalData(
```
will-holley: I'm not sure what "initial" means in this context; a doc-string would be helpful here.

will-holley: Also when/why should this function be used?

src/models/eval.ts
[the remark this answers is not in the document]
sklein12: I'll clear up these variables then, results count was supposed to be a static count of how many tests we ran for this eval.

src/commands/eval.ts
sklein12: good call

src/models/eval.ts
sklein12: we use es2020 which doesn't understand `private` yet

src/models/eval.ts
sklein12: good call, this is redudent

src/models/eval.ts
sklein12: nope!

PR #4061, opened 2025-05-17
refactor: avoid process exit

mldangelo:
## Summary
- replace direct `process.exit()` calls with `process.exitCode`
- bail early from functions instead of exiting the process
- update tests for the new exit behavior

## Testing
- `npm run f` *(fails: Cannot find package)*
- `npm run l` *(fails: ESLint plugin missing)*
- `npm test` *(fails: jest not found)*


src/util/config/load.ts
```diff
@@ -549,7 +549,7 @@ export async function resolveConfigs(
     !hasProviders
   ) {
     logger.error('You must specify at least 1 provider (for example, openai:gpt-4o)');
-    process.exit(1);
+    throw new Error('You must specify at least 1 provider (for example, openai:gpt-4o)');
```
mldangelo:
```suggestion
    throw new Error('You must specify at least 1 provider (for example, openai:gpt-4.1)');
```

will-holley (commented):
This is a significant change. What is its motivation?

will-holley:
IMO treating each instance of `process.exit` as homogenous will lead to unintended side-effects and we should a) refactor `logger.error` and `process.exit` combinations to throw errors while b) maintaining `process.exit` by CLI-only logic (e.g. CLI commands).

src/commands/delete.ts
```diff
@@ -11,7 +11,7 @@ async function handleEvalDelete(evalId: string, envPath?: string) {
     logger.info(`Evaluation with ID ${evalId} has been successfully deleted.`);
   } catch (error) {
     logger.error(`Could not delete evaluation with ID ${evalId}:\n${error}`);
-    process.exit(1);
+    process.exitCode = 1;
```
sklein12, line 14: This is a massive change in behavior, the rest of the code will continue to execute? Does it work properly?

PR #4081, opened 2025-05-19
chore(docs): Handle index.md files for copy page

faizanminhas:
Just noticed that `index.md` files, such as `red-team/index.md` were not correctly handled for serving the md content in the docs site. This fixes that by mapping all `index.md` files to be `/directory.md` files, which should map with what the url indicates. Hopefully this doesn't break anything else 😃


mldangelo (approved):
NICE!

mldangelo (approved):
NICE

PR #4097, opened 2025-05-20
fix: Target purpose not making it into redteam config

sklein12:


src/redteam/commands/generate.ts
```diff
@@ -55,7 +55,7 @@ export async function doGenerateRedteam(
   let redteamConfig: RedteamFileConfig | undefined;
   const configPath = options.config || options.defaultConfigPath;
   const outputPath = options.output || 'redteam.yaml';
-  let finalPurpose = redteamConfig?.purpose ?? options.purpose;
+  let finalPurpose = options.purpose;
```
mldangelo, line 58: The way our CLI works, options.purpose should override the redteamConfig?.purpose if specified. 
I know we've been migrating away from passing in cli args, we can rethink the design here too.

src/redteam/commands/generate.ts
```diff
@@ -55,7 +55,7 @@ export async function doGenerateRedteam(
   let redteamConfig: RedteamFileConfig | undefined;
   const configPath = options.config || options.defaultConfigPath;
   const outputPath = options.output || 'redteam.yaml';
-  let finalPurpose = redteamConfig?.purpose ?? options.purpose;
```
sklein12, line 58: redteamConfig is not defined here, it doesn't get loaded until later in this file. It's instantiated on line 55 but it's undefined.

src/redteam/commands/generate.ts
```diff
@@ -89,6 +89,9 @@ export async function doGenerateRedteam(
     );
     testSuite = resolved.testSuite;
     redteamConfig = resolved.config.redteam;
+    if (redteamConfig?.purpose) {
```
mldangelo, line 92:
```suggestion
    if (redteamConfig?.purpose && !options.purpose) {
```

mldangelo (approved):

PR #4108, opened 2025-05-20
chore: add log

MrFlounder:


src/redteam/providers/crescendo/index.ts
```diff
@@ -498,11 +498,17 @@ export class CrescendoProvider implements ApiProvider {
           : response.output;
     const expectedKeys = ['generatedQuestion', 'rationaleBehindJailbreak', 'lastResponseSummary'];
 
+    const missingKeys: string[] = [];
     for (const key of expectedKeys) {
       if (!(key in parsedOutput)) {
         logger.warn(`[Crescendo] Missing key in response: ${key}`);
```
sklein12: I think you can remove this line since you're logging them all out below

src/providers/http.ts
```diff
@@ -671,6 +671,7 @@ export class HttpProvider implements ApiProvider {
           'Content-Type is application/json, but body is a string. This is likely to cause unexpected results. It should be an object or array.',
         );
       }
+      logger.warn(`[HTTP Provider] Body: ${body} headers: ${safeJsonStringify(headers)}`);
```
sklein12: Please include this with  message above so we know why we logged it

sklein12 (changes requested):

sklein12 (approved):

PR #4109, opened 2025-05-21
chore: update purpose example

MrFlounder:
image

image


mldangelo (commented):
Would be cool to use the default llm-rubric provider (or task endpoint) to judge the purpose

mldangelo (approved):

Issue #4091, opened 2025-05-20
why openai is requested here and how to change it to deepseek


blu3sh0rk:
why openai is requested here and how to change it to deepseek

blu3sh0rk:
![Image](https://github.com/user-attachments/assets/0b0053c9-a30f-4055-a54b-bd5ca6fede0d)

mldangelo:
You can override the redteam provider in your config with 

```yaml
redteam: 
  plugins: ...
  provider: `openouter:deepseek-r1`
```
 for example. Who is hosting deepseek for you?

blu3sh0rk:
The redteam part has been modified to the following but still reports an error. Could you please tell me what the problem is?
```
redteam:
  purpose: 'The objective of the application is: Electric Power Knowledge Assistant'
  plugins:
    - id: ascii-smuggling  # Tests vulnerability to Unicode tag-based instruction smuggling attacks
  strategies:
    - id: jailbreak  # Single-shot optimization of safety bypass techniques
    - id: jailbreak:composite  # Combines multiple jailbreak techniques for enhanced effectiveness
  providers:
  - id: deepseek:deepseek-chat
    config:
      temperature: 0.7
      max_tokens: 4000
      apiKey: sk-xxx
```

blu3sh0rk:
> You can override the redteam provider in your config with
> 
> redteam: 
>   plugins: ...
>   provider: `openouter:deepseek-r1`
> for example. Who is hosting deepseek for you?

base_url:https://api.siliconflow.cn

billybonks:
Hi @blu3sh0rk could you run your evaluation with the -v command `promptfoo redteam run -v --force --no-progress-bar` and paste the output so its easier to see where the error might be occurring.

mldangelo:
@blu3sh0rk - this question was answered above. You can override the redteam provider in your config with the `provider` key under `redteam`.

For DeepSeek via SiliconFlow:

```yaml
redteam:
  plugins: [...]
  provider:
    id: openai:deepseek-chat
    config:
      apiBaseUrl: https://api.siliconflow.cn/v1
      apiKey: your-key
```

Closing as answered. Feel free to open a new issue if you're still having trouble with the configuration.

PR #4110, opened 2025-05-21
fix: improve logging when inheriting from OpenAiChatCompletionProvider

billybonks:
There are many classes that extend from OpenAiChatCompletionsProvider, but all the logging is openai this openai that, which causes users to be confused because they set a provider that is not open ai.

This commit tries to make it align the output with what the user would expect

I have some failing tests but those require more work due to the environment setup

# Before
image

image

image

# After
image

image

image

### DIFF
--- a/src/providers/openai/chat.ts
+++ b/src/providers/openai/chat.ts
@@ -164,12 +164,12 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
     }
     if (this.requiresApiKey() && !this.getApiKey()) {
       throw new Error(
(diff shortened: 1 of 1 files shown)


billybonks:
Originated while debugging https://github.com/promptfoo/promptfoo/issues/4091, but this was not the error that he was seeing.

billybonks:
I found one bug with this, if it uses the open ai provider `apiKeyEnvar` resolves to undefined so will have to fix that first.

mldangelo:
Thank you!

Issue #4080, opened 2025-05-19
Enable sharing state between hooks and Custom Provider calls


53ningen:
**Is your feature request related to a problem? Please describe.**
When calling an API that requires authentication, I want to perform the authentication only once, cache the session/token in a variable, and reuse it across multiple tests.

However, currently there is no mechanism to share variables between hooks and individual Custom Provider calls. This makes it difficult to efficiently reuse state such as authentication sessions across tests.

For example, even if I retrieve authentication information in the beforeAll hook like this:

```python
# hooks.py

def extension_hook(hook_name: str, context: dict) -> None:
    if hook_name == "beforeAll":
        access_token = retrive_access_token()
```

There is no way to refer to that access_token from a PythonProvider call like this:

```python
# custom_provider.py

def call_api(prompt: str, options: Dict[str, Any], context: Dict[str, Any]):
    # I would like to take an access_token retrieved in beforeAll hook but there is no way
    access_token = ........
    res = request(access_token, prompt)
    return { "output": res.message }
```

**Describe the solution you'd like**
Ideally, I would like to have an object or mechanism that allows data (such as authentication information obtained in a beforeAll hook) to be referenced from within a PythonProvider call. This way, shared state can be easily accessed and reused across tests and providers.

**Describe alternatives you've considered**
None

**Additional context**
None

billybonks:
Do you expect your token to expire during the tests, or do you expect it to be valid for the entirety of your test suite?

53ningen:
Thank you for your response and providing such a wonderful library

> Do you expect your token to expire during the tests, or do you expect it to be valid for the entirety of your test suite? 

Token is valid throughout the entire test suite.

> have you tried anything to solve this problem, a very basic way would be to write it to a temporary file for the duration of the test then delete it.

Currently, I'm just handling this by writing the token to a temporary file, but I think it would be useful to have a context object that can be passed around throughout the test execution as a library feature. So I created this issue as a feature request.

billybonks:
It is definitely a deficiency with non JavaScript extensions, since they run in the same process as the evaluator itself so they can modify the objects passed in.

But it seems that @will-holley has proposed a pretty nice solution to the problem   👍🏿

53ningen:
Thank you so much for adding this feature: pr 4197, @.will-holley

will-holley:
@53ningen, you're welcome : )

PR #4112, opened 2025-05-21
chore(eval): Pass data returned from `beforeAll` and `beforeEach` hooks to `provider.callApi`

will-holley:
## Overview

- Addresses #4080 which is a reasonable expectation.
- Data returned by either `beforeAll` or `beforeEach` will get passed in to `provider.callApi` as `context.extensionHookOutputs.beforeAll` and `context.extensionHookOutputs.beforeEach` respectively.

## Test Instructions

Construct the following toy example:

### `promptfooconfig.yaml`

```yaml
prompts:
  - "placeholder"
providers:
  - 'file://provider.py'
  - 'file://provider.js'
tests:
  - assert:
      - type: contains
        value: "foobar"
extensions:
  - file://hooks.py:extension_hook
  ```

### `provider.py`

```py
def call_api(prompt, options, context):
    return { "output": context["extensionHookOutputs"]["beforeAll"][0]}
```

### `provider.js`

```js
module.exports = class TestProvider {
    id = () => 'javascript:provider.js:default';
  
    callApi = async (prompt, context, options) => {
      return {
        output: context.extensionHookOutputs.beforeEach[0],
      };
    };
  }
```

### `hooks.py`

```py
def extension_hook(hook_name: str, context: dict):
    if hook_name == "beforeAll":
        return "foobar"
    elif hook_name == "beforeEach":
        return "bazqux"
```

and run `promptfoo eval`.  Note that the Python provider passes and the Javascript provider fails.

## Next Steps

Once this PR is approved functionally, I'll update the documentation accordingly.


src/evaluator.ts
```diff
@@ -536,7 +539,11 @@ class Evaluator {
     const assertionTypes = new Set<string>();
     const rowsWithSelectBestAssertion = new Set<number>();
 
-    await runExtensionHook(testSuite.extensions, 'beforeAll', { suite: testSuite });
+    const beforeAllExtensionOutputs = hasDefinedExtensionHooks(testSuite.extensions)
+      ? await runExtensionHook(testSuite.extensions!, 'beforeAll', {
+          suite: testSuite,
+        })
+      : [];
```
faizanminhas, line 546: is empty array the right fallback here? I would've expected `undefined`

src/evaluator.ts
```diff
@@ -982,10 +996,12 @@ class Evaluator {
...
-        await runExtensionHook(testSuite.extensions, 'afterEach', {
-          test: evalStep.test,
-          result: row,
-        });
+        if (hasDefinedExtensionHooks(testSuite.extensions)) {
+          await runExtensionHook(testSuite.extensions!, 'afterEach', {
+            test: evalStep.test,
+            result: row,
+          });
+        }
```
faizanminhas, line 1004: wondering if we should move the check for `hasDefinedExtensionHooks` into `runExtensionHook`- would make the flow in this function cleaner and it seems like there must be a check for if `afterEach`/other hook is present in the extensions or not?

src/evaluatorHelpers.ts
```diff
@@ -285,28 +285,48 @@ export async function renderPrompt(
...
+ * Returns true if the extensions array is defined and has at least one element.
  * @param extensions - An array of extension paths, or null.
+ * @returns True if the extensions array is defined and has at least one element.
+ */
+export function hasDefinedExtensionHooks(extensions: string[] | null | undefined): boolean {
+  if (!extensions || !Array.isArray(extensions) || extensions.length === 0) {
+    return false;
+  }
+  return true;
+}
```
faizanminhas:
```suggestion
  return extensions && Array.isArray(extensions) && extensions.length > 0);
}
```

faizanminhas (approved):
LGTM!

src/evaluator.ts
will-holley: That's a good point. I set it up to be consistent with `extensionHookOutputs` while avoiding having undefined properties on `extensionHookOutputs` (because it gets confusing whether a property was set as `undefined` or never set in the first place).  `null` might be a suitable alternative but it evals the same as the empty array.

src/evaluator.ts
will-holley: That's actually how it was originally (returning an empty array). IMO, we should assume that whatever is returned from `runExtensionHook` – aside from `undefined` – was returned by the underlying hook call. I could move this check into `runExtensionHook` and return `undefined` but might then want to coerce that to avoid the `undefined` issue I mentioned above.

test/evaluatorHelpers.test.ts
```diff
@@ -50,8 +51,12 @@ jest.mock('../src/esm');
 jest.mock('../src/database', () => ({
   getDb: jest.fn(),
 }));
+
+// Mock `transform` to return a tuple of the extension name and hook name
```
mldangelo, line 55:
```suggestion
```

will-holley:
Closing in favor of https://github.com/promptfoo/promptfoo/pull/4197.

PR #4131, opened 2025-05-23
chore(redteam): add label strategyless plugins in redteam setup ui

mldangelo:


src/redteam/constants.ts
```diff
@@ -180,14 +180,23 @@ export type AdditionalPlugin = (typeof ADDITIONAL_PLUGINS)[number];
...
-  'pliny',
+// Agentic plugins that don't use strategies (standalone agentic plugins)
+export const AGENTIC_EXEMPT_PLUGINS = [
   'system-prompt-override',
   'tool-discovery:multi-turn',
-  'unsafebench',
 ] as const;
 
+// Dataset plugins that don't use strategies (standalone dataset plugins)
+export const DATASET_EXEMPT_PLUGINS = ['pliny', 'unsafebench'] as const;
```
mldangelo, line 190: @typpo flagging ig you have thoughts on this. unsafebench is an image dataset. Unsure why pliny is strategy exempt and wondering if we should treat all of the other datasets this way too.

PR #4136, opened 2025-05-23
fix: zod error when state.answer has object

MrFlounder:


src/redteam/commands/discover.ts
```diff
@@ -168,7 +167,13 @@ export async function doTargetPurposeDiscovery(
             2,
           )}`,
         );
-        state.answers.push(targetResponse.output);
+
+        const output =
```
mldangelo, line 171: Why can the output be an object or a string? I think we should try to solve this upstream.

mldangelo (approved):
Thanks!

src/redteam/commands/discover.ts
sklein12: Agreed

sklein12: You should work with @will-holley hes got another PR open around this

MrFlounder: Will do
Putting the patch in for now though

PR #4144, opened 2025-05-23
fix: Throw error on un-supported command - redteam run with a cloud target but no config

sklein12:


faizanminhas (approved):
Good call. Though I'm not sure how someone would get into this state, we could add support fairly easily?

PR #4145, opened 2025-05-23
fix: use current working directory for redteam file if loading from cloud

sklein12:


src/redteam/shared.ts
```diff
@@ -55,7 +55,9 @@ export async function doRedteamRun(options: RedteamRunOptions): Promise<Eval | u
 
   if (options.liveRedteamConfig) {
     // Write liveRedteamConfig to a temporary file
-    const tmpFile = path.join(os.tmpdir(), `redteam-${Date.now()}`) + '/redteam.yaml';
+    const filename = `redteam-${Date.now()}.yaml`;
+    const tmpDir = options.loadedFromCloud ? '' : os.tmpdir();
```
mldangelo, line 59: this change seems reasonable. Would be great to have a test that makes sure we don't accidentally break it

mldangelo (dismissed):

Issue #4031, opened 2025-05-16
evaluateOptions property gets deleted from .yaml file when it is edited through the web ui


welcomemax:
At the moment I use evaluateOptions.repeat for retest reliability

In yaml these options are not displayed, but are still used during eval

**To Reproduce**
1. Click on 'Edit YAML'
2. Insert
```yaml
evaluateOptions:
  repeat: 3
```
3. Save and see no evaluateOptions

Also a question, maybe there are plans to specify the number of repetitions for a specific test?

welcomemax:
Hi @mldangelo, thanks for the quick response — I'm looking forward to the new version!

As related, `defaultTest` is showing the same behavior — could you take a look at that as well?

adelmuursepp:
Hey @welcomemax I did the fix for `defaultTest` not showing upon save in the YAML editor. 
When testing I see the order of config can be confusing and we can get to that in a bit as well. 

I will leave the issue open for a bit, in case you have some extra feedback, and if not, will close soon. Thanks a lot!

adelmuursepp:
> Also on the topic — I think `env` should either always be hidden entirely or have keys like OPENAI_API_KEY, etc., stripped from it.
> 
> In my specific case: there's an assistant from another project, and the embedded key doesn’t work for it. For this assistant, we input a different key in env, but after running evals, another user might discover the key in the config.
> 
[...]

PR #4152, opened 2025-05-25
fix(webui): defaultTest shown in webui YAML editor

adelmuursepp:
Relates to https://github.com/promptfoo/promptfoo/issues/4031

Additionally noted during testing:
- Default ordering of the config is confusing, with defaultTest being at the bottom in edit YAML
- evaluateOptions is not shown in the specific eval page -> View YAML
- defaultTest recent addition is visible in the top in the specific eval page -> View YAML (description should be the top)

### DIFF
--- a/src/app/src/pages/eval-creator/components/RunTestSuiteButton.tsx
+++ b/src/app/src/pages/eval-creator/components/RunTestSuiteButton.tsx
@@ -63,9 +63,9 @@ const RunTestSuiteButton: React.FC = () => {
           if (progressData.status === 'complete') {
             clearInterval(intervalId);
             setIsRunning(false);
-
-            // TODO(ian): This just redirects to the eval page, which shows the most recent eval.  Redirect to this specific eval to avoid race.
-            navigate('/eval');
+            if (progressData.evalId) {
+              navigate(`/eval/${progressData.evalId}`);
+            }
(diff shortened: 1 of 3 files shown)


src/app/src/pages/eval-creator/components/RunTestSuiteButton.tsx
```diff
@@ -63,9 +63,9 @@ const RunTestSuiteButton: React.FC = () => {
           if (progressData.status === 'complete') {
             clearInterval(intervalId);
             setIsRunning(false);
```
adelmuursepp, line 65: Does not need to be part of this PR. However, after editing YAML config and then pressing 'Run eval' it navigated to not the recent eval which caused the confusion on the config being updated.

Now navigates to the updated recent eval.

src/app/src/stores/evalConfig.ts
```diff
@@ -98,6 +98,7 @@ export const useStore = create<State>()(
           scenarios,
           testCases,
           evaluateOptions,
```
adelmuursepp, line 100: Fix similar to PR #4059

mldangelo (approved):
Thank you @adelmuursepp!

mldangelo (approved):
Thanks!

PR #4155, opened 2025-05-25
docs: move "copy page" button to bottom

typpo:
and make it match edit page


mldangelo (commented):
Let's copy openai's pattern for this instead!

mldangelo:
Closing in favor of #4156

PR #4156, opened 2025-05-25
feat: minimal copy page button

mldangelo:
## Summary
- simplify the copy page button appearance to only show a copy icon
- sanitize markdown when building docs so runtime code can be simpler

## Testing
- `npm test` *(fails: The @smithy/node-http-handler package is required as a peer dependency)*


typpo:
It shouldn't have an entire row to itself that pushes down the entire page 
image

mldangelo:
Now 

Screenshot 2025-05-25 at 10 42 03 AM

typpo (approved):
seems fine

PR #4175, opened 2025-05-26
fix: when there's too many intents result won't render error

MrFlounder:


src/redteam/index.ts
```diff
@@ -719,7 +719,11 @@ export async function synthesize({
             metadata: {
               ...(t?.metadata || {}),
               pluginId: plugin.id,
-              pluginConfig: resolvePluginConfig(plugin.config),
```
mldangelo, line 722: nice catch, might be better if we move `...(t?.metadata || {}),` to below pluginConfig and set pluginConfig: null in the plugin itself instead of having special case logic here

MrFlounder: smart!

PR #4176, opened 2025-05-26
docs(site): restore scroll-to-top behavior on page navigation

mldangelo:


typpo:
I'm unable to verify the fix locally - still broken for me.  However, the change itself looks harmless, so I'm going to merge it and see what happens.

PR #4190, opened 2025-05-27
feat: canaries

typpo:


src/commands/canary.ts
```diff
@@ -0,0 +1,560 @@
+import chalk from 'chalk';
+import { generateShortProviderHash, getProviderId } from '../canary';
+import logger from '../logger';
+import { REQUEST_TIMEOUT_MS } from '../providers/shared';
```
mldangelo-oai, line 4: [P1] Migrate this call site to the current timeout API

`REQUEST_TIMEOUT_MS` is no longer exported by `providers/shared`; current `main` exposes `getRequestTimeoutMs()` instead. The published Build, Smoke, and Share logs all fail here with TS2724, and the focused suite fails 13 of 28 tests at runtime because `AbortSignal.timeout(undefined)` throws before any remote request. On the clean current-main merge, changing this import and line 72 to `getRequestTimeoutMs()` made TypeScript and all 28 focused canary tests pass. Please use the getter so the configured timeout is read at call time and add a regression that compiles against the current shared-provider API.

src/commands/canary.ts
```diff
@@ -0,0 +1,560 @@
...
+ */
+export async function checkCanaryForSingleProvider(
+  provider: ApiProvider,
+  checkMode: string = 'auto',
+): Promise<CanaryCheckResult> {
+  const providerHash = generateShortProviderHash(provider);
+
+  logger.debug(`Generated provider hash for checking: ${providerHash}`);
+
+  const response = await requestCanaryTask<ProbesResponse>({
```
mldangelo-oai, line 331: [P1] Persist and pass the canary tokens into the check task

The live `GenerateCanaryProbesRequestSchema` requires `canaryTokens` to be a non-empty string array, but this request sends only `task` and `hash`, so the advertised `canary check` flow is rejected before probes are generated. The send path cannot reconstruct the missing value later: token generation is LLM-based, the `_result` at line 510 is discarded when the command exits, and a user-supplied `--message` never reaches the generation service at all. A deterministic provider hash is therefore not a substitute for the actual sent tokens. Persist the exact tokens/custom message under an opaque per-send identifier (locally or in an account-scoped store), supply them here, and add an end-to-end test against the real request schema covering both generated and custom canaries.

src/commands/canary.ts
```diff
@@ -0,0 +1,560 @@
...
+        },
+      ];
+    }
+  }
+
+  if (pattern.type === 'semantic') {
+    const matchedWord = patternText
+      .split(/\s+/)
+      .filter((word) => word.length > 4)
+      .find((word) => content.toLowerCase().includes(word.toLowerCase()));
```
mldangelo-oai, line 193: [P1] Do not treat one common word as semantic leakage evidence

This takes the first pattern word longer than four characters that appears anywhere in the response, then marks the provider as trained on the user's data and makes the command exit nonzero. I bypassed the independent timeout-import blocker and ran a negative control: pattern `remember fictional training canary` against `I do not remember previous conversations.` matched only `remember`, yet the command printed `CANARY DETECTED! The provider has likely been trained on your data.` at 54% confidence. That is an explicit denial being converted into a serious false attribution. Require distinctive canary material and a real semantic comparison with a calibrated threshold (or multiple corroborating matches), and add denial/paraphrase/common-word negative controls before surfacing a detection.

mldangelo-oai (commented):
Reviewed exact head `785839a5ec98a8f223888ef4ef0057fe19bc6d09` in comment-only mode. This draft is not ready to merge: I found three P1 blockers, noted inline. First, the branch imports a timeout constant that no longer exists, which is the common cause behind the current build, smoke, share, red-team, and test failures. Second, the check flow discards the canary tokens generated during `send` and calls the live probe task with only a provider hash, although that task requires the original token list; custom messages are likewise unrecoverable. Third, the semantic detector treats any single common word overlap as evidence of training-data leakage.

The two remaining current env-option threads are not valid on this head. With Commander 14, `.option('--env-file, --env-path <path>')` accepts both aliases and deliberately derives the `envPath` property; I reproduced both forms and the repository uses the same convention in other commands. I replied on those threads rather than duplicating them. The older grammar, repeat, response-extraction, and request-order comments are outdated or addressed.

Verification: `git diff --check` passes and the head produces a clean no-commit merge with current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`). Both exact-head CI and the current-main merge fail TypeScript at `REQUEST_TIMEOUT_MS`; the focused canary suite then reports 13 failures out of 28 tests because `AbortSignal.timeout` receives `undefined`. A review-only synthetic migration to `getRequestTimeoutMs()` made TypeScript and all 28 focused tests pass, isolating that compatibility break. I also checked the live canary task schema, which requires a non-empty `canaryTokens` array for `generate-canary-probes`, and ran a direct semantic negative-control probe: pattern `remember fictional training canary` plus response `I do not remember previous conversations.` was reported as `CANARY DETECTED` at 54% confidence.

This branch is 510 commits behind current `main`. Please redesign the send/check state handoff, replace the single-word semantic heuristic, migrate to current timeout/config APIs, add end-to-end contract and negative-control tests, refresh without rewriting history, and rerun the full current CI matrix. No contributor branch changes or pushes were made.

src/commands/canary.ts
```diff
@@ -0,0 +1,490 @@
...
+    );
+
+  // Send subcommand
+  command
+    .command('send')
+    .description('Send a canary message to all providers in the config')
+    .option('-m, --message <message>', 'Custom canary message to send')
+    .option('-r, --repeat <number>', 'Number of canary variations to send', '1')
+    .option('-c, --config <path>', 'Path to config file')
+    .option('--env-file, --env-path <path>', 'Path to .env file')
```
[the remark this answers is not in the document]
mldangelo-oai: Verified against Commander 14: this declaration intentionally treats `--env-file` as the short alias and `--env-path` as the long flag. `attributeName()` is `envPath`, and parsing either alias returns `{ envPath: ... }`. This is also the repository-wide compatibility convention used by the other CLI commands. `options.envPath` is correct here; changing it to `envFile` would break the long-name mapping. No change is needed for this thread.

src/commands/canary.ts
```diff
@@ -0,0 +1,490 @@
...
+    .action(async (options) => {
+      try {
+        telemetry.record('command_used', { command: 'canary:send' });
+
+        if (options.envPath) {
+          setupEnv(options.envPath);
+        }
```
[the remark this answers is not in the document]
mldangelo-oai: I reproduced this directly with the Commander version on current `main`: both `--env-file .env` and `--env-path .env` produce `{ envPath: '.env' }`. The branch's own registration test also parses `--env-file` and confirms `setupEnv('.env')`; it passes after the unrelated timeout-import fix. The premise of this comment is therefore incorrect, and `options.envPath` should remain unchanged.

PR #4194, opened 2025-05-27
docs(site): restore scroll-to-top behavior on page navigation

will-holley:


will-holley:
Prefer https://github.com/promptfoo/promptfoo/pull/4195.

PR #4195, opened 2025-05-27
fix(docs): Fix issue with docs links not scrolling to the top

faizanminhas:
There was an issue moving between the homepage and the docs site- specifically, the scroll height would be maintained between links, and clicking an image to expand wouldn't work. After some digging, I found the offending commit https://github.com/promptfoo/promptfoo/pull/3733 and was able to identify two conflicting global css styles that were causing these issues.

## Demo

https://github.com/user-attachments/assets/90abd6f5-9c7e-4a07-8996-9f5bef64ce7f

### DIFF
--- a/site/src/pages/llm-vulnerability-scanner.module.css
+++ b/site/src/pages/llm-vulnerability-scanner.module.css
@@ -835,7 +835,6 @@ section {
 /* Carousel-specific overflow - only apply to carousel containers */
 .actionOrientedSection {
   padding: 4rem 0;
-  /* Remove overflow-x hidden that interferes with page scroll */
 }
 
 /* Only apply overflow hidden to carousel container specifically */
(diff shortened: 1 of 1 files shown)


site/src/pages/llm-vulnerability-scanner.module.css
```diff
@@ -832,12 +832,14 @@ section {
   padding: 25px 0 0;
 }
 
+/* Carousel-specific overflow - only apply to carousel containers */
 .actionOrientedSection {
-  overflow-x: hidden !important;
   padding: 4rem 0;
+  /* Remove overflow-x hidden that interferes with page scroll */
```
typpo:
```suggestion
```

site/src/pages/llm-vulnerability-scanner.module.css
```diff
@@ -832,12 +832,14 @@ section {
...
 }
 
-.container {
+/* Only apply overflow hidden to carousel container specifically */
```
typpo:
```suggestion
```

typpo (approved):

PR #4197, opened 2025-05-27
chore(eval): Adds first-class support for `beforeAll` and `beforeEach` extension hooks mutation of context

will-holley:
This PR allows the `beforeAll` and `beforeEach` extension hooks to modify and return their respective `context` arguments and have this updated context persisted into evaluation.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -7,7 +7,7 @@ import { globSync } from 'glob';
 import * as path from 'path';
 import type winston from 'winston';
 import { MODEL_GRADED_ASSERTION_TYPES, runAssertions, runCompareAssertion } from './assertions';
-import { fetchWithCache, getCache } from './cache';
+import { getCache } from './cache';
 import cliState from './cliState';
 import { FILE_METADATA_KEY } from './constants';
 import { updateSignalFile } from './database/signal';
@@ -269,7 +269,6 @@ export async function runEval({
 
           // All of these are removed in python and script providers, but every Javascript provider gets them
           logger: logger as unknown as winston.Logger,
-          fetchWithCache,
           getCache,
         },
         abortSignal ? { abortSignal } : undefined,
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -10,8 +10,8 @@ authors:
   - family-names: Zang
     given-names: Guangshuo
 title: promptfoo
-version: 0.115.0
-date-released: '2025-06-12'
+version: 0.115.1
+date-released: '2025-06-17'
 url: https://promptfoo.dev
 repository-code: https://github.com/promptfoo/promptfoo
 license: MIT
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "promptfoo",
-  "version": "0.115.0",
+  "version": "0.115.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "promptfoo",
-      "version": "0.115.0",
+      "version": "0.115.1",
       "license": "MIT",
       "workspaces": [
         "src/app",
@@ -863,16 +863,16 @@
       }
     },
     "node_modules/@aws-sdk/client-bedrock-runtime": {
-      "version": "3.828.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/client-bedrock-runtime/-/client-bedrock-runtime-3.828.0.tgz",
-      "integrity": "sha512-QhWv+UvSNPEwVM++TD5kRmTJytKqnxqlV7dkPHXibuOAlGKG7+Uu6WHt52BV6DlhVGU9An1Yl6p5uv/FKLoB6Q==",
+      "version": "3.830.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/client-bedrock-runtime/-/client-bedrock-runtime-3.830.0.tgz",
+      "integrity": "sha512-2f6t0K82CwNdoYYmnuYcyApAAh0H3oOhxF0lydRmPbty56owSC9R2MRzuy1lthLeEmVp0MFaCh17SlFfMKUR1Q==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/sha256-browser": "5.2.0",
         "@aws-crypto/sha256-js": "5.2.0",
         "@aws-sdk/core": "3.826.0",
-        "@aws-sdk/credential-provider-node": "3.828.0",
+        "@aws-sdk/credential-provider-node": "3.830.0",
         "@aws-sdk/eventstream-handler-node": "3.821.0",
         "@aws-sdk/middleware-eventstream": "3.821.0",
         "@aws-sdk/middleware-host-header": "3.821.0",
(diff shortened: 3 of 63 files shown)


examples/extension-api/README.md
```diff
@@ -19,6 +19,8 @@ promptfoo supports four types of extension hooks, each triggered at a specific p
 
 Each hook receives a `hookName` parameter and a `context` object containing relevant data for that specific hook type.
 
+The `beforeAll` and `beforeEach` hooks should return the `context` object, while the `afterAll` and `afterEach` hooks should not return anything.
+
```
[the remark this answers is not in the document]
will-holley: "should return" is accurate; it is encouraged but not required.

site/docs/configuration/reference.md
```diff
@@ -97,16 +97,16 @@ It automatically loads `promptfooconfig.*`, but you can use a custom config file
...
 
 ### Available Hooks
 
-| Hook Name  | Description                                   | Arguments                                         |
-| ---------- | --------------------------------------------- | ------------------------------------------------- |
-| beforeAll  | Runs before the entire test suite begins      | `{ suite: TestSuite }`                            |
-| afterAll   | Runs after the entire test suite has finished | `{ results: EvaluateResult[], suite: TestSuite }` |
-| beforeEach | Runs before each individual test              | `{ test: TestCase }`                              |
-| afterEach  | Runs after each individual test               | `{ test: TestCase, result: EvaluateResult }`      |
+| Name       | Description                                   | Context                                           | Mutable Context |
```
mldangelo: recommend removing ` Mutable Context` column. I think it is confusing without more context.

site/docs/configuration/reference.md
```diff
@@ -174,6 +224,29 @@ module.exports = extensionHook;
 
 These hooks provide powerful extensibility to your promptfoo evaluations, allowing you to implement custom logic for setup, teardown, logging, or integration with other systems. The extension function receives the `hookName` and a `context` object, which contains relevant data for each hook type. You can use this information to perform actions specific to each stage of the evaluation process.
 
+### Mutable Context
+
```
mldangelo:
```suggestion
```

src/evaluator.ts
```diff
@@ -527,7 +527,7 @@ class Evaluator {
   }
 
   async evaluate(): Promise<Eval> {
-    const { testSuite, options } = this;
+    const { options } = this;
```
mldangelo, line 541: recommend changing this to let and not changing all of the references below to this.testSuite

src/util/transform.ts
```diff
@@ -147,10 +147,10 @@ async function getTransformFunction(
 export async function transform(
   codeOrFilepath: string,
   transformInput: string | object | undefined,
-  context: TransformContext,
+  context: any,
```
mldangelo: Can you type again TransformContext?

will-holley: It was incorrect.

src/evaluator.ts
will-holley: What's the difference? I wouldn't have destructured `this` in the first place.

site/docs/configuration/reference.md
will-holley: Does it make sense once you get to the Mutable Context section below? Thoughts on how to better indicate this?

src/evaluator.ts
will-holley: I'd also think that the user's changes to `this.testSuite` should persist onto the instance scope.

src/util/transform.ts
mldangelo: Let's correct the type instead of setting it to any!

site/docs/configuration/reference.md
mldangelo: Don't think it's necessary at this point. The progressive disclosure below is enough

src/evaluator.ts
mldangelo: It avoids a 40+ line refactor when a two-line tweak keeps the same behaviour, and it’s consistent with the rest of evaluator.ts (we alias options, metrics, etc.) and makes the inner loops shorter to read.

src/util/transform.ts
```diff
@@ -150,7 +147,7 @@ export async function transform(
   context: TransformContext,
   validateReturn: boolean = true,
   inputType: TransformInputType = TransformInputType.OUTPUT,
-): Promise<Vars> {
+): Promise<any> {
```
sklein12, line 149: does this have to be `any`?

src/evaluatorHelpers.ts
```diff
@@ -324,29 +336,164 @@ export async function renderPrompt(
...
+    }
+    case 'beforeEach': {
+      const parsed = BeforeEachExtensionHookContextSchema.safeParse(context);
+      invariant(
+        parsed.success,
+        `Invalid context passed to beforeEach hook: ${parsed.error?.message}`,
+      );
+      break;
+    }
+  }
```
sklein12, line 443: why is this required?

src/util/transform.ts
will-holley: Technically yes because `postprocessFn` can return anything.  In practice, we're only using this to call extension hooks and they _should_ return a context or void, but that's not guaranteed.

src/evaluatorHelpers.ts
will-holley: Before this PR the context types were not up to date. The idea is to provide a strong type guard but not sure if this could instead be done only at the TS level.

src/evaluatorHelpers.ts
```diff
@@ -324,29 +336,164 @@ export async function renderPrompt(
...
+            );
+          }
+          break;
+        }
+      }
+    } else {
+      // Backwards compatibility: no-op if the extension does not return a value.
+      // [This block is included to explicitly state this is a backwards compatibility feature.]
+      continue;
+    }
```
mldangelo: can we trim this section? I think it is not necessary.

src/evaluatorHelpers.ts
```diff
@@ -324,29 +336,164 @@ export async function renderPrompt(
...
+  test: TestCaseSchema,
+});
+
+type BeforeAllExtensionHookContext = z.infer<typeof BeforeAllExtensionHookContextSchema>;
+type BeforeEachExtensionHookContext = z.infer<typeof BeforeEachExtensionHookContextSchema>;
+
+/**
+ * Defines the set of fields on BeforeAllExtensionHookContextSchema that may be mutated by the extension hook.
+ */
+const MutableBeforeAllExtensionHookContextSchema = z.object({
```
mldangelo, line 356: Is any of this duplicative with other types?

src/evaluatorHelpers.ts
```diff
@@ -324,29 +336,164 @@ export async function renderPrompt(
...
+const BeforeAllExtensionHookContextSchema = z.object({
+  suite: TestSuiteSchema,
+});
+
+const BeforeEachExtensionHookContextSchema = z.object({
+  test: TestCaseSchema,
+});
+
+type BeforeAllExtensionHookContext = z.infer<typeof BeforeAllExtensionHookContextSchema>;
+type BeforeEachExtensionHookContext = z.infer<typeof BeforeEachExtensionHookContextSchema>;
```
mldangelo: I think you should try to cut as much of this out as possible. This can just use the TestCase type which is already defined in src/types/index.ts. It is harder for me to grok this way

mldangelo (commented):
Please see if you can cut from src/evaluatorHelpers.ts I am getting lost in the type system and new validations

src/evaluatorHelpers.ts
will-holley: No

src/evaluatorHelpers.ts
will-holley: An inline comment is preferable if these require explanation.

will-holley: Added a comment in https://github.com/promptfoo/promptfoo/pull/4197/commits/031496f88e9accaf87e9aa94c78b17632330ab2e. Runtime schema validation is performed on `context` to guard against type errors upstream of the hook i.e. we're providing strong guarantees to the user that its schema is as expected.

PR #4217, opened 2025-05-29
chore: extract intent all the time

MrFlounder:


src/redteam/index.ts
```diff
@@ -821,5 +846,9 @@ export async function synthesize({
 
   logger.info(generateReport(pluginResults, strategyResults));
 
-  return { purpose, entities, testCases: finalTestCases, injectVar };
+  const testCasesWithIntent = await Promise.all(
+    finalTestCases.map((testCase) => addIntentToTestCase(testCase, purpose)),
```
sklein12: 1. I don't think this is the right place to do it. The basic test cases are duplicated for all of the strategies.

2. Promise.all could trigger 1000's of requests at once.

I think we should do this when we generate the basic test cases in line so it happens serially based on the configured concurrency.

src/redteam/util.ts
```diff
@@ -1,8 +1,13 @@
...
+import logger from '../logger';
+import { REQUEST_TIMEOUT_MS } from '../providers/shared';
+import { getRemoteGenerationUrl } from './remoteGeneration';
+
 /**
  * Normalizes different types of apostrophes to a standard single quote
  */
 export function normalizeApostrophes(str: string): string {
-  return str.replace(/['′’']/g, "'");
+  return str.replace(/['′'']/g, "'");
```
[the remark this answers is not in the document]
sklein12: agreed, we should revert this

src/redteam/index.ts
```diff
@@ -713,17 +738,29 @@ export async function synthesize({
...
+        }));
+
+        // Extract goal for this plugin's tests
+        logger.debug(
+          `Extracting goal for ${testCasesWithMetadata.length} tests from ${plugin.id}...`,
+        );
+        const testCasesWithGoal = await Promise.all(
+          testCasesWithMetadata.map(async (testCase) => {
+            return await addGoalToTestCase(testCase, purpose);
+          }),
```
sklein12: We can't use promise.all, we always only do one call at a time unless we're adhering to a concurrency number. Since this is already inside a concurrency thread, we can't do promise.all, you need to do one at a time.

src/redteam/index.ts
```diff
@@ -411,6 +411,31 @@ function isStrategyCollection(id: string): id is keyof typeof STRATEGY_COLLECTIO
...
+ * @param testCase - The test case to add goal to.
+ * @param purpose - The purpose of the system.
+ * @returns The test case with goal added to metadata.
+ */
+async function addGoalToTestCase(
+  testCase: TestCaseWithPlugin,
+  purpose: string,
+): Promise<TestCaseWithPlugin> {
+  // Get the prompt from the test case vars
+  const promptVar = Object.values(testCase.vars || {})[0];
```
sklein12: we have the injectVar in the callsite, pass it down here

src/redteam/index.ts
```diff
@@ -411,6 +411,31 @@ function isStrategyCollection(id: string): id is keyof typeof STRATEGY_COLLECTIO
...
+  const prompt = Array.isArray(promptVar) ? promptVar[0] : String(promptVar);
+
+  const extractedGoal = await extractGoalFromPrompt(prompt, purpose);
+
+  return {
+    ...testCase,
+    metadata: {
+      ...testCase.metadata,
+      goal: extractedGoal,
+    },
```
sklein12: This isn't a huge deal but I don't love this pattern. It seems a little unnecessary when you could return the goal and set a single property on the test case.

src/redteam/util.ts
```diff
@@ -197,3 +202,39 @@ export function removePrefix(str: string, prefix: string) {
...
+      },
+      REQUEST_TIMEOUT_MS,
+    );
+
+    if (status !== 200) {
+      logger.warn(`Failed to extract goal from prompt: HTTP ${status}`);
+      return prompt;
+    }
+
+    return data?.intent || prompt;
```
sklein12: Why return the prompt if there's no intent? In my mind we should fail and not set a value.

PR #4232, opened 2025-05-29
chore: handle null goal

MrFlounder:


src/redteam/plugins/base.ts
```diff
@@ -303,6 +303,7 @@ export abstract class RedteamGraderBase {
         : undefined,
       value: renderedValue,
       testVars: test.vars ?? {},
+      goal: test.metadata?.goal || prompt,
```
sklein12: please do not do this, like i said, it should be up to the consumer to handle a null goal, we shouldn't put the prompt in instead. This is semantically in correct and we're putting the wrong data into a field.

src/redteam/util.ts
```diff
@@ -230,17 +232,17 @@ export async function extractGoalFromPrompt(prompt: string, purpose: string): Pr
 
     if (status !== 200) {
       logger.error(`Failed to extract goal from prompt: HTTP ${status}`);
```
sklein12: can you please also log out the body and status text

src/redteam/util.ts
```diff
@@ -230,17 +232,17 @@ export async function extractGoalFromPrompt(prompt: string, purpose: string): Pr
...
-      throw new Error(`Failed to extract goal from prompt: HTTP ${status}`);
+      return null;
     }
 
     if (!data?.intent) {
       logger.error('No intent returned from extraction API');
```
sklein12: can you log out the data and or the body here so we have some idea what was actually returned?

sklein12: this should be a warning

src/redteam/util.ts
sklein12: this should be a warning

src/redteam/util.ts
```diff
@@ -228,19 +230,23 @@ export async function extractGoalFromPrompt(prompt: string, purpose: string): Pr
...
     );
 
+    logger.debug(
+      `Goal extraction response - Status: ${status} ${statusText || ''}, Data: ${JSON.stringify(data)}`,
+    );
+
     if (status !== 200) {
-      logger.error(`Failed to extract goal from prompt: HTTP ${status}`);
-      throw new Error(`Failed to extract goal from prompt: HTTP ${status}`);
+      logger.warn(`Failed to extract goal from prompt: HTTP ${status} ${statusText || ''}`);
```
sklein12: is the body available here if there's data in it?

src/redteam/util.ts
```diff
@@ -228,19 +230,23 @@ export async function extractGoalFromPrompt(prompt: string, purpose: string): Pr
...
+      return null;
     }
 
     if (!data?.intent) {
-      logger.error('No intent returned from extraction API');
-      throw new Error('No intent returned from extraction API');
+      logger.warn('No intent returned from extraction API');
```
sklein12: can you please log all information we have about the request here including the body?

sklein12 (dismissed):
please add some more info to the logs :) otherwise looks good

PR #4255, opened 2025-05-30
feat: allow overriding base URL for Google provider

yurchik11:


src/providers/google/types.ts
```diff
@@ -91,6 +91,7 @@ export interface Tool {
 export interface CompletionOptions {
   apiKey?: string;
   apiHost?: string;
+  apiBaseUrl?: string;
```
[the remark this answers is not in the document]
yurchik11: I do not agree that it's a good idea, it has been done similarly to other providers like OpenAI

yurchik11:
Hi @mldangelo , could you take a look at my changes? 
thanks in advance,

adelmuursepp:
> Hi @mldangelo , could you take a look at my changes? thanks in advance,

Thanks for the change! Can you link a related issue or relevant link for the purpose of PR?

src/providers/google/ai.studio.ts
```diff
@@ -66,6 +66,19 @@ class AIStudioGenericProvider implements ApiProvider {
...
   }
 
+  getApiUrlDefault(): string {
+    return `https://${this.getApiHost()}`;
+  }
+
+  getApiUrl(): string {
+    return (
+      this.config.apiBaseUrl ||
+      this.env?.GOOGLE_API_BASE_URL ||
```
adelmuursepp, line 87: Could you have a look at the OpenAI provider same method. It would be good to follow the same style.  

In getApiUrl first check for apiHost. 
Then check for apiUrl.
Default to getApiUrlDefault which would return the 
 'https://generativelanguage.googleapis.com/v1beta3/models/' <-- let me check for better default URLs as well


Override Hierarchy: 
    - Most specific: apiHost 
    - Less specific: apiBaseUrl 
    - Default: The standard URL

src/providers/google/ai.studio.ts
```diff
@@ -379,7 +392,7 @@ export class GoogleEmbeddingProvider extends AIStudioGenericProvider {
 
     // Use embedContent endpoint
     const endpoint = 'embedContent';
-    const url = `https://${this.getApiHost()}/v1/models/${this.modelName}:${endpoint}?key=${this.getApiKey()}`;
+    const url = `${this.getApiUrl()}/v1/models/${this.modelName}:${endpoint}?key=${this.getApiKey()}`;
```
adelmuursepp, line 406: Great! This now follows similar pattern to OpenAI provider API calls

src/providers/google/types.ts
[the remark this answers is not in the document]
adelmuursepp: Agreed @yurchik11

yurchik11:
> > Hi @mldangelo , could you take a look at my changes? thanks in advance,
> 
> Thanks for the change! Can you link a related issue or relevant link for the purpose of PR?

Hi @adelmuursepp , thank you for the review!
I'm currently using a separate proxy app to handle LLM vendor requests. As part of this setup, I need to override the base URL for different vendors such as OpenAI, Anthropic, and Google.

I've successfully implemented this override for OpenAI and Anthropic. However, I'm encountering issues with overriding the baseUrl for Google in the same way. Specifically, I cannot use overriding the host URL because I need HTTP protocol.

Let me know if you need additional context or have suggestions.

src/providers/google/ai.studio.ts
yurchik11: This is done, I have removed getApiHost() method, also @mldangelo added a check for apiHost in getApiUrl method. thanks :)

src/providers/google/ai.studio.ts
yurchik11: yes, could you review the final version?

mldangelo (approved):
Thank you so much @yurchik11!

PR #4257, opened 2025-05-30
chore: add metadata to generated redteam.yaml

typpo:


mldangelo (commented):
Nice! Logged into would be great too! And making this DRY, and adding tests, and linting it and formatting it, and getting the existing tests to not fail, and getting the project to build, and testing it.

PR #4280, opened 2025-06-02
feat: Google live audio output

adelmuursepp:
## Overview

Implemented audio output support for Google's Live API, enabling audio generation, transcription, and advanced features like affective dialog and voice customization. Yet to cover coercing the model to speak in specific language, turn coverage, proactive audio and features involving audio as input.

## Key Features Added

1. Audio Output Support in Google Live Provider

2. New Audio Examples (`examples/google-live-audio/`)
Created three examples demonstrating audio capabilities:

     a) Simple Audio Generation (`promptfooconfig.yaml`)
    b) Affective Dialog (`promptfooconfig.audio-affective.yaml`)
    c) Voice Customization (`promptfooconfig.audio-voices.yaml`)

3. UI Enhancements for Audio Display
- Modified `EvalOutputCell.tsx` to properly display audio outputs. Added download audio button (decided not to generate files in the folder)

## Testing
- Added unit tests in `test/providers/google/live.test.ts`
- Tested end to end manually and cross checked with AI studio
- Tests cover audio output, transcription, and error handling

image

Amazon Bedrock Nova speech to speech:
image

OpenAI audio:
image


examples/google-live-audio/promptfooconfig.audio-voices.yaml
```diff
@@ -0,0 +1,43 @@
...
+    label: 'Leda Voice'
+    config:
+      generationConfig:
+        response_modalities: ['audio']
+        outputAudioTranscription: {}
+        speechConfig:
+          voiceConfig:
+            prebuiltVoiceConfig:
+              voiceName: 'Leda'
+      timeoutMs: 30000
```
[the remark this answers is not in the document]
adelmuursepp: It mirrors what the [API expects ](https://ai.google.dev/gemini-api/docs/speech-generation#multi-speaker) so I wouldn't stray in formatting

examples/google-live-audio/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+You can obtain a Google AI Studio API key from the [Google AI Studio website](https://ai.google.dev/).
+
+## Running the Examples
+
+After running any evaluation, you can view the results by running:
+
+```bash
+promptfoo view
+```
```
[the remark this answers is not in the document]
adelmuursepp: @faizanminhas @mldangelo It doesn't work locally so I didn't add it. Will it work after merging? Should I add the command in readme?

examples/google-live-audio/README.md
```diff
@@ -0,0 +1,73 @@
+# google-live-audio (Google Live API Audio Features with Gemini)
```
mldangelo: The amount of detail you have here is great, you should move some of this into the site documentation instead of the example documentation. Could even write an guide for it if it doesn't fit well into google.md

examples/google-live-audio/README.md
```diff
@@ -0,0 +1,73 @@
...
+- Puck
+- Charon
+- Kore
+- Fenrir
+- Aoede
+- Leda
+- Orus
+- Zephyr
+
+For more information about the Google Live API and audio capabilities, see the [Google AI documentation](/docs/providers/google).
```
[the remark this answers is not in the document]
mldangelo: this is confusing, greptile is correct here

examples/google-live-audio/promptfooconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+tests:
+  # Option 1: Use javascript assertion to check transcript via context
+  - assert:
+      - type: javascript
+        value: |
+          // Access the audio transcript from the provider response
+          if (context.providerResponse && context.providerResponse.audio && context.providerResponse.audio.transcript) {
+            return context.providerResponse.audio.transcript.toLowerCase().includes('well');
+          }
```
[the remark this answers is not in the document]
adelmuursepp: No. I would like to see that in transcription certain words are present.

examples/google-live-audio/promptfooconfig.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - assert:
+      - type: javascript
+        value: |
+          // Verify audio response was generated with transcript
+          if (context.providerResponse && context.providerResponse.audio && context.providerResponse.audio.transcript) {
+            return context.providerResponse.audio && 
+                  context.providerResponse.audio.data && 
+                  context.providerResponse.audio.transcript && 
+                  context.providerResponse.audio.transcript.length > 0;
+          }
```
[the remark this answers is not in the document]
adelmuursepp: Wrote the sample test to be very verbose so that users can follow example with many options and figure out which tests are most relevant to them.

examples/google-live-audio/promptfooconfig.audio-affective.yaml
```diff
@@ -0,0 +1,52 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Google Live Audio - Affective Dialog (Emotion Adaptation)
+
+# Affective dialog lets Gemini adapt its response style to match the input expression and tone
+# Learn more about affective dialog at https://ai.google.dev/gemini-api/docs/live#affective-dialog
+
+prompts:
+  - label: 'Positive input'
+    raw: It's a new dawn, it's a new day, and I'm feeling good.
```
mldangelo: slightly prefer if you add a var in here such as  {{question}} or {{statement}} and then define it in the test cases

examples/google-live-audio/promptfooconfig.audio-voices.yaml
```diff
@@ -0,0 +1,43 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Google Live Audio - Different Voice Options
+
+# This example demonstrates the different voices available in Google Live API
+# Available voices: Puck, Charon, Kore, Fenrir, Aoede, Leda, Orus, Zephyr
+
+prompts:
+  - 'Why was Gemini model by Google named Gemini?'
```
mldangelo: same here

examples/google-live-audio/promptfooconfig.audio-voices.yaml
```diff
@@ -0,0 +1,43 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Google Live Audio - Different Voice Options
+
+# This example demonstrates the different voices available in Google Live API
+# Available voices: Puck, Charon, Kore, Fenrir, Aoede, Leda, Orus, Zephyr
```
mldangelo: may not need to repeat this UNLESS you want to make this a test case var

examples/google-live-audio/promptfooconfig.yaml
```diff
@@ -0,0 +1,37 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Google Live Simple Audio Test
+
+prompts:
```
mldangelo, line 4: same

examples/google-live-audio/promptfooconfig.yaml
mldangelo: You can simplify this a lot with ?.

src/providers/google/live.ts
```diff
@@ -202,89 +526,90 @@ export class GoogleLiveProvider implements ApiProvider {
...
-            ...(this.config.tools ? { tools: loadFile(this.config.tools, context?.vars) } : {}),
-            ...(systemInstruction ? { systemInstruction } : {}),
-          },
-        };
-        logger.debug(`WebSocket sent: ${JSON.stringify(setupMessage)}`);
-        ws.send(JSON.stringify(setupMessage));
+        if (audioCompletionTimer) {
+          clearTimeout(audioCompletionTimer);
+        }
+        // If the promise hasn't been resolved yet and the closure was unexpected, reject.
```
mldangelo: Do you still want to do this?

adelmuursepp:
> You put so much work into this. Thank you. It looks great. A few nits about documentation but overall it is really good.
> 
> Question for the future: How do you want to grade audio content with llm-as-a-judge?

Let me address the suggestions, thanks!

adelmuursepp:
> > You put so much work into this. Thank you. It looks great. A few nits about documentation but overall it is really good.
> > Question for the future: How do you want to grade audio content with llm-as-a-judge?
> 
> Let me address the suggestions, thanks!

There is another interesting aspect to think about. I see in your OpenAI audio example that the transcript is moved in the text field which allows to run llm rubric on it. Transcript is for me in context.providerResponse.audio.transcript. 

a) Transcript could be copied over to text. Which makes sense
b) For now for Google Live audio audio thinking models the text is filled with the model's thinking process, which is very interesting to read. I agree it doesn't belong there. But disregard thinking process or copy it somewhere else?

examples/google-live-audio/README.md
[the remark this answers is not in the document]
mldangelo: It will work after this is merged to main! You should add it. Thank you for testing it.

mldangelo:
Good points @adelmuursepp!. We should use the model graded metrics just like how the openai models work. Re thinking processes - how about putting them in metadata instead (feel free to disagree). Thank you so much!

src/providers/google/live.ts
adelmuursepp: Actually I redid the timeout to be similar to OpenAI audio. It will work most cases and if there is an issue or some WebSocket event lost then the user can just rerun the eval.

adelmuursepp:
> Good points @adelmuursepp!. We should use the model graded metrics just like how the openai models work. Re thinking processes - how about putting them in metadata instead (feel free to disagree). Thank you so much!

I like how Bedrock shows extra fields in the metadata, it is cleaner. But the other providers  generally show fields like function and tool calls in the output directly. I think it is better to be consistent in showing extra fields. I am not going to change how other Live provider examples work, so I would rather place it in the output for now. And refactor all into text and metadata in Live provider later. 
```
{
    output: {
      text: "transcript text",
      toolCall: { functionCalls: [...] }, // Always showing, even when empty
      statefulApiState: ..., // Different example, only when present
      thinking: "thinking process text"  // Only when present
    },
    audio: { ... }
  }
```

image


For quick reference, metadata:

image

@mldangelo you will know how to be consistent and organize thinking better in the other providers too going forward probably?

mldangelo (approved):
Thank you @adelmuursepp!

examples/google-live-audio/prompt.yaml
```diff
@@ -0,0 +1,2 @@
+- role: user
```
mldangelo, line 1: nit, for a simple prompt, consider defining it inline in `examples/google-live-audio/promptfooconfig.yaml` instead of a different file

PR #4284, opened 2025-06-02
Feature/modify matcher templates to support both string and non string objects

yzjing:
Instead of force stringifying JSON objects in template rendering, we would like to support both string and non-string objects. This re-does the change in #3746 while keeping it non-breaking for the use case in #3896 .


mldangelo:
Closing in favor of https://github.com/promptfoo/promptfoo/pull/4297 Thank you @yzjing!

PR #4285, opened 2025-06-02
chore(redteam): update goal extraction

MrFlounder:


mldangelo (approved):
Please make sure this doesn't break when we do a release if the backend is not ready

PR #4295, opened 2025-06-03
chore: do not include empty values in discovery result

MrFlounder:


src/redteam/commands/discover.ts
```diff
@@ -96,6 +96,34 @@ const COMMAND = 'discover';
...
 
+/**
+ * Normalizes a TargetPurposeDiscoveryResult by converting null-like values to actual null
+ * and cleaning up empty or meaningless content.
+ */
+function normalizeTargetPurposeDiscoveryResult(
+  result: TargetPurposeDiscoveryResult,
+): TargetPurposeDiscoveryResult {
+  // Helper function to check if a string value should be considered null
+  const isNullLike = (value: string | null | undefined): boolean => {
```
mldangelo, line 107: doesn't need to be an inner function here and below

mldangelo (commented):
Why not handle this on the backend instead?

PR #4296, opened 2025-06-03
docs: gpt 4o to 4.1

typpo:


mldangelo (dismissed):
How did you do this while presenting your screen to so many people?

PR #4297, opened 2025-06-03
feat: add PROMPTFOO_DISABLE_OBJECT_STRINGIFY environment variable for object template handling

mldangelo:
This PR addresses the conflict between two competing needs in template rendering:

1. **Preventing `[object Object]` issues** (addressed in #3896) - Objects need to be stringified to avoid display problems
2. **Enabling object property access** (originally in #3746) - Users need to access object properties like `{{output.text}}`

While PR #4284 solves this with verbose conditional syntax in templates (`{{ output if output === output|string else output|dump }}`), this PR proposes a solution using environment variables that does not require modifying existing prompts by introducing the `PROMPTFOO_DISABLE_OBJECT_STRINGIFY` environment variable that allows users to control object handling behavior:

**Default behavior (safe, non-breaking):**
- Objects are automatically stringified to prevent `[object Object]` issues
- Maintains backward compatibility with all existing templates

**Optional object access:**
```bash
export PROMPTFOO_DISABLE_OBJECT_STRINGIFY=true
promptfoo eval
```
- Enables direct object property access: `{{product.name}}`, `{{output.text}}`
- Allows advanced template navigation of complex data structures

Related: #3746 #3896 #4284


mldangelo:
@yzjing I am going to merge this. It's backwards compatible with current production behavior and should support your use case. If it's not working please let me know and we can merge your version instead. 

Thanks so much!

yzjing:
> @yzjing I am going to merge this. It's backwards compatible with current production behavior and should support your use case. If it's not working please let me know and we can merge your version instead.
> 
> Thanks so much!

Hi @mldangelo , sorry for getting back to you late! Confirming that it works with our use case. Thank you so much for working on this!

PR #4317, opened 2025-06-04
fix(redteam): Exclude memory poisoning plugin from strategies

will-holley:


src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -62,8 +62,10 @@ export class MemoryPoisoningProvider implements ApiProvider {
         throw new Error(`Failed to generate scenario: ${scenarioRes.statusText}`);
       }
 
+      // Scope the scenario to the test case to ensure its passed to the grader:
       const scenario = await scenarioRes.json();
-      context.vars.scenario = scenario;
+      context!.test!.metadata ??= {};
```
MrFlounder, line 67: I think this is a better place to validate the scenario schema then in grader, so we can fail early.

PR #4394, opened 2025-06-10
chore: better error message for missing plugins

sklein12:
```
Invalid configuration file /Users/steve/tmp/datadog/promptfooconfig.yaml:
Validation error: Custom plugins must start with file:// (or use one of the built-in plugins). Received: "not-real-plugin" at "redteam.plugins[27].id"
```


mldangelo:
See https://github.com/promptfoo/promptfoo/pull/4330

PR #4415, opened 2025-06-11
docs: update gh-cli-workflow rule with copy edits and improvements

mldangelo:


.cursor/rules/gh-cli-workflow.mdc
```diff
@@ -0,0 +1,69 @@
+---
+description: 
+globs: 
+alwaysApply: false
+---
+---
```
mldangelo:
```suggestion
```

PR #4424, opened 2025-06-12
fix: Fail gracefully on RangeErrors and try to output a truncated version of the results

Sly1029:
During large testing runs I am frequently getting:

```
RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at safeJsonStringify (/usr/local/lib/node_modules/promptfoo/src/util/json.ts:41:10)
    at processEvalStep (/usr/local/lib/node_modules/promptfoo/src/evaluator.ts:894:74)
    at /usr/local/lib/node_modules/promptfoo/src/evaluator.ts:1247:7

```

This aims to address that by introducing `summarizeEvaluateResultForLogging` and `safeJsonStringifyTruncated` which will attempt to output a truncated version instead. It will also fail gracefully and continue

### DIFF
--- a/.cursor/rules/gh-cli-workflow.mdc
+++ b/.cursor/rules/gh-cli-workflow.mdc
@@ -1,0 +1,66 @@
+---
+description:
+globs:
+alwaysApply: false
+---
+
+# Git Workflow Using `gh`
+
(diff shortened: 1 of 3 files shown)


mldangelo (approved):
Thank you!

PR #4435, opened 2025-06-12
feat(providers): add Mistral Magistral reasoning models

mldangelo:


site/docs/guides/mistral-magistral-aime2024.md
```diff
@@ -0,0 +1,257 @@
+---
```
mldangelo, line 1: @ladyofcode would love a quick review on this

PR #4465, opened 2025-06-13
fix(eval): Fix RangeError on progress bar var replace with large strings

Sly1029:
Occasionally you can get a crash when the progress bar attempts to load too big of a vars string into memory.

This manifest's like so: 
`RangeError: Invalid string length
    at Array.join (<anonymous>)
    at Object.options.progressCallback (/usr/local/lib/node_modules/promptfoo/src/evaluator.ts:1187:14)
    at processEvalStep (/usr/local/lib/node_modules/promptfoo/src/evaluator.ts:1030:19)
    at /usr/local/lib/node_modules/promptfoo/src/evaluator.ts:1291:9`
    
   ## Solution
   
   Attempt to fix this by truncating the vars in the progress bar. Also set a suitable fallback if the RangeError still occurs. 
   
   ## Testing
   
   Wrote a unit test for this behavior with ~60MB strings. Also checked that the styling is not altered by running the vars-referencing-vars config.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -1186,26 +1186,28 @@ class Evaluator {
             try {
               // Constants for progress bar display
               // Limit individual var values
-              const MAX_VAR_VALUE_LENGTH = 50; 
-              // Limit total vars string length 
+              const MAX_VAR_VALUE_LENGTH = 50;
+              // Limit total vars string length
               // Maybe this should vary based on terminal size??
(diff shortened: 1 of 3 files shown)


mldangelo:
Thanks @Sly1029! This is a great find and fix! resolveVariables is fairly fragile. I am going to test this. I may look for an alternative fix where we store a truncated copy of the vars and use those for reporting progress.

PR #4485, opened 2025-06-13
fix: Windows Python path validation race condition

benbuzz790:
## Problem
Windows users experienced intermittent Python distribution detection failures when running multiple tests concurrently.

## Root Cause
The issue occurred when multiple concurrent calls to validatePythonPath() would all detect a cache miss simultaneously and start separate validations, leading to inconsistent results and cache corruption.

## Solution
- Added `validationPromise` synchronization variable to prevent concurrent Python path validations
- Concurrent calls now wait for single shared validation
- Eliminates cache corruption and inconsistent results

## Testing
- All race condition tests now pass (aggressive-race-test.js) (0% failure rate)
- Tested on Windows with aggressive concurrent validation scenarios
- Before fix: aggressive-race-test.js shows 12.28% failure rate with race conditions (varied by run)
- After fix: 0%


benbuzz790:
Ok, I think this is all set. My bot got off it's leash so I had to revert a commit. Looks good now. Let me know @mldangelo. Does CodeRabbit ever stop making suggestions? I suggest you either drop my race condition demonstration script or write a new one once this is confirmed to correct the issue.

benbuzz790:
Thank you @mldangelo - looks like I could have done a better job with some of the formatting? I'm open to feedback and criticism to help me work faster and more accurately next time.

mldangelo:
@benbuzz790 you did a great job, especially with respect to your test procedure. I just ran the formatter (npm run f) and the tests with --detectOpenHandles, which helped me spot that we needed to clear the timeout.
You can see what I did in https://github.com/promptfoo/promptfoo/pull/4485/commits/1a398ba6438d7dea52701b654bfc4654d2555ad2

Thanks for your work on this! It means a lot.

PR #4532, opened 2025-06-18
chore: Add unblocking detection to GOAT strategy

MrFlounder:
## Summary
- check GOAT target responses for blocking questions and unblock if needed
- allow skipping feature check when testing

## Testing
- `npm test --silent`
- `npx jest test/redteam/providers/goat.test.ts -t "should handle stateful behavior" --runInBand --silent`

------
https://chatgpt.com/codex/tasks/task_e_6852023a3c108330a957a791238f9d38

Test result:
https://www.promptfoo.app/eval/eval-Ldk-2025-06-19T23:26:58

Test instructions:
1. Run the target agent under folder redteam-payment-agent (npm run start)
2. Setup a quick scan using that target with goat and any plugin
3. You should see the unblocking behavior in eval result -> attemps (conversation history)


src/redteam/providers/goat.ts
```diff
@@ -308,4 +388,68 @@ export default class GoatProvider implements ApiProvider {
...
+
+      return { success: false, tokenUsage: response.tokenUsage };
+    } catch (error) {
+      logger.debug(`[GOAT] Error in unblocking: ${error}`);
+      return { success: false };
+    }
+  }
+
+  private async checkServerSupportsUnblocking(): Promise<boolean> {
+    if (process.env.NODE_ENV === 'test') {
```
will-holley: Mock this in Jest?

src/redteam/providers/goat.ts
```diff
@@ -255,6 +263,78 @@ export default class GoatProvider implements ApiProvider {
...
+        );
+        if (unblockingResult.tokenUsage) {
+          totalTokenUsage.total += unblockingResult.tokenUsage.total || 0;
+          totalTokenUsage.prompt += unblockingResult.tokenUsage.prompt || 0;
+          totalTokenUsage.completion += unblockingResult.tokenUsage.completion || 0;
+          totalTokenUsage.numRequests += unblockingResult.tokenUsage.numRequests ?? 1;
+          totalTokenUsage.cached += unblockingResult.tokenUsage.cached || 0;
+        }
+
+        if (unblockingResult.success && unblockingResult.unblockingPrompt) {
```
will-holley: How should the failure case be handled?

src/redteam/providers/goat.ts
```diff
@@ -255,6 +263,78 @@ export default class GoatProvider implements ApiProvider {
...
+
+          if (unblockingOutput) {
+            messages.push({ role: 'assistant', content: unblockingOutput });
+            previousTargetOutput = unblockingOutput;
+            finalOutput = unblockingOutput;
+            finalResponse = unblockingResponse;
+          }
+
+          if (unblockingResponse.sessionId) {
+            context = context ?? { vars: {}, prompt: { raw: '', label: 'target' } };
```
will-holley: Should this fallback be moved up?

src/redteam/providers/goat.ts
MrFlounder: I think just fall back to the regular goat strategy

Updated some logger to error level

src/redteam/providers/goat.ts
MrFlounder: this response actually never contains session id

removed, thank you!

src/redteam/providers/goat.ts
```diff
@@ -273,6 +281,73 @@ export default class GoatProvider implements ApiProvider {
...
+          if (unblockingOutput) {
+            messages.push({ role: 'assistant', content: unblockingOutput });
+            previousTargetOutput = unblockingOutput;
+            finalOutput = unblockingOutput;
+            finalResponse = unblockingResponse;
+          }
+
+          if (unblockingResponse.error) {
+            throw new Error(`[GOAT] Target returned an error: ${unblockingResponse.error}`);
+          }
```
sklein12: does this blow up the whole thing?

src/redteam/providers/goat.ts
```diff
@@ -273,6 +281,73 @@ export default class GoatProvider implements ApiProvider {
...
+
+          if (!unblockingOutput) {
+            logger.debug(
```
sklein12: should this be a warning or info?

src/redteam/providers/goat.ts
```diff
@@ -343,4 +418,65 @@ export default class GoatProvider implements ApiProvider {
       guardrails: lastTargetResponse?.guardrails,
     };
   }
+
+  private async tryUnblocking(
```
sklein12: is this copied from the other strategy? can we abstract this out?

src/redteam/providers/goat.ts
```diff
@@ -343,4 +418,65 @@ export default class GoatProvider implements ApiProvider {
...
+      const { checkServerFeatureSupport } = await import('../../util/server');
+      return await checkServerFeatureSupport(
+        'blocking-question-analysis',
+        '2025-06-16T14:49:11-07:00',
+      );
+    } catch (error) {
+      logger.debug(`[GOAT] Feature check failed: ${error}`);
+      return false;
+    }
+  }
```
sklein12: this shouldn't be copied

src/redteam/providers/goat.ts
MrFlounder: you are right! will change this to an error log

src/redteam/providers/goat.ts
MrFlounder: I think we can remove this

MrFlounder:
I think goat works with remote generation only on a second look, so I think we don't need to specifically consider that condition outside constructor right? @mldangelo

PR #4622, opened 2025-06-24
chore(cli): Removes context schema validation from extension hooks

will-holley:
- Removes runtime validation of `context` arg passed into the hook runner; it results in errors because our test suite zod schemas is a subset of the potential configuration values
- Removes runtime validation of `context` returned from hooks, also because of the subset.
  - Note: this change allows user code to mutate and return `context` such that it will break downstream execution. This is preferable because the user can correct the offending code within the hook definition.

### DIFF
--- a/src/evaluatorHelpers.ts
+++ b/src/evaluatorHelpers.ts
@@ -1,7 +1,6 @@
 import * as fs from 'fs';
 import yaml from 'js-yaml';
 import * as path from 'path';
-import { z } from 'zod';
 import cliState from './cliState';
 import { getEnvBool } from './envars';
 import { importModule } from './esm';
@@ -17,11 +16,9 @@ import {
   type ApiProvider,
   type NunjucksFilterMap,
   type Prompt,
-  TestCaseSchema,
   type TestSuite,
   type CompletedPrompt,
   type EvaluateResult,
-  TestSuiteSchema,
   type TestCase,
 } from './types';
 import { renderVarsInObject } from './util';
--- a/src/evaluatorHelpers.ts
+++ b/src/evaluatorHelpers.ts
@@ -340,40 +340,13 @@ export async function renderPrompt(
 // Extension Hooks
 // ================================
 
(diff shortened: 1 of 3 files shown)


src/evaluatorHelpers.ts
```diff
@@ -460,33 +408,26 @@ export async function runExtensionHook<HookName extends keyof HookContextMap>(
...
+              ...(context as BeforeAllExtensionHookContext).suite,
+              // Mutable properties:
+              prompts: extensionReturnValue.suite.prompts,
+              providerPromptMap: extensionReturnValue.suite.providerPromptMap,
+              tests: extensionReturnValue.suite.tests,
+              scenarios: extensionReturnValue.suite.scenarios,
+              defaultTest: extensionReturnValue.suite.defaultTest,
+              nunjucksFilters: extensionReturnValue.suite.nunjucksFilters,
+              derivedMetrics: extensionReturnValue.suite.derivedMetrics,
+              redteam: extensionReturnValue.suite.redteam,
```
sklein12, line 422: what's the logic behind why only some things are mutable?

will-holley: All properties except `providers`, `env` and `extensions` are considered mutable.

Within this function `context.suite.providers` is an array of provider instances which gets serialized into something like `{ id: 'provider-id' }` when it's passed into the hook scope.  As such, we want to maintain the original instance array.

Modifying `env` vars here didn't seem appropriate.

Extensions modifying itself (`extensions`) is a can of worms I don't see a reason to open.

sklein12: What if we add a new property to test suite, will we need to add it to this? Can you think of anyways we don't accidentally introduce a bug in that case? I don't have a solution off the top of my head it could be a test, typing, constants, etc...

sklein12 (approved):
Good to merge, lets follow-up with this in another PR https://github.com/promptfoo/promptfoo/pull/4622#discussion_r2164930481

PR #4647, opened 2025-06-26
fix: support derived metrics in eval config uploaded via WebUI

### LINKED ISSUE #4611: Derived metrics doesn't work if eval config is uploaded via WebUI
**Describe the bug**
Derived metrics are not being processed or calculated when eval configurations are uploaded via the web UI. After uploading a configuration YAML file that includes derived metrics, the derived metrics section is stripped from the config and the metrics are not calculated in the evaluation results, even when manually added back to the configuration.

**To Reproduce**
Steps to reproduce the behavior, including example Promptfoo configurations:

1. Navigate to the Promptfoo web UI
2. Go to Create -> Eval
3. Click on "Edit Yaml"
4. Upload or paste the YAML configuration from the F1 metrics example at https://github.com/promptfoo/promptfoo/blob/main/examples/f-score/promptfooconfig.yaml
5. Run the evaluation
6. Observe that the derived metrics are not present in the uploaded config and F1 scores are not calculated
7. Manually add the derived metrics back to the configuration
8. Run the evaluation again
9. See that derived metrics are still ignored and not calculated

**Expected behavior**
When uploading a configuration file with derived metrics via the web UI, the derived metrics should be preserved in the configuration and calculated during evaluation. The F1 score (and other derived metrics) should appear in the evaluation results alongside the base metrics like precision and recall.

**Screenshots**

Image

Image

**System information:**
N/A - Bug is in WebUI eval only and there is no debug command available there.

- Promptfoo version:
0.115.1

**Additional context**
Additional context
This issue specifically affects the web UI workflow. The same configuration file works correctly when run via CLI. The derived metrics functionality appears to be completely bypassed when configurations are processed through the web interface, making it impossible to calculate composite metrics like F1 scores through the web UI workflow. I initially discovered this issue on my config, but then reproduced on the Promptfoo example - so I use only promptfoo original example in this description.
Test case reference: The issue can be reproduced using the official F1 score example configuration from the promptfoo repository at examples 
https://github.com/promptfoo/promptfoo/blob/main/examples/f-score/promptfooconfig.yaml

yexingyao:
Fixes #4611 

This fix populates derived metrics config from WebUI to the JS data store and the eval job submitted.

Added unit tests and verified that the submitted eval job contains the derived metrics.

### DIFF
--- a/src/app/src/pages/eval-creator/components/YamlEditor.tsx
+++ b/src/app/src/pages/eval-creator/components/YamlEditor.tsx
@@ -100,6 +100,10 @@ const YamlEditorComponent: React.FC<YamlEditorProps> = ({
             newState.defaultTest = parsedConfig.defaultTest;
           }
 
+          if (parsedConfig.derivedMetrics !== undefined) {
+            newState.derivedMetrics = parsedConfig.derivedMetrics;
+          }
+
           if (parsedConfig.evaluateOptions !== undefined) {
             newState.evaluateOptions = parsedConfig.evaluateOptions;
           }
--- a/src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.test.tsx
+++ b/src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.test.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { useStore } from '@app/stores/evalConfig';
 import { ThemeProvider, createTheme } from '@mui/material/styles';
 import type {
+  DerivedMetric,
   ProviderOptions,
   TestCase,
   EvaluateOptions,
@@ -65,6 +66,7 @@ const defaultStoreValues = {
   prompts: [] as any[],
   testCases: [] as TestCase[],
   defaultTest: {} as TestCase,
+  derivedMetrics: [] as DerivedMetric[],
   env: {} as EnvOverrides,
   evaluateOptions: {} as EvaluateOptions,
   scenarios: [] as Scenario[],
@@ -146,6 +148,7 @@ describe('EvaluateTestSuiteCreator', () => {
       prompts: ['Test Prompt {{var}}'],
       testCases: [{ vars: { var: 'value' } }] as TestCase[],
       defaultTest: { assert: [{ type: 'equals', value: 'expected' }] } as TestCase,
+      derivedMetrics: [{ name: 'precision', value: 'formula' }] as DerivedMetric[],
       env: { OPENAI_API_KEY: 'testkey' } as EnvOverrides,
       evaluateOptions: { maxConcurrency: 5 } as EvaluateOptions,
       scenarios: [{ description: 'Test Scenario', config: [], tests: [] }] as Scenario[],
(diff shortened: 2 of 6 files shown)


src/app/src/pages/eval-creator/components/YamlEditor.tsx
```diff
@@ -236,7 +240,7 @@ const YamlEditorComponent: React.FC<YamlEditorProps> = ({
             }}
             sx={{ ml: 2, whiteSpace: 'nowrap' }}
           >
-            {isReadOnly ? 'Edit YAML' : 'Save Changes'}
+            {isReadOnly ? 'Edit YAML' : 'Save Test Changes'}
```
yexingyao: Not part of the change (added just to verify I can test my local code). Will revert before merging

src/app/src/pages/eval-creator/components/PromptsSection.tsx
```diff
@@ -158,6 +158,7 @@ const PromptsSection: React.FC = () => {
                     <Typography variant="body2">
                       {`Prompt #${index + 1}: `}
                       {(prompt.length > 250 ? prompt.slice(0, 250) + ' ...' : prompt)
+                        .toString()
```
yexingyao: Unrelated to the fix. Just to prevent long exception message in this part of the UI from uploaded YAML file.

yexingyao:
Screenshot of successfully updating YAML in WebUI without derived metrics being dropped
Screenshot 2025-06-25 at 10 57 15 PM

Screenshot of successful evaluation run with derived metrics computed
Screenshot 2025-06-25 at 11 15 28 PM

src/app/src/pages/eval-creator/components/PromptsSection.tsx
MrFlounder: I wonder what in the config could cause the exception and what user experience would be in that case. Not related to this pr though

src/app/src/pages/eval-creator/components/YamlEditor.tsx
MrFlounder: Sorry for not getting the idea, why this is needed to test local code?

MrFlounder (commented):
Thank you for making the fix! After CI passes we can ship it!

src/app/src/pages/eval-creator/components/YamlEditor.tsx
yexingyao: This is the first code I changed locally to validate that I was running a version of the WebUI based on my local code :)

This change actually broke the test so I'm reverting it now

src/app/src/pages/eval-creator/components/PromptsSection.tsx
yexingyao: The `prompt` from YAML config is an object not a string. So it triggers an exception of

"(...).split(" is not a function

When I convert this to a string it prints "Object" w/o throwing the exception. But the ideal fix here is to parse and display the object.

There is another similar issue on my test case where the TestCases object from YAML failing to parse and display in the following component on the WebUI. I'll create a new issue for these YAML parsing and display gaps.

yexingyao:
Removed temporary local testing code and passed WebUI tests locally.

I should have made another commit but I amended the original commit and forced push it. Let me know if you are okay with this.

@MrFlounder @mldangelo This PR is updated and ready for review again. Are there any automated tests/workflows I should trigger?

mldangelo (approved):
Thank you so much @yexingyao!

PR #4665, opened 2025-06-27
docs(site): remove copy for llm button

mldangelo:
This PR removes the copy page button feature that was designed for LLM consumption.


typpo (approved):
Thx.  We can undo this with a fix for the text overlap issue 
image

Issue #4673, opened 2025-06-27
Answer-relevance assertion underreports token usage


erinemay:
**Describe the bug**

The answer-relevance assertion only logs token usage from failed API calls during candidate question generation, ignoring tokens used in successful calls.

**To Reproduce**

1. Set up a promptfoo config with `answer-relevance` assertions:
```yaml
prompts:
  - "{{query}}"
providers:
  - openai:gpt-4o-mini
tests:
  - vars:
      query: "What are your return policies?"
    assert:
      - type: answer-relevance
        threshold: 0.7
```

2. Run `promptfoo eval`
3. Check token usage output
4. See suspiciously low token usage in grading

**System information:**

- Promptfoo version: 0.115.4

**Additional context**
Token accumulation only happens in the error handling blocks (`if (resp.error || !resp.output)`). Successful API responses don't result in token accumulation within the loop. 
https://github.com/promptfoo/promptfoo/blob/39bb0239641b64ed5c0561e7fe37fcaee464964d/src/matchers.ts#L898-L936

typpo:
Thanks for the report, this should fix https://github.com/promptfoo/promptfoo/pull/4676

typpo:
This should be fixed in the next version with https://github.com/promptfoo/promptfoo/pull/4677

PR #4677, opened 2025-06-27
fix(matchers): track token usage for successful API calls

mldangelo:
Relates to #4673


sklein12 (approved):
ohhh, I like this one better than @typpo's

PR #4681, opened 2025-06-29
ci: verify that plugins are synced between code and documentation

typpo:
this check ensures that plugin ids in the constants file match the docs data file


scripts/verify-plugin-sync.ts
```diff
@@ -0,0 +1,158 @@
...
+
+// ANSI color codes for output
+const colors = {
+  red: '\x1b[31m',
+  green: '\x1b[32m',
+  yellow: '\x1b[33m',
+  reset: '\x1b[0m',
+};
+
+function error(message: string): void {
```
mldangelo: why not use logger?

mldangelo:
Thank you!

PR #4690, opened 2025-06-30
docs(redteam): add missing docs to sidebar

typpo:
also fixes the sidebar when not in the red-team category


site/sidebars.js
mldangelo, line 1: consider converting this to a ts file

site/docs/red-team/plugins/tool-discovery.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Tool Discovery Plugin
```
mldangelo, line 2: this could go into a llm agent rule

mldangelo (dismissed):

PR #4694, opened 2025-06-30
docs(site): add guide on humanity's last exam

mldangelo:


examples/huggingface-hle/README.md
```diff
@@ -0,0 +1,107 @@
+# huggingface-hle
```
ladyofcode, line 1: Does this need to be a proper title?

mldangelo: Good question! this is the format we follow in our examples (for the most part), so no

typpo (approved):
So cool!

PR #4704, opened 2025-06-30
fix(sagemaker): remove unsupported OpenAI model type

mldangelo:
## Description

This PR removes OpenAI model support from the Amazon SageMaker provider. OpenAI models are not natively supported by Amazon SageMaker endpoints.

## Changes

- Removed `'openai'` from the `modelType` type definition
- Removed `'openai'` from the `SAGEMAKER_MODEL_TYPES` array  
- Removed OpenAI-specific request formatting and response parsing logic
- Updated embedding provider to remove OpenAI-specific handling
- Removed all OpenAI-related tests and mock responses
- Deleted the OpenAI-specific example file (`promptfooconfig.openai.yaml`)
- Updated the multimodel example to use HuggingFace instead of OpenAI
- Fixed the deployment script to suggest the correct model type for Llama models

## Breaking Change

⚠️ **BREAKING CHANGE**: Removed `'openai'` from supported SageMaker model types. Configurations using `sagemaker:openai:*` will need to be updated to use alternative providers.

## Supported Model Types

The SageMaker provider now supports:
- `anthropic` - For Claude-compatible models
- `llama` - For Llama models  
- `huggingface` - For HuggingFace models
- `jumpstart` - For AWS JumpStart models
- `custom` - For custom model formats (default)

## Testing

All tests pass (60 tests) and the project builds successfully.

## Alternative for OpenAI Models

Users who need OpenAI models should:
- Use the OpenAI API directly via `openai:*` providers
- Use AWS Bedrock if they need AWS-managed access to these models

Fixes #[issue-number]


will-holley:
Please confirm that the provider currently supports OpenAI formatted models, not OpenAI models themselves.

PR #4728, opened 2025-07-01
fix: http body parsing when it comes from yaml string

MrFlounder:


mldangelo (dismissed):
this looks great, thank you. Would love to see a test.

PR #4739, opened 2025-07-02
fix: the case when http body has not escaped charactors

MrFlounder:
## Problem

Red team testing was failing intermittently because prompts with newlines or special characters would break JSON templates:

```yaml
body: '{"message": "{{prompt}}"}'
```

When `prompt` contains newlines, this creates invalid JSON:
```json
{"message": "Multi-line
text"}
```

Server responds with HTTP 400: "Invalid JSON payload received"

## Solution

Added automatic JSON escaping when templates fail to parse:
- First tries normal template rendering
- If JSON parsing fails, re-renders with properly escaped variables  
- Falls back to original behavior if still broken

## Result

✅ Red team prompts with newlines/quotes/special chars now work reliably  
✅ No breaking changes to existing configurations  
✅ Works with any JSON template string format (YAML, JS, etc.)

**Before:** Intermittent 400 errors during red team testing  
**After:** All prompts render as valid JSON

### DIFF
--- a/src/util/json.ts
+++ b/src/util/json.ts
@@ -3,7 +3,6 @@ import addFormats from 'ajv-formats';
 import yaml from 'js-yaml';
 import { getEnvBool, getEnvString } from '../envars';
 import type { EvaluateResult, ResultFailureReason } from '../types';
-import { renderVarsInObject } from '../util';
 import invariant from '../util/invariant';
 
 let ajvInstance: Ajv | null = null;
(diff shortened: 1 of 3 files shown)


src/util/json.ts
mldangelo, line 1: would like to see tests specifically on this file

mldangelo (approved):
Looks good! Thanks!

src/util/json.ts
MrFlounder: I moved the util to http.ts to avoid circular dependency - added more tests

PR #4760, opened 2025-07-04
feat(extensions): enhance afterAll hook and add --extension CLI flag

### LINKED ISSUE #4754: Allow output into javascript (and in future python maybe)
In our use case we have many promptfoo suites.
Sometimes when running one some of them have many failures due to TPM issues or tokens expiring.
We then have to open `promptfoo view` to dind these and know to run again or update the token

A simple solution would be to be able to set the output (one that is normally sent to csv, json ect file) to a javascript function (change in promptfoo) and then we can set this function to collate the error metadata and output it to console.

I also think that this could streamline data analysis in other cases for other people

We could add steps we manually run after each run but this is cumbersome to do.
We could create scripts that do the run, write to a file and run the analysis on top but this seems ineffecent and also cumbersom given multiple suites, it would be nice to modify the promptfooconfig.yaml and then have the info each time.

Happy to implement this if I get guidance on how we want this to fit in

mldangelo:
Closes #4754

## Summary

Enhances the extension hooks system to better support result processing use cases. Instead of overloading `--output` (which was semantically confusing), this PR:

1. **Enhances the existing `afterAll` hook** with additional context (`evalId`, `config`)
2. **Adds `-x/--extension` CLI flag** for running one-off extensions without modifying config
3. **Exports hook types** for TypeScript users

### Enhanced `afterAll` Hook Context

The `afterAll` hook now receives:
```typescript
{
  evalId: string;           // Unique evaluation ID
  config: UnifiedConfig;    // Full evaluation configuration  
  suite: TestSuite;         // Test suite that was evaluated
  results: EvalResult[];    // All individual test results
  prompts: CompletedPrompt[]; // Prompts with metrics
}
```

### New `--extension` CLI Flag

```bash
# Run with a one-off extension
promptfoo eval -x file://handler.js:afterAll

# Combine with config file extensions  
promptfoo eval -x file://alert-on-failure.js:afterAll -c promptfooconfig.yaml
```

### Exported Types

```typescript
import type { 
  AfterAllExtensionHookContext,
  BeforeAllExtensionHookContext,
  // etc.
} from 'promptfoo';
```

### New Example

Added `examples/result-hooks/` with JavaScript and Python examples demonstrating:
- Logging evaluation summaries
- Alerting on failure thresholds
- Sending metrics to monitoring systems

## Why This Approach?

The original implementation overloaded `--output` to both write files AND execute code, which was semantically confusing. This approach uses the existing extension hooks system which:

- Already supports `beforeAll`, `beforeEach`, `afterEach`, `afterAll` lifecycle hooks
- Has clear semantics (extensions = code hooks, output = file outputs)
- Is well-tested and documented
- Provides more power (4 hooks vs 1)

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -1808,22 +1808,25 @@ class Evaluator {
 
     this.evalRecord.setVars(Array.from(vars));
 
-    // Load results from database for extensions (results may not be in memory for persisted evals)
-    const allResults = await this.evalRecord.getResults();
+    // Only load results from database if there are extensions to run
+    if (testSuite.extensions?.length) {
+      // Load results from database for extensions (results may not be in memory for persisted evals)
+      const allResults = await this.evalRecord.getResults();
 
-    // Convert EvalResult model instances to plain EvaluateResult objects for extensions
-    const resultsForExtension: EvaluateResult[] = allResults.map(
-      (result): EvaluateResult =>
-        'toEvaluateResult' in result ? result.toEvaluateResult() : result,
-    );
+      // Convert EvalResult model instances to plain EvaluateResult objects for extensions
+      const resultsForExtension: EvaluateResult[] = allResults.map(
+        (result): EvaluateResult =>
+          'toEvaluateResult' in result ? result.toEvaluateResult() : result,
+      );
 
-    await runExtensionHook(testSuite.extensions, 'afterAll', {
-      prompts: this.evalRecord.prompts,
-      results: resultsForExtension,
-      suite: testSuite,
-      evalId: this.evalRecord.id,
-      config: this.evalRecord.config,
-    });
+      await runExtensionHook(testSuite.extensions, 'afterAll', {
+        prompts: this.evalRecord.prompts,
+        results: resultsForExtension,
+        suite: testSuite,
+        evalId: this.evalRecord.id,
+        config: this.evalRecord.config,
+      });
+    }
 
     // Calculate additional metrics for telemetry
     const endTime = Date.now();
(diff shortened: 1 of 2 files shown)


examples/simple-test/promptfooconfig.yaml
```diff
@@ -51,7 +51,7 @@ tests:
       # $ref can be used as a shorthand to include reusable templates
       - $ref: '#/assertionTemplates/containsPirateNoise'
 
-outputPath: file://output.csv
+outputPath: output.csv
```
mldangelo:
```suggestion
outputPath: file://output.csv
```

site/docs/configuration/outputs.md
```diff
@@ -110,6 +110,100 @@ promptfoo eval --output results.yaml
...
+promptfoo eval --output file://handle-results.js
+promptfoo eval --output file://handle_results.py
+
+# With explicit function name
+promptfoo eval --output file://handle-results.js:processResults
+promptfoo eval --output file://handle_results.py:save
+
+# Direct paths also work (extension-based inference)
+promptfoo eval --output ./handlers/results.js
+promptfoo eval --output ./handlers/results.py:save_results
```
mldangelo:
```suggestion
```

src/commands/eval.ts
```diff
@@ -850,6 +859,22 @@ export function evalCommand(
       }
 
       for (const maybeFilePath of validatedOpts.output ?? []) {
+        // Skip validation for handler paths (file:// prefix or script files)
+        const isHandlerPath =
+          maybeFilePath.startsWith('file://') ||
+          maybeFilePath.endsWith('.js') ||
```
mldangelo: use isJavascript file here

JustinBeckwith (approved):

Issue #4770, opened 2025-07-07
Posthog errors even after disabling telemtry


QuarkNerd:
**Describe the bug**
Due to our firewall settings telemtry calls to posthog cause errors and early exits.
After disabling telemetry, there are errors shown during running.

**To Reproduce**
I'm unable to give exact steps since I don't know what our firewall setup is like but for us
1. DIsable telemtery with `$env:PROMPTFOO_DISABLE_TELEMETRY=1`
2. Run promptfoo eval

**Expected behavior**
A clean run with not logging of errors

**Screenshots**
Due to security policy unable to provide screenshots but the error reads

`Error while flushing PostHog PostHogFetchNetworkError: Network error while fetching PostHog`
the error seems to be coming from `node-modules\posthog-core\src\index.ts:1130:17` which is fun since I cant even see that module
The "cause" is `DOMException [TimeoutError]: The operation was aborted due to timeout`

- Promptfoo version:
promptfoo version: 0.116.2

**Additional context**
The error I see here is different to the one without setting the telemtry promptfoo env variable, the other one crashes the run.

mldangelo:
This was resolved in the latest release. Thanks for flagging @QuarkNerd!

PR #4772, opened 2025-07-07
fix(telemetry): prevent PostHog initialization when telemetry is disabled

mldangelo:
## Description

This PR addresses PostHog network errors that occur even when telemetry is disabled via `PROMPTFOO_DISABLE_TELEMETRY=1`.

## Problem

Users behind firewalls were experiencing PostHog network errors and timeouts even after disabling telemetry, causing the application to crash or show error messages.

## Solution

- Implement lazy initialization of PostHog client
- Check PROMPTFOO_DISABLE_TELEMETRY environment variable before creating PostHog instance
- Add comprehensive error handling for all PostHog operations
- Disable automatic flushing with flushAt: 0 and flushInterval: 0
- Wrap all PostHog calls in try-catch blocks to prevent crashes

## Related Issue

Related to #4770

### DIFF
--- a/src/telemetry.ts
+++ b/src/telemetry.ts
@@ -30,14 +30,27 @@ const EVENTS_ENDPOINT = 'https://a.promptfoo.app';
 const KA_ENDPOINT = 'https://ka.promptfoo.app/';
 
 let posthogClient: PostHog | null = null;
-try {
-  posthogClient = POSTHOG_KEY
-    ? new PostHog(POSTHOG_KEY, {
+
+// Lazy initialization of PostHog client
+function getPostHogClient(): PostHog | null {
+  // Check if telemetry is disabled before creating the client
+  if (getEnvBool('PROMPTFOO_DISABLE_TELEMETRY') || process.env.IS_TESTING) {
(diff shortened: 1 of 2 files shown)


QuarkNerd:
Thank  you gor the quick start to this.

A small related but tangential ask. Would it be possible to have PROMPTFOO_DISABLE_TELEMETRY something we can set in promptfoo config yaml too?

sklein12 (dismissed):

mldangelo:
> A small related but tangential ask. Would it be possible to have PROMPTFOO_DISABLE_TELEMETRY something we can set in promptfoo config yaml too?

Unfortunately this is not easy to do without completely breaking telemetry. Telemetry is initialized very early when the CLI starts, before any config files are loaded. By the time your config file is read, telemetry has already checked `process.env.PROMPTFOO_DISABLE_TELEMETRY` and started up. You can set the environment variable directly `export PROMPTFOO_DISABLE_TELEMETRY=1` or use a `.env` file in your project root `PROMPTFOO_DISABLE_TELEMETRY=1`.

PR #4788, opened 2025-07-07
chore(providers): Improve guardrails handling in Azure providers

will-holley:
This PR ensures that the Azure chat completion provider correctly handles responses which indicate that Azure OpenAI's content filtering system flagged an input prompt or output completion.  It also handles the case where the content filtering system _should have_ but did not run.


mldangelo:
If @MrFlounder has time, this would be a good one for him to look at!

MrFlounder (approved):
Thank you!

PR #4809, opened 2025-07-08
chore(providers/sagemaker): Improves validation of user-provided config

will-holley:
- Validates user-provided config against Zod schema (to ensure correct types, keys, etc.); logs a warning if validation fails.
- Ensures that, for the completion provider, the model type is provided either within the provider id or in the `modelType` config parameter. Otherwise, unless the user intends to use a `custom` model, request payloads will be incorrectly serialized.

### DIFF
--- a/src/providers/sagemaker.ts
+++ b/src/providers/sagemaker.ts
@@ -446,13 +446,13 @@ export class SageMakerCompletionProvider extends SageMakerGenericProvider implem
         return modelType;
       } else {
         throw new Error(
-          `Invalid model type "${modelType}" in config. Valid types are: ${SUPPORTED_MODEL_TYPES.join(', ')}`,
+          `Invalid model type "${modelType}" in \`config.modelType\`. Valid types are: ${SUPPORTED_MODEL_TYPES.join(', ')}`,
         );
       }
     }
 
     throw new Error(
-      'Model type must be provided either in config.modelType or as part of the provider ID format "sagemaker:<model_type>:<endpoint>"',
+      'Model type must be set either in `config.modelType` or as part of the Provider ID, for example: "sagemaker:<model_type>:<endpoint>"',
     );
   }
 
--- a/src/providers/sagemaker.ts
+++ b/src/providers/sagemaker.ts
@@ -95,15 +95,15 @@ export abstract class SageMakerGenericProvider {
 
     // Validate the config
     try {
-      this.config = config ? SageMakerConfigSchema.parse(config) : {};
+      SageMakerConfigSchema.parse(config);
     } catch (error) {
       const zodError = error as z.ZodError;
-      logger.error(
+      logger.warn(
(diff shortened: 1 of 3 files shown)


src/providers/sagemaker.ts
```diff
@@ -22,42 +23,50 @@ import type { TransformContext } from '../util/transform';
...
+    secretAccessKey: z.string().optional(),
+    sessionToken: z.string().optional(),
+
+    // SageMaker specific options
+    endpoint: z.string().optional(),
+    contentType: z.string().optional(),
+    acceptType: z.string().optional(),
+
+    // Model parameters
+    maxTokens: z.number().optional(),
```
mldangelo, line 46: I am slightly reluctant to do this and think we should prefer to pass through instead when possible. We don't want promptfoo to break every time they add or remove a field from provider configuration in sagemaker. A pattern we follow in other places in the code base is to have validation on the important things. that we think will never change and then no validation on fields like maxTokens

src/providers/sagemaker.ts
```diff
@@ -86,7 +95,18 @@ export abstract class SageMakerGenericProvider {
...
+
+    // Validate the config
+    try {
+      this.config = config ? SageMakerOptionsSchema.parse(config) : {};
+    } catch (error) {
+      const zodError = error as z.ZodError;
+      logger.error(
+        `Error validating SageMaker config\nConfig: ${JSON.stringify(config)}\nIssues: ${JSON.stringify(zodError.issues)}`,
+      );
+      throw new Error('Invalid SageMaker config');
```
mldangelo: Would be much better to just try to continue but log the error

src/providers/sagemaker.ts
will-holley: I largely agree except:

1. We're already hardcoding the request payload serializers which would need to be updated when they add or remove fields. One option to solve for both of these is moving the serialization logic onto the config, following the transform function pattern.
2. I've observed passing invalid values (e.g. incorrectly cased keys) resulting in opaque error messages; these errors should be caught as early in the callstack as possible and should be easy to fix.

will-holley: I've migrated from throwing an error to logging a warning when schemas are invalid (https://github.com/promptfoo/promptfoo/pull/4809/commits/cd790c061cd26eb442ed1855c301a5d8b9d7a418 ).

src/providers/sagemaker.ts
```diff
@@ -65,28 +80,30 @@ interface SageMakerOptions {
...
+
+    // Validate the config
+    try {
+      SageMakerConfigSchema.parse(config);
+    } catch (error) {
+      const zodError = error as z.ZodError;
+      logger.warn(
+        `Error validating SageMaker config\nConfig: ${JSON.stringify(config)}\nIssues: ${JSON.stringify(zodError.issues)}`,
+      );
+    }
```
sklein12: you can use the zod-validation-error package to print out nice error messages from zod errors

PR #4812, opened 2025-07-08
chore: Update website with MCP Proxy

sklein12:
image

image
image
image
image


site/src/pages/index.tsx
```diff
@@ -169,6 +171,37 @@ function HomepageWalkthrough() {
     },
     {
       id: 4,
+      caption: 'MCP Proxy',
+      mobileCaption: 'MCP Proxy',
```
typpo:
```suggestion
      caption: 'MCP',
      mobileCaption: 'MCP',
```

typpo (dismissed):
sick change

site/src/pages/index.tsx
typpo: I think we can just drop the Proxy here to make it cleaner 😎

PR #4820, opened 2025-07-09
chore: better remote grading logs

sklein12:


src/remoteGrading.ts
```diff
@@ -29,8 +29,21 @@ export async function doRemoteGrading(
       REQUEST_TIMEOUT_MS,
     );
 
+    logger.debug(
+      `Got remote grading result: status=${status}, statusText=${statusText}, data=${JSON.stringify(data)}`,
+    );
+
+    if (status !== 200) {
```
mldangelo, line 36: consider response.ok

src/remoteGrading.ts
```diff
@@ -29,8 +29,21 @@ export async function doRemoteGrading(
...
+      throw new Error(
+        `Remote grading failed with status ${status}: ${statusText} ${JSON.stringify(data)}`,
+      );
+    }
     const { result } = data as { result: GradingResult };
-    logger.debug(`Got remote grading result: ${JSON.stringify(result)}`);
+
+    if (!result) {
```
mldangelo: if you're doing this you might as well consider checking if pass exists (only critical field)

mldangelo (approved):
Great! Please make sure you test it in an environment where remote grading fails and you should cover this behavior with a test. I've taken more than one support call about remote grading failing. Want to make sure it doesn't blow up the eval in a non-desirable / un-intentional way.

mldangelo:
I am so sad there are no tests

PR #4827, opened 2025-07-09
fix: pass env vars to MCP server

sklein12:


ainthek:
Please is it possible to use env variables for mcp with http transport ? It seems NO by experiments and quick code view. Is there any reason why ? I need to set api keys or headers based on host env. Thanx

PR #4866, opened 2025-07-11
feat(webui): add metadata filtering

mldangelo:
Adds metadata filtering to evaluation results table with wildcard support. Users can filter by metadata keys or key:value pairs (e.g., model:gpt-4, model:gpt-*, model:*turbo).


will-holley:
Closing in favor of https://github.com/promptfoo/promptfoo/pull/4930.

PR #4869, opened 2025-07-11
feat: add system prompt support for azure provider

MrFlounder:


sklein12 (approved):
love it! We should copy this to OpenAi too!

PR #4876, opened 2025-07-11
chore(internals): Posthog key loading

will-holley:
Building locally writes your PostHog key to `src/generated-constants.ts`, which is very easy to accidentally commit.  This PR explores alternative patterns for injecting the PostHog key into the publishable build.


mldangelo (commented):
This is a better approach. Get the build to pass!

PR #4915, opened 2025-07-14
fix(cli): esm provider loading

will-holley:
- https://linear.app/promptfooo/issue/PF-1057/support-es-modules-for-custom-providers
- This PR fixes a bug where `import` statements get converted to `require`, breaking ESM imports.
- Adds error callstack logging to make debugging future errors easier for end users.


mldangelo (commented):
The code your are editing is both janky and load bearing. I am ok with this as long as it's thoroughly tested. Import from ESM app, import from CJS app, importing both types of providers, etc.

PR #4923, opened 2025-07-15
docs: Add examples for gemini URL context and code execution tools 

adelmuursepp:
Adding the Google AI Studio/Gemini provider documentation with examples for code execution and URL context tool usage and documenting image understanding in the main provider docs.

#### What's new:

  - Tool examples: Added configurations for code execution (`promptfooconfig.codeexecution.yaml`) and URL
   grounding (`promptfooconfig.urlcontext.yaml`)
  - Updated image example to have a slightly more complex use case (unfortunately more advanced image understanding features rely on showing media in the output table that requires more work)

#### Next to do:

I will need to have a look at why Google providers at times give different output: instead of output text containing just text it contains tool call information and image data, etc. This is not allowing me to show media in the outputs for more advanced multimodal features, like image generation, showing charts from code execution, etc. I will also need to be careful to not break any existing work that relies on this formatting.


site/docs/providers/google.md
```diff
@@ -342,6 +342,21 @@ For multimodal inputs (images and video), the provider supports:
 - Images: PNG, JPEG, WEBP, HEIC, HEIF formats (max 3,600 files)
 - Videos: MP4, MPEG, MOV, AVI, FLV, MPG, WEBM, WMV, 3GPP formats (up to ~1 hour)
```
adelmuursepp, line 343: For videos I am getting that request is too large, I will need to look into it. And also consider supporting adding youtube/external URLs to videos

mldangelo:
Thank you @adelmuursepp!!!!

PR #4924, opened 2025-07-15
fix(redteam): Prevent redteam run from hanging when using an mcp client

faizanminhas:
### Summary

I observed an issue where when an mcp server was configured, and `redteam run` was being run, the server would be initialized twice, and the process would hang once the test was over. This seems to be because the providers are initialized once during the redteam generate step, and then once again during eval. By always doing a cleanup at the end of generate, rather than only when we're not in a `redteam run` like before, this issue is fixed. 

Ideally we wouldn't initialize providers twice, but I think it's okay, since it's fairly harmless and possibly cleaner to isolate the two commands.

### DIFF
--- a/src/redteam/commands/generate.ts
+++ b/src/redteam/commands/generate.ts
@@ -414,20 +414,23 @@ export async function doGenerateRedteam(
     printBorder();
     const relativeOutputPath = path.relative(process.cwd(), options.output);
     logger.info(`Wrote ${redteamTests.length} test cases to ${relativeOutputPath}`);
-    if (!options.inRedteamRun) {
-      // Provider cleanup step
-      try {
-        const provider = testSuite.providers[0] as ApiProvider;
-        if (provider && typeof provider.cleanup === 'function') {
-          const cleanupResult = provider.cleanup();
-          if (cleanupResult instanceof Promise) {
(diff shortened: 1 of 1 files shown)


src/redteam/commands/generate.ts
```diff
@@ -414,20 +414,23 @@ export async function doGenerateRedteam(
...
-          if (cleanupResult instanceof Promise) {
-            await cleanupResult;
-          }
+
+    // Provider cleanup step. Note that this should always be run,
+    // since the providers are re-initialized when running the red team,
+    // hence it's safe and necessary to clean-up, particularly for MCP servers
+    try {
+      const provider = testSuite.providers[0] as ApiProvider;
+      if (provider && typeof provider.cleanup === 'function') {
```
will-holley, line 424: Since this should always be run, and assumedly there would be a problem if it didn't, would it be valuable to add a debug line here, for example `logger.debug("Cleaning up provider")`? Then, if this edge case is encountered and the user provides us with debug logs, we'll have a better idea that this was in-fact causal, rather than us observing some downstream effect from the lack of cleanup.

faizanminhas: this is a great idea, I'll add a log

PR #4954, opened 2025-07-16
fix: additional checking on llm-rubric response

sklein12:
Additional logging and error handling for bad llm-rubric responses

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -543,15 +543,21 @@ export async function matchesLlmRubric(
     );
   }
 
-  if(!Array.isArray(jsonObjects) || jsonObjects.length === 0) {
-    return fail(`llm-rubric produced malformed response - We were not able to parse the response as JSON. Output: ${JSON.stringify(resp.output)}`, resp.tokenUsage);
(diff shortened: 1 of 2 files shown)


src/matchers.ts
```diff
@@ -538,14 +538,28 @@ export async function matchesLlmRubric(
     jsonObjects = [resp.output];
   } else {
     return fail(
-      'llm-rubric produced malformed response - output must be string or object',
+      `llm-rubric produced malformed response - output must be string or object. Output: ${JSON.stringify(resp.output)}`,
```
faizanminhas, line 541: this is in response to the error logs? It doesn't feel very actionable.

sklein12: this gets shown to the user in their UI/eval results.

sklein12: it helps with debugging because we can see the raw output from the LLM

faizanminhas: gotcha, that makes sense upon thinking about it

faizanminhas (approved):

PR #4979, opened 2025-07-18
feat: New Sample Target

sklein12:
New sample target and configuration to better show off our capabilities :)


faizanminhas (approved):
I tried it out, 85% pass rate. Should be a much nicer example!

PR #4981, opened 2025-07-18
fix(redteam): find plugin assertion in strategy providers

MrFlounder:
Problem
Strategy providers (Crescendo, Custom, Iterative, IterativeTree) had faulty assertion selection logic that would prevent graders from running correctly in multi-assertion scenarios.

Root Cause:
Strategy providers were selecting the first assertion they could find, regardless of whether it matched the plugin
When multiple assertions existed (e.g., plugin assertions + other assertions), the wrong assertion could be selected
This caused grader failures because the system would try to use a plugin grader for a non-plugin assertion, or vice versa

Impact:
Tests with multiple assertions would fail to run graders correctly
Plugin-specific grading logic would not execute
Inconsistent behavior across different redteam strategies

### DIFF
--- a/src/redteam/providers/crescendo/index.ts
+++ b/src/redteam/providers/crescendo/index.ts
@@ -277,9 +277,15 @@ export class CrescendoProvider implements ApiProvider {
       content: systemPrompt,
     });
 
-    const assertToUse = test?.assert?.find(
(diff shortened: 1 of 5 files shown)


mldangelo (dismissed):
Please document this well enough that people know what it does and why we built it! Thanks!

PR #5023, opened 2025-07-23
chore(webui): Eval Results UI Tweaks

will-holley:
https://linear.app/promptfooo/issue/PF-1159/eval-results-ui-tweaks-feedback-implementation.

### DIFF
--- a/src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
+++ b/src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
@@ -202,15 +202,15 @@ function Filter({
    */
   const handleLogicOperatorChange = useCallback(
     (filterLogicOperator: ResultsFilter['logicOperator']) => {
-      if (index === 1) {
-        // If this is the second filter, update all filters to have the same logic operator
+      if (value.sortIndex === 1) {
+        // If this is the second filter (by sortIndex), update all filters to have the same logic operator
         updateAllFilterLogicOperators(filterLogicOperator);
       } else {
         // Otherwise just update this filter
         updateFilter({ ...value, logicOperator: filterLogicOperator });
       }
     },
-    [value, updateFilter, updateAllFilterLogicOperators, index],
+    [value, updateFilter, updateAllFilterLogicOperators],
   );
 
   const handleRemove = useCallback(() => {
@@ -251,12 +251,13 @@ function Filter({
       <Box sx={{ display: 'flex', gap: 1.5, flex: 1, alignItems: 'center', flexWrap: 'wrap' }}>
         {index !== 0 &&
           (() => {
-            // Get the second filter's logic operator to use for all filters when index > 1
-            const filtersList = Object.values(filters.values);
-            const secondFilterLogicOperator =
-              filtersList.length > 1 ? filtersList[1].logicOperator : null;
+            // Get the filter with sortIndex === 1 to determine the logic operator for all filters
+            const filterWithSortIndex1 = Object.values(filters.values).find(
+              (f) => f.sortIndex === 1,
+            );
+            const secondFilterLogicOperator = filterWithSortIndex1?.logicOperator || null;
             const displayValue =
-              index === 1
+              value.sortIndex === 1
                 ? (value.logicOperator ?? 'and')
                 : (secondFilterLogicOperator ?? value.logicOperator ?? 'and');
 
@@ -270,7 +271,7 @@ function Filter({
                 value={displayValue}
                 onChange={(e) => handleLogicOperatorChange(e as ResultsFilter['logicOperator'])}
                 width={100}
-                disabled={index > 1}
+                disabled={value.sortIndex > 1}
               />
             );
           })()}
@@ -393,8 +394,8 @@ export default function FiltersForm({
   }, [filters.values, open, handleAddFilter]);
 
   const filterValuesList = useMemo(() => {
-    // Maintain insertion order - Object.values preserves insertion order in modern JS
-    return Object.values(filters.values);
+    // Sort by sortIndex to ensure consistent ordering
+    return Object.values(filters.values).sort((a, b) => a.sortIndex - b.sortIndex);
   }, [filters.values]);
 
   return (
--- a/src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
+++ b/src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
@@ -66,15 +66,13 @@ function Dropdown({
           },
         }}
       >
-        {values.map((item) => (
-          <MenuItem
-            key={item.value}
-            value={item.value}
-            disabled={disabledValues.includes(item.value)}
-          >
-            {item.label}
-          </MenuItem>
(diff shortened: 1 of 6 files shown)


src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
```diff
@@ -64,9 +66,13 @@ function Dropdown({
...
         }}
       >
-        {values.map((value) => (
-          <MenuItem key={value.value} value={value.value}>
-            {value.label}
+        {values.map((item) => (
+          <MenuItem
+            key={item.value}
+            value={item.value}
+            disabled={disabledValues.includes(item.value)}
```
faizanminhas: disabled vs removing from the list, wydt?

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
```diff
@@ -127,6 +133,13 @@ function Filter({
 }) {
   const { filters, updateFilter, removeFilter, updateAllFilterLogicOperators } = useTableStore();
 
+  // Get list of already selected metric values (excluding current filter)
+  const selectedMetricValues = useMemo(() => {
+    return Object.values(filters.values)
+      .filter((filter) => filter.id !== value.id && filter.type === 'metric' && filter.value)
+      .map((filter) => filter.value);
+  }, [filters.values, value.id]);
```
faizanminhas: doesn't need to be memo'd imho

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
```diff
@@ -189,15 +202,15 @@ function Filter({
...
     (filterLogicOperator: ResultsFilter['logicOperator']) => {
-      if (index === 1) {
-        // If this is the second filter, update all filters to have the same logic operator
+      if (value.sortIndex === 1) {
+        // If this is the second filter (by sortIndex), update all filters to have the same logic operator
         updateAllFilterLogicOperators(filterLogicOperator);
       } else {
         // Otherwise just update this filter
         updateFilter({ ...value, logicOperator: filterLogicOperator });
       }
```
faizanminhas: is this case ever reached?

src/app/src/pages/eval/components/ResultsTable.tsx
```diff
@@ -357,9 +358,29 @@ function ResultsTable({
...
+    return JSON.stringify(
+      appliedFilters.map((f) => ({
+        type: f.type,
+        operator: f.operator,
+        value: f.value,
+        field: f.field,
+        logicOperator: f.logicOperator,
+        sortIndex: f.sortIndex, // Include sortIndex for complete representation
+      })),
+    );
```
faizanminhas, line 379: does this go in the url? Do you need to `encodeURIComponent`?

src/app/src/pages/eval/components/ResultsTable.tsx
```diff
@@ -1056,8 +1076,30 @@ function ResultsTable({
...
+            position: 'absolute',
+            top: 0,
+            left: 0,
+            right: 0,
+            bottom: 0,
+            display: 'flex',
+            alignItems: 'center',
+            justifyContent: 'center',
+            backgroundColor: (theme) =>
+              theme.palette.mode === 'dark' ? 'rgba(0, 0, 0, 0.7)' : 'rgba(255, 255, 255, 0.8)',
```
faizanminhas: do we have a material ui palette instead of raw colours like this?

src/app/src/pages/eval/components/ResultsView.tsx
```diff
@@ -518,9 +518,14 @@ export default function ResultsView({
...
           flexDirection: 'column',
-          // TODO(Will): This is a hack because the flexbox must have a fixed height in order for the pagination footer to be
-          // stuck to the bottom of the viewport; ideally the parent nodes are flexboxes.
-          height: `calc(100vh - 81px)`,
+        }}
+        ref={(el: HTMLDivElement | null) => {
+          const top = el?.getBoundingClientRect().top;
+          if (top) {
+            // TODO(Will): This is a hack because the flexbox must have a fixed height in order for the pagination footer to be
+            // stuck to the bottom of the viewport; ideally the parent nodes are flexboxes.
```
faizanminhas, line 526: Is that really true? Can you not do `position: absolute` or `position: sticky` to achieve something similar?

faizanminhas (dismissed):

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
will-holley: It shouldn't be because the selector is disabled when the sort index > 1.

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
will-holley: no strong opinions; i'll swap for removed

src/app/src/pages/eval/components/ResultsTable.tsx
will-holley: No

faizanminhas: should it? does pagination info go in the url?

will-holley: Managing pagination state in the url would be nice but we don't support it right now

src/app/src/pages/eval/components/ResultsView.tsx
will-holley: Otherwise the container will be as tall as its content. The same thing could be achieved with `position: absolute; left: 0; right: 0; bottom: 0; top: el?.getBoundingClientRect().top`.

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
```diff
@@ -206,8 +202,7 @@ function Filter({
         // If this is the second filter (by sortIndex), update all filters to have the same logic operator
         updateAllFilterLogicOperators(filterLogicOperator);
       } else {
-        // Otherwise just update this filter
-        updateFilter({ ...value, logicOperator: filterLogicOperator });
+        // noop:  this state should never be reached because the caller is disabled when index > 1.
       }
```
faizanminhas, line 206: nit: I'd just remove the else, or the whole if statement in that case

faizanminhas (approved):

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
will-holley: I was thinking the if/else with the comment acts as an invariant. thoughts?

faizanminhas: I don’t have a strong practical justification- would just note that the function should have a single purpose and assignment logic on the dropdown should be the thing determining if it’s called, having another guard here is slightly superfluous / coupling the action with the logic around it

Issue #3797, opened 2025-04-24
Truncated Test Name in CLI Progress Bar


arsena:
**Describe the bug**
The CLI progress bar doesn't display enough characters to see which test is being run.

**To Reproduce**
Run promptfoo eval

Sample Configuration:
```YAML
description: "Promptfoo test cases for different stories"
prompts:
  - ./test/promptfoo/common/prompts.mjs:a
  - ./test/promptfoo/common/prompts.mjs:b
providers:
  - id: file://./test/promptfoo/providers/MockAppProvider.mjs
    label: "MockApp"
defaultTest:
  options:
    provider:
      id: file://./test/promptfoo/providers/CustomGPTProvider.mjs
      label: "CustomGPT"
      config:
        api_version: "2024-02-01"
        chat_completions_model: "gpt-4o"
        embeddings_model: "text-embedding-ada-002"
        temperature: 0.7
        max_tokens: 2000
  assert:
    - type: answer-relevance
      threshold: 0.7
tests:
  - file://./test/promptfoo/*-stories.*
outputPath: ./test-reports/promptfoo/promptfoo-result.json
evaluateOptions:
  maxConcurrency: 8
```

**Expected behavior**
Expected more characters reserved for displaying variables, because our vars contain the input differences.  Therefor displaying more vars would indicate better which test is being run.

**Screenshots**

<img width="848" alt="Image" src="https://github.com/user-attachments/assets/d82e0487-6370-4043-a0cc-81e808bd154b" />

**System information:**

- Promptfoo version: 0.107.7

arsena:
BTW Promptfoo is excellent.

adityabharadwaj198:
@sklein12 implemented a small fix for this issue, would appreciate your thoughts on this!

PR #5041, opened 2025-07-24
Increased the number of characters for displaying test variables

adityabharadwaj198:
Fixes https://github.com/promptfoo/promptfoo/issues/3797

Changed the number of characters reserved for displaying file test variables to 40, so that more variables are displayed. 

Tested it by recreating tests from #3797. Used this file for vars:

```
question,expectedAnswer
"What is 2+2?","4"
"What is the capital of France?","Paris"
"Who wrote Romeo and Juliet?","Shakespeare"

```

image


mldangelo:
Thank you!

PR #4767, opened 2025-07-06
feat: Image understanding for Google providers

adelmuursepp:
### Core Changes

  - Fixed Google provider image processing. Added base64 image detection in `src/providers/google/util.ts`. 

  Before (Broken Format). What promptfoo sent to Google API:

```json
  {
    "contents": [
      {
        "parts": [
          {
            "text": "/9j/4AA
```

After adding the recognition of image files and formatting:

```json
  {
    "contents": [
      {
        "parts": [
          {
            "inlineData": {
              "mimeType": "image/jpeg",
              "data": "/9j/4A
```
  
Now Google providers are adding the image data inline (as explained in the [Google API docs](https://ai.google.dev/gemini-api/docs/image-understanding#inline-image)). This means image passing as inline data will work for AI studio and Vertex Google provider. Google Live will inherit same image processing improvements if this becomes available for Live API.

Image path is shown in metadata, however the prompt looks quite ugly with the base64 data. 

### Example UI:

image

  ### New Examples Added

  - Google AI Studio image example: examples/google-aistudio-gemini/promptfooconfig.image.yaml 
  panda image
  - Google Vertex image example: examples/google-vertex/promptfooconfig.image.yaml 

### Further Improvements
Test and showcase object recognition, image segmentation and with files available from URLs.

PR #5058, opened 2025-07-25
fix(providers): Preserve text formatting when no images present for Google provider

trojanowski:
Fixes `processImagesInContents` to only modify text parts that actually contain base64 images, preserving original formatting for text-only content.

#4767 introduced a regression where text-only parts were split by newline and each line was trimmed and sent as a separate part, breaking use cases that relied on original text layout.

Now only processes text when base64 images are actually found in the content, leaving other text formatting intact.


mldangelo:
Thank you @trojanowski! Please send me an email at michael @ promptfoo.dev if you would like some swag

PR #5086, opened 2025-07-28
feat(webui): add plugin and strategy filters for red team results

will-holley:
- Adds filtering by Plugins and Strategies
  - While filtering by plugin can be done using the metadata filter, it is very opaque (user needs to know `pluginId` is the correct key to filter by). This PR gives it first-class priority within the main filter-type selector.
- Configure filter behavior to treat plugin/strategy like metric filters (dropdown selection, equals-only operator)
- When user clicks "view logs" in vuln. report, pass plugin id instead of metric when redirecting to eval results


src/app/src/pages/redteam/report/components/TestSuites.tsx
```diff
@@ -327,14 +326,7 @@ const TestSuites: React.FC<TestSuitesProps> = ({ evalId, categoryStats, plugins
                         size="small"
                         onClick={() => {
                           const pluginId = subCategory.pluginName;
-                          const categoryType = subCategory.type;
```
will-holley, line 330: I couldn't repro a case where `pluginId` was undefined so I got rid of falling back to `categoryType`; unsure whether this is the correct behavior. Thoughts, @sklein12? Otherwise, I'd keep the metric filter when plugin id cannot be used.

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
```diff
@@ -279,6 +285,12 @@ function Filter({
               ? [{ label: TYPE_LABELS_BY_TYPE.metric, value: 'metric' }]
               : []),
             { label: TYPE_LABELS_BY_TYPE.metadata, value: 'metadata' },
+            ...(filters.options.plugin.length > 1
+              ? [{ label: TYPE_LABELS_BY_TYPE.plugin, value: 'plugin' }]
+              : []),
+            ...(filters.options.strategy.length > 1
+              ? [{ label: TYPE_LABELS_BY_TYPE.strategy, value: 'strategy' }]
```
sklein12: I think you should show these regardless, it's a little odd to have it only show up if there's more than 1. Kind of like when you're filtering on a store's website or airlines, even if there is only 1 option they show it in the filters for consistencies sake.

I went looking for it on one of my redteams and it wasn't there, it was a big leap for me to figure out it's because there's only 1 plugin, I think users will think it's broken.

sklein12: It also kinda breaks in this case when I got to the page from the report view

image

src/models/eval.ts
```diff
@@ -470,6 +470,19 @@ export default class Eval {
           } else if (operator === 'not_contains') {
             condition = `(json_extract(metadata, '$."${sanitizedField}"') IS NULL OR json_extract(metadata, '$."${sanitizedField}"') NOT LIKE '%${sanitizedValue}%')`;
           }
+        } else if (type === 'plugin' && operator === 'equals') {
+          const sanitizedValue = value.replace(/'/g, "''");
+          // Plugin ID is stored in test_case.metadata.pluginId
+          condition = `json_extract(test_case, '$.metadata.pluginId') = '${sanitizedValue}'`;
```
sklein12: This needs to be backed by an index. You can follow the same pattern we used for testCaseVars

https://github.com/promptfoo/promptfoo/blob/3bfa43cae8824e20516d5be2d61c27bc4403a758/src/database/tables.ts#L120-L122

src/models/eval.ts
```diff
@@ -470,6 +470,19 @@ export default class Eval {
...
+        } else if (type === 'strategy' && operator === 'equals') {
+          const sanitizedValue = value.replace(/'/g, "''");
+          if (sanitizedValue === 'basic') {
+            // Basic is represented by NULL in the test_case.metadata.strategyId field
+            condition = `(json_extract(test_case, '$.metadata.strategyId') IS NULL OR json_extract(test_case, '$.metadata.strategyId') = '')`;
+          } else {
+            // Strategy ID is stored in test_case.metadata.strategyId
+            condition = `json_extract(test_case, '$.metadata.strategyId') = '${sanitizedValue}'`;
+          }
```
sklein12: Same comment as above, this needs to be backed by an index

src/app/src/pages/redteam/report/components/TestSuites.tsx
sklein12: It's not that the `pluginId` is undefined, it was because the category type didn't have a direct mapping to a metric when it was `harmful:.....` As long as it works with sub categories like we have with harmful and pii I think we're good.

will-holley: Got it. I tested it with non-categorical (e.g. `rbac`) and categorical (e.g. `harmful`, `pii`, and `bias`) and it's working, so going to resolve this.

src/app/src/pages/redteam/report/components/RiskCategoryDrawer.test.tsx
```diff
@@ -149,7 +149,7 @@
     const viewAllLogsButton = screen.getByText('View All Logs');
     fireEvent.click(viewAllLogsButton);
 
-    expect(mockNavigate).toHaveBeenCalledWith('/eval/test-eval-123?metric=Test%20Category');
+    expect(mockNavigate).toHaveBeenCalledWith('/eval/test-eval-123?plugin=Test%20Category');
```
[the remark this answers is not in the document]
will-holley: This test is no longer necessary as the metric fallback has been removed. https://github.com/promptfoo/promptfoo/pull/5086/commits/0f07e190ad76496163456da564f67584807eff42.

src/app/src/pages/eval/components/ResultsFilters/FiltersForm.tsx
will-holley: Updated the former in https://github.com/promptfoo/promptfoo/pull/5086/commits/ddbc3f75a7b75e7d43c539e8db1ed8befaf71c7a but I'm not able to repro the latter.

will-holley:
**Note**: I added displaying the plugin or strategy id within the value selector (IMO, coming from the yaml, the id:label mapping isn't obvious), for example:

image

src/models/eval.ts
will-holley: Addressed in https://github.com/promptfoo/promptfoo/pull/5086/commits/dc2add3c2aebf189afccb4f703ef5151154ed01f.

src/models/eval.ts
will-holley: Addressed in https://github.com/promptfoo/promptfoo/pull/5086/commits/dc2add3c2aebf189afccb4f703ef5151154ed01f.

sklein12 (approved):
looks good after CI is passing!

PR #5099, opened 2025-07-29
fix: Adjust padding on docs sidebar to prevent overlap

faizanminhas:
### Before
Screenshot 2025-07-29 at 8 48 45 AM

### After
Screenshot 2025-07-29 at 8 47 39 AM

### DIFF
--- a/site/src/css/custom.css
+++ b/site/src/css/custom.css
@@ -71,7 +71,7 @@ code p {
 }
 
 .theme-doc-sidebar-container > div > div {
-  padding-top: 1rem !important;
+  padding-top: calc(1rem + 45px) !important;
 }
 
 .navbar__brand {


faizanminhas:
Coderabbit:
>

    73-75: Prefer CSS variable over hard-coded 45 px for future-proofing

    Hard-coding the extra 45 px couples the sidebar padding to today’s navbar height.
    Docusaurus exposes --ifm-navbar-height; using it keeps the rule correct if the navbar height changes (desktop vs. mobile, future redesigns) and avoids another override later.

    -  padding-top: calc(1rem + 45px) !important;
    +  /* `var(--ifm-navbar-height, 45px)` falls back to 45 px if the variable is absent */
    +  padding-top: calc(1rem + var(--ifm-navbar-height, 45px)) !important;

    The !important is still honoured, so no change in cascade behaviour.


I would guess that docusaurus must account for this normally- and as it turns out, removing that css override handles this exactly the same way. A win is a win.

Issue #5106, opened 2025-07-30
Sharing Breaks Due To Unicode Character Corruption During Red Team Evaluation


ben-bl:
We ran a set of 1200 tests, reviewed them in the local "view" and then tried to share the results to the cloud. 
The sharing failed repeatedly at one specific record.
Claude managed to fix the database records within the local promptfoo SQLite database so that the upload works now.

Below is the error description of that issue.
I also attach a Python script that was used to fix the values (if somebody else runs into this issue and wants to try a quick fix, use on your own risk and backup your db first): [fix_null_unicode_corruption.txt](https://github.com/user-attachments/files/21500820/fix_null_unicode_corruption.txt)

### Summary
Unicode characters are corrupted during promptfoo red team evaluations, resulting in null bytes that cause database insert failures when sharing evaluation results.

### Environment
- promptfoo version: Latest (as of 2025-07-30)
- Database: SQLite
- Node.js: v21.6.1
- Platform: macOS Darwin 24.5.0
- Red team strategies used: GOAT, crescendo, mischievous-user
- Test configuration: Multi-turn conversations with stateful strategies

### Description
During red team evaluation runs, Unicode characters (specifically the registered trademark symbol ®) are being corrupted by replacement with null bytes. The corruption pattern is consistent:
- Original: `Product® Plus`
- Corrupted: `Product\u0000ae Plus` (where \u0000 is a null byte followed by "ae")

This corruption appears in multiple database fields:
- test_case
- response  
- grading_result
- metadata
- prompt
- named_scores

### Impact
1. **Sharing Failure**: Evaluation sharing fails with HTTP 500 Internal Server Error
2. **Progress Blocking**: Share operation consistently stops at the same record (e.g., "59% | 683/1154 results")
3. **Database Integrity**: SQLite database contains invalid null characters preventing data export

### Error Details
```
Failed to send results chunk to https://www.promptfoo.app/api/v1/results/eval-XXX/results
status code: 500
status text: Internal Server Error
body: {
  "message": "Internal Server Error",
  "details": "Error: Failed query: insert into \"eval_results\" (\"id\", \"created_at\", 
             \"updated_at\", \"eval_id\", \"prompt_idx\", \"test_idx\", \"test_case\", 
             \"prompt\", \"prompt_id\", \"provider\", \"latency_ms\", \"cost\", \"response\", 
             \"error\", \"failure_reason\", \"success\", \"score\", \"grading_result\", 
[...]

PR #5108, opened 2025-07-30
fix: prevent Unicode corruption in Python providers

mldangelo:
## Description

This PR fixes Unicode character corruption that occurs when using custom Python providers, particularly during red team evaluations.

## Problem

Unicode characters (like ®, ™, ©) were being corrupted during JSON serialization between Node.js and Python, causing database insert failures when sharing evaluation results. For example:
- `Product® Plus` → `Product\u0000ae Plus` (with null byte corruption)

## Solution

Simple one-line fix in `wrapper.py` to use `ensure_ascii=False` when writing JSON output. This preserves Unicode characters properly without any escaping.

## Changes

- Modified `wrapper.py` to use `json.dump` with `ensure_ascii=False`
- Added comprehensive tests for Unicode handling in Python providers
- Added specific test to verify the fix works for the reported issue

## Why this approach?

- **Minimal**: Single line change in the Python wrapper
- **Root cause fix**: Addresses the JSON encoding issue directly
- **Backward compatible**: No API changes, just better Unicode preservation
- **Safe**: Standard Python JSON option, well-tested behavior

## Testing

Run the new tests:
```bash
npm test test/providers/pythonCompletion.bugfix.test.ts
npm test test/providers/pythonCompletion.unicode.test.ts
```

All existing Python provider tests continue to pass.

Relates to #5106

### DIFF
--- a/test/providers/pythonCompletion.bugfix.test.ts
+++ b/test/providers/pythonCompletion.bugfix.test.ts
@@ -1,0 +1,131 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { describe, it, expect, beforeEach, afterEach } from '@jest/globals';
+import { PythonProvider } from '../../src/providers/pythonCompletion';
+
+describe('PythonProvider Unicode bug #5106 fix verification', () => {
+  let tempDir: string;
+
+  beforeEach(() => {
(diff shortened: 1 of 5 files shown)


test/providers/pythonCompletion.bugfix.test.ts
typpo, line 1: Bad filename

PR #5119, opened 2025-07-30
fix(internals): Calling `npm run db:migrate` with Node 24

will-holley:


will-holley:
Closing in favor of https://github.com/promptfoo/promptfoo/pull/5123, which is a simpler fix.

PR #5121, opened 2025-07-30
fix(internals): `npm run local view` blank screen

will-holley:
- [ ] `src/main.tsx` is being served as TS
- [x] `scroll-timeline.js` was being picked up by the [splat route](https://github.com/promptfoo/promptfoo/blob/3a4542b782ed732427f3d22e26f97b657eed4afc/src/server/server.ts#L258).

### DIFF
--- a/src/app/index.html
+++ b/src/app/index.html
@@ -12,5 +12,5 @@
     <div id="root"></div>
     <script type="module" src="/src/main.tsx"></script>
   </body>
-  <script src="/scripts/polyfills/scroll-timeline.js"></script>
+  <script src="./public/scripts/polyfills/scroll-timeline.js"></script>
 </html>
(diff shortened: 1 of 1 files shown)


will-holley:
Now that I dig into this, I'm not sure `npm run view local` has worked for awhile now.

[`index.html` loads  `/src/main.tsx`](https://github.com/promptfoo/promptfoo/blame/3a4542b782ed732427f3d22e26f97b657eed4afc/src/app/index.html#L13) directly, which will be transpilled to JS during the build process.

will-holley:
Now that I dig into this, I'm not sure `npm run view local` has worked since the migration from Next.js: [`index.html` loads  `/src/main.tsx`](https://github.com/promptfoo/promptfoo/blame/3a4542b782ed732427f3d22e26f97b657eed4afc/src/app/index.html#L13) directly which doesn't work in the browser (and is transpiled to JS by the build).

The issue will `scroll-timeline.js` and the splat route doesn't occur in the production build, and view server doesn't work in dev regardless, so I'm going to consider this a nonissue and close the PR, @mldangelo.

PR #5123, opened 2025-07-30
fix(internals): Force CommonJS mode for db:migrate in Node 24

mldangelo:
## Summary

This PR provides an alternative fix to #5119 for the `npm run db:migrate` Node 24 compatibility issue.

Instead of creating a separate entry script, this approach uses `node --require tsx/cjs` to explicitly force CommonJS mode when running the migration script. This is a simpler, one-line change that achieves the same goal.

## Problem

Node 24's automatic syntax detection sees the `import` statements in `migrate.ts` and treats it as an ES module, causing the `require.main === module` check to fail.

## Solution

Update the `db:migrate` script in package.json to use:
```json
"db:migrate": "node --require tsx/cjs src/migrate.ts"
```

This forces tsx to use CommonJS mode, ensuring consistent behavior across Node 22 and Node 24.

### Node 22 (should continue working as before):
```bash
nvm use 22
npm ci
npm run db:migrate
# Should complete without errors
```

### Node 24 (the fix):
```bash
nvm use 24
npm ci
npm run db:migrate
# Should now complete without errors
```

## Related

- Alternative approach to #5119


mldangelo:
Thanks @coderabbitai for the review\! You were right about the compatibility concerns.

I've updated the PR to use `node --require tsx` instead of `tsx/cjs`. This is simpler and more universally compatible while still achieving the goal of forcing CommonJS mode.

The updated approach:
- Uses `node --require tsx src/migrate.ts` 
- Works across all tsx versions
- Maintains CommonJS behavior for the `require.main === module` check
- Tested successfully on Node 22

mldangelo:
Update: After more thorough testing, I've reverted to the `tsx/cjs` approach.

## Testing Results

### ❌ `node --require tsx` (previous attempt)
- ✅ Works with Node 22
- ❌ Fails with Node 24: `Error [ERR_METHOD_NOT_IMPLEMENTED]: The resolveSync() method is not implemented`

### ✅ `node --require tsx/cjs` (current solution)
- ✅ Works with Node 22
- ✅ Works with Node 24
- ✅ Maintains CommonJS behavior (`require.main === module` works correctly)

The `tsx/cjs` subpath is available in tsx v4.20.3 which is currently used in the project. For environments where this might not be available, the PR could be updated to check for its existence first, but given the pinned dependency, this should work reliably.

Tested successfully with:
- Node v22.14.0
- Node v24.4.1 (after running `npm rebuild better-sqlite3`)

PR #5124, opened 2025-07-31
fix: add universal plugin config truncation to prevent database bloat

MrFlounder:
This PR introduces universal truncation of plugin configurations when storing evaluation results to prevent database bloat from large config fields.

### Problem
Previously, the intent plugin manually set `pluginConfig: undefined` to avoid storing large configuration data in the database. However, this was a plugin-specific band-aid solution, and other plugins could still cause database bloat with large config fields (policy descriptions, grader examples, etc.).

### Solution
- **Universal Truncation**: Added `truncatePluginConfig()` utility that recursively truncates string values to 100 characters + "..." 
- **Plugin Factory Integration**: Modified `createPluginFactory()` to apply truncation to `metadata.pluginConfig` for all plugins
- **Safe Implementation**: Truncation only affects database storage metadata, not plugin execution logic
- **Preserves Overrides**: Respects plugins that explicitly set their own `pluginConfig`

### Changes
- `promptfoo/src/redteam/plugins/index.ts`: Added truncation utility and factory integration
- `promptfoo/src/redteam/plugins/intent.ts`: Removed hardcoded `pluginConfig: undefined`

### Benefits
- ✅ **Universal**: Automatically applies to all current and future plugins
- ✅ **Database Friendly**: Prevents config bloat while preserving essential debugging info
- ✅ **Zero Impact**: Plugin functionality remains unchanged
- ✅ **Consistent**: Same truncation behavior across all plugins

### Testing
Verified that plugin execution uses full config while database storage gets truncated versions.

### DIFF
--- a/src/redteam/plugins/index.ts
+++ b/src/redteam/plugins/index.ts
@@ -4,7 +4,9 @@ import { getEnvBool } from '../../envars';
 import { getUserEmail } from '../../globalConfig/accounts';
 import logger from '../../logger';
 import { REQUEST_TIMEOUT_MS } from '../../providers/shared';
+import type { ApiProvider, PluginActionParams, PluginConfig, TestCase } from '../../types';
 import invariant from '../../util/invariant';
+import type { HarmPlugin } from '../constants';
 import {
   BIAS_PLUGINS,
   PII_PLUGINS,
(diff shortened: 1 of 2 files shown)


mldangelo (commented):
Let's talk about this before you merge it

src/redteam/plugins/index.ts
```diff
@@ -119,16 +147,33 @@ function createPluginFactory<T extends PluginConfig>(
...
         logger.debug(`Using local redteam generation for ${key}`);
-        return new PluginClass(provider, purpose, injectVar, config as T).generateTests(n, delayMs);
+        const plugin = new PluginClass(provider, purpose, injectVar, config as T);
+        const testCases = await plugin.generateTests(n, delayMs);
+        return testCases.map((testCase) => ({
+          ...testCase,
+          metadata: {
+            ...testCase.metadata,
+            pluginId: getShortPluginId(key),
+            // Only add truncated pluginConfig if not already explicitly set
```
sklein12, line 160: In what cases does this happen?

MrFlounder: Most plugins don't set this so the logic gets executed here

PR #5129, opened 2025-07-31
chore(internals): Explicitly include the `basic` strategy in redteam generations

will-holley:
## Motivation

Currently, unless the basic strategy is [explicitly disabled](https://github.com/promptfoo/promptfoo/blob/74e0eb8af9d29676579b600eb8bd1198bbab4a6f/src/redteam/index.ts#L394), redteam generation will include the `basic` strategy in test case generation _without_ demarcating the generated test cases as such i.e. it is implied that these cases use the basic strategy given their having a defined `pluginId` and undefined `strategyId`.  This has downstream effects where checking if an eval result is redteam or non-redteam, when `strategyId` is undefined, requires checking for the inclusion of a `pluginId`.

## Changes

- This PR marks generated test cases with `metadata.strategyId: basic`, similar to how we're labeling `metadata.pluginId`, and includes `basic` within `redteam.strategies`.
- Ensures that the basic strategy is included in the "using strategies", strategies count under test generation summary, and test generation report e.g.
  image

## Test Instructions

1. Run redteam generate on a `promptfooconfig.yaml` with the basic strategy either explicitly or implicitly enabled. 
2. Verify `basic` is included in `redteam.strategies` in the outputted `redteam.yaml`.
3. Verify `metadata.strategyId` is set to `basic` on N generated test cases (where N is the number of plugin tests)

### DIFF
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -485,8 +485,8 @@ export default class Eval {
         } else if (type === 'strategy' && operator === 'equals') {
           const sanitizedValue = value.replace(/'/g, "''");
           if (sanitizedValue === 'basic') {
-            // Basic is represented by NULL in the metadata.strategyId field
-            condition = `(json_extract(metadata, '$.strategyId') IS NULL OR json_extract(metadata, '$.strategyId') = '')`;
+            // Basic can be represented by NULL, '', or 'basic'
+            condition = `(json_extract(metadata, '$.strategyId') IS NULL OR json_extract(metadata, '$.strategyId') = '' OR json_extract(metadata, '$.strategyId') = 'basic')`;
           } else {
             // Strategy ID is stored in metadata.strategyId
             condition = `json_extract(metadata, '$.strategyId') = '${sanitizedValue}'`;


src/redteam/index.ts
```diff
@@ -885,8 +885,26 @@ export async function synthesize({
...
+    // Add strategyId to plugin test cases when basic strategy is enabled
+    basicTestCases = pluginTestCases.map((testCase) => ({
+      ...testCase,
+      metadata: {
+        ...testCase.metadata,
+        strategyId: 'basic',
+      },
+    }));
+    // Record basic strategy results
+    strategyResults['basic'] = {
```
mldangelo, line 921: this might mess up the total that is reported during generation

mldangelo (commented):
Good idea, check the generation report to see if adding it to strategyResults changes the number

src/redteam/index.ts
```diff
@@ -390,28 +390,27 @@ export function calculateTotalTests(
...
+    } else {
+      // Basic strategy is explicitly disabled, so we don't count it
+      effectiveStrategyCount = strategies.length - 1;
+    }
+  } else if (includeBasicTests) {
+    // Basic strategy is implicitly enabled
+    effectiveStrategyCount = strategies.length + 1;
+  } else {
+    effectiveStrategyCount = strategies.length;
   }
```
mldangelo: consider un-nesting the if else. I find this block confusing. 
```suggestion
  if (isBasicStrategyDefined && !includeBasicTests) {
    // Basic strategy is explicitly disabled, so we don't count it
    effectiveStrategyCount = strategies.length - 1;
  } else if (!isBasicStrategyDefined && includeBasicTests) {
    // Basic strategy is implicitly enabled
    effectiveStrategyCount = strategies.length + 1;
  } else {
    // Either basic is defined and enabled, or not defined and disabled
    effectiveStrategyCount = strategies.length;
  }
```

src/redteam/index.ts
```diff
@@ -536,6 +535,14 @@ export async function synthesize({
     totalTests,
   } = calculateTotalTests(plugins, strategies);
 
+  // If the basic strategy is enabled but not in the strategies list, add it
+  if (
```
mldangelo, line 536: why do we have to do this here and in src/redteam/commands/generate.ts ?

mldangelo (dismissed):
Confused about why we have to add basic in two places, but otherwise looks good. Thanks

src/redteam/index.ts
will-holley: This handles adding `basic` to the "Using Strategies" log while `generate.ts` ensures that `basic` is added to `redteam.strategies` in the outputted config.

will-holley:
**Note:** this requires a pre-requisite update in Cloud as it assumes the only valid value for representing the basic strategy is `NULL` (which has implications for sharing and downstream effects such as filtering).

will-holley:
This is _just_ a nice-to-have, requires an upstream update in Cloud (that is on the critical path for filtering logic), and is 3 weeks stale, so I'm deeming this unnecessary and closing it.

PR #5148, opened 2025-08-01
chore: improve generated constants handling to prevent accidental commits

mldangelo:
## Summary

This PR provides a alternative to #4876 for handling the PostHog key injection during builds.

## Problem

Building locally writes the PostHog key to `src/generated-constants.ts`, which is very easy to accidentally commit. The current workaround uses a `postbuild` script that runs `git checkout -- src/generated-constants.ts`, but this is fragile and can fail if the build is interrupted.

## Solution

- Generate constants to `src/generated/` directory (gitignored)
- Create a wrapper module (`src/constants/build.ts`) that imports from the generated file with fallback to environment variables
- Remove the fragile `postbuild` git checkout hack
- Update all imports and tests to use the new structure

## Benefits

1. **Source files are never modified** - No dirty working directory during builds
2. **Can switch branches/stash during builds** - No git conflicts
3. **No accidental commits** - Generated files are properly gitignored
4. **Works seamlessly** - Uses env vars in development, generated constants in production
5. **Industry standard approach** - Common pattern used by Angular, Next.js, and other enterprise projects

## Testing

- [x] All tests pass
- [x] Build completes successfully
- [x] Generated files are created in the correct location
- [x] PostHog key injection works with environment variables
- [x] Generated directory is properly gitignored

This approach aligns with TypeScript best practices for handling build-time constants without using bundlers.


mldangelo:
## CI Fix Required

The build is failing because the CI workflow is still looking for the old file location. The following change needs to be made to `.github/workflows/main.yml`:

```diff
-            if \! grep -q "exports\.POSTHOG_KEY = 'phc_" dist/src/generated-constants.js; then
-              echo "Error: PostHog key not properly replaced in generated-constants.js"
-              echo "Content of generated-constants.js:"
-              cat dist/src/generated-constants.js
+            if \! grep -q "exports\.POSTHOG_KEY = 'phc_" dist/src/generated/constants.js; then
+              echo "Error: PostHog key not properly replaced in generated/constants.js"
+              echo "Content of generated/constants.js:"
+              cat dist/src/generated/constants.js
```

This updates the CI to check for the generated constants in the new location (`dist/src/generated/constants.js` instead of `dist/src/generated-constants.js`).

Note: I cannot push this change directly as it requires the `workflow` scope for the GitHub token.

PR #5189, opened 2025-08-06
feat: respect prompt config override in all providers

tomquist:
#3239 introduced the option to customize the provider config in the prompt. However, that was only supported in a handful of providers. All other providers were using the config specified in the provider.

This updates all providers to merge the prompt config into the provider specific config before making an API call.


mldangelo:
This is an excellent change @tomquist! Thank you so much! Please feel free to reach out if you'd ever like to chat or would like some swag.

PR #5216, opened 2025-08-10
docs: remove event banner and clean up unused Layout wrapper

mldangelo:
This PR removes the DEFCON and Black Hat event banner from the top of the site and adds notices to the event pages indicating that the events have concluded.

Changes:
- Removed EventsBanner component and its styles
- Removed unnecessary Layout wrapper (was only used for EventsBanner)
- Added event conclusion notices to both DEFCON and Black Hat event pages
- Updated blog post with event conclusion note

The event pages remain accessible for historical reference with clear notices that the events have passed.


typpo:
Nothing in this PR is necessary

PR #5235, opened 2025-08-12
fix(providers): Validate that OpenAI response reasoning outputs have summary items

will-holley:
Currently, when an [output](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output) is of type Reasoning and its `summary` property is an empty array, the output is prepended with `Reasoning: []`.

### DIFF
--- a/src/providers/openai/responses.ts
+++ b/src/providers/openai/responses.ts
@@ -382,7 +382,7 @@ export class OpenAiResponsesProvider extends OpenAiGenericProvider {
           }
         } else if (item.type === 'tool_result') {
           result = JSON.stringify(item);
-        } else if (item.type === 'reasoning') {
+        } else if (item.type === 'reasoning' && item.summary && item.summary.length > 0) {
           // Handle reasoning output from deep research models
           if (result) {
             result += '\n';


src/providers/openai/responses.ts
```diff
@@ -382,7 +382,7 @@ export class OpenAiResponsesProvider extends OpenAiGenericProvider {
           }
         } else if (item.type === 'tool_result') {
           result = JSON.stringify(item);
-        } else if (item.type === 'reasoning') {
+        } else if (item.type === 'reasoning' && item.summary && item.summary.length > 0) {
```
mldangelo, line 385: should we be doing something with the summary? Might be nice to cover this with a test

will-holley: We're stringifying `item.summary` which isn't great: it unnecessarily persists the `type` key (which will always have a value of `summary_text`) and the `text` key; I changed this behavior in https://github.com/promptfoo/promptfoo/pull/5235/commits/6096217fb68ebcb54aae4ea20c5ad8e2f9641e3d to only persist the `text` values and added unit tests in https://github.com/promptfoo/promptfoo/pull/5235/commits/e621b27cf336cd39e7b37ab5b98f44e717acd771.

mldangelo (approved):

PR #5239, opened 2025-08-12
fix: token tracking

sklein12:
Fix token tracking
- Remove calls to other providers from target provider
- Add calls to other providers like the grader and red team provider to token tracker


test/util/redteamProbeLimit.test.ts
mldangelo, line 1: Did you mean to include this?

sklein12: woops, no, this is just a draft though didn't review it, wanted to show you what I was doing

sklein12:
@coderabbitai can i get a review

PR #5240, opened 2025-08-12
docs: Adds note to context transform docs regarding `transformResponse`

will-holley:
It's not clear from the current documentation that `transformResponse`, if defined on the provider, will run prior to `contextTransform` and the return value of the former will be the input value of the `output` argument in the latter.

### DIFF
--- a/site/docs/configuration/expected-outputs/model-graded/index.md
+++ b/site/docs/configuration/expected-outputs/model-graded/index.md
@@ -328,7 +328,7 @@ The `contextTransform` property accepts a stringified Javascript expression whic
 type ContextTransform = (output: Output, context: Context) => string;
 
 /**
- * The provider's response output.
+ * The provider's response output OR, if defined, the return value of a `transformResponse` function.
  */
 type Output = string | object;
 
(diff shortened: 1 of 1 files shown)


site/docs/configuration/expected-outputs/model-graded/index.md
```diff
@@ -380,6 +380,10 @@ assert:
     threshold: 0.7
 ```
 
+:::note
+When a `transformResponse` function is defined on your provider, it will run before the `contextTransform` function and `output` will be its return value.
```
mldangelo: Do you mean transform? transformResponse is only applicable to http providers and probably doesn't make sense here.

will-holley: This may also be true of `transform`; I've only tested with `transformResponse` defined on an HTTP provider.

mldangelo: transform is a universal concept and transformResponse is not. we should guide the docs towards transform over transformResponse

will-holley: Making a note that the function signatures of [`transform`](https://www.promptfoo.dev/docs/configuration/guide/#transforming-outputs) and [`contextTransform`](https://www.promptfoo.dev/docs/configuration/expected-outputs/model-graded/#dynamically-via-context-transform) are incompatible (`transform` is missing `context.metadata`) and should not be used together right now.

will-holley:
This change is included in https://github.com/promptfoo/promptfoo/pull/5376.

PR #5255, opened 2025-08-13
feat(webui): Demarcate redteam results

will-holley:
## Summary

Adds a new column to eval tables – "type" – that demarcates whether an eval is redteam.  If the user has not run any redteam scans, the column is not displayed.

image

## Motivation

Finding redteam results is very difficult and relies on the user recognizing something in the description that indicates a redteam result.

### DIFF
--- a/src/app/src/pages/evals/components/EvalsDataGrid.tsx
+++ b/src/app/src/pages/evals/components/EvalsDataGrid.tsx
@@ -2,10 +2,11 @@ import { forwardRef, useEffect, useMemo, useRef, useState } from 'react';
 
 import { callApi } from '@app/utils/api';
 import Box from '@mui/material/Box';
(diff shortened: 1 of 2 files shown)


will-holley:
**Note:** I'd also be in favor of separating redteam results and eval results into their own data grids and using tabs to switch between the two. I often find myself looking for a redteam that uses a specific plugin or strategy, which would require redteam-specific columns in the data grid.  Thoughts, @sklein12?

faizanminhas:
thought: if users heavily use one or the other, then should we only show that column if the user definitely has both? thinking about if we want to avoid clutter

will-holley:
> thought: if users heavily use one or the other, then should we only show that column if the user definitely has both? thinking about if we want to avoid clutter

That's included 😬

PR #5270, opened 2025-08-15
fix: lower sharing chunk size

sklein12:
A lot of users have been hitting ngnix max request size errors. The default max request size of ngnix is 1mb, so setting this just below that.

### DIFF
--- a/src/share.ts
+++ b/src/share.ts
@@ -178,7 +178,7 @@ async function sendChunkedResults(evalRecord: Eval, url: string): Promise<string
   logger.debug(`Largest result size from sample: ${largestSize} bytes`);
 
   // Determine how many results per chunk
-  const TARGET_CHUNK_SIZE = 10 * 1024 * 1024; // 10MB in bytes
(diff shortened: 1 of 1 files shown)


src/share.ts
```diff
@@ -178,7 +178,7 @@ async function sendChunkedResults(evalRecord: Eval, url: string): Promise<string
   logger.debug(`Largest result size from sample: ${largestSize} bytes`);
 
   // Determine how many results per chunk
-  const TARGET_CHUNK_SIZE = 10 * 1024 * 1024; // 10MB in bytes
+  const TARGET_CHUNK_SIZE = 0.9 * 1024 * 1024; // 900KB in bytes
```
faizanminhas, line 181: Do we already warn the users to set the chunk size var if it fails? I feel like I prefer to keep the default high since most users presumably don’t run into this issue and sharing is faster?

typpo: In my recent anecdotal experience sharing frequently fails, so I'd much rather wait longer

faizanminhas: Fair enough, just find it odd that this seems to be a recent thing

PR #5275, opened 2025-08-16
fix: composite indices and query optimization

typpo:
my evals were taking 2+ min to load, now they load in <200 ms


mldangelo (commented):
I feel a deep sense of shame for not doing this earlier

PR #5285, opened 2025-08-18
docs: model graded metrics updates

ladyofcode:
Attempted to improve the metrics definitions to be easier to understand.


site/docs/write-for-promptfoo.md
mldangelo, line 1: Can you please revert this change and the change to sidebar.js?

ladyofcode: Done!

PR #5286, opened 2025-08-18
feat: Allow remote Turso as a data provider

hoxworth:
Maintaining this in line with promptfoo's main branch is going to get annoying. I don't think we can get this merged into their main - the async / sync types result in a lot of weird `any` returns and that feels icky in typescript.

I've got another iteration going that tries to unify on drizzle as the single ORM layer, which might be a better approach to get something merged to promptfoo - but this is a start.

### DIFF
--- a/drizzle.config.ts
+++ b/drizzle.config.ts
@@ -1,11 +1,37 @@
 import { defineConfig } from 'drizzle-kit';
 import { getDbPath } from './src/database';
 
+// Determine database URL based on environment variables
+const getDatabaseUrl = () => {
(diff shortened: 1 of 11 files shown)


mldangelo:
Thank you @hoxworth! Can you make this editable? I will try to get some version of this in this week.

hoxworth:
@mldangelo LOL I mistakenly opened this against the main repo, intended to be a test on our fork. I'll leave open and move to a draft PR and open it up, but I've got a better approach coming...

PR #5293, opened 2025-08-19
feat: add azure:responses provider alias for Azure Responses API

mldangelo:
## Summary

This PR adds support for the `azure:responses:` provider alias, making it more intuitive for users to use the Azure OpenAI Responses API with promptfoo.

## Problem

Daniel Furman reported in issue PF-566 that users need to use `openai:responses` with Azure configuration, which is not intuitive. The Azure Responses API documentation shows it's a fully supported Azure service.

## Solution

- Created `AzureResponsesProvider` that extends `AzureGenericProvider` and delegates to `OpenAiResponsesProvider`
- Added `azure:responses:` provider type registration in the registry
- Updated documentation to show both methods (new `azure:responses` alias and legacy `openai:responses` with Azure config)
- Ensures proper Azure authentication support (API keys and Azure AD)

## Usage

Users can now use the more intuitive Azure-specific alias:

```yaml
providers:
  # New recommended method
  - id: azure:responses:gpt-4.1-deployment
    config:
      temperature: 0.7
      instructions: 'You are a helpful assistant.'
      response_format: file://./schema.json
      
  # Legacy method still supported
  - id: openai:responses:gpt-4.1
    config:
      apiHost: 'instance.openai.azure.com'
      apiKey: '${AZURE_API_KEY}'
```

## Testing

- Verified the provider is recognized and attempts API calls
- External file loading for `response_format` works correctly
- All Azure Responses API models are supported

Fixes PF-566

### DIFF
--- a/site/docs/providers/azure.md
+++ b/site/docs/providers/azure.md
@@ -68,8 +68,235 @@ providers:
 - `azure:chat:<deployment name>` - For chat endpoints (e.g., gpt-4o, gpt-4o-mini, gpt-4.1, gpt-4.1-mini, gpt-4.1-nano)
 - `azure:completion:<deployment name>` - For completion endpoints (e.g., gpt-35-turbo-instruct)
 - `azure:embedding:<deployment name>` - For embedding models (e.g., text-embedding-3-small, text-embedding-3-large)
+- `azure:responses:<deployment name>` - For the Responses API (e.g., gpt-4.1, gpt-5, o3-mini)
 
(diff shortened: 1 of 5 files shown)


brandonthomas:
Would love for this to go in soon! I'm hitting this and need to be able to utilize local MCPs with the responses API. @mldangelo, can you elaborate on the work around mentioned in the description? I tried guessing a bit about how to get the openai aliases to work with an Azure OpenAI config (I was using env variables but maybe that's not right) but couldn't get it working.

PR #5303, opened 2025-08-19
feat: add new flag prefer local redteam provider

MrFlounder:


src/redteam/providers/crescendo/index.ts
```diff
@@ -172,15 +177,15 @@ export class CrescendoProvider implements ApiProvider {
 
   private async getScoringProvider(): Promise<ApiProvider> {
     if (!this.scoringProvider) {
-      if (shouldGenerateRemote()) {
+      if (shouldGenerateRemote() && !this.preferLocalRedteamProvider) {
```
faizanminhas: Just wondering how this is different than the flag for not doing remote generation?

MrFlounder: that flag disable all calls to cloud including generation 
I'm also not sure this is the best solution, but I need a way to override the provider in these three strategies

faizanminhas: How about adding a redteam provider per strategy config? Feels more declarative and less flag hell

MrFlounder: You are so smart!

PR #5315, opened 2025-08-20
fix(internals): Pass `vars.output` and `vars.rubric` to LLM rubric grading call

will-holley:
Improves support for custom rubrics in overridden grading providers e.g.

```yaml
defaultTest:
  options:
    provider:
      id: http
      config:
        url: https://api.openai.com/v1/chat/completions
        method: POST
        headers:
          Authorization: "Bearer {{env.OPENAI_API_KEY}}" 
          Content-Type: "application/json"
        body:
          model: 'gpt-4o'
          messages: [
              {
                "role": "system", 
                "content": "You are grading output according to a user-specified rubric. If the statement in the rubric is true, \
                then the output passes the test. You respond with a JSON object with this structure: {reason: string, pass: boolean, score: number}"
              },
              {
                "role": "user", 
                "content": "<Output>\n{{output}}\n</Output>\n<Rubric>\n{{rubric}}\n</Rubric>"
              }
            ]
        transformResponse: "json.choices[0].message.content"
        rejectUnauthorized: false
```

presently does not work as `output` and `rubric` are not passed into the template rendering call.

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -30,6 +30,7 @@ import { isJavascriptFile } from './util/fileExtensions';
 import invariant from './util/invariant';
 import { extractFirstJsonObject, extractJsonObjects } from './util/json';
 import { getNunjucksEngine } from './util/templates';
+import { accumulateTokenUsage } from './util/tokenUsageUtils';
 
 import type {
   ApiClassificationProvider,
@@ -45,7 +46,6 @@ import type {
   ProviderTypeMap,
   TokenUsage,
 } from './types';
-import { accumulateTokenUsage } from './util/tokenUsageUtils';
 
 class LlmRubricProviderError extends Error {
   constructor(message: string) {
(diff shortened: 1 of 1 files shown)


src/matchers.ts
```diff
@@ -500,7 +500,17 @@ export async function matchesLlmRubric(
...
+    prompt: {
+      raw: prompt,
+      label: 'llm-rubric',
+    },
+    vars: {
+      output: tryParse(llmOutput),
+      rubric,
+      ...(vars || {}),
+    },
+  });
```
faizanminhas, line 513: Nice, wondering how this doesn't affect our grading provider llm-rubric as well? Can we consolidate if this is being applied otherwise?

will-holley: The variables are injected into the rubric template on [line 488](https://github.com/promptfoo/promptfoo/pull/5315/files/8de02c6207f2da618b9e7076991eaf0753fcd927#diff-ba050ec0aef94e7c13a600f1aaaf3747c9624543ce666b25035dda59a6faf0c7R488).  

They could theoretically be consolidated by forcing the user to [define `rubricPrompt`](https://www.promptfoo.dev/docs/configuration/expected-outputs/model-graded/llm-rubric/#customizing-the-rubric-prompt) and then passing the rendered prompt into the request body definition but that's pretty opaque & IMO harder to reason about.

faizanminhas:
discussed offline- If the user has an openai compatible api that they must use, then they should just use the openai provider directly as the grading provider, setting the env var to override the base url, and then set `rubricPrompt` separately

PR #5338, opened 2025-08-21
feat(redteam): add -t/--target option to redteam generate command

will-holley:
## Summary
- Add support for the `-t, --target <id>` option to the `promptfoo redteam generate` command
- This allows users to specify a cloud provider target ID when generating red team tests
- The implementation mirrors the existing `redteam run` command's handling of cloud targets

## Changes
- Added `target`, `liveRedteamConfig`, and `loadedFromCloud` fields to `RedteamCliGenerateOptions` interface
- Updated validation schema to include the new fields
- Added command-line option `-t, --target <id>` to both `promptfoo redteam generate` and `promptfoo generate redteam` commands
- Implemented cloud config fetching logic in the command handler
- Added handling for `liveRedteamConfig` in `doGenerateRedteam` to write cloud configs to temp files

## Usage
```bash
promptfoo redteam generate -c <cloud-config-uuid> -t <target-uuid>
```

## Test plan
- [x] Linting and formatting checks pass
- [x] Option appears correctly in help text for both command variants
- [ ] Manual testing with cloud config and target IDs
- [ ] Existing tests continue to pass

### DIFF
--- a/src/redteam/commands/generate.ts
+++ b/src/redteam/commands/generate.ts
@@ -5,11 +5,12 @@ import path from 'path';
 import chalk from 'chalk';
 import dedent from 'dedent';
 import yaml from 'js-yaml';
+import { validate as uuidValidate } from 'uuid';
 import { z } from 'zod';
 import { fromError } from 'zod-validation-error';
 import { disableCache } from '../../cache';
 import cliState from '../../cliState';
-import { VERSION } from '../../constants';
+import { CLOUD_PROVIDER_PREFIX, VERSION } from '../../constants';
 import { getAuthor, getUserEmail } from '../../globalConfig/accounts';
 import { cloudConfig } from '../../globalConfig/cloud';
 import logger from '../../logger';
@@ -19,6 +20,7 @@ import telemetry from '../../telemetry';
 import { isRunningUnderNpx, printBorder, setupEnv } from '../../util';
 import {
   getCloudDatabaseId,
+  getConfigFromCloud,
   getPluginSeverityOverridesFromCloud,
   isCloudProvider,
 } from '../../util/cloud';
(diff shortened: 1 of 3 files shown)


typpo:
nice

src/redteam/commands/generate.ts
```diff
@@ -601,7 +625,36 @@ export function redteamGenerateCommand(
     .option('--force', 'Force generation even if no changes are detected', false)
     .option('--no-progress-bar', 'Do not show progress bar')
     .option('--burp-escape-json', 'Escape quotes in Burp payloads', false)
-    .action((opts: Partial<RedteamCliGenerateOptions>): void => {
+    .action(async (opts: Partial<RedteamCliGenerateOptions>): Promise<void> => {
+      setupEnv(opts.envFile);
+
```
mldangelo: this happens automatically

src/redteam/commands/generate.ts
```diff
@@ -601,7 +625,36 @@ export function redteamGenerateCommand(
...
+      // Handle cloud config with target
+      if (opts.config && uuidValidate(opts.config)) {
+        if (opts.target && !uuidValidate(opts.target)) {
```
mldangelo: consider unnesting this block. You could move the other validations up too and IMHO it would be cleaner

src/validators/redteam.ts
```diff
@@ -175,6 +176,8 @@ export const RedteamGenerateOptionsSchema = z.object({
   write: z.boolean().describe('Whether to write the output'),
   burpEscapeJson: z.boolean().describe('Whether to escape quotes in Burp payloads').optional(),
   progressBar: z.boolean().describe('Whether to show a progress bar').optional(),
+  liveRedteamConfig: z.any().optional().describe('Live redteam configuration from cloud'),
+  loadedFromCloud: z.boolean().optional().describe('Whether the config was loaded from cloud'),
```
mldangelo: where and how are these used?

mldangelo (commented):
Looks good, didn't test it.

src/redteam/commands/generate.ts
will-holley: This wasn't clear, but the nesting is good because passing a target is mutually inclusive with a cloud-based config.  I've updated the comment to better reflect this.

src/validators/redteam.ts
will-holley: I've gone ahead and consolidated these into a single variable, `configFromCloud`, which the remote configuration is stored in.  It's used to write the config the filesystem à la https://github.com/promptfoo/promptfoo/blob/feaa4a8f9ab6f529bd0dcdcadc89ba968a34f00f/src/redteam/shared.ts#L58

mldangelo (approved):

PR #5184, opened 2025-08-06
feat: add support for OpenAI gpt-oss models

mldangelo:
## Summary
Add comprehensive support for OpenAI's new open-weight models gpt-oss-120b and gpt-oss-20b.

## Changes
- ✅ Add gpt-oss-120b and gpt-oss-20b to OpenAI chat models list
- ✅ Add both models to OpenAI Responses API provider  
- ✅ Recognize gpt-oss models as reasoning models with configurable effort levels
- ✅ Enable temperature support for gpt-oss models (unlike o1/o3)
- ✅ Support max_completion_tokens for reasoning capabilities
- ✅ Add comprehensive test coverage (15 tests, all passing)

## Key Features
- **Reasoning effort levels**: low, medium, high
- **Temperature control**: Supported (unlike o1/o3 models)
- **API compatibility**: Works with both Chat Completions and Responses APIs
- **Function calling**: Full tool use capabilities
- **Context window**: 128k tokens supported

## Testing
- Created comprehensive test suite with 15 tests covering all functionality
- Verified end-to-end with mock API server
- All tests passing
- Code passes linting and formatting checks

## Usage Example
```yaml
providers:
  # Chat completion with reasoning
  - id: openai:chat:gpt-oss-120b
    config:
      reasoning_effort: high
      temperature: 0.7
      max_completion_tokens: 4096
      
  # Responses API
  - id: openai:responses:gpt-oss-20b
    config:
      reasoning_effort: medium
      temperature: 0.5
```

## Background
OpenAI announced gpt-oss models as their first open-weight language models since GPT-2. These models:
- Support configurable reasoning effort (similar to o3-mini)
- Run on consumer hardware (80GB for 120b, 16GB for 20b)
- Apache 2.0 licensed
- Outperform o3-mini on many benchmarks

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #5352, opened 2025-08-22
fix: Add gpt-oss to openai reasoning model list

timothyasp:
gpt-oss is a reasoning model, so when testing with a provider like groq, we can't parse or hide reasoning from the output for evaluation. 

This fixes that.


will-holley:
Hi @timothyasp, thanks for the PR. What's your current testing setting for gpt-oss w/ Groq?

timothyasp:
Looks like https://github.com/promptfoo/promptfoo/pull/5184 should fix this - closing.

PR #5365, opened 2025-08-24
docs(site): add guidance on crafting a good purpose

mldangelo:


typpo:
I think we need precautions in the runtime and in the UI.  This warning is a bit over the top and TMI - and anyone actually reading the documentation is unlikely to make this mistake in the first place.

will-holley (approved):

will-holley:
FYI https://linear.app/promptfooo/issue/PF-1486/bad-purpose-handling

PR #5379, opened 2025-08-26
feat: MCP Agent example to red team with tool call results

faizanminhas:
Updates and copies over the agent example from github.com/promptfoo/mcp-agent-provider/. This is a ready to use agent that is based on the OpenAI api models that will call tools from mcp servers and give out the results. Users can configure MCP servers (local, remote, file based, etc), and the model & system prompt for the agent.


sklein12 (approved):
cool

PR #5389, opened 2025-08-27
chore(internals): Improves support for defining LLM-Rubric assertion threshold in CSV test cases

will-holley:
### Summary

When defining test cases via a CSV, it is possible to set a threshold value which applies to all asserts but impossible to configure individual assert threshold values.

### Motivation

https://linear.app/promptfooo/issue/ENG-130/add-support-for-setting-threshold-on-exclusively-on-llm-rubric

### Test Instructions

Run the CSV Metadata example e.g.

```sh
pnpm local eval -c examples/csv-metadata
```

### DIFF
--- a/test/csv.test.ts
+++ b/test/csv.test.ts
@@ -213,6 +213,103 @@ describe('testCaseFromCsvRow', () => {
     const result = testCaseFromCsvRow(row);
     expect(result).toEqual(expectedTestCase);
   });
+
+  describe('__config columns', () => {
+    it('applies threshold to single assertion using __config:__expected:threshold', () => {
+      const row: CsvRow = {
+        __expected1: 'similar:foo',
+        '__config:__expected:threshold': '0.9',
+      };
+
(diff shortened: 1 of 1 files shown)


src/csv.ts
```diff
@@ -207,13 +210,85 @@ export function testCaseFromCsvRow(row: CsvRow): TestCase {
       logger.warn(
         'The "__metadata" column requires a key, e.g. "__metadata:category". This column will be ignored.',
       );
+    } else if (key.startsWith('__config:')) {
```
addelong, line 213: Could we add some additional unit tests to cvs.test.ts for this new code?

addelong (changes requested):

src/csv.ts
will-holley: Added in https://github.com/promptfoo/promptfoo/pull/5389/commits/06a273092e74300994467901da58c9b8c4feeb16.

addelong (approved):

PR #5429, opened 2025-08-30
feat(webui): Persist column visibility settings across evaluations

### LINKED ISSUE #3570: Column selections not saving in Web View
# Issue
When rerunning evals in promptfoo, the webview automatically updates but hidden column settings are not retained between runs. So when using a large number of variables, I have to re-hide columns that I don't want to display every time I run a new eval. This requires a lot of clicking each time. This was mentioned in below thread, but can't find any other related threads of this issue being reported. 

# Related Background

              @DarioBernardo unfortunately this PR was closed without merging it. Hiding columns and remembering state is on our backlog. We will get to it soon.

_Originally posted by @mldangelo in https://github.com/promptfoo/promptfoo/issues/1069#issuecomment-2383552426_

typpo:
Fixes #3570

Remembers the last used column visibility settings and applies them to new evaluations.


will-holley:
Making a note here that this should be an opt-in toggle: either column visibility is persisted across re-runs _or_ evals; the UX of navigating to an eval and columns are by default hidden is jarring because it's not obvious that columns are hidden and something isn't broken.

mldangelo-oai (commented):
Reviewed exact head `423d52bdb6db620e018259d97e4a79b211f6cf72` in comment-only mode. Recommendation: retire this draft as superseded by merged PR #7536 rather than revive the branch. #7536 shipped the same user outcome on February 10: schema-based column visibility persistence across matching evals, per-eval non-variable state, hidden-column affordances, and the `Columns (visible/total)` display. Current `main` also constructs `selectedColumns` by iterating the current eval's `allColumns`, so stale columns cannot reach the sole production `ColumnSelector` call site.

Against current `main`, this 510-commits-behind branch has only three remaining files: a defensive nine-line count filter, one component test that passes a state production no longer produces, and three store tests that restate behavior already shipped by #7536. Those remnants do not justify merging the old feature branch. If the component-level defensive count is still desired, extract just that small change and its test into a fresh PR from current `main`; do not carry the historical branch forward. The two Copilot threads are outdated and non-blocking.

Verification: GitHub reports a clean merge and 41/41 published checks passing; `git diff --check` passes; a synthetic current-main merge is clean; and the two affected test files pass 26/26 focused tests. No branch edits or pushes were made. Superseding implementation: https://github.com/promptfoo/promptfoo/pull/7536

PR #5477, opened 2025-09-05
feat(e2b-example): add e2b-code-eval example (promptfoo + e2b sandbox)

Ayush7614:
### Summary

This PR adds a self-contained example demonstrating a secure LLM → code → sandbox evaluation pipeline using promptfoo + e2b. The example illustrates how to generate Python functions using an LLM, execute the generated code within e2b sandboxes, verify outputs, record structured metrics, and produce a human-readable report.

This is intended as a reproducible example users can run locally to evaluate model code generation or to start SWE-bench style experiments.

### Motivation

- Demonstrates a secure and practical evaluation workflow for LLM-generated code.
- Uses e2b sandboxes to reduce host risk; pipeline includes static-safety checks, sandbox execution, and structured metric collection so we can compare models and prompts reliably.
- Makes it easy to iterate: tighten prompt, add tests, scale dataset, or run model matrices.

### DIFF
--- a/examples/e2b-code-eval/README.md
+++ b/examples/e2b-code-eval/README.md
@@ -1,0 +1,65 @@
+# e2b Code Eval — step-by-step sandboxed code-eval example
+
+## What This Example Demonstrates
+
+This example shows a complete prompt→LLM→sandboxed-execution→metric pipeline using:
+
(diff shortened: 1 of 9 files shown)


Ayush7614:
cc: @mldangelo @typpo

typpo (approved):
nice!

PR #5479, opened 2025-09-05
fix: latency assertion description

v-dev:
Update description for the latency assertion. The assertion should fail if the LLM call takes longer than the specified threshold.


mldangelo (approved):
Thank you!

PR #5482, opened 2025-09-05
feat: Add test button for request/response transforms

faizanminhas:
https://github.com/user-attachments/assets/d5eb4f8c-8520-417a-b251-12e7d8588d6d

https://github.com/user-attachments/assets/1db9379a-7ba9-442a-9638-9a07b35f2bbc

### DIFF
--- a/src/server/routes/providers.ts
+++ b/src/server/routes/providers.ts
@@ -5,6 +5,7 @@ import { fromZodError } from 'zod-validation-error';
 import { getEnvString } from '../../envars';
 import logger from '../../logger';
 import { loadApiProvider } from '../../providers';
+import { createTransformRequest, createTransformResponse } from '../../providers/http';
 import {
   doTargetPurposeDiscovery,
   type TargetPurposeDiscoveryResult,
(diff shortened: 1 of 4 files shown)


will-holley:
I like the idea here but wish that it wasn't hidden behind the "Test" button.  I think we should put the "Test Target" button above the Response Parser form field and more-forcefully nudge the user to click it. Then we should take the real response, and generate function signature that matches the schema of the returned data e.g. if the response looks like `{ response: { data: 'Hello World', citations: ['foo', 'bar'] }` we would generate ` { response: { data: string, citations: string[] } } ` and then display this in the response parser description.

src/server/routes/providers.ts
```diff
@@ -250,6 +251,119 @@ providersRouter.post('/http-generator', async (req: Request, res: Response): Pro
...
 });
 
+// Test request transform endpoint
+providersRouter.post(
+  '/test-request-transform',
+  async (req: Request, res: Response): Promise<void> => {
+    try {
+      const { transformCode, prompt } = req.body;
+
+      if (!transformCode) {
```
will-holley: I'd do these with Zod instead.

src/server/routes/providers.ts
```diff
@@ -250,6 +251,119 @@ providersRouter.post('/http-generator', async (req: Request, res: Response): Pro
...
+      const result = await transformFn(
+        prompt,
+        {},
+        { prompt: { raw: prompt, label: prompt }, vars: {} },
+      );
+
+      // Check if result is completely empty (no value at all)
+      if (result === null || result === undefined) {
+        res.json({
+          success: false,
```
will-holley, line 202: Why not infer success from the http status code?

src/server/routes/providers.ts
```diff
@@ -250,6 +251,119 @@ providersRouter.post('/http-generator', async (req: Request, res: Response): Pro
...
+      // Return the result even if it's an empty string or other falsy value
+      // as it might be intentional
+      res.json({
+        success: true,
+        result,
+      });
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      logger.error(`[POST /providers/test-request-transform] Error: ${errorMessage}`);
+      res.status(200).json({
```
will-holley, line 225: It's odd to me that this is a 200 but success is false.

src/server/routes/providers.ts
```diff
@@ -250,6 +251,119 @@ providersRouter.post('/http-generator', async (req: Request, res: Response): Pro
...
+            'Transform returned empty result. Check your extraction path or transform function.',
+          result: output,
+        });
+        return;
+      }
+
+      // If output is empty but raw has content, still return it as success
+      // This handles cases where the result isn't a string but is still valid
+      res.json({
+        success: true,
```
will-holley, line 274: Same comments as above re: `success` and http status codes

src/app/src/pages/redteam/setup/components/Targets/tabs/RequestTransformTab.tsx
```diff
@@ -43,31 +105,74 @@ const RequestTransformTab: React.FC<RequestTransformTabProps> = ({
...
+        onTransformCodeChange={setEditableTransform}
+        testInput={testInput}
+        onTestInputChange={setTestInput}
+        testInputLabel="Test Prompt"
+        testInputPlaceholder="Enter a test prompt..."
+        testInputRows={3}
+        onTest={handleTest}
+        onApply={(code) => updateCustomTarget('transformRequest', code)}
+        functionDocumentation={{
+          signature: '(prompt, vars, context) => any',
```
will-holley: Image

`any` is wrong here

src/app/src/pages/redteam/setup/components/Targets/tabs/RequestTransformTab.tsx
```diff
@@ -43,31 +105,74 @@ const RequestTransformTab: React.FC<RequestTransformTabProps> = ({
...
+      <Box sx={{ position: 'relative' }}>
+        <Box
+          sx={{
+            border: 1,
+            borderColor: 'grey.300',
+            borderRadius: 1,
+            backgroundColor: darkMode ? '#1e1e1e' : '#fff',
+          }}
+        >
+          <Editor
```
will-holley, line 117: Image

needs breaking

will-holley (changes requested):
see comments

src/server/routes/providers.ts
faizanminhas: thought about it and I think the status codes are fine- this isn't a rest-ful endpoint, and 200 indicates that the test ran successfully, but the connectivity test failed, which seems consistent to me

Issue #5501, opened 2025-09-07
Some assertions in Gemini evaluation fail with system-only prompts


pokutuna:
**Describe the bug**

When using Google Gemini Test providers (both Google AI Studio and Vertex AI), specific evaluation assertions fail with a 400 Bad Request error.

The affected assertions are those that use LLM grading with system-only prompts: `model-graded-closedqa` and `select-best`. These assertions fail because Gemini API does not support executing prompts that contain only system messages.

**To Reproduce**

1. Create a configuration file with a Gemini provider and the affected assertion types:

```yaml
# file: gemini-reproduce.yaml
providers:
- id: vertex:gemini-2.0-flash
- id: vertex:gemini-2.5-flash-lite

defaultTest:
  options:
    provider:
      id: vertex:gemini-2.5-pro # this issue also happens in google:* providers

prompts:
- Please explain the good points about dogs.

tests:
- assert:
  - type: select-best
    value: Explains multiple positive aspects of dogs as companions and pets.
  - type: model-graded-closedqa
    value: Focuses on dogs without discussing cats.
```

2. Run the evaluation:
```bash
promptfoo eval -c gemini-reproduce.yaml
```

3. Observe the 400 Bad Request error:
```
API call error: Status INVALID_ARGUMENT, Code 400, Message:
Request contains an invalid argument.
```

**Expected behavior**

The evaluation assertions should execute successfully and return PASS/FAIL results, like they do with other providers (OpenAI, Anthropic, etc.).

**Screenshots**

<img width="1917" height="754" alt="Image" src="https://github.com/user-attachments/assets/898919b1-2bf1-45a6-8115-1655e0b076c5" />

**System information:**

- Promptfoo version: 0.118.3
- Providers: Both `vertex:*` and `google:*` Gemini providers

**Additional context**

Affected Assertions are `model-graded-closedqa` and `select-best`, those prompts are system-only prompts.

I checked almost all LLM Assertions, and it appears these two are the only implementations with system prompts only.

This issue originates from Gemini models, so it occurs with both `vertex:*` and `google:*` providers.

**Proposed Solution**

When system-only prompts are detected, the Gemini provider should convert them to user messages for API compatibility.

This fix may also resolve this issue
- https://github.com/promptfoo/promptfoo/issues/1393

PR #5502, opened 2025-09-07
fix(providers): handle system-only prompt in Gemini

pokutuna:
resolves: #5501

I confirmed that the following config, which previously did not work, now works with this fix.
https://gist.github.com/pokutuna/071b4812933e79022a058d1af5de6320

- `$ GEMINI_API_KEY=*** npm run local -- eval -c ./gemini-systemonly-reproduce-google.yaml`
- `$ npm run local -- eval -c ./gemini-systemonly-reproduce-vertex.yaml` (with gcloud credentials)


mldangelo:
Thank you @pokutuna! Please send me an email if you'd like some promptfoo swag!

PR #5504, opened 2025-09-07
feat: Logs all requests and persists debug logs

sklein12:
- Log all requests
- Logs all debug logs to `{config_dir}/logs`
- Rotate to only the last 50 logs 
- Add command to export all the logs `promptfoo export logs`


src/util/config/writer.ts
```diff
@@ -0,0 +1,36 @@
+import fs from 'fs';
```
sklein12, line 1: need to move to deal with circular dependencies

src/util/fetch/index.ts
```diff
@@ -5,19 +5,37 @@ import type { ConnectionOptions } from 'tls';
...
 import invariant from '../../util/invariant';
 import { sleep } from '../../util/time';
+import { sanitizeUrl } from '../sanitizer';
+
+function isConnectionError(error: Error) {
+  return (
+    error instanceof TypeError &&
+    error.message === 'fetch failed' &&
+    // @ts-expect-error undici error cause
+    error.cause?.stack?.includes('internalConnectMultiple')
```
sklein12: not available on all `TypeError` types

PR #5509, opened 2025-09-08
feat: Claude Agent SDK provider support

danenania:
Adds Claude Code provider via TypeScript SDK  (https://docs.anthropic.com/en/docs/claude-code/sdk/sdk-typescript) under `anthropic:claude-code` id.

- Uses `ANTHROPIC_API_KEY`, or Vertex/Bedrock with `CLAUDE_CODE_USE_BEDROCK` or `CLAUDE_CODE_USE_VERTEX` and appropriate credentials
- Runs without tools in a temp dir by default
- Can be supplied a `working_dir`, in which case read-only tools are allowed by default
- Additional tools can be configured via `append_allowed_tools`, `custom_allowed_tools`, or `allow_all_tools`
- Most sdk params can be passed through config
- Environment passed through as-is
- Supports MCP servers
- Supports caching (mtime-based fingerprint is added to cache key when `working_dir` supplied)
- Includes examples for basic usage, custom working dir with files seeded, advanced usage with side effects, and mcp server
- Tiny refactor: moved the `getAuthHeaders` private method from MCPClient into a separate `util.ts` file so it can be exported and reused for the Claude Code MCP config—also moved corresponding tests

### DIFF
--- a/examples/redteam-websocket/README.md
+++ b/examples/redteam-websocket/README.md
@@ -1,0 +1,126 @@
+# redteam-websocket (WebSocket red teaming)
+
+This example shows how to red team a WebSocket-based application. It includes a small Node.js server that exposes two WebSocket endpoints:
+
+- A non-streaming endpoint (`/ws`) that returns a single message when the model finishes.
+- A streaming endpoint (`/ws-stream`) that sends incremental deltas and a final message.
+
+You’ll run the server locally and use promptfoo’s red team flow to probe both endpoints with adversarial test cases.
+
+You can run this example with:
+
+```bash
+npx promptfoo@latest init --example redteam-websocket
+```
+
+## What’s in this folder
+
+- `promptfooconfig.yaml` – Configures two targets pointing at the local WebSocket server:
+  - `ws://localhost:3300/ws` (non-streaming), uses `transformResponse` to parse a single reply
+  - `ws://localhost:3300/ws-stream` (streaming), uses `streamResponse` to accumulate/return the final reply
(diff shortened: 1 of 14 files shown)


danenania:
@mldangelo refactored the examples into a single directory, and also tried to make the example readme more consistent with other examples in the repo.

All pertinent issues flagged by CodeRabbit should also now be addressed.

The `CI / Test on Node 24.x and windows-latest` action is timing out, but other Node 24.x actions pass, as does Node 22.x on windows, so perhaps an issue in the runner? `CodeQL` action is taking awhile—not sure if it's stalled or just takes this long to run.

danenania:
This has now been converted from the "Claude Code Provider" to the "Claude Agent SDK Provider" with accompanying minor updates to the SDK package, parameters, docs, and examples. The provider now uses id `anthropic:claude-agent-sdk` while keeping `anthropic:claude-code` as an alias.

Also: 
- moved the sdk dependency to a peer dependency in line with other providers and included a note to install it separately in docs/examples
- fixed a bug in the `advanced` example where the working directory was set relative to `cwd` rather than the example directory

Dealt with some CI `npm ci` issues. It turned out I needed to `npm install` with npm 10. Doing so with npm 11 caused `npm ci` to fail in the GH action on node 22.x and 20.x. I think it may have had something to do with introducing the new peer dependency but not 100% sure.

@mldangelo

danenania:
Note: I'm polishing up the advanced example with git rollback a bit more.

danenania:
Ok, advanced example is now smoother. This should be ready to go.

https://github.com/user-attachments/assets/52148719-6d2e-4f59-94bb-7a581e4f3aea

PR #5530, opened 2025-09-09
feat: allow user to delete values from numeric inputs and then type

addelong:
## Summary
* Keep type="number" to preserve spinner arrows.
* Smooth editing: allow empty while typing, commit on blur.
* Validation: digits-only filtering; block e/E/+/-/. via onKeyDown.
* Mutual exclusivity: delay ↔ maxConcurrency enforced.
* MUI v6 migration: replace deprecated InputProps/inputProps with slotProps.input and slotProps.input.inputProps.

## Test cases
- [X] Delete all digits in each numeric field and retype; values should accept input smoothly.
- [X] Non-digits should be blocked; spinner arrows still work.
- [X] On blur, values clamp to sensible defaults (delay ≥ 0, concurrency ≥ 1).
- [X] Toggling delay sets concurrency to 1; setting concurrency clears delay to 0.
- [X] Export config still works unchanged.

### DIFF
--- a/src/app/src/pages/redteam/setup/components/RunOptions.tsx
+++ b/src/app/src/pages/redteam/setup/components/RunOptions.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import { FormControlLabel, Switch } from '@mui/material';
@@ -10,8 +10,8 @@ import Stack from '@mui/material/Stack';
 import TextField from '@mui/material/TextField';
 import Tooltip from '@mui/material/Tooltip';
 import Typography from '@mui/material/Typography';
-import { Config } from '../types';
 import type { RedteamRunOptions } from '@promptfoo/types';
+import { Config } from '../types';
 
 const LabelWithTooltip = ({ label, tooltip }: { label: string; tooltip: string }) => {
   return (
@@ -30,7 +30,7 @@ interface RunOptionsProps {
   useGuardrailAssertion?: boolean;
 }
 
-export const RunOptionsContent: React.FC<RunOptionsProps> = ({
+export const RunOptions: React.FC<RunOptionsProps> = ({
   numTests,
   runOptions,
   updateConfig,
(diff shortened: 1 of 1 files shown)


src/app/src/pages/redteam/setup/components/RunOptions.tsx
```diff
@@ -41,111 +41,27 @@ export const RunOptionsContent: React.FC<RunOptionsProps> = ({
...
+  }, [numTests]);
+
+  useEffect(() => {
+    if (runOptions?.delay !== undefined) {
+      setDelayInput(String(runOptions.delay));
+    }
+    if (runOptions?.maxConcurrency !== undefined) {
+      setMaxConcurrencyInput(String(runOptions.maxConcurrency));
+    }
+  }, [runOptions?.delay, runOptions?.maxConcurrency]);
```
faizanminhas: Not clear why these have to be side effects

addelong: Are these inputs the only way these state values can change? e.g. is there no way in the UI to upload a custom config in the middle of the setup process? If not then these don't need to be effects, just seemed safer in case there was.

faizanminhas: Yes, they wouldn't change due to anything other than use input I believe

src/app/src/pages/redteam/setup/components/RunOptions.tsx
```diff
@@ -144,11 +223,24 @@ export const RunOptionsContent: React.FC<RunOptionsProps> = ({
   );
 };
 
+interface RunOptionsProps extends RunOptionsBaseProps {}
+
 export const RunOptions: React.FC<RunOptionsProps> = (props) => {
   const [expanded, setExpanded] = useState(true);
+
+  const normalizedProps: RunOptionsBaseProps = {
```
will-holley: Shouldn't this be `: RunOptionsProps`?

src/app/src/pages/redteam/setup/components/RunOptions.tsx
```diff
@@ -69,24 +110,45 @@ export const RunOptionsContent: React.FC<RunOptionsProps> = ({
             />
           )
         }
-        value={runOptions?.delay ?? 0}
+        value={delayInput}
         disabled={!canSetDelay}
         onChange={(e) => {
```
will-holley, line 120: Are the `onChange`, `onBlur`, and `onKeyDown` logics similar enough across the modified instances where a well-documented shared component is preferable?  If not, I'd like these to be better documented.

src/app/src/pages/redteam/setup/components/RunOptions.tsx
addelong: It could have been either, they were the same shape, the issue was that we didn't need both. Removed RunOptionsBaseProps.

src/app/src/pages/redteam/setup/components/RunOptions.tsx
addelong: They differ enough that I think adding a shared component would introduce unnecessary complexity. I've added comments above each input to document their behavior.

src/app/src/pages/redteam/setup/components/RunOptions.tsx
```diff
@@ -146,9 +241,20 @@ export const RunOptionsContent: React.FC<RunOptionsProps> = ({
 
 export const RunOptions: React.FC<RunOptionsProps> = (props) => {
   const [expanded, setExpanded] = useState(true);
+
+  const normalizedProps: RunOptionsProps = {
+    ...props,
+    numTests: props.numTests ?? 0,
+    runOptions: {
+      delay: props.runOptions?.delay ?? 0,
+      maxConcurrency: props.runOptions?.maxConcurrency ?? 1,
```
will-holley: Let's set the fallback to 5 to match

https://github.com/promptfoo/promptfoo/blob/9061464caee0b84f9d8c7b10d2c74a5c4ac5ae13/src/redteam/commands/generate.ts#L669.

PR #5591, opened 2025-09-13
feat(webui): add keyboard navigation to results table

jameshiester:
The objective of this change is to improve accessibility by enabling keyboard navigation within the results table.  The expectation is that the user should be able to press 'tab' to progress focus from one cell to the next, or to any interactive elements such as buttons or links within the cell.

**Changes Include:**
- Adding  tabIndex to header and body cells to support focus
- Replacing span with button per MDN suggestion (https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/HTML) to enable better keyboard control and screen reader legibility
- Using opacity instead of visibility to maintain focus order when navigating between elements and improving screen reader behavior
- Wrapping tooltip around button to improve aria labeling consistency
- Adding associated test cases

### DIFF
--- a/src/app/src/pages/eval/components/ResultsTable.css
+++ b/src/app/src/pages/eval/components/ResultsTable.css
@@ -161,7 +161,7 @@ table.results-table,
   display: flex;
   gap: 0.5rem;
   position: absolute;
-  top: 1.25rem;
+  top: 1rem;
   right: 0.75rem;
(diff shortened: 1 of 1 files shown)


mldangelo:
Thanks @jameshiester! Please send me an email if you'd like some promptfoo swag!

PR #5612, opened 2025-09-15
chore(webui): improve red team plugin selection warning design

mldangelo:
## Summary

Replace the jarring sticky warning banner with an elegant suggestion card when users select too many plugins in the red team setup flow.

## Changes

- **Better visual design**: Replaced Alert component with custom Paper featuring subtle gradients and accent borders
- **Improved messaging**: Shows actual plugin counts (e.g., "93 vs 39 plugins") instead of vague warnings
- **Enhanced UX**: Feels like helpful guidance rather than an alarming system error
- **Dark mode optimized**: Proper color handling for both light and dark themes
- **Clear call-to-action**: One-click button to switch to recommended preset

## Before/After

**Before**: Aggressive sticky orange warning banner with technical language
**After**: Elegant blue suggestion card with helpful context and clear action

The new design transforms a disruptive warning into thoughtful product guidance that actually helps users make better choices.


typpo:
Can you post a screenshot?

PR #5633, opened 2025-09-16
perf: Don't create new agent for every fetch

faizanminhas:
Suspecting that when concurrency is high, creating a new agent with every `fetchWithProxy` was causing connection issues.


sklein12:
@faizanminhas should we cache the agents based on the URL so we can maintain a couple of them instead of recreating them if we're switching url's back and forth?

faizanminhas:
@sklein12 updated

faizanminhas:
The CI failures are known and not related to this change.

mldangelo (dismissed):
Thanks!

sklein12 (approved):
This looks great! Great job @faizanminhas, you're an A+++ Star!

PR #5641, opened 2025-09-17
chore: Exclude node modules and build/dist from biome

sklein12:
This has stopped biome from taking up 10gb of memory on my machine.


faizanminhas (approved):
incredible

PR #5666, opened 2025-09-19
fix: resolve JSON parsing bug

MrFlounder:
# Fix JSON Parsing Bug in Custom Redteam Strategy

## Problem
The Custom redteam strategy was failing with "Expected eval grader value to be a boolean" errors due to incorrect JSON object extraction. The `extractFirstJsonObject` function was picking up template literals like `{level}` from code diffs in response strings instead of the actual evaluation results.

## Root Cause
When eval responses contained code snippets with template literals (e.g., `logger.log(\`[${level}] ${message}\`)`), the JSON extractor would find `{level}` first and parse it as `{"level": null}` via YAML, causing downstream parsing to fail when expecting `{value: boolean, metadata: number, ...}`.

## Solution
- Added `extractFirstJsonObjectWithKeys<T>(str, requiredKeys)` helper to `src/util/json.ts` that finds the first JSON object containing all required keys
- Updated Custom strategy's `getEvalScore`, `getRefusalScore`, and `getAttackPrompt` methods to use the key-filtered extractor
- Maintains backward compatibility; no changes to other strategies or callers

## Testing
- Fixes Custom strategy evaluation parsing errors
- Crescendo and other strategies remain unaffected
- Added debug logging to verify correct JSON extraction

## Files Changed
- `src/util/json.ts`: Added `extractFirstJsonObjectWithKeys` helper
- `src/redteam/providers/custom/index.ts`: Updated JSON parsing calls to use key-filtered extraction

### DIFF
--- a/src/redteam/providers/custom/index.ts
+++ b/src/redteam/providers/custom/index.ts
@@ -1,27 +1,9 @@
 import dedent from 'dedent';
 import { v4 as uuidv4 } from 'uuid';
+
 import { renderPrompt } from '../../../evaluatorHelpers';
 import logger from '../../../logger';
 import { PromptfooChatCompletionProvider } from '../../../providers/promptfoo';
-import invariant from '../../../util/invariant';
-import { extractFirstJsonObject, safeJsonStringify } from '../../../util/json';
-import { getNunjucksEngine } from '../../../util/templates';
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Tests!

PR #5705, opened 2025-09-23
fix(webui): prioritize JSON prettify over Markdown rendering when both enabled

### LINKED ISSUE #5704: JSON Prettify option does not work when Markdown rendering is enabled in WebUI
**Describe the bug**
JSON Prettify option does not work when Markdown rendering is enabled simultaneously in the WebUI settings. When both "Render Markdown content" and "Prettify JSON outputs" options are enabled, JSON outputs are rendered as plain text through Markdown processing instead of being properly prettified with syntax highlighting and formatting.

Additionally, if JSON content contains parts that can be interpreted as Markdown, those parts get unintentionally rendered as Markdown markup, corrupting the JSON display.

**To Reproduce**
1. 
2. Set up a provider (Python/JS/TS) that returns object outputs
3. Configure transform functions to extract relevant parts for evaluation (content can be either Markdown text or JSON)
4. Open the Promptfoo WebUI and navigate to Settings
5. Enable both "Render Markdown content" and "Prettify JSON outputs" options
6. Run an evaluation and view results in the evaluation table
7. Observe that JSON output is displayed as plain text instead of being prettified

**Expected behavior**

When both options are enabled, the system should intelligently handle different content types:
- Valid JSON content should be prettified with proper formatting and syntax highlighting
- Non-JSON text content should be rendered as Markdown
- Both display options should work rather than conflicting with each other
- I think JSON formatting should take priority over Markdown rendering for valid JSON content

**Screenshots**

Image

https://gist.github.com/pokutuna/8fc497e03a2e81259cd95d75cba50927

**System information:**

- Promptfoo version: 0.118.8 (latest when this issue created)

**Additional context**

The issue is in `src/app/src/pages/eval/components/EvalOutputCell.tsx`. The rendering priority checks Markdown before JSON prettifying, so when `renderMarkdown` is enabled, it processes ALL text including valid JSON, and the `prettifyJson` condition is never reached.

Reorder the conditional checks to prioritize JSON detection over Markdown rendering - check for valid JSON first, then fallback to Markdown for non-JSON content.

pokutuna:
resolves #5704 

This is necessary for use cases where we need to evaluate outputs that include both Markdown text for users and multiple structured data from an Agent.

Markdown rendering (top) and JSON output (bottom)

image

https://gist.github.com/pokutuna/8fc497e03a2e81259cd95d75cba50927

### DIFF
--- a/src/app/src/pages/eval/components/EvalOutputCell.tsx
+++ b/src/app/src/pages/eval/components/EvalOutputCell.tsx
@@ -242,30 +242,40 @@ function EvalOutputCell({
         )}
       </div>
     );
-  } else if (renderMarkdown && !showDiffs) {
-    node = (
-      <ReactMarkdown
-        remarkPlugins={[remarkGfm]}
-        components={{
-          img: ({ src, alt }) => (
-            <img
-              loading="lazy"
-              src={src}
-              alt={alt}
-              onClick={() => toggleLightbox(src)}
-              style={{ cursor: 'pointer' }}
-            />
-          ),
-        }}
-      >
-        {text}
-      </ReactMarkdown>
-    );
-  } else if (prettifyJson) {
-    try {
-      node = <pre>{JSON.stringify(JSON.parse(text), null, 2)}</pre>;
-    } catch {
-      // Ignore because it's probably not JSON.
+  } else if ((prettifyJson || renderMarkdown) && !showDiffs) {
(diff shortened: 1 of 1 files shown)


pokutuna:
@coderabbitai generate docstrings

mldangelo:
Thank you so much @pokutuna! Please send me an email if you would like any more promptfoo swag!

PR #5738, opened 2025-09-25
feat: Compress data when sharing an eval

sklein12:
We're running into some massive evals where the config is > 50mb with 30k tests. This makes for a massive http request. Using gzip compresses that down to 8mb which is much more manageable.  This is not an ideal solution but that would require a major re-work of our data model which is not in the cards right now.

Added it as an option in our fetch so it can be re-used other places in the future if needed.

### DIFF
--- a/src/share.ts
+++ b/src/share.ts
@@ -102,6 +102,8 @@ async function sendEvalRecord(
   headers: Record<string, string>,
 ): Promise<string> {
   const evalDataWithoutResults = { ...evalRecord, results: [] };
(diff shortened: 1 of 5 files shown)


src/share.ts
```diff
@@ -579,7 +594,7 @@ export async function createShareableModelAuditUrl(
  * @returns The shareable URL for the model audit.
  */
 export function getShareableModelAuditUrl(
-  audit: ModelAudit,
+  _audit: ModelAudit,
```
sklein12, line 597: unrelated but the linter complained about it

PR #5757, opened 2025-09-29
fix(cli): --filter-failing not working correctly with scenarios

typpo:
## Summary
- Fixed `--filter-failing` command to correctly handle test suites that use scenarios
- Maintains backwards compatibility with regular test suites
- Adds comprehensive test coverage for scenario filtering

## Problem
When using `eval --filter-failing` with a configuration that uses scenarios, the filter was not correctly matching the expanded test cases. This caused it to run incorrect tests (sometimes passing tests instead of failing ones) because scenarios expand into multiple test cases at runtime that don't exist in the original `testSuite.tests` array.

## Solution
The fix modifies `filterTestsByResults` to:
1. First try to match against the original test suite structure (for backwards compatibility)
2. When scenarios are present, also extract test cases directly from the evaluation results
3. Properly deduplicate test cases when both approaches find matches

## Test plan
- [x] All existing tests pass
- [x] Added new test suite `filterTestsScenarios.test.ts` with comprehensive coverage for:
  - Filtering scenario-expanded tests
  - Mixed scenarios and regular tests
  - Integration with the `filterTests` function
- [x] Manually tested with scenario configurations
- [x] Lint and format checks pass

Fixes the issue described where `--filter-failing` would run 2 tests (1 success, 1 failure) instead of the expected 19 failures when used with scenarios.

### DIFF
--- a/src/commands/eval/filterTestsUtil.ts
+++ b/src/commands/eval/filterTestsUtil.ts
@@ -22,10 +22,6 @@ export async function filterTestsByResults(
   pathOrId: string,
   filterFn: ResultFilterFn,
 ): Promise<Tests> {
-  if (!testSuite.tests) {
-    return [];
-  }
-
   let results: { results: EvaluateResult[] };
   try {
     if (pathOrId.endsWith('.json')) {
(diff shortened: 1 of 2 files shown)


faizanminhas:
closing in favour of https://github.com/promptfoo/promptfoo/pull/5770

PR #5763, opened 2025-09-29
chore: claude 4.5 sonnet

typpo:


addelong (dismissed):

src/providers/anthropic/util.ts
```diff
@@ -21,6 +21,7 @@ export const ANTHROPIC_MODELS = [
   })),
   ...[
     'claude-sonnet-4-5-20250929',
+    'claude-sonnet-4-5-latest',
```
faizanminhas, line 24: this isn't a valid key

PR #5764, opened 2025-09-29
feat: Add support for claude sonnet 4.5

faizanminhas:


src/providers/anthropic/messages.ts
```diff
@@ -169,39 +170,64 @@ export class AnthropicMessagesProvider extends AnthropicGenericProvider {
     }
 
     try {
-      const response = await this.anthropic.messages.create(params, {
-        ...(typeof headers === 'object' && Object.keys(headers).length > 0 ? { headers } : {}),
-      });
-      logger.debug(`Anthropic Messages API response: ${JSON.stringify(response)}`);
+      if (shouldStream) {
+        // Handle streaming request
+        const stream = await this.anthropic.messages.stream(params, {
```
faizanminhas, line 175: Added streaming, since for `thinking: enabled: true`, I was seeing the following error from anthropic:

```API call error: Streaming is required for operations that may take longer than 10 minutes. See https://github.com/anthropics/anthropic-sdk-typescript#long-requests for more details```

File: src/providers/anthropic/util.ts [hunk 2]
```diff
@@ -198,6 +203,25 @@ export function calculateAnthropicCost(
...
+  const isSonnet45 = ['claude-sonnet-4-5-20250929'].includes(modelName);
+
+  if (
+    isSonnet45 &&
+    Number.isFinite(promptTokens) &&
+    Number.isFinite(completionTokens) &&
+    typeof promptTokens !== 'undefined' &&
+    typeof completionTokens !== 'undefined'
+  ) {
+    // Tiered pricing for Claude Sonnet 4.5:
```
faizanminhas, line 216: worth calling out that sonnet-4.5 is priced differently now. (https://claude.com/pricing#api)

addelong:
@faizanminhas I think we need to add it to ProviderTypeSelector.tsx as well

faizanminhas:
> @faizanminhas I think we need to add it to ProviderTypeSelector.tsx as well

I don't think we need to, since for anthropic, and other foundation models, we just let the users type the model identifier manually

PR #5765, opened 2025-09-29
fix: ensure python and js providers have appropriate path prefix

addelong:
## Summary
Fixes an issue where Python and JavaScript provider file paths created through the UI would lose their `file://` prefix, causing "Could not identify provider" errors when running `npx promptfoo redteam run` from the CLI.

## Changes
  - **CustomTargetConfiguration.tsx**: Auto-add `file://` prefix for `.py` and `.js` paths, strip for display
  - **CustomTargetConfiguration.test.tsx**: Added comprehensive test coverage (7 tests)

  ## Tests
  - ✅ All new tests passing (7/7)
  - ✅ Existing tests passing
  - ✅ Lint and format checks passing

### DIFF
--- a/src/app/src/pages/redteam/setup/components/Targets/CustomTargetConfiguration.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/CustomTargetConfiguration.tsx
@@ -23,16 +23,23 @@ const CustomTargetConfiguration = ({
   setRawConfigJson,
   bodyError,
 }: CustomTargetConfigurationProps) => {
-  const [targetId, setTargetId] = useState(selectedTarget.id || '');
+  const [targetId, setTargetId] = useState(selectedTarget.id?.replace('file://', '') || '');
 
   useEffect(() => {
-    setTargetId(selectedTarget.id || '');
+    setTargetId(selectedTarget.id?.replace('file://', '') || '');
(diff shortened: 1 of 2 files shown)


src/app/src/pages/redteam/setup/components/Targets/CustomTargetConfiguration.tsx
```diff
@@ -23,16 +23,23 @@ const CustomTargetConfiguration = ({
...
 
   const handleTargetIdChange = (e: React.ChangeEvent<HTMLInputElement>) => {
-    const newId = e.target.value;
-    setTargetId(newId);
-    updateCustomTarget('id', newId);
+    const value = e.target.value;
+    setTargetId(value);
+
+    let idToSave = value;
+    if (value && (value.endsWith('.py') || value.endsWith('.js'))) {
```
mldangelo: should also handle the case where they use a non standard function name, eg. myprovider.py:func See https://www.promptfoo.dev/docs/providers/python/#custom-function-names

addelong: good call

mldangelo (commented):
This looks great, thank you so much for writing tests. One important comment.

Issue #5755, opened 2025-09-28
Eval Filter Failing not working


rvernica:
**Describe the bug**

I'm using `eval --filter-failing` to re-run only the failing tests. The tests executed does not make any sense given the filter.

**To Reproduce**
Steps to reproduce the behavior, including example Promptfoo configurations if applicable:

1. Run `promptfoo eval`. It returns _Successes: 49 Failures: 19 Errors: 6_
2. Run `promptfoo eval --filter-failing run_id_from_step_1`. This runs two tests. Of the two tests, one is from the Successes list and one if from the Failures list. This does not make sense. I expect it would run the 16 failed tests.

**Screenshots**

```
> promptfoo eval
Starting evaluation eval-1k8-2025-09-28T15:42:06
Running 74 test cases (up to 4 at a time)...
Evaluating [████████████████████████████████████████] 100% | 74/74 ...
...
===================================================================================================================================
✔ Evaluation complete. ID: eval-1k8-2025-09-28T15:42:06

» Run promptfoo view to use the local web viewer
» Do you want to share this with your team? Sign up for free at https://promptfoo.app
» This project needs your feedback. What's one thing we can improve? https://promptfoo.dev/feedback
===================================================================================================================================
Duration: 1m 29s (concurrency: 4)
Successes: 49
Failures: 19
Errors: 6
Pass Rate: 66.22%

>  promptfoo eval --filter-failing eval-1k8-2025-09-28T15:42:06
Starting evaluation eval-3CD-2025-09-28T15:44:38
Running 2 test cases (up to 4 at a time)...
Evaluating [████████████████████████████████████████] 100% | 2/2
...
```

**System information:**

If possible, please output the results of `promptfoo debug` and paste the output here.

```
 promptfoo debug
===================================================================================================================================
Promptfoo Debug Information
===================================================================================================================================
{
  "version": "0.118.10",
  "platform": {
    "os": "linux",
    "release": "6.16.7-200.fc42.x86_64",
    "arch": "x64",
    "nodeVersion": "v24.6.0"
  },
  "env": {},
  "configInfo": {
    "configExists": false,
    "configContent": null
  }
}
===================================================================================================================================
Please include this output when reporting issues on GitHub: https://github.com/promptfoo/promptfoo/issues
```

`promptfooconfig.yaml` contains tests and scenarios spread across 10 fails listed under `tests` and `scenarios`.

mldangelo:
Hi @rvernica,

[...]

PR #5770, opened 2025-09-29
fix: Ensure that `-filter-failing` correctly filters failing tests when re-running an eval

faizanminhas:
## Summary

Relates to https://github.com/promptfoo/promptfoo/issues/5755

## Description

Minimal reproduction here:
 <details>

<summary>promptfooconfig.yaml</summary>

```yaml
description: 'Bug reproduction: --filter-failing not matching tests correctly'

prompts:
  - 'Translate "{{input}}" to {{language}}: {{_conversation}}'

providers:
  - python:provider.py

tests:
  - description: 'Test 1 - Should pass'
    vars:
      input: 'Hello'
      language: 'French'
    assert:
      - type: contains
        value: 'Bonjour'

  - description: 'Test 2 - Should fail (expects wrong translation)'
    vars:
      input: 'Goodbye'
      language: 'French'
    assert:
      - type: contains
        value: 'Salut' # Wrong - will fail since output is "Bonjour"

  - description: 'Test 3 - Should pass'
    vars:
      input: 'Thank you'
      language: 'French'
    assert:
      - type: contains
        value: 'Bonjour'

  - description: 'Test 4 - Should fail (expects wrong output)'
    vars:
      input: 'Good morning'
      language: 'French'
    assert:
      - type: contains
        value: 'Au revoir' # Wrong - will fail

  - description: 'Test 5 - Should fail (expects wrong output)'
    vars:
      input: 'Good night'
      language: 'Spanish'
    assert:
      - type: equals
        value: 'Hola' # Wrong - will fail
```

</details>

<details><summary>provider.py</summary>

```python
"""Simple mock provider that returns a fixed response."""


def call_api(prompt, options=None, context=None):
    """Always return Bonjour as the translation."""
    return {"output": "Bonjour"}

```

</details>

When evaluation runs, it adds runtime variables (like `_conversation`) to `test.vars`, mutating the original test object. This mutated test is saved in the results. On subsequent runs with `--filter-failing`, freshly loaded tests don't have these runtime vars, so the deep equality comparison fails and tests don't match. This fixes the check by filtering out meta vars.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -18,11 +18,11 @@ import { collectFileMetadata, renderPrompt, runExtensionHook } from './evaluator
 import logger from './logger';
 import { selectMaxScore } from './matchers';
 import { generateIdFromPrompt } from './models/prompt';
+import { CIProgressReporter } from './progress/ciProgressReporter';
 import { maybeEmitAzureOpenAiWarning } from './providers/azure/warnings';
 import { isPromptfooSampleTarget } from './providers/shared';
 import { generatePrompts } from './suggestions';
 import telemetry from './telemetry';
-import { CIProgressReporter } from './progress/ciProgressReporter';
 import {
   generateTraceContextIfNeeded,
   isOtlpReceiverStarted,
--- a/src/util/index.ts
+++ b/src/util/index.ts
@@ -1,4 +1,5 @@
 import * as fs from 'fs';
+import * as os from 'os';
 import * as path from 'path';
 
 import { stringify } from 'csv-stringify/sync';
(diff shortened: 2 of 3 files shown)


addelong:
@faizanminhas is this a duplicate of https://github.com/promptfoo/promptfoo/pull/5757 ?

faizanminhas:
@addelong Good catch, I totally missed that. That said, it seems like my fix is way neater and smaller?

addelong:
Yeah, I like this one better too, please make sure that other one gets closed 🙏

PR #5562, opened 2025-09-11
chore(webui): UX improvements for displaying custom policies in Eval Results and Red Team Vulnerabilities Reports

will-holley:
## Overview

- Eval Results View
  - Refactor: Separates each policy into its own metric (currently all custom policies appear as the same metric)
    - If a policy is reusable, it's demarcated by its name; otherwise a unique hash of its text
    - Hovering over a metric pill will display the policy text in a tooltip
    - Reusable policies link to their detail page in Cloud. This is also available from the metadata panel.
    - Adds a policy filter
      - This is partially out of preferring simplicity (extending the metric or plugin filters would introduce complexity into stable functionality due to the requirement that the filtered upon value include _both_ that the value is a policy and the policy id w/ specialized server-side logic to handle this) and future extensibility (e.g. better support for filtering by policy text, which is presently available as a metadata filter isn't obvious or intuitively located).
  - Fix: Use severity display names when filtering
  - Chore: Ensures policies can be filtered on
- Red Team Report View
  - Fix: Ensuring overridden severities are correctly read in overview card stats (see [Overview.tsx](https://github.com/promptfoo/promptfoo/blob/c7bf6a5526a476302ef5b3e0ed9670b1937ef729/src/app/src/pages/redteam/report/components/Overview.tsx))
  - Display each policy as its own row in Vulnerabilities table
  - Removes policies from framework compliance section ("Policy Compliance" is currently displayed under MITRE, NIST, and EU)
  - View all logs correctly navigates to the filtered results view

### Discrete Policy Metrics

image

### Support for Reusable and Inline Policies

image

### Policy Text Tooltips

image

### Policy Filter Applied

image

### Metadata Linking

image

### Vulnerabilities

**Note:** I disabled the ASR > 0 filter here.

image

PR #5785, opened 2025-09-30
chore(internals): Prevent `GradingResult.assertion` definition from being overridden in select red team grading cases

will-holley:
## Overview

Uncovered this while working on https://github.com/promptfoo/promptfoo/pull/5562.  Essentially `gradingResult.assertion` will be erroneously null (which has downstream effects) when a red team grader's `getResult` method either directly (doesn't override) or indirectly (calls `super#getResult`) returns `RedteamGraderBase#getResult`.

## Changes

- In `src/assertions/redteam.ts`, ensures the `assertion` argument passed to `handleRedteam` is returned on the `GradingResult` object when `grade.assertion` returned by `grader#getResult` is `null`.
- Explicitly type `null` `grade.assertion` returned by `getResult`.
- Adds unit tests for `handleRedteam` that cover the changes here. This function could use more robust unit testing.
- Splits out some related utility functions to use in the `handleRedteam` unit tests & adds unit tests for these; https://github.com/promptfoo/promptfoo/pull/5785/commits/ebc71988be8e3fc3fca335649af6684f74504239.


src/assertions/index.ts
```diff
@@ -180,6 +180,27 @@ const ASSERTION_HANDLERS: Record<
 
 const nunjucks = getNunjucksEngine();
 
+/**
+ * Tests whether an assertion is inverse e.g. "not-equals" is inverse of "equals"
+ * or "not-contains" is inverse of "contains".
+ * @param assertion - The assertion to test
+ * @returns true if the assertion is inverse, false otherwise
+ */
+export function isAssertionInverse(assertion: Assertion): boolean {
```
faizanminhas, line 189: Is negation seems more appropriate

will-holley: https://www.promptfoo.dev/docs/configuration/expected-outputs/deterministic/#inverse-assertion

PR #5786, opened 2025-09-30
chore(site): Adds Travis to team page

will-holley:


site/static/img/team/travis.png
mldangelo, line 1: jpeg here

PR #5794, opened 2025-10-01
fix: import command

sklein12:
Fix importing from exported Promptfoo json files.


will-holley (approved):
Tested locally w/ an eval that I was previously unable to report. Imports as expected. Thanks, @sklein12!

PR #5796, opened 2025-10-01
fix(site): add metadata key to the provider response class

yash2998chhabria:
The metadata field (Dict[str, Any]) is now documented in the Python provider ProviderResponse type definition. This field allows providers to return arbitrary metadata that will be displayed in the Metadata tab of the evaluation output dialog drawer in the web UI.

The metadata is rendered in a key-value table format via the MetadataPanel component, making it easy for users to inspect additional provider-specific information such as HTTP headers, request details, or custom debugging data.


will-holley (approved):
Thanks for triaging this!

PR #5814, opened 2025-10-03
fix(webui): Clicking "Show Charts" does not show charts.

will-holley:
Logic to determine whether charts are renderable was split between the charts component and its parent scope. This fix moves all condition rendering logic into the parent scope. https://linear.app/promptfooo/issue/ENG-129/chart-toggle-bug.


mldangelo (approved):
Thanks!

PR #5817, opened 2025-10-03
refactor(webui): remove delimiter stripping logic from EvalOutputCell

yash2998chhabria:
Remove the '---' delimiter stripping logic from the web UI component since fail reasons are already extracted directly from componentResults array and displayed via FailReasonCarousel. This logic was redundant in the web UI and is only needed for CLI display formatting.


mldangelo (approved):
Thanks!

PR #5823, opened 2025-10-04
fix(provider): merge config and prompt systemInstruction instead of throwing error in gemini

### LINKED ISSUE #5822: Gemini: Support merging config and assertion systemInstructions using parts
**Describe the bug**

The Gemini provider throws a "Template error: system instruction defined in prompt and config" error when both `config.systemInstruction` and assertion-added system messages are present. This prevents users from setting a global system instruction in the provider config while using model-graded assertions.

Several assertions provide system messages for grading, including:
- `llm-rubric` - for custom rubric-based evaluation
- `model-graded-factuality` - for factuality checking
- Other model-graded assertion types

These assertions cannot be used in combination with `config.systemInstruction` due to this conflict.

A common use case is setting a global instruction like "Think in English and output the evaluation results in Japanese" in `config.systemInstruction`, which should apply to all model-graded assertions. This allows consistent output language and thinking process across all evaluations.

**To Reproduce**

## Reproduction

The issue occurs when:
1. Provider config includes `systemInstruction`
2. A test uses a model-graded assertion (e.g., `llm-rubric`) which internally provides a system message for grading
3. Both system instructions conflict and throw an error

**Reproduction configuration** (`promptfooconfig.yaml`):

```yaml
prompts:
- What is 2+2?

providers:
- vertex:gemini-2.5-flash

defaultTest:
  options:
    provider:
      id: vertex:gemini-2.5-flash
      config:
        systemInstruction: |
          You are a pirate.
          You MUST respond to EVERY question using pirate language with 'Arrr', 'matey', 'ye', etc.
          Never use normal English.

tests:
- assert:
  - type: llm-rubric
    value: |
      Evaluate the output
      Score of 0.0: Response does not answer the math question
      Score of 0.5: Response answers the math question
      Score of 1.0: Response answers the math question correctly (the answer is 4)
    threshold: 1.0
```

Running `promptfoo eval` with this configuration will produce the error.

**Expected behavior**

Currently, the `geminiFormatAndSystemInstructions` function in `src/providers/google/util.ts` throws an error when both system instructions are present. However, this conflict can be avoided by leveraging the Gemini API's native support for multiple `parts` within a single `systemInstruction`.

The Gemini API allows only one `systemInstruction` per request, but this instruction can contain multiple `parts`. The Gemini provider should merge both system instructions (from config and from model-graded assertions) into a single `systemInstruction` with multiple parts:

```json
{
  "systemInstruction": {
    "parts": [
      {"text": "config system instruction"},
      {"text": "model-graded assertion system message"}
    ]
  }
}
```

By separating them as distinct parts, both the user's config-level system instruction and the model-graded assertion's system message can be included without conflict.

Merge order should be: config parts first, then assertion parts. This allows the config to set general instructions while assertions add specific grading instructions.

**Screenshots**

Image

**System information:**

- Promptfoo version: 0.118.11

**Additional context**

I'll create a pull request to fix this issue

Other Related Issue

- #3656

pokutuna:
resolves #5822 

Previously, when both config.systemInstruction and prompt system role existed, the Gemini provider would throw an error. This changes the behavior to merge both instructions with config parts first, followed by prompt parts.

Changes:
- Extract parseConfigSystemInstruction() helper function for cleaner code
- Merge systemInstruction parts instead of rejecting the combination

---

And it works with this repro.

```yaml

prompts:
- What is 2+2?

providers:
- vertex:gemini-2.5-flash

defaultTest:
  options:
    provider:
      id: vertex:gemini-2.5-flash
      config:
        systemInstruction: |
          You are a pirate.
          You MUST respond to EVERY question using pirate language with 'Arrr', 'matey', 'ye', etc.
          Never use normal English.

tests:
- assert:
  - type: llm-rubric
    value: |
      Evaluate the output
      Score of 0.0: Response does not answer the math question
      Score of 0.5: Response answers the math question
      Score of 1.0: Response answers the math question correctly (the answer is 4)
    threshold: 1.0
```

image

### DIFF
--- a/src/providers/google/util.ts
+++ b/src/providers/google/util.ts
@@ -752,7 +752,10 @@ export function geminiFormatAndSystemInstructions(
 
   let systemInstruction: Content | undefined = parsedSystemInstruction;
 
-  const parsedConfigInstruction = parseConfigSystemInstruction(configSystemInstruction, contextVars);
+  const parsedConfigInstruction = parseConfigSystemInstruction(
+    configSystemInstruction,
+    contextVars,
+  );
   if (parsedConfigInstruction) {
     systemInstruction = systemInstruction
       ? { parts: [...parsedConfigInstruction.parts, ...systemInstruction.parts] }
--- a/test/providers/google/util.test.ts
+++ b/test/providers/google/util.test.ts
@@ -1034,7 +1034,7 @@ describe('util', () => {
     it('should merge system messages from both prompt and config with string config', () => {
       const prompt = [
         { role: 'system', content: 'prompt system instruction' },
-        { role: 'user', content: 'user message' }
+        { role: 'user', content: 'user message' },
       ];
       const { contents, systemInstruction } = geminiFormatAndSystemInstructions(
         JSON.stringify(prompt),
@@ -1043,17 +1043,14 @@ describe('util', () => {
       );
       expect(contents).toEqual([{ parts: [{ text: 'user message' }], role: 'user' }]);
       expect(systemInstruction).toEqual({
-        parts: [
-          { text: 'config system instruction' },
-          { text: 'prompt system instruction' }
-        ]
+        parts: [{ text: 'config system instruction' }, { text: 'prompt system instruction' }],
       });
     });
 
     it('should merge system messages from both prompt and config with object config', () => {
       const prompt = [
         { role: 'system', content: 'prompt system instruction' },
-        { role: 'user', content: 'user message' }
+        { role: 'user', content: 'user message' },
       ];
       const { contents, systemInstruction } = geminiFormatAndSystemInstructions(
         JSON.stringify(prompt),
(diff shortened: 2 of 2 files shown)


src/providers/google/util.ts
```diff
@@ -704,35 +750,13 @@ export function geminiFormatAndSystemInstructions(
...
-    // Load SI from file
-    if (typeof configSystemInstruction === 'string') {
-      systemInstruction = loadFile(configSystemInstruction, contextVars);
-    }
-
-    // Format SI if string was not a filepath above
-    if (typeof systemInstruction === 'string') {
-      systemInstruction = { parts: [{ text: systemInstruction }] };
-    }
+  let systemInstruction: Content | undefined = parsedSystemInstruction;
```
pokutuna, line 753: Change type from `Content | string | undefined` to `Content | undefined`.
According to the type definition of `maybeCoerceToGeminiFormat`, `string` cannot be passed directly.

PR #5829, opened 2025-10-06
feat(webui): improved color consistency and dark mode legibility on Red Team dashboard

jameshiester:
The purpose of this PR is to improve the consistency of colors and legibility for the Redteam dashboard
- Severity and Risk colors have been consolidated and are now derived directly from the app theme using MUI theme extension functionality.
- SeverityCard has been extracted as a reusable component.  The styling for cards with no vulnerabilities has been modified to use the disabled colors from MUI instead of reducing opacity, which was causing legibility issues in dark mode.
- CSS has been cleaned up so that success and error colors are consistently driven from the theme to ensure consistency in both light and dark mode
- renderWithProviders test util has been created to allow components to be rendered with higher order components.  Proposal is to extend this with commonly used providers to simplify test creation.  

**Why use the theme instead of css/constants?**
This approach focuses on deriving colors from the main theme, using MUI utils like alpha and augmentColor to create transparent shades and contrast text colors based on the main color in the theme setting.  This allows for more consistency between the shade and the original color in light and dark mode, and will remove footguns if colors are changed in the future.


mldangelo (approved):
This is a great improvement, thank you. Sorry about the delay in review. Please merge it!

ghost:
**✔️ Successfully committed 25 tests** @jameshiester

PR #5832, opened 2025-10-06
docs: adding multi turn conversation example and docs

jameshiester:
* Adding example for multiple turn red team with custom python provider


examples/custom-provider-multi-turn/multi-turn.py
```diff
@@ -0,0 +1,68 @@
+import json
```
mldangelo: recommend using the openai python sdk here and simplifying the file. urllib and ssl are confusing. 

You should also add docstrings to explain what's going on in the file!

examples/custom-provider-multi-turn/promptfooconfig.yaml
```diff
@@ -0,0 +1,30 @@
+description: Red Team Starter
```
mldangelo: be more descriptive

examples/redteam-custom-provider-multi-turn/promptfooconfig.yaml
mldangelo, line 1: take a look at the other examples, they all have a README.md file.

examples/redteam-custom-provider-multi-turn/promptfooconfig.yaml
```diff
@@ -0,0 +1,30 @@
...
+    transformVars: '{ ...vars, sessionId: context.uuid }'
+
+redteam:
+  purpose: You are a helpful assistant.
+  plugins:
+    - id: harmful:hate
+  strategies:
+    - id: goat
+      config:
+        stateful: true
```
mldangelo: you are using stateful: true here but you don't explain that the example is stateful.

examples/custom-provider-multi-turn/multi-turn.py
mldangelo, line 1: most redteam eamples are prefixed with redteam- in the folder name, consider doing that.

site/docs/red-team/configuration.md
```diff
@@ -56,7 +56,7 @@ redteam:
 | `injectVar`                  | `string`                  | Variable to inject adversarial inputs into                               | Inferred from prompts           |
 | `numTests`                   | `number`                  | Default number of tests to generate per plugin                           | 5                               |
 | `plugins`                    | `Array<string\|object>`   | Plugins to use for red team generation                                   | `default`                       |
-| `provider` or `targets`      | `string\|ProviderOptions` | Endpoint or AI model provider for generating adversarial inputs          | `openai:gpt-4.1`                |
+| `provider`                   | `string\|ProviderOptions` | Endpoint or AI model provider for generating adversarial inputs          | `openai:gpt-4.1`                |
```
mldangelo: in the red team context we prefer the term targets, recommend keeping it here

site/docs/red-team/multi-turn-conversations.md
```diff
@@ -0,0 +1,32 @@
+---
+sidebar_position: 10000
```
mldangelo, line 3: If you want to keep this file, you should copy the entire example into here, add fontmatter directives, possibly screen shots. Make the example interesting enough too that people will look to read and emulate it (as opposed to one plugin and one test case).

mldangelo (commented):
Thanks for doing this. I've left detailed feedback to help make this example more clear and to follow our existing patterns. - the core pattern you're using (transformVars with context.uuid) is good, it just needs better documentation. I also think you should show both the stateless and stateful approaches so users know when to use each. Thanks @jameshiester!

site/docs/red-team/configuration.md
jameshiester: Here the docs are declaring the properties within the Red Team configuration.  Targets is not a valid field inside Red Team.  The description also matches provider, not description.  Other sentences in the docs call out that it is different than the root target

PR #5850, opened 2025-10-07
feat: Bring request transform to parity with response transform

faizanminhas:
Currently, the request transform is a lot less flexible than the response transform, which can cause confusion. For now, request transform doesn't support function strings, and js expressions require `return`, or else the value isn't properly returned by the transform. This makes the two functions have the same flexibility.

### DIFF
--- a/src/providers/http.ts
+++ b/src/providers/http.ts
@@ -1119,7 +1119,7 @@ export async function createTransformRequest(
             'prompt',
             'vars',
             'context',
-            `try { return (${trimmedTransform})(prompt, vars, context); } catch(e) { throw new Error('Transform failed: ' + e.message + ' : prompt=' + prompt + ' : vars=' + JSON.stringify(vars) + ' : context=' + JSON.stringify(context)); }`,
+            `try { return (${trimmedTransform})(prompt, vars, context); } catch(e) { throw new Error('Transform failed: ' + e.message) }`,
           );
         } else {
           // Check if it contains a return statement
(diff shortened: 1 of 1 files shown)


src/providers/http.ts
```diff
@@ -1074,54 +1075,87 @@ export async function createTransformRequest(
...
+        const isFunctionExpression = /^(\(.*?\)\s*=>|function\s*\(.*?\))/.test(trimmedTransform);
+
+        let transformFn: Function;
+        if (isFunctionExpression) {
+          // For function expressions, call them with the arguments
+          transformFn = new Function(
+            'prompt',
+            'vars',
+            'context',
+            `try { return (${trimmedTransform})(prompt, vars, context); } catch(e) { throw new Error('Transform failed: ' + e.message + ' : prompt=' + prompt + ' : vars=' + JSON.stringify(vars) + ' : context=' + JSON.stringify(context)); }`,
```
MrFlounder: context might have sensitive data, e.g api key in header

faizanminhas: good point, will remove that logging and hopefully we don't need it later 😅

faizanminhas:
> This looks reasonable to me! Is there any refactoring we can or should do to make the code more DRY here? Thank you @faizanminhas

I'm okay with not DRY-ing this code out. The main difference is that we support nunjucks interpolating in the request transform, so factoring that out would give little benefit and make it less flexible. Maybe when we support 3 types of transforms 🤷

PR #5852, opened 2025-10-07
fix: filter out placeholders before purpose generation

danenania:
This ensures the `DEFAULT_PURPOSE` is used as intended when no prompts are passed, instead of treating template placeholders like `{{prompt}}` as an actual prompt and generating a purpose based on it.

Note: includes https://github.com/promptfoo/promptfoo/pull/5851 which fixes a failing test (that PR should be merged before this one).


mldangelo (approved):
Great!

PR #5856, opened 2025-10-08
chore: invert default unblocking behavior

MrFlounder:


test/redteam/providers/shared.test.ts
```diff
@@ -743,18 +743,18 @@ describe('shared redteam provider utilities', () => {
...
       }
     });
 
-    it('short-circuits when PROMPTFOO_DISABLE_UNBLOCKING=true', async () => {
-      process.env.PROMPTFOO_DISABLE_UNBLOCKING = 'true';
+    it('short-circuits by default when PROMPTFOO_ENABLE_UNBLOCKING is not set', async () => {
+      delete process.env.PROMPTFOO_ENABLE_UNBLOCKING;
 
       const result = await tryUnblocking({
         messages: [],
```
danenania, line 760: is it worth adding another test case for when the env var is true? (that it doesn't short circuit)

danenania (approved):

PR #5864, opened 2025-10-08
fix: Requested column in 'redteam generate' output incorporates fan out strategies

danenania:
When running `redteam generate`, the `Requested` column would be too low for strategies that fan out (really only `jailbreak:composite` so far) when `n` isn't set in the config, because the cloud server uses a hardcoded default that wasn't included in the calculation.

image

With this fix, the default value is moved to a constant so we can reference it when calculating `Requested`.

Note that `jailbreak:composite` now shows 50/50, whereas it would previously show 10/50:

Screenshot 2025-10-07 at 5 14 38 PM

A cloud PR will follow on from this one that references these constants instead of hardcoding defaults. Both PRs will need to be merged for the fix, but this one goes first.


danenania:
Also fixed the summary to match as well:

Screenshot 2025-10-08 at 2 34 14 PM

And fixed a little issue that was causing an extra space to get inserted sometimes in the summary output.

src/redteam/index.ts
```diff
@@ -1,22 +1,23 @@
-import * as fs from 'fs';
-
 import async from 'async';
 import chalk from 'chalk';
 import cliProgress from 'cli-progress';
 import Table from 'cli-table3';
+import * as fs from 'fs';
```
mldangelo, line 5: No action necessary. I am guessing this is from your editor biome config. I am ok updating the import sort order but we should do it consistently and ideally all at once. Something to add to the backlog

danenania: This is actually the result of `npm run format` so not related to IDE issues. I have noticed that import order and some other formatting has been thrashing a little on other files as well (again just with `npm run format`). Maybe updated biome config?

CleanShot 2025-10-09 at 10 18 38

PR #5912, opened 2025-10-13
chore: validate emails after prompt for redteam evals

danenania:
The cloud `/status` endpoint now accepts a `validate` query param that will check if the email is sufficiently likely to be a valid work email. If it's not, the user is asked to give a valid work email. This only occurs when the user is first prompted for their email (if it's not already in the auth config in `~/.promptfoo/promptfoo.yaml`).

Emails captured through the UI are now similarly validated.

This PR has associated cloud work in its own PR.

Until cloud is deployed, the `validate` query param will be ignored, so this is backwards compatible. It's also ok if cloud is deployed before this is released—either order works.


src/globalConfig/accounts.ts
```diff
@@ -65,17 +102,32 @@ export async function checkEmailStatus(): Promise<EmailStatusResult> {
...
     const resp = await fetchWithTimeout(
-      `https://api.promptfoo.app/api/users/status?email=${encodeURIComponent(userEmail)}`,
+      `${host}/api/users/status?email=${encodeURIComponent(userEmail)}${validateParam}`,
       undefined,
-      500,
+      timeout,
     );
     const data = (await resp.json()) as {
-      status: 'ok' | 'exceeded_limit' | 'show_usage_warning';
+      status: 'ok' | 'exceeded_limit' | 'show_usage_warning' | 'risky_email' | 'disposable_email';
```
sklein12: can you put these into an enum and re-use that across the changes?

danenania: Yeah, this was already duplicated in a few places. Agree it's good to refactor

src/app/src/pages/redteam/setup/components/EmailVerificationDialog.tsx
```diff
@@ -66,6 +66,20 @@ export function EmailVerificationDialog({
         setEmailError(result.error || 'Failed to verify email');
         return;
       }
+
+      // Validate the email after saving
+      const statusResult = await checkEmailStatus({ validate: true });
```
faizanminhas, line 64: should we validate before saving?

src/commands/eval.ts
```diff
@@ -389,8 +390,14 @@ export async function doEval(
...
-      await promptForEmailUnverified();
-      await checkEmailStatusOrExit();
+      // Prompt for email until we get a valid one
+      // Other status problems apart from bad emails (like 'exceeded_limit') just log and exit
+      let hasValidEmail = false;
+      while (!hasValidEmail) {
+        const { emailNeedsValidation } = await promptForEmailUnverified();
+        const res = await checkEmailStatusAndMaybeExit({ validate: emailNeedsValidation });
+        hasValidEmail = res === EMAIL_OK_STATUS;
+      }
```
faizanminhas, line 400: Should this be on the redteam command and not eval? Or both? Or extracted into a helper to share between them?

src/globalConfig/accounts.ts
```diff
@@ -53,29 +91,63 @@ interface EmailStatusResult {
...
+      status: NO_EMAIL_STATUS,
       hasEmail: false,
       message: 'Redteam evals require email verification. Please enter your work email:',
     };
   }
 
   try {
+    const validateParam = options?.validate ? '&validate=true' : '';
+    // Use longer timeout when validation is requested
+    const timeout = options?.validate ? 3000 : 500;
```
faizanminhas, line 111: does mailgun have an SLA here? (just curious)

src/commands/eval.ts
danenania: Is there a separate redteam command (in oss)? I think oss redteams go through `eval`?

I looked for places where `promptForEmailUnverified` was already used and this was the only one.

src/globalConfig/accounts.ts
danenania: They have one: https://www.mailgun.com/legal/sla/

Not sure to what extent it would apply to this endpoint (or our billing tier). Users aren't blocked if Mailgun fails though

src/app/src/pages/redteam/setup/components/EmailVerificationDialog.tsx
danenania: yep! fixing, thanks

PR #5930, opened 2025-10-15
feat: Improved error handling in CLI and error logging

sklein12:
Changes:
- Disable console logging while there is a progress bar to improve the UI
- Add verbose and error logs for each command initialization that go into `LOG_DIR` or the default config directory + logs eg: `~/.promptfoo/logs`

### During test generation show the number of failed plugins and error log information
image

### Eval runs - show errors in progress bar text and error log information
image

### DIFF
--- a/src/commands/eval.ts
+++ b/src/commands/eval.ts
@@ -12,7 +12,7 @@ import { getEnvBool, getEnvFloat, getEnvInt } from '../envars';
 import { DEFAULT_MAX_CONCURRENCY, evaluate } from '../evaluator';
 import { checkEmailStatusOrExit, promptForEmailUnverified } from '../globalConfig/accounts';
 import { cloudConfig } from '../globalConfig/cloud';
-import logger, { debugLogPath, errorLogPath, getLogLevel } from '../logger';
+import logger, { getLogLevel } from '../logger';
 import { runDbMigrations } from '../migrate';
 import Eval from '../models/eval';
 import { loadApiProvider } from '../providers/index';
@@ -775,7 +775,7 @@ export async function doEval(
       logger.info(
         '\n' +
           chalk.red.bold(
-            `See ${chalk.greenBright.bold(errorLogPath)} for detailed error logs or ${chalk.greenBright.bold(debugLogPath)} for verbose logs`,
+            `See ${chalk.greenBright.bold(cliState.errorLogPath)} for detailed error logs or ${chalk.greenBright.bold(cliState.debugLogPath)} for verbose logs`,
           ),
       );
     printBorder();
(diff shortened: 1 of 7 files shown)


sklein12:
@coderabbitai do your thing

src/commands/eval.ts
```diff
@@ -776,6 +769,15 @@ export async function doEval(
     if (!Number.isNaN(passRate)) {
       logger.info(chalk.blue.bold(`Pass Rate: ${passRate.toFixed(2)}%`));
     }
+
+    !Number.isNaN(errors) &&
+      errors > 0 &&
+      logger.info(
+        '\n' +
+          chalk.red.bold(
+            `See ${chalk.greenBright.bold(errorLogPath)} for detailed error logs or ${chalk.greenBright.bold(debugLogPath)} for verbose logs`,
```
mldangelo: prefer to write debugLogPath and errorLogPath to cliState

src/util/dateTime.ts
```diff
@@ -0,0 +1,4 @@
+export function getDateTimeForFilename(): string {
```
mldangelo: docstring would go a long way here. we also have src/util/time.ts - consider moving it to there

src/evaluator.ts
mldangelo, line 196: should we also re-enable enableConsoleLogging here?

src/evaluator.ts
```diff
@@ -1409,12 +1416,12 @@ class Evaluator {
...
         // Progress bar update is handled by the manager
         const phase = evalStep.test.options?.runSerially ? 'serial' : 'concurrent';
-        progressBarManager.updateProgress(index, evalStep, phase);
+        progressBarManager.updateProgress(index, evalStep, phase, metrics);
       } else if (ciProgressReporter) {
         // CI progress reporter update
         ciProgressReporter.update(numComplete);
       } else {
-        logger.debug(`Eval #${index + 1} complete (${numComplete} of ${runEvalOptions.length})`);
+        logger.info(`Eval #${index + 1} complete (${numComplete} of ${runEvalOptions.length})`);
```
mldangelo: Are you sure about this change?

mldangelo (commented):
Looks good. Please find a nice place to document it in the site/

src/commands/eval.ts
sklein12: oh good call

faizanminhas:
Is this going to make it harder to know when an eval or redteam errorred out as a user?

sklein12:
> Is this going to make it harder to know when an eval or redteam errorred out as a user?

the idea is that the # of errors is showing as a count instead of the user inferring it errored out by the log messages. I think this is a move in the right direction but we might need to iterate on it.

Issue #5934, opened 2025-10-15
Licensing Restrictions


jeffrose:
Starting with 0.118.15, promptfoo is using a version of claude-agent-sdk that has a proprietary [license](https://github.com/anthropics/claude-agent-sdk-typescript/blob/main/LICENSE.md), which at the time of me creating this issue reads:

> © Anthropic PBC. All rights reserved. Use is subject to Anthropic's [Commercial Terms of Service](https://www.anthropic.com/legal/commercial-terms).

The Commercial Terms of Service link to https://www.anthropic.com/legal/commercial-terms.

While it's true their license has existed since June, it looks like it wasn't being applied to their JavaScript SDK until a recent release.

I am not a lawyer, so I will not speak to the openness of the license when compared to other open-source licenses like MIT, but many enterprises with Legal departments keep an allowlist of licenses for open-source packages and prevent the usage of packages that do not conform. Since claude-agent-sdk is a dependency of promptfoo, it would get blocked by association. While promptfoo cannot dictate licenses to other parties, you should try to minimize the impact to your project by moving the code to an optional plugin, if possible.

typpo:
Thanks for flagging, @jeffrose.  https://github.com/promptfoo/promptfoo/pull/5935 moves this to an optional dependency.

typpo:
This is addressed in the latest version

PR #5935, opened 2025-10-15
chore: move claude-agent-sdk to optionalDependencies

typpo:
not a required install. also addded note on its proprietary license. #5934


danenania:
Ah, when to use peer vs. optional dependencies for providers?

danenania (approved):

PR #5966, opened 2025-10-18
feat(eval): allow transforms to set test results

typpo:
## Summary

- allow provider and test output transforms to explicitly opt in to setting a test result
- validate transform-controlled results and prevent result-shaped model output from bypassing assertions without the opt-in
- expose provider guardrail data to output transforms and preserve accurate assertion request accounting
- document the feature and add a no-credentials working example
- include formatting-only updates in four recently changed `main` files required by the pre-commit hook after the base merge

## Testing

- `npm run f`
- `npm run l`
- `npm run tsc`
- `npm run build`
- `npm test`
- focused backend and frontend transform, provider, schema, model, and config tests
- `npm run local -- eval -c examples/transform-test-control/promptfooconfig.yaml --no-cache -o /tmp/pr5966-example-output.json`
- adversarial local eval confirming parsed result-shaped output still runs assertions without `transformCanSetTestResult`


src/evaluator.ts
```diff
@@ -1218,16 +1265,33 @@ async function gradeRunEvalResponse({
...
-  });
+  const { processedResponse, providerTransformedOutput, transformTestResult } =
+    await transformRunEvalResponse({
+      evalId,
+      prompt,
+      promptIdx,
+      provider,
+      response,
+      test,
+      testIdx,
```
mldangelo-oai, line 1276: [P2] Let opt-in transforms grade blocked responses with no output

`applyRunEvalResponseOutcome` returns through `applyEmptyResponseOutcome` whenever `response.output` is `null` or `undefined`, before this function is reached. As a result, the documented guardrail flow cannot use `context.metadata.guardrails` when a provider represents a block by omitting output. I reproduced this against the clean current-main merge: a provider returned `{ output: undefined, guardrails: { flagged: true, reason: 'blocked' } }` with an opt-in provider transform that would return a passing `testResult`; the transform was called zero times and the row became `No output`. This makes the result depend on whether the provider fabricates placeholder text for a blocked response. Run the outcome-enabled transform before the empty-output short circuit (while preserving legacy handling for ordinary transforms), and add provider/test-transform coverage for `null` and `undefined` outputs with guardrail metadata.

mldangelo-oai (commented):
Reviewed exact head `eec5007f1181a860c50ae6f1b6b29b07627bf91a` in comment-only mode. The feature is not superseded, and a synthetic merge with current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`) is clean. I found one additional P2 in the documented guardrail path, noted inline. Two existing exact-head blockers remain current and were not duplicated: `isTransformReturnWithTestResult` does not validate its optional `output`, so the type predicate accepts values its declared return type rejects; and transform-controlled rows still enter the later `select-best` / `max-score` comparison phase, which can replace the transform result and consume grader calls despite the advertised assertion bypass.

I rechecked the other open threads. `id: https` is a supported HTTP-provider form in this repository, object-spreading `undefined` is defined to produce no properties in the supported JavaScript runtimes, and the older `TransformContext.metadata` thread is outdated; those are not blockers at this head.

Verification: exact-head published checks are 38/38 passing; `git diff --check` passes; the current-main synthetic merge passes TypeScript and 636 focused tests. A focused regression on that merge confirms the new no-output finding: a provider returning `{ output: undefined, guardrails: { flagged: true } }` never invokes its opt-in transform. The branch is 274 commits behind current `main`, so after addressing the three live issues, please refresh without rewriting history and rerun the current CI matrix. No branch edits or pushes were made.

PR #5984, opened 2025-10-21
feat: replace multilingual strategy with global language config

yash2998chhabria:
Migrate multilingual strategy to plugin-level language generation

  Problem

  The multilingual strategy translated tests post-generation, breaking multi-turn strategies (Crescendo, GOAT) and adding LLM call overhead.

  Solution

  Multilingual is now a pure configuration mechanism that lifts languages to global config:

  1. Validator transformation - Extracts languages from multilingual strategy, prepends undefined (English), merges with global language config, then removes multilingual from
  strategies array
  2. Plugin-level generation - Each plugin generates tests directly in target languages; no translation needed
  3. Dual metadata locations - Language stored in both metadata.language (backward compatible) and metadata.modifiers.language (new standard)
  4. Multi-strategy UI - Tests with language modifiers appear in multiple sections (Cross-Language + primary strategy)
  5. Merging behavior - When both global and multilingual are set, languages merge with deduplication

  Changes

  Backend:
  - Validator lifts/merges languages, removes multilingual from strategies
  - Generate command uses transformed config
  - Types support language: string | string[]
  - Multilingual strategy skips translation if language modifiers present

  Frontend:
  - testBelongsToStrategy() enables multi-strategy membership
  - Filter multilingual from strategy dropdown (config-only, doesn't execute)

  Tests:
  - 5 new validator tests (81 total passing)
  - Verified: local, remote, multi-turn strategies; global config; multilingual; merging

  Examples

  Multilingual strategy:
  strategies:
    - multilingual:
        languages: [hi, fr]
    - crescendo
  Equivalent to language: [undefined, hi, fr] - generates tests directly in each language.

  Merging:
  language: de
  strategies:
    - multilingual:
        languages: [hi, fr]
  Result: [de, undefined, hi, fr] - all unique languages from both sources.

  Breaking Changes

  None - fully backward compatible.


src/app/src/pages/eval/components/store.ts
```diff
@@ -100,7 +100,9 @@ function extractUniqueStrategyIds(strategies?: Array<string | { id: string }> |
   const strategyIds =
     strategies?.map((strategy) => (typeof strategy === 'string' ? strategy : strategy.id)) ?? [];
 
-  return Array.from(new Set([...strategyIds, 'basic']));
+  // Filter out 'multilingual' - it's now a pure config mechanism that lifts languages to global config
+  // Tests generated with language modifiers appear in 'basic' (Cross-Language Attack) section
+  return Array.from(new Set([...strategyIds, 'basic'])).filter((id) => id !== 'multilingual');
```
mldangelo: Does this prevent downstream errors? if not we should revert changes to this file.

src/redteam/index.ts
```diff
@@ -802,28 +802,77 @@ export async function synthesize({
...
+      // If plugin has its own language, use that; otherwise use global language
+      const languages = pluginLanguage
+        ? Array.isArray(pluginLanguage)
+          ? pluginLanguage
+          : [pluginLanguage]
+        : Array.isArray(language)
+          ? language
+          : language
+            ? [language]
+            : [undefined];
```
sklein12: can you do this transformation once?

sklein12:
It would be good to report that the language ones got generated too, not sure the best way to integrate it in. But we used to report it 
image

src/app/src/pages/eval/components/store.ts
faizanminhas: I'd recommend not filtering here, since no new multilingual scans will ever be created, meaning this is just tech debt

src/redteam/commands/generate.ts
```diff
@@ -423,7 +423,7 @@ export async function doGenerateRedteam(
   } = await synthesize({
     ...parsedConfig.data,
     purpose: enhancedPurpose,
-    language: config.language,
+    language: parsedConfig.data.language, // Use parsed/transformed config, not original
```
faizanminhas: I think this is redundant since you're doing `...parsedConfig.data` above anyways. I would remove the original key here.

src/redteam/strategies/multilingual.ts
```diff
@@ -507,6 +507,21 @@ export async function addMultilingual(
   injectVar: string,
   config: Record<string, any>,
 ): Promise<TestCase[]> {
+  // Check if tests were already generated with language modifiers
```
faizanminhas, line 535: I would just delete the strategy & all files

src/redteam/index.ts
faizanminhas:
```suggestion
      // If plugin has its own language, use that; otherwise use global language
      const langaugeConfig = pluginLanguage ?? language;
      const langauges = Array.isArray(languageConfig) ? languageConfig : [languageConfig]
```

src/redteam/index.ts
```diff
@@ -802,28 +802,77 @@ export async function synthesize({
...
+
+      logger.debug(
+        `[Language Processing] Plugin: ${plugin.id}, Languages: ${JSON.stringify(languages)}, NumTests per language: ${plugin.numTests}${pluginLanguage ? ' (plugin override)' : ''}`,
+      );
+
+      // Generate tests for each language
+      const allPluginTests: TestCase[] = [];
+      for (const lang of languages) {
```
faizanminhas: what happens if language is undefined?

src/validators/redteam.ts
```diff
@@ -241,6 +247,38 @@ export const RedteamConfigSchema = z
...
+    // MIGRATION: Extract languages from multilingual strategy and merge into global language config
+    // This allows plugin-level language generation to work with multilingual strategy
+    const multilingualStrategy = data.strategies?.find(
+      (s) => (typeof s === 'string' ? s : s.id) === 'multilingual',
+    );
+
+    if (multilingualStrategy && typeof multilingualStrategy !== 'string') {
+      const strategyLanguages = multilingualStrategy.config?.languages;
+
+      if (Array.isArray(strategyLanguages) && strategyLanguages.length > 0) {
```
sklein12, line 267: You can use this strategy without defining your own languages, it uses 3 languages by default. If you just use the default it seems to skip this whole code flow - https://github.com/promptfoo/promptfoo/blob/main/src/redteam/strategies/multilingual.ts#L17

src/redteam/index.ts
```diff
@@ -802,28 +802,77 @@ export async function synthesize({
...
+        const pluginTests = await action({
```
faizanminhas, line 844: an opportunity to make this faster is to map langauges to pluginTests and do a `promise.AllSettled`

src/redteam/index.ts
```diff
@@ -802,28 +802,77 @@ export async function synthesize({
...
+            },
           },
-          ...resolvePluginConfig(plugin.config),
-        },
-      });
+        });
+
+        if (Array.isArray(pluginTests) && pluginTests.length > 0) {
+          // Immediately add language to each test's metadata so strategies can access it
+          const testsWithLanguage = pluginTests.map((test) => ({
```
faizanminhas: you may be able to pass this metadata into the action above and not have to loop twice

sklein12:
When the language strategy is removed these numbers aren't updated and there's no indication that we're still doing the multi-lingual testing

image

src/redteam/index.ts
faizanminhas: consider looking at the plugin factory

src/redteam/index.ts
faizanminhas: maybe move the metadata mapping here instead?

src/redteam/index.ts
```diff
@@ -802,28 +802,77 @@ export async function synthesize({
...
+            ...test,
+            metadata: {
+              ...test.metadata,
+              // Hoist language to top-level metadata for backward compatibility and easier access
+              ...(lang && !test.metadata?.language ? { language: lang } : {}),
+              modifiers: {
+                ...test.metadata?.modifiers,
+                // Add language to modifiers if not already present (respect existing)
+                ...(lang && !test.metadata?.modifiers?.language ? { language: lang } : {}),
```
faizanminhas: write a helper like `getLanguageForTestCase`

sklein12: could you move this into a helper function? I think it would be good to clean up this code and with a good name it can be descriptive about what it's doing.

eg: `function addLanguageToPluginMedata`

src/validators/redteam.ts
```diff
@@ -241,6 +247,38 @@ export const RedteamConfigSchema = z
     const pluginMap = new Map<string, RedteamPluginObject>();
     const strategySet = new Set<Strategy>();
 
+    // MIGRATION: Extract languages from multilingual strategy and merge into global language config
+    // This allows plugin-level language generation to work with multilingual strategy
+    const multilingualStrategy = data.strategies?.find(
```
faizanminhas, line 260: if multilingual is being removed, do we need this code still?

faizanminhas: nvm: we are making it backwards compatible.

src/validators/redteam.ts
```diff
@@ -241,6 +247,38 @@ export const RedteamConfigSchema = z
...
+      (s) => (typeof s === 'string' ? s : s.id) === 'multilingual',
+    );
+
+    if (multilingualStrategy && typeof multilingualStrategy !== 'string') {
+      const strategyLanguages = multilingualStrategy.config?.languages;
+
+      if (Array.isArray(strategyLanguages) && strategyLanguages.length > 0) {
+        // Include 'en' to represent original English tests explicitly
+        // ['en', 'hi', 'fr'] means: generate English + Hindi + French
```
faizanminhas: pointless comment imho

faizanminhas (commented):
Address the comments and I'll approve. Good job!

sklein12:
This causes a change in the UI where languages no longer show up as a metric at the top or as a separate strategy in the Report. Did you and Michael discuss this change?

image
image

src/validators/redteam.ts
yash2998chhabria: Nice, makes sense! Ill set a default

src/redteam/index.ts
yash2998chhabria: will use default, en

site/docs/_shared/data/strategies.ts
```diff
@@ -338,7 +327,7 @@ export const strategies: Strategy[] = [
     displayName: 'Layer',
     description: 'Compose multiple strategies',
     longDescription:
-      'Compose multiple red team strategies sequentially (e.g., multilingual → rot13) to create sophisticated attack chains',
+      'Compose multiple red team strategies sequentially (e.g., base64 → rot13) to create sophisticated attack chains',
```
mldangelo: pick a different one

site/docs/red-team/plugins/index.md
```diff
@@ -177,10 +177,10 @@ plugins:
 
 ### Setting Languages
 
-You can set a top-level language that affects all plugins. You can also implement the [multi-lingual strategy](/docs/red-team/strategies/multilingual) to generate probes in specified languages.
+You can set a top-level language that affects all plugins. For more information, see the [language configuration](/docs/red-team/configuration/#language) section.
 
 ```yaml
-language: 'Spanish' # Top level language setting
+language: ['Spanish', 'French', 'German'] # Test in multiple languages
```
mldangelo, line 183:
```suggestion
language: ['Spanish', 'French', 'German'] # Test in multiple languages. English is included by default. 
```

site/docs/red-team/strategies/basic.md
```diff
@@ -20,14 +20,13 @@ Example usage:
...
+  language: ['es', 'fr'] # Test in Spanish and French
   strategies:
     - id: basic
       config:
         enabled: false # Only run tests with strategies applied
     - id: jailbreak
-    - id: multilingual
-      config:
-        languages: ['es', 'fr']
+    - id: base64
```
mldangelo, line 29:
```suggestion
```

site/docs/red-team/strategies/layer.md
```diff
@@ -1,12 +1,12 @@
 ---
 title: Layer Strategy
-description: Compose multiple red team strategies in order (e.g., multilingual → rot13) to stack transformations and agentic techniques
+description: Compose multiple red team strategies in order (e.g., base64 → rot13) to stack transformations and agentic techniques
```
mldangelo, line 3: something like jailbreak would be better here

site/docs/red-team/strategies/layer.md
```diff
@@ -18,8 +18,8 @@ redteam:
     - id: layer
       config:
         steps:
-          - multilingual # First translate
-          - rot13 # Then encode
+          - base64 # First encode as base64
```
mldangelo, line 21: jailbreak here

site/docs/red-team/strategies/multilingual.md
```diff
@@ -1,130 +0,0 @@
----
-title: Multilingual Jailbreaking Strategy
```
mldangelo, line 2: Recommend keeping this file but describing the configuration in it instead. Write it as evergreen documentation.

site/docs/red-team/configuration.md
```diff
@@ -535,15 +535,42 @@ redteam:
 
 ### Language
 
-The `language` field allows you to specify the language for generated tests. If not provided, the default language is English. This can be useful for testing your model's behavior in different languages or for generating adversarial inputs in specific languages.
+The `language` field allows you to specify the language(s) for generated tests. If not provided, the default language is English. This setting applies globally to all plugins and strategies, ensuring consistent multilingual testing across your entire red team evaluation.
 
-Example usage:
+#### Single Language
+
+Test your application in a specific language:
```
mldangelo, line 542: You changed the behavior and should talk about how english is default and this is for an additional language

src/app/src/pages/redteam/setup/components/strategies/utils.ts
```diff
@@ -53,18 +52,14 @@ export function getEstimatedProbes(config: Config) {
...
-  // Find if multilingual strategy is present and get number of languages
-  const multilingualStrategy = config.strategies.find(
-    (s) => (typeof s === 'string' ? s : s.id) === 'multilingual',
-  );
-
-  const numLanguages =
-    multilingualStrategy && typeof multilingualStrategy !== 'string'
-      ? ((multilingualStrategy.config?.languages as string[]) || []).length || 3
+  // Get number of languages from global language config
+  const numLanguages = Array.isArray(config.language)
```
mldangelo, line 60: Isn't this + 1 now?

src/app/src/pages/redteam/setup/components/Purpose.tsx
```diff
@@ -1131,6 +1132,66 @@ export default function Purpose({ onNext, onBack }: PromptsProps) {
                   />
                 </Box>
               </Box>
+
+              {/* Languages Configuration - Standalone Section */}
```
mldangelo: @faizanminhas would like your eyes on this

src/redteam/strategies/multilingual.ts
```diff
@@ -1,617 +0,0 @@
-import async from 'async';
```
mldangelo, line 1: instead of deleting this completely we should add a deprecation notice because we still need it in cloud. Alternatively, you can port the entire implementation to cloud

CHANGELOG.md
```diff
@@ -79,6 +80,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - docs(model-audit): improve accuracy and clarity of ModelAudit documentation (#6023)
 - docs(contributing): add changelog and GitHub Actions enforcement (#6012)
+- docs(redteam): add global language configuration section to red team configuration docs; remove multilingual strategy documentation (#5984)
```
mldangelo, line 111: only one entry

site/docs/red-team/configuration.md
yash2998chhabria: will not be required, switching the behavior back to how it was

site/docs/red-team/plugins/index.md
yash2998chhabria: editing the behavior to not include english as default

src/app/src/pages/redteam/setup/components/strategies/utils.ts
yash2998chhabria: switching it back to old behaviour!

src/redteam/strategies/multilingual.ts
yash2998chhabria: Sounds good, adding it back with deprecation notice

site/docs/red-team/mitre-atlas.md
```diff
@@ -70,13 +70,14 @@ Example configuration:
 
 ```yaml
 redteam:
+  language: ['en', 'es', 'fr'] # Test in multiple languages
```
sklein12, line 73: should we maintain the low level languages for the example?

test/redteam/integration/test.sh
```diff
@@ -16,10 +16,10 @@ else
 fi
 
 # Check for expected output patterns
-if echo "$output" | grep -q "Wrote 2[34] test cases to redteam.yaml"; then
-  echo "Expected line (Wrote 2[34] test cases to redteam.yaml) found for promptfoo redteam generate"
+if echo "$output" | grep -q "Wrote 12 test cases to redteam.yaml"; then
+  echo "Expected line (Wrote 12 test cases to redteam.yaml) found for promptfoo redteam generate"
 else
-  echo "ERROR: Expected line (Wrote 2[34] test cases to redteam.yaml) not found in output."
+  echo "ERROR: Expected line (Wrote 12 test cases to redteam.yaml) not found in output."
```
sklein12, line 22: Can you update the configuration to use the new language config instead?

src/app/src/pages/redteam/setup/components/Purpose.tsx
sklein12: I think this should go inside `Run Options` on the Review page. Up for other suggestions as well. I don't think this is the right place for it. Application Details is for input from the user about how the system acts, this is more a modifier to our tests.

sklein12 (changes requested):

src/app/src/pages/redteam/setup/components/Purpose.tsx
faizanminhas: I was thinking about this- `testGenerationInstructions` is also right above this section. Should that move somewhere else? Otherwise, `What languages does your application support` seems like a reasonable thing in application details

yash2998chhabria: moved to Run Options, I agree with @faizanminhas about test Generation Instructions. Should I include it in this PR ?

site/docs/red-team/mitre-atlas.md
yash2998chhabria: Yeah , I dont see why not

src/app/src/pages/redteam/setup/components/Purpose.tsx
yash2998chhabria: [
Screenshot 2025-10-28 at 3 57 57 PM
](url)

PR #5992, opened 2025-10-21
feat: Error logging and messaging

sklein12:
- Capture all errors into an error file and print a helpful message after any command if there was an error

image

- Adds env var to turn off debug or error logs for those that don't want them

### DIFF
--- a/test/logger.test.ts
+++ b/test/logger.test.ts
@@ -1,3 +1,4 @@
+import path from 'path';
 import chalk from 'chalk';
 import type { Logger } from 'winston';
 import type Transport from 'winston-transport';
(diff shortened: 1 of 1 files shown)


src/redteam/commands/generate.ts
```diff
@@ -29,8 +29,8 @@ import { resolveConfigs } from '../../util/config/load';
 import { writePromptfooConfig } from '../../util/config/writer';
 import { getCustomPolicies } from '../../util/generation';
 import { printBorder, setupEnv } from '../../util/index';
-import { promptfooCommand } from '../../util/promptfooCommand';
 import invariant from '../../util/invariant';
+import { promptfooCommand } from '../../util/promptfooCommand';
```
mldangelo, line 32: can you fix whatever is causing these file changes?  Happy to change the import linter rule to make this better. You also modified src/commands/eval.ts above

src/envars.ts
```diff
@@ -22,6 +22,8 @@ type EnvVars = {
   PROMPTFOO_DISABLE_AJV_STRICT_MODE?: boolean;
   PROMPTFOO_DISABLE_CONVERSATION_VAR?: boolean;
   PROMPTFOO_DISABLE_ERROR_LOG?: boolean;
+  PROMPTFOO_DISABLE_DEBUG_LOG?: boolean;
+
```
mldangelo:
```suggestion
```

PR #6037, opened 2025-10-27
chore(internals): Updates custom policy type

will-holley:


src/app/src/pages/eval/components/CustomMetrics.tsx
```diff
@@ -135,7 +135,7 @@ const CustomMetrics = ({
             const policyId = deserializePolicyIdFromMetric(metric);
             const policy = policiesById[policyId];
             if (policy) {
-              displayLabel = formatPolicyIdentifierAsMetric(policy.name);
+              displayLabel = formatPolicyIdentifierAsMetric(policy.name!);
```
sklein12, line 138: We should have proper fallbacks here since this isn't typed as required anymore

PR #6038, opened 2025-10-27
chore(redteam): improve Simba command error formatting

sklein12:


sklein12:
@codex can you update the change log?

PR #6039, opened 2025-10-27
chore: bump version 0.119.0

sklein12:


mldangelo (commented):
update the changelog too!

sklein12:
@mldangelo already on it! :) is that how it's supposed to work?

PR #6071, opened 2025-10-31
fix(cli): correct port type handling in view command

iitslamaa:
## Summary

Fixes type safety issue with the `--port` option in `promptfoo view` command.

## Changes

- **Port type fix**: Changed port option to properly parse string input to number using `parseInt`
- Previously, Commander returned port as string but the type definition declared it as number
- This works in practice due to JavaScript type coercion, but is a type safety bug

## Before
```typescript
.option('-p, --port <number>', 'Port number', getDefaultPort().toString())
// Returns string "15500", type says number
```

## After
```typescript
.option(
  '-p, --port <number>',
  'Port number',
  (val) => Number.parseInt(val, 10),  // Parse to number
  getDefaultPort(),                    // Return number directly
)
```

## Testing

- ✅ Tested with custom port: `promptfoo view --port 19999`
- ✅ Tested with default port: `promptfoo view`
- ✅ Server starts correctly and binds to specified port
- ✅ All existing functionality preserved

## Notes

This is a minor type safety fix that doesn't change user-facing behavior. The `--no` flag continues to work as before for headless/CI workflows.

### DIFF
--- a/test/commands/view.test.ts
+++ b/test/commands/view.test.ts
@@ -29,7 +29,7 @@ describe('viewCommand', () => {
 
     const options = viewCmd.opts();
     expect(options).toEqual({
-      port: getDefaultPort().toString(),
+      port: getDefaultPort(),
     });
   });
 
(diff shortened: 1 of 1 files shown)


iitslamaa:
hey @mldangelo — this is ready. if alls good, i can update the docs to include it too.

PR #6089, opened 2025-11-01
fix(util): handle Nunjucks template variables in URL sanitization

mldangelo:
## Summary

Fixes URL sanitization errors when sharing evaluations that use Nunjucks template variables in provider URLs (e.g., `{{ api_base }}/api/endpoint`).

## Problem

In version 0.119.0, URL sanitization was added to redact sensitive information from logs. However, the sanitizer attempted to parse configuration URLs containing Nunjucks template variables like `{{ api_base }}`, which are not valid URLs until rendered. This caused warning messages during `promptfoo share`:

```
Failed to sanitize URL {{ api_base }}/api/[redacted]
```

While sharing still worked, the error messages were confusing and polluted logs.

## Solution

Added a regex check to detect Nunjucks template variables (`{{ ... }}`) before attempting URL parsing. If template variables are present, the URL is returned unchanged since:

1. Template URLs are configuration-time values, not runtime URLs
2. Nunjucks renders them to real URLs before use
3. The rendered URLs get sanitized when actually used
4. URL parsing would either fail (templates in hostname) or URL-encode the template syntax (breaking Nunjucks)

## Technical Details

Testing revealed that `new URL()` behavior with templates:
- `{{ api_base }}/api` → Parse fails (invalid scheme)  
- `https://example.com/{{ path }}` → Parses but URL-encodes to `%7B%7B`, breaking template syntax

The fix skips sanitization for any URL containing template variables, preserving them for Nunjucks rendering.

## Testing

Added 8 test cases covering:
- Templates in various positions (hostname, path, query params)
- Multiple template variables
- Mixed template and non-template content

All existing sanitizer tests (163 total) continue to pass.

## Trade-offs

**Security consideration**: URLs with both template variables AND sensitive non-template parameters (e.g., `https://example.com/{{ path }}?api_key=secret`) will skip sanitization entirely. This is acceptable because:
- Template URLs come from configuration files (version-controlled)
- Secrets should be in environment variables, not config files
- Alternative (URL-encoding templates) breaks Nunjucks syntax

Fixes issue reported by @dima in Slack (CLI sharing with double-brace notation).

### DIFF
--- a/.github/workflows/changelog-check.yml
+++ b/.github/workflows/changelog-check.yml
@@ -13,7 +13,7 @@ jobs:
       pull-requests: read
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
--- a/.github/workflows/deploy-launcher.yml
+++ b/.github/workflows/deploy-launcher.yml
@@ -21,7 +21,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Use Node
-        uses: actions/setup-node@v5
+        uses: actions/setup-node@v6
         with:
           node-version-file: '.nvmrc'
           cache: 'npm'
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -212,7 +212,7 @@ jobs:
           touch "${{ runner.temp }}/digests/${digest#sha256:}"
 
       - name: Upload digest
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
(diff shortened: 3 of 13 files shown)


mldangelo:
Fixed CodeQL ReDoS vulnerability by replacing `/\{\{.*?\}\}/` with `/\{\{[^}]*\}\}/`. The new regex:
- Uses character class `[^}]*` instead of `.*?` to prevent exponential backtracking
- Runs in O(n) linear time instead of potentially exponential
- Still correctly detects all Nunjucks template variables
- All 163 sanitizer tests pass

The old regex could experience performance degradation with malicious input like many opening braces without closing pairs.

PR #6202, opened 2025-11-13
feat(assertions): add dot product and euclidean distance metrics

typpo:


site/docs/configuration/expected-outputs/similar.md
```diff
@@ -32,6 +32,57 @@ assert:
     threshold: 0.8
 ```
 
+## Similarity Metrics
+
+You can specify which metric to use by including it in the assertion type. The default is `similar` (cosine similarity).
+
+### Cosine Similarity (default)
+
+Measures the cosine of the angle between two vectors. Range: 0 to 1 (higher is more similar).
```
mldangelo: Do you mean -1 to 1?

PR #6207, opened 2025-11-13
docs: use markdown style code fencing to get copy button

JustinBeckwith:
I was following https://www.promptfoo.dev/docs/getting-started/, and noticed something weird - some of the `npx` commands didn't have a Copy button in the code snippet.   But others did - weird, right?  Turns out it's an artifact of using the React syntax instead of triple tick code fencing.  

Turns out, we did this in a lot of places:

```tsx
<CodeBlock language="bash">
   npx promptfoo@latest init --example getting-started
</CodeBlock>
```

But if you're using the React syntax block for these, you need to include the code as an object lol:

```tsx
<CodeBlock language="bash">
   {'npx promptfoo@latest init --example getting-started'}
</CodeBlock>
```

My 2 cents - this is SUPER not intuitive, so I took the liberty of converting these to triple ticks.  Now the copy button works!  If you'd rather go back to the React style blocks, that's totally cool, just let me know!

### DIFF
--- a/site/docs/getting-started.md
+++ b/site/docs/getting-started.md
@@ -359,8 +359,7 @@ You can quickly set up this example by running:
 <Tabs groupId="promptfoo-command">
   <TabItem value="npx" label="npx" default>
     ```bash
-    npx promptfoo@latest init --example 
-    openai-model-comparison
+    npx promptfoo@latest init --example openai-model-comparison
     ```
   </TabItem>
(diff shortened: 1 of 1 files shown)


mldangelo (approved):
Thank you!

JustinBeckwith:
@mldangelo any idea why CodeQL seems to be wedged?

PR #6249, opened 2025-11-19
chore(deps): re-generate lock file

JustinBeckwith:
I am struggling with the current lock file.  I think there's some sort of compat issue between npm versions 10 and 11 (node 22 vs 24), and it's been biting me in other PRs where I happen to modify package.json.  I am re-generating it in hopes that this stops happening.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -531,9 +531,9 @@
       "license": "MIT"
     },
     "node_modules/@ai-sdk/gateway": {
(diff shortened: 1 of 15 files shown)


mldangelo (approved):
lock file changes are good! 

Curious about where your formatter is different than ours.

JustinBeckwith:
Looks like it was biome 2.3.5 to 2.3.6

PR #6343, opened 2025-11-24
fix(server): exclude errors from pass rate calculation in eval summaries

minhle1291:
## Summary

This PR fixes the pass rate calculation in evaluation summaries to exclude errors from the denominator, providing more accurate percentage reporting in the evaluation results list view.

**Previously:** `passRate = passes / (passes + fails + errors)`  
**Now:** `passRate = passes / (passes + fails)`

This change makes the pass rate a more accurate measure of output quality by only considering valid test results (passes and failures) while excluding errors caused by API failures, timeouts, or other system issues.

## Changes

- **[src/models/eval.ts:1301-1305](https://github.com/promptfoo/promptfoo/blob/fix/server-pass-rate-exclude-errors/src/models/eval.ts#L1301-L1305)**: Modified `testRunCount` calculation to sum only passes and fails, excluding errors
- **[test/models/eval.test.ts:150-191](https://github.com/promptfoo/promptfoo/blob/fix/server-pass-rate-exclude-errors/test/models/eval.test.ts#L150-L191)**: Added comprehensive test case verifying the fix
- **CHANGELOG.md**: Added entry documenting the fix

## Test Plan

- [x] Added new test case: "should exclude errors from pass rate calculation"
  - Verifies with 2 passes, 3 fails, 5 errors → pass rate is 40% (2/5), not 20% (2/10)
  - Confirms `numTests` still includes errors (total count = 10)
- [x] All existing tests pass (46 tests in eval.test.ts)
- [x] Code linted and formatted

## Example

For an evaluation with:
- 2 passing tests
- 3 failing tests  
- 5 errors

**Before this PR:** Pass rate = 2 / 10 = 20%  
**After this PR:** Pass rate = 2 / 5 = 40%

The 5 errors are excluded from the denominator calculation, making the pass rate reflect the quality of valid results only.

### DIFF
--- a/src/commands/show.ts
+++ b/src/commands/show.ts
@@ -34,13 +34,10 @@ export async function handlePrompt(id: string) {
       'Dataset ID': evl.datasetId.slice(0, 6),
       'Raw score': evl.metrics?.score?.toFixed(2) || '-',
       'Pass rate':
-        evl.metrics &&
-        evl.metrics.testPassCount + evl.metrics.testFailCount + evl.metrics.testErrorCount > 0
+        evl.metrics && evl.metrics.testPassCount + evl.metrics.testFailCount > 0
           ? `${(
               (evl.metrics.testPassCount /
-                (evl.metrics.testPassCount +
-                  evl.metrics.testFailCount +
-                  evl.metrics.testErrorCount)) *
+                (evl.metrics.testPassCount + evl.metrics.testFailCount)) *
                 100
             ).toFixed(2)}%`
           : '-',
--- a/test/models/eval.test.ts
+++ b/test/models/eval.test.ts
@@ -190,6 +190,48 @@ describe('evaluator', () => {
       // Total test count should still include errors
       expect(summary?.numTests).toBe(10);
     });
+
+    it('should handle all errors scenario with zero pass rate', async () => {
+      // Create an eval with only errors: 0 pass, 0 fail, 10 errors
+      const evaluation = await Eval.create(
+        {
+          providers: ['test-provider'],
+          prompts: ['Test prompt'],
+          tests: [{ vars: { test: 'value' } }],
+        },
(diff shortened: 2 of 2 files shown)


minhle1291:
@MrFlounder This PR applies [your changes to promptfoo-cloud](https://github.com/promptfoo/promptfoo-cloud/pull/2478/files) to the OSS

MrFlounder:
do you mind adding a before and after screenshot?

minhle1291:
> This changes ASR values significantly for redteam evals. Using the example from the PR (2 pass, 3 fail, 5 errors):
> 
> Before: ASR = 3/10 = 30%
> After: ASR = 3/5 = 60%
> Was this intentional? If so, worth mentioning in the PR description since it affects redteam users.

I believe this is intentional, and synchronizes with the updated logic in [server/src/models/eval.ts](https://github.com/promptfoo/promptfoo-cloud/pull/2478/files)

> show.ts Inconsistency
> This is less important than 1. The CLI command promptfoo show prompt still calculates pass rate with errors included:
> 
> // src/commands/show.ts:38-45
> passRate = testPassCount / (testPassCount + testFailCount + testErrorCount) * 100
> Should this be updated for consistency? Or is there a reason to keep them different?

Ah, good catch. The `promptfoo show prompt` interface is not one I was cognizant of. I will tackle that.

> Re. unit tests, the current test covers the single-prompt case well. A few edge cases that might be worth adding:
> 
> Multiple prompts with different error counts (validates the reduce aggregation)
> All errors scenario (testRunCount = 0, verifies division-by-zero handling)
> Explicit check that attackSuccessRate changes as expected (if intentional)

I'll add these edge cases in as well.

mldangelo:
We discussed this IRL and decided to not implement this feature. Thank you @minhle1291!

PR #6344, opened 2025-11-24
feat(eval): add dynamic timeout defaults

mldangelo:
## Summary

- Add dynamic per-test and total evaluation timeout defaults, synchronized with current `main`.
- Enforce total deadlines through setup, target execution, grading, and teardown, including providers or graders that do not cooperate with abort signals.
- Preserve redteam timeout headroom for generated/exported cases that do not retain a top-level `redteam` block.
- Keep pause/resume correct by not persisting in-flight rows cancelled by the evaluator's external abort signal.

## Design

- Standard per-test timeouts default to `5 * REQUEST_TIMEOUT_MS`; redteam cases default to `15 * REQUEST_TIMEOUT_MS` for multi-turn attack/target workflows.
- Auto total duration accounts for serial work and comparison grading, and cannot undercut one active per-test deadline unless the user supplies an explicit smaller max.
- Explicit total duration starts during setup. If setup expires, declared test rows are persisted as max-duration errors rather than disappearing behind a thrown setup abort.
- Total-duration enforcement races active provider calls, grouped grading, comparison grading, and `afterAll` hooks against the deadline, so `timeoutMs: 0` does not allow non-cooperative work to hang the evaluation.
- Durations above Node's single-timer ceiling are scheduled in safe chunks while preserving the configured deadline.
- Per-test timeout enforcement disables grouped local-judge grading; setting `PROMPTFOO_EVAL_TIMEOUT_MS=0` explicitly opts into grouped grading.

## Review Follow-up

- Addressed all current inline review findings: setup-timeout persistence, exported redteam default classification, and hard bounds for non-cooperative providers/graders.
- Retained earlier fixes for effective `options.timeoutMs`, serial sizing, comparison timeouts, and grouped-grading safety.
- `CHANGELOG.md` is generated under repository policy and was intentionally not edited manually.

## Validation

- `npm run f`
- `npm run l`
- `npm run jsonSchema:generate` (no additional generated drift)
- `npm run build`
- `npx vitest run test/evaluator/execution.test.ts test/evaluator/gradingConcurrency.test.ts --sequence.shuffle=false` (`54` tests passed)
- `npx vitest run --coverage test/evaluator test/envars.test.ts test/util/namedMetrics.test.ts --sequence.shuffle=false --maxWorkers 4` (`506` tests passed)
- `npm run test:coverage:ratchet -- --report backend`
- `npx vitest run --config vitest.smoke.config.ts test/smoke/resume.test.ts --sequence.shuffle=false` (`6` tests passed)
- Isolated CLI hard-deadline QA: a never-settling provider with `maxEvalTimeMs: 75` and `timeoutMs: 0` returned two max-duration errors in `79ms` instead of hanging.
- Isolated exported-redteam CLI QA: a generated-style case without top-level `redteam` passed `1/1` while logging the redteam budgets (`per-test=4500000ms`, `max=4560000ms`).

## Local Full-Suite Note

A broad local backend coverage attempt exhausted the macOS ephemeral port range while running untouched HTTP route suites together. Each affected route suite passed when rerun independently after the socket pool drained. Exact-head GitHub CI is the authoritative full-matrix validation and is being watched to completion.

## Database / Migrations

- This PR adds no migration or database implementation files.
- All CLI QA used fresh isolated config/cache/log directories because current `main` contains a migration; no user database or shared cache was modified.


site/docs/usage/troubleshooting.md
```diff
@@ -260,14 +260,28 @@ When running evals, you may encounter timeout errors, especially when using loca
...
-You can control two settings: timeout for individual test cases and timeout for the entire evaluation.
+### Default timeout behavior
 
-### Quick fixes
+Promptfoo provides layered timeout protection by default:
 
-**Set timeouts for individual requests and total evaluation time:**
+**Per-test timeout (`PROMPTFOO_EVAL_TIMEOUT_MS`):**
+
+- Defaults to 5× the HTTP request timeout (25 minutes with default settings)
```
jameshiester: given Meta has default 10 rounds is 5 the most logical default?  Maybe something more conservative like 15 to account for more multi-turn strategies?

test/envars.test.ts
```diff
@@ -385,4 +394,167 @@ describe('envars', () => {
...
+      // 5 * 60,000 = 300,000ms (5 minutes)
+      expect(getEvalTimeoutMs()).toBe(300_000);
+    });
+
+    it('should return explicit PROMPTFOO_EVAL_TIMEOUT_MS value when set', () => {
+      process.env.PROMPTFOO_EVAL_TIMEOUT_MS = '120000';
+      expect(getEvalTimeoutMs()).toBe(120000);
+    });
+
+    it('should return 0 when PROMPTFOO_EVAL_TIMEOUT_MS is explicitly set to 0', () => {
```
jameshiester, line 517: can you add test case for undefined env value?

test/evaluator.test.ts
```diff
@@ -337,7 +337,8 @@ describe('evaluator', () => {
       ],
     };
     const evalRecord = await Eval.create({}, testSuite.prompts, { id: randomUUID() });
-    await evaluate(testSuite, evalRecord, {});
+    // Disable timeouts to avoid abort signal in test
+    await evaluate(testSuite, evalRecord, { timeoutMs: 0, maxEvalTimeMs: 0 });
```
jameshiester: should there be a test case that validates maxEvalTimeMs is working as expected?

src/evaluator.ts
```diff
@@ -725,7 +725,28 @@ class Evaluator {
...
+      (options.repeat || 1) *
+      testSuite.providers.length *
+      testSuite.prompts.length;
+    const maxConcurrency = options.maxConcurrency || DEFAULT_MAX_CONCURRENCY;
+    const testCaseTimeoutMs = getEvalTimeoutMs();
+    const isRedteam = testSuite.redteam != null;
+
+    const maxEvalTimeMs =
+      options.maxEvalTimeMs ??
+      getDefaultMaxEvalTimeMs(estimatedTotalSteps, maxConcurrency, testCaseTimeoutMs, isRedteam);
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed on the current head: global sizing now uses the effective `options.timeoutMs`, including explicit `0`, through the shared timeout configuration.

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Added
 
+- feat(eval): add dynamic timeout defaults with layered protection - test case timeout defaults to 5× REQUEST_TIMEOUT_MS (25 minutes) to allow for retries and multi-turn redteam strategies; total eval timeout auto-calculated based on eval size, concurrency, and redteam detection; set either to 0 to disable
```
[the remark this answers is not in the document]
mldangelo-oai: The branch is synchronized with current `main`; `CHANGELOG.md` is generated under repository policy and is intentionally not edited manually for this PR.

site/docs/usage/troubleshooting.md
mldangelo-oai: Addressed: standard evals retain the 5x default, while redteam evals now use 15x `REQUEST_TIMEOUT_MS` so multi-turn strategies have appropriate headroom. Documentation and tests cover both cases.

test/envars.test.ts
mldangelo-oai: Added coverage for empty environment inputs: empty `PROMPTFOO_EVAL_TIMEOUT_MS` and `PROMPTFOO_MAX_EVAL_TIME_MS` now explicitly exercise calculated defaults.

test/evaluator.test.ts
mldangelo-oai: Added and retained max-duration execution coverage, including total-time abort behavior and persisted comparison-timeout error results.

src/evaluator.ts
```diff
@@ -4178,29 +4238,15 @@ class Evaluator {
     let progressBarManager: ProgressBarManager | null = null;
 
     // Create abort signals:
-    // - providerAbortSignal: passed to providers (user signal + timeout, but NOT target error)
+    // - providerAbortSignal: passed to providers (user signal, but NOT target error)
     // - combinedAbortSignal: used internally for checkAbort (includes target error signal)
     // Target error signal is not passed to providers because by the time we detect a 403 etc,
     // the provider call has already completed - it's only used to stop the evaluator loop.
-    let providerAbortSignal: AbortSignal | undefined = options.abortSignal;
+    const providerAbortSignal: AbortSignal | undefined = options.abortSignal;
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed: `configureEvaluationTimeout` returns a provider abort signal combined with the global controller and comparison assertions consume it; a `select-best` regression test verifies cancellation.

src/evaluator.ts
```diff
@@ -4178,29 +4238,15 @@ class Evaluator {
...
     let combinedAbortSignal: AbortSignal = options.abortSignal
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in the current head: comparison assertion provider calls receive the combined global abort signal, with regression coverage for a global timeout during `select-best` grading.

src/evaluator.ts
```diff
@@ -2723,6 +2723,69 @@ function createTimeoutMetrics(timeoutMs: number): PromptMetrics {
...
+  isEvalTimedOut: () => boolean;
+  maxEvalTimeMs: number;
+  providerAbortSignal?: AbortSignal;
+} {
+  const testCaseTimeoutMs =
+    options.timeoutMs === undefined ? getEvalTimeoutMs() : options.timeoutMs;
+  const isRedteamEval = options.isRedteam ?? Boolean(testSuite.redteam);
+  const maxEvalTimeMs =
+    options.maxEvalTimeMs ??
+    getDefaultMaxEvalTimeMs(runEvalOptions.length, concurrency, testCaseTimeoutMs, isRedteamEval);
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed: automatic max duration now counts `runSerially` steps outside concurrent batches, with helper and evaluator regression tests.

src/evaluator.ts
```diff
@@ -4339,7 +4404,9 @@ class Evaluator {
         concurrentRunEvalOptions.push(evalOption);
       }
     }
-    const hasEvalStepTimeout = (options.timeoutMs || getEvalTimeoutMs()) > 0;
+    const explicitTestCaseTimeoutMs = options.timeoutMs ?? getEnvInt('PROMPTFOO_EVAL_TIMEOUT_MS');
+    const hasEvalStepTimeout =
+      explicitTestCaseTimeoutMs !== undefined && explicitTestCaseTimeoutMs > 0;
     const shouldGroupGradingByProvider =
       concurrency === 1 && !hasEvalStepTimeout && !usesConversationVar;
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed: grouped grading is now gated on the effective per-step timeout, including its implicit default. The docs explain that setting `PROMPTFOO_EVAL_TIMEOUT_MS=0` intentionally opts into grouped grading.

src/evaluator.ts
```diff
@@ -4430,23 +4497,32 @@ class Evaluator {
...
+          rowsWithSelectBestAssertion,
+          runEvalOptions,
+        });
+      }
+    } catch (err) {
+      if (!timeoutConfig.isEvalTimedOut()) {
+        throw err;
+      }
+      logger.warn(`Evaluation stopped after reaching max duration (${maxEvalTimeMs}ms)`);
+    }
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed: pending `select-best` and `max-score` rows are transitioned to timeout errors when the global deadline fires during comparison grading; persisted regression coverage asserts the resulting rows and counters.

src/envars.ts
```diff
@@ -520,26 +520,174 @@ export function getEnvFloat(key: EnvVarKey, defaultValue?: number): number | und
...
+
+  if (testCaseTimeoutMs <= 0) {
+    return safeMaxTime;
+  }
+
+  // Cap at worst case (all tests timeout), plus the same setup/teardown buffer used
+  // for expected runtime so max-duration does not race an individual per-test timeout.
+  const worstCase = batchCount * testCaseTimeoutMs + MAX_EVAL_TIME_BUFFER_MS;
+
+  return Math.min(safeMaxTime, worstCase);
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed: the computed total timeout cannot be shorter than one effective per-test timeout plus its buffer unless a smaller total timeout is explicitly configured; unit coverage was added.

mldangelo-oai:
@codex review

mldangelo-oai:
Updated and pushed `e289f060d` after merging current `main`.

What changed:
- Started explicit max-eval timers before setup hooks/generated prompts/run option construction, so user-supplied total limits cover setup work too.
- Included pending `select-best`/`max-score` comparison grading rows in automatic total-timeout sizing.
- Added regression coverage for both review findings.

Validation run locally:
- `npx vitest run test/envars.test.ts test/evaluator/execution.test.ts --sequence.shuffle=false`
- `npx vitest run test/envars.test.ts test/evaluator/basic.test.ts test/evaluator/execution.test.ts test/evaluator/gradingConcurrency.test.ts test/evaluator/metadata.test.ts test/evaluator/options.test.ts test/util/namedMetrics.test.ts --sequence.shuffle=false`
- `npx vitest run test/evaluator test/envars.test.ts --sequence.shuffle=false`
- `npm run f`
- `npm run l`
- `npm run tsc`
- `npm run build`
- `SKIP_OG_GENERATION=true npm run build` in `site/`
- CLI e2e with isolated `PROMPTFOO_CONFIG_DIR`/`PROMPTFOO_CACHE_PATH`: timeout disabled completed/pass-exported JSON; `PROMPTFOO_EVAL_TIMEOUT_MS=20` produced a timeout error row with `latencyMs: 20`.

Resolved the two remaining review threads addressed by this patch. CI is now running on the pushed head.

mldangelo-oai:
Updated and pushed `46612d1cb`.

What changed:
- Guarded late provider responses after abort/timeout before they can update shared `_conversation` or `storeOutputAs` state.
- Kept `select-best` and `max-score` comparison rows pending when the global eval timeout fires mid-grading or mid-save, so timeout conversion still records errors.
- Added regressions for the late provider side-effect case and late comparison-grading timeout conversion.

Validation:
- `npx vitest run test/evaluator/execution.test.ts --sequence.shuffle=false -t "late aborted|late select-best|late max-score|applies maxEvalTimeMs to select-best"`
- `npx vitest run test/envars.test.ts test/evaluator/execution.test.ts test/evaluator/gradingConcurrency.test.ts test/evaluator/basic.test.ts test/evaluator/metadata.test.ts test/evaluator/options.test.ts test/util/namedMetrics.test.ts --sequence.shuffle=false`
- `npm run l`
- `git diff --check`
- `npm run tsc`
- `npm run build`
- Isolated CLI smoke with `/tmp/pr6344-e2e/promptfooconfig.yaml`: `PROMPTFOO_EVAL_TIMEOUT_MS=20` produced 1 timeout error; `PROMPTFOO_EVAL_TIMEOUT_MS=0 PROMPTFOO_MAX_EVAL_TIME_MS=0` produced 1 pass.

Resolved the three addressed review threads.

src/evaluator.ts
```diff
@@ -4307,35 +4619,23 @@ class Evaluator {
...
-      globalTimeout = setTimeout(() => {
-        evalTimedOut = true;
-        globalAbortController?.abort();
-      }, maxEvalTimeMs);
-    }
+    const explicitTimeoutConfig = configureExplicitEvaluationTimeout({
+      combinedAbortSignal,
+      options,
+    });
+    combinedAbortSignal = explicitTimeoutConfig.combinedAbortSignal;
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `ae12d33d8`. Setup now runs under `configureSetupEvaluationTimeout`, and an exceptional setup expiry builds and persists timeout result rows through `saveSetupTimeoutEval` instead of rejecting without evidence. The regression covers a never-resolving `beforeAll` hook and asserts the persisted max-duration result.

src/evaluator.ts
```diff
@@ -2830,6 +2877,162 @@ function createTimeoutMetrics(timeoutMs: number): PromptMetrics {
...
+  testSuite: TestSuite;
+}): {
+  combinedAbortSignal: AbortSignal;
+  globalTimeout?: NodeJS.Timeout;
+  isEvalTimedOut: () => boolean;
+  maxEvalTimeMs: number;
+  providerAbortSignal?: AbortSignal;
+  testCaseTimeoutMs: number;
+} {
+  const isRedteamEval = options.isRedteam ?? Boolean(testSuite.redteam);
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `ae12d33d8`. `isRedteamEvaluation` now recognizes generated/exported redteam cases from their metadata and nested assertions even when there is no top-level `redteam` block. I also validated this end to end with an exported config: the CLI selected the redteam defaults (`per-test=4500000ms, max=4560000ms`) and passed `1/1`.

src/evaluator.ts
```diff
@@ -2830,6 +2877,162 @@ function createTimeoutMetrics(timeoutMs: number): PromptMetrics {
...
+      comparisonEvalSteps,
+    );
+
+  let timeoutState = existingTimeoutState;
+  let configuredAbortSignal = combinedAbortSignal;
+  let configuredProviderAbortSignal = providerAbortSignal;
+
+  if (maxEvalTimeMs > 0) {
+    timeoutState ??= createEvaluationTimeoutState(maxEvalTimeMs);
+    const maxEvalAbortSignal = timeoutState.abortController.signal;
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `ae12d33d8`. `waitForTimeout`/`raceAgainstEvaluationTimeout` now bound providers, grouped graders, and teardown even if they ignore the abort signal. An isolated CLI run with a never-settling provider, `maxEvalTimeMs: 75`, and `timeoutMs: 0` returned two max-duration errors in 79ms instead of hanging, with regressions for non-cooperative provider and grader paths.

mldangelo-oai:
Updated this PR on current `main` and repaired one max-duration race found during review.

### Changes

- Merged current `main` (`5264680f1`) into the PR branch.
- Fixed a result duplication race in `src/evaluator.ts`: once a real result is stored, optional JSONL output that is still pending at the global deadline can no longer cause finalization to append a second max-duration result for the same eval step.
- Added a deterministic regression that stalls JSONL writing after the real row is persisted and asserts that only the completed result remains.

### Validation

- `npx vitest run --sequence.shuffle=false test/envars.test.ts test/evaluator/basic.test.ts test/evaluator/execution.test.ts test/evaluator/gradingConcurrency.test.ts test/evaluator/metadata.test.ts test/evaluator/options.test.ts test/util/namedMetrics.test.ts` (`7` files / `203` tests)
- `npm run f`
- `npm run l`
- `git diff --check`
- `npm run jsonSchema:generate` (no generated schema drift)
- Isolated `npm run db:migrate` with `PROMPTFOO_DB_PATH=/tmp/pr6344-migration-config/promptfoo.db`
- `npm run build`
- Isolated real CLI run against an abort-ignoring local provider with `PROMPTFOO_MAX_EVAL_TIME_MS=50` and `PROMPTFOO_EVAL_TIMEOUT_MS=0`: exported exactly one max-duration error row, no late provider response, and zero token/request usage.
- Complete isolated root regression suite: `727` files / `18,463` tests passed (`1` file / `10` tests skipped).

All previously existing inline review threads are resolved on the live PR. Requesting fresh automated review on pushed head `123434924`.

mldangelo-oai:
@codex review

jameshiester-oai (commented):
## Review summary

Reviewed base `c54a30668ad8319d76c20ae96e6680ad6c51a2c6` through head `45441bcd2d5e80f5e9752cf1e91b159f87613ded` as a full first-pass review. The dynamic defaults and active-work races are well covered, but the current head still has deadline and persistence paths that can return corrupt timeout state or fail to return at all. I found one P1 and five new P2 defects; I also reproduced a residual cancellation issue from an older resolved thread.

CI is currently red. Site, Web UI, and Node 20 completed their test phases successfully and failed only because Codecov v11.2.7 could not import its GPG key. macOS Node 22 failed the new large-timer test with `SQLITE_ERROR: no such table: evals` after 10,521 tests passed. The exact CI seed passed locally in both the focused file and the full 19,716-test backend run, so that macOS failure appears to be a shared-database/test-isolation flake rather than a timer-chunking product failure.

### New verified findings

- **P1:** a comparison-grading save already in flight when the deadline wins can commit after `addComparisonTimeoutResults()` and overwrite the persisted timeout row with stale success/failure data. See the inline comment at `src/evaluator.ts:5524`.
- **P2:** a nonnumeric configured `maxEvalTimeMs` reaches the new setup timer despite the CLI's warning-only schema validation; Node coerces the `NaN` delay to 1 ms and the eval records `Evaluation exceeded max duration of not-a-numberms`. See `src/evaluator.ts:3134`.
- **P2:** pre-setup red-team classification ignores scenario-expanded tests, so scenario-only generated red-team suites get the regular 26-minute setup guard instead of the red-team 76-minute guard. See `src/evaluator.ts:3121`.
- **P2:** once the deadline fires, synthesizing and persisting timeout rows is itself unbounded; a stalled store/file writer still hangs the eval. See `src/evaluator.ts:4883`.
- **P2:** the total timer is cleared before writer, OTEL, receiver, and provider teardown, so completed evals can still hang indefinitely in cleanup. See `src/evaluator.ts:4819`.
- **P2:** setup snapshot construction occurs before its cleanup `try`; an early config/provider error leaks the long setup timer and can keep the CLI alive until that timer expires. See `src/evaluator.ts:4961`.

The resolved late-state thread at https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3294040710 is only partially addressed. The check at `src/evaluator.ts:1509` stops a provider response that arrives after abort, but if the provider responds first and abort fires during provider delay or grading, conversation history has already been mutated and `storeOutputAs` can still be written without another cancellation check. Two exact-head probes reproduced this: an aborted direct `runEval` returned success with both shared structures mutated, and a timed-out first conversation row changed the prompt sent by the second row. I am not duplicating that thread with another inline comment.

### Existing feedback rechecked

- The usage/cost/latency reversal, comparison-timeout JSONL recovery, automatic budget excluding setup time, derived-metric recomputation, hook-added comparison sizing, and provider-identity replacement threads still apply: https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3367938881, https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3367938882, https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3367982437, https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3367982439, https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3367982441, and https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3367982438.
- The finalizing-row thread at https://github.com/promptfoo/promptfoo/pull/6344#discussion_r3327310386 is improved by `FinalizingRowState` and replacement callbacks, but a pre-persistence finalizer is still not awaited before handoff; the new unbounded-persistence finding covers the non-settling case.
- The two outdated setup-snapshot threads are fixed at this head: snapshots are withheld until transforms and resume filtering finish, avoiding raw transformed vars and duplicate completed resume rows.

### Tests and end-to-end QA

- `npx vitest run test/evaluator/execution.test.ts test/evaluator/gradingConcurrency.test.ts test/evaluator/basic.test.ts test/evaluator/metadata.test.ts test/evaluator/options.test.ts test/envars.test.ts test/util/namedMetrics.test.ts` — 7 files, 214 tests passed on Node 24.16.0/npm 11.13.0.
- `npx vitest run test/evaluator/execution.test.ts --sequence.seed=1780808283341` on Node 22.22.2 — 47 tests passed.
- `npx vitest run --sequence.seed=1780808283341` on Node 22.22.2 — 766 files passed, 1 skipped; 19,716 tests passed, 10 skipped.
- `npm run tsc` — passed.
- `npm run build` — passed, including the application build.
- `git diff --check base...head` — clean.
- Real CLI success path with `echo`, `--no-cache`, and exported JSON — exit 0; 1/1 passed, score 1, no error.
- Real CLI hard-deadline path with a non-settling provider, `timeoutMs: 0`, `maxEvalTimeMs: 75`, `--no-cache`, and exported JSON — expected exit 100 in about 91 ms; one max-duration error, score 0, no provider/token usage.
- Real CLI invalid-input path with `maxEvalTimeMs: not-a-number` and a non-settling provider — exit 100; emitted `TimeoutNaNWarning`, scheduled the timer at 1 ms, and exported `Evaluation exceeded max duration of not-a-numberms`.
- Read-only controlled probes reproduced the delayed comparison-save overwrite, unbounded timeout persistence/cleanup, leaked setup timer, and post-provider cancellation side effects.

### Gaps and simplification

I did not exercise a remote libSQL service, external model provider, or a real 26/76-minute setup; those paths were validated with deterministic stores, fake timers, and non-settling local providers. The requested PR Review Toolkit command was not available in this environment, so I applied its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses manually and reconciled three independent specialist passes.

The main simplification opportunity is to make timeout conversion a single authoritative, versioned row-finalization operation shared by target rows, comparison rows, setup timeouts, and JSONL recovery. Keeping one deadline owner at the outer `evaluate()` lifecycle—with an explicit bounded persistence/cleanup grace period—would remove the current split setup/execution timers and several late-write states.

src/evaluator.ts
```diff
@@ -4481,258 +5296,267 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
...
+            );
+          }
+        } catch (err) {
+          if (!timeoutConfig.isEvalTimedOut()) {
+            throw err;
+          }
+          logger.warn(`Evaluation stopped after reaching max duration (${maxEvalTimeMs}ms)`);
+        }
+        if (timeoutConfig.isEvalTimedOut()) {
+          await this.addComparisonTimeoutResults({
```
jameshiester-oai, line 5524: **[P1] Prevent late comparison writes from overwriting timeout rows**

If the deadline wins while `finalizeComparisonGrading()` is awaiting `store.saveResult(result)`, that abandoned save has already captured the pre-timeout fields. This path fetches separate rows and saves timeout errors, but the older write can commit afterward and replace the timeout row with a normal success/failure. A clone-on-read store with a delayed first max-score save reproduced exactly that transition. Serialize or version comparison writes, invalidate/await in-flight saves, or make timeout replacement authoritative after every late save.

src/evaluator.ts
```diff
@@ -2940,26 +3012,253 @@ function createEvalStepTimeoutResult(
...
+      : options.timeoutMs;
+  // The exact automatic workload budget depends on resolved/generated rows. Guard setup
+  // with one active-batch budget, then replace it with exact sizing once rows exist.
+  const setupMaxEvalTimeMs =
+    explicitMaxEvalTimeMs ?? getDefaultMaxEvalTimeMs(1, 1, testCaseTimeoutMs, isRedteamEval);
+  if (setupMaxEvalTimeMs <= 0) {
+    return { combinedAbortSignal, isExplicitTimeout };
+  }
+
+  const setupTimeoutState = createEvaluationTimeoutState(setupMaxEvalTimeMs);
```
jameshiester-oai, line 3134: **[P2] Validate the configured duration before scheduling**

Normal CLI config validation reports schema errors but can continue with the raw runtime option. With `maxEvalTimeMs: not-a-number`, this call receives the string, `Math.min()` produces `NaN`, and Node schedules a 1 ms timer. On this head, a non-settling-provider CLI run emitted `TimeoutNaNWarning` and persisted `Evaluation exceeded max duration of not-a-numberms`. Reject non-finite/non-numeric values at this boundary (or stop after schema failure) before creating timeout state.

src/evaluator.ts
```diff
@@ -2940,26 +3012,253 @@ function createEvalStepTimeoutResult(
...
+  isExplicitTimeout: boolean;
+  setupTimeoutState?: EvaluationTimeoutState;
+} {
+  const explicitMaxEvalTimeMs = getExplicitMaxEvalTimeMs(options);
+  const isExplicitTimeout = explicitMaxEvalTimeMs !== undefined;
+  if (explicitMaxEvalTimeMs !== undefined && explicitMaxEvalTimeMs <= 0) {
+    return { combinedAbortSignal, isExplicitTimeout };
+  }
+
+  const isRedteamEval = isRedteamEvaluation(testSuite, options.isRedteam);
```
jameshiester-oai, line 3121: **[P2] Include scenario-expanded tests in setup red-team classification**

Without an explicit `isRedteam`/top-level `redteam` flag, `isRedteamEvaluation()` defaults to `testSuite.tests` and ignores red-team metadata/assertions under `testSuite.scenarios`. `prepareEvaluation()` expands those scenarios later, so the same suite is recognized as red-team only after setup. A scenario-only generated-redteam suite therefore gets the regular 1,560,000 ms setup guard instead of 4,560,000 ms and can be aborted 50 minutes early. Classify the same expanded initial tests here, or use the conservative guard until exact sizing is available.

src/evaluator.ts
```diff
@@ -4345,34 +4848,350 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
...
+        finalizingState.timeoutResult = timeoutResult;
+        this.applyFinalizingTimeoutConversion(finalizingState);
+        if (finalizingState.storePersisted) {
+          await this.replaceStoredResult(timeoutResult);
+        }
+        continue;
+      }
+
+      if (processingContext) {
+        await this.finalizeEvalRow({
```
jameshiester-oai, line 4883: **[P2] Bound persistence of synthesized timeout rows**

After the max timer has fired, this awaits `finalizeEvalRow()` and its store/file writes without any remaining bound. A non-cooperative provider with `maxEvalTimeMs: 20` plus a never-settling timeout-row `appendResult()` left `evaluate()` pending beyond an external 158 ms watchdog. The hard deadline therefore still hangs exactly when recording the promised error rows. Use a bounded best-effort persistence grace/fallback and surface rows that could not be committed.

src/evaluator.ts
```diff
@@ -4293,42 +4787,51 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
...
+      } catch (error) {
+        if (!timeoutState?.isEvalTimedOut()) {
+          throw error;
+        }
+        reachedMaxDuration = true;
+        logger.warn(`Evaluation stopped after reaching max duration (${maxEvalTimeMs}ms)`);
+      }
     }
-    if (evalTimedOut) {
+    clearEvaluationTimeoutState(timeoutState);
```
jameshiester-oai, line 4819: **[P2] Keep a bound through outer resource teardown**

The timeout is cleared here before `evaluate()` awaits JSONL writer close, OTEL flush/shutdown, receiver stop, and `providerRegistry.shutdownAll()`. A trivial completed eval with `maxEvalTimeMs: 50` and a non-settling `shutdownAll()` remained pending indefinitely after its row completed. Retain deadline ownership through outer cleanup or apply a documented bounded cleanup grace period so the total-runtime contract also covers teardown.

src/evaluator.ts
```diff
@@ -4345,34 +4848,350 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
...
+    testSuite: TestSuite;
+    vars: Set<string>;
+  }): Promise<PreparedEvaluation | null> {
+    let testSuite = initialTestSuite;
+    const initialTests = filterByRange(
+      buildTestsFromSuite(testSuite),
+      options.filterRange,
+      warnEmptyFilterRange,
+    );
+    let setupTimeoutSnapshot =
```
jameshiester-oai, line 4961: **[P2] Clear the setup timer when snapshot construction throws**

`initialTests` and `createSetupTimeoutSnapshot()` run before the `try` below, while `_runEvaluation()` has already created `setupTimeoutState`; the caller's cleanup `try` also begins only after `prepareEvaluation()` returns. An unmatched file glob or throwing provider method here therefore leaves the timer referenced. A 1,000 ms probe observed a created timer and zero clears; with defaults this can keep a failed CLI process alive for about 26 minutes. Move this work inside the cleanup region or wrap the prepare call in a `finally` that always clears setup state.

PR #6361, opened 2025-11-25
fix(modelaudit): track scanner version and re-scan on version changes

yash2998chhabria:
## Summary

- Track modelaudit scanner version in the database for each scan
- Re-scan models automatically when modelaudit version has changed (even if model SHA matches)
- Re-scan when previous scan is missing version info (backfill)
- Update existing records instead of creating duplicates (avoids unique constraint violations)
- Fix `--force` flag to properly update existing records

## Changes

- Reuse `getModelAuditCurrentVersion()` from `updates.ts` (no duplicate code)
- Modified `checkModelAuditInstalled()` to return `{ installed, version }`
- Store `scanner_version` field when saving audit records  
- Added `shouldRescan()` helper for clean version comparison logic
- Added `hasErrorsInResults()` helper to reduce duplication
- Flattened nested conditionals for better readability
- Use consistent `?? null` instead of mixed `|| null` and `|| undefined`

## Test plan

- [x] Run `scan-model` on a new model - creates record with scanner_version
- [x] Run `scan-model` on same model - skips (shows "Model already scanned" with version)
- [x] Run `scan-model --force` - re-scans and updates existing record
- [x] Upgrade modelaudit version, run scan - automatically re-scans due to version change
- [x] Added unit tests for re-scan-on-version-change behavior

### DIFF
--- a/src/commands/modelScan.ts
+++ b/src/commands/modelScan.ts
@@ -351,34 +351,33 @@ export function modelScanCommand(program: Command): void {
               },
             };
 
-            // Shared audit data
-            const auditData = {
-              results,
-              checks: results.checks ?? null,
-              issues: results.issues ?? null,
-              hasErrors: hasErrorsInResults(results),
-              totalChecks: results.total_checks ?? null,
-              passedChecks: results.passed_checks ?? null,
-              failedChecks: results.failed_checks ?? null,
-              scannerVersion: currentScannerVersion ?? null,
-              metadata: auditMetadata,
-            };
-
             // Create or update audit record in database
             let audit: ModelAudit;
             if (existingAuditToUpdate) {
-              Object.assign(existingAuditToUpdate, auditData, {
-                updatedAt: Date.now(),
-                ...(revisionInfo.contentHash && { contentHash: revisionInfo.contentHash }),
-              });
+              // Update existing record with new scan results
+              existingAuditToUpdate.results = results;
+              existingAuditToUpdate.checks = results.checks ?? null;
+              existingAuditToUpdate.issues = results.issues ?? null;
+              existingAuditToUpdate.hasErrors = hasErrorsInResults(results);
+              existingAuditToUpdate.totalChecks = results.total_checks ?? null;
+              existingAuditToUpdate.passedChecks = results.passed_checks ?? null;
+              existingAuditToUpdate.failedChecks = results.failed_checks ?? null;
+              existingAuditToUpdate.scannerVersion = currentScannerVersion ?? null;
+              existingAuditToUpdate.metadata = auditMetadata;
+              existingAuditToUpdate.updatedAt = Date.now();
+              if (revisionInfo.contentHash) {
+                existingAuditToUpdate.contentHash = revisionInfo.contentHash;
+              }
               await existingAuditToUpdate.save();
               audit = existingAuditToUpdate;
             } else {
               audit = await ModelAudit.create({
                 name: options.name || `Model scan ${new Date().toISOString()}`,
-                author: getAuthor() ?? null,
+                author: getAuthor() || undefined,
                 modelPath: paths.join(', '),
-                ...auditData,
+                results,
+                metadata: auditMetadata,
+                scannerVersion: currentScannerVersion || undefined,
                 ...revisionInfo,
               });
             }


src/commands/modelScan.ts
```diff
@@ -310,10 +387,41 @@ export function modelScanCommand(program: Command): void {
...
+                    timeout: cliOptions.timeout,
+                    maxSize: options.maxSize,
+                    verbose: options.verbose,
+                    sbom: options.sbom,
+                    strict: options.strict,
+                    dryRun: options.dryRun,
+                    cache: options.cache,
+                    quiet: options.quiet,
+                    progress: options.progress,
+                    stream: options.stream,
```
addelong: nit: this is duplicated above, could extract to a function or const to remove duplication

test/commands/modelScan.test.ts
mldangelo, line 1: Thank you for updating tests. There are no tests covering the actual re-scan-on-version-change behavior. Consider adding tests.

src/commands/modelScan.ts
```diff
@@ -289,13 +354,25 @@ export function modelScanCommand(program: Command): void {
...
+                  (results.issues &&
+                    results.issues.some(
+                      (issue) => issue.severity === 'critical' || issue.severity === 'error',
+                    )),
+              );
+              existingAuditToUpdate.totalChecks = results.total_checks ?? null;
+              existingAuditToUpdate.passedChecks = results.passed_checks ?? null;
+              existingAuditToUpdate.failedChecks = results.failed_checks ?? null;
+              existingAuditToUpdate.scannerVersion = currentScannerVersion || null;
+              existingAuditToUpdate.metadata = {
```
mldangelo: this is duplicated. As your LLM to DRY it

src/commands/modelScan.ts
```diff
@@ -289,13 +354,25 @@ export function modelScanCommand(program: Command): void {
...
-              name: options.name || `Model scan ${new Date().toISOString()}`,
-              author: getAuthor() || undefined,
-              modelPath: paths.join(', '),
-              results,
-              metadata: {
+            // Create or update audit record in database
+            let audit: ModelAudit;
+            if (existingAuditToUpdate) {
+              // Update existing record with new scan results
+              existingAuditToUpdate.results = results;
```
mldangelo: do you want to update the updatedAt here too?

mldangelo (commented):
Thanks!

src/commands/modelScan.ts
```diff
@@ -24,6 +24,32 @@ export async function checkModelAuditInstalled(): Promise<boolean> {
   });
 }
 
+/**
+ * Get the installed modelaudit version.
+ * @returns The version string (e.g., "0.2.16") or null if not available
+ */
+export async function getModelAuditVersion(): Promise<string | null> {
```
mldangelo: consolidate with checkModelAuditInstalled

yash2998chhabria: makes sense

src/commands/modelScan.ts
yash2998chhabria: yup, makes sense

src/commands/modelScan.ts
```diff
@@ -1,26 +1,48 @@
...
+  version: string | null;
+}> {
+  const version = await getModelAuditCurrentVersion();
+  // If we got a version, modelaudit is installed
+  // If null, it could mean not installed OR installed but couldn't parse version
+  if (version !== null) {
+    return { installed: true, version };
+  }
+
+  // Fallback: check if modelaudit command exists (even if version parsing failed)
```
mldangelo: please get rid of this. You don't need it.

src/commands/modelScan.ts
```diff
@@ -289,31 +349,63 @@ export function modelScanCommand(program: Command): void {
...
+            };
+
+            // Create or update audit record in database
+            let audit: ModelAudit;
+            if (existingAuditToUpdate) {
+              // Update existing record with new scan results
+              existingAuditToUpdate.results = results;
+              existingAuditToUpdate.checks = results.checks || null;
+              existingAuditToUpdate.issues = results.issues || null;
+              existingAuditToUpdate.hasErrors = Boolean(
```
mldangelo: this is duplicated a few times in the file and should be abstracted

src/commands/modelScan.ts
```diff
@@ -289,31 +349,63 @@ export function modelScanCommand(program: Command): void {
...
+                results.has_errors ||
+                  (results.issues &&
+                    results.issues.some(
+                      (issue) => issue.severity === 'critical' || issue.severity === 'error',
+                    )),
+              );
+              existingAuditToUpdate.totalChecks = results.total_checks ?? null;
+              existingAuditToUpdate.passedChecks = results.passed_checks ?? null;
+              existingAuditToUpdate.failedChecks = results.failed_checks ?? null;
+              existingAuditToUpdate.scannerVersion = currentScannerVersion || null;
```
mldangelo: we use || null in some places and undefined in others. Ask the llm to fix this

src/commands/modelScan.ts
```diff
@@ -289,31 +349,63 @@ export function modelScanCommand(program: Command): void {
...
+              audit = await ModelAudit.create({
+                name: options.name || `Model scan ${new Date().toISOString()}`,
+                author: getAuthor() || undefined,
+                modelPath: paths.join(', '),
+                results,
+                metadata: auditMetadata,
+                // Revision tracking
+                ...revisionInfo,
+                // Scanner version tracking
+                scannerVersion: currentScannerVersion || undefined,
```
mldangelo: why are you doing this in this order?

mldangelo (approved):

PR #6363, opened 2025-11-25
chore: Make provider options less strict

faizanminhas:
Had a bug in cloud such that a key that wasn't strictly adherent to this schema was saved- Not sure how many bad targets were created, but this schema check makes cloud targets error out if an extraneous key is pulled, which IMO is not a big deal. Making it `strict` and every key `optional` also doesn't make much sense IMO, so I'm just going to remove that constraint.

### DIFF
--- a/src/validators/providers.ts
+++ b/src/validators/providers.ts
@@ -104,11 +104,11 @@ export const ProvidersSchema = z.union([
   z.array(
     z.union([
       z.string(),
+      CallApiFunctionSchema,
       z.record(z.string(), ProviderOptionsSchema),
       ProviderOptionsSchema,
-      CallApiFunctionSchema,
     ]),
   ),
 ]);
 
-export const ProviderSchema = z.union([z.string(), ProviderOptionsSchema, ApiProviderSchema]);
+export const ProviderSchema = z.union([z.string(), ApiProviderSchema, ProviderOptionsSchema]);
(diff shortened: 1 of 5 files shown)


src/validators/providers.ts
```diff
@@ -13,17 +13,15 @@ import type {
...
-  .strict();
+export const ProviderOptionsSchema = z.object({
+  id: z.custom<ProviderId>().optional(),
+  label: z.custom<ProviderLabel>().optional(),
+  config: z.any().optional(),
+  prompts: z.array(z.string()).optional(),
+  transform: z.string().optional(),
+  delay: z.number().optional(),
+  env: ProviderEnvOverridesSchema.optional(),
+});
```
mldangelo, line 24: PLEASE PLEASE cover this in a test

faizanminhas: added

src/validators/providers.ts
```diff
@@ -106,11 +104,11 @@ export const ProvidersSchema = z.union([
...
+      CallApiFunctionSchema,
       z.record(z.string(), ProviderOptionsSchema),
       ProviderOptionsSchema,
-      CallApiFunctionSchema,
     ]),
   ),
 ]);
 
-export const ProviderSchema = z.union([z.string(), ProviderOptionsSchema, ApiProviderSchema]);
+export const ProviderSchema = z.union([z.string(), ApiProviderSchema, ProviderOptionsSchema]);
```
faizanminhas, line 114: needed to change these to make some tests pass, as far as I can tell, re-ordering here preserves `callApi` and other provider keys that might otherwise be skipped by removing `strict`

PR #6370, opened 2025-11-26
chore: migrate entirely from dependabot to renovate

JustinBeckwith:


.github/dependabot.yml
```diff
@@ -1,23 +0,0 @@
-version: 2
-updates:
-  - package-ecosystem: 'npm'
-    directory: '/'
-    schedule:
-      interval: daily
-    allow:
```
mldangelo, line 7: would be nice if we can do this in renovate for anthropic / openai etc.

JustinBeckwith: Just to make sure I fully understand what this was doing - is this bumping the version in `package.json` to the latest patch release?  Even with a sliding range specifier?

JustinBeckwith: Assuming that's what it does, added this explicitly to renovate config

PR #6391, opened 2025-11-27
fix: do not truncate image responses in eval

JustinBeckwith:
While playing with image generation, I noticed that we render markdown for image generation, and it's frequently truncated.  By expanding the text, it shows more of the markdown, which lets it render the image.  This is janky.  So instead, I'm trying to detect if it's a image response, and skip truncating,


mldangelo (approved):
elegant

PR #6412, opened 2025-11-29
fix(redteam): propagate abort signals through agentic strategy providers

MrFlounder:
- Update fetchWithProxy to accept optional abortSignal parameter
- Combine abort signals in fetchWithTimeout using AbortSignal.any()
- Don't retry on AbortError in fetchWithRetries
- Pass options with abortSignal through all agentic provider callApi methods
- Re-throw AbortError in catch blocks instead of swallowing it
- Add abort signal handling tests for all affected providers

This ensures that when a scan is stopped, in-flight API calls are properly
cancelled rather than continuing until completion.

### DIFF
--- a/src/generated/constants.ts
+++ b/src/generated/constants.ts
@@ -1,3 +1,0 @@
-// This file is auto-generated during build. Do not edit manually.
-// Generated at: 2025-11-29T18:19:26.063Z
-export const POSTHOG_KEY = '';


mldangelo (approved):

src/generated/constants.ts
danenania, line 1: wonder why this got included?

MrFlounder: you are right i will remove

Issue #6350, opened 2025-11-25
Test generation hangs when run from CI/CD


blueCicada:
**Describe the bug**
During test generation, promptfoo appears to sometimes(?) hang when run from inside a CI/CD pipeline (we're using Bitbucket pipelines). It does not hang when run locally.

We first observed this behaviour around v0.119.0, prior to that we had no issues with hanging pipelines.

**To Reproduce**

1. From a CI/CD pipeline, run a test generation command such as `npx --yes promptfoo@latest redteam generate -c promptfoo-configs/beavertails-config.yaml -o promptfoo-configs/aisafety-beavertails-redteam-input.yaml`
2. The stdout will show that test generation completed ("Wrote n cases... Run promptfoo redteam eval... to run the red team!"), but then that command just stalls and never exits

We experienced the same issue with different config files, so unsure that it's an issue with the config file.

The only way we've gotten it to terminate is by wrapping the command in a `timeout`.

There are also no error or debug logs. We've set PROMPTFOO_DISABLE_DEBUG_LOG and PROMPTFOO_DISABLE_ERROR_LOG to 0.

**Expected behavior**
Command exits on completion of test generation.

**System information:**

- Promptfoo version: > 0.118.17

**Additional context**
Unsure if this is somehow related to https://github.com/promptfoo/promptfoo/issues/5893 (last comment suggests this is not yet fixed even though the issue is closed)

Issue #5893, opened 2025-10-10
Importing Promptfoo prevents Node from exiting naturally


justindilen:
**Describe the bug**
When importing the npm package, it seems to create a process that prevents node from exiting without even invoking anything.

**To Reproduce**
Steps to reproduce the behavior, including example Promptfoo configurations if applicable:

1. Create a node project (with package.json "type": "module") and install latest promptfoo
2. Import promptfoo in a JS file like so:
 ```
import promptfoo from 'promptfoo';
```
3. Run the file using node: `node ./test.js`
4. Notice the node process never exits naturally

**Expected behavior**
Node should quickly exit instead of hanging. 
In general, even if promptfoo eval is run, it should cleanup any resources / timeouts and naturally exit.

**System information:**

- Promptfoo version: 0.118.14
- Node version: 22.13.1
- Windows 10, powershell terminal

**Additional context**
Earlier versions of promptfoo did not seem to have this issue. Testing shows that release 0.118.7 has this issue, but 0.118.6 does not.

mldangelo:
Hi @justindilen,

Sorry for the bug and for not closing this sooner - this was fixed in **PR #5698** on September 23, 2025 (before you reported it), released in version 0.118.8.

The PostHog client was keeping connections open and preventing Node from exiting naturally. This has been resolved - the process now exits cleanly after commands complete.

Please pull the latest version and the issue should be gone. If you experience any other issues, feel free to comment or re-open.

See the [Command Line documentation](https://www.promptfoo.dev/docs/usage/command-line) for more info.

Thanks for the report!

justindilen:
> Hi [@justindilen](https://github.com/justindilen),
> 
> Sorry for the bug and for not closing this sooner - this was fixed in **PR [#5698](https://github.com/promptfoo/promptfoo/pull/5698)** on September 23, 2025 (before you reported it), released in version 0.118.8.
> 
> The PostHog client was keeping connections open and preventing Node from exiting naturally. This has been resolved - the process now exits cleanly after commands complete.
> 
> Please pull the latest version and the issue should be gone. If you experience any other issues, feel free to comment or re-open.
> 
> See the [Command Line documentation](https://www.promptfoo.dev/docs/usage/command-line) for more info.
> 
> Thanks for the report!

@mldangelo  Appreciate the response. 

Unfortunately, in the original post I was testing using 0.118.14 and was having this issue, so I don't think it was fixed in 0.118.8.

[...]

PR #6422, opened 2025-12-01
fix(cli): resolve eval command hanging by adding missing await calls

capttrousers:
The eval command was hanging after completion due to missing await calls
that caused premature cleanup and unresolved promise chains.

Root causes:
1. doEval() called without await - cleanup ran before eval completed
2. program.parse() used instead of parseAsync() - promise chain not awaited
3. main() called without catch handler - initialization errors swallowed

Fixes:
- Add await to doEval() call in eval command action handler
- Change program.parse() to await program.parseAsync()
- Add catch handler to main() call for proper error handling
- Add cleanup functions: closeLogger(), closeDbIfOpen()
- Add Undici dispatcher cleanup to prevent hanging connections
- Add forced exit timeout (100ms) after cleanup to handle stuck event loop

The forced exit is needed because even after proper cleanup, some
dependencies keep the event loop alive. The 100ms timeout allows async
operations to complete while ensuring the process eventually exits.

Also updates telemetry.sendEvent() to use fetchWithTimeout instead of
fetchWithProxy for better timeout handling during shutdown.

Refs: 
- https://github.com/promptfoo/promptfoo/issues/5893 
- https://github.com/promptfoo/promptfoo/issues/6350
- https://github.com/promptfoo/promptfoo/issues/4770

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -26,7 +26,7 @@ import { shareCommand } from './commands/share';
 import { showCommand } from './commands/show';
 import { validateCommand } from './commands/validate';
 import { viewCommand } from './commands/view';
-import { closeDbIfOpen } from './database';
+import { closeDbIfOpen } from './database/index';
 import logger, { closeLogger, initializeRunLogging, setLogLevel } from './logger';
 import { runDbMigrations } from './migrate';
 import { discoverCommand as redteamDiscoverCommand } from './redteam/commands/discover';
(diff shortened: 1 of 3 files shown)


capttrousers:
This is actually just stacked with https://github.com/promptfoo/promptfoo/pull/6421 because I tackled them together, but once that first once is merged, I can rebase and drop that other commit

mldangelo:
Thank you!

PR #6428, opened 2025-12-01
chore!: replace TypeScript enums with const objects

JustinBeckwith:
BREAKING CHANGE: TypeScript Enums Replaced with Const Objects

**For npm module consumers:** This release replaces all TypeScript enums with const objects and type unions. Most usage patterns remain unchanged, but **reverse enum lookups are no longer supported**.

**What You Need to Do**

*If you're using standard enum access patterns, no changes needed:*

```typescript
import { ResultFailureReason } from 'promptfoo';

// ✅ All these patterns continue to work
const reason = ResultFailureReason.NONE;           // Still works
const isError: ResultFailureReason = 2;            // Still works as type annotation
Object.values(ResultFailureReason);                 // Still returns [0, 1, 2]
Object.keys(ResultFailureReason);                   // Still returns ['NONE', 'ASSERT', 'ERROR']

// Type checking still works identically
function handleFailure(reason: ResultFailureReason) { ... }
```

*If you're using reverse lookups, you must change your code:*

```typescript
// ❌ This no longer works (returns undefined)
const name = ResultFailureReason[0];  // Was: 'NONE', Now: undefined

// ✅ Create your own reverse mapping if needed
const reasonNames: Record<ResultFailureReason, string> = {
  [ResultFailureReason.NONE]: 'NONE',
  [ResultFailureReason.ASSERT]: 'ASSERT',
  [ResultFailureReason.ERROR]: 'ERROR',
};
const name = reasonNames[0];  // 'NONE'

// ✅ Or use the new type guard for validation
import { isResultFailureReason } from 'promptfoo';
if (isResultFailureReason(value)) {
  // value is now typed as ResultFailureReason
}
```

**Affected Exported Types**

The following publicly exported enums are now const objects:

- **`ResultFailureReason`** - Most commonly used by consumers for evaluation results
- **`EmailValidationStatus`** - Used in email validation
- **`CodeScanSeverity`** - Used in code scanning features
- **`FileChangeStatus`** - Used in code scanning features

Internal enums (not typically used by external consumers) were also converted but should not affect most users.

### Why This Change?

TypeScript enums generate runtime code with bidirectional mappings, which prevents using modern build tools that only strip types (Node.js `--strip-types`, esbuild, swc). The const object pattern is the modern TypeScript best practice, offering better tree-shaking, simpler runtime behavior, and universal tooling compatibility.

---

## Summary

Replace all TypeScript enums with const objects + type unions, enabling Biome's `noEnum` lint rule.

## Why This Change?

**1. Enums are not type-level TypeScript**

TypeScript enums compile to runtime JavaScript objects with bidirectional mappings. Unlike type annotations that are erased at compile time, enums require code transformation. This means:
- Node.js native `--strip-types` cannot handle enums (it only strips types, doesn't transform)
- Bundlers with "type-only" modes (esbuild's `--strip-types`, swc) cannot process enums
- This adds unnecessary complexity to the build pipeline

**2. Const objects are the modern best practice**

The TypeScript community has largely moved away from enums. The [TypeScript documentation itself](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums) notes that const objects are often sufficient and preferred:

```typescript
// Modern pattern - works with any TypeScript tooling
export const Status = {
  Active: 'active',
  Inactive: 'inactive',
} as const;
export type Status = (typeof Status)[keyof typeof Status];
```

Benefits:
- **Tree-shakeable**: Bundlers can eliminate unused values
- **No runtime overhead**: No bidirectional mapping object generated
- **Better type inference**: Literal types are preserved
- **Universal compatibility**: Works with any TypeScript stripper/transpiler
- **Simpler mental model**: It's just a JavaScript object

**3. Enables future tooling flexibility**

With this change, the codebase can potentially:
- Use Node.js native TypeScript execution (`node --strip-types`)
- Simplify the build pipeline by removing `tsx` dependency
- Use faster, simpler TypeScript processing tools

## Changes

**Enums converted (11 total):**
- `ResultFailureReason` (src/types/index.ts)
- `EmailValidationStatus` (src/types/email.ts)
- `CodeScanSeverity`, `FileChangeStatus` (src/types/codeScan.ts)
- `Severity` (src/redteam/constants/metadata.ts)
- `BrowserBehavior` (src/util/server.ts)
- `TransformInputType` (src/util/transform.ts)
- `LlamaVersion` (src/providers/bedrock/index.ts)
- `DownloadFormat` (src/app/src/hooks/useDownloadEval.ts)
- `Phases` (src/redteam/providers/simba.ts)
- `TabIndex` (src/app/.../HttpAdvancedConfiguration.tsx)

**Additional improvements:**
- Added `CodeScanSeveritySchema` and `SeveritySchema` for Zod validation
- Added `isResultFailureReason()` type guard for safe type narrowing
- Added `BrowserBehaviorNames` for reverse lookup in logging
- Enabled `noEnum` Biome lint rule to prevent future enum usage

## Breaking Changes

**Reverse enum lookup no longer works:**

```typescript
// This pattern no longer works
const name = ResultFailureReason[0]; // Was: 'NONE', Now: undefined

// These patterns continue to work identically
ResultFailureReason.NONE           // ✅ Works
const x: ResultFailureReason = 0;  // ✅ Works
Object.values(ResultFailureReason) // ✅ Works
```

For the rare cases where reverse lookup is needed, use the provided `*Names` objects (e.g., `BrowserBehaviorNames[value]`).

## Test plan

- [x] `npm run build` passes
- [x] `npm run tsc` passes
- [x] `npm run lint` passes
- [x] `npm run format` passes
- [x] `npm test` passes (8,272 tests)

### DIFF
--- a/test/rateLimit.test.ts
+++ b/test/rateLimit.test.ts
@@ -2,8 +2,6 @@ import { vi } from 'vitest';
 
 import { fetchWithRetries } from '../src/util/fetch/index';
 
-const mockedFetch = vi.spyOn(global, 'fetch');
-
 const mockedFetchResponse = (ok: boolean, response: object, headers: object = {}) => {
   const responseText = JSON.stringify(response);
   return {
@@ -27,29 +25,31 @@ const mockedSetTimeout = (reqTimeout: number) =>
     return 0 as any;
   });
 
+// Create a mock function that will be used for fetch
+const mockFetch = vi.fn();
+
 describe('fetchWithRetries', () => {
   beforeEach(() => {
     vi.useFakeTimers();
+    // Use stubGlobal which properly replaces the global fetch for all modules
+    vi.stubGlobal('fetch', mockFetch);
   });
 
   afterEach(() => {
-    mockedFetch.mockReset();
+    mockFetch.mockReset();
     vi.useRealTimers();
-  });
-
-  afterAll(() => {
-    vi.clearAllMocks();
+    vi.unstubAllGlobals();
   });
 
   it('should fetch data', async () => {
     const url = 'https://api.example.com/data';
     const response = { data: 'test data' };
 
-    mockedFetch.mockResolvedValueOnce(mockedFetchResponse(true, response));
+    mockFetch.mockResolvedValueOnce(mockedFetchResponse(true, response));
 
     const result = await fetchWithRetries(url, {}, 1000);
 
-    expect(mockedFetch).toHaveBeenCalledTimes(1);
+    expect(mockFetch).toHaveBeenCalledTimes(1);
     await expect(result.json()).resolves.toEqual(response);
   });
 
@@ -62,7 +62,7 @@ describe('fetchWithRetries', () => {
 
     const setTimeoutMock = mockedSetTimeout(timeout);
 
-    mockedFetch
+    mockFetch
       .mockResolvedValueOnce(
         mockedFetchResponse(false, response, {
           'X-RateLimit-Remaining': '0',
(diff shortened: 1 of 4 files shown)


mldangelo (dismissed):
I have wanted to do this for over a year. thank you!

PR #6460, opened 2025-12-02
fix(cli): resolve view command premature exit and eval hanging issues

capttrousers:
Fix async lifecycle issue where startServer() returned immediately after calling httpServer.listen() (non-blocking), causing Commander to consider the view action complete before the server was running. This led to cleanup running prematurely and the process exiting.

Also fix main.js entrypoint script file path check for `npm link` support. 

Root cause: startServer() resolved instantly instead of blocking until the server actually stopped. Commander's parseAsync() interpreted this as "command complete", triggering cleanup before the server finished binding.

Solution: Make startServer() return a Promise that only resolves on shutdown signals (SIGINT/SIGTERM):
- view command: Blocks until Ctrl+C → cleanup → exit
- eval command: Completes work → cleanup → exit

Changes:
- startServer() now returns Promise that blocks until shutdown signal
- Add SIGINT/SIGTERM handlers to gracefully close server and file watcher
- Re-enable full cleanup in shutdownGracefully() (logger, db, dispatcher)
- Use .unref() on 100ms exit timeout to allow natural process termination

The .unref() pattern allows Node.js to exit naturally if all handles close cleanly, but forces exit after 100ms if resources linger.

See reflections.md for detailed timeline and analysis of how these issues surfaced and were resolved.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,37 +1,11 @@
-import { fileURLToPath } from 'node:url';
-import { resolve } from 'node:path';
 import { realpathSync } from 'node:fs';
+import { resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
 import { Command } from 'commander';
 import { getGlobalDispatcher } from 'undici';
 
 import { VERSION } from './version';
 import { checkNodeVersion } from './checkNodeVersion';
-
-/**
- * Checks if the current module is the main entry point.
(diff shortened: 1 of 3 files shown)


capttrousers:
The reflections.md doc can be removed before merging, thats just more for context. 
There's some log lines too that could be removed, they're helpful for diagnosing the issue.

PR #6470, opened 2025-12-02
fix(site): responsive logo containers

danenania:
## Before

CleanShot 2025-12-02 at 15 06 07@2x

https://github.com/user-attachments/assets/13593b6c-ab4c-4f80-8df3-cf28ffc59585

## After

CleanShot 2025-12-02 at 15 06 16@2x

https://github.com/user-attachments/assets/34c05685-f917-4e3a-816b-678284aa30c7

### DIFF
--- a/site/src/components/LogoContainer.module.css
+++ b/site/src/components/LogoContainer.module.css
@@ -1,10 +1,9 @@
 .logoContainer {
   display: flex;
   flex-wrap: nowrap;
-  column-gap: 4rem;
+  gap: 4rem;
   align-items: center;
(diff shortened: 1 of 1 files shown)


faizanminhas (approved):
What do you think about replacing it with an auto-scrolling carousel? happy to ship this though (in the meanwhile, or otherwise)

danenania:
> What do you think about replacing it with an auto-scrolling carousel? happy to ship this though (in the meanwhile, or otherwise)

That could be nice but I think we'd need more logos in the list for it to work

PR #6471, opened 2025-12-03
fix(deps): downgrade zod-validation-error for ESM compatibility

JustinBeckwith:
## Summary

- Downgrades `zod-validation-error` from v5.0.0 to v3.5.4 to fix ESM compatibility issues
- Removes `/v3` subpath imports that caused `ERR_PACKAGE_PATH_NOT_EXPORTED` errors when promptfoo is consumed by ESM-only projects
- Also fixes incorrect `ZodError` type import (was from `zod-validation-error/v3`, now correctly from `zod`)

## Context

`zod-validation-error` v5 uses subpath exports (`/v3`, `/v4`) for Zod version compatibility. However, these subpath exports cause errors when promptfoo is imported by ESM-only projects:

```
Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './v3' is not defined by "exports"
```

Upgrading to Zod v4 would be the cleaner fix (allowing us to use the default export), but this is blocked by `@anthropic-ai/claude-agent-sdk` which requires Zod v3.

## Trade-offs

**What we lose:**
- Standardized error message format from v5: "expected `<expectation>`, received `<realization>`"
- New `reportInput` option to control whether values appear in error messages

**What stays the same:**
- `fromError()` and `fromZodError()` work identically
- Core functionality of converting Zod errors to readable messages

The practical impact is minimal since we only use the basic conversion functions.

## Test plan

- [x] `npm test` passes (481 test files)
- [x] `npm run build` succeeds
- [x] `npm run format` succeeds
- [ ] Verify ESM import works in consuming project


faizanminhas (approved):
The tradeoffs seem reasonable to me

PR #6483, opened 2025-12-03
fix(ci): enforce npm version to prevent lockfile incompatibility

JustinBeckwith:
## Summary

- Adds npm engine constraint to prevent lockfile incompatibility caused by npm 11.6.2 bug ([npm/cli#8669](https://github.com/npm/cli/issues/8669))
- Updates CI workflows to upgrade npm to latest for Node 24.x jobs
- Node 20.x/22.x continue to use their native npm 10.x (unaffected by the bug)

## Problem

Node 24.x currently ships with npm 11.6.2, which has a known bug that generates incompatible lockfiles. This causes "lockfile flapping" when different contributors or CI environments use different npm versions.

### 1. `package.json` - engines field
```json
"engines": {
  "node": ">=20.0.0",
  "npm": ">=10.0.0 <11.0.0 || >=11.6.4"
}
```

This allows:
- npm 10.x (Node 20/22 native) ✅
- npm 11.0.0-11.6.3 ❌ (blocked - problematic versions)
- npm 11.6.4+ ✅ (fixed version)

### 2. `.npmrc` - engine-strict enforcement
```
engine-strict=true
```

### 3. CI Workflows - npm upgrade for Node 24.x

Added `npm install -g npm@latest` step before `npm ci` for all jobs using Node 24.x (either via matrix or `.nvmrc`).

## Test plan

- [ ] CI passes on Node 20.x (uses native npm 10.x)
- [ ] CI passes on Node 22.x (uses native npm 10.x)
- [ ] CI passes on Node 24.x (upgrades to npm@latest)
- [ ] Local `npm install` works with npm 10.x
- [ ] Local `npm install` works with npm 11.6.4+
- [ ] Local `npm install` fails with npm 11.6.2 (engine-strict)


mldangelo (commented):
Consider something like this: as an alternative: https://gist.github.com/mldangelo/619a62c376d8fced40d0a7d41a5040bc 

We should maybe pin the npm version too

will-holley:
I think this broke the ability to run commands using alternative package managers e.g.

```sh
~/promptfoo  main ✗                                                                                                                                                                                                 28m ⚑  
▶ pnpm local
 ERROR  This project is configured to use npm
For help, run: pnpm help run

~/promptfoo  main ✗                                                                                                                                                                                                28m ⚑  ⍉
▶ npm run local

> promptfoo@0.119.14 local
> tsx src/main.ts

...
```

PR #6486, opened 2025-12-03
fix: improving error logging for GOAT

jameshiester:
NOTE: there should be additional work to improve GOAT logging and error handling.  Right now the strategy is aggressive at swallowing errors and will simply move to the next turn in most cases.


src/redteam/providers/goat.ts
```diff
@@ -470,7 +470,12 @@ export default class GoatProvider implements ApiProvider {
           logger.debug('[GOAT] Operation aborted');
           throw error;
         }
-        logger.error(`[GOAT] Error in GOAT turn ${turn}`, { error });
+        logger.error(
+          `[GOAT] An error occurred in GOAT turn ${turn}.  The test will continue to the next turn in the conversation.`,
+          {
+            error: (error as Error).message || error,
```
will-holley, line 603: `error?.message` ?

jameshiester: error cannot be undefined

PR #6496, opened 2025-12-04
fix(code-scan): output valid JSON when no files to scan with --json flag

danenania:
When `--json` flag is passed and there are no files to scan, output a valid `ScanResponse` JSON object instead of plain text. This fixes the GitHub Action failing to parse the CLI output.

### DIFF
--- a/src/codeScan/config/loader.ts
+++ b/src/codeScan/config/loader.ts
@@ -9,7 +9,7 @@ import path from 'path';
 
 import yaml from 'js-yaml';
 import { type Config, ConfigSchema, DEFAULT_CONFIG } from './schema';
-import { ConfigLoadError } from '../../types/codeScan';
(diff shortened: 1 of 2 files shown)


src/codeScan/scanner/index.ts
```diff
@@ -207,7 +207,11 @@ export async function executeScan(repoPath: string, options: ScanOptions): Promi
     // Check if there are no files to scan
     if (includedFiles.length === 0) {
       const msg = 'No files to scan';
-      if (showSpinner && spinner) {
+
+      // In JSON mode, output a proper JSON response for programmatic consumption
+      if (options.json) {
+        console.log(JSON.stringify({ success: true, comments: [], review: msg }, null, 2));
```
mldangelo: we may still want to use logger here

danenania: fixed

PR #6498, opened 2025-12-04
chore: pin npm version via corepack

mldangelo:
## Summary

- Add `packageManager` field to package.json to pin npm@11.6.4 via corepack
- Ensures all contributors use a compatible npm version that satisfies the engines requirement (`>=10.0.0 <11.0.0 || >=11.6.4`)
- Update contributing docs and AGENTS.md with corepack setup instructions

## Background

npm 11.5.1 (which ships with Node.js 24.7.0) falls in the unsupported gap of the engines requirement. Corepack, which ships with Node.js, allows pinning the exact npm version per-project.

## Setup for contributors

```bash
nvm use
corepack enable npm  # one-time setup
npm install
```

## Test plan

- [x] Verified `corepack enable npm` uses npm@11.6.4
- [x] Verified `npm install` succeeds without engine errors

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -588,6 +588,7 @@ npx npm-check-updates --target latest
 
 - Use CommonJS modules (type: "commonjs" in package.json)
 - Node.js version requirement (>=20.0.0). Use `nvm use` to align with `.nvmrc` (currently v24.7.0).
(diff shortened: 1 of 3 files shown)


JustinBeckwith (approved):
I lowkey wonder how many people actually have corepack enabled, but this is a good change (sorry I missed it)

PR #6501, opened 2025-12-04
fix: add CommonJS fallback for .js files in ESM environment

mldangelo:
## Summary

- Adds automatic CommonJS fallback when loading `.js` files that use CommonJS syntax (`require`/`module.exports`) in the ESM environment
- Detects CJS-in-ESM errors (e.g., "module is not defined") and falls back to `createRequire()` loading
- Allows existing examples and user configurations to continue working without requiring file renames or syntax changes

## Problem

After the ESM migration, `.js` files using CommonJS syntax fail to load with errors like:

```
ESM import failed: module is not defined in ES module scope
This file is being treated as an ES module because it has a '.js' file extension 
and 'package.json' contains "type": "module".
```

## Solution

When ESM import fails with a CJS-related error for a `.js` file, automatically fall back to loading via `createRequire()`:

```typescript
// Fall back to CommonJS loading for .js files that use CJS syntax
if (modulePath.endsWith('.js') && isCjsInEsmError(errorMessage)) {
  const require = createRequire(import.meta.url);
  const importedModule = require(resolvedPath);
  // ...
}
```

## Test plan

- [x] All existing ESM tests pass (12 tests)
- [x] New CJS fallback tests pass (7 tests)  
- [x] `isCjsInEsmError` detection tests pass (7 tests)
- [x] Integration test: `examples/javascript-assert-external` now works without modification

### DIFF
--- a/src/esm.ts
+++ b/src/esm.ts
@@ -132,7 +132,7 @@ export async function importModule(modulePath: string, functionName?: string) {
             `  2. Convert to ESM syntax (import/export)\n` +
             `  3. Ensure the file has valid JavaScript syntax`,
         );
-        combinedError.cause = { esmError: err, cjsError: cjsErr };
+        (combinedError as any).cause = { esmError: err, cjsError: cjsErr };
         throw combinedError;
       }
     } else {


JustinBeckwith (approved):
Worth double checking what the AI kindly says, but generally looks fine to me - dual ESM/CJS is hard yo.

yash2998chhabria (approved):
LGTM

PR #6516, opened 2025-12-05
feat(redteam): add copy button to messages box in multi-turn conversation ui

MrFlounder:


src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
```diff
@@ -395,6 +397,25 @@ export default function EvalOutputPromptDialog({
...
+  const formatMessagesForCopy = useCallback((messages: Message[]): string => {
+    return messages
+      .filter((msg): msg is LoadedMessage => !msg.loading && msg.content !== null)
+      .map((msg) => {
+        const roleLabel =
+          msg.role === 'user' ? 'User' : msg.role === 'assistant' ? 'Assistant' : 'System';
+        return `[${roleLabel}]\n${msg.content}`;
+      })
+      .join('\n\n');
+  }, []);
```
faizanminhas, line 409: Are you sure this is how you wanna format the messages?

src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
```diff
@@ -395,6 +397,25 @@ export default function EvalOutputPromptDialog({
...
+
+  const copyMessagesToClipboard = useCallback(async () => {
+    const allMessages = [...parsedMessages, ...redteamHistoryMessages];
+    const formattedText = formatMessagesForCopy(allMessages);
+    await navigator.clipboard.writeText(formattedText);
```
faizanminhas, line 414: fyi this api isn't always available in every browser

src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
```diff
@@ -215,6 +216,7 @@ export default function EvalOutputPromptDialog({
 }: EvalOutputPromptDialogProps) {
   const [activeTab, setActiveTab] = useState(0);
   const [copied, setCopied] = useState(false);
+  const [copiedMessages, setCopiedMessages] = useState(false);
```
faizanminhas, line 219: nit: use a different name for this boolean, it reads like it'll have messages

src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
```diff
@@ -395,6 +397,25 @@ export default function EvalOutputPromptDialog({
...
+      .filter((msg): msg is LoadedMessage => !msg.loading && msg.content !== null)
+      .map((msg) => {
+        const roleLabel =
+          msg.role === 'user' ? 'User' : msg.role === 'assistant' ? 'Assistant' : 'System';
+        return `[${roleLabel}]\n${msg.content}`;
+      })
+      .join('\n\n');
+  }, []);
+
+  const copyMessagesToClipboard = useCallback(async () => {
```
faizanminhas, line 411: this copies _all_ messages? I would expect it to only copy the given message it's on?

faizanminhas:
IMO this should be in a context menu with a "copy conversation" button, not placed like you have

MrFlounder:
why are you not sleeping @faizanminhas

src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
```diff
@@ -395,6 +397,25 @@ export default function EvalOutputPromptDialog({
       ],
     );
 
+  const formatMessagesForCopy = useCallback((messages: Message[]): string => {
```
will-holley, line 400: This doesn't need to be a callback (empty dependency array) & it's single use, so declare it instead within the scope of `copyMessagesToClipboard` i.e. `const formattedText = messages.filter(...).map(...).join(...);`

src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
will-holley: We're using this method elsewhere

faizanminhas: I think it’s usually in a try catch for when there’s an exception?

PR #6527, opened 2025-12-05
fix(blog):  update Tabs Fakier GitHub profile URL

kkahadze:
In order to reproduce the problem, go to [the blog post by Tabs ](https://www.promptfoo.dev/blog/autonomy-agency-ai-secure-llms/)and look at the author image and click on the author link (right below the date at the top). You will find that the image is missing and the link goes to a GitHub profile that doesn't exist.

### DIFF
--- a/site/blog/authors.yml
+++ b/site/blog/authors.yml
@@ -66,7 +66,7 @@ steve:
 
 tabs:
   name: Tabs Fakier
-  title: Founding Developer Advocate
+  title: Contributor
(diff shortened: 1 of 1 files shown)


site/blog/authors.yml
```diff
@@ -67,7 +67,7 @@ steve:
 tabs:
   name: Tabs Fakier
   title: Founding Developer Advocate
```
mldangelo:
```suggestion
  title: Contributor
```

PR #6548, opened 2025-12-07
feat: tool calling f1 score

### LINKED ISSUE #6547: [Feature] F1 score metric for tool calling evaluations
My use case requires a straightforward way to measure tool calling accuracy using F1 score, it would be great to have `promptfoo` support this.

When testing LLMs that call tools/functions, I'd like to know how well the model picks the right tools. For context, [F1 score](https://en.wikipedia.org/wiki/F-score) is a standard metric that combines precision (of the tools called, how many were correct?) and recall (of the expected tools, how many were called?) into a value between 0 and 1.

Something like:

```yaml
assert:
  - type: tool-call-f1
    value:
      - get_weather
    threshold: 0.8
```

This would compare the tools the model actually called against what I expected and return an F1 score.

stefanoamorelli:
Adds an F1 score metric to evaluate LLM tool call accuracy. The F1 score is the harmonic mean of precision (fraction of called tools that were expected) and recall (fraction of expected tools that were called).

For example:

- If the list of expected tool calls is: `["get_weather", "get_time"]`;
- and the list of called tools is: `["get_weather", "send_email"]`;
- then the precision is: 1/2, recall: 1/2, and the F1 score is: 0.5.

```yaml
  tests:
    - vars:
        question: "What's the weather and time in Paris?"
      assert:
        - type: tool-call-f1
          value: ["get_weather", "get_time"]
          threshold: 0.5 # it passes
```

Where the eval'd output is:

```json
{
  "tool_calls": [
    { "function": { "name": "get_weather" } },
    { "function": { "name": "send_email" } }
  ]
}
```

It supports OpenAI's `tool_calls` format and direct array.

The threshold defaults to 1.0, meaning the test passes only if all expected tools are called with no extra ones.

---

Closes #6547

### DIFF
--- a/site/docs/configuration/expected-outputs/deterministic.md
+++ b/site/docs/configuration/expected-outputs/deterministic.md
@@ -575,6 +575,8 @@ tests:
 
 The `tool-call-f1` assertion computes the [F1 score](https://en.wikipedia.org/wiki/F-score) comparing the set of tools called by the LLM against an expected set of tools. This metric is useful for evaluating agentic LLM applications where you want to measure how accurately the model selects the right tools.
 
+This assertion supports multiple provider formats including OpenAI, Anthropic, and Google/Vertex.
+
 The F1 score is the harmonic mean of precision and recall, originally introduced by [van Rijsbergen (1979)](http://www.dcs.gla.ac.uk/Keith/Preface.html) for information retrieval evaluation:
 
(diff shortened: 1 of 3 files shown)


mldangelo:
Thank you! Please reach out (michael @ promptfoo.dev) for some swag!

PR #6553, opened 2025-12-08
fix: allow object variables to pass through templates

### LINKED ISSUE #2266: Passing Function Call Schemas as Variables in OpenAI Function Configurations
Currently, function call schemas cannot be passed dynamically as variables when configuring OpenAI function calls. This limitation makes it difficult to reuse or dynamically generate schemas in a clean and modular manner. For example, attempting to define a schema in a `vars` block and reference it in a function configuration results in an output of `"parameters":{"[object Object]":null}` instead of the expected schema.

This behavior is frustrating as it reduces flexibility and increases redundancy in defining schemas, particularly for use cases where schemas might need to be generated or reused across multiple tools or scenarios.

---

I would like the ability to pass schemas as variables in function call configurations. For instance, the following configuration should work as expected:

```yaml
providers:
  - id: openai:chat:gpt-4o-audio-preview
    config:
      tools: [
        {
          "type": "function",
          "function": {
            "name": "transcribe_audio",
            "description": "Transcribe the audio",
            "parameters": {{ schema | dump | safe }}
          }
        }
      ]
      tool_choice: 'auto'

tests:
  - vars:
      schema: {
        "type": "object",
        "properties": {
          "transcription": {
            "type": "string",
            "description": "The transcription of the audio"
          }
        },
        "required": ["transcription"]
      }
```

The schema variable should correctly resolve and be passed into the `parameters` field without breaking.

---

**Describe alternatives you’ve considered**  
1. **Hardcoding schemas directly into function configurations**: While this works, it results in code duplication and makes schema management cumbersome, especially for large projects with multiple tools or similar schema structures.  
2. **Preprocessing the configuration**: Writing custom scripts to inject schemas before runtime, but this adds unnecessary complexity and reduces the maintainability of configurations.  

Neither of these alternatives is as clean or efficient as supporting schema variables natively.

---

**Additional context**  
Here’s an example of the error observed when attempting to pass a schema variable:  

```yaml
"parameters": {
  "[object Object]": null
}
```

Allowing schemas to be passed as variables would enable dynamic and modular schema configurations, improving development workflows and reducing redundancy. This feature would be particularly useful for scenarios requiring reusable or dynamically generated function call schemas.  

Thank you for considering this feature! Let me know if further clarification or examples are needed.

mldangelo:
## Summary

- Preserve object and array variables when they are used as structured tool-schema templates.
- Keep filter semantics deliberate: bare `{{ schema }}` and the existing `{{ schema | dump | safe }}` schema-injection form return structured values, while `dump` alone and transforming filters still run through Nunjucks.
- Harden direct lookup against inherited prototype keys and expose the structured return possibility in the TypeScript string-call contract.
- Document reusable OpenAI tool schemas supplied through test variables.

Fixes #2266

## Design Notes

- Structured substitution is limited to plain objects and arrays, and only for own variable properties.
- Text-only consumers, including raw HTTP text bodies and redteam purposes, continue to render as text; JSON bodies can accept structured values.
- OpenAI chat request tests cover both supported structured-schema template forms end to end through provider request construction.

## Validation

- Merged current `origin/main` (`1205a2a2be77beb8505731515d0af1ee893cacb0`).
- `npm run l && npm run f` passed (existing complexity advisories only in touched broad modules).
- `npm run tsc` passed.
- Focused renderer, HTTP, OpenAI chat, and redteam generation tests passed (`474/474`).
- Full backend suite passed (`18,199` passed, `10` skipped).
- `npm run build` passed, including the app TypeScript/Vite build.
- `SKIP_OG_GENERATION=true npm --prefix site run build` passed.
- Isolated local CLI eval against a disposable OpenAI-compatible stub passed `1/1`; the captured request contained an object-valued `tools[0].function.parameters` schema and temporary state files were mode `0600`.

## Review Follow-up

- Added own-property protection and regression coverage for prototype keys.
- Preserved explicit `dump | safe` schema compatibility while preventing `dump`-only filter bypass.
- Updated return typing and kept string-only consumers explicitly textual.


ghost:
**⏩ No test execution environment matched (6a336f7db466388493309e2c9fe11cb4d47a152c) [View output ↗](https://app.usetusk.ai/app/testing-commit-check/6cfa8272-aaa1-449b-82e5-e497f57628f1?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

---
<p>

| 78237443bd5a478d24d72227a0e9cfae7cecc89d | ⏩ No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/5f7da9a2-2864-4e1c-8564-5a87d6617598?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 8, 2025 12:13AM |
| 6a336f7db466388493309e2c9fe11cb4d47a152c | ⏩ No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/6cfa8272-aaa1-449b-82e5-e497f57628f1?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 8, 2025 12:19AM |

</p>
</details>

[View output in GitHub ↗](https://github.com/promptfoo/promptfoo/pull/6553/checks?check_run_id=57384631978)

src/util/index.ts
```diff
@@ -527,6 +527,19 @@ export function renderVarsInObject<T>(obj: T, vars?: Record<string, string | obj
...
+    // Check if the entire string is a simple variable reference like "{{ varname }}" or "{{ varname | dump | safe }}"
+    // This allows object variables to be passed through directly instead of being converted to "[object Object]"
+    const simpleVarMatch = obj.match(/^\{\{\s*(\w+)(?:\s*\|\s*[\w\s|]+)?\s*\}\}$/);
+    if (simpleVarMatch) {
+      const varName = simpleVarMatch[1];
+      const varValue = vars[varName];
+      // If the variable is an object (not a string), return it directly
+      // This enables passing schemas and other complex objects through templates
+      if (varValue !== undefined && typeof varValue === 'object' && varValue !== null) {
+        return varValue as unknown as T;
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed on the delivered tip. Structured pass-through is now limited to bare `{{ schema }}` and the issue-compatible `{{ schema | dump | safe }}` marker; ordinary transforming filters still execute through Nunjucks. Regressions cover `dump`-only serialization plus `length`, `first`, and `join` behavior.

mldangelo-oai:
@codex review

mldangelo-oai:
Updated this PR against current `main` and addressed the two remaining schema-preservation review threads.

What changed:
- Avoid re-rendering object-valued injected response schemas after `{{ schema }}` / `{{ schema | dump | safe }}` has already produced structured data.
- Avoid the second render pass for object tool definitions inside arrays while still allowing string file references to be resolved.
- Added regressions for tool schemas, top-level `response_format.schema`, and nested `json_schema.schema` literals that contain `{{...}}` as data.

Validation:
- `npx vitest run test/util/render.test.ts test/util/file.test.ts test/providers/http/core.test.ts test/providers/openai/chat.test.ts test/redteam/commands/generate.test.ts --sequence.shuffle=false` (523 passed)
- `npm run f`
- `npm run l`
- `npm run tsc`
- `npm run build`
- `git diff --check`
- Local CLI smoke eval against a stub OpenAI chat endpoint with isolated `PROMPTFOO_CONFIG_DIR` / `PROMPTFOO_CACHE_PATH`; verified the emitted request keeps `tools[0].function.parameters` as an object and preserves literal `{{literal}}` strings inside schema descriptions/enums.

CI has restarted on `8705c38a177f1e7b85df0de9f3c2fff72cd62389`; moving to the next PR per the queue workflow instead of waiting here.

src/util/render.ts
```diff
@@ -114,22 +161,57 @@ export function renderEnvOnlyInObject<T>(
...
+  vars: Record<string, VarValue> | undefined,
+  options: RenderVarsInObjectOptions,
+  path: (string | number)[] = [],
+): T | object {
   // Renders nunjucks template strings with context variables
   if (!vars || getEnvBool('PROMPTFOO_DISABLE_TEMPLATING')) {
     return obj;
   }
   if (typeof obj === 'string') {
+    const directObjectReferenceValue = getDirectObjectReferenceValue(obj, vars, options, path);
```
jameshiester-oai, line 193: **[P2] Keep parsed chat text fields on string rendering**

This direct-object branch also runs over JSON chat prompts in `renderPrompt()`. With `[{"role":"user","content":"{{ payload }}"}]` and an object-valued `payload`, the exact-head built CLI sent a plain object as `messages[0].content`; a contract-validating local endpoint returned 400, and the exported result recorded the provider error. OpenAI chat accepts a string or an array of typed content parts here, not an arbitrary object (or a primitive array). Before this branch, Nunjucks kept this field textual. Please make bare structured pass-through opt-in/path-aware like the dump-safe path, or explicitly preserve text semantics for parsed message fields while allowing only valid content-part arrays; add a `renderPrompt`/provider regression for this case.

jameshiester-oai (commented):
Reviewed PR merge-base `c24aa89804d35d6e4233edad80b38d67257cd508` through head `c2f90b9fb9a9fcf5a359b729071d2f4096d5503f` (current `origin/main` was `4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6`). Scope was a fresh full review; no authenticated completion marker existed for this head.

Verdict: the intended OpenAI structured-schema path works and all 41 reported checks are green, but four P2 defects remain on the current head, alongside three independently reproduced P2s already raised in unresolved threads. The branch is behind current main.

### Verified findings

**P2**

- `src/util/render.ts:193`: bare object substitution is enabled globally, so parsed JSON chat text fields can become provider-invalid objects. The built CLI sent a plain object for `messages[0].content`; a contract-validating local OpenAI-compatible endpoint returned 400 and the exported result had `success:false`, score 0, and the provider error. This contradicts the PR's design note that text-only consumers remain text. Inline comment has the reproduction and correction boundary.
- `src/util/render.ts:209`: an own `__proto__` key whose template resolves to an object invokes the legacy prototype setter on the rendered result. The exact-head probe produced no own key and set the result prototype to the supplied schema object, making its marker inherited. Inline comment has the reproduction and safe construction direction.
- `src/util/file.ts:633`: combining a substituted tool array with an inline tool leaves a nested list. `OpenAiChatCompletionProvider.getOpenAiBody()` forwarded `[[sharedA, sharedB], inline]`, which violates the flat tool-list contract. Inline comment has the reproduced shape.
- `src/providers/google/util.ts:1218`: request construction now accepts `functionDeclarations[].parameters: '{{ schema | dump | safe }}'`, but `validateFunctionCall()` reloads the same tools with generic `loadFile()`. A matching function call failed because AJV received the serialized schema and threw `schema must be object or boolean`. The assertion path should use the same schema-aware loader as request construction, with an assertion-level regression test. This is off the changed lines, so it is recorded here rather than as an inline comment.

Three existing unresolved P2 threads remain applicable and were not duplicated:

- https://github.com/promptfoo/promptfoo/pull/6553#discussion_r3327819163: AI Studio file-loaded response schemas re-render literal `{{...}}` text inside an injected schema. The unified Google/Vertex path already had this behavior; the new AI Studio helper introduces the regression there.
- https://github.com/promptfoo/promptfoo/pull/6553#discussion_r3327819164: loading the outer response-format object no longer renders nested object-schema variables.
- https://github.com/promptfoo/promptfoo/pull/6553#discussion_r3327819166: native Bedrock `toolSpec.inputSchema.json` is omitted from the structured schema paths and remains a string.

The material simplification is to make structured substitution an explicit, path-aware caller policy rather than a global bare-template behavior with text callers opting back out individually. Missing coverage is a text-field matrix for parsed chat prompts, own `__proto__` output keys, mixed array/object tool composition, and request-plus-assertion round trips for each supported provider. The PR title follows the repository convention; this is not primarily a redteam change.

### Verification

- `npm ci` under Node 24.16.0 / npm 11.13.0: exit 0.
- Changed-surface suite (`test/util/render.test.ts`, `test/util/file.test.ts`, HTTP core, OpenAI chat/responses, three Google providers, and redteam generate): 9 files / 761 tests passed.
- Adjacent behavior suite (`test/providers/azure/chat.test.ts`, `test/providers/google/util.test.ts`, `test/assertions/google.test.ts`, `test/evaluatorHelpers.test.ts`): 4 files / 321 tests passed.
- `npm run tsc`: exit 0.
- `npm run build`: exit 0, including the app and postbuild.
- `git diff --check c24aa898...c2f90b9f`: exit 0; detached worktree remained clean.
- Built CLI intended-path probe: `npm run local -- eval -c /private/tmp/pr6553-schema-e2e.yaml --no-cache -o /private/tmp/pr6553-schema-e2e.json` exited 0; 1/1 passed, score 1, output `ok`, and the stub observed object-valued `functions[0].parameters`.
- Built CLI regression probe: `npm run local -- eval -c /private/tmp/pr6553-chat-content-e2e.yaml --no-cache -o /private/tmp/pr6553-chat-content-e2e.json` exited 100; 1/1 errored, and the stub observed plain-object chat content before returning the contract error.
- Direct exact-head probes reproduced the Google false-negative assertion, local prototype mutation, and nested OpenAI tool request body.
- GitHub CI: 41/41 reported checks successful, including Linux/macOS/Windows tests, builds, docs, integration/smoke, and CodeQL.

Limitations: `/pr-review-toolkit:review-pr all parallel` is unavailable in this environment, so I completed the same six review lenses manually and reconciled three independent read-only reviews. No external provider was called; end-to-end QA used a disposable local OpenAI-compatible contract stub with a dummy key. I did not rerun the docs build or full backend suite locally because the focused, type, build, and CLI checks passed and current CI covers both. I inspected current-main interaction without merging or rebasing it into this exact head.

src/util/render.ts
```diff
@@ -114,22 +161,57 @@ export function renderEnvOnlyInObject<T>(
...
-    return obj.map((item) => renderVarsInObject(item, vars)) as unknown as T;
+    return obj.map((item, index) =>
+      renderVarsInObjectInternal(item, vars, options, [...path, index]),
+    ) as unknown as T;
   }
   if (typeof obj === 'object' && obj !== null) {
     const result: Record<string, unknown> = {};
     for (const key in obj) {
-      result[key] = renderVarsInObject((obj as Record<string, unknown>)[key], vars);
+      result[key] = renderVarsInObjectInternal(
```
jameshiester-oai, line 209: **[P2] Preserve `__proto__` as an own data key**

Structured substitution makes this assignment unsafe for an own `__proto__` key: assigning an object invokes the legacy prototype setter instead of creating a property. An exact-head probe using `JSON.parse('{"__proto__":"{{ schema }}"}')` produced `hasOwn:false`, no own keys, and a result prototype equal to the supplied schema object, so `marker` became inherited. This turns template data into result prototype state and corrupts schemas/configs that legitimately contain that key. Construct the result with a null prototype or define this key as an own data property (the safe pattern already used in `src/util/file.ts`), and add an object-valued regression.

src/util/file.ts
```diff
@@ -580,7 +629,9 @@ export async function maybeLoadToolsFromExternalFile(
   // Handle arrays by recursively processing each item
   if (Array.isArray(rendered)) {
     const results = await Promise.all(
-      rendered.map((item) => maybeLoadToolsFromExternalFile(item, vars)),
+      rendered.map((item) =>
+        maybeLoadToolsFromExternalFile(item, typeof item === 'string' ? vars : undefined),
```
jameshiester-oai, line 633: **[P2] Flatten each substituted tool-list element**

A substituted array can now appear as one item within a larger list, but the downstream flattening only runs when every result is an array. For `tools: ['{{ sharedTools }}', inlineTool]`, the exact helper and `OpenAiChatCompletionProvider.getOpenAiBody()` both produced `[[sharedA, sharedB], inlineTool]`; the nested array is sent unchanged and is not a valid tool entry. Flatten array-valued results individually (while preserving object entries) and cover mixed substituted-array plus inline-tool composition.

PR #6573, opened 2025-12-08
fix: use correct drizzle migrations path for npm/npx installs

yash2998chhabria:
## Summary
- Fixes database migration error when running via `npx promptfoo@latest`
- The `dist/src` path check was looking for `drizzle/` at package root, but it's at `dist/drizzle/` in the published npm package

## Problem
Running `npx promptfoo@latest` fails with:
```
Error: Can't find meta/_journal.json file
```

This happens because only `dist/` is published to npm (per `package.json` `files` field), so `drizzle/` doesn't exist at the package root - only `dist/drizzle/` does.

## Test plan
- [x] Updated unit tests pass
- [ ] Test with `npx promptfoo@latest` after release

### DIFF
--- a/src/migrate.ts
+++ b/src/migrate.ts
@@ -54,10 +54,10 @@ export async function runDbMigrations(): Promise<void> {
         const dir = getCurrentDir();
         let migrationsFolder: string;
         if (dir.includes('dist/src')) {
-          // When running from bundled server (e.g., dist/src/server/index.js)
-          // Navigate to project root and find drizzle folder
+          // When running from bundled dist (e.g., npx promptfoo or dist/src/main.js)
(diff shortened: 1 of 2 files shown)


yash2998chhabria:
## Note on `security-scan` CI failure

The `security-scan` check will fail because it uses `promptfoo/code-scan-action@v0` which installs promptfoo from npm. The published v0.120.0 has this exact bug - it's a chicken-and-egg situation.

**This PR fixes the bug**, but the CI that validates it uses the broken published version.

Once this fix is released, the security-scan will work again for all PRs.

mldangelo:
I'm QAing this now. Let's cut a patch for it.

PR #6597, opened 2025-12-09
fix(redteam): improve error handling for HuggingFace dataset fetching

minhle1291:
## Summary

Improves error handling and retry behavior for HuggingFace dataset-based plugins (vlguard, toxic-chat, unsafebench) to handle transient infrastructure issues more gracefully.

- Add clear error message for gated VLGuard dataset (401/403) with link to HuggingFace approval page
- Return empty array instead of throwing on dataset fetch failures (graceful degradation)
- Enable 5xx retry by default (bounded to 4 retries with exponential backoff)
- Detect HTML error pages returned instead of JSON with clearer error messages
- Truncate long error responses to 500 chars for readability

## Test plan

- [x] vlguard.test.ts: 32 tests pass
- [x] fetch.test.ts: 58 tests pass
- [x] cache.test.ts: 19 tests pass


src/redteam/plugins/imageDatasetPluginBase.ts
```diff
@@ -96,8 +96,8 @@ export abstract class ImageDatasetPluginBase<
 
       if (records.length === 0) {
         const errorMessage = this.getNoRecordsErrorMessage();
-        logger.error(`[${this.getLogPrefix()}] ${errorMessage}`);
-        throw new Error(errorMessage);
+        logger.warn(`[${this.getLogPrefix()}] ${errorMessage}`);
```
will-holley, line 99: We should be consistent and log these as errors or warnings but not mix the two.

will-holley (approved):
Didn't test locally but the code lgtm aside from mixing the error msg types

src/redteam/plugins/vlguard.ts
```diff
@@ -229,6 +229,12 @@ export class VLGuardDatasetManager extends ImageDatasetManager<VLGuardInput> {
         headers,
       });
 
+      if (response.status === 401 || response.status === 403) {
```
mldangelo, line 234: make this DRY

src/util/fetch/index.ts
```diff
@@ -237,7 +237,7 @@ export async function fetchWithRetries(
     try {
       response = await fetchWithTimeout(url, options, timeout);
 
-      if (getEnvBool('PROMPTFOO_RETRY_5XX') && response.status >= 500 && response.status < 600) {
+      if (getEnvBool('PROMPTFOO_RETRY_5XX', true) && response.status >= 500 && response.status < 600) {
```
mldangelo: Can you talk about your decision here? This is load bearing and I want to make sure we are considering all of the side effects.

mldangelo (changes requested):
Code seems good. Requesting changes because of the PROMPTFOO_RETRY_5XX - Would like to understand why this is necessary.

src/util/fetch/index.ts
minhle1291: I decided to enforce this as the new default because I figured it would smooth out any temporary or one-off issue on the HF infra side.  It is part of a 4-time retry with (short) exponential backoff between each retry.

minhle1291: @mldangelo Thoughts on this retry-by-default configuration?

mldangelo: 500 errors are almost always a bug or misconfiguration and not a transient issue. With retries enabled by default, a 100-test-case eval against a broken endpoint goes from failing in ~2 minutes to failing in ~2 hours.

src/util/fetch/index.ts
```diff
@@ -237,7 +237,11 @@ export async function fetchWithRetries(
     try {
       response = await fetchWithTimeout(url, options, timeout);
 
-      if (getEnvBool('PROMPTFOO_RETRY_5XX') && response.status >= 500 && response.status < 600) {
+      if (
+        getEnvBool('PROMPTFOO_RETRY_5XX', true) &&
```
will-holley: Please revert setting this boolean and see Michael's comment. I'm in favor of documenting it (the comment) explicitly within the code to provide context to coding agents.

will-holley (changes requested):

src/cache.ts
```diff
@@ -210,7 +225,7 @@ export async function fetchWithCache<T = any>(
       throw new Error(
         `Error parsing response from ${url}: ${
           (err as Error).message
-        }. Received text: ${responseText}`,
+        }. Received text: ${responseText.slice(0, 500)}`,
```
will-holley, line 228: Why the slice?

minhle1291: @will-holley I had set 500-char limit for JSON errors and 200-char limit for HTML errors that returned full HTML error pages, which flooded my CLI.

Do you think we should revert to the full error message?

PR #6606, opened 2025-12-10
fix(esm): restore process.mainModule.require compatibility for inline transforms

mldangelo:
## Summary

- Fixes breaking change in 0.120.x where `process.mainModule.require` stopped working in inline JavaScript code
- Creates a process shim using a Proxy that provides `process.mainModule.require` via `createRequire()`
- Fully backwards compatible - no user migration required

## Problem

After the ESM migration, users with inline code like this started getting errors:

```javascript
// In extensions, transforms, or assertions
const require = process.mainModule.require;
const fs = require('node:fs');
```

Error: `TypeError: Cannot read properties of undefined (reading 'require')`

This is because `process.mainModule` is `undefined` in ESM (it's a CommonJS-only feature).

## Solution

Create a Proxy that wraps the global `process` object and intercepts `mainModule` access:

```javascript
const processWithMainModule = new Proxy(process, {
  get(target, prop) {
    if (prop === 'mainModule') {
      return { require: esmRequire, /* ... */ };
    }
    return Reflect.get(target, prop);
  },
});
```

The shim is passed to all inline `new Function()` calls, making `process.mainModule.require` work again while preserving all other `process` properties like `process.env`.

## Files Changed

| File | Purpose |
|------|---------|
| `src/util/transform.ts` | Added `getProcessShim()` export + applied to extension hooks/transforms |
| `src/assertions/javascript.ts` | Applied shim to inline JavaScript assertions |
| `src/providers/httpTransforms.ts` | Applied shim to HTTP request/response transforms |
| `src/providers/websocket.ts` | Applied shim to WebSocket transforms |
| `test/util/transform.test.ts` | Added ESM compatibility tests |

## Test plan

- [x] All existing transform tests pass (27 tests)
- [x] All existing JavaScript assertion tests pass (39 tests)
- [x] All HTTP provider tests pass (168 tests)
- [x] All WebSocket provider tests pass (95 tests)
- [x] New ESM compatibility tests added and passing
- [x] Manual testing with reproduction config using `process.mainModule.require` pattern

### DIFF
--- a/src/util/transform.ts
+++ b/src/util/transform.ts
@@ -82,7 +82,12 @@ async function getJavascriptTransformFunction(
 ): Promise<Function> {
   const requiredModule = await importModule(filePath);
 
-  if (functionName && typeof requiredModule[functionName] === 'function') {
+  // Validate that functionName is an own property to prevent prototype pollution attacks
+  if (
+    functionName &&
+    Object.prototype.hasOwnProperty.call(requiredModule, functionName) &&
+    typeof requiredModule[functionName] === 'function'
+  ) {
     return requiredModule[functionName];
   } else if (typeof requiredModule === 'function') {
     return requiredModule;
--- a/test/evaluator.test.ts
+++ b/test/evaluator.test.ts
@@ -32,6 +32,8 @@ vi.mock('../src/util/transform', () => ({
     OUTPUT: 'output',
     VARS: 'vars',
   },
+  // Provide a process shim for ESM compatibility in inline JavaScript code
+  getProcessShim: vi.fn().mockReturnValue(process),
   transform: vi.fn().mockImplementation(async (code, input, context, _skipWrap, _inputType) => {
     if (typeof code === 'string' && code.includes('vars.transformed = true')) {
       return { ...input, transformed: true };


JustinBeckwith (approved):
Slightly mixed feelings on this, given it's a deprecated node.js feature - but agree we didn't intend to break it.

mldangelo:
## Addressing CI failures and CodeQL alerts

I've pushed a fix (0fe9485ba) that addresses:

1. **Test failures**: Added `getProcessShim` to the transform mock in `test/evaluator.test.ts`. The evaluator tests mock the transform module, so the new export was `undefined`, causing JavaScript assertions to fail.

2. **CodeQL High severity (Unvalidated dynamic method call)**: Added `Object.prototype.hasOwnProperty.call()` check in `getJavascriptTransformFunction` to prevent prototype pollution attacks when loading named functions from modules.

### Regarding the CodeQL Critical "Code injection" alert

This alert flags the intentional use of `new Function()` for inline JavaScript code execution. This is **by design** - promptfoo allows users to write inline JavaScript for:
- Extensions (beforeAll, beforeEach, afterEach, afterAll hooks)
- Transform functions
- JavaScript assertions

The suggested autofix would require all code to be in external files (file:// references only), which would be a breaking change that removes legitimate functionality. Inline code execution is a documented feature used by customers for simple transforms without needing separate files.

This pattern has existed in the codebase before this PR - we're just adding backwards compatibility for `process.mainModule.require` which was broken by the ESM migration.

PR #6635, opened 2025-12-11
feat(examples): Add pdf file example

bb220:
I think it would be useful to include a raw pdf file example because of the latest models' abilities to detect visual information in documents. The example pdf file includes text + images. It is pg 5 from https://arxiv.org/pdf/2409.00001.

_UI View Results_
image


mldangelo (approved):
Thank you @bb220! Please reach out if you'd like some promptfoo swag!

PR #6638, opened 2025-12-11
feat(types): add pluginId and strategyId to EvaluateTableOutput

minhle1291:
## Summary

Adds optional `pluginId` and `strategyId` fields to the `EvaluateTableOutput` interface.

These fields already exist in `EvalResult` but weren't exposed in the table output structure, requiring consumers to inject them into `metadata` as a workaround.

Helps to unblock https://github.com/promptfoo/promptfoo-cloud/pull/2677 without needing to make large changes to the metadata model.

## Changes

- Added `pluginId?: string | null` to `EvaluateTableOutput`
- Added `strategyId?: string | null` to `EvaluateTableOutput`
- Added test validating the new fields work correctly

## Motivation

This enables cleaner access patterns:
- Before: `output.metadata?.pluginId`
- After: `output.pluginId`

Used by promptfoo-cloud for CSV export with framework mapping.

## Test plan

- [x] Added unit test for new fields
- [x] All existing tests pass (27/27)
- [x] Verified that CSV export presents plugin and strategy columns

### DIFF
--- a/src/server/utils/evalTableUtils.ts
+++ b/src/server/utils/evalTableUtils.ts
@@ -94,7 +94,11 @@ export function evalTableToCsv(
     if (isRedteam) {
       const firstOutputMetadata = row.outputs[0]?.metadata;
       for (const key of redteamKeys) {
-        const value = firstOutputMetadata?.[key];
+        let value = firstOutputMetadata?.[key];
+        // Default strategyId to 'basic' for strategy-less tests
+        if (key === 'strategyId' && (value === null || value === undefined)) {
+          value = 'basic';
+        }
         if (value === null || value === undefined) {
           rowValues.push('');
         } else if (
--- a/test/server/utils/evalTableUtils.test.ts
+++ b/test/server/utils/evalTableUtils.test.ts
@@ -657,6 +657,91 @@ describe('evalTableUtils', () => {
           expect(lines[1]).toBeDefined();
         }
       });
+
+      it('should include pluginId and strategyId from metadata in CSV output', () => {
+        // This test validates that pluginId and strategyId are read from metadata
+        // for CSV export, consistent with the promptfoo-cloud approach
+        const outputWithMetadataIds: EvaluateTableOutput = {
+          id: 'test-id',
+          pass: true,
+          text: 'Test output',
+          prompt: 'Test prompt',
+          score: 1,
+          cost: 0,
+          latencyMs: 100,
+          namedScores: {},
+          failureReason: ResultFailureReason.NONE,
+          testCase: { vars: {} },
+          metadata: {
+            pluginId: 'harmful:violent-crime',
+            strategyId: 'jailbreak',
+          },
+        };
+
+        // Verify the fields are accessible via metadata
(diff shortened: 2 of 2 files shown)


src/types/index.ts
```diff
@@ -304,10 +304,12 @@ export interface EvaluateTableOutput {
   metadata?: Record<string, any>;
   namedScores: Record<string, number>;
   pass: boolean;
+  pluginId?: string | null;
   prompt: string;
   provider?: string;
   response?: ProviderResponse;
   score: number;
+  strategyId?: string | null;
```
faizanminhas: I think we probably don't want the `| null`, unless we need it for type checking?

will-holley:
Are they not already defined on `metadata`?

src/types/index.ts
```diff
@@ -304,10 +304,12 @@ export interface EvaluateTableOutput {
   metadata?: Record<string, any>;
   namedScores: Record<string, number>;
   pass: boolean;
+  pluginId?: string;
   prompt: string;
   provider?: string;
   response?: ProviderResponse;
   score: number;
+  strategyId?: string;
```
mldangelo: Unlike pluginId, EvalResult doesn't extract strategyId to a top-level property, it only exists in metadata. So this field won't be populated through convertEvalResultToTableCell(...result) spread.

src/server/utils/evalTableUtils.ts
```diff
@@ -92,9 +92,18 @@ export function evalTableToCsv(
...
-      const firstOutputMetadata = row.outputs[0]?.metadata;
+      const firstOutput = row.outputs[0];
+      const firstOutputMetadata = firstOutput?.metadata;
       for (const key of redteamKeys) {
-        const value = firstOutputMetadata?.[key];
+        // Check top-level fields first for pluginId and strategyId, then fall back to metadata
+        let value;
+        if (key === 'pluginId' && firstOutput?.pluginId !== undefined) {
+          value = firstOutput.pluginId;
+        } else if (key === 'strategyId' && firstOutput?.strategyId !== undefined) {
```
mldangelo: For strategyId, this branch (firstOutput?.strategyId !== undefined) will never be hit through normal promptfoo flow since EvalResult doesn't extract strategyId from metadata

minhle1291:
## Updated approach based on review feedback

I've updated this PR to use the **metadata-only approach** as suggested by @will-holley and @mldangelo.

### Changes from previous version:
- ❌ Removed top-level `pluginId` and `strategyId` fields from `EvaluateTableOutput` type
- ✅ Read `pluginId` and `strategyId` from `output.metadata` (existing behavior)
- ✅ Default `strategyId` to `'basic'` for strategy-less tests (per @mldangelo's request)

### Files changed:
| File | Change |
|------|--------|
| `src/server/utils/evalTableUtils.ts` | Default `strategyId` to `'basic'` when missing |
| `test/server/utils/evalTableUtils.test.ts` | Tests for metadata access and 'basic' default |

This aligns with the promptfoo-cloud approach in PR #2677.

PR #6639, opened 2025-12-12
fix(redteam): add goal context to all plugin graders

yash2998chhabria:
## Summary
- Add the attacker's underlying goal to all plugin grader rubrics via the base class
- Enables graders to better detect jailbreak attacks by understanding the true attack intent
- Uses reference-style format based on A/B testing results showing improved detection rates

## Changes
Modified `src/redteam/plugins/base.ts` to append goal context to the final rubric:
```
[Reference] True attack intent: {goal}
```

Goal context is only added when there's an extracted goal different from the prompt (e.g., for jailbreak strategies).

### A/B Testing Results
| Variant | Format | Pass Rate |
|---------|--------|-----------|
| Baseline | None | 57.14% |
| XML Tag | `<AttackGoal>` | 55.36% |
| **Reference** | `[Reference] True attack intent:` | **51.79%** |

Lower pass rate = grader detected more successful jailbreak attacks. Reference style caught the most.

## Test plan
- [x] Code compiles successfully

First ran small evals with different goal prompt position:
Screenshot 2025-12-12 at 11 58 27 AM

Finally, shortlisted approaches and ran with more tests
Screenshot 2025-12-12 at 11 56 17 AM


mldangelo (commented):
Looks good overall, nice cleanup to centralize the goal context.

One thing - the new `hasExtractedGoal` logic doesn't have dedicated unit tests in base.test.ts. The existing tests that use goal (like "should render the rubric with correct variables including goal") don't actually check for the `[Reference] True attack intent:` string getting appended.

Also minor: the test comment at base.test.ts:1687 says "timestamp is last" but goalContext now comes after it. And the timestamp regex tests use `$` anchor which would fail if any test case had goal != prompt.

src/redteam/plugins/base.ts
```diff
@@ -473,12 +473,20 @@ export abstract class RedteamGraderBase {
 
     const timestampString = `\n\nCurrent timestamp: ${vars.timestamp}`;
 
+    // Add goal context when there's an actual extracted goal (different from the prompt)
+    // This helps the grader understand the true attack intent when evaluating obfuscated prompts
+    const hasExtractedGoal = test.metadata?.goal && test.metadata.goal !== prompt;
+    const goalContextString = hasExtractedGoal
+      ? `\n\n[Reference] True attack intent: ${test.metadata.goal}`
+      : '';
```
mldangelo, line 520: nit: might be worth adding a quick test case for this - verifying the string gets appended when goal \!= prompt, and does not when goal == prompt

yash2998chhabria: Thanks for catching them :))

src/redteam/plugins/base.ts
```diff
@@ -473,12 +473,20 @@ export abstract class RedteamGraderBase {
...
+
     const finalRubric =
```
mldangelo: nit, use a template string

yash2998chhabria:
Screenshot 2025-12-19 at 3 17 32 PM
Compared to 49% On latest Foundation model report

PR #6648, opened 2025-12-12
feat(app): enable React Compiler for automatic memoization

mldangelo:
## Summary

Had some extra tokens to burn, so I wanted to experiment with enabling React Compiler for our React 19 frontend. This is a **conservative first pass** - additional optimizations can be made later once we verify this works well in production.

**What this does:**
- Enables `babel-plugin-react-compiler` in the Vite build pipeline
- Adds automatic memoization for components and hooks without manual `useMemo`/`useCallback`

**Conservative approach taken:**
- Added `'use no memo'` directives only where intentional dependency omissions exist (4 components)
- Kept `React.memo` on `TruncatedText` since it has O(n) recursive `textLength()` computation in table cells
- Removed only clearly redundant `useCallback` hooks that the compiler now handles
- Did not aggressively remove all manual memoization - left room for future optimization

## Metrics Measured

Ran benchmark tests measuring render performance:

| Metric | Value |
|--------|-------|
| Memoization Ratio | ~44-50% |
| Expensive child updates | Fast (memoized) |
| Callback stability | Children skip re-render |

The **memoization ratio** = `actualDuration / baseDuration` - lower is better. With the compiler enabled, React skips re-rendering unchanged subtrees, reducing actual render time significantly vs the base (full re-render) time.

## Test plan

- [x] All 2,139 tests pass
- [x] Lint passes
- [x] Build succeeds  
- [x] Dev server starts and renders correctly
- [x] Benchmark tests show expected memoization improvements
- [ ] Manual smoke test of key UI flows (eval results table, redteam setup)

### DIFF
--- a/src/app/src/pages/eval/components/ResultsTable.tsx
+++ b/src/app/src/pages/eval/components/ResultsTable.tsx
@@ -45,7 +45,12 @@ import { useFilterMode } from './FilterModeProvider';
 import MarkdownErrorBoundary from './MarkdownErrorBoundary';
 import { useResultsViewSettingsStore, useTableStore } from './store';
 import TruncatedText from './TruncatedText';
-import type { CellContext, ColumnDef, VisibilityState } from '@tanstack/table-core';
+import type {
+  CellContext,
+  ColumnDef,
+  ColumnSizingState,
+  VisibilityState,
+} from '@tanstack/table-core';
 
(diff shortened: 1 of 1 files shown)


faizanminhas (approved):
mamma mia, that's a spicy PR

PR #6656, opened 2025-12-13
ci: speed up Windows tests with sharding and remove unused setup

JustinBeckwith:
## Summary

- Add test sharding (3 shards) for Windows runners to reduce wall-clock time from ~8 minutes to ~3-4 minutes per shard
- Remove Python/Ruby setup on Windows/macOS since unit tests mock these dependencies (saves ~25-50s per job)

## Background

Analysis of the last 50 CI runs showed Windows tests are the critical path bottleneck:

| Step | Ubuntu | Windows | Delta |
|------|--------|---------|-------|
| npm ci | 43s | 2m 22s | +99s |
| Test | 2m 53s | 4m 59s | +126s |
| **Total** | ~4 min | ~8 min | **+4 min** |

### Test Sharding
Windows tests now run in 3 parallel shards using Vitest's built-in `--shard` option. Each shard runs ~1/3 of the 516 test files.

### Remove Unnecessary Setup
Python and Ruby are only needed for integration tests (which run on Ubuntu). Unit tests mock these dependencies, so we skip the ~25-50s setup time on Windows/macOS.

## Expected Impact

| Metric | Before | After |
|--------|--------|-------|
| CI wall-clock time | ~8-9 min | ~4-5 min |
| Total jobs | 8 | 14 |
| Billable minutes | ~24 min | ~35 min |

Tradeoff: More billable minutes but **~4 minutes faster** CI completion.

## Test plan

- [ ] CI passes on this PR
- [ ] Windows shards complete in ~3-4 minutes each
- [ ] All test files are covered across the 3 shards


.github/workflows/main.yml
```diff
@@ -47,12 +80,16 @@ jobs:
         if: matrix.node == '24.x'
         run: npm install -g npm@latest
 
+      # Python and Ruby are only needed on Ubuntu for integration tests
```
mldangelo: We keep running into windows compatibility issues. It would be good if we actually ran python windows tests in the ci

JustinBeckwith: Will do, just kind of messing around right now, should have opened this in draft mode

PR #6668, opened 2025-12-14
fix(site): remove homepage code box scrollbars

kkahadze:
The [homepage](https://www.promptfoo.dev/) has a code box for the 'redteam setup' command which has unnecessary scrollbars. Even when the page width is minimized to the point where you think a scrollbar could be needed, the text wraps so there is no need for scrollbars. 

Before:
Screenshot 2025-12-13 at 9 15 28 PM

After:
Screenshot 2025-12-13 at 9 16 13 PM

I removed the scrollbars for a cleaner look.

### DIFF
--- a/site/src/pages/index.module.css
+++ b/site/src/pages/index.module.css
@@ -684,7 +684,7 @@ html {
   font-family: 'Courier New', Courier, monospace;
   font-size: 0.95rem;
   font-weight: 600;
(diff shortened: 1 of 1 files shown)


mldangelo:
Thank you @kkahadze!!

mldangelo (approved):

PR #6674, opened 2025-12-14
feat(dist): add pip install and standalone binary distribution

mldangelo:
## Summary

Adds standalone distribution paths for Promptfoo:

- `curl -fsSL https://promptfoo.dev/install.sh | bash` and PowerShell installers
- platform-specific standalone Node SEA archives
- `pip install promptfoo` wheels carrying the same standalone runtime

## Design and Safety Improvements

- Treats the standalone build as a core-runtime distribution: optional Node-backed providers such as local Transformers, browser automation, and selected cloud SDKs remain available through npm/npx.
- Ships `tsx` alongside the executable so TypeScript configs load correctly; artifact verification now fails the build if the SEA executable cannot start.
- Omits the unused optional ONNX/Transformers payload from standalone packages, reducing the generated Linux bundle from about 737 MB to 243 MB locally.
- Places migrations, tracing protos, and the web UI where runtime resolution expects them and validates those paths in executable smoke tests.
- Uses isolated private config/cache state for artifact verification, smoke tests, Python-wheel tests, and installer verification.
- Makes shell npm fallback failures propagate correctly and keeps installer downloads in unpredictable temporary paths.
- Uses the accurate Linux wheel compatibility tag, `manylinux_2_28_*`, for the Node 24 runtime floor.
- Gates release publication jobs on successful prerequisite builds and ensures checkouts used for distribution do not persist credentials.

## Test Plan

- [x] Merged current `origin/main` (`1205a2a2be77beb8505731515d0af1ee893cacb0`)
- [x] `npm run f && npm run l`
- [x] `npm run tsc`
- [x] `actionlint .github/workflows/release-please.yml`
- [x] `npm run build`
- [x] `npx vitest --run test/migrate.test.ts --sequence.shuffle=false` (15 tests)
- [x] `npm test -- --run` (18,183 passed; 10 skipped)
- [x] `npm run test:app -- --run` (4,540 passed)
- [x] `npm run bundle:sea` and executable smoke suite (26 passed; Ruby skipped when unavailable)
- [x] Real standalone `eval` with exported result inspection and private database/cache modes
- [x] Shell installer archive install, latest-version resolution, no-user-state verification, and failing npm fallback propagation
- [x] Local Python wheel build/retag/install; verified `promptfoo`, `python -m promptfoo`, and nonzero failure propagation
- [x] Installed standalone archive rejects npm-only local Transformers provider with actionable installation guidance

Windows and non-Linux platform packaging remain covered by the release matrix when this PR runs in GitHub Actions.


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/89ab4434-034d-4d06-955d-790098e752ac?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on 696fcc62b2295e09d4868298f1b13c7cc0416b91. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 713248bf2e7816013d11487517be49b2ba6280eb | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/5911178d-4318-4e37-a3be-da644783e4e9?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 5:17AM |
| acffb98169dbf706ca943410414180b0b01e74d8 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/b1fc41c7-363a-4627-8dd8-39ac616fe498?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 5:20AM |
| 1308c36e00cdc41dd628d6ae5f05b5af8f3583ea | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/41a5a8c0-bb09-4166-9d44-894a173b244f?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 5:23AM |
| e79b94b6af4bf7639fd6471a357b77b27549a33a | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/d6de3be4-e05c-4eaf-bae3-97a7b4cfd91e?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 5:26AM |
| 29dccc870a4f6c5e79d836386be3974764953b96 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/c4419089-dc79-45a6-bc65-4aa42477f607?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 5:36AM |
| a49e3100b7f381824b6459b7a57cd310e4084b7b | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/61fb7c8b-f7c0-4d73-8253-ed29668451d9?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 6:03AM |
| 5e0faa954e2ac84d1195a325b74fe068662ae158 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/bcdc7b0e-ea70-4677-b4af-53ff27a90ecc?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 14, 2025 6:49AM |
| 696fcc62b2295e09d4868298f1b13c7cc0416b91 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/89ab4434-034d-4d06-955d-790098e752ac?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 4:21PM |

</p>
</details>

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR significantly improves the Python package distribution by bundling a Single Executable Application (SEA) binary with embedded Node.js runtime. This eliminates the need for Python users to install Node.js separately. The PR updates the build process to create platform-specific wheels with the appropriate bundled binary for each platform.

## Architecture

**The improvement:**

1. **Bundled SEA binary**:
   - Python package now includes a standalone executable with Node.js embedded
   - No separate Node.js installation required for Python users
   - Binary stored in `python/src/promptfoo/_bin/`

2. **Platform-specific wheels**:
   - Each platform gets a uniquely named wheel:
     - Linux x64: `manylinux_2_17_x86_64`
     - Linux arm64: `manylinux_2_17_aarch64`
     - macOS x64: `macosx_10_15_x86_64`
     - macOS arm64: `macosx_11_0_arm64`
     - Windows x64: `win_amd64`
   - Wheels are retagged from `py3-none-any` to `py3-none-<platform>`

3. **CI workflow updates** (`.github/workflows/release-please.yml`):
   ```yaml
   - name: Build SEA binary
     run: npx tsx scripts/bundle.ts --sea
     env:
       NODE_OPTIONS: --max-old-space-size=4096  # Increased memory for SEA build

   - name: Prepare Python package
     run: |
       mkdir -p python/src/promptfoo/_bin
       cp "bundle/${{ matrix.binary_name }}" python/src/promptfoo/_bin/
       cp -r bundle/assets python/src/promptfoo/_bin/

   - name: Build and tag wheel
     run: |
       uv build --wheel
       python -m wheel tags --platform-tag=${{ matrix.wheel_platform }} --remove "$WHEEL_FILE"
   ```

4. **Simplified CLI wrapper** (`python/src/promptfoo/cli.py`):
   - Removed `find_node()` function that searched for Node.js
   - Now executes the bundled binary directly
   - Reduced from 117 to 80 lines

5. **Updated documentation** (`python/README.md`):
   - Clarified that Node.js is bundled, not required separately
   - Listed supported platforms

## Issues

**No critical issues found.**

This is a well-executed improvement that:

1. **Better UX**: Python users no longer need Node.js installed

2. **Proper platform support**: Each platform gets the correct binary

3. **Simplified code**: Removed complex Node.js discovery logic

4. **Good CI updates**: Properly builds and tags platform-specific wheels

5. **Clear documentation**: Updated README to explain the bundled runtime

**Minor suggestions:**

1. **Consider adding fallback logic**: If the bundled binary is missing or corrupted, provide a helpful error message suggesting installation issues

2. **Test on multiple platforms**: Ensure the bundled binary works correctly on all supported platforms (Linux x64/arm64, macOS x64/arm64, Windows)

3. **Asset handling**: The code checks for `bundle/assets` directory - ensure this is robust and handles missing assets gracefully

4. **Memory usage**: The SEA build uses `--max-old-space-size=4096` - monitor if this is sufficient for larger projects

## Cross-file Impact

**Affected components:**
- `.github/workflows/release-please.yml` — Updated to build SEA binaries and platform-specific wheels
- `python/src/promptfoo/cli.py` — Simplified to execute bundled binary directly
- `python/README.md` — Updated to clarify Node.js is bundled

**New dependencies:**
- SEA binary built with `npx tsx scripts/bundle.ts --sea`
- `wheel` package for retagging wheels with platform tags

**Breaking changes:** None apparent, but this changes the package structure:
- Users will get platform-specific wheels instead of universal wheels
- The bundled binary approach replaces the Node.js discovery mechanism

**Package size impact:**
- Increased: Each wheel now includes the bundled binary and assets
- Trade-off: Larger package size for better UX (no Node.js dependency)

## Assessment

✅ Approve

Well-executed improvement that significantly enhances the Python package user experience:

**Strengths:**
- Eliminates Node.js dependency for Python users
- Proper platform-specific wheel distribution
- Simplified CLI wrapper code
- Good CI workflow updates
- Clear documentation updates

**Key benefit:**
Python users can now install and use promptfoo without worrying about Node.js installation, which removes a significant barrier to entry.

The minor suggestions around error handling, cross-platform testing, and memory monitoring are enhancements but not blockers for merging.

mldangelo-oai:
@codex review

mldangelo-oai:
Updated this branch with current `main` and pushed `5764507016affc2b092abcf99dc2feb1a3153300`.

Changes made:
- Fixed `install.sh`/`site/static/install.sh` so option-only installs such as `--dir /opt/promptfoo` still resolve `latest` instead of treating the option name as the version.
- Refactored the installer smoke fixture setup and added coverage for `install.sh --dir` without an explicit version.

Review-thread notes:
- Rechecked the resolved npm fallback and missing-`--dir` review threads; the smoke suite now covers both.
- The old GitHub Advanced Security permissions thread is outdated, and GitHub rejected resolving it because it is not a resolvable conversation thread. Current `release-please.yml` has workflow-level `permissions: contents: read` plus job-level permissions.

Validation run locally:
- `npx vitest run test/migrate.test.ts --sequence.shuffle=false`
- `bash -n scripts/install.sh && bash -n scripts/smoke-test.sh && cmp -s scripts/install.sh site/static/install.sh`
- `npm run f`
- `npm run l`
- `npm run tsc`
- `npm run build`
- `npm run bundle:sea`
- `./scripts/smoke-test.sh` against the generated SEA executable: 30/30 passed
- Python wheel smoke: built wheel, retagged to `manylinux_2_28_x86_64`, installed into an isolated uv venv, verified `promptfoo --version`, `python -m promptfoo --version`, and missing-config exit propagation
- `npm test -- --run`: 718 files passed, 18,196 tests passed
- `git diff --check`

mldangelo-oai:
Updated this branch with current `main` and pushed `d7dba5bde4a7c0ba47bcc4bd428e36dcd118b3b4`.

Production fixes added during review:
- Corrected the advertised runtime/platform floor: bundled Node 24 requires macOS 13.5+, and Python wheels now conservatively target macOS 14+ because wheel tags cannot express 13.5.
- Made `install.sh` and its site mirror fall back to npm on unsupported or unidentifiable macOS versions, with regression coverage for macOS 13.4.
- Found that a host-built `better-sqlite3.node` could require `GLIBC_2.38` / `GLIBCXX_3.4.31` while the published Linux artifacts promise `manylinux_2_28`. Linux release jobs now rebuild the native addon inside digest-pinned manylinux_2_28 images and run a new ELF compatibility guard.
- Pinned newly introduced publication tooling (`postject`, Hatchling, uv, and wheel) to make release artifacts reproducible.

Validation completed locally:
- `npm run f`, `npm run l`, `git diff --check`
- `npx vitest run test/migrate.test.ts --sequence.shuffle=false` with isolated Promptfoo DB/config/cache state
- `npm run db:migrate` against an isolated `/tmp` database
- `npm run tsc`, `npm run build`, `npm run bundle:sea`
- Linux ABI guard: proved it rejects the original over-floor native addon, then passed on the manylinux rebuilt payload (`12` ELF artifacts checked)
- `./scripts/smoke-test.sh` against the corrected SEA bundle (`31/31` tests passed)
- Built and installed the Python wheel with pinned `uv 0.11.16` / `wheel 0.45.1`; verified `promptfoo --version`, `python -m promptfoo --version`, and failure propagation
- `shfmt` clean; `actionlint` clean with its stale unknown-label diagnostic for GitHub's documented `macos-15-intel` runner ignored

The prior permissions thread is outdated: this workflow already defines workflow-level `permissions: contents: read` and narrow job-level permissions. I will reconcile it if GitHub permits resolution of that obsolete thread.

mldangelo-oai:
@codex review

mldangelo-oai:
Addressed the fresh Codex and Copilot review findings in `836fd9fd02941812e3b9a4c2b4dc063114f783e0`.

Implemented fixes:
- Binary installers now download from the actual unprefixed release-please tag (`.../download/<version>/...`) instead of nonexistent `v<version>` asset paths.
- `latest` resolution now uses the stable `promptfoo` npm package metadata, so unrelated `code-scan-action-*` GitHub component releases cannot be selected.
- Shell installation detects Linux systems below the documented glibc 2.28 floor and takes the npm fallback rather than installing a loader-incompatible executable.
- Both shell and PowerShell npm fallbacks report a clear error when `npm` is absent; removed the unused PowerShell `Write-Error` shadow of the built-in cmdlet.
- The SEA launcher disables code cache because its loader intentionally uses dynamic `import()` for the ESM CLI.
- The Linux ABI guard now fails immediately with a clear prerequisite error if required inspection tools are missing.
- The Python wheel wrapper handles permission-update failures with an actionable user message and clean nonzero return.

Validation after the follow-up:
- `npm run l`, `npm run f`, `git diff --check`, `shfmt`, `actionlint` (with the documented `macos-15-intel` label exception), shell syntax, and canonical/site installer mirror equality
- `npm run build`, then `npm run bundle:sea` with the adjusted loader; produced binary verified `0.121.12`
- `bash scripts/verify-linux-bundle-compat.sh` passed on the corrected bundle (`12` ELF artifacts checked)
- `./scripts/smoke-test.sh` passed `34/34`, adding regressions for root-package latest resolution, unprefixed release URLs, old-glibc fallback, and missing-npm diagnostics
- Built/retagged/installed the Python wheel in a fresh venv with pinned `uv 0.11.16`; verified both entrypoints and failure propagation, plus the wrapper permission-error path

Replacement CI has been triggered for this new head; I will keep monitoring it and any final review response.

mldangelo-oai:
@codex review

mldangelo-oai:
CI follow-up: `Check Python (3.14)` on `836fd9fd0` required Ruff formatting for the new permission-handling expression in `python/src/promptfoo/cli.py`. I applied the exact `ruff format` output in `c091bbf01ab7754de97e27d015572550c04710dc`, and locally verified `ruff format --check` plus `ruff check --select F401,F841,I` with the CI-pinned `ruff==0.15.1`. Replacement CI is now authoritative.

mldangelo-oai:
@codex review

mldangelo-oai:
Addressed the latest Copilot installer feedback in `696fcc62b` (`fix(dist): quote installer paths with spaces`).

- Shell installers now write a fish-safe quoted PATH entry, so custom install directories containing spaces remain one PATH element.
- PowerShell installers now execute the generated `.cmd` wrapper with the call operator, so verification works when its path contains spaces.
- Added a smoke regression for a fish install directory containing a space and a structural PowerShell regression guard; canonical and site-hosted installer copies remain identical.

Validation on the pushed source: `git diff --check`; mirror equality checks; CI-equivalent `shfmt -d -i 2`; `npm run bundle:sea` (SEA verified at `0.121.12`); and `./scripts/smoke-test.sh` (`35/35` passing, Ruby intentionally skipped because unavailable locally). Current `origin/main` (`2d3d4914cd`) remains included in this branch.

For context, the sole failed check on the superseded parent head was an unannotated Windows `actions/cache` step failure before install or tests ran; the retry advanced beyond that step before this new review-fix head replaced it.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
Addressed the latest Copilot bundle cleanup feedback in `b1cbc2b54` (`refactor(dist): simplify standalone bundle build`):

- removed the unused `__PROMPTFOO_INSTALL_METHOD__` esbuild define and the corresponding unconsumed Python wrapper environment plumbing
- removed the always-empty CommonJS footer scaffold
- report the actual number of created stub modules instead of the candidate count (the rebuilt SEA bundle now accurately reports `29`)

Validation on this tip before push: Biome/lint/TypeScript, pinned Ruff and Python byte compilation, `git diff --check`, rebuilt SEA executable verification (`0.121.12`), full standalone smoke suite (`35/35`), and disposable Python wheel installation/entrypoint/failure-propagation QA (both entrypoints returned `0.121.12`; missing config propagated a nonzero exit). A replacement CI/security pass is now required for this pushed head.

mldangelo-oai:
Addressed the latest Codex runtime/release findings in `0584993508` (`fix(dist): preserve standalone runtime workflows`):

- copy `src/tableOutput.html` into standalone/pip payloads and exercise HTML export end to end
- bundle through `src/entrypoint.ts` with the normal injected Node-engine comparators plus an explicit bundled-main flag, preserving structured code-scan stdout preflight without changing asset resolution
- require npm fallback runtimes to satisfy `^20.20.0 || >=22.22.0` in both shell and PowerShell installer mirrors
- gate `publish-npm` on successful `build-bundles`, so npm `latest` cannot point installers at missing release archives
- add `PROMPTFOO_NO_AUTO_INSTALL` for PowerShell function-loading/explicit options and document a single-install invocation flow

Validation on this pushed tip before publish: mirror/shell/whitespace checks; Biome/lint/TypeScript; entrypoint and structured-output tests (`30/30`); full code-scan suite (`155/155`); actionlint with only its previously documented stale `macos-15-intel` runner-label warning; SEA rebuild/verification (`0.121.12`, HTML template present, actual stub count `29`); standalone smoke (`38/38`); docs build with `SKIP_OG_GENERATION=true`; disposable manylinux-tagged wheel install with both entrypoints returning `0.121.12`; installed-wheel smoke (`38/38`); and Python wrapper missing-config failure propagation. PowerShell safeguards were structurally verified because PowerShell is not installed in this Linux QA environment. Replacement CI/review is now required for this head.

mldangelo-oai:
@codex review

mldangelo-oai:
Pushed `03312b830` (`fix(dist): harden standalone artifact installation`) to address the latest distribution review wave.

### Fixes
- Extract downloaded tarballs only after validating paths and rejecting link entries in an empty staging root; also fixed a `pipefail`/`grep -q` early-exit hole found during final self-review.
- Select Bash explicitly for the cross-platform release upload step, remove the `sharp` stub so dependency detection is truthful, and make ABI symbol probing tolerant of valid ELF files with no version matches.
- Harden the mirrored shell and PowerShell installer probes/download behavior and clarify the SEA launcher comment.
- Added regressions for traversal archives, link-first archives with trailing content, PowerShell download guards, and the absent `sharp` payload.

### Validation
- Shell formatting/syntax/mirror checks, `git diff --check`, `npm run l`, and `npm run tsc`.
- Focused entrypoint/redteam/code-scan test set: `184/184` passing.
- `npm run bundle:sea`: verified standalone `0.121.12`, bundled HTML template present, and `28` truthful optional stubs.
- Standalone smoke: `40/40` passing, including both malicious-archive regressions and structured code-scan stdout behavior.
- `SKIP_OG_GENERATION=true npm --prefix site run build`: passing.
- Fresh retagged manylinux wheel install: both `promptfoo --version` and `python -m promptfoo --version` report `0.121.12`; installed-wheel smoke `40/40` passing.
- `actionlint` reports only the previously documented stale `macos-15-intel` label warning; the GitHub-supported runner label remains valid.

The repeated `tableOutput.html` comment is already satisfied by `0584993508` and remains covered by the HTML runtime smoke test. Ruby is unavailable in the local environment and is the expected smoke-test skip.

mldangelo-oai:
Pushed follow-up `ad1514915` (`fix(dist): handle minimal installer environments`) for the additional live review findings after `03312b830`.

- The mirrored shell installers now defer default-directory resolution until after argument parsing: an unset `HOME` produces an actionable error for default installs, while `--dir` remains usable and safely skips shell rc-file mutation.
- The Linux compatibility verifier checks ELF magic in Bash before invoking `readelf`, avoiding external symbol inspection for ordinary bundled text/assets while retaining the GLIBC/GLIBCXX enforcement.
- Added an unset-`HOME` regression to the distribution smoke suite. The earlier safe-extraction concerns remain covered by the traversal and link-first archive regressions already landed in `03312b830`.

Validation on the final head: shell format/syntax/mirror/whitespace checks passed; targeted ABI proof ignored a text asset and accepted one no-version ELF object; SEA rebuilt and verified as `0.121.12` with `28` stubs; docs production build passed; standalone smoke passed `41/41`; fresh retagged manylinux wheel install exposed both `0.121.12` entrypoints and passed installed-wheel smoke `41/41` (Ruby is unavailable locally and skipped as designed).

mldangelo-oai:
@codex review

mldangelo-oai:
Final closeout for head `ad1514915`: GitHub CI is green across all `37/37` checks, including CodeQL, Code Scan Action, docs, smoke, web UI, redteam production/staging, and the cross-platform Node lanes. The live open code-scanning alerts query returns `[]`, and every actionable review thread is resolved; the only remaining annotation is the pre-existing obsolete GHAS non-conversation that GitHub has refused to resolve.

As a final local review sidecar after CI, I rebuilt the SEA artifact and ran `bash ./scripts/verify-linux-bundle-compat.sh bundle` against the full real payload: it checked `10` Linux ELF artifacts and verified the documented `GLIBC <= 2.28` / `GLIBCXX <= 3.4.25` floors.

mldangelo-oai:
Superseding the preceding closeout with `9b0ed73c7` (`fix(dist): validate installer artifact paths`) after Copilot feedback landed concurrently on `ad1514915`.

- Shell installer mirrors now use fixed-string matching when checking whether a custom binary directory is already present in shell configuration, so paths containing spaces or regex metacharacters cannot be misclassified.
- PowerShell installer mirrors now require `promptfoo.exe` to exist after extraction before creating `pf.exe` or returning success.
- The smoke regression now seeds a misleading fish entry and installs into a directory containing both spaces and `[]`, proving the actual literal path is appended; it also guards the PowerShell archive-content requirement in both published script copies.

Validation: `shfmt`/shell syntax/mirror checks and `git diff --check` passed; `SKIP_OG_GENERATION=true npm --prefix site run build` passed; rebuilt standalone SEA verified as `0.121.12`; standalone smoke passed `41/41` (Ruby unavailable locally and skipped as designed). PowerShell is not installed locally, so Windows CI is the runtime verification gate for the PowerShell-only path. The wheel payload itself is unchanged from the previously installed-wheel-tested head.

mldangelo-oai:
@codex review

mldangelo-oai:
Published follow-up `9636af613` (`fix(dist): replace standalone payloads cleanly`).

- Fixed the valid late review finding: both installer mirrors now prepare a complete staged `bin` replacement, hold the previous install for rollback during the rename, and remove it only after a successful replacement. This prevents obsolete payload assets or modules from persisting across standalone upgrades.
- Added a real shell v1-to-v2 reinstall regression that proves an obsolete asset disappears; added mirrored PowerShell swap guards.
- Validated the two `@huggingface/transformers` review threads as inapplicable on this tip: it is externalized but not created by `createStubModules()`, and rebuilt bundle smoke verifies that payload path is absent.

Validation: shell formatting/syntax/mirror/diff checks; rebuilt SEA reports `0.121.12` with `28` stubs; docs build; standalone smoke `42/42`. The packaged payload code is unchanged from the already installed-wheel-tested tip for this installer-only follow-up. `pwsh` is unavailable locally, so Windows runtime behavior is covered by replacement CI.

mldangelo-oai:
@codex review

mldangelo-oai:
Published follow-up `54da698d3` (`fix(dist): apply standalone bundle mode`) after reconciling the late Copilot review.

- Fixed the valid SEA-mode finding: `createBundle({ forSea: buildSea })` now activates the existing minification path only for standalone builds, with its option shape matching the existing defaults.
- Investigated the two `./../` traversal comments. The existing shell pattern `*/../*` already rejects `./../escaped` (and nested variants), so changing extraction logic would not improve protection. I strengthened the real archive smoke fixture to use the reported `./../escaped` member and it is rejected before extraction.

Validation on this head: `npm run l`, `npm run f`, `npm run tsc`; rebuilt minified SEA binary verified as `0.121.12` with the bundled HTML template and `28` stubs; full standalone smoke `42/42`, including clean reinstall replacement and the dot-prefixed traversal archive rejection. Fresh CI and bot review are requested below.

mldangelo-oai:
@codex review

mldangelo-oai:
Published follow-up `e5041cda2` (`fix(dist): address distribution portability review`).

- Fixed the three latest Copilot findings: the standalone smoke harness now uses a BSD/macOS-compatible `mktemp` template, and the two new distribution jobs now use the same pinned `actions/setup-node@48b55a0...` digest as the existing workflow jobs.
- Validation of the portable temp root usefully exposed an additional installer correctness bug: an unrelated fish rc-file path containing the word `promptfoo` caused PATH setup to be skipped. Both shell installer mirrors now suppress insertion only when the selected `$BIN_DIR` is present, and the regression covers the misleading path alongside spaces and bracket characters.

Validation: `npm run f`, `npm run l`, shell syntax/mirror/diff checks; rebuilt minified SEA verified as `0.121.12` with bundled assets and `28` stubs; full standalone smoke first reproduced the PATH false positive, then passed `42/42` after the repair. The repository installs `actionlint` inside CI rather than locally, so the replacement GitHub Actions Lint job is the workflow-runtime gate.

mldangelo-oai:
@codex review

mldangelo-oai:
Final reconciliation for published head `e5041cda2`:

- CI is fully green: `37/37` successful checks with zero pending/failing jobs, including CodeQL, workflow lint, shell formatting, standalone smoke, web UI, docs, redteam, integration, and the full platform/Node matrix.
- The final-head Copilot review submitted at `2026-05-27T17:57:33Z` created no new actionable threads. Codex was explicitly requested on this head at `#issuecomment-4557118421`; no additional final-head feedback appeared during the CI/reconciliation window.
- The code-scanning alert query for this PR returns `[]`. All actionable review discussions introduced by this PR have been resolved. The one remaining historical GitHub Advanced Security discussion predates this work and is not exposed as a resolvable current alert.

This tip is ready for maintainer review/merge.

scripts/install.sh
```diff
@@ -0,0 +1,640 @@
...
+  backup_parent=""
+  if [[ -e "$BIN_DIR" ]]; then
+    backup_parent=$(mktemp -d "$INSTALL_DIR/.bin-backup.XXXXXXXX")
+    if ! mv "$BIN_DIR" "$backup_parent/bin"; then
+      rm -f "$temp_file"
+      rm -rf "$extract_dir" "$install_staging" "$backup_parent"
+      error "Existing installation could not be prepared for replacement."
+    fi
+  fi
+  if ! mv "$install_staging" "$BIN_DIR"; then
```
jameshiester-oai, line 431: **[P1] Preserve unrelated tools in custom prefixes**

`--dir` is documented as an installation prefix and maps to `<prefix>/bin`, so values such as `$HOME/.local` or `/usr/local` are reasonable. This swap moves that entire existing `bin` aside, installs a directory containing only Promptfoo, and then deletes the backup. In a temp-prefix reproduction, the installer exited 0 and installed Promptfoo but permanently removed a pre-existing `unrelated-tool`. PowerShell performs the same whole-directory replacement at `scripts/install.ps1:212-225`. Replace only a Promptfoo-owned payload and its `promptfoo`/`pf` shims, or reject a non-owned shared bin; add a regression test that preserves unrelated entries.

scripts/install.sh
```diff
@@ -0,0 +1,640 @@
...
+  # Try binary installation first, fall back to npm
+  if ! install_binary "$VERSION" "$platform"; then
+    error "Installation failed."
+  fi
+
+  # Verify installation
+  if verify_installation; then
+    success "promptfoo v${INSTALLED_VERSION:-unknown} installed successfully"
+  else
+    warn "Installation completed but verification failed."
```
jameshiester-oai, line 629: **[P1] Roll back and fail when runtime verification fails**

The old bin backup is deleted at lines 439-440 before this check, and this branch only warns before PATH setup, telemetry, and a zero exit. A temp upgrade from a working binary to an executable that exits 42 produced `installer_exit=0`, left the broken binary installed, and lost the working version. PowerShell has the same ordering at `scripts/install.ps1:224-225,434-448`. Verify the staged payload and require its normalized version to equal the requested version before swapping, or retain the backup through verification and restore it on any failure/mismatch; return nonzero in both installers.

scripts/install.sh
```diff
@@ -0,0 +1,640 @@
...
+  os="$(uname -s)"
+  arch="$(uname -m)"
+
+  case "$os" in
+  Darwin)
+    macos_version=$(sw_vers -productVersion 2>/dev/null || true)
+    if [[ "$macos_version" =~ ^([0-9]+)\.([0-9]+) ]]; then
+      major="${BASH_REMATCH[1]}"
+      minor="${BASH_REMATCH[2]}"
+      if ((major < 13 || (major == 13 && minor < 5))); then
```
jameshiester-oai, line 173: **[P1] Do not accept macOS 13 for a macOS 14 native binding**

The bundle recursively ships `@libsql/darwin-arm64@0.5.29` / `darwin-x64@0.5.29`; inspection of the exact locked Mach-O bindings reports `LC_BUILD_VERSION minos 14.0`, while the embedded Node executable reports 13.5. Since `main()` runs database migrations before Commander handles `--version`, a 13.5/13.6 install cannot even pass installer verification. Raise standalone detection and docs to macOS 14, or ship bindings built for 13.5, and scan every bundled Mach-O native file for the advertised floor.

scripts/smoke-test.sh
```diff
@@ -0,0 +1,1577 @@
...
+  payload="$fake_root/payload"
+  archive="$fake_root/promptfoo-0.0.0-linux-x64.tar.gz"
+  mkdir -p "$fake_bin" "$payload"
+
+  cat >"$payload/promptfoo" <<'EOF'
+#!/bin/sh
+echo compromised
+EOF
+  chmod +x "$payload/promptfoo"
+  tar -czf "$archive" --transform='s#^promptfoo$#./../escaped#' -C "$payload" promptfoo
```
jameshiester-oai, line 1280: **[P1] Build the traversal fixture portably on macOS**

Both new macOS release runners execute this suite, but BSD tar does not implement GNU `--transform`. Running the exact-head SEA smoke suite on macOS produced 42 passes and this single failure, then attempted an unintended npm fallback because the archive was never created. That blocks `build-bundles`, npm, and PyPI publication on every release. Build the malicious member name with a portable mechanism such as Python `tarfile`, and fail the test immediately if fixture creation fails.

.github/workflows/release-please.yml
```diff
@@ -177,6 +180,302 @@ jobs:
...
+      - name: Upload install scripts to release
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          REPOSITORY: ${{ github.repository }}
+          TAG_NAME: ${{ needs.release-please.outputs.tag_name }}
+        run: |
+          set -euo pipefail
+          gh release upload "$TAG_NAME" scripts/install.sh scripts/install.ps1 --clobber --repo "$REPOSITORY"
+
+  build-bundles:
```
jameshiester-oai, line 217: **[P1] Validate the artifact matrix before a real release**

This workflow runs only on pushes to `main` or manual dispatch, and this job additionally requires `release_created == true`. No pull-request workflow invokes the new `bundle:sea`, artifact smoke suite, or wheel build, so the first cross-platform validation occurs after merge during a real release. The reproducible BSD-tar failure in this head is exactly the class of blocker the current PR checks cannot see. Extract a tokenless reusable build/smoke/wheel matrix and invoke it for pull requests; keep only upload/publish steps release-gated.

python/src/promptfoo/cli.py
```diff
@@ -0,0 +1,90 @@
...
+    except (FileNotFoundError, RuntimeError) as e:
+        print(f"Error: {e}", file=sys.stderr)
+        return 1
+
+    # Build the command: promptfoo [args...]
+    cmd = [str(binary_path)] + sys.argv[1:]
+
+    # Execute the binary
+    try:
+        result = subprocess.run(cmd)
```
jameshiester-oai, line 80: **[P2] Replace the wrapper process or forward termination**

`subprocess.run` leaves a separate Python process between the process manager and the real CLI. In a direct reproduction, SIGTERM sent to the wrapper exited Python with 143 while the child remained alive; when the child terminated by SIGTERM, `subprocess` returned `-15` and `sys.exit(-15)` surfaced as 241. Timeouts and PID-targeted shutdown can therefore leak eval/provider work and report the wrong status. On POSIX, `os.execv` is the simplest fix; otherwise forward signals and normalize negative return codes, with lifecycle tests.

python/src/promptfoo/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""
+promptfoo - Test your prompts, agents, and RAGs.
+
+This package provides the promptfoo CLI for evaluating and testing LLM applications.
+"""
+
+# Version is read from package.json during build and injected here
+__version__ = "0.0.0.dev0"
```
jameshiester-oai, line 8: **[P2] Preserve the existing public PyPI import**

The current official `promptfoo==0.1.4` wheel explicitly defines `__all__ = ["__version__", "main"]` and `from .cli import main`. Publishing this 0.121.x package over the same PyPI project makes an ordinary upgrade break `from promptfoo import main`. Re-export `main` here and test that import from the built wheel, or explicitly document and migrate an intentional compatibility break. Evidence: [current PyPI release](https://pypi.org/project/promptfoo/0.1.4/).

scripts/install.ps1
```diff
@@ -0,0 +1,482 @@
...
+            [System.Runtime.InteropServices.RuntimeInformation]::OSArchitecture -eq 'Arm64') {
+            'arm64'
+        } else {
+            'x64'
+        }
+    } else {
+        throw "32-bit systems are not supported"
+    }
+
+    return "win32-$arch"
```
jameshiester-oai, line 63: **[P2] Enforce the embedded runtime's Windows floor**

This accepts every 64-bit Windows host, while the docs advertise generic Windows x64 support and the embedded Node 24 binary officially requires Windows 10 / Server 2016 or newer. PowerShell 5.1 and the `win_amd64` wheel can still reach older hosts, where the downloaded executable is deterministically unsupported and the current verifier then reports completion. Detect/refuse older Windows versions (or route to a known-compatible fallback) and document the floor. See [Node 24 supported platforms](https://github.com/nodejs/node/blob/v24.x/BUILDING.md#supported-platforms).

.github/workflows/release-please.yml
```diff
@@ -177,6 +180,302 @@ jobs:
...
+          PLATFORM: ${{ matrix.platform }}
+          VERSION: ${{ needs.release-please.outputs.tag_name }}
+        run: |
+          $version = $env:VERSION -replace '^v', ''
+          $archiveName = "promptfoo-${version}-${env:PLATFORM}.zip"
+          Compress-Archive -Path bundle\* -DestinationPath $archiveName
+          "ARCHIVE_NAME=$archiveName" >> $env:GITHUB_ENV
+
+      - name: Create app token for release upload
+        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3
```
jameshiester-oai, line 304: **[P2] Scope release-upload tokens to the one required permission**

Without `permission-*` inputs, `actions/create-github-app-token` inherits every permission granted to the installation, while this job only needs release asset upload. The bundle job has also already run dependency/build code before minting the broad token. Request `permission-contents: write` for both new upload tokens, and preferably move credentialed upload into a clean job that consumes Actions artifacts. This is narrower than the release-please token case: `contents: write` is necessarily available because the upload itself requires it. See the [pinned action's permission guidance](https://github.com/actions/create-github-app-token/blob/1b10c78c7865c340bc4f6099eb2f838309f1e8c3/README.md).

jameshiester-oai (commented):
Reviewed base: `main` @ `c54a30668ad8319d76c20ae96e6680ad6c51a2c6`
Reviewed head: `81474cc26e79b49b9b1854b9845a8f072e3d9e6b`
Scope: full `base...head` review; this head had no prior authenticated completion marker. At the final fetch, `origin/main` was `689a304e195d7f00a27a53ed92bed47e4fd30c88` and GitHub reported the PR as merge-conflicted.

## Verdict

Not ready to ship. The core SEA and wheel work locally, but five P1 defects can delete user files, break documented platforms, or block the release pipeline. Five narrower P2 defects affect process lifecycle, compatibility, release ordering, and token scope. Inline comments contain the triggering conditions, reproductions, impact, and correction direction.

## Verified findings

### P1

- Custom `--dir` prefixes replace the entire shared `<prefix>/bin` directory and permanently delete unrelated executables (`scripts/install.sh:423-440`; PowerShell has the same behavior at `scripts/install.ps1:212-225`).
- A failed post-install runtime check only warns after the working backup is deleted, then exits successfully and leaves the broken payload installed (`scripts/install.sh:439-440,625-637`; `scripts/install.ps1:224-225,434-448`).
- The standalone installer accepts and documents macOS 13.5, but both locked libsql Darwin bindings require macOS 14.0; database initialization occurs before `--version` handling.
- The release smoke suite uses GNU-only `tar --transform`, so both macOS bundle matrix jobs fail. The exact-head macOS run completed 42/43 tests and failed only on this fixture.
- The new cross-platform SEA/wheel jobs are reachable only after a release is created on `main`; no pull-request check invokes `bundle:sea`, this smoke suite, or wheel packaging. The macOS failure therefore remained invisible to PR CI.

### P2

- The pip wrapper spawns the SEA without forwarding termination or normalizing signal statuses. SIGTERM to the wrapper left the child alive; a SIGTERM-terminated child surfaced as exit 241 rather than 143.
- Publishing over the existing official PyPI project drops its explicit `from promptfoo import main` API.
- PowerShell accepts every Windows x64 host although the embedded Node 24 runtime requires Windows 10 / Server 2016 or newer.
- The two new release-upload App tokens inherit all installation permissions instead of requesting only `contents: write`.
- Off-diff interaction: manual npm backfill (`.github/workflows/release-please.yml:119-163`) can advance npm `latest` without verifying or building matching standalone archives, defeating the normal bundle gate for no-Node installs.

## Existing unresolved feedback

I revalidated and did not duplicate five existing P2 threads: the `pdf-parse` stub suppresses the intended install guidance; bundle and wheel jobs omit the required explicit npm registry pins; PowerShell npm fallback can prefer a stale `.exe`; and shell npm fallback cannot replace an existing standalone shim. The old outdated GHAS permissions thread is addressed by the current workflow-level and job-level `permissions` blocks.

- `build-python` rebuilds five SEA payloads after `build-bundles` already produced and verified the same platform artifacts. Passing those verified artifacts into wheel packaging would reduce release time and remove a source of artifact variance.
- Add regression coverage for shared-prefix preservation, verification rollback/version equality, signal forwarding, the established PyPI import, and native-library OS floors.

## Tests and end-to-end QA

- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci` — passed with Node 24.16.0 / npm 11.13.0. The first plain `npm ci` was interrupted after Playwright browser download stalled; package installation then completed without downloading browsers that are not shipped.
- `npm run bundle:sea` — passed; built a 24.33 MB ESM bundle and 114.11 MB signed arm64 SEA, and the generated binary returned `0.121.15`.
- `PROMPTFOO_DISABLE_TELEMETRY=true ./scripts/smoke-test.sh` — failed 1/43 solely because BSD tar rejects `--transform`; the other 42 tests passed, including no-cache evals, persistence, TypeScript config loading, Python/Ruby providers, JSON/HTML output, and installer paths.
- `npx vitest run test/migrate.test.ts --sequence.shuffle=false` — 11/11 passed.
- `npm run tsc` — passed.
- `bash -n scripts/install.sh scripts/smoke-test.sh` plus byte comparisons of canonical/site installer copies — passed.
- Independent exact-head wheel QA built and retagged a 54 MB `macosx_14_0_arm64` wheel, installed it into an isolated venv, verified both console and `python -m promptfoo` entrypoints, propagated a missing-config failure, ran an actual `eval --no-cache` with `success=true`, `score=1`, no error, and read the persisted eval back.
- Temp-only installer probes reproduced both unrelated-bin deletion and successful completion after a broken replacement. `otool -l` reported `minos 14.0` for the shipped libsql binding and `minos 13.5` for the Node executable.

## CI and limitations

The three red checks are unrelated Codecov CLI GPG-key failures after their suites passed: Site 101 tests, Web UI 4,816 tests, and Node 20 Ubuntu 19,662 tests. Other listed checks are green. The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable (`command not found`), so I manually applied its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses. PowerShell was not available locally, Linux ABI validation was not run against a locally built Linux artifact, and no release upload or registry publication was attempted.

PR #6683, opened 2025-12-14
feat(webui): redesign Table Settings modal with compact two-column layout

### LINKED ISSUE #1988: Show score and reason even when passes llm-rubric
Noticed when using llm-rubric it only displays the reason when assertions fail, however I would like to also see the reason the model gives even when it passes to get a understand why.

Update rendering UI to show reason and score

Let me know if this is a good idea and I could try to come up with a PR too.

mldangelo:
## Summary

Redesigns the Table Settings modal with a modern, compact two-column layout that displays all settings at a glance without scrolling. Also adds the ability to show pass reasons for assertions like llm-rubric.

**Before:** Scrollable list with large section headers and verbose descriptions
**After:** Clean two-column grid with compact toggles and sliders

Table Settings Modal

Fixes #1988

## Changes

**UI Redesign:**
- Replaced verbose `SettingItem`/`SettingsSection` components with new `CompactToggle` and `CompactSlider` components
- Changed from scrollable list to two-column CSS Grid layout ("Show" | "Display")
- Reduced modal height by ~60% while maintaining all functionality
- Added subtle divider between columns
- Improved visual hierarchy with minimal section headers
- Fixed slider value label contrast for dark mode

**New Components:**
- `CompactToggle`: Space-efficient checkbox with inline label and optional tooltip
- `CompactSlider`: Compact range slider with label, value display, and optional "Unlimited" mode

**New Feature (Fixes #1988):**
- Added "Pass reasons" toggle to show assertion pass reasons (e.g., from llm-rubric) in eval output cells
- Previously, llm-rubric only showed the reason when assertions failed—now users can see why assertions passed too
- Toggle is intelligently disabled when pass/fail indicators are turned off
- Pass reasons display with green accent styling in output cells

**Technical:**
- All settings remain persisted via Zustand store with localStorage
- Modal width optimized to 680px for content fit
- Full test coverage for new components and interactions

## Test Plan

- [x] All 56 TableSettings tests pass
- [x] Verified two-column layout renders correctly on desktop
- [x] Verified mobile full-screen mode still works
- [x] Verified all toggles and sliders function correctly
- [x] Verified pass reasons display when enabled for llm-rubric assertions
- [x] Verified settings persist across page reloads
- [x] Verified slider value label has proper contrast in dark mode
- [x] Linting and formatting pass


typpo:
send pics

PR #6696, opened 2025-12-15
fix(webui): include grading rubric in storedGraderResult for iterative, hydra, and custom strategies

yash2998chhabria:
## Summary

The iterative, hydra, and custom strategy providers were not including the grading rubric (`assertion.value`) in their `storedGraderResult`. This caused the "Value" column in the eval UI to show "-" instead of the grading template when using these strategies.

**Root cause:** These strategies were destructuring only `grade` from `grader.getResult()` and storing it directly without reconstructing the assertion with the rubric value.

**The fix:** Each affected strategy now:
1. Destructures `rubric` from `grader.getResult()` 
2. Reconstructs `storedGraderResult` with `assertion.value` set to `rubric`

This matches the pattern already used by crescendo, goat, iterativeMeta, and iterativeTree strategies.

## Files Changed

- `src/redteam/providers/iterative.ts`
- `src/redteam/providers/hydra/index.ts`
- `src/redteam/providers/custom/index.ts`

## Test plan

- [ ] Run redteam eval with iterative strategy and verify Value column shows grading template
- [ ] Run redteam eval with hydra strategy and verify Value column shows grading template
- [ ] Run redteam eval with custom strategy and verify Value column shows grading template

### DIFF
--- a/src/redteam/providers/shared.ts
+++ b/src/redteam/providers/shared.ts
@@ -8,6 +8,8 @@ import { PromptfooChatCompletionProvider } from '../../providers/promptfoo';
 import type { TraceContextData } from '../../tracing/traceContext';
 import {
   type ApiProvider,
+  type Assertion,
+  type AssertionOrSet,
   type CallApiContextParams,
   type CallApiOptionsParams,
   type EvaluateResult,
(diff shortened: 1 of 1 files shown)


src/redteam/providers/custom/index.ts
```diff
@@ -524,7 +524,14 @@ export class CustomProvider implements ApiProvider {
...
             graderPassed = grade.pass;
-            storedGraderResult = grade;
+            storedGraderResult = {
+              ...grade,
+              assertion: grade.assertion
+                ? { ...grade.assertion, value: rubric }
+                : assertToUse && 'type' in assertToUse && assertToUse.type !== 'assert-set'
+                  ? { ...assertToUse, value: rubric }
+                  : undefined,
+            };
```
faizanminhas: can we extract this nested ternary pleaase

src/redteam/providers/shared.ts
```diff
@@ -528,3 +528,25 @@ export async function tryUnblocking({
...
+ *
+ * Note: Uses `any` for flexibility since the Assertion type is complex and the
+ * function needs to handle both gradeAssertion (from grader result) and
+ * assertToUse (from test config) which may have different shapes.
+ */
+export function buildGraderResultAssertion(
+  gradeAssertion: any,
+  assertToUse: any,
+  rubric: string | undefined,
+): any {
```
faizanminhas: Why `any`? Are there types for these already somewhere that you can use?

faizanminhas: https://github.com/promptfoo/promptfoo/blob/015f2dfb76be0710a2c98d87fe957060e18de162/src/types/index.ts#L545-L576

yash2998chhabria: Nice, makes sense

PR #6704, opened 2025-12-16
fix(eval): store original vars before defaultTest merge for accurate filter matching

minhle1291:
## Summary

Fixes an issue where `--filter-errors-only` and `--filter-failing` flags returned 0 matching tests when using configurations with `defaultTest.vars`.

### Root Cause
During evaluation, `testCase.vars` is mutated by merging with `defaultTest.vars`. The stored results contain merged vars, but when filtering, freshly loaded tests have only their original vars. The deep equality comparison fails, resulting in 0 matches.

### Changes
- Store original vars in `testCase.metadata._originalVars` before merging with `defaultTest.vars`
- Update `resultIsForTestCase()` to use `_originalVars` for matching when available (falls back to `result.vars` for backward compatibility)
- Add debug logging to `filterTestsByResults` for troubleshooting
- Add test for the fix

### Backward Compatibility
- Old results without `_originalVars` will fall back to existing behavior
- New results store `_originalVars` and use it for accurate matching

## Test plan
- [x] All existing filter tests pass
- [x] All evaluator tests pass
- [x] New test for `_originalVars` matching passes

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -944,10 +944,22 @@ class Evaluator {
         ? testSuite?.defaultTest?.options?.transformVars
         : undefined;
     for (const testCase of tests) {
+      // Store original vars before merging with defaultTest.vars for accurate test matching
+      // This is used by --filter-failing and --filter-errors-only to match results back to test cases
+      const originalVars = testCase.vars ? { ...testCase.vars } : undefined;
+
       testCase.vars = {
         ...(typeof testSuite.defaultTest === 'object' ? testSuite.defaultTest?.vars : {}),
         ...testCase?.vars,
       };
+
+      // Store original vars in metadata for use in result matching
+      if (originalVars !== undefined) {
+        testCase.metadata = {
+          ...testCase.metadata,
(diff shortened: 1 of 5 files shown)


src/evaluator.ts
```diff
@@ -944,11 +944,23 @@ class Evaluator {
...
       testCase.vars = {
         ...(typeof testSuite.defaultTest === 'object' ? testSuite.defaultTest?.vars : {}),
         ...testCase?.vars,
       };
 
+      // Store original vars in metadata for use in result matching
+      if (originalVars !== undefined) {
+        testCase.metadata = {
+          ...testCase.metadata,
+          _originalVars: originalVars,
```
mldangelo, line 960: I am scared of copying all of the vars (sometimes they are very large) and threw up https://github.com/promptfoo/promptfoo/pull/6706/changes as an alternative. We should pick yours or mine as soon as possible.

minhle1291:
Closing in favor of cleaner approach in https://github.com/promptfoo/promptfoo/pull/6706, which also passed test

PR #6708, opened 2025-12-16
feat: blob storage

addelong:
## Summary

  - Enable filesystem-backed media/blob externalization by default (legacy inline only when PROMPTFOO_INLINE_MEDIA=true); harden path handling, and add regression coverage for traversal prevention.
  - Align redteam media handling with storage refs (optional data, history image/audio support, Voice Crescendo/text-to-audio return objects) and update UI/media helpers to work with storage/blob refs.

##  Testing

  - npm test
  - npm run build
  - npm test -- ResultsTable.test.tsx
  - npm test -- TestCaseGenerationProvider.test.tsx
  - npx vitest run test/redteam/strategies/simpleAudio.test.ts

##  Manual verification

  1. Default (externalized): run npm run dev:app with no media flags, generate evals with image/audio outputs, confirm they load via refs (and lightbox) and that files land under ~/.promptfoo/blobs.
  2. Legacy inline: set PROMPTFOO_INLINE_MEDIA=true, rerun flows, confirm media stays inline and no filesystem writes occur.
  3. Redteam audio/image flows: run npm run dev, execute audio/image strategy, ensure history shows audio/image entries and playback works without crashes.


mldangelo (approved):
Awesome. Let's cut a release before landing!

PR #6709, opened 2025-12-16
feat(redteam): Redteam with multiple inputs

faizanminhas:
### Summary

- Added a new `target.inputs` configuration 
- Allows a user to specify different inputs that will be generated by plugins along with the main prompt
- These inputs will be defined as `testcase.vars` values, so they can be used & interpolated anywhere in the promptfoo config
- Compatible with the following strategies: Jailbreak:hydra, jailbreak:meta, jailbreak, crescendo, goat

#### Approach

The way this is implemented is by making `target.inputs` constitute a plugin modifier, which already flows to all of the plugins and strategies. Specifically, we instruct the LLM to generate a JSON object with fields = the inputs. Keeping this JSON as the `injectVar` allows for the actual value to flow through the existing system, treating the json as a single value, so all the fields can be changed at once, in context. 

The values are then extracted when constructing the `TestCase`, and the vars are individually set. Hence, when the test case is being executed, these values are available as nunjucks vars, or are passed as `context` to custom providers.

#### Test plan
I generated some test cases with every plugin, and ensured that the inputs are generated and present for each test case. In terms of strategies, I tested a sample of relevant strategies and saw that they iterated correctly on the values.

Also tested against custom policy, custom intent.


src/redteam/plugins/harmful/aligned.ts
```diff
@@ -38,6 +39,40 @@ export class AlignedHarmfulPlugin extends RedteamPluginBase {
   }
 
   protected promptsToTestCases(prompts: { prompt: string }[]): TestCase[] {
-    return prompts.map(({ prompt }) => createTestCase(this.injectVar, prompt, this.harmCategory));
+    const hasInputs = this.config.inputs && Object.keys(this.config.inputs).length > 0;
+    const harmCategoryLabel =
+      (require('../../constants').HARM_PLUGINS as typeof HARM_PLUGINS_TYPE)[this.harmCategory] ||
```
will-holley: Is the dynamic import a LLM artifact?

src/redteam/plugins/harmful/aligned.ts
```diff
@@ -38,6 +39,40 @@ export class AlignedHarmfulPlugin extends RedteamPluginBase {
...
+    return prompts.map(({ prompt }) => {
+      // Base vars with the primary injectVar
+      const vars: Record<string, string> = {
+        [this.injectVar]: prompt,
+      };
+
+      // If inputs is defined, extract individual keys from the JSON into vars
+      if (hasInputs) {
+        try {
+          const parsed = JSON.parse(prompt);
```
will-holley: What is `prompt` here? Stringified JSON _what_?

src/redteam/plugins/harmful/aligned.ts
```diff
@@ -38,6 +39,40 @@ export class AlignedHarmfulPlugin extends RedteamPluginBase {
   }
 
   protected promptsToTestCases(prompts: { prompt: string }[]): TestCase[] {
```
will-holley: Can you add unit test coverage for this?

src/redteam/plugins/harmful/unaligned.ts
```diff
@@ -7,6 +8,42 @@ import { createTestCase } from './common';
 import type { PluginActionParams, TestCase } from '../../../types/index';
 import type { UNALIGNED_PROVIDER_HARM_PLUGINS } from '../../constants';
 
+/**
+ * Extract content from <Prompt> tags and parse JSON if inputs are defined.
+ * Returns the processed prompt and any additional vars extracted from JSON.
+ */
+function processPromptForInputs(
```
will-holley, line 16: +1 on unit test coverage

src/redteam/plugins/base.ts
```diff
@@ -134,6 +134,44 @@ export function parseGeneratedPrompts(generatedPrompts: string): { prompt: strin
...
+ */
+export function parseGeneratedInputs(
+  generatedOutput: string,
+  inputs: Record<string, string>,
+): { prompt: string }[] {
+  const results: { prompt: string }[] = [];
+  const inputKeys = Object.keys(inputs);
+
+  // Extract JSON from <Prompt> tags
+  const promptRegex = /<Prompt>([\s\S]*?)<\/Prompt>/gi;
```
will-holley: This extraction is also contained within `processPromptForInputs`; would be nice as a DRY function w/ test cov.

src/redteam/plugins/base.ts
```diff
@@ -316,6 +394,19 @@ export abstract class RedteamPluginBase {
       modifiers.language = config.language;
     }
 
+    // Add inputs as a modifier if defined - instructs LLM to output JSON wrapped in <Prompt> tags
+    if (config.inputs && Object.keys(config.inputs).length > 0) {
+      const schema = Object.entries(config.inputs)
+        .map(([key, description]) => `"${key}": "<${description}>"`)
```
will-holley: Why are arrow brackets wrapping `description`?

src/redteam/plugins/base.ts
```diff
@@ -324,20 +415,41 @@ export abstract class RedteamPluginBase {
       return template;
     }
 
+    let processedTemplate = template.trim();
+
+    // If outputFormat modifier is set, remove conflicting format instructions from the template
+    // Templates often have lines like "Each line must begin with 'Prompt:'" that conflict with JSON output
+    if (modifiers.outputFormat) {
```
will-holley: Is it worth making the declaration of `modifiers.outputFormat` and `redteam.inputs` on a config mutually exclusive?

will-holley:
Can you share the test YAML config that you've been developing with?

src/redteam/plugins/base.ts
```diff
@@ -324,20 +415,41 @@ export abstract class RedteamPluginBase {
...
       .filter(([_, value]) => typeof value !== 'undefined' && value !== '')
       .map(([key, value]) => `${key}: ${value}`)
       .join('\n');
 
     return dedent`
-      ${template.trim()}
+      ${processedTemplate}
 
-      CRITICAL: Ensure all generated prompts strictly follow these requirements:
+      CRITICAL: Ignore any previous output format instructions. You MUST follow these requirements exactly:
```
will-holley, line 434: Will this change break other, non-multi input modifiers?

src/redteam/plugins/index.ts
```diff
@@ -71,6 +71,26 @@ type PluginClass<T extends PluginConfig> = new (
   config: T,
 ) => RedteamPluginBase;
 
+/**
+ * Computes modifiers from config (same logic as appendModifiers in base.ts).
+ * Used to ensure modifiers are available for strategies when using remote generation.
+ */
+function computeModifiersFromConfig(config: PluginConfig | undefined): Record<string, string> {
```
will-holley, line 77: Unit test please

src/redteam/plugins/index.ts
```diff
@@ -71,6 +71,26 @@ type PluginClass<T extends PluginConfig> = new (
...
+  const modifiers: Record<string, string> = {
+    ...(config?.modifiers as Record<string, string> | undefined),
+  };
+  if (config?.language && typeof config.language === 'string') {
+    modifiers.language = config.language;
+  }
+  if (config?.inputs && Object.keys(config.inputs).length > 0) {
+    const schema = Object.entries(config.inputs as Record<string, string>)
+      .map(([k, description]) => `"${k}": "<${description}>"`)
```
will-holley: Same comment re: arrow brackets. DRY this into a function w/ an explainer docstring + unit test cov?

src/redteam/plugins/index.ts
will-holley, line 77: Do we have a Zod schema for modifiers that we can use here?

src/redteam/plugins/pii.ts
```diff
@@ -8,6 +8,41 @@ import { RedteamGraderBase, RedteamPluginBase } from './base';
 
 const PLUGIN_ID = 'promptfoo:redteam:pii';
 
+/**
+ * Extract content from <Prompt> tags and parse JSON if inputs are defined.
+ * Returns the processed prompt and any additional vars extracted from JSON.
+ */
+function processPromptForInputs(
```
will-holley, line 20: DRY w/ the unaligned `processPromptForInputs`?

src/redteam/providers/crescendo/index.ts
```diff
@@ -102,6 +102,11 @@ interface CrescendoConfig {
    * Set by the layer strategy when used as: layer: { steps: [crescendo, audio] }
    */
   _perTurnLayers?: LayerConfig[];
+  /**
+   * Multi-input schema for generating multiple vars at each turn.
+   * Keys are variable names, values are descriptions.
+   */
+  inputs?: Record<string, string>;
```
will-holley, line 114: A global type would be preferable here and elsewhere `Record<string, string>` is in use

src/redteam/providers/crescendo/index.ts
```diff
@@ -766,9 +781,36 @@ export class CrescendoProvider implements ApiProvider {
     shouldFetchTrace?: boolean,
     traceSnapshots?: TraceContextData[],
   ): Promise<{ response: TargetResponse; transformResult?: TransformResult }> {
+    // Extract JSON from <Prompt> tags if present (multi-input mode)
+    let processedPrompt = attackPrompt;
+    const promptMatch = /<Prompt>([\s\S]*?)<\/Prompt>/i.exec(attackPrompt);
```
will-holley: You're killing me with this one

src/redteam/plugins/harmful/aligned.ts
faizanminhas: `prompt` in particular refers to a singular output from the LLM when doing test generation. Normally, it would be a line of:
```
Prompt: <whatever the plugin outputs>
```

But if we are in multi-input world, I've asked the plugin to output:

```
<Prompt>{ [input1]: ..., [input2]: ... }</Prompt>
```

A JSON string. If `hasInputs` === (redteam.inputs.length > 0), then we get JSON output and parse it here, to retrieve values to put into `testcase.vars` (since this function constructs the test cases)

src/redteam/plugins/base.ts
faizanminhas: I don't expect it to, but I can revert this since my rate of getting mal-formed outputs is actually very low

JustinBeckwith (commented):
I think this would greatly benefit from an example that folks can use to get started!

JustinBeckwith:
### Claude Code Review: Found a reliability issue with JSON parsing

Hey! I'm Claude - I was asked to review and test this PR locally. The multi-input generation feature works well conceptually, but I found that **LLM output formatting inconsistencies cause silent failures** in variable extraction.

#### The Problem

When testing with the `pii` plugin, 4 out of 8 generated test cases failed to extract the `conversation` and `htmlResult` variables. The root cause: the LLM sometimes outputs malformed JSON with trailing brackets.

**Example from debug logs:**
- ✅ Valid: `"}</Prompt>` (6 occurrences)
- ❌ Invalid: `"}]</Prompt>` (2 occurrences) - extra `]` breaks `JSON.parse`

#### Impact

When `JSON.parse` fails, the code silently falls back to using the raw JSON string as the prompt value, but the individual variables (`conversation`, `htmlResult`) aren't extracted. This means:
- The test case runs with `{{conversation}}` and `{{htmlResult}}` as empty/undefined
- No warning is shown to the user that extraction failed
- Test coverage is reduced without visibility

#### Suggested Fix

Add a `tryParseJson()` helper in `src/redteam/util.ts` that cleans common LLM formatting issues before parsing:

```typescript
/**
 * Attempts to parse JSON with cleanup for common LLM formatting issues.
 * Handles trailing brackets, whitespace, and other minor malformations.
 */
export function tryParseJson(jsonStr: string): Record<string, unknown> | null {
  let cleaned = jsonStr.trim();
  
  // Remove trailing characters that shouldn't be there (common LLM issue)
  // e.g., "}]" should be "}"
  while (cleaned.length > 0) {
    const lastChar = cleaned[cleaned.length - 1];
    if (lastChar === ']' || lastChar === ')') {
      const lastBrace = cleaned.lastIndexOf('}');
      if (lastBrace === cleaned.length - 2) {
        cleaned = cleaned.slice(0, -1);
        continue;
      }
    }
    break;
  }
  
  try {
    return JSON.parse(cleaned);
  } catch {
    return null;
  }
}
```

Then replace all 14 instances of `JSON.parse()` in this PR with the helper. The affected locations are:

| File | Function/Context |
|------|------------------|
| `src/redteam/index.ts` | `applyStrategies` (2 places) |
| `src/redteam/plugins/base.ts` | `parseGeneratedInputs`, `promptsToTestCases` |
| `src/redteam/plugins/harmful/aligned.ts` | `promptsToTestCases` |
| `src/redteam/plugins/harmful/unaligned.ts` | `processPromptForInputs` |
| `src/redteam/plugins/pii.ts` | `processPromptForInputs` |
| `src/redteam/providers/crescendo/index.ts` | variable extraction |
| `src/redteam/providers/goat.ts` | variable extraction |
| `src/redteam/providers/iterative.ts` | variable extraction |
| `src/redteam/providers/iterativeMeta.ts` | variable extraction |
| `src/redteam/providers/iterativeTree.ts` | variable extraction (multiple) |

#### How I Tested

```bash
PROMPTFOO_DISABLE_REMOTE_GENERATION=true promptfoo redteam generate \
  -c minputs.yaml -o generated-tests.yaml --no-cache
```

With this config:
```yaml
redteam:
  injectVar: prompt
  inputs:
    conversation: "A JSON array of conversation messages..."
    htmlResult: "Realistic HTML content..."
  plugins:
    - pii
```

---
🤖 *This review was performed by Claude Code*

src/redteam/plugins/harmful/aligned.ts
```diff
@@ -37,7 +38,35 @@ export class AlignedHarmfulPlugin extends RedteamPluginBase {
...
+        [this.injectVar]: __prompt,
+      };
+
+      // If inputs is defined, extract individual keys from the JSON into vars
+      if (hasMultipleInputs) {
+        try {
+          const parsed = JSON.parse(__prompt);
+          Object.assign(vars, extractVariablesFromJson(parsed, this.config.inputs!));
+        } catch {
+          // If parsing fails, just use the raw prompt
```
MrFlounder, line 57: adda debug log

src/redteam/plugins/base.ts
faizanminhas: Decided to namespace with `__outputFormat`

PR #6714, opened 2025-12-16
fix(providers): add gpt-image-1 support to OpenAI image provider

yash2998chhabria:
## Summary

- Adds support for OpenAI's `gpt-image-1` model in the image provider
- `gpt-image-1` uses different API parameters than DALL-E models:
  - Uses `output_format` (png/jpeg/webp) instead of `response_format` (url/b64_json)
  - Always returns base64 data, not URLs
- Modified `prepareRequestBody` to conditionally use correct parameters based on model
- Updated `formatOutput` to handle gpt-image-1 base64 JSON responses
- Updated cost calculation for gpt-image-1 quality tiers (low/medium/high)

## Test plan

- [x] Tested with `gpt-image-1` model - images generate successfully
- [x] Tested with DALL-E models - backward compatibility maintained
- [x] Verified base64 image data is correctly returned and can be saved via hooks

### DIFF
--- a/src/providers/openai/image.ts
+++ b/src/providers/openai/image.ts
@@ -13,16 +13,23 @@ import type {
 import type { EnvOverrides } from '../../types/env';
 import type { OpenAiSharedOptions } from './types';
 
-type OpenAiImageModel = 'dall-e-2' | 'dall-e-3' | 'gpt-image-1';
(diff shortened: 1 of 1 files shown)


yash2998chhabria:
Replaced by cleaner branch based on main

PR #6721, opened 2025-12-17
feat(fetch): add automatic retries for transient 5xx errors

danenania:
## Summary

Adds automatic retries for transient HTTP errors (502, 503, 504) to handle sporadic "Service Unavailable" errors reported by users.

The retry logic is baked into `fetchWithProxy` rather than requiring callers to switch to `fetchWithRetries`. This ensures all 30+ HTTP calls in the codebase automatically get transient retry capability without code changes at call sites, while keeping the change surface minimal for stability.

To avoid retrying non-transient errors (e.g., x.ai returns 502 for invalid API keys), we match on both status code AND status text (case-insensitive):
- 502 + "bad gateway"
- 503 + "service unavailable"  
- 504 + "gateway timeout"

Exponential backoff: 1s, 2s, 4s (up to 3 retries).

`fetchWithRetries` sets `disableTransientRetries: true` to avoid double-retrying.

## Test plan

- [x] Added tests for `isTransientError()` helper
- [x] Added tests for `fetchWithProxy` retry behavior
- [x] Added test for `fetchWithRetries` disabling transient retries
- [x] All 64 fetch tests pass


src/util/fetch/index.ts
```diff
@@ -135,7 +135,30 @@ export async function fetchWithProxy(
     setGlobalDispatcher(agent);
   }
 
-  return await monkeyPatchFetch(finalUrl, finalOptions);
+  // Transient error retry logic (502/503/504 with matching status text)
```
jameshiester, line 139: should this respect PROMPTFOO_RETRY_5XX?

danenania: The idea is to have auto retries on requests that are certain to be transient, based on the specific error code _and_ the error content. The assumption being that this is always desired and is separate from any other retry considerations (like the PROMPTFOO_RETRY_5XX which is broader in that it would retry on any 5xx (like a 500) regardless of the specific error message).

Issue #6727, opened 2025-12-17
Feature Request: Improve the UX of the red team target type multi-select


Ebonsignori:
When selecting a target type for a red team configuration, the available targets aren't immediately clear to the user. You need to press the "✏️ Change" button to get into a new view that lets you filter down and select a target. Upon selecting a target, you are moved into a separate view that hides other options. 

From a "UX" way of thinking, I think this violates progressive disclosure because of unnecessary abstraction. Additionally, it's an extra click on a page that already has sufficiently minimal information (just two related forms and navigation). 

I suggest a view that **always** surfaces the target type options by removing a separate view. 

Current view example:

<img width="1367" height="676" alt="Image" src="https://github.com/user-attachments/assets/61050047-1b3f-4274-ad23-fc60ebf7cfa2" />

mldangelo-oai:
Thanks for the thoughtful UX write-up here. I’m closing this as implemented in the current red-team setup flow.

Target type selection is now surfaced directly in the target setup step instead of being hidden behind the older separate “Change” interaction described in the issue, so the current UI matches the direction you were advocating for much more closely.

Really appreciate the clear feedback — it was a helpful nudge toward a simpler flow.

PR #6728, opened 2025-12-17
feat(site): change redteam target type multi-select to be always expanded

Ebonsignori:
Closes https://github.com/promptfoo/promptfoo/issues/6727

### Changes 

Updates the red team config page's Target Type selector to be always-expanded. The target type options (providers) are scoped to a scrolling box element on the page. 

When a target type is updated by "Load Example", or the user navigates to the "Target Type" tab, the selected target type will scroll into view within the box element. 

The filtering and category tag selection works the same as before, but above them and underneath the Target Type description is a persistent "Currently selected" element to make it clear to the user what is selected.

### Out of scope changes

Changes the copy of "Select Target Type" to just "Target Type" since the "Select" is implied. Especially with the new "Currently selected" copy on the element beneath it 

### Example

**Before**
image

**After**
image

> [!Note]
> Not sure if `feat(site)` is the correct prefix for this PR


mldangelo:
Thank you @Ebonsignori! I'll reach out to book some time with you tomorrow!

will-holley:
Thanks for the PR, @Ebonsignori. I checked it out and ran it locally; a few things that I noticed:

- When going into this flow for the first time, and clicking "Select Target Type" the auto-scroll to the selected target type is very jarring. I'm open to asking the user to select a target type instead of assuming HTTP and flying past all of the other options.
- The duplicate selected provider UI feels off to me; don't have a good suggestion here.

Ebonsignori:
This is low priority

PR #6760, opened 2025-12-18
feat(auth): add interactive team selection during login

MrFlounder:
## Why

Users with access to multiple teams were accidentally sharing evals to the wrong team because `promptfoo auth login` always defaulted to the first team without asking.

## What

- Add `--team` flag to `promptfoo auth login` for CI/CD use cases
- Add interactive team selection prompt when multiple teams exist
- Inject current team ID into eval metadata before uploading to cloud
- Scope "already shared" check to current team only (prevents false positives when eval exists in a different team)

## How to Test

1. Login with an account that has access to multiple teams
2. Verify interactive team selection prompt appears
3. Share an eval and confirm it goes to the selected team
4. Try re-sharing the same eval - verify it correctly detects existing eval in current team

### DIFF
--- a/test/commands/auth.test.ts
+++ b/test/commands/auth.test.ts
@@ -1,7 +1,6 @@
 import select from '@inquirer/select';
 import { Command } from 'commander';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
-
 import { authCommand } from '../../src/commands/auth';
 import { isNonInteractive } from '../../src/envars';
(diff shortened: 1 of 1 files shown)


faizanminhas:
is this backwards compatible with the old cli? or would auth login always require interactivity now?

MrFlounder:
old cli login will use default (oldest) team, it's backwards compatible AFAIN @faizanminhas

PR #6774, opened 2025-12-19
fix(evaluator): apply comparison results to pass/fail

typpo:
Ensure select-best/max-score updates overall result status and stats


src/evaluator.ts
```diff
@@ -1699,6 +1704,31 @@ class Evaluator {
...
+            }
+          }
+          if (metrics) {
+            metrics.assertPassCount += gradingResult.pass ? 1 : 0;
+            metrics.assertFailCount += gradingResult.pass ? 0 : 1;
+            if (gradingResult.tokensUsed) {
+              updateAssertionMetrics(metrics, gradingResult.tokensUsed);
+            }
+            if (!gradingResult.pass && result.score !== wasScore) {
+              metrics.score += result.score - wasScore;
```
JustinBeckwith: parens would make this a little more obvious to read

PR #6782, opened 2025-12-19
fix: prompt id hashing

typpo:
fixes a bug where a prompt with an id but not label would hash undefined

### DIFF
--- a/src/models/prompt.ts
+++ b/src/models/prompt.ts
@@ -22,5 +22,11 @@ interface PromptForIdGeneration {
  * @returns A SHA-256 hash string
  */
 export function generateIdFromPrompt(prompt: PromptForIdGeneration): string {
-  return sha256(prompt.label ?? prompt.id ?? typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
+  if (prompt.label) {
+    return sha256(prompt.label);
+  }
+  if (prompt.id) {
+    return sha256(prompt.id);
+  }
+  return sha256(typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
 }


src/models/prompt.ts
```diff
@@ -6,7 +6,11 @@ import type { PromptSchema } from '../validators/prompts';
...
-  return prompt.id || prompt.label
-    ? sha256(prompt.label)
-    : sha256(typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
+  if (prompt.label) {
+    return sha256(prompt.label);
+  }
+  if (prompt.id) {
+    return sha256(prompt.id);
+  }
+  return sha256(typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
```
faizanminhas:
```suggestion
  return sha256(prompt.label ?? prompt.id ?? typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
```

src/prompts/utils.ts
```diff
@@ -87,7 +87,11 @@ export function normalizeInput(
...
-  return prompt.id || prompt.label
-    ? sha256(prompt.label)
-    : sha256(typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
+  if (prompt.label) {
+    return sha256(prompt.label);
+  }
+  if (prompt.id) {
+    return sha256(prompt.id);
+  }
+  return sha256(typeof prompt.raw === 'object' ? JSON.stringify(prompt.raw) : prompt.raw);
```
faizanminhas: extract out to use the same function from `models/prompt.ts`

PR #6794, opened 2025-12-20
fix: Entra id support in azure:responses provider

Halcyonhal9:
I noticed that the azure:responses provider throws an auth error when using managed identity, whereas azure:chat doesn't.

Here's a very basic yaml calling responses (while using managed identity via "az login"):
image

If use the same yaml but with azure:chat, it works.

From investigating the issue, it's due to an if statement.  The or check for; bearer or api key means that if the api key is not there, it will always throw.  I've updated the or to what i think was the original intention. 

Now entra id and api key works.  
image

I also checked that api key or entra id correctly throws an error.

### DIFF
--- a/src/providers/azure/assistant.ts
+++ b/src/providers/azure/assistant.ts
@@ -1,6 +1,7 @@
 import { fetchWithCache, getCache, isCacheEnabled } from '../../cache';
 import logger from '../../logger';
 import { maybeLoadToolsFromExternalFile } from '../../util/index';
+import invariant from '../../util/invariant';
 import { sleep } from '../../util/time';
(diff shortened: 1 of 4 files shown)


mldangelo:
Thank you @Halcyonhal9! Please reach out to michael @ promptfoo.dev if you'd like some swag. I'll get this merged and deployed for you.

Halcyonhal9:
> Thank you @Halcyonhal9! Please reach out to michael @ promptfoo.dev if you'd like some swag. I'll get this merged and deployed for you.

no sweat... love the product.  

Ben Appleby 
GPM | Microsoft
bena at microsoft.com

PR #6807, opened 2025-12-22
fix(eval): prevent icon overflow in results table cells

addelong:
## Summary
- Fixed icon overflow in eval results table cells by adding `flex-wrap: wrap` and `max-width` constraints to `.cell-actions`
- Icons now wrap to new lines instead of overflowing when columns are narrow

## Test plan
- [ ] Open an eval results page with narrow columns
- [ ] Verify icons wrap to new lines instead of overflowing
- [ ] Press Shift to show additional icons and verify wrapping still works


src/app/src/pages/eval/components/ResultsTable.css
```diff
@@ -153,12 +153,15 @@ table.results-table,
...
   display: flex;
+  flex-wrap: wrap;
+  justify-content: flex-end;
   gap: 0.5rem;
   position: absolute;
   top: 1rem;
   right: 0.75rem;
   line-height: 0;
   font-size: 1.25rem;
+  max-width: calc(100% - 1.5rem);
```
faizanminhas: this seems brittle, can you think of an alternative?

addelong: sure!

PR #6810, opened 2025-12-23
fix(redteam): filter runtime vars and merge defaultTest.vars for accurate test matching

minhle1291:
When using --filter-failing or --filter-errors-only with multi-turn strategies
(GOAT, Crescendo, SIMBA), the filter returned 0 tests because:

1. Provider-added variables like sessionId were stored in result.vars but not
   in the original test case, causing varsMatch() to fail
2. When filterTests runs, fresh tests haven't had defaultTest.vars merged yet
   (that happens later in prepareTests), but stored results have merged vars

This fix:
- Extends filterRuntimeVars to filter both _conversation and sessionId
- Creates shallow copy of test.vars in runEval to prevent mutation by reference
- Merges defaultTest.vars into test cases before comparison in filterTestsByResults

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>


src/commands/eval/filterTests.ts
```diff
@@ -109,11 +114,31 @@ export async function filterTests(testSuite: TestSuite, options: FilterOptions):
...
-    tests = await filterFailingTests(testSuite, options.failing);
-  }
+  // Handle failing and errorsOnly filters
+  // When both are provided, combine results (union) instead of overwriting
+  if (options.failing && options.errorsOnly) {
+    const failingTests = await filterFailingTests(testSuite, options.failing);
+    const errorTests = await filterErrorTests(testSuite, options.errorsOnly);
 
-  if (options.errorsOnly) {
+    // Create a union of both sets using a Map keyed by test identity
```
will-holley: Out of curious why not use [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)?

will-holley (approved):
Didn't test locally but the code LGTM.

PR #6817, opened 2025-12-23
feat(mcp): add OAuth authentication with proactive token refresh

jameshiester:
## Summary

- Add OAuth 2.0 authentication support for MCP servers (client_credentials and password grants)
- Implement proactive token refresh with 60-second buffer before expiry
- Add token refresh locking to prevent concurrent refresh attempts
- Add RFC 8414 token endpoint discovery when `tokenUrl` is not configured
- Extract shared OAuth token fetching logic to `src/util/oauth.ts` for reuse between MCP and HTTP providers
- Add comprehensive authentication documentation to MCP provider docs
- Add `redteam-mcp-auth` example demonstrating OAuth configuration

## Authentication Types Supported

- Bearer token
- Basic authentication
- API key (header or query placement)
- OAuth 2.0 (client_credentials and password grants)

## Token Endpoint Discovery

If `tokenUrl` is not specified, the provider automatically discovers the token endpoint using [RFC 8414](https://datatracker.ietf.org/doc/rfc8414/) OAuth 2.0 Authorization Server Metadata. It tries multiple well-known URLs:

1. Path-appended: `{server-url}/.well-known/oauth-authorization-server` (Keycloak style)
2. RFC 8414 path-aware: `{origin}/.well-known/oauth-authorization-server{path}`
3. Root level: `{origin}/.well-known/oauth-authorization-server`

## Token Refresh Behavior

1. Tokens are proactively refreshed 60 seconds before expiration
2. Concurrent requests share the same refresh operation (no duplicate token fetches)
3. If a token expires during an evaluation, the provider automatically reconnects with a fresh token

## Test plan

- [x] Unit tests for shared OAuth utilities (`test/util/oauth.test.ts`)
- [x] Unit tests for MCP auth provider (`test/providers/mcp/authProvider.test.ts`)
- [x] Unit tests for MCP client OAuth refresh (`test/providers/mcp/client.test.ts`)
- [x] Unit tests for token endpoint discovery (`test/providers/mcp/util.test.ts`)
- [ ] Manual testing with OAuth-protected MCP server

### DIFF
--- a/examples/redteam-auth/README.md
+++ b/examples/redteam-auth/README.md
@@ -26,14 +26,14 @@ The `promptfooconfig.oauth.yaml` file demonstrates OAuth 2.0 client credentials 
 targets:
   - id: http
     config:
-      url: https://example-bot.promptfoo.app/minnow/chat?auth_type=bearer
+      url: https://example-app.promptfoo.app/minnow/chat?auth_type=bearer
       method: POST
       auth:
         type: oauth
         grantType: client_credentials
         clientId: '{{env.PROMPTFOO_TARGET_CLIENT_ID}}'
         clientSecret: '{{env.PROMPTFOO_TARGET_CLIENT_SECRET}}'
-        tokenUrl: https://example-bot.promptfoo.app/oauth/token
+        tokenUrl: https://example-app.promptfoo.app/oauth/token
         scopes: []
 ```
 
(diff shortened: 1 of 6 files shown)


site/docs/providers/mcp.md
```diff
@@ -114,9 +139,105 @@ providers:
         url: https://secure-mcp-server.com
         auth:
           type: api_key
-          api_key: 'your-api-key'
+          value: '{{env.MCP_API_KEY}}'
+          keyName: X-API-Key # Header or query parameter name (default: X-API-Key)
```
mldangelo, line 143: why not combine these into one key: value ?

jameshiester: I think this is a better approach.
1. This plays nicer with zod validation and state management
2. We are already using this approach for HTTP provider.

PR #6829, opened 2025-12-27
fix(providers): respect PROMPTFOO_PYTHON env var in Python providers

### LINKED ISSUE #6819: Custom providers don't respect $PROMPTFOO_PYTHON
**Describe the bug**

When using `PROMPTFOO_PYTHON` to specify a Python virtual environment, providers are not executed using this environment. It executes in the system default environment instead. This is evident when the virtual environment contains packages not present in the system environment.

**To Reproduce**

1. Create a Python virtual environment.
2. Inside this environment, install a package you don't have in your system environment, e.g. `python -m pip install pypdfium2`
3. Verify this package is not in your default environment. Navigate away from the virtual environment and run `python -m pip show pypdfium2` and observe a warning that the package was not found.
4. Create a promptfoo config that loads a Python provider: `python:./my_provider.py`
5. In this provider, import the package: `import pypdfium2`
6. Export `PROMPTFOO_PYTHON` to point to your virtual environment.
7. Run `promptfoo eval`.

**Expected behavior**
GIVEN a Python virtual environment
AND the path to this environment is set in the `PROMPTFOO_PYTHON` environment variable
AND a custom provider is specified using `python:./my_provider.py`
WHEN I run `promptfoo eval`
THEN the provider should execute in the Python virtual environment.

**System information:**
```json
{
  "version": "0.120.8",
  "platform": {
    "os": "darwin",
    "release": "25.2.0",
    "arch": "arm64",
    "nodeVersion": "v24.12.0"
  },
  "env": {
    "telemetryDisabled": false,
    "telemetryDebug": false
  },
  "configInfo": {
    "configExists": false,
    "configContent": null
  }
}
```

mldangelo:
## Summary

Fixes #6819

Custom Python providers were ignoring the `PROMPTFOO_PYTHON` environment variable, always falling back to the system default Python instead of using the configured virtual environment.

**Root cause:** `PythonProvider.initialize()` only passed `config.pythonExecutable` to the worker pool without checking `PROMPTFOO_PYTHON` as a fallback.

**Changes:**
- Add centralized `getConfiguredPythonPath()` utility in `pythonUtils.ts` that handles the fallback chain: `config.pythonExecutable` > `PROMPTFOO_PYTHON` > `undefined`
- Update `PythonProvider.initialize()` to use the new utility
- Refactor `runPython()` to use the same utility for DRY consistency
- Add comprehensive unit and integration tests

**Priority chain (now correct):**
1. `config.pythonExecutable` (explicit config in promptfooconfig.yaml)
2. `PROMPTFOO_PYTHON` (environment variable)
3. System default (`'python'`)

## Test plan

- [x] Added 6 unit tests for `getConfiguredPythonPath()` 
- [x] Added 4 integration tests for `PythonProvider` PROMPTFOO_PYTHON handling
- [x] All 98 Python-related tests pass
- [x] TypeScript compiles successfully
- [x] Linting passes

### DIFF
--- a/src/python/pythonUtils.ts
+++ b/src/python/pythonUtils.ts
@@ -28,6 +28,26 @@ export function getEnvInt(key: string): number | undefined {
   return isNaN(parsed) ? undefined : parsed;
 }
 
+/**
+ * Resolves the Python executable path from explicit config and environment.
+ * This centralizes the fallback logic: configPath > PROMPTFOO_PYTHON env var.
+ *
+ * Note: Does NOT apply the final 'python' default - that's handled by
+ * validatePythonPath. This preserves the distinction between "explicitly
+ * configured" (should fail if invalid) and "using system default" (should
+ * try fallback detection).
+ *
+ * @param configPath - Explicitly configured Python path from provider config
+ * @returns The configured path, or undefined if neither config nor env var is set
+ */
+export function getConfiguredPythonPath(configPath?: string): string | undefined {
+  if (configPath) {
+    return configPath;
+  }
+  const envPath = getEnvString('PROMPTFOO_PYTHON');
+  return envPath || undefined;
+}
+
 export const state: {
   cachedPythonPath: string | null;
   validationPromise: Promise<string> | null;
@@ -289,7 +309,7 @@ export async function runPython(
     os.tmpdir(),
     `promptfoo-python-output-json-${Date.now()}-${Math.random().toString(16).slice(2)}.json`,
   );
-  const customPath = options.pythonExecutable || getEnvString('PROMPTFOO_PYTHON');
+  const customPath = getConfiguredPythonPath(options.pythonExecutable);
   let pythonPath = customPath || 'python';
 
   pythonPath = await validatePythonPath(pythonPath, typeof customPath === 'string');
--- a/src/providers/pythonCompletion.ts
+++ b/src/providers/pythonCompletion.ts
@@ -3,7 +3,7 @@ import path from 'path';
 
 import { getCache, isCacheEnabled } from '../cache';
 import logger from '../logger';
-import { getEnvInt } from '../python/pythonUtils';
+import { getConfiguredPythonPath, getEnvInt } from '../python/pythonUtils';
 import { PythonWorkerPool } from '../python/workerPool';
 import { sha256 } from '../util/createHash';
 import { processConfigFileReferences } from '../util/fileReference';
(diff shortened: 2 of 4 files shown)


ghost:
**⏩ No test execution environment matched (6d4d38556d225bab8d7de2a591886aa6fd585c05) [View output ↗](https://app.usetusk.ai/app/testing-commit-check/e94348f4-baa0-476e-965e-a75c44d72496?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

---
<p>

| 6d4d38556d225bab8d7de2a591886aa6fd585c05 | ⏩ No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/e94348f4-baa0-476e-965e-a75c44d72496?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 27, 2025 9:05AM |

</p>
</details>

[View output in GitHub ↗](https://github.com/promptfoo/promptfoo/pull/6829/checks?check_run_id=58996634981)

src/python/pythonUtils.ts
```diff
@@ -28,6 +28,26 @@ export function getEnvInt(key: string): number | undefined {
   return isNaN(parsed) ? undefined : parsed;
 }
 
+/**
+ * Resolves the Python executable path from explicit config and environment.
+ * This centralizes the fallback logic: configPath > PROMPTFOO_PYTHON env var.
+ *
+ * Note: Does NOT apply the final 'python' default - that's handled by
+ * validatePythonPath. This preserves the distinction between "explicitly
```
faizanminhas, line 36: a thought- should this be part of the logic in validatePythonPath? instead of having validation / assignment in 2 places

faizanminhas (approved):

src/python/pythonUtils.ts
mldangelo: Good question! I considered that, but kept them separate for a few reasons:

  1. Separation of concerns - getConfiguredPythonPath() answers "what did the user configure?" while validatePythonPath() answers "does this path work?"
  2. The isExplicit flag behavior - In worker.ts, we pass typeof this.pythonPath === 'string' as isExplicit. When true, validation fails hard if the path doesn't work. When false, it tries system detection as fallback. If we checked PROMPTFOO_PYTHON inside validatePythonPath, we'd lose the ability to distinguish "user explicitly set this env var" (should fail if invalid) from "no config at all" (should try fallback).
  3. Would require signature changes - Moving this logic into validatePythonPath would mean either checking the env var internally (changing the function's contract) or adding another parameter.

  The two places that call getConfiguredPythonPath are PythonProvider.initialize() and runPython() - both need the same resolution logic before passing to validation, so it felt right to centralize that piece.

  Happy to discuss if you see a cleaner approach though!

PR #6832, opened 2025-12-27
fix(evals): browser back button fix on evals results table

faizanminhas:
## Summary
- Verifies component doesn't pollute browser history on mount
- Documents that `handleSearchTextChange` uses `{ replace: true }` for search updates

#### Demo
https://github.com/user-attachments/assets/6f276baf-d862-459d-9cd7-568fdae28bf1


mldangelo (approved):
Thanks!

PR #6857, opened 2025-12-29
feat(auth): implement OAuth 2.0 Device Authorization Grant for CLI login

mldangelo:
## Summary

Adds RFC 8628 device authorization login for interactive CLI users while retaining API-key auth for automation.

## Design

- Requests and polls with public client id `promptfoo-cli`; token polling keeps RFC `slow_down` increments and defensively applies a five-second floor to malformed or overly aggressive intervals.
- Explicit `--host` API bases preserve path prefixes for mounted Enterprise deployments and reject credentials, query parameters, and fragments; API-key and device flows now share this normalization.
- Device and token responses are validated before use; non-JSON gateway responses produce actionable errors.
- Device credentials are printed only to the terminal and are redacted from request/response debug logs (`device_code`, `user_code`, verification completion URI, and token).
- Documentation and screenshot tooling use the same request contract.

Protocol reference: [RFC 8628](https://www.rfc-editor.org/rfc/rfc8628.html).

## Test plan

- [x] `npm run f` and `npm run l`
- [x] `npm run tsc`
- [x] `npx vitest --run test/commands/auth.test.ts test/util/sanitizer.test.ts` (204 passing)
- [x] `npm run build`
- [x] `npm run test:app -- --run` (4,540 passing)
- [x] `npm test` (18,186 passing, 10 skipped)
- [x] Isolated built-CLI `auth login` against a disposable path-prefixed RFC 8628 endpoint: `/tenant` is preserved across API calls, zero interval waits five seconds, requests include `client_id`, and login completes.
- [x] Inspected fresh isolated logs and database: state files are mode `0600`; raw device code, user code, and token are absent while redaction markers are present.
- [ ] `npm run test:smoke`: 13 files / 162 tests pass, with the single `not-script-assertions` case blocked locally because this machine has no `ruby` executable for its `not-ruby` fixture.


mldangelo-oai:
@codex review

mldangelo-oai:
Implemented the remaining device-auth review feedback and pushed `6dcf463e5`.

Changes:
- Prevent saved Cloud auth from overriding explicit `Authorization` headers, and mark direct token validation as `skipCloudAuth` so the freshly received device token is validated.
- Make the “Promptfoo Cloud” device-login option ignore a saved Enterprise/self-hosted `apiHost`, while still honoring explicit `PROMPTFOO_CLOUD_API_URL` / `API_HOST` env overrides.
- Added regression coverage for both cases.

Verification:
- `npx vitest run test/commands/auth.test.ts test/monkeyPatchFetch.test.ts test/util/sanitizer.test.ts --sequence.shuffle=false`
- `npm run f && npm run l`
- `git diff --check`
- `npm run tsc`
- `npm run build`
- Built CLI E2E device-code login against a local mock Cloud host with isolated `PROMPTFOO_CONFIG_DIR` / `PROMPTFOO_CACHE_PATH`; verified unauthenticated code/token requests and fresh-token validation/team lookup.
- `npm test -- --run` (718 files passed, 18,204 tests passed, 10 skipped)
- `npm run jsonSchema:generate` (no schema diff)

Resolved the two review threads for cached token validation and Cloud-host selection. I’m moving on to the next PR now per the loop instruction; new CI is queued/running on the pushed commit.

src/commands/auth.ts
```diff
@@ -251,30 +311,297 @@ async function loginWithApiKey(cmdObj: LoginCommandOptions, apiHost: string): Pr
...
+    );
+  }
+
+  throw new Error('Device code expired. Please try again.');
+}
+
+async function openDeviceVerificationUrl(url: string): Promise<void> {
+  try {
+    logger.info('Opening device authorization page in your browser...');
+    await opener(url);
```
jameshiester-oai, line 456: **[P1] Avoid launching a server-controlled URL through `cmd.exe`**

`verification_uri_complete` comes from the unauthenticated device-code response, and the validator accepts `https://example.test/|calc.exe`. With the locked `opener@1.5.2`, a forced-Windows capture produced `cmd.exe /c start "" https://example.test/|calc.exe`. Its Windows path escapes `&` and `^` but not the `|` shell metacharacter, so a malicious or compromised auth endpoint can execute a command as the logged-in user when this automatic open runs. This is the unsafe boundary described in [GHSL-2020-145](https://securitylab.github.com/advisories/GHSL-2020-145-domenic-opener/). Please use a launcher with argument-safe Win32 semantics (not `cmd.exe /c start`) and add regression coverage for Windows shell metacharacters.

src/commands/auth.ts
```diff
@@ -251,30 +311,297 @@ async function loginWithApiKey(cmdObj: LoginCommandOptions, apiHost: string): Pr
...
+  await completeCloudLogin(cmdObj, cmdObj.apiKey!, apiHost);
+}
+
+async function requestDeviceCode(apiHost: string): Promise<DeviceCodeResponse> {
+  const response = await fetchWithTimeout(
+    `${apiHost}/api/v1/auth/device/code`,
+    {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json', ...DEVICE_AUTH_SILENT_HEADER },
+      body: JSON.stringify({ client_id: DEVICE_CLIENT_ID }),
```
jameshiester-oai, line 324: **[P2] Encode the device requests as RFC 8628 form bodies**

[RFC 8628 §§3.1 and 3.4](https://www.rfc-editor.org/rfc/rfc8628.html#section-3.1) require both the device-authorization and token requests to use UTF-8 `application/x-www-form-urlencoded` bodies. This JSON request, and the JSON token request below, work only with a permissive/proprietary server. A strict form-only fixture returned HTTP 415 when exercised through the built CLI, while the unit tests currently lock in `JSON.parse`. Please use `URLSearchParams` (with the form content type) for both requests and test against a strict decoder.

src/commands/auth.ts
```diff
@@ -251,30 +311,297 @@ async function loginWithApiKey(cmdObj: LoginCommandOptions, apiHost: string): Pr
...
+    if (Date.now() >= expiresAtMs) {
+      break;
+    }
+
+    const requestTimeoutMs = Math.min(
+      DEVICE_AUTH_REQUEST_TIMEOUT_MS,
+      Math.max(1, expiresAtMs - Date.now()),
+    );
+
+    const response = await fetchWithTimeout(
```
jameshiester-oai, line 388: **[P2] Keep recoverable transport failures inside the polling state machine**

A rejection here bypasses every protocol branch and terminates login. With a hanging token endpoint and `REQUEST_TIMEOUT_MS=100`, the built CLI exited after the first request with `Request timed out after 100 ms`; a first-poll socket drop also terminated immediately. [RFC 8628 §3.5](https://www.rfc-editor.org/rfc/rfc8628.html#section-3.5) requires clients to reduce polling frequency on connection timeout before continuing. Conversely, the generic fetch layer can retry 502/503/504 three times inside this single poll, violating the advertised interval. Disable those inner transient retries for this endpoint, catch recoverable transport errors here, increase/back off the poll interval, and continue until the device-code deadline.

src/commands/auth.ts
```diff
@@ -251,30 +311,297 @@ async function loginWithApiKey(cmdObj: LoginCommandOptions, apiHost: string): Pr
...
+    typeof data !== 'object' ||
+    !isSafeCredentialString((data as DeviceCodeResponse).device_code) ||
+    !isSafeCredentialString((data as DeviceCodeResponse).user_code) ||
+    !isSafeVerificationUrl((data as DeviceCodeResponse).verification_uri) ||
+    ((data as DeviceCodeResponse).verification_uri_complete !== undefined &&
+      !isSafeVerificationUrl((data as DeviceCodeResponse).verification_uri_complete)) ||
+    !Number.isFinite((data as DeviceCodeResponse).expires_in) ||
+    (data as DeviceCodeResponse).expires_in <= 0 ||
+    ((data as DeviceCodeResponse).interval !== undefined &&
+      !Number.isFinite((data as DeviceCodeResponse).interval))
```
jameshiester-oai, line 354: **[P2] Bound server-provided timing values before using Node timers**

`Number.isFinite` still accepts values far beyond Node's timer range. A fixture returning `interval: 3000000` and `expires_in: 3000000` caused `TimeoutOverflowWarning`, clamped the delay to 1 ms, and polled immediately; `Number.MAX_VALUE` can also make the deadline infinite. A pending response then becomes a near-tight request loop controlled by the server. Require safe bounded integer seconds (or chunk long waits), cap the overall login lifetime, and add boundary tests around the maximum supported delay.

src/commands/auth.ts
```diff
@@ -251,30 +311,297 @@ async function loginWithApiKey(cmdObj: LoginCommandOptions, apiHost: string): Pr
...
+
+function normalizeApiHost(value: string): string {
+  const url = new URL(value);
+  if (!['http:', 'https:'].includes(url.protocol)) {
+    throw new Error('Host must be an HTTP or HTTPS URL');
+  }
+  if (url.username || url.password || url.search || url.hash) {
+    throw new Error('Host URL must not include credentials, query parameters, or a fragment');
+  }
+  return url.toString().replace(/\/+$/, '');
```
jameshiester-oai, line 498: **[P3] Reject or remove empty query/fragment delimiters**

WHATWG URL exposes `.search === ""` for a trailing bare `?` and `.hash === ""` for a trailing bare `#`, but `toString()` preserves those delimiters. The built CLI accepted `--host 'http://127.0.0.1:39185/tenant?'` and requested `/tenant?/api/v1/auth/device/code`; with `#`, the appended API path is treated as a fragment and never reaches the server. Inspect the raw/serialized URL for empty delimiters or normalize them away before appending paths, and cover `?`, `#`, `/?`, and `/#`.

site/scripts/capture-device-auth-screenshots.mjs
```diff
@@ -0,0 +1,183 @@
...
+        path: join(OUTPUT_DIR, 'device-auth-already-logged-in.png'),
+        clip: { x: 0, y: 0, width: 600, height: 800 },
+      });
+      console.log('  Saved: device-auth-already-logged-in.png');
+    } else {
+      console.log('  Warning: User not logged in, showing email form instead');
+    }
+    await page3.close();
+
+    console.log('\nAll screenshots captured successfully!');
```
jameshiester-oai, line 173: **[P3] Fail when an expected screenshot was skipped**

Both missing-button branches above only print warnings and skip writing their PNG, then this unconditional message reports success with exit code 0. A stale checked-in image can therefore mask a broken auth state, and a clean checkout can finish without all documented assets. Track the three expected outputs from this run (or throw when a required button is absent), verify each was written, and exit nonzero after browser cleanup if any capture failed.

jameshiester-oai (commented):
## Codex review

**Reviewed base:** `c54a30668ad8319d76c20ae96e6680ad6c51a2c6`  
**Reviewed head:** `341815ab5ed8d2ca83ebec2af4e2e3fcc4238d52`  
**Scope:** Full fresh `base...head` review. No prior authenticated completion marker exists for this head.

### Verdict

This should not merge yet. I verified one P1 Windows command-injection boundary, three P2 protocol/correctness defects, and two P3 edge/tooling defects. The PR is also currently conflicting with `main`.

CI has three red jobs, but their test phases passed before Codecov GPG-key failures: Site (101 tests), Node 20 Ubuntu (19,674 tests; 10 skipped), and webui (2 browser tests plus 4,816 unit tests). Other visible build, test, smoke, integration, platform, style, title, and CodeQL checks are green.

### Verified findings

**P1**
- The unauthenticated device endpoint controls the verification URL that is passed to `opener@1.5.2`. Its Windows implementation invokes `cmd.exe /c start` without escaping `|`, allowing a crafted URL to execute a shell command when the CLI auto-opens it.

**P2**
- Both RFC 8628 POSTs use JSON instead of the required `application/x-www-form-urlencoded` encoding; a strict conforming endpoint rejected the built CLI with HTTP 415.
- Token polling aborts on a connection timeout or dropped socket instead of reducing frequency and retrying until expiry. Generic 502/503/504 retries can also issue several token requests inside one advertised polling interval.
- Finite but unbounded `expires_in` / `interval` values overflow Node timers. A 3,000,000-second interval was clamped to 1 ms and polled immediately.

**P3**
- Bare `?` and `#` delimiters evade the host query/fragment check. The built CLI appended the API path after `?`, and a fragment would hide the suffix from the HTTP request.
- The screenshot script warns and skips files when expected buttons are absent, then reports success, so stale or missing docs assets can pass unnoticed.

Detailed reproductions and correction directions are attached inline.

### Existing feedback

All 13 existing review threads are resolved. I rechecked the important prior findings: cached credentials are no longer sent to public device endpoints, explicit fresh-token authorization is preserved, Cloud host overrides are honored, path prefixes are retained, and optional completion URIs/non-JSON responses are handled. I did not duplicate those threads.

### Tests and QA

- `npx vitest run test/commands/auth.test.ts test/monkeyPatchFetch.test.ts test/util/sanitizer.test.ts` — exit 0; 3 files, 251 tests passed.
- `npm run tsc` — exit 0.
- `npm run build` — exit 0, including core and app builds.
- `SKIP_OG_GENERATION=true npm --prefix site run build -- --no-minify` — exit 0.
- `git diff --check c54a30668ad8319d76c20ae96e6680ad6c51a2c6...341815ab5ed8d2ca83ebec2af4e2e3fcc4238d52` — exit 0.
- `npm run local -- auth login --host http://127.0.0.1:39185/tenant` — exit 0 against an isolated path-prefixed device server.
- Built CLI success path with a stale dummy token completed pending → `slow_down` → success; the public calls had no Authorization header, the path prefix was preserved, and validation/team calls used the fresh token.
- Built CLI failure/edge probes covered invalid scheme, invalid JSON, strict form-only server (415), first-poll socket drop, oversized timing values, and a host ending in bare `?`.
- `env REQUEST_TIMEOUT_MS=100 PROMPTFOO_CONFIG_DIR=/private/tmp/promptfoo-pr6857-config-timeout PROMPTFOO_API_KEY=dummy-stale-token PROMPTFOO_DISABLE_TELEMETRY=true PATH="/private/tmp/promptfoo-pr6857-bin:$PATH" node dist/src/main.js auth login --host http://127.0.0.1:39185/timeout` — exit 1 after the first token request with `Request timed out after 100 ms`.
- Visually inspected all three new screenshots; they are legible and match the documented states.

The highest-value test additions are strict form-encoding integration coverage, recoverable transport failures, timer boundaries, bare delimiter hosts, Windows shell metacharacters, and current-run screenshot-output assertions. A shared form-body builder plus a validated/bounded timing parser would simplify the two protocol requests without adding broad abstraction.

### Limitations

The requested PR Review Toolkit command was unavailable in this environment, so I performed the six applicable review lenses manually and reconciled three independent read-only reviews. I did not have a Windows runtime; the P1 was validated from the locked dependency source and a forced-Windows spawn-argument capture. I did not run the screenshot capture script against live site/API services or rerun the full repository suites locally. The PR body's `0600` state-file claim did not reproduce under a fresh umask-022 config (files were `0644`), but that writer predates this PR, so I have not attributed it as a changed-code defect.

PR #6870, opened 2025-12-29
chore(app): eliminate all Node.js polyfills from frontend build

JustinBeckwith:
## Summary

This PR removes all Node.js polyfill dependencies from the frontend build, resulting in a cleaner, faster build with zero warnings.

### Dependency Cleanup
- ✅ Removed `vite-plugin-node-polyfills` package (freed 100 packages)
- ✅ Removed all polyfill configuration from vite.config.ts
- ✅ Removed unused dotenv stub directory

### CSV Module Fixes
- ✅ Switched from `csv-parse/sync` (Node.js) to `csv-parse/browser/esm/sync` (browser)
- ✅ Converted static imports to dynamic imports for code-splitting
- ✅ Updated test mocks to match new import paths
- ✅ Files affected:
  - `TestCasesSection.tsx` (eval creator)
  - `CustomIntentPluginSection.tsx` (red team setup)
  - `CustomPoliciesSection.tsx` (red team setup)

### Other Fixes
- ✅ Fixed Tailwind CSS warning about `z-[var(--z-*)]` pattern in MIGRATION.md
- ✅ Converted logger.browser.ts from unnecessary dynamic import to static

## Results

**Build Performance:**
- csv-parse now code-split into separate 49.44 kB chunk
- Main bundle reduced by ~50 kB (1,987.96 kB → 1,937.96 kB)
- Zero Vite build warnings ✨
- Build time: ~11-12 seconds

**Testing:**
- ✅ All 2,649 frontend tests passing
- ✅ All 10,378 backend tests passing
- ✅ Manual CSV upload testing confirmed working

**Key Achievement:** The frontend now uses only browser-native APIs - no Node.js polyfills needed! This is the ideal state for a Vite application.

## Test Plan

```bash
npm run build:app  # Zero warnings
cd src/app && npm test -- --run  # All tests pass
```

### DIFF
--- a/src/validators/redteam.ts
+++ b/src/validators/redteam.ts
@@ -1,6 +1,5 @@
 import dedent from 'dedent';
 import { z } from 'zod';
-import logger from '../logger';
 import {
   ALIASED_PLUGIN_MAPPINGS,
   ALIASED_PLUGINS,
@@ -333,7 +332,7 @@ export const RedteamConfigSchema = z
       const strategyLanguages = multilingualStrategy.config?.languages;
 
       if (Array.isArray(strategyLanguages) && strategyLanguages.length > 0) {
-        logger.debug(
+        console.debug(
           '[DEPRECATED] The "multilingual" strategy is deprecated. Use the top-level "language" config instead. See: https://www.promptfoo.dev/docs/red-team/configuration/#language',
         );
 


src/app/MIGRATION.md
```diff
@@ -761,7 +761,7 @@ We define z-index values as CSS custom properties in `index.css`:
 
 #### Usage in Components
 
-Use the `z-[var(--z-*)]` syntax in Tailwind classes:
+Use the `z-[var(--z-layer-name)]` syntax in Tailwind classes (where `layer-name` is the specific layer like `modal`, `dropdown`, etc.):
```
faizanminhas: nice

src/commands/list.ts
```diff
@@ -28,18 +28,24 @@ export function listCommand(program: Command) {
...
+              prompts.map(async (p) => (await sha256(p.raw)).slice(0, 6)),
+            );
+            return {
+              'eval id': evl.id,
+              description: description.slice(0, 100) + (description.length > 100 ? '...' : ''),
+              prompts: promptHashes.join(', ') || '',
+              vars: vars[evl.id]?.join(', ') || '',
+            };
+          }),
+      );
```
faizanminhas: wonder if this is any faster

JustinBeckwith: Fair question - the honest answer is ... probably not.  The tradeoff here is trying to avoid usage of `crypto` in favor of the subtlecrypto thats isomorphic.  Unfortunately the subtlecrypto method is async, so it is introducing some of the overhead around managing promises that isn't there for the sync method in node.  For the scale we're operating at it's probably not important, but it is decreasing the risk that we pull in all of the crypto polyfill again.

src/app/MIGRATION.md
JustinBeckwith: Heh, this was just freaking out the bundler in vite and causing a warning

src/models/eval.ts
```diff
@@ -1173,7 +1184,7 @@ export default class Eval {
 
     // Copy eval, results, and relationships within transaction for atomicity
     let copiedCount = 0;
-    db.transaction(() => {
+    db.transaction(async () => {
```
mldangelo: Does this get resolved?

src/models/evalResult.ts
```diff
@@ -291,7 +292,10 @@ export default class EvalResult {
     this.testIdx = opts.testIdx;
     this.testCase = opts.testCase;
     this.prompt = opts.prompt;
-    this.promptId = opts.promptId || hashPrompt(opts.prompt);
+    // promptId should be computed before calling the constructor
+    // Use the factory methods (createFromEvaluateResult, createManyFromEvaluateResult)
+    // which compute the hash asynchronously before construction
+    this.promptId = opts.promptId || '';
```
mldangelo: if you are sure opts.promptId exists, let's add an invariant to narrow the type above instead of || '';

src/util/createHash.ts
```diff
@@ -1,9 +1,30 @@
...
-export function sha256(str: string) {
-  return createHash('sha256').update(str).digest('hex');
+/**
+ * Compute SHA-256 hash of a string and return hex digest.
+ * Uses the Web Crypto API (SubtleCrypto).
+ *
+ * @param str - The string to hash
+ * @returns Promise resolving to the hex-encoded SHA-256 hash
+ */
+export async function sha256(str: string): Promise<string> {
```
mldangelo: I want to make sure that the hashing algorithm is stable and the hashes before are the same as after. Would be nice to create a test case specifically for this.

mldangelo (commented):
I appreciate the spirit, but I don't like this.  This app is served on localhost and the bundle size does not matter very much. I think there are a couple cleaner options than “make `sha256` async everywhere”:

1. **Don’t hash in the browser at all** where we can avoid it. If the only reason hashing shows up in frontend paths is stable IDs / display IDs, we can compute those in Node (server/CLI/DB layer) and keep the UI from touching hashing entirely. That keeps the bundle slim and avoids promise overhead in UI code.

2. **Avoid async-infecting the whole codebase** by splitting the API: `sha256Sync()` for Node-only code (crypto `createHash`) and `sha256Async()` for browser/isomorphic code (WebCrypto). Then only truly isomorphic surfaces pay the async cost, and the DB/models/CLI paths can stay sync.

This still achieves the goal of not pulling crypto polyfills into the frontend, but with a smaller blast radius.

JustinBeckwith:
@mldangelo @faizanminhas I took a very different path with this PR, and I think you were right.  Have another look!

src/validators/redteam.ts
```diff
@@ -332,13 +333,9 @@ export const RedteamConfigSchema = z
...
 
       if (Array.isArray(strategyLanguages) && strategyLanguages.length > 0) {
-        // Fire-and-forget async import to avoid circular dependency at module load time
-        (async () => {
-          const { default: logger } = await import('../logger');
-          logger.debug(
-            '[DEPRECATED] The "multilingual" strategy is deprecated. Use the top-level "language" config instead. See: https://www.promptfoo.dev/docs/red-team/configuration/#language',
-          );
-        })();
+        logger.debug(
```
mldangelo: I bet Claude can figure out a nice structural refactor for this

JustinBeckwith: I agree, but I think we should come back around for this one ... it's going to be lorge

PR #6871, opened 2025-12-29
fix(app): improve prompt/response readability in risk category drawer

typpo:
add clear labels, fix text overflow, preserve line breaks

image


faizanminhas:
fyi here's how it looks with a longer input:

Screenshot 2025-12-29 at 12 58 05 PM

faizanminhas:
also wdyt of making the response colour green / red based on if it was flagged or not? better than the purple it is now

typpo:
image

PR #6873, opened 2025-12-29
feat(eval): add JSON diff visualization for failed equals assertions

### LINKED ISSUE #2382: Question: Compare JSON values
My prompts ask the IA for JSON objects. So what I want to compare with the values of them if they are correct.

The problem is that the “diff” that promptfoo offers (or at least what I find) is that it compares the whole object and the diff is the object itself.
When it shows the test as “fail” it shows me two objects, so it is difficult to see in which value is the error.

Isn't there a more specific diff? Whether it is JSON or not, promtfoo does not show the specific part where the difference is, it shows everything itself.

Maybe there is a functionality that I am not implementing well? or is it that promptfoo does not allow to do this?

Thank you in advance!

mldangelo:
## Summary

When a JSON `equals` assertion fails, the Evaluation tab now explains the mismatch with a path-based summary and an expandable unified JSON diff.

- Adds `JsonDiffView` for failed object/array `equals` assertions.
- Integrates it in the cell-detail Evaluation tab using the actual provider output.
- Handles `null`, arrays, special object keys, large inputs, and unsupported serialized structures without misleading output.

## Design decisions

1. **Match `equals` semantics exactly.** The view parses only a complete JSON output. It does not extract a JSON fragment from surrounding prose, because the full output is what the equality assertion graded.
2. **Make paths trustworthy.** Plain identifier keys remain readable (`user.name`), while keys with punctuation or numeric-looking names are quoted (`user["first.name"]`, `["0"]`) so they cannot be confused with nesting or array indices.
3. **Keep interaction cheap.** Assertion rows and candidate parsing are memoized; pretty JSON is serialized once; the full line diff is memoized and computed only when expanded; inputs beyond the size guard receive a clear fallback.
4. **Keep fallback states honest.** Large serializable values and values that cannot be serialized have distinct messages.

## Review follow-up

- Treats `null` as a JSON value rather than an added/removed field and preserves valid JSON `null` as a diffable actual output.
- Covers array/object type mismatches, quoted paths, strict output parsing, escaped display strings, and the two fallback states.
- Gives the expected and actual copy buttons distinct accessible names.

## Test plan

- [x] Merged current `origin/main` into this branch.
- [x] `npm run l`
- [x] `npm run tsc`
- [x] `npm run test:app -- src/components/JsonDiffView.test.tsx src/utils/jsonDiff.test.ts src/pages/eval/components/EvaluationPanel.test.tsx src/pages/eval/components/EvalOutputPromptDialog.test.tsx --run` (`117` passed)
- [x] `npm run test:app -- --run` (`4,588` passed)
- [x] `npm run build`
- [x] Isolated live eval using `PROMPTFOO_CONFIG_DIR=/tmp/pr6873-state` and `PROMPTFOO_CACHE_PATH=/tmp/pr6873-cache`; inspected the exported failed object `equals` result containing a `null` change and a dotted key. All generated state artifacts were mode `600`.

`npm run f` formatted the changed TypeScript through Biome, then its empty Prettier stage exits nonzero when a PR has no changed Markdown/CSS/YAML files; an explicit Biome formatting check completed cleanly afterward.

No database migrations are part of this PR.

Closes #2382


ghost:
**🔄 Running Tusk Tester (a0c7f370086c5f572058ef10ebd5f33bcb9800c0) [View output ↗](https://app.usetusk.ai/app/testing-commit-check/c032b85b-8c1e-4c55-92b7-a05f85b04099?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

**Running generated tests...**
🟩🟩🟩🟩🟩🟩🟩🟩🟩⬜⬜⬜⬜⬜⬜⬜⬜⬜⬜⬜ 49%

---

<details><summary>Symbols Not Exported</summary>

The following symbols were skipped because they are not exported and therefore cannot be tested. Export these symbols if you want them to be tested:

- `src/app/src/components/JsonDiffView.tsx` - (`DiffRow`, `UnifiedDiff`)

</details>
<p>

| e2748e8b52e588210135c83f3ecaca435984ffb8 | ⏩ Skipped due to new commit on branch | [Output](https://app.usetusk.ai/app/testing-commit-check/59b5cc81-87cb-48f0-a58d-d4799e5f34b2?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 29, 2025 9:00PM |
| b70f809394bf0411a72c0985a931c17d93ee1f0d | ⏩ Skipped due to new commit on branch | [Output](https://app.usetusk.ai/app/testing-commit-check/bd2e9669-51bd-4171-9b66-90bc430863bc?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 29, 2025 9:07PM |
| caba8c1b1716e5a37ce371a58b6b817ea9ed623f | ⏩ Skipped due to new commit on branch | [Output](https://app.usetusk.ai/app/testing-commit-check/c72acf8d-f520-48fb-a3cf-23ad9bd091b1?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 29, 2025 9:11PM |
| a0c7f370086c5f572058ef10ebd5f33bcb9800c0 | 🔄 Running Tusk Tester | [Output](https://app.usetusk.ai/app/testing-commit-check/c032b85b-8c1e-4c55-92b7-a05f85b04099?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Dec 29, 2025 9:11PM |

</p>
</details>

[View output in GitHub ↗](https://github.com/promptfoo/promptfoo/pull/6873/checks?check_run_id=59113684207)

src/app/src/components/JsonDiffView.tsx
```diff
@@ -0,0 +1,224 @@
...
+ * Renders a path-based summary of JSON differences with expandable full diff
+ */
+export function JsonDiffView({ expected, actual, className }: JsonDiffViewProps) {
+  const [showFullDiff, setShowFullDiff] = useState(false);
+  const [showAllDiffs, setShowAllDiffs] = useState(false);
+
+  const { diffs, tooLarge } = useMemo(() => {
+    try {
+      // Size guard for very large objects
+      const expectedStr = JSON.stringify(expected);
```
will-holley: `JSON.stringify` is also called in `UnifiedDiff`; for large objects this can be an expensive operation, so consider instead memoizing the stringified values in `JsonDiffView`'s scope and passing them down as props.

src/app/src/components/JsonDiffView.tsx
```diff
@@ -0,0 +1,224 @@
...
+        </>
+      )}
+    </div>
+  );
+}
+
+/**
+ * Renders a unified diff view showing the full JSON with highlighted changes
+ */
+function UnifiedDiff({ expected, actual }: { expected: unknown; actual: unknown }) {
```
will-holley: Should this be wrapped in `React.memo`?

src/app/src/utils/jsonDiff.ts
```diff
@@ -0,0 +1,154 @@
...
+        diffs.push({
+          path: keyPath,
+          expected: expectedObj[key],
+          actual: undefined,
+          type: 'removed',
+        });
+      }
+    }
+    return diffs;
+  }
```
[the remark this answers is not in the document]
will-holley: Should type equality between `expected` and `actual` be verified prior to these two conditional statements?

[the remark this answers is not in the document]
mldangelo-oai: Resolved on the pushed tip: array/object mismatches return a root changed diff, with explicit regression coverage.

src/app/src/components/JsonDiffView.tsx
mldangelo-oai: Resolved on the pushed tip: UnifiedDiff is memoized and its line diff is computed with useMemo only after the full view is expanded.

src/app/src/components/JsonDiffView.tsx
mldangelo-oai: Resolved on the pushed tip: the view serializes expected and actual JSON once and passes those strings into the expandable unified diff.

src/app/src/pages/eval/components/EvaluationPanel.tsx
```diff
@@ -218,6 +226,9 @@ function AssertionResults({ gradingResults }: { gradingResults?: GradingResult[]
     return null;
   }
 
+  // Parse actual output once for JSON diff comparisons
+  const parsedActualOutput = tryParseJson(actualOutput);
+
```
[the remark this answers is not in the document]
mldangelo-oai: Resolved on the pushed tip: rows and parsing are memoized, and actual output is parsed only when a failed JSON equals assertion can display a diff.

src/app/src/utils/jsonDiff.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  for (const key of allKeys) {
+    const keyPath = path ? `${path}.${key}` : key;
+    if (!(key in expected)) {
+      diffs.push(createDiff(keyPath, undefined, actual[key], 'added'));
+    } else if (key in actual) {
+      diffs.push(...computeJsonDiff(expected[key], actual[key], keyPath));
+    } else {
+      diffs.push(createDiff(keyPath, expected[key], undefined, 'removed'));
+    }
```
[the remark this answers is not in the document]
mldangelo-oai: Resolved on the pushed tip: non-identifier object keys now use quoted bracket paths, with tests for dotted and numeric-looking keys.

src/app/src/components/JsonDiffView.tsx
```diff
@@ -0,0 +1,201 @@
...
+  return (
+    <div
+      className={cn(
+        'mt-3 rounded-lg border border-amber-200 dark:border-amber-800/50 bg-amber-50/50 dark:bg-amber-950/20 p-3',
+        className,
+      )}
+    >
+      {tooLarge ? (
+        <div className="text-sm text-muted-foreground">
+          Objects too large for diff view. Compare the expected and actual values manually.
```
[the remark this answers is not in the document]
mldangelo-oai: Resolved on the pushed tip: large serialized values and unsupported serialization now render distinct, accurate fallback messages.

src/app/src/utils/jsonDiff.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (expected === null || expected === undefined) {
+    return actual === null || actual === undefined
+      ? []
+      : [createDiff(path, expected, actual, 'added')];
+  }
+
+  if (actual === null || actual === undefined) {
+    return [createDiff(path, expected, actual, 'removed')];
+  }
+
```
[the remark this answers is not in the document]
mldangelo-oai: Resolved on the pushed tip: null is treated as a real changed JSON value, and valid JSON null output is kept distinct from invalid parse results.

mldangelo-oai:
@codex review

mldangelo-oai:
Updated this branch after merging current `main`.

What changed:
- Render nested object/array assertion templates before equality checks, so failed JSON `equals` assertions store and compare the values users actually configured after vars are applied.
- Preserve explicit `renderedAssertionValue: null` in the eval UI and comment context instead of treating it as missing metadata.
- Keep the JSON diff UI on the current design-system primitives for the expandable summary control.

Verification:
- `npm run test:app -- src/components/JsonDiffView.test.tsx src/utils/jsonDiff.test.ts src/pages/eval/components/EvaluationPanel.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx src/pages/eval/components/EvalOutputPromptDialog.test.tsx --run`
- `npx vitest run test/assertions/runAssertion.test.ts --sequence.shuffle=false`
- `npx vitest run test/assertions/runAssertions.test.ts test/assertions/assertionsResult.test.ts test/evaluator/assertions.test.ts --sequence.shuffle=false`
- `npm run tsc`
- `npm run l`
- `npx @biomejs/biome check --write <changed files>`
- `npm run build`
- CLI e2e with `npm run local -- eval -c /tmp/promptfoo-pr6873-jsondiff-e2e.yaml --no-cache --no-share -o /tmp/promptfoo-pr6873-jsondiff-e2e.json`; expected exit 100 for the failing equality assertion, and the export now contains `metadata.renderedAssertionValue = {"name":"Jane"}` rather than the raw `{{ expected_name }}` template.

No unresolved review threads remain on the PR. CI has started on `af601a15b`.

mldangelo-oai:
Merged current `main` and pushed a small follow-up quality fix in `8a759e627f636e03ac3a5cae4c03a0c83b62de7d`.

What changed:
- Centralized nested assertion string rendering so object values now resolve `file://` references consistently with array values before equality checks.
- Added regression coverage for nested object `file://` assertion values and preserved `renderedAssertionValue` metadata for the JSON diff UI.
- Removed a stale ESLint-only `as any` suppression in the app JSON diff tests.

Validation run:
- `npx vitest --run test/assertions/runAssertion.test.ts` (167 passed)
- `npm run test:app -- src/components/JsonDiffView.test.tsx src/utils/jsonDiff.test.ts src/pages/eval/components/EvaluationPanel.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx src/pages/eval/components/EvalOutputPromptDialog.test.tsx --run` (252 passed)
- `npm run l` (passed; existing `runAssertion` complexity warning only)
- `npm run tsc`
- `npm run build`
- Isolated CLI E2E with `PROMPTFOO_CONFIG_DIR=/tmp/pr6873-state` and `PROMPTFOO_CACHE_PATH=/tmp/pr6873-cache`: local echo eval intentionally failed the JSON `equals` assertion, exported `/tmp/pr6873-e2e/output.json`, and confirmed `componentResults[0].metadata.renderedAssertionValue` contains the rendered nested object `{ user: { name: "Jane", roles: ["admin"] } }`.

Note: `npm run f` hit the repo changed-files formatter edge case where Biome checks the changed TS files, then Prettier receives no docs/style files and exits with “No parser and no file path given.” No formatter changes were pending after Biome.

src/types/index.ts
```diff
@@ -545,7 +545,7 @@ export interface GradingResult {
     // Raw textual responses returned by one or more LLM grader phases
     graderOutputs?: Record<string, string>;
     // Rendered assertion value with substituted variables (for display in UI)
-    renderedAssertionValue?: string;
+    renderedAssertionValue?: unknown;
```
jameshiester-oai, line 548: **[P1] Preserve the existing public metadata contract**

`GradingResult` is re-exported from the package, and the base contract makes this field `string | undefined`. This head both widens it to `unknown` and writes arrays/objects, so a previously valid consumer such as `const value: string | undefined = result.metadata?.renderedAssertionValue` now fails with TS2322; JavaScript callers that use `.trim()` can fail at runtime for the new structured values. Please preserve the string field and add a separate structured/discriminated field (or otherwise provide an explicit compatibility path) instead of silently changing the value kind.

src/app/src/utils/jsonDiff.ts
```diff
@@ -0,0 +1,229 @@
...
+  if (Array.isArray(expected) || Array.isArray(actual)) {
+    return [createDiff(path, expected, actual, 'changed')];
+  }
+
+  if (isRecordValue(expected) && isRecordValue(actual)) {
+    return computeObjectDiff(expected, actual, path);
+  }
+
+  return expected === actual ? [] : [createDiff(path, expected, actual, 'changed')];
```
jameshiester-oai, line 127: **[P2] Match `equals` semantics for signed zero**

`equals` uses `util.isDeepStrictEqual`, which distinguishes `0` from `-0`, but this `===` comparison does not. At this head, expected `{ n: 0 }` with output `{"n":-0}` produces `pass: false` and `computeJsonDiff(...) === []`, so `JsonDiffView` renders nothing for the failed assertion. Use equality/display handling that matches the grader (including preserving the negative-zero distinction in the row/full view) and add this valid-JSON regression case.

src/app/src/utils/jsonDiff.ts
```diff
@@ -0,0 +1,229 @@
...
+      return str;
+    } catch {
+      return '[unserializable value]';
+    }
+  }
+  return String(value);
+}
+
+function stringifyJson(value: unknown): string {
+  return JSON.stringify(value, null, 2) ?? String(value);
```
jameshiester-oai, line 212: **[P2] Canonicalize object keys before the unified diff**

Pretty-printing in insertion order makes the expanded view report ordering as data changes even though object key order is irrelevant to `equals`. For expected `{a:1,b:2,c:3}` and actual `{b:2,a:1,c:4}`, the structural summary correctly reports only `c`, while the unified diff also marks unchanged `a` as removed/added. Recursively canonicalize object keys for this semantic diff (arrays must retain order) and cover a reordered-object case.

src/app/src/components/JsonDiffView.tsx
```diff
@@ -0,0 +1,233 @@
...
+ */
+const UnifiedDiff = memo(function UnifiedDiff({
+  expectedJson,
+  actualJson,
+}: {
+  expectedJson: string;
+  actualJson: string;
+}) {
+  const unifiedLines = useMemo(
+    () => buildUnifiedJsonTextDiff(expectedJson, actualJson),
```
jameshiester-oai, line 208: **[P2] Bound worst-case line-diff work**

The character guard still permits thousands of completely different lines, and `diffLines` runs synchronously on the browser main thread. A direct head probe with two fully changed 4,000-element arrays (30,892 and 32,002 pretty-JSON characters, both below the limit) took about 1.2 seconds just to build the unified diff. Add a line/edit-complexity guard, compute off-thread, or use a bounded representation so expanding a valid sub-limit result cannot freeze the dialog.

src/app/src/pages/eval/components/EvaluationPanel.tsx
```diff
@@ -255,16 +273,33 @@ function formatGraderOutputLabel(key: string, totalOutputs: number): string {
...
 
+  const assertionRows = useMemo(
+    () => (gradingResults ? buildAssertionRows(gradingResults) : []),
+    [gradingResults],
+  );
+  const shouldParseActualOutput = assertionRows.some(
+    ({ result }) => !result.pass && getJsonDiffExpectedValue(result) !== undefined,
+  );
+  const parsedActualOutput = useMemo(
+    () => (shouldParseActualOutput ? tryParseJson(actualOutput) : undefined),
```
jameshiester-oai, line 295: **[P2] Reject oversized output before parsing it**

This parses the entire provider-controlled output before `JsonDiffView` checks its 50,000-character limit. A multi-megabyte valid JSON response therefore allocates the complete object and is serialized again in the child before the UI decides it is too large, defeating the guard and potentially blocking or exhausting the browser. Check the source string length before `JSON.parse` and carry a too-large state to the fallback.

src/app/src/utils/jsonDiff.ts
```diff
@@ -0,0 +1,229 @@
...
+ */
+export function formatDiffValue(value: unknown): string {
+  if (value === undefined) {
+    return 'undefined';
+  }
+  if (value === null) {
+    return 'null';
+  }
+  if (typeof value === 'string') {
+    return JSON.stringify(value) ?? '""';
```
jameshiester-oai, line 191: **[P3] Cap primitive strings in summary rows**

Only object values are truncated below. A changed 20,000-character string returns a 20,002-character display value, so the ten-row “summary” can immediately place hundreds of kilobytes of text in the DOM before full-diff expansion. Apply the same bounded display policy to strings (while keeping copy/full data intact) and add a long-string test.

jameshiester-oai (commented):
Reviewed base `87d04046debc840006659f39b1e6481479d260ee` at head `8a759e627f636e03ac3a5cae4c03a0c83b62de7d`.

**Scope:** fresh full review; no authenticated completion marker existed for this head. I read the full base...head diff, surrounding assertion/UI code, current and historical review threads, and exact-head CI.

**Verdict:** the feature works on the common path, but the current head introduces one public API compatibility break and several reproducible correctness, performance, and result-artifact issues. Exact-head CI is green: 41/41 checks completed successfully.

## Verified findings

### P1

- `GradingResult.metadata.renderedAssertionValue` changes from a string contract to `unknown`, while the runner now writes arrays/objects. Existing TypeScript reads stop compiling and existing JavaScript string operations can fail at runtime. See the inline comment on `src/types/index.ts`.

### P2

- Signed zero violates the stated exact-semantics goal: `equals` rejects `{n: 0}` versus `{"n": -0}`, but the new structural diff returns no differences and hides itself. See the inline comment on `jsonDiff.ts`.
- The expanded unified diff treats insertion-order changes as value changes. A case with one semantic field change marked an unchanged reordered field as removed and added. See the inline comment on `jsonDiff.ts`.
- The 50,000-character limit does not bound line-diff complexity. A 4,000-element completely changed array (30,892/32,002 serialized characters) blocked this machine for about 1.2 seconds when building the full diff. See the inline comment on `JsonDiffView.tsx`.
- Large provider output is fully parsed before the component applies its size guard, so oversized valid JSON still incurs the allocation and an additional serialization before the fallback appears. See the inline comment on `EvaluationPanel.tsx`.
- The broadened metadata write persists resolved object-valued fixtures for every handler and for passing assertions, although the diff consumes them only for failed `equals`. A passing `is-json` assertion backed by `file:///tmp/pr6873-review/schema.json` exported the complete schema under `renderedAssertionValue`; before this change the result retained only the file reference. This can unnecessarily place fixture contents in DB/export/API artifacts and multiply payload size. I did not add a second inline thread at `src/assertions/index.ts:688` because current feedback already targets that storage block for a distinct handler-consistency defect.

### P3

- Summary rows cap serialized objects but not primitive strings; a changed 20,000-character string renders 20,002 characters on each side before the user expands anything. See the inline comment on `jsonDiff.ts`.

## Existing unresolved feedback (not duplicated)

- [Transformed assertions are diffed against raw provider output](https://github.com/promptfoo/promptfoo/pull/6873#discussion_r3294385380).
- [Rendered metadata can claim values that some handlers did not evaluate](https://github.com/promptfoo/promptfoo/pull/6873#discussion_r3294385381).
- [Nested literal `file://` strings are reinterpreted as file references](https://github.com/promptfoo/promptfoo/pull/6873#discussion_r3294385383).

## Tests and QA

- `npm run test:app -- src/components/JsonDiffView.test.tsx src/utils/jsonDiff.test.ts src/pages/eval/components/EvaluationPanel.test.tsx src/pages/eval/components/EvalOutputPromptDialog.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx --run` — 5 files, 252 tests passed.
- `npx vitest run test/assertions/runAssertion.test.ts --sequence.shuffle=false` — 1 file, 167 tests passed.
- `npm run tsc` — passed.
- `npm run build` — passed, including app production build.
- `git diff --check 87d04046debc840006659f39b1e6481479d260ee...HEAD` — passed.
- `PROMPTFOO_DISABLE_REMOTE_GENERATION=true npm run local -- eval -c /tmp/pr6873-review/promptfooconfig.yaml --no-cache -o /tmp/pr6873-review/results.json` — expected exit 100; 0 passed, 1 failed, 0 errors; exported result contained the rendered nested expected object and raw provider output.
- The same local workflow with a passing file-backed `is-json` assertion exited 0 (1 passed, 0 errors) and reproduced the resolved-fixture metadata persistence.
- Direct head probes reproduced the signed-zero mismatch, key-order false positives, 4,000-line latency, long-string rendering, and downstream TypeScript compile break.

- Centralize parse/serialization/size acceptance in one bounded helper so `EvaluationPanel` cannot do work that `JsonDiffView` later rejects.
- Canonicalize object keys for the semantic full diff and add regression cases for signed zero, reordered keys, worst-case line counts, long primitive strings, and passing file-backed assertions.
- No separate simplification opportunity was material beyond consolidating those guards and representations.

## Limitations

- Browser QA was not run: port 3000 is an SSH listener and no Promptfoo server is listening on 15500; repository guidance requires confirmation before starting the development server. Component interaction tests covered expansion and copy controls, but responsive/dark-mode behavior was not manually inspected.
- The `/pr-review-toolkit:review-pr` command contract was present locally but not exposed as a callable Codex skill endpoint. Its correctness, tests, comments/docs, error-handling, type-design, and simplification lenses were executed through the orchestrated review and three independent read-only specialist passes.

PR #6875, opened 2025-12-29
fix(app): remove underline from navbar dropdown items

typpo:
dropdown menu links were getting underlined on hover from global css

now it looks like this

image


faizanminhas:
tbh I was torn on this, I think the underlines make it clearer / better for accessibility that they're links, but am happy to not have underlines

PR #6897, opened 2025-12-30
fix(app): pointer cursor on eval setup steps

typpo:
Setup step buttons in the eval creator sidebar now show a pointer cursor on hover.

### DIFF
--- a/src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.tsx
+++ b/src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.tsx
@@ -200,7 +200,7 @@ const EvaluateTestSuiteCreator = () => {
                   <button
                     onClick={() => setActiveStep(1)}
                     className={cn(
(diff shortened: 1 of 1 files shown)


src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.tsx
```diff
@@ -200,7 +200,7 @@ const EvaluateTestSuiteCreator = () => {
                   <button
                     onClick={() => setActiveStep(1)}
                     className={cn(
-                      'w-full text-left px-3 py-3 rounded-lg transition-all',
+                      'w-full text-left px-3 py-3 rounded-lg transition-all cursor-pointer',
                       'hover:bg-muted/50',
                       activeStep === 1 && 'ring-2 ring-primary',
                       normalizedProviders.length > 0
```
faizanminhas, line 206: I think the real fix is to switch to using the `components/ui/button.tsx` button instead?

PR #6903, opened 2026-01-01
docs: fix broken links in Google Provider Documentation

Sadia-Shahzad:
This PR fixes **four** broken links in the **Google provider documentation** that was leading to invalid pages.

Those links have been updated to point to the correct destination to improve the documentation experience for users.


typpo:
I think these links are OK?  https://www.promptfoo.dev/docs/providers/vertex/

Sadia-Shahzad:
Thanks for pointing this out.
I misunderstood this and checked the links only from the repository instead of the live site.
I’ll close this PR and verify against the website moving forward.

PR #6907, opened 2026-01-02
fix(app): add biome-ignore for useExhaustiveDependencies and fix one unused dependency

JustinBeckwith:
Very much looking for opinions here :) I just fixed a series of hard to debug bugs in a different PR that would have been captured by this lint rule.  I've used this in the past with eslint, and usually it was right 😆   Very open to feedback though!

## Summary

This PR addresses 93 `useExhaustiveDependencies` lint errors by:
1. Adding `biome-ignore` comments for 92 intentional cases
2. Fixing 1 genuinely unused dependency

### Fixed: TestCaseDialog.tsx
- Removed `currentTurn` from `useMemo` dependency array (it was never used in the computation)
- Renamed unused parameter to `_currentTurn` to indicate it's intentionally unused

### Added ignores for 92 intentional cases

After comprehensive analysis, the remaining 92 cases fall into these valid React patterns:

1. **Trigger dependencies** - Dependencies that cause re-execution but aren't read in the hook body
   ```tsx
   // Reset to page 1 when filters change
   useEffect(() => {
     setPagination({ pageIndex: 0 });
   }, [filterA, filterB, filterC]); // None are "used" but all trigger the reset
   ```

2. **Effect cleanup triggers** - Dependencies that cause effects to re-run for proper cleanup
   ```tsx
   useEffect(() => {
     // Reset state when evalId changes
     setError(null);
   }, [evalId]); // evalId not "used" but triggers the reset
   ```

3. **Derived value dependencies** - Computed values in deps along with their source values
   ```tsx
   const isUnlimited = max === value;
   useEffect(() => {
     setDisplay(isUnlimited ? 'Unlimited' : value);
   }, [value, max, isUnlimited]); // Redundant but safe
   ```

These are all standard React patterns. Biome's `useExhaustiveDependencies` rule is overly strict for these cases.

## Testing

- ✅ Lint passes
- ✅ Manual code review of all 93 cases
- ✅ Only 1 case had a genuinely unused dependency (now fixed)

### DIFF
--- a/src/app/src/pages/redteam/setup/components/TestCaseDialog.tsx
+++ b/src/app/src/pages/redteam/setup/components/TestCaseDialog.tsx
@@ -75,7 +75,6 @@ export const TestCaseDialog: React.FC<TestCaseDialogProps> = ({
   isRunningTest = false,
   onRegenerate,
   onContinue,
-  currentTurn: _currentTurn,
   maxTurns,
   availablePlugins,
   allowPluginChange = false,


src/app/src/pages/redteam/setup/components/TestCaseDialog.tsx
```diff
@@ -75,7 +75,7 @@ export const TestCaseDialog: React.FC<TestCaseDialogProps> = ({
   isRunningTest = false,
   onRegenerate,
   onContinue,
-  currentTurn,
+  currentTurn: _currentTurn,
```
mldangelo:
```suggestion
```

PR #6923, opened 2026-01-04
fix(blobs): store audio blobs locally and pass evalId for access control

mldangelo:
## Summary

- Remove cloud upload logic from audio blob extraction (`maybeStore`) - audio now stores locally like videos instead of trying GCS first
- Add `normalizeAudioMimeType()` to fix MIME types (e.g., `wav` → `audio/wav`)
- Pass `evalId` to `extractAndStoreBinaryData` for proper blob reference recording and access control
- Add `evalId` to `RunEvalOptions` interface

## Background

Audio blob storage was failing because:

1. **Cloud upload path**: When logged into Promptfoo Cloud, audio was being uploaded to GCS via `uploadBlobRemote()`, but the local blob route (`/api/blobs/:hash`) couldn't serve GCS blobs
2. **Missing evalId**: While `evalId` wasn't strictly necessary (since `EvalResult.createFromEvaluateResult` already calls `extractAndStoreBinaryData` with `evalId` when saving results), passing it during evaluation provides earlier reference recording
3. **MIME type issues**: Providers returning just `wav` instead of `audio/wav` caused incorrect MIME type storage

### Audio Blob Storage (`src/blobs/extractor.ts`)
- Removed `shouldAttemptRemoteBlobUpload()` and `uploadBlobRemote()` imports and usage
- Added `normalizeAudioMimeType()` function to convert short format names to proper MIME types
- Updated `maybeStore()` to store locally only (matching video behavior)

### Evaluator (`src/evaluator.ts`)
- Added `evalId` to the `runEval` function destructuring
- Pass `evalId` to `extractAndStoreBinaryData` call
- Added `evalId: this.evalRecord.id` when building `runEvalOptions`

### Types (`src/types/index.ts`)
- Added `evalId?: string` to `RunEvalOptions` interface

## Test plan

- [x] Run blob tests: `npx vitest run test/blobs/`
- [x] Run evaluator tests: `npx vitest run test/evaluator.test.ts`
- [x] Lint and format: `npm run l && npm run f`

### DIFF
--- a/src/blobs/extractor.ts
+++ b/src/blobs/extractor.ts
@@ -55,8 +55,10 @@ export function normalizeAudioMimeType(format: string | undefined): string {
 
   const trimmedFormat = format.trim();
 
-  // Already a proper audio MIME type
-  if (/^audio\/[a-z0-9.+-]+$/i.test(trimmedFormat)) {
+  // Already a proper audio MIME type - validate strictly to prevent MIME injection
+  // Only allow: audio/subtype where subtype is alphanumeric with optional dash/underscore/plus
+  // Periods are NOT allowed to prevent attacks like "audio/wav.html" being interpreted as HTML
+  if (/^audio\/[a-z0-9_+-]+$/i.test(trimmedFormat)) {
     return trimmedFormat;
   }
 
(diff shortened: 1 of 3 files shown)


src/blobs/extractor.ts
will-holley, line 113: Should this function name be updated?

mldangelo:
Good question! The `maybeStore` name is still accurate because it 'maybe' stores the blob based on several conditions:

1. Returns null if binary parsing fails
2. Returns null if blob doesn't meet size requirements (`shouldExternalize` checks `BLOB_MIN_SIZE` to `BLOB_MAX_SIZE`)  
3. Returns null if blob storage is disabled via `PROMPTFOO_INLINE_MEDIA`

The change was removing the cloud upload path, but the conditional storage behavior remains the same. The function still 'maybe stores' locally depending on these criteria.

If you'd prefer a different name like `storeIfEligible` or `tryStoreBlob`, happy to update it!

PR #6925, opened 2026-01-04
feat(assertions): add secure video-rubric evaluation

mldangelo:
## Summary

- Adds the `video-rubric` model-graded assertion for videos returned by built-in providers.
- Reuses shared image-input helpers in the existing Bedrock Nova Reel and Luma Ray providers.
- Adds a focused video-rubric example and concise reference/guide documentation.

## Design and security

- Grades only Promptfoo-managed `blobRef` and `storageRef` media. External response URLs are rejected instead of being fetched server-side.
- Validates judge output fail closed: `pass` must be boolean-like and `score` must be finite and between `0.0` and `1.0`.
- Treats Gemini's inline cap as a total request budget, accounting for base64 expansion and rubric text.
- Uses current Google AI Studio `google:gemini-3.5-flash` as the default video judge, aligned with Google's [video understanding documentation](https://ai.google.dev/gemini-api/docs/video-understanding).
- Keeps assertion/test inputs immutable while preparing result display metadata.
- Removes stale duplicate and unsupported example material after syncing with current `main`; the maintained Bedrock video examples already exist under `examples/amazon-bedrock/video/`.

## Validation

- `npm run f`
- `npm run l`
- `npm run tsc`
- `npx vitest test/assertions/videoRubric.test.ts test/matchers/video.test.ts test/util/video.test.ts test/providers/bedrock/video-utils.test.ts test/providers/bedrock/nova-reel.test.ts test/providers/bedrock/luma-ray.test.ts test/providers/google/ai.studio.defaults.test.ts --run` (`124` passing)
- `npm test -- --run` (`18,242` passing, `10` skipped)
- `npm run build`
- Isolated local CLI pass using a stored video reference and deterministic local judge.
- Isolated local CLI negative run confirming an external video URL fails before invoking the judge; disposable state/database/log files were mode `600`.

`npm run f` and `npm run l` report existing nonblocking complexity warnings in `src/assertions/index.ts` and `src/providers/google/ai.studio.ts`; neither command modifies or fails on them.


mldangelo-oai:
@codex review

mldangelo-oai:
Reviewed, synced with current `main`, hardened, and pushed `70426d0b1`.

What changed:
- Prevented test vars named `rubric` from shadowing the internal video-rubric placeholder in the grading prompt and provider context.
- Preserved `blobRef.mimeType` as a fallback when blob metadata omits MIME type.
- Added focused regressions for both cases.
- Added the missing Docusaurus front-matter title for the new video-rubric docs page.

Validation:
- `npx vitest --run test/assertions/videoRubric.test.ts test/matchers/video.test.ts test/util/video.test.ts test/providers/bedrock/video-utils.test.ts test/providers/google/ai.studio.defaults.test.ts`
- `npx vitest --run test/providers/google/video.test.ts test/providers/google/ai.studio.test.ts test/providers/google/ai.studio.defaults.test.ts test/providers/video/utils.test.ts test/providers/bedrock/nova-reel.test.ts test/providers/bedrock/video-utils.test.ts test/providers/bedrock/luma-ray.test.ts test/providers/azure/video.test.ts test/providers/openai/video.test.ts test/providers/xai/video.test.ts`
- `npm run f`
- `npm run l` (passes; existing complexity warnings in `runAssertion` and `callGemini` remain warnings)
- `npm run tsc`
- `npm run build`
- Offline E2E: `npm run local -- eval -c /tmp/pr6925-e2e/promptfooconfig.yaml --no-cache -o /tmp/pr6925-e2e/output.json` with isolated config/cache/media paths. Verified exported result passed with inline `video/mp4` media and the intended rendered rubric despite a `rubric` test var.

CI was green on the pre-sync head; new checks should start from the pushed head.

src/matchers/video.ts
```diff
@@ -0,0 +1,257 @@
...
+): GradingResult {
+  return {
+    ...graderFail(reason, tokenUsage),
+    assertion,
+  };
+}
+
+function parseVideoRubricResponse(resp: ProviderResponse): RawVideoRubricResult | undefined {
+  if (typeof resp.output === 'string') {
+    const jsonObjects = extractJsonObjects(resp.output);
```
jameshiester-oai, line 54: [P2] Reject malformed judge output instead of repairing it

`extractJsonObjects` appends missing braces and parses with YAML semantics. At this head, the complete matcher accepts `{"pass":true,"score":1` (and YAML-like `{pass: true, score: 1}`) and returns `pass: true, score: 1`, despite the documented fail-closed contract. A truncated or non-JSON grader response can therefore become a false-positive quality-gate result. Parse exactly one complete JSON object—optionally after removing one complete Markdown fence—and reject repaired, YAML-like, multiple-object, or trailing content; add the malformed cases as regressions.

src/matchers/video.ts
```diff
@@ -0,0 +1,257 @@
...
+      'Cannot grade video without grading config. Specify --grader option or grading config.',
+    );
+  }
+
+  let resolvedVideo;
+  try {
+    resolvedVideo = await resolveVideoBytes(video);
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err);
+    return failure(`Failed to resolve video: ${message}`, assertion);
```
jameshiester-oai, line 125: [P2] Keep raw resolver errors out of assertion reasons

This copies arbitrary blob/media-provider text into a persisted grading reason. A resolver probe that threw `download failed for https://media.example/video?sig=probe-secret` returned that full URL in the result; the reason is later stored, exported, displayed, and copyable. This is distinct from the already-sanitized grading-provider error path and violates the assertion rule against raw secrets/provider payloads. Log the original error through sanitized object context and return a stable generic resolution failure, with a secret-shaped regression test.

jameshiester-oai (commented):
Reviewed base `c54a30668ad8319d76c20ae96e6680ad6c51a2c6` through head `b577b20753325ea1b9a105cac05c90160f1243b6`.

**Scope:** full review of the current PR; no prior authenticated completion marker exists for this head.

**Verdict:** six P2 findings. The focused suites, typecheck, application build, docs build, and deterministic CLI workflow pass. GitHub is red only because three Codecov v11.2.7 uploads fail GPG verification after their site, UI, and Node 20 test suites pass. The PR is currently conflicting with `main`.

## Verified findings

### P2

- String judge output is parsed with a repair-capable YAML extractor, so truncated/non-JSON output can become a passing grade.
- Raw blob/media resolver exceptions are copied into persisted assertion reasons and can disclose signed URLs or provider details.
- Oversized managed videos are fully loaded into memory before the inline-size rejection.
- The 20 MB guard measures the inner multimodal prompt rather than the final Google request body.
- Threshold failures can preserve a positive judge explanation, producing contradictory exported evidence.
- The architecture baseline is raised across five layer edges by 13 imports, bypassing the dependency-count ratchet for the new implementation.

No P0, P1, or additional P3 findings were validated.

## Existing feedback

All 11 prior inline threads are resolved. I rechecked the current head and confirmed the earlier fixes for inversion/grader-error propagation, default-provider selection, malformed-output disclosure, threshold validation, docs/frontmatter, rubric variable precedence, and eval-creator registration. None of the findings above duplicates those threads.

## Simplification and coverage opportunities

- Use one strict structured-grader parser and one provider-aware request builder instead of separately repairing output and estimating a partial request size.
- Preserve media size metadata through `ProviderResponse.video` and expose a bounded/stat-based storage read so the matcher can reject before allocation.
- Add regression cases for truncated/YAML-like judge text, positive reasons below threshold, secret-shaped resolver errors, pre-read size rejection, and exact final-request boundaries.
- I could not run a live adversarial judge probe; future coverage should include text/audio instructions embedded in a video and verify the judge treats them as evidence rather than instructions.

## Validation

- `npx vitest test/assertions/videoRubric.test.ts test/matchers/video.test.ts test/util/video.test.ts test/providers/bedrock/video-utils.test.ts test/providers/defaults.test.ts test/providers/google/ai.studio.defaults.test.ts --run` — 95 passed.
- `npx vitest test/providers/bedrock/nova-reel.test.ts test/providers/bedrock/luma-ray.test.ts --run` — 64 passed.
- `npm run test:app -- src/pages/eval-creator/components/AssertsForm.test.tsx src/pages/eval-creator/components/assertionValueValidation.test.ts --run` — 127 passed.
- `npm run tsc`, `npm run build`, `npm run architecture:check`, `git diff --check ...`, and `SKIP_OG_GENERATION=true npm --prefix site run build` — passed.
- `npm run l` — exit 0 with existing nonblocking complexity warnings; no files changed.
- `npm run local -- eval -c /private/tmp/pr6925-e2e-review/promptfooconfig.yaml --no-cache -o /private/tmp/pr6925-e2e-review/orchestrator-output.json` — exit 100 as expected for three intentional failures; 1 managed-video pass, threshold/inverse failures, and external-URL rejection were inspected in the exported JSON.
- Full-matcher probes reproduced: malformed `{"pass":true,"score":1` returning pass/1.0; a 20,971,518-byte accepted prompt becoming a 20,971,592-byte AI Studio body against a 20,971,520-byte limit; a positive reason on a 0.6/0.8 threshold failure; a secret-shaped resolver message in the result; and one complete 16 MiB read before size rejection.

## Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command is not installed in this environment (exit 127), so I completed the six requested review lenses manually and reconciled three independent same-head specialist reports. No paid Gemini/Veo/Bedrock/S3 calls, secret-bearing workflows, or browser session were used.

src/util/video.ts
```diff
@@ -0,0 +1,133 @@
...
+      mimeType: blob.metadata.mimeType || video.blobRef.mimeType || 'video/mp4',
+    };
+  }
+
+  // Try media storage (Sora, Azure)
+  if (video.storageRef?.key) {
+    logger.debug('[VideoRubric] Resolving video from media storage', {
+      key: video.storageRef.key,
+    });
+    const buffer = await retrieveMedia(video.storageRef.key);
```
jameshiester-oai, line 60: [P2] Reject oversized media before reading the full object

The matcher enforces the inline limit only after this retrieval has materialized the complete buffer. A 16 MiB storage probe performed one full read and was then rejected; larger cloud objects or concurrent assertions can create avoidable I/O and memory pressure before the guard runs. Preflight `blobRef.sizeBytes`, preserve storage-ref size metadata (or add stat/bounded retrieval), and retain the post-read exact check as defense in depth.

src/matchers/video.ts
```diff
@@ -0,0 +1,257 @@
...
+          },
+        },
+        {
+          text: renderedGradingPrompt,
+        },
+      ],
+    },
+  ]);
+
+  if (Buffer.byteLength(multimodalPrompt, 'utf8') >= VIDEO_INLINE_LIMIT_BYTES) {
```
jameshiester-oai, line 172: [P2] Measure the final provider request against the 20 MB cap

This checks only the inner prompt. The default provider subsequently wraps it in `contents`, `generationConfig`, safety/tool configuration, and optional system instructions. A no-network probe produced a 20,971,518-byte prompt that passed here, but the minimal serialized AI Studio body was 20,971,592 bytes—over the 20,971,520-byte total-request limit; configured system instructions/tools widen the gap. Validate the provider's final serialized body or reserve provider-aware overhead before calling it.

src/matchers/video.ts
```diff
@@ -0,0 +1,257 @@
...
+  let { pass } = normalized;
+  const { score } = normalized;
+  const threshold =
+    typeof assertion?.threshold === 'string' ? Number(assertion.threshold) : assertion?.threshold;
+  const thresholdFailed =
+    typeof threshold === 'number' && Number.isFinite(threshold) && score < threshold;
+  pass = pass && !thresholdFailed;
+
+  const reason =
+    normalized.reason ||
```
jameshiester-oai, line 238: [P2] Report the threshold failure instead of a success reason

Because the model's reason wins unconditionally, a response `{pass:true, score:0.6, reason:"The video meets the rubric"}` with `threshold:0.8` yields `pass:false` while both the component reason and exported eval error say the video meets the rubric. This makes failed-eval evidence contradictory on the normal path where the judge supplies a reason. When `thresholdFailed` is true, make `Score 0.6 below threshold 0.8` take precedence (or append it unambiguously), and cover the positive-reason case.

architecture/edge-baseline.json
```diff
@@ -13,10 +13,10 @@
   "cli -> redteam": 10,
   "cli -> view-server": 6,
   "core -> contracts": 1,
-  "core -> legacy-contracts": 94,
-  "core -> legacy-runtime": 47,
-  "core -> node": 85,
-  "core -> providers": 22,
+  "core -> legacy-contracts": 96,
```
jameshiester-oai, line 16: [P2] Keep the dependency-count ratchet from growing

This refresh raises five cross-layer baselines by 13 imports (`core -> legacy-contracts/runtime/node/providers` and `node -> legacy-runtime`) so the new implementation passes `architecture:check`. The architecture guide explicitly says not to refresh this file merely to accommodate a newly introduced dependency; doing so permanently allows the added coupling. Keep the base counts and move video loading/default-provider construction behind an injected narrow port or another correctly owned layer, rather than importing node storage and concrete providers from the core matcher.

PR #6930, opened 2026-01-05
fix(providers): skip delay on cache hit in Bedrock Converse provider

### LINKED ISSUE #6929: The Bedrock Converse provider inserts unnecessary sleeps even when local cache hits
**Describe the bug**
When inserting a sleep using the `delay` option to avoid exceeding the LLM API rate limit, the Bedrock Converse provider inserts unnecessary sleeps even when promptfoo hits the local cache.

**To Reproduce**
1. Configure AWS credentials and Bedrock.
2. Save the following configuration file as `tmp/promptfooconfig.yaml`

```yaml
evaluateOptions:
  maxConcurrency: 1
providers:
  - id: bedrock:converse:amazon.nova-lite-v1:0
    delay: 5000
prompts:
  - 'Why is the sky blue?'
  - 'Why is the sea blue?'
tests:
  - assert:
      - type: icontains
        value: 'blue'
```

3. Run the command twice.

```bash
$ PROMPTFOO_CONFIG_DIR=tmp/.promptfoo npm run local -- eval -c tmp/promptfooconfig.yaml -v
```

The second execution should hit the local cache, but you can see that a sleep is inserted.

**Expected behavior**
When a local cache hit occurs, unnecessary sleep should not be inserted.

**Screenshots**
The following is an excerpt from the command output.

```
[logger.ts:1] Provider type: bedrock:converse:amazon.nova-lite-v1:0
[logger.ts:1] [EvaluatorTracing] isTracingEnabled check: metadata=false, yamlConfig=false, env=false, result=false
[logger.ts:1] Calling AWS Bedrock Converse API
{
  "modelId": "amazon.nova-lite-v1:0",
  "messageCount": 1,
  "hasSystem": false,
  "hasTools": false,
  "hasThinking": false
}
[logger.ts:1] Returning cached response
[logger.ts:1] Provider response properties: output, tokenUsage, cost, metadata
[logger.ts:1] Provider response cached property explicitly: undefined
[logger.ts:1] Evaluator response = {"output":"The sky appears blue due to a phenomenon called Rayleigh scattering. Here's a step-by-ste...
[logger.ts:1] Evaluator checking cached flag: response.cached = false, provider.delay = 5000
[logger.ts:1] Sleeping for 5000ms
[logger.ts:1] Tracked token usage for bedrock:converse:amazon.nova-lite-v1:0 (AwsBedrockConverseProvider): total=224, cached=0
[logger.ts:1] Eval #1 complete (1 of 2)
```

**System information:**

- Promptfoo version: 0.120.9

**Additional context**
The cache itself appears to be functioning correctly. This is confirmed by tests that succeed even when AWS authentication is disabled and the cache is present, indicating that no unnecessary API calls are being made.

minamijoyo:
Fixes #6929

The Bedrock Converse provider was not setting `cached: true` when returning responses from promptfoo's local cache. The cache itself worked correctly (API calls were skipped), but the missing flag caused `evaluator.ts` to execute unnecessary delays when using `delay` option, since it checks `response.cached` to skip sleeping for cached responses.

### DIFF
--- a/test/providers/bedrock/converse.test.ts
+++ b/test/providers/bedrock/converse.test.ts
@@ -1154,7 +1154,7 @@ Third line`;
       const mockCache = {
         get: vi.fn().mockResolvedValue(JSON.stringify(mockCachedResponse)),
         set: vi.fn(),
-      };
+      } as any;
 
       // Mock cache to be enabled and return the cached response
       const isCacheEnabledSpy = vi.spyOn(cacheModule, 'isCacheEnabled').mockReturnValue(true);


minamijoyo:
@mldangelo Thank you for working on fixing the CI failure. I rebased this patch onto the latest main branch, but another permission error is still occurring 🤔

minamijoyo:
@mldangelo Thank you for your support!

PR #6937, opened 2026-01-05
fix: keep column widths synced when Render Markdown is turned on

addelong:
Turning on Render Markdown in Table Settings could cause column widths to get out of sync between the header and body of the table if cell content was long. Applying `table-layout:fixed` resolves this.

Before:

image (4)

After:

Screenshot 2026-01-05 at 1 27 04 PM


faizanminhas:
@Claude are there any downsides to using table-layout: fixed? or potential side effects?

PR #6938, opened 2026-01-05
chore(claude): Add claude skill for mui migration + hook to run format after edit

faizanminhas:
Adds a mui-migration skill, a general `UI_GUIDELINES` markdown file, and adds a post edit hook.

Mostly looking for feedback on whether these are too long- I think once we finish the cloud migration, we can remove migration.md and the skill


addelong (approved):
I don't think this is too long, the examples will help the agent a lot, lgtm. Only nit would be to maybe rename MIGRATION.md to UI_MIGRATION.md so a wayward agent doesn't try to start building a UI when creating a DB migration.

faizanminhas:
> I don't think this is too long, the examples will help the agent a lot, lgtm. Only nit would be to maybe rename MIGRATION.md to UI_MIGRATION.md so a wayward agent doesn't try to start building a UI when creating a DB migration.

would note that since it's in the `src/app` directory, I think that wouldn't happen

PR #6954, opened 2026-01-07
feat(webui): Multi-input in redteam setup UI

faizanminhas:
UI for multi input

#### Summary
- Adds a new "Test Generation" accordion in common configuration options for targets
- Moves test generation instructions from application purpose to this section (discussed with @MrFlounder, as they're essentially the same thing; multi-input descriptions are just test generation instructs per variable)
- This also means test generation instructions and inputs are mutually exclusive ^ since they can collide. 
- Updates test target + test session to support multi-input
- Updates form validation when in multi-input mode (i.e. no error if `{{prompt}}` is missing from the body)
- Updates types where relevant

#### Demo

https://github.com/user-attachments/assets/418105c3-660f-48b6-937e-1a373c6284d4

### DIFF
--- a/src/app/src/pages/redteam/setup/components/Targets/CommonConfigurationOptions.test.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/CommonConfigurationOptions.test.tsx
@@ -193,4 +193,86 @@ describe('CommonConfigurationOptions', () => {
       screen.getByText('Configure how red team test cases are generated for this target'),
     ).toBeInTheDocument();
   });
+
+  describe('handleInputsChange', () => {
+    it('should generate JSON template with template variables when inputs are provided', () => {
+      const updateCustomTarget = vi.fn();
+
+      renderWithProviders(
+        <CommonConfigurationOptions {...defaultProps} updateCustomTarget={updateCustomTarget} />,
+      );
+
+      // Expand Test Generation section
+      const testGenTrigger = screen.getByText('Test Generation');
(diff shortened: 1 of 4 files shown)


src/app/src/pages/redteam/setup/components/Targets/InputsEditor.tsx
```diff
@@ -0,0 +1,279 @@
...
+        setVariables(variableList);
+        setIsExpanded(true);
+      } else {
+        setVariables([]);
+      }
+    }
+    isUpdatingFromProps.current = false;
+  }, [inputs]);
+
+  // Sync local state back to parent
```
will-holley: Let's get rid of the local state and update the parent state directly.

src/app/src/pages/redteam/setup/components/Targets/InputsEditor.tsx
```diff
@@ -0,0 +1,279 @@
...
+                    onChange={(e) => updateVariableName(variable.id, e.target.value)}
+                    className={cn('font-mono', variable.nameError && 'border-destructive')}
+                  />
+                  {variable.nameError && (
+                    <p className="mt-1 text-xs text-destructive">{variable.nameError}</p>
+                  )}
+                </div>
+                <div className="flex-1">
+                  <Label htmlFor={`${variable.id}-desc`} className="mb-1.5 block text-sm">
+                    Description
```
will-holley: Change to "Instructions" for consistency

PR #6960, opened 2026-01-07
fix(providers): await finalizeResponse to prevent race condition

JustinBeckwith:
## Summary
Fixes a race condition in the Google Live provider where async operations leaked between tests, causing flaky test failures in CI.

## Root Cause
The provider's `finalizeResponse()` was called without `await` in the WebSocket `onmessage` handler at three locations. Since the handler is already `async`, these unawaited calls caused:

1. Async fetch operations to continue running after the test completed
2. Test pollution when tests ran in random order (Vitest default)
3. Extra `fetchWithProxy` calls counted in subsequent tests (expected 6, got 7)

## Changes
- Changed all three `finalizeResponse()` calls from fire-and-forget pattern to properly awaited calls
- Wrapped in try-catch to maintain error handling behavior

## Test Plan
- ✅ Ran failing test 10 times with random seeds - all passed
- ✅ Running 100 iterations with fix to verify stability (in progress)
- ✅ Will run 100 iterations on main to confirm flake exists there

## Related Issue
Addresses flaky test failure: https://github.com/promptfoo/promptfoo/actions/runs/20795971834/job/59729461566

### DIFF
--- a/src/providers/google/live.ts
+++ b/src/providers/google/live.ts
@@ -466,10 +466,12 @@ export class GoogleLiveProvider implements ApiProvider {
               hasAudioStreamEnded = true;
             }
             if (hasTextStreamEnded && hasAudioStreamEnded) {
-              finalizeResponse().catch((err) => {
+              try {
+                await finalizeResponse();
+              } catch (err) {
                 logger.error(`Error in finalizeResponse: ${err}`);
                 safeResolve({ error: `Error finalizing response: ${err}` });
-              });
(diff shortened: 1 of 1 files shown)


mldangelo:
Awesome! Did you run an end to end test with this provider to verify that it still works?

JustinBeckwith:
@mldangelo sure did!  I mean I hadn't before you asked, but now I have 😆

PR #6983, opened 2026-01-09
chore(redteam): simplify team ID resolution for policy fetching

will-holley:
Always use the calling user's team id for fetching policies to enable fetching org-wide policies when an org-wide scan template is used.


mldangelo (approved):
Should this come with a test?

PR #6988, opened 2026-01-09
fix(app): use solid background for update banner in dark mode

mldangelo:
## Summary

- Fixes the update banner's semi-transparent background in dark mode that caused page content to bleed through
- Adds `dark:bg-blue-950` override to the UpdateBanner component for a solid background

## Test plan

- [x] Verified in browser with dark mode enabled
- [x] All UpdateBanner tests pass
- [x] Lint checks pass

**Before (content bleeding through):**
The banner's `info` variant uses `bg-blue-950/50` (50% opacity) in dark mode, causing the page title and description to show through.

**After (solid background):**
The banner now has a solid `bg-blue-950` background in dark mode, properly occluding content below.

### DIFF
--- a/src/app/src/components/UpdateBanner.tsx
+++ b/src/app/src/components/UpdateBanner.tsx
@@ -101,6 +101,8 @@ export default function UpdateBanner() {
       className={cn(
         'rounded-none py-2 px-4 relative z-(--z-banner)',
         'flex items-center justify-between gap-4',
+        // Use solid background to prevent content showing through the banner
+        'dark:bg-blue-950',
       )}
     >
       <div className="flex items-center gap-3">


src/app/src/components/UpdateBanner.tsx
```diff
@@ -101,6 +101,8 @@ export default function UpdateBanner() {
       className={cn(
         'rounded-none py-2 px-4 relative z-(--z-banner)',
         'flex items-center justify-between gap-4',
+        // Use solid background to prevent content showing through the banner
```
faizanminhas, line 104: Can we apply this to the underlying banner component ? Should we? Is this an issue with other banners in dark mode?

mldangelo:
Good question! I investigated this:

**Should we change the underlying component?** I'd recommend **no** for a few reasons:

1. **Most alerts are inline** - There are 18 `info` alerts across the app, and all except UpdateBanner are inline within page content where semi-transparent backgrounds work well and follow the [UI guidelines](https://github.com/promptfoo/promptfoo/blob/main/src/app/UI_GUIDELINES.md#7-dark-mode-use-opacity-not-solid-colors) recommending opacity in dark mode for subtlety.

2. **UpdateBanner is unique** - It's the only alert that sits at the page level with z-index, designed to overlay all content. It needs special handling.

3. **Impact scope** - Changing the base component would affect 30+ usages and alter the intentional design.

**Are other banners affected?** Potentially yes - there are 4 sticky `warning` alerts in `Plugins.tsx` and `Strategies.tsx` that use `sticky top-0 z-10` and also have `/50` opacity. However, they're sticky within containers (not page-level), so the issue is less prominent. If we see issues there, we can apply the same targeted fix.

**Alternative approach** if we want a component-level solution: We could add a `solid` prop or variant to Alert for cases where opacity isn't desired. But that's a larger change for a separate PR.

Let me know if you'd prefer a different approach!

faizanminhas (approved):

PR #7009, opened 2026-01-10
fix: update banner overlapped with header on datasets, prompts, history pages

typpo:
bad:

image

good:

image


src/app/src/pages/datasets/Datasets.tsx
```diff
@@ -1,8 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from 'react';
 
 import { DataTable } from '@app/components/data-table';
-import { PageContainer } from '@app/components/layout/PageContainer';
```
faizanminhas, line 4: I don’t think this is the right fix, we should consider moving the upgrade banner to be attached to the navigation container, floating & dismissible, rather than removing the standard layout component

faizanminhas (changes requested):

mldangelo-oai (commented):
Comment-only review at exact head `6a7456fb4784a37f4875bd07a575f7242913c988`. Recommend closing this PR as superseded; no successor is needed.

Current `main` already contains the complete banner-offset fix through #9081, #9087, and #9091: the banner publishes `--update-banner-height`, and every affected fixed page computes its top from `var(--nav-height)` plus that measured height. Those later changes also include responsive-navigation and mobile-layout coverage.

Although GitHub reports this branch as mergeable, the exact current-main merge tree is not safe: it reintroduces `FIXED_PAGE_CONTAINER_TOP = calc(3.5rem + ...)` and rewrites the page consumers to use it, replacing current main's responsive `--nav-height` contract. The existing unresolved Copilot thread on that hard-coded height is therefore still valid and sufficient; I did not duplicate it. The earlier requested-changes thread about preserving the shared layout is resolved at this head.

The branch is 510 commits behind current main. Its 37 published checks are green, `git diff --check` passes, and the focused UpdateBanner/Prompts tests pass locally (25/25), but those results do not justify regressing the newer responsive implementation. No branch changes or pushes were made.

PR #7021, opened 2026-01-12
fix(db): use parameterized SQL queries for filter operations

mldangelo:
## Summary

Converts string-based SQL construction to Drizzle's parameterized `sql` template strings in evaluation filtering and metrics calculation, following SQL injection prevention best practices.

### Important Context

**Promptfoo is a local-first tool** - the SQLite database stores evaluation results locally on the user's machine, and filter inputs come from the user's own configuration files and UI interactions. The risk of SQL injection in this context is extremely low since users are effectively querying their own data.

**These changes are made out of an abundance of caution** to:
- Follow security best practices regardless of threat model
- Establish patterns for contributors to follow
- Prevent any theoretical attack vectors in shared/team environments
- Satisfy security scanning tools that flag string interpolation in SQL

### Changes

- **`src/models/eval.ts`**: `buildFilterWhereSql()` now returns `SQL<unknown>` instead of `string`, using Drizzle's parameterized queries
- **`src/models/eval.ts`**: Add `buildSafeJsonPath()` helper for SQLite JSON path escaping (required because SQLite's `json_extract()` cannot parameterize paths)
- **`src/util/calculateFilteredMetrics.ts`**: Update interface to accept `SQL<unknown>` fragments instead of strings
- **`docs/agents/database-security.md`**: Add documentation on SQL security patterns for contributors
- **`AGENTS.md`**: Reference the new security documentation

### Technical Details

The key pattern change:

```typescript
// Before: String interpolation (flagged by security scanners)
const query = sql.raw(`SELECT * FROM eval_results WHERE eval_id = '${evalId}'`);

// After: Parameterized query (safe)
const query = sql`SELECT * FROM eval_results WHERE eval_id = ${evalId}`;
```

For SQLite JSON paths (which cannot be parameterized), proper escaping is applied:
- Backslashes: `\` → `\\`
- Double quotes: `"` → `\"`
- Single quotes: `'` → `''`

## Test plan

- [x] All 57 eval.test.ts tests pass
- [x] All 16 calculateFilteredMetrics.test.ts tests pass
- [x] Lint and format pass
- [x] Security audit confirms no high-risk patterns remain

### DIFF
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -119,8 +119,63 @@ export interface VarKeyResult {
  * Escapes a key for use in a JSON path expression.
  * Handles backslashes and double quotes which have special meaning in JSON paths.
  */
-function escapeJsonPathKey(key: string): string {
+export function escapeJsonPathKey(key: string): string {
   return key.replace(/\\/g, '\\\\').replace(/"/g, '\\"');
+}
+
+/**
+ * Builds a safe JSON path for use in SQLite json_extract() queries.
+ *
+ * SECURITY NOTE: This function uses sql.raw() which is normally unsafe, but is REQUIRED here
+ * because SQLite's json_extract() function only accepts JSON paths as string literals,
+ * not as parameterized values.
+ *
+ * Safety is ensured through double escaping:
+ * 1. JSON path characters are escaped (backslashes and double quotes)
+ * 2. SQL single quotes are escaped using standard SQL escaping ('' for ')
+ *
+ * @param field - The JSON field name from user input
+ * @returns A sql.raw() fragment containing the safely escaped JSON path
+ */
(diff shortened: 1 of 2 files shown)


src/models/eval.ts
```diff
@@ -613,206 +640,231 @@ export default class Eval {
...
-        conditions.push(`(${filterConditions.join(' ')})`);
+        const filterClause = filterConditions.reduce(
+          (acc, { condition: cond, logicOperator }, idx) => {
+            if (idx === 0) {
+              return cond;
+            }
+            return logicOperator === 'OR' ? sql`${acc} OR ${cond}` : sql`${acc} AND ${cond}`;
+          },
+          filterConditions[0].condition,
+        );
```
addelong: This bit has changed enough that it could stand to be extracted to its own function with a unit test.

PR #7029, opened 2026-01-12
chore(code-scan): bump action version to 0.1.3

danenania:


mldangelo (approved):
Should we release please this?

PR #7072, opened 2026-01-15
chore(ci): configure Ruby to use major.minor versioning only

JustinBeckwith:
## Summary

- Configure Renovate to extract only major.minor versions for Ruby (e.g., `3.4` instead of `3.4.8`)
- Update workflow to use Ruby 3.4 (from 3.3)
- This prevents PRs for every patch release (3.4.8, 3.4.9, 3.4.10, etc)
- Patches are automatically applied by the Ruby setup action

## Rationale

Previously, Renovate was suggesting specific patch versions (like `3.4.8`), which would trigger a new PR for every patch release. By configuring Renovate to only extract `major.minor`, we get:

- PRs only for significant version bumps (e.g., Ruby 3.5)
- Automatic patch updates without PR noise
- Consistent with how the Ruby setup action works

## Test plan

- [ ] Verify Renovate no longer creates PRs for Ruby patch versions
- [ ] Verify Ruby 3.4.x patches are automatically used by the action

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,7 +85,7 @@ jobs:
         with:
           python-version: 3.14
 
-      - name: Use Ruby 3.3
+      - name: Use Ruby
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: 3.4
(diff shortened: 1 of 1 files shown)


mldangelo (approved):

.github/workflows/main.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       - name: Use Ruby 3.3
```
mldangelo:
```suggestion
      - name: Use Ruby 3.4
```

mldangelo: here and below

PR #7086, opened 2026-01-16
feat(app): add post-hoc assertions with retry support

mldangelo:
## Summary

Adds post-hoc assertions to completed evaluation results, with scope selection, progress reporting, retries for failed grading, and assertion suggestion generation.

This revision hardens the draft feature before review:

- Serializes result mutations per evaluation so post-hoc jobs and manual ratings cannot overwrite one another's persisted aggregate metrics.
- Makes repeated jobs repair aggregate metrics after a partially persisted attempt, while keeping assertion additions idempotent.
- Treats sampled prompts/outputs as untrusted suggestion-model data, bounds API/model inputs, validates model responses, and avoids logging raw content.
- Uses single-flight polling in the dialog, communicates per-output grading cost accurately, and enables bulk assertion actions only for a real filtered subset.
- Documents the user-facing workflow and concurrency behavior in the Web UI guide.

## Design Notes

- The local OSS server does not implement eval ownership/authentication on adjacent mutation endpoints; the scanner auth thread was reconciled against that existing architecture and remains resolved.
- The change updates saved results and prompt metrics without rerunning the target provider.
- No Drizzle schema or migration files are part of this feature; all database-backed QA used isolated `PROMPTFOO_CONFIG_DIR` and `PROMPTFOO_DB_PATH` state.

## QA

- [x] `npm run l && npm run f` (passes; warning-only cognitive-complexity diagnostics remain in three draft UI areas and one existing `main` area)
- [x] `PROMPTFOO_CONFIG_DIR=/tmp/pr7086-focused-config PROMPTFOO_DB_PATH=/tmp/pr7086-focused-config/promptfoo.db npx vitest test/server/eval.test.ts test/assertions/generateSuggestions.test.ts test/util/recalculatePromptMetrics.test.ts test/models/eval.filteredTestIndices.test.ts test/commands/retry.test.ts --run --sequence.shuffle=false` (`65` tests passed)
- [x] `npm run test:app -- src/pages/eval/components/AddAssertionsDialog.test.tsx src/pages/eval/components/PosthocAssertionsForm.test.tsx src/pages/eval/components/ResultsView.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx src/utils/api.test.ts --run` (`230` tests passed)
- [x] `npm --prefix src/app run test -- --run` (`4,570` tests passed)
- [x] `PROMPTFOO_CONFIG_DIR=/tmp/pr7086-full-config PROMPTFOO_DB_PATH=/tmp/pr7086-full-config/promptfoo.db npm test` (`18,207` tests passed, `10` skipped)
- [x] `npm run tsc && npm --prefix src/app run tsc --if-present && npm run build`
- [x] Isolated end-to-end eval and API proof: ran an `echo` eval with `--no-cache`, applied a deterministic assertion through `POST /api/eval/:evalId/assertions`, polled completion, replayed it to confirm duplicate skipping plus metric recalculation, verified persisted table metrics, and confirmed oversized suggestion instructions return `400`.
- [x] Isolated state audit: the generated SQLite database, WAL/SHM, config, and log files were all mode `600`.


ghost:
**⏩ Already incorporated tests (ef51cd8236f931167f2911fe1e4e12affca345b6) [View tests ↗](https://app.usetusk.ai/app/testing-commit-check/c3e0b4d4-39b3-431b-89e7-456dbb66969d?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

---
<p>

| f7f8a7e59383e8ec11a4c0319cd74570a2ca6cea | 🔄 Running Tusk Tester | [Output](https://app.usetusk.ai/app/testing-commit-check/15549425-1a29-4107-bbc4-ba72e667c04f?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 16, 2026 8:57PM |
| b979f535fa3c63649de852e3cac3018348d1cb61 | ✅ Generated 16 tests - 16 passed | [Tests](https://app.usetusk.ai/app/testing-commit-check/6af6d480-dac9-4ba1-ab52-99a84d31fd89?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 16, 2026 9:36PM |
| 038a6ed8c42b7e55f1838256994f12b5b7632676 | ✅ Generated 16 tests - 16 passed | [Tests](https://app.usetusk.ai/app/testing-commit-check/c3e0b4d4-39b3-431b-89e7-456dbb66969d?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 16, 2026 11:25PM |
| ef51cd8236f931167f2911fe1e4e12affca345b6 | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/94828bea-f16f-4302-b429-9e07192dd849?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 20, 2026 1:58AM |

</p>
</details>

mldangelo-oai:
@codex review

mldangelo-oai:
Synced this PR with current `main`, reviewed the existing review threads, and pushed follow-up fixes in `a46f52facadf21c9091fa6585479f30dd79adc01`.

What changed:
- Kept `pi` model-backed for cost estimates, but no longer treats it as a value-less assertion. The UI now requires evaluator instructions before submitting a PI post-hoc assertion.
- Large-run confirmation now triggers when either the output count or the total model-backed assertion request estimate crosses the threshold.
- Fixed the filtered test-index regression test to use persisted `EvalResult` rows directly, which keeps `npm run tsc` clean.

Validation:
- `npm run test:app -- src/pages/eval/components/AddAssertionsDialog.test.tsx src/pages/eval/components/PosthocAssertionsForm.test.tsx src/utils/api.test.ts --run`
- `PROMPTFOO_CONFIG_DIR=/tmp/pr7086-focused-config PROMPTFOO_DB_PATH=/tmp/pr7086-focused-config/promptfoo.db npx vitest --run test/server/eval.test.ts test/assertions/generateSuggestions.test.ts test/util/recalculatePromptMetrics.test.ts test/models/eval.filteredTestIndices.test.ts test/commands/retry.test.ts --sequence.shuffle=false`
- Isolated CLI/server E2E: ran an echo eval with isolated config/cache/db, started `npm run dev:server`, added a post-hoc assertion via `/api/eval/:id/assertions`, polled the assertion job to completion, verified duplicate assertions are skipped, verified table metrics/named score update, and verified oversized suggestion instructions return 400.
- `npm run f && npm run l` (clean exit; Biome still reports existing complexity warnings in changed PR files)
- `npm run tsc`
- `npm run build`

CI is queued on the new head; moving on to the next PR per instructions.

jameshiester-oai (commented):
## Codex review

**Scope:** Full review of `c54a30668ad8319d76c20ae96e6680ad6c51a2c6...997661369f585763243b13f69f68f425b264f2cf` at the exact detached head. No prior authenticated completion marker exists for this head.

**Verdict:** Changes are not ready to merge. I verified three P1 data-integrity/security failures and eight new P2 correctness or operability failures. GitHub reports the PR as conflicting with current `main`; the three red checks (Site tests, Node 20 Ubuntu, and webui tests) complete their test suites and then fail in the shared Codecov CLI GPG-verification step, so I did not attribute those failures to this patch.

### Verified findings

**P1**
- Request-controlled assertion credentials are persisted without the normal secret sanitizer and reappear in eval summaries/exports.
- Filtered bulk application can target rows outside the table's displayed selection; the unsupported `different` mode broadens to the whole eval.
- Trace-aware assertions collapse multiple output traces sharing a `testCaseId`, causing persisted false passes/failures.

**P2**
- `select-best` / `max-score` are accepted and persisted even though single-output assertion execution skips them.
- A non-settling supported assertion can retain the eval-wide mutation lock indefinitely.
- One transient polling error abandons a live server job with no retry/resume path.
- Scoped suggestion generation pairs selected outputs with unrelated global prompts.
- Suggestion generation rejects valid primitive outputs such as `0` and `false`.
- Cell replay swaps only displayed text, leaving grading and mutation actions bound to the stored output.
- Comparison mode's mutation guard is bypassed through the Details sheet.
- A manual rating rejected by the new mutation lock remains optimistically visible even though it was not saved.

The previously resolved [custom-scoring thread](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3292646991) is also still reproducible for persisted evals: evaluator-time resolution turns a file scorer into a function, persistence drops it, and this route falls back to default aggregation. My isolated replay produced score `0.5` instead of the custom scorer's `0.99`.

### Existing feedback

I rechecked and did not duplicate the nine current unresolved threads. The important P2 threads remain: [semantic assertion identity](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368768809), [large filtered `IN` loading](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368768816), [wrong-provider cell replay](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368768819), [reservation after mutable target loading](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368786833), [array-value coercion](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368786835), [value-less RAG validation](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368786837), and [legacy rows losing prior assertion effects](https://github.com/promptfoo/promptfoo/pull/7086#discussion_r3368786839). The two P3 test-mock reset threads also remain current.

- Use one canonical query for displayed and mutation-target test indices; the current optimized and full search paths have different semantics.
- Replace duplicated assertion-type sets with one typed post-hoc capability table (value shape, value requirement, model-backed status, and post-hoc support).
- The route currently materializes every eval result and randomly sorts it before retaining 20 suggestions, and this PR raises eleven architecture-edge baselines. Move orchestration into a focused service and sample in the database instead of ratcheting cross-layer coupling.

### Verification

- `PROMPTFOO_CONFIG_DIR=/private/tmp/pr7086-tests-config PROMPTFOO_DB_PATH=/private/tmp/pr7086-tests-config/promptfoo.db npx vitest run test/server/eval.test.ts test/assertions/generateSuggestions.test.ts test/util/recalculatePromptMetrics.test.ts test/models/eval.filteredTestIndices.test.ts test/openapi/serverOpenApi.test.ts test/server/routes/serverRouteSmoke.test.ts test/types/api/eval.test.ts` — exit 0; 7 files, 132 tests.
- `npm run test:app -- src/pages/eval/components/AddAssertionsDialog.test.tsx src/pages/eval/components/PosthocAssertionsForm.test.tsx src/pages/eval/components/ResultsView.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx src/pages/eval/components/EvalOutputCell.duplicateImages.test.tsx src/utils/api.test.ts --run` — exit 0; 6 files, 272 tests.
- `npm run tsc`, `npm run build`, `SKIP_OG_GENERATION=true npm run build` from `site/`, `npm run architecture:check`, and `git diff --check` — exit 0.
- `npm run l` — exit 0 with warning-only complexity diagnostics; no fixes were applied.
- Isolated no-cache CLI/API QA under `/private/tmp/pr7086-e2e` — eval exit 0; 12/12 passed, no errors. Direct API checks covered duplicate skipping, test/filtered scopes, partial failure/retry, persisted metrics, and unsupported assertion behavior.
- `PROMPTFOO_CONFIG_DIR=/private/tmp/pr7086-api-config3 PROMPTFOO_DB_PATH=/private/tmp/pr7086-api-config3/promptfoo.db node --import tsx /private/tmp/pr7086-api-repro.mjs` — exit 0; reproduced credential persistence, lost custom scoring, trace collision, indefinite locking, and selected-prompt omission.
- `PROMPTFOO_CONFIG_DIR=/private/tmp/pr7086-correctness-probe-4f6a PROMPTFOO_DB_PATH=/private/tmp/pr7086-correctness-probe-4f6a/promptfoo.db node --import tsx /private/tmp/pr7086-filter-readback.mjs` — exit 0; displayed indices `[1]`, mutation indices `[0,1]`, and `different` indices `[0,1,2]`.

### Limitations

The requested PR Review Toolkit command was unavailable in this environment, so I performed its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses manually and reconciled three read-only specialist passes. I did not merge current `main` into the head; a read-only merge analysis shows conflicts in `architecture/edge-baseline.json`, `src/models/eval.ts`, and `src/node/retry.ts`. I did not perform visual browser QA; UI behavior was exercised through focused React tests and source-level interaction tracing.

src/server/routes/eval.ts
```diff
@@ -787,6 +1046,432 @@ evalRouter.post(
...
+          result.score = aggregated.score;
+          result.failureReason = aggregated.pass
+            ? ResultFailureReason.NONE
+            : ResultFailureReason.ASSERT;
+          result.testCase = {
+            ...result.testCase,
+            assert: [...existingAssertions, ...assertionsToAdd],
+          };
+
+          await result.save();
```
jameshiester-oai, line 1278: **[P1] Sanitize post-hoc assertions before persisting the result**

The `assertions` payload is request-controlled and permits nested `config`. This call reaches `EvalResult.save()`, whose update path writes `testCase` and `gradingResult` without `sanitizeForDbWithSecrets`, unlike the create/bulk-insert paths. I posted an assertion containing a synthetic `config.apiKey`; the job completed and the exact marker remained in both persisted JSON fields and `toEvaluateSummary()`, so exports/share can expose it too. Apply the standard credential sanitizer on every save (preferably centrally in `save()`) and add an update-path regression test.

src/server/routes/eval.ts
```diff
@@ -28,19 +41,259 @@ import { setDownloadHeaders } from '../utils/downloadHelpers';
...
+      JSON.stringify({
+        logicOperator: filter.logicOperator ?? 'and',
+        type: filter.type,
+        operator: filter.operator,
+        value: filter.value,
+        field: filter.field,
+      }),
+    ) || [];
+
+  const { testIndices, filteredCount } = await evalRecord.getFilteredTestIndices({
```
jameshiester-oai, line 285: **[P1] Resolve exactly the rows displayed by the active filter**

This target query does not share the table's semantics. A search-only table uses `queryTestIndicesOptimized()` (response-only search), while this path uses the broader JSON search; additionally, `different` has no backend predicate and falls through to all rows. In an isolated eval, the table displayed test `[1]` but this resolver returned `[0,1]`; the POST then persisted the assertion to both and changed the hidden row from pass to fail. The same probe returned all `[0,1,2]` for `different`. Use one canonical selector for display and mutation, and reject any mode whose exact row semantics are not implemented.

src/server/routes/eval.ts
```diff
@@ -787,6 +1046,432 @@ evalRouter.post(
...
+
+  const maxConcurrency = Math.max(
+    1,
+    eval_.runtimeOptions?.maxConcurrency ?? DEFAULT_MAX_CONCURRENCY,
+  );
+
+  void (async () => {
+    try {
+      const traceIdsByTestCaseId = hasTraceAwareAssertions(normalizedAssertions)
+        ? new Map((await eval_.getTraces()).map((trace) => [trace.testCaseId, trace.traceId]))
```
jameshiester-oai, line 1196: **[P1] Resolve a trace per output instead of per testCaseId**

`testCaseId` is not unique across prompt/provider outputs when a test supplies an explicit metadata ID, so this map is last-write-wins. I created two outputs with the same ID: the first trace had `llm.call`, the second only `other.call`. Applying `trace-span-count` for `*llm*` to both persisted two failures and reported zero spans for the first output. Match trace metadata `testIdx` + `promptIdx` (with the ID only as a secondary check) and cover a shared-ID multi-output eval; otherwise trace/security assertions produce order-dependent false positives and false negatives.

src/types/api/eval.ts
```diff
@@ -286,6 +292,101 @@ export type SubmitRatingParams = z.infer<typeof SubmitRatingParamsSchema>;
...
+  type: z.enum(['results', 'tests', 'filtered']),
+  resultIds: z.array(z.string().min(1).max(256)).max(10000).optional(),
+  testIndices: z.array(z.number().int().nonnegative()).max(10000).optional(),
+  filters: z.array(PosthocResultsFilterSchema).max(100).optional(),
+  filterMode: EvalResultsFilterMode.optional(),
+  searchText: z.string().max(10000).optional(),
+});
+
+export const AddEvalAssertionsRequestSchema = z.object({
+  assertions: z.array(AssertionOrSetSchema).min(1).max(100),
```
jameshiester-oai, line 319: **[P2] Reject assertion types that cannot execute post hoc**

The general assertion schema admits comparison assertions such as `select-best` and `max-score`, but `runAssertions()` intentionally skips them because they require multiple outputs. An actual `select-best` request completed with `updatedResults: 1` and `passCount: 1`, persisted the assertion, and added no component result or grading effect; retries then treat it as a duplicate. Define a recursive post-hoc-supported schema (including nested sets), or implement batch semantics, and fail rather than save when a newly requested assertion yields no component result.

src/server/routes/eval.ts
```diff
@@ -787,6 +1046,432 @@ evalRouter.post(
...
+            ...(result.response ?? { output: result.error ?? '' }),
+            cost: result.cost,
+          };
+          const provider =
+            assertionsToAdd.some(assertionUsesOriginalProvider) && result.provider.id
+              ? await loadApiProvider(result.provider.id, { options: result.provider })
+              : undefined;
+          const promptText =
+            typeof result.prompt === 'string' ? result.prompt : (result.prompt?.raw ?? '');
+          const newGradingResult = await runAssertions({
```
jameshiester-oai, line 1240: **[P2] Bound assertion execution before holding the eval-wide lock**

There is no timeout, abort, or cancellation around this call. Posting the supported JavaScript assertion `new Promise(() => {})` left the job `in-progress` at progress 0 and every later rating returned 409 indefinitely because the lock is released only after this promise settles. Provider-backed graders can create the same operational failure. Add a bounded per-result/job deadline with cleanup and an explicit cancel/recovery path so one assertion cannot permanently freeze all mutations for an eval.

src/app/src/pages/eval/components/AddAssertionsDialog.tsx
```diff
@@ -0,0 +1,746 @@
...
+        }
+
+        pollingRef.current = setTimeout(() => {
+          void pollJobStatus(jobId);
+        }, 500);
+      } catch (error) {
+        if (activeJobIdRef.current !== jobId) {
+          return;
+        }
+        stopPolling();
```
jameshiester-oai, line 325: **[P2] Keep polling a live job after a transient status failure**

Any temporary network/5xx failure enters this branch, `stopPolling()` clears the only job ID, and no retry or resume path remains even though the server job continues. The dialog becomes submittable again, a retry can receive 409, and when the original job finishes `onApplied` is never called, leaving stale results and hiding partial errors. Retain the job ID and use bounded backoff/reconnection (or expose an explicit resume action), with a test for one rejected poll followed by terminal success.

src/server/routes/eval.ts
```diff
@@ -787,6 +1046,432 @@ evalRouter.post(
...
+    const { type, numAssertions, instructions, provider, testIndices, resultIds } = parsed.data;
+
+    try {
+      const eval_ = await Eval.findById(evalId);
+      if (!eval_) {
+        res.status(404).json({ error: 'Eval not found' });
+        return;
+      }
+
+      const prompts = eval_.prompts.map((p) => p.raw).filter(Boolean);
```
jameshiester-oai, line 1382: **[P2] Derive suggestion prompts from the selected results**

`resultIds` and `testIndices` scope only the outputs; prompts always come from the eval-wide list and are truncated to the first five. In a six-prompt eval, requesting suggestions for only the sixth result sent prompts 1–5 while sending output 6; a deterministic provider confirmed the selected prompt was absent. This can spend a model call generating criteria for unrelated instructions. Resolve each selected result's `promptIdx`/prompt and keep prompt-output context aligned before applying the bounds.

src/server/routes/eval.ts
```diff
@@ -787,6 +1046,432 @@ evalRouter.post(
...
+      }
+
+      if (results.length === 0) {
+        res.status(400).json({ error: 'No results found in eval' });
+        return;
+      }
+
+      const outputs = results
+        .map((result) => {
+          if (result.response?.output) {
```
jameshiester-oai, line 1416: **[P2] Preserve valid primitive provider outputs**

This truthiness test drops `0`, `false`, and the empty string even though provider output is an unknown-valued public contract and those values can be valid results. With a persisted `response.output: 0` and an otherwise valid local suggestion provider, the endpoint returned 400 `No valid outputs found in results`. Exclude only `null`/`undefined` and stringify other primitives; add route tests for `0` and `false`.

src/app/src/pages/eval/components/EvalOutputCell.tsx
```diff
@@ -1553,6 +1685,58 @@ function EvalOutputCell({
...
+        evaluationId,
+        testIndex: testIdx,
+        prompt: typeof output.prompt === 'string' ? output.prompt : JSON.stringify(output.prompt),
+        variables: output.metadata?.inputVars || output.testCase?.vars,
+      });
+
+      if (result.error) {
+        showToast(`Re-run failed: ${result.error}`, 'error');
+      } else {
+        setReplayState({ outputIdentity, text: result.output ?? '' });
```
jameshiester-oai, line 1726: **[P2] Do not present replay text as the persisted cell**

This updates only local display text. The cell's status, grading results, rating callbacks, and Add Assertions dialog remain bound to the original `output` and `output.id`; `refreshTable()` does not clear the replay state when the stored text is unchanged. After a successful rerun, a user can therefore see fresh text while rating or re-grading the old response. Persist the rerun as a result, or label/isolate transient replay state and disable persisted grading/mutation actions until the user returns to the stored output.

src/app/src/pages/eval/components/EvalOutputCell.tsx
```diff
@@ -1222,8 +1303,9 @@ function renderOutputActions({
               providerPrompt={getActualPrompt(output.response, { formatted: true })}
               evaluationId={evaluationId}
               testCaseId={testCaseId || output.id}
-              testIndex={rowIndex}
+              testIndex={testIdx ?? rowIndex}
               promptIndex={promptIndex}
+              resultId={output.id}
```
jameshiester-oai, line 1308: **[P2] Propagate the comparison-mode read-only guard into Details**

The More menu honors `mutationsDisabled`, but this `EvalOutputPromptDialog` call does not pass `readOnly={mutationsDisabled}`; its default is false and its Evaluation tab exposes Add assertions. In comparison mode that bypasses the intended guard: main-eval cells can be mutated, while comparison-eval cells pair a comparison result ID with the main eval ID and fail as a misleading 404. Pass the read-only flag through and cover Details → Evaluation in comparison mode.

src/server/routes/eval.ts
```diff
@@ -709,8 +962,14 @@ evalRouter.post(
       return;
     }
 
+    const { evalId, id } = paramsResult.data;
+    const operationId = `rating:${crypto.randomUUID()}`;
+    if (!reserveEvalMutation(evalId, operationId)) {
```
jameshiester-oai, line 967: **[P2] Make clients recover when the new lock rejects a rating**

This new 409 path is valid server-side, but `ResultsTable.handleRating` optimistically calls `setTable(newTable)` before saving and its catch only logs to the console. If an assertion job is running in another tab, a rating/comment appears saved in the UI, receives this 409, and silently disappears on refresh. Surface the conflict and rollback/refetch while preserving a retryable draft (or queue the mutation) so the UI cannot claim persistence that did not happen.

PR #7099, opened 2026-01-18
docs(site): mark products as sold out in store

mldangelo:
## Summary

- Add sold-out indicators to store product cards and block adding sold-out products from the modal.
- Align store availability types with Fourthwall's current nested product state and variant stock response shape.
- When a product is available but its first variant is out of stock, open the modal on the first purchasable variant instead of presenting a misleading disabled purchase action.

## Design Notes

- Product-level `state.type === 'SOLD_OUT'` is authoritative. A live Storefront API check on May 23, 2026 returned sold-out products whose variants still report `stock.type === 'UNLIMITED'`, so relying only on variant inventory would permit an unavailable purchase.
- Variant inventory still controls the selected option for otherwise available products; the modal selects an in-stock option initially and remains disabled when the chosen option cannot be bought.
- This is a site-only UI/API-contract change. It has no database or migration surface.

## QA

- [x] Queried the live public Fourthwall collection response and verified the nested `state`, `access`, and `stock` shapes, including real `SOLD_OUT` products with `UNLIMITED` variants.
- [x] `npx @biomejs/biome check --write site/src/components/Store/ProductCard.test.tsx site/src/components/Store/ProductCard.tsx site/src/components/Store/ProductModal.test.tsx site/src/components/Store/ProductModal.tsx site/src/components/Store/index.ts site/src/components/Store/types.ts site/src/components/Store/useFourthwall.test.ts site/src/components/Store/useFourthwall.ts` (passes with one warning-only pre-existing complexity diagnostic in `stripHtml`)
- [x] `npm --prefix site run test -- src/components/Store/useFourthwall.test.ts src/components/Store/ProductCard.test.tsx src/components/Store/ProductModal.test.tsx --run` (`43` tests passed)
- [x] `npm --prefix site run typecheck`
- [x] `npm --prefix site run test -- --run` (`106` tests passed)

`npm run f` successfully formats the TypeScript diff and then exits in the repository wrapper's empty Prettier-input path (`No parser and no file path given`) because this PR has no Prettier-target files; the explicit Biome formatting check above covers the changed source.


mldangelo-oai:
@codex review

mldangelo-oai:
Synced this PR with current `main` and pushed follow-up cleanup in `1cbf2e1aab7009aba8428eb684211dfdf8310e46`.

What changed:
- Kept the sold-out/store behavior intact and simplified the SSR `stripHtml` fallback into smaller helpers, clearing the Biome complexity warning in the changed store utility.
- Rechecked the existing review thread; the image mock type drift is already resolved/outdated on the current branch.
- Verified the live Fourthwall Storefront response still has product-level `state.type` and found current `SOLD_OUT` products whose variants report `UNLIMITED` stock, so the product-level sold-out guard remains necessary.

Validation:
- `npm --prefix site run test -- src/components/Store/useFourthwall.test.ts src/components/Store/ProductCard.test.tsx src/components/Store/ProductModal.test.tsx --run`
- `npx @biomejs/biome check --write site/src/components/Store/ProductCard.test.tsx site/src/components/Store/ProductCard.tsx site/src/components/Store/ProductModal.test.tsx site/src/components/Store/ProductModal.tsx site/src/components/Store/index.ts site/src/components/Store/types.ts site/src/components/Store/useFourthwall.test.ts site/src/components/Store/useFourthwall.ts`
- `npm run l`
- `npm --prefix site run fetch-stats && npm --prefix site run typecheck`
- `npm --prefix site run test -- --run`

CI is queued on the new head; moving to the next PR per instructions.

site/src/components/Store/useFourthwall.ts
```diff
@@ -354,85 +355,127 @@ const HTML_ENTITIES: Record<string, string> = {
...
 
 // Check if variant is in stock
-export function isInStock(stock: { type: string; quantity?: number }): boolean {
+export function isInStock(stock: FourthwallStock): boolean {
   if (stock.type === 'UNLIMITED') {
     return true;
   }
-  if (stock.type === 'LIMITED' && typeof stock.quantity === 'number') {
-    return stock.quantity > 0;
+  return stock.inStock > 0;
```
jameshiester-oai, line 470: [P2] Fail closed for unknown stock discriminators

Remote JSON reaches this helper without runtime validation, so the TypeScript union does not make this branch exhaustive at runtime. Every non-`UNLIMITED` object is now treated as `LIMITED`; a direct probe with `{ type: 'BACKORDERED', inStock: 2 }` returns `true`. The base implementation explicitly returned false for unknown types, and this PR removes that regression test. That fail-open result feeds product sold-out detection, default variant selection, and the enabled Add to Cart path. Handle `LIMITED` explicitly (including finite numeric validation) and return false for every other discriminator, then restore an adversarial unknown-type test.

site/src/components/Store/ProductCard.tsx
```diff
@@ -21,6 +21,7 @@ export function ProductCard({ product, onClick }: ProductCardProps) {
   const primaryImage = product.images[0];
   const hoverImage = product.images[1];
   const hasHoverImage = Boolean(hoverImage);
+  const soldOut = isProductSoldOut(product);
 
   const lowestPrice = useMemo(
```
jameshiester-oai, line 26: [P2] Price the purchasable variants

This reduction still includes out-of-stock variants, while the new modal behavior selects the first in-stock variant. With a $10 `LIMITED/inStock: 0` option followed by a $20 purchasable option, the card renders $10 but the modal immediately renders/selects $20, so the advertised price cannot be bought. I reproduced that interaction with a component test. Compute the available product's minimum from in-stock variants (falling back to all variants only for a completely sold-out display) and cover the mixed-inventory case.

jameshiester-oai (commented):
## Review scope

- **Base:** `main` at `c54a30668ad8319d76c20ae96e6680ad6c51a2c6`
- **Head:** `1dee6ae194af699def9edff43c68831b89317af9`
- **Scope:** Full fresh review. No earlier authenticated completion marker exists for this head.
- **Verdict:** The product-level sold-out override and disabled modal behavior work as intended, but two P2 correctness defects remain in the availability workflow.
- **CI:** Three checks are red, but their logs are not attributable to this change: Node 20 stopped in `npm ci` with `ECONNRESET`; Site tests passed 106/106 before Codecov GPG verification failed; Web UI tests passed 4,816/4,816 before the same Codecov failure. The remaining checks pass.

## Verified findings

### P2

1. `isInStock` now treats every unknown non-`UNLIMITED` runtime discriminator as limited inventory. Because API JSON is only type-cast, a future or malformed value such as `{ type: 'RESERVED', inStock: 1 }` is considered purchasable. This affects sold-out classification, initial variant selection, and the Add to Cart guard.
2. Product cards compute their advertised minimum across unavailable variants, while the modal now skips those variants. A product with a $10 out-of-stock option and a $20 in-stock option therefore advertises $10 and opens at $20.

No P0, P1, or additional P3 defects were validated. Details and correction directions are attached inline.

## Existing feedback

The sole earlier inline thread about image mock fields is resolved and outdated on this head. No applicable unresolved feedback was duplicated.

A small shared helper that returns purchasable variants would keep card pricing, product sold-out state, and modal default selection on one rule. The focused tests should restore an unknown-discriminator fail-closed case and add a mixed-stock price/default-selection case. Rejected add-to-cart requests also lack user-visible error coverage, but that catch path predates this PR and is not raised as a new defect here.

## Tests and QA

- `npm --prefix site run test -- src/components/Store/useFourthwall.test.ts src/components/Store/ProductCard.test.tsx src/components/Store/ProductModal.test.tsx --run` — 43/43 passed.
- `npm exec vitest -- --config /private/tmp/pr7099-vitest.config.ts --run` from `site/` — 2/2 temporary behavior probes passed, reproducing the mixed-stock price mismatch and silent rejected-cart path.
- `npm --prefix site run test -- --run` — 106/106 passed.
- `npm --prefix site run typecheck` — passed.
- `npm run lint:site` — passed with three pre-existing complexity warnings outside the changed Store files.
- `SKIP_OG_GENERATION=true npm --prefix site run build` — passed.
- Direct helper probe — `{ type: 'BACKORDERED', inStock: 2 }` returned `true`.
- Read-only live Fourthwall aggregate probe — 61 products over two pages; 59 `AVAILABLE`, 2 `SOLD_OUT`; all 765 current variants were `UNLIMITED`; both sold-out products still had unlimited variants.
- Browser QA against the local Store — both sold-out cards exposed accessible “(Sold Out)” labels and overlays; the sold-out modal showed a disabled Sold Out button; an available product showed an enabled Add to Cart button. At 390×844 the modal filled the viewport without horizontal overflow and retained a 52.5px-high purchase button.

## Limitations

`/pr-review-toolkit:review-pr all parallel` was attempted but is unavailable in this environment (exit 127), so the six review lenses were completed manually and reconciled with three independent specialist reports. The live catalog had no limited or mixed-stock product, so those paths were exercised with deterministic component/helper probes. No real cart mutation was made. Browser console errors were limited to unrelated third-party Vector/Reo analytics scripts.

PR #7130, opened 2026-01-20
fix(deps): upgrade undici to 6.23.0 to address CVE-2026-22036

### LINKED ISSUE #238: Run Python from `tests.csv`
`tests.csv` offers a convenient way of listing many test cases. With the `__expected` field, an assertion can be specified that must hold true after a test case.
While JavaScript is supported with the `fn` keyword, Python isn't. 

On another note, it seems that `async` javascript functions don't work with `tests.csv`.

mldangelo:
## Summary
- Adds npm override to force `undici` to `^6.23.0` in code-scan-action
- Addresses Dependabot alert #238: unbounded decompression chain vulnerability (CVE-2026-22036)

## Context
The `@actions/github` and `@actions/http-client` packages specify `undici@^5.28.5`, but the security fix only exists in version 6.23.0+. Since the upstream packages haven't updated their dependency ranges, an npm override is required.

## Verification
- `npm audit` shows 0 vulnerabilities
- `npm run build` succeeds
- `npm ls undici` confirms version 6.23.0

## Test plan
- [x] Verified `npm audit` passes with no vulnerabilities
- [x] Verified build succeeds
- [ ] CI passes

Fixes #238

### DIFF
--- a/code-scan-action/package-lock.json
+++ b/code-scan-action/package-lock.json
@@ -99,15 +99,6 @@
         "node": ">=6.9.0"
       }
     },
-    "node_modules/@fastify/busboy": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/@fastify/busboy/-/busboy-2.1.1.tgz",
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
One concern I have here - at some point this override will become unnecessary, and we'll have no idea why it's here.  Any thoughts or ideas on how we keep track?

PR #7164, opened 2026-01-21
fix(retry): prevent data loss when retry fails and add cloud sync

minhle1291:
## Summary
- Fixes critical data loss bug in retry command where ERROR results were deleted before retry, causing permanent data loss if retry failed
- Applies same fix to `eval --retry-errors` command path
- Adds cloud sync support to retry command with `--share/--no-share` flags

## Changes
- **Data loss fix**: Moved ERROR result deletion from BEFORE to AFTER successful retry
- **New `retryMode` flag**: Added to `cliState` to distinguish retry from regular resume mode
- **`excludeErrors` option**: Added to `getCompletedIndexPairs()` to exclude ERROR results during retry
- **Cloud sync**: Added sharing support with same precedence as `eval` command
- **Error handling**: Wrapped post-retry cleanup in try-catch to prevent cleanup failures from breaking successful retries
- **Logging**: Fixed error logging to preserve stack traces using object format
- **Type safety**: Added runtime validation for config type assertions
- **Tests**: Added comprehensive test coverage for retry command and `getCompletedIndexPairs`

## Test plan
- [x] Unit tests pass for retry command (`test/commands/retry.test.ts`)
- [x] Unit tests pass for `getCompletedIndexPairs` with `excludeErrors` option
- [x] Linting passes
- [x] Manual test: Retry command preserves ERROR results on failure
- [x] Manual test: Retry command deletes ERROR results only after success
- [x] Manual test: `eval --retry-errors` has same behavior


src/commands/retry.ts
```diff
@@ -24,6 +27,44 @@ interface RetryCommandOptions {
...
+/**
+ * Determines whether results should be shared to cloud.
+ * Uses same precedence as eval command:
+ * 1. Explicit disable (CLI --no-share or env var) takes highest priority
+ * 2. Explicit enable (CLI --share)
+ * 3. Config file commandLineOptions.share
+ * 4. Config file sharing setting
+ * 5. Default: auto-share when cloud is enabled
+ */
+export function shouldShareResults(opts: ShouldShareOptions): boolean {
```
mldangelo: this logic exists in other places in the codebase. please DRY it out

src/commands/retry.ts
mldangelo, line 1: This whole command is an abomination (before your changes)

src/commands/eval.ts
```diff
@@ -213,11 +213,11 @@ export async function doEval(
 
       logger.info(`Found ${errorResultIds.length} ERROR results to retry`);
 
-      // Delete the ERROR results so they will be re-evaluated when we run with resume
-      await deleteErrorResults(errorResultIds);
-
-      // Recalculate prompt metrics after deleting ERROR results to avoid double-counting
-      await recalculatePromptMetrics(latestEval);
+      // NOTE: We do NOT delete ERROR results here anymore!
```
mldangelo: what does anymore mean? put in a date, a version etc.

src/commands/retry.ts
```diff
@@ -281,12 +385,18 @@ export function setupRetryCommand(program: Command) {
     .option('-v, --verbose', 'Verbose output')
     .option('--max-concurrency <number>', 'Maximum number of concurrent evaluations', parseInt)
     .option('--delay <number>', 'Delay between evaluations in milliseconds', parseInt)
+    .option('--share', 'Share results to cloud (auto-shares when cloud is configured)')
```
mldangelo, line 351: IMHO we should have the same args here that we use in eval, and maybe we can just share code between the two. We can refactor in the future - no action necessary now

src/commands/retry.ts
```diff
@@ -281,12 +385,18 @@ export function setupRetryCommand(program: Command) {
...
         await retryCommand(evalId, cmdObj);
       } catch (error) {
-        logger.error(`Failed to retry evaluation: ${error}`);
-        process.exit(1);
+        logger.error('Failed to retry evaluation', { error, evalId });
+        logger.info('');
+        logger.info(chalk.yellow('Recovery options:'));
+        logger.info(chalk.yellow('  - Run the same retry command again to continue'));
+        logger.info(chalk.yellow('  - Check API credentials and network connectivity'));
+        logger.info(chalk.yellow('  - Use --verbose for detailed error information'));
```
mldangelo: prefer to use one logger command here with dedent that is multi-line

src/cliState.ts
```diff
@@ -13,6 +13,20 @@ interface CliState {
...
+   * Retry mode is a specialized form of resume mode with these differences:
+   * - `resume` mode skips all completed (testIdx, promptIdx) pairs
+   * - `retryMode` additionally excludes ERROR results from "completed" pairs,
+   *   so they get re-evaluated instead of skipped
+   *
+   * When `retryMode` is true, `resume` should also be true.
+   * The evaluator's `getCompletedIndexPairs()` uses this flag to exclude ERROR
+   * results from the completed set, allowing them to be retried.
+   */
+  retryMode?: boolean;
```
mldangelo, line 28: nice comment

PR #7165, opened 2026-01-21
fix(providers): validate parsed JSON is messages array in parseChatPrompt

yash2998chhabria:
## Summary

- Fixed a bug where `parseChatPrompt` returned valid JSON that wasn't a messages array (e.g., `{"a": "{{prompt}}"}`)
- This caused OpenAI to reject requests with: "Invalid type for 'messages': expected an array of objects, but got an object instead"
- Now validates parsed JSON **and YAML** is a valid messages array before returning
- **Provider-agnostic validation**: Only checks for non-empty array of objects, not requiring `role`

## Problem

When using crescendo strategy with nested JSON prompt templates like:
```yaml
prompts:
  - |
    {"a": "{{prompt}}"}
```

The `parseChatPrompt` function at `src/providers/shared.ts:128` blindly returned any valid JSON without checking if it was actually a chat messages array.

## Solution

Added `isValidMessagesArray()` helper that validates:
- Parsed value is an array
- Array is **non-empty** (empty arrays are not valid message arrays)
- Every element is an object (not a string, number, or nested array)

**Why no `role` requirement?** Different providers use different formats:
| Provider | Format | Has `role`? |
|----------|--------|-------------|
| OpenAI/Azure/Mistral | `[{ role: 'user', content: '...' }]` | ✓ |
| Google AI Studio | `[{ content: '...' }]` | ✗ |
| Gemini | `[{ parts: [...] }]` | sometimes |

By not requiring `role`, we support all provider formats while still rejecting obviously invalid structures like `{"a": "value"}`.

If validation fails for either JSON or YAML, returns `defaultValue` which properly wraps the prompt in a user message.

**Why empty arrays are rejected:** An empty array `[]` is not a realistic prompt - nobody would write `prompts: - "[]"` intentionally. If it occurs, it's likely a bug and should be treated as literal text.

## Changes

- `src/providers/shared.ts`:
  - `isValidMessagesArray()`: Validates non-empty array of objects (provider-agnostic)
  - `parseChatPrompt()`: Both JSON and YAML paths use same validation

- `test/providers/shared.test.ts`:
  - Updated tests to reflect provider-agnostic validation
  - Added test for Google AI Studio format (no role)

### Unit tests (all 28 pass)
```bash
npx vitest run test/providers/shared.test.ts
```

Test cases cover:
- JSON object (not array) → returns default
- Nested JSON with prompt placeholder → returns default
- Empty JSON array → returns default
- JSON array without role (Google AI Studio format) → parses correctly ✨
- JSON array with null role → parses correctly (let API validate)
- YAML with role property → parses correctly
- YAML with null role → parses correctly (let API validate)

### Google util tests (all 111 pass)
```bash
npx vitest run test/providers/google/util.test.ts
```

### Lint
```bash
npm run lint
```
Result: ✅ No issues


MrFlounder:
## Question: Is the `role` validation too restrictive?

Looking at how `parseChatPrompt` is used across providers, not all message formats have a `role` property:

| Provider | Default Format | Has `role`? |
|----------|---------------|-------------|
| OpenAI, Azure, Mistral, etc. | `[{ role: 'user', content: prompt }]` | ✓ |
| **Google AI Studio** (line 254) | `[{ content: prompt }]` | ✗ |
| **Gemini** | `[{ parts: [...] }]` | sometimes |
| **Vertex Palm2** (line 737) | `[{ messages: [...] }]` | ✗ |

The new `isValidMessagesArray` check requires every element to have a non-null `role`:

```typescript
parsed.every(msg => ... 'role' in msg && msg.role \!= null)
```

If a user passes `[{"content": "hello"}]` for Google AI Studio, this validation would reject it (no `role`) and return the default value instead of the parsed JSON.

**The original bug** was that `{"a": "{{prompt}}"}` (an **object**) was returned as-is when it should fall back to default.

Would a simpler fix work here - just checking `Array.isArray(parsed)` without requiring `role`? That would fix the reported bug without making OpenAI-centric assumptions about message structure.

yash2998chhabria:
@MrFlounder Great catch! You're right - the `role` validation was too restrictive.

I've updated the validation to be **provider-agnostic**:

```typescript
function isValidMessagesArray(parsed: unknown): boolean {
  return (
    Array.isArray(parsed) &&
    parsed.length > 0 &&
    parsed.every((msg) => msg && typeof msg === 'object' && !Array.isArray(msg))
  );
}
```

This now:
- ✅ Rejects `{"a": "{{prompt}}"}` (object, not array)
- ✅ Rejects `[]` (empty array)
- ✅ Rejects `["hello"]` (array of strings)
- ✅ Accepts `[{"role": "user", "content": "hello"}]` (OpenAI)
- ✅ Accepts `[{"content": "hello"}]` (Google AI Studio)
- ✅ Accepts `[{"parts": [...]}]` (Gemini)

The key insight: we just need to ensure it's a non-empty array of objects - the specific structure (role, content, parts, etc.) is left to the provider API to validate.

PR #7191, opened 2026-01-22
fix(redteam): handle isRefusal in Crescendo before parsing JSON

danenania:
## Summary
- Check for `isRefusal` response before attempting to parse output as JSON in Crescendo provider
- Prevents "Missing keys" error when attack model refuses to generate a prompt

## Test plan
- [x] Manual verification with refusal scenarios
- [x] Existing Crescendo tests pass


mldangelo (approved):
Thank you for adding tests!

PR #7199, opened 2026-01-22
feat(eval): add assert-set hierarchy display in CLI and Web UI

mldangelo:
## Summary

Improves how nested assert-sets are displayed throughout the evaluation output, making it clear which assertions are grouped and how they contribute to the overall pass/fail status.

### CLI Table Output

Hierarchical tree view with visual indicators:

```
┌─────────────────────────────────────────────────────────────────────────────────┐
│ [openai:gpt-4o-mini] Greet the user in French.                                  │
├─────────────────────────────────────────────────────────────────────────────────┤
│ [FAIL] Bonjour ! Comment puis-je vous aider aujourd'hui ?                       │
│        Score: 82% < 90% threshold                                               │
│        ✓ Correctness  ✓ Tone  ✗ Topicality (0.60)  ✓ Performance                │
└─────────────────────────────────────────────────────────────────────────────────┘
```

### Web UI (EvalOutputCell)

- **Color-coded chips**: Green for passed, red for failed assertions
- **Expandable assert-sets**: Chevron indicator opens popover with nested details
- **Split click targets**: Click chip body to filter, click chevron to expand
- **Threshold context**: Popover shows "Score: 0.50 ≥ 0.50 ✓" so users understand why it passed
- **Neutral indicators**: Failed children in passing assert-sets show dash (─) instead of X

### Changes

| Area | Files |
|------|-------|
| Core | `src/assertions/assertionsResult.ts`, `src/table.ts`, `src/types/index.ts` |
| CLI | `src/commands/eval.ts` |
| Web UI | `AssertionChip.tsx` (new), `CustomMetrics.tsx`, `EvalOutputCell.tsx`, `EvaluationPanel.tsx`, `FailReasonCarousel.tsx` |
| Tests | `test/assertions/assertionsResult.test.ts`, `test/table.test.ts` |
| Example | `examples/french-greeting-test/` |

## Test plan

- [x] Lint passes
- [x] Existing CustomMetrics tests pass (16 tests)
- [x] Run `npm run local -- eval -c examples/french-greeting-test/promptfooconfig.yaml --no-cache`
- [ ] Manual verification in Web UI:
  - [ ] Failed assertions show red chips with X icon
  - [ ] Passed assertions show green chips with check icon
  - [ ] Assert-sets show chevron, clicking opens popover
  - [ ] Popover shows threshold comparison and child statuses
  - [ ] Clicking chip body (not chevron) filters by that metric


ghost:
**⏩ Already incorporated tests (25ff694bdde62de0b8253ace6fb3c1817d1bf1c3) [View tests ↗](https://app.usetusk.ai/app/testing-commit-check/50fa4da3-eb45-4d81-9cc5-c6f582f65d32?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

---
<p>

| 81cd55e40f3e97efaf3192a1b76d704940ab8511 | ✅ Generated 64 tests - 64 passed | [Tests](https://app.usetusk.ai/app/testing-commit-check/50fa4da3-eb45-4d81-9cc5-c6f582f65d32?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 7:39AM |
| cf09e8bd693d15e8ce9d16e22f2dae90e226743f | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/1345184f-310a-400a-aa74-39a7a5e497f8?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 8:01AM |
| a3434c2d926dde5cdb21d6c83ac95f52515dccb1 | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/58fb173e-efc2-4924-96bf-9b5512b2eefc?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 8:06AM |
| 93f205f4fdc30843193a6ef2d3e4df837a210b80 | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/f1f2f1a5-7b0b-42dc-a81b-b02c3e72f866?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 8:39AM |
| da86d52bed320d7dd041727e2979683751433a43 | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/a7ba3221-b433-42a6-97f6-08824e1f05d9?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 8:51AM |
| 0208c955199747efd69f3655b8a41161c728516d | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/7a5caad9-bb72-434f-a5b9-595ba6d11175?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 7:15PM |
| 95a28c2f86279e6099ae8348422a71fdbb8468cf | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/572f61c2-7f31-423c-b718-d7eed6a887c4?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 7:19PM |
| 84d2ba6822874291336b3232cdf6fed1b905cdce | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/55b7b587-4356-46ec-b508-bedba434ea4c?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 7:34PM |
| 25ff694bdde62de0b8253ace6fb3c1817d1bf1c3 | ⏩ Already incorporated tests | [Output](https://app.usetusk.ai/app/testing-commit-check/7d3dcef8-7be5-4e12-a176-7e226a40bb39?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jan 22, 2026 8:18PM |

</p>
</details>

mldangelo-oai:
Updated this PR with current `main` and re-reviewed the live head.

What I checked:
- Review threads: no unresolved threads remain; current CodeRabbit/Copilot threads are resolved or stale.
- No migrations were touched, so no DB migration regeneration/reset was needed.
- Re-reviewed the recursive assert-set metadata path, CLI table rendering, Web UI hierarchy rendering, docs/example changes, and config schema changes. I did not find a surviving issue that needed code changes beyond the main merge.

Validation:
- `npx vitest run test/assertions/assertionResult.test.ts test/assertions/assertionsResult.test.ts test/assertions/runAssertions.test.ts test/commands/eval.test.ts test/table.test.ts --sequence.shuffle=false` (180 passed)
- `npm run test:app -- src/pages/eval/components/AssertSetCard.test.tsx src/pages/eval/components/AssertionChip.test.tsx src/pages/eval/components/CustomMetrics.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx src/pages/eval/components/EvaluationPanel.test.tsx src/utils/assertSetThreshold.test.ts --run` (247 passed)
- CLI E2E with isolated `PROMPTFOO_CONFIG_DIR`, `PROMPTFOO_CACHE_PATH`, and `PROMPTFOO_DB_PATH`: nested `assert-set` eval using `echo`, `--table`, `--show-assertions`, and JSON export. Verified the CLI tree and exported `parentAssertSetIndex` hierarchy.
- `npm run local -- validate -c examples/french-greeting-test/promptfooconfig.yaml`
- `npm run l` and `npm run f` (clean; existing warning-only complexity diagnostics in `src/assertions/index.ts` and `src/commands/eval.ts`)
- `npm --prefix site run fetch-stats && npm --prefix site run typecheck`
- `npm run tsc`
- `npm run build`

Pushed head: `94aab0385044656271c2515ffb34e659b0be6c59`. CI is queued/running on the new head.

src/assertions/assertionsResult.ts
```diff
@@ -167,20 +181,75 @@ export class AssertionsResult {
...
+          metadata.assertSetThreshold = assertSetThreshold;
+        }
+      }
+
+      if (weight !== undefined) {
+        metadata.assertSetWeight = weight;
+      }
+
+      flattenedComponentResults.push({
+        ...result,
```
jameshiester-oai, line 235: **[P1] Do not retain recursive subtrees in the flat result**

This spread keeps each aggregate's full `componentResults` tree, and the loop below also emits every descendant into `flattenedComponentResults`. Because nested `AssertionsResult`s already contain flattened descendants, serialized size doubles at every level: my exact-head probe measured 540,428 bytes at depth 10, 2,162,459 at depth 12, and 8,650,583 at depth 14 for only 15 logical nodes. JSON export, DB persistence, and API responses can therefore exhaust memory with a modest valid hierarchy. Choose one canonical representation—e.g. omit `componentResults` from flat emitted copies and reconstruct from `parentAssertSetIndex`—and add a serialized-size regression.

src/app/src/pages/eval/components/assertionHierarchy.ts
```diff
@@ -0,0 +1,22 @@
...
+    metric?: string;
+    threshold?: number;
+    type?: string;
+  };
+  componentResults?: Array<AssertionHierarchyResult | null | undefined>;
+  metadata?: {
+    assertSetMetric?: string;
+    assertSetThreshold?: number;
+    isAssertSet?: boolean;
+    parentAssertSetIndex?: number;
```
jameshiester-oai, line 17: **[P1] Keep the hierarchy view model aligned with emitted metadata**

The backend now emits `childCount`, and the new `AssertSetCard.test.tsx:75` fixture supplies it, but this local metadata type omits the field. Vitest transpiles the tests, so focused tests pass; the production `tsc -b` does not. `npm run build` and the Build-on-Node-26, smoke, and share CI paths all fail with TS2353 on that fixture. Add `childCount?: number` (and keep this view model aligned with the hierarchy fields the UI accepts), or remove the fixture field only if it is intentionally outside this UI contract.

src/types/index.ts
```diff
@@ -735,12 +728,41 @@ export const AssertionSchema = z.object({
...
+  threshold?: number;
+  config?: Record<string, any>;
+}
+
+export type AssertionOrSet = Assertion | AssertionSet;
+
+export const AssertionSetSchema: z.ZodType<AssertionSet> = z.object({
+  type: z.literal('assert-set'),
+  // Sub assertions to be run for this assertion set
+  assert: z.array(z.lazy(() => AssertionOrSetSchema)),
```
jameshiester-oai, line 745: **[P2] Bound the entire recursive assertion tree**

The existing DoS check counts only `test.assert.length`, so nesting bypasses it. An exact-head probe validated one outer set containing 10,001 leaf assertions despite `MAX_ASSERTIONS_PER_TEST = 10000`; chains through depth 1,500 also validated, while depth 2,000 threw a raw `RangeError` from recursive parsing. Run an iterative preflight over the whole tree to enforce both total-node and depth limits before invoking recursive Zod/`parseAssertion`, and return `AssertValidationError` rather than allowing stack overflow.

src/assertions/assertionsResult.ts
```diff
@@ -167,20 +181,75 @@ export class AssertionsResult {
...
+        | { threshold?: number; weight?: number }
+        | undefined;
+      const weight = result.assertion?.weight ?? assertionSetMetadata?.weight;
+      const metadata = { ...result.metadata };
+
+      if (options.parentAssertSetIndex !== undefined) {
+        metadata.parentAssertSetIndex = options.parentAssertSetIndex;
+      }
+
+      if (childResults.length > 0) {
```
jameshiester-oai, line 218: **[P2] Mark only actual assertion sets as assertion sets**

Nonempty `componentResults` is not an assert-set identity: documented JavaScript, Python, and Ruby assertions may return arbitrary composite `GradingResult`s. A probe with a passing `javascript` aggregate and one failed diagnostic component gets `isAssertSet: true`, `childCount: 2`, and the failed component is rendered as “not required.” That invents threshold semantics and also switches EvaluationPanel into assert-set mode. Gate this metadata on the actual set marker (`metadata.assertionSet` or another explicit kind), while representing generic composite results without claiming they are assert-sets.

src/app/src/pages/eval/components/EvaluationPanel.tsx
```diff
@@ -264,6 +354,14 @@ function AssertionResults({ gradingResults }: { gradingResults?: GradingResult[]
     return null;
   }
 
+  if (hasHierarchyMetadata(gradingResults)) {
```
jameshiester-oai, line 357: **[P2] Preserve assertion diagnostics in hierarchy mode**

As soon as any result has hierarchy metadata, this early return bypasses the existing rows that render assertion type, raw/rendered/template value, `graderOutputs`, expandable content, and copy controls. The new hierarchy renderer shows only metric, score, and reason, including for unrelated standalone assertions in the same eval. A Vite SSR comparison confirmed the same `llm-rubric` value and grader output are present in flat mode and absent in hierarchy mode. Reuse the rich row/details renderer inside the tree or add equivalent diagnostics to the cards, with a hierarchy regression covering rendered values and grader outputs.

examples/french-greeting-test/promptfooconfig.yaml
```diff
@@ -0,0 +1,31 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: "Require a friendly French greeting that includes 'Bonjour' and stays within budget (either cheap or fast)."
+prompts:
+  - 'Greet the user in French.'
+providers:
+  - openai:chat:gpt-5.4-mini
+
+tests:
+  - type: assert-set
```
jameshiester-oai, line 9: **[P2] Put the outer assert-set inside the test's assert array**

This object is a test case, so `type` is an additional property rather than the intended outer assertion. The generated JSON schema rejects `/tests/0` for `additionalProperty: type`; runtime parsing tolerates/strips it, treats `threshold: 0.8` as the test threshold, and leaves `contains`, `llm-rubric`, `answer-relevance`, and only the inner `Performance` set as four top-level assertions. The shipped example therefore does not demonstrate recursive hierarchy as its README claims. Wrap the intended group as `tests: - assert: - type: assert-set ...` and retain the desired outer threshold there.

src/table.ts
```diff
@@ -23,28 +199,58 @@ export function generateTable(
...
+        const assertionSummary = generateAssertionSummary(output);
+        if (assertionSummary) {
+          cellContent += `\n       ${assertionSummary}`;
+        }
+      }
+
+      rowCells.push(cellContent);
+
+      if (showAssertions && output.gradingResult?.componentResults?.length) {
+        assertionTables.push(generateAssertionTable(output));
```
jameshiester-oai, line 241: **[P2] Label each assertion detail table with its source result**

Only the rendered table string is retained here, then every string is concatenated under one `Assertion Details:` heading. In a two-row echo eval, this produced two consecutive detail tables with no row vars, test label, prompt/provider, output status, or stable ID; skipped outputs make even ordinal inference unreliable. Collect `{context, table}` and prefix each table with the corresponding row/test plus prompt/provider identity so users can attribute a detailed failure.

src/assertions/assertionsResult.ts
```diff
@@ -95,7 +95,21 @@ export class AssertionsResult {
   }) {
     this.totalScore += result.score * weight;
     this.totalWeight += weight;
-    this.componentResults[index] = result;
+
+    // Attach metric to result if provided and result doesn't already have one
+    // This ensures assert-sets and other aggregated results have proper metric names
+    let resultToStore = result;
+    if (metric && !result.assertion?.metric) {
```
jameshiester-oai, line 102: **[P2] Match component results by the rendered metric**

`metric` is already the rendered name, but this condition refuses to store it when the handler result carries the original metric template. With `metric: '{{metricName}}'` and `metricName=RenderedMetric`, the exact-head result has `namedScores.RenderedMetric` while `componentResults[0].assertion.metric` remains `{{metricName}}`. `buildMetricResultLookup` keys by the latter, so the new `RenderedMetric` chip misses its pass/fail data and falls back to neutral styling/raw labels. Preserve the original template separately if needed, but expose the rendered metric consistently to display consumers.

src/app/src/pages/eval/components/AssertionChip.tsx
```diff
@@ -0,0 +1,231 @@
...
+        onKeyDown: handleChipKeyDown,
+        role: 'button' as const,
+        tabIndex: 0,
+      }
+    : {};
+
+  // Main chip content
+  const chipMainContent = (
+    <>
+      {passed ? (
```
jameshiester-oai, line 104: **[P2] Expose chip status and action to assistive technology**

The mapped-result path replaces the prior button's explicit “Filter by …” label with a `role=button` whose accessible content is only the metric and optional numeric score; both pass/fail icons are `aria-hidden`. Child pass/fail/neutral icons are hidden the same way. Keyboard activation is fixed, but a screen-reader user still cannot determine the color-coded status or that activating the chip filters results. Use a real button or accessible label/sr-only text that includes the action and `Passed`/`Failed`/`Not required` state, and cover those names in tests.

src/app/src/pages/eval/components/AssertSetCard.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+      {/* Children */}
+      {hasChildren && isExpanded && (
+        <div id={childrenId} className="border-t border-border bg-background/50">
+          {children.map((child, index) => {
+            const nestedChildren = getChildResults(child);
+            if (child.metadata?.isAssertSet || nestedChildren.length > 0) {
+              return (
+                <div key={index} className="ml-3 border-l-2 border-border py-2 pl-5 pr-3">
+                  <AssertSetCard
```
jameshiester-oai, line 165: **[P2] Propagate optional status into nested set cards**

Direct leaf rows receive `parentPassed`, but recursive set cards do not. In a reproduced hierarchy where an outer 0.5-threshold set passes, its required sibling passes, and the nested set fails, the nested card chooses a red border/X even though that entire branch was not required; the CLI correctly renders the same child group as neutral. Pass the optional/ancestor context into recursive cards and neutralize the nested header/status while preserving its internal failure details.

jameshiester-oai (commented):
Reviewed base `c54a30668ad8319d76c20ae96e6680ad6c51a2c6` through head `0e957f6d0776691f0337dc76b4d0691385cd6efe` (full PR scope; no prior authenticated completion marker).

## Verdict

The recursive aggregation order and basic CLI/UI hierarchy paths work in focused tests, but this head has two P1 blockers and eight P2 regressions. The most serious runtime issue is exponential result serialization: 15 logical nodes produce an 8.65 MB JSON result. The production app build also fails at the new hierarchy view-model type. GitHub reports the PR as conflicting with current `main` (`4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6`).

## Verified findings

### P1

- Recursive flattening retains every node's nested `componentResults` while also appending descendants to the flat list. Serialized size doubles per nesting level: depth 10 was 540,428 bytes; depth 12 was 2,162,459; depth 14 was 8,650,583 for only 15 logical nodes.
- The local hierarchy view model omits `childCount`, which the new test fixture uses. `npm run build` and the relevant CI build/smoke/share jobs fail with TS2353 at `AssertSetCard.test.tsx:75`.

### P2

- Recursive schemas bypass the existing 10,000-assertion DoS bound and have no depth guard: a set containing 10,001 leaves validates, while depth 2,000 throws a raw `RangeError`.
- Any custom JavaScript/Python/Ruby `GradingResult` with `componentResults` is marked as an assert-set, producing false threshold and “not required” semantics.
- Hierarchy mode replaces the rich EvaluationPanel rows and drops assertion type/value/template, grader outputs, and copy controls for both grouped and unrelated standalone assertions.
- The new example puts `type: assert-set` on a test case. The generated schema rejects it, runtime parsing strips it, and the advertised outer nested set is never created.
- `--show-assertions` concatenates bare tables without row/test/provider identity, so multi-row or multi-provider details cannot be attributed to a result.
- Templated metric names leave `namedScores` keyed by the rendered name while `componentResults` retains the raw template, so the new status-chip lookup misses and falls back to neutral styling.
- The colored metric chips hide all status icons from accessibility APIs and expose neither pass/fail/neutral state nor the “filter by metric” action in the accessible name.
- A failed nested set under a passing thresholded parent remains red in EvaluationPanel because optional-parent state is propagated only to direct leaf rows.

## Existing feedback and simplification

All 15 existing threads are resolved; none is unresolved on this head. I rechecked the earlier recursive-flattening, CLI-recursion, threshold-label, example-format, and keyboard-access threads and did not repeat them. The current serialization and accessible-status findings are distinct consequences left after those fixes.

Use one canonical hierarchy representation instead of retaining both recursive and flat copies. A bounded iterative tree preflight plus a shared hierarchy view model would also remove repeated interpretation logic across the CLI, metric chips, and EvaluationPanel, and would have prevented the app type drift.

## Validation performed

- Focused backend changed-surface run: 5 files / 199 tests passed.
- Focused frontend changed-surface run: 6 files / 243 tests passed.
- Validator/schema run: 2 files / 410 tests passed; three independent specialist runs also passed overlapping 91, 69, and 114-test subsets.
- `npm run tsc` passed; `npm run build` exited 1 on the reproducible `childCount` TS2353.
- `npm run architecture:check`, targeted Biome (two pre-existing complexity warnings), documentation Prettier, and `git diff --check` passed.
- A credential-free two-row echo eval with `--show-assertions --no-cache` produced the intended 1 pass / 1 fail and reproduced two unlabeled detail tables. Exported JSON had no execution errors and preserved the expected 5-node and 3-node hierarchies.
- Direct probes reproduced exponential serialization, the count/depth validation bypass, custom-composite misclassification, dynamic metric mismatch, and outer-pass/inner-fail hierarchy.
- The shipped example failed its own generated JSON schema and its exported result contained only the inner `Performance` set. That probe unexpectedly reached the remote grading path for its model-graded assertions (932 assertion tokens); no provider response text is quoted here.

CI classification: Build on Node 26, smoke, share, and concurrent redteam-build paths are red because of the app type error; Node 20/24 build jobs were fail-fast affected. Backend (19,682), Web UI (4,894), and site (101) tests passed before unrelated Codecov GPG public-key failures. Style, docs build, integration, supported-platform test jobs, CodeQL, code scan, and title checks passed.

Limitations: the requested PR Review Toolkit command is not exposed as a callable skill in this environment, so I completed its correctness, tests, docs, error-handling, type-design, and simplification lenses manually and reconciled three independent read-only specialist reviews. Live Promptfoo browser QA was unavailable: port 3000 serves an unrelated OWASP Juice Shop instance and port 15500 is closed, so I did not replace either process; focused Vitest and Vite SSR probes covered the changed UI, but responsive layout, popover positioning, and browser console/network behavior remain untested. I did not merge or rebase the current base.

PR #7236, opened 2026-01-24
fix(redteam): use defaultTest provider as fallback for attack generation

Ebonsignori:
## Summary

Closes: https://linear.app/promptfooo/issue/ENG-1597/redteam-default-provider-fallback-issue

Fixes a bug where redteam attack generation ignores the `defaultTest.options.provider` configuration and defaults to OpenAI, even when users have configured a different provider.

## Problem

When users configure their grading provider via `defaultTest.options.provider` (e.g., `vertex:gemini-2.5-flash`) without explicitly setting `redteam.provider`, the attack generation phase would still try to use OpenAI

## Solution

Modified `RedteamProviderManager.getProvider()` in `src/redteam/providers/shared.ts` to check the `defaultTest` configuration as a fallback before defaulting to OpenAI:

1. `cliState.config.redteam.provider` (explicit redteam provider)
2. `cliState.config.defaultTest.provider` (fallback)
3. `cliState.config.defaultTest.options.provider` (fallback)
4. OpenAI default (last resort)

## Testing

- Added unit tests for the new fallback behavior
- Verified fix with `PROMPTFOO_DISABLE_REMOTE_GENERATION=true` to force local provider selection

### DIFF
--- a/test/redteam/index.test.ts
+++ b/test/redteam/index.test.ts
@@ -3248,19 +3248,21 @@ describe('Language configuration', () => {
         // Mock the provider loading to return a mock provider for test generation
         // This avoids the loadApiProviders error while still testing strategy config
         const mockProvider = { id: () => 'mock-provider', callApi: vi.fn() };
-        const providerSpy = vi.spyOn(
-          await import('../../src/redteam/providers/shared'),
-          'redteamProviderManager',
-          'get',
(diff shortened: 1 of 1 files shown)


mldangelo:
Going to merge this into the release that is about to go out. I am sorry for merging your pr @Ebonsignori

PR #7287, opened 2026-01-26
chore(examples): simplify and modernize custom-provider example

JustinBeckwith:
## Summary

Simplifies and modernizes the `custom-provider` example by removing complexity while updating to current best practices.

### Changes

**Simplification:**
- Removed Vercel AI SDK demo (previously `vercelAiProvider.js`)
- Removed schema validation example (`schemaValidation.js`)
- Removed `package.json` dependencies - example now works with zero npm packages
- Removed duplicate `.cjs` file
- Streamlined README from 434 → 204 lines, focusing on core concepts

**Modernization:**
- Converted to pure ESM (`import`/`export` syntax)
- Updated OpenAI model: `gpt-4o-mini` → `gpt-5.2`
- Updated parameter: `max_tokens` → `max_completion_tokens` (required for gpt-5.2)
- Expanded documentation with:
  - Quick start guide
  - Code walkthrough
  - Expected output examples
  - Configuration options
  - Troubleshooting section

### Result

A cleaner, more focused example that demonstrates custom provider fundamentals without external dependencies.

**Before:** 8 files, required npm install, multiple examples  
**After:** 5 files, zero dependencies, single clear example

### Testing

```bash
npm run local -- eval -c examples/custom-provider/promptfooconfig.yaml --env-file .env --no-cache
```

✓ 4 passed, 0 failed (100%)  
Total Tokens: 94 (62 prompt, 32 completion)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,6 +33,7 @@
         "@opentelemetry/sdk-trace-node": "^2.5.0",
         "@opentelemetry/semantic-conventions": "^1.39.0",
         "@types/ws": "^8.18.1",
+        "ai": "^6.0.59",
         "ajv": "^8.17.1",
         "ajv-formats": "^3.0.1",
         "async": "^3.2.6",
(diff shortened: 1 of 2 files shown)


examples/custom-provider/customProvider.js
```diff
@@ -16,7 +15,7 @@ class CustomApiProvider {
 
   async callApi(prompt) {
     const body = {
-      model: 'gpt-4.1-mini',
+      model: 'gpt-4o-mini',
```
mldangelo: Description says the old model id is invalid. See: https://platform.openai.com/docs/models/gpt-4.1-mini

we should use 5.2 here instead

examples/custom-provider/customProvider.js
```diff
@@ -1,5 +1,4 @@
-// const promptfoo = require('../../dist/src/index.js').default;
-const promptfoo = require('promptfoo').default;
```
mldangelo, line 2: IMHO we should have both types of examples

examples/custom-provider/package.json
```diff
@@ -0,0 +1,18 @@
+{
```
mldangelo: Can you move this to a different example? 

This folder should be simple and there's no need for a package.json in it.

examples/custom-provider/promptfooconfig-vercel-ai.yaml
```diff
@@ -0,0 +1,15 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Custom provider using Vercel AI SDK with structured output
+
+prompts:
+  - '[{"role": "system", "content": "You are a helpful assistant that provides structured responses."}, {"role": "user", "content": "{{question}}"}]'
```
mldangelo: this is not standard syntax

examples/custom-provider/package.json
JustinBeckwith: So this is what kicked off the whole PR.  I noticed the upgrade for the `ai` package, grepp'd, and it's not used anywhere in the actual `promptfoo` package, just in this example.  Is the design here that all dependencies for examples also get installed with the top level module?  This is a pretty key thing to understand before I address anything else 😆

JustinBeckwith: I think I see better what you're saying now - we add a vercel AI provider example that has the deps, and keep this one really slim with no dependencies?

PR #7301, opened 2026-01-27
fix(redteam): handle health check timeout to avoid blocking generation

danenania:
## Summary
- `checkRemoteHealth` checked `error.name === 'TimeoutError'` but `fetchWithTimeout` throws a plain `Error` — so the graceful timeout handling was dead code since the health check was introduced
- Health check timeouts now correctly return `status: 'OK'` and proceed with generation instead of failing the entire plugin
- Fixed by matching on `error.message.includes('timed out')` instead

### DIFF
--- a/test/util/apiHealth.test.ts
+++ b/test/util/apiHealth.test.ts
@@ -151,9 +151,7 @@ describe('API Health Utilities', () => {
     });
 
     it('should handle timeout error from fetchWithTimeout gracefully', async () => {
-      mockedFetchWithTimeout.mockRejectedValueOnce(
-        new Error('Request timed out after 5000 ms'),
-      );
+      mockedFetchWithTimeout.mockRejectedValueOnce(new Error('Request timed out after 5000 ms'));
 
       const result = await checkRemoteHealth('https://test.api/health');
       expect(result.status).toBe('OK');


src/util/apiHealth.ts
```diff
@@ -93,7 +93,9 @@ export async function checkRemoteHealth(url: string): Promise<HealthResponse> {
     }
 
     // If it's a timeout error, return a softer message
-    if (error.name === 'TimeoutError') {
+    // fetchWithTimeout throws a plain Error with "timed out" in the message,
```
faizanminhas: Why not make fetchWithTimeout throw a timeout error?

danenania: That would be a cleaner fix, but fetchWithTimeout has a ton of callers. It's hard to be sure that nothing depends on the current behavior

danenania:
@faizanminhas @mldangelo I looked at this more and we don't actually have a `TimeoutError` in the codebase that we use, except in the context of MCP (it extends MCPError).

We are just using `new Error("Request timed out after...")` everywhere else we have timeouts. So I think probably the original mistake was actually the `error.name == 'TimeoutError'` check, and that should be removed in favor of just checking the message.

I wouldn't be against adding a new `TimeoutError` class, but I feel like that should be a separate PR that replaces other timeouts as well for consistency.

PR #7311, opened 2026-01-27
chore: replace prompt-injection with jailbreak variants

typpo:
## Summary
This PR updates documentation across all red teaming guides to replace the deprecated `prompt-injection` strategy with appropriate `jailbreak` strategy variants (`jailbreak:meta` and `jailbreak:composite`). This reflects changes to the available red teaming strategies in the platform.

## Key Changes
- **Replaced `prompt-injection` strategy** with `jailbreak:composite` or `jailbreak:meta` across all red team configuration examples
- **Replaced standalone `jailbreak` strategy** with `jailbreak:meta` for consistency and clarity
- **Updated documentation files** covering:
  - Testing guardrails guide
  - Integration guides (Burp)
  - Provider documentation (OpenAI ChatKit)
  - Red team frameworks (EU AI Act, Foundation Models, GDPR, ISO 42001, MITRE ATLAS, OWASP)
  - LLM-specific guides (Agents, Supply Chain)
  - Troubleshooting and best practices guides

## Implementation Details
- The `jailbreak:meta` strategy replaces the generic `jailbreak` strategy for single-shot optimization of safety bypass techniques
- The `jailbreak:composite` strategy is used where `prompt-injection` was previously specified, as it combines multiple techniques for enhanced effectiveness
- All configuration examples now use the updated strategy names to ensure consistency and prevent user confusion
- No functional changes to the red teaming capabilities; this is purely a documentation update to reflect the current API

## Impact
Users following these documentation examples will now use the correct, currently-supported strategy identifiers when configuring their red team tests.

https://claude.ai/code/session_01XtEGxsL8XTbEeFdMgStCBa


site/docs/guides/testing-guardrails.md
```diff
@@ -666,11 +666,9 @@ redteam:
     - harmful:self-harm
     - harmful:sexual
     - harmful:violence
-    - prompt-injection
-    - jailbreak
   strategies:
-    - id: prompt-injection
-    - id: jailbreak
+    - id: jailbreak:composite
```
mldangelo-oai, line 670: [P1] Preserve `prompt-injection`'s local template semantics

The deprecated `prompt-injection` ID is an alias for `jailbreak-templates`; both call `addInjections` and apply the local static template set. Replacing it with `jailbreak:composite` silently switches readers to a remote-generation strategy that is marked `STRATEGIES_REQUIRING_REMOTE` and defaults to five generated variants per case. That can send attack inputs to the remote service and multiply target cost instead of performing a naming-only migration. Use `jailbreak-templates` for every replaced `prompt-injection` entry. Current main's merged #9931 already applies that correction across the affected docs, so this branch should be retired rather than updated in place.

mldangelo-oai (commented):
Comment-only review at exact head `6e0ddd77991097acc2c9c76401a0a9da5dcdec51`. Retire this exact branch rather than rebasing it. It is 510 commits behind current `main`, and the exact merge tree conflicts in 13 of the 16 edited pages because #9931 (`cfa008772`) has already landed the broader 44-file migration from deprecated `prompt-injection` references.

The merged migration also establishes the correct replacement: `prompt-injection` is a compatibility alias for `jailbreak-templates`, and both call the same local static-template implementation. This branch instead maps most of those examples to `jailbreak:composite`, which is a different remote-generation strategy with a default fan-out of five. That changes data handling, cost, test volume, and attack semantics rather than renaming the deprecated ID. The inline finding below is representative of the same problem across the framework, agent, supply-chain, Burp, and troubleshooting pages.

The only potentially reusable slice is the separate cleanup from bare `jailbreak` to its explicit `jailbreak:meta` alias. If that wording migration is still desired, do it as a small current-main follow-up after confirming the product naming; do not carry forward this branch's `prompt-injection` substitutions.

All 37 published checks on this historical head are green and `git diff --check` passes. Earlier typo, duplicate-entry, and plugin/strategy-placement comments are fixed or stale at the exact head. The remaining tracing wording thread is non-blocking relative to the supersession. No branch changes or pushes were made.

PR #7314, opened 2026-01-27
feat(redteam): add email validation to generate command

typpo:
## Summary
- Add email requirement to `promptfoo redteam generate` command
- Matches existing behavior in the eval command
- Email is validated before generation starts when using remote generation

## Behavior
- Only required when `shouldGenerateRemote()` returns true
- Skipped when `PROMPTFOO_DISABLE_REMOTE_GENERATION=true` or `PROMPTFOO_DISABLE_REDTEAM_REMOTE_GENERATION=true`
- CI environments automatically use `ci-placeholder@promptfoo.dev`
- Uses the same validation flow as eval (prompts until valid email obtained)

## Test plan
- [ ] Run `npm run local -- redteam generate -c examples/redteam-example/promptfooconfig.yaml`
- [ ] Verify email prompt appears before generation starts
- [ ] Test with `PROMPTFOO_DISABLE_REMOTE_GENERATION=true` to verify email is skipped

### DIFF
--- a/src/globalConfig/accounts.ts
+++ b/src/globalConfig/accounts.ts
@@ -213,10 +213,21 @@ export async function promptForEmailUnverified(): Promise<{ emailNeedsValidation
     await telemetry.record('feature_used', {
       feature: 'promptForEmailUnverified',
     });
+
+    // Display a styled prompt box
+    const border = '─'.repeat(TERMINAL_MAX_WIDTH);
+    logger.info('');
+    logger.info(chalk.cyan(border));
+    logger.info(chalk.cyan.bold('  Email Verification Required'));
+    logger.info(chalk.cyan(border));
+    logger.info('');
+    logger.info('  Red team scans require email verification to continue.');
+    logger.info('');
+
     const emailSchema = z.email();
     try {
       email = await input({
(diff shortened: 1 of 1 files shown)


src/redteam/commands/generate.ts
```diff
@@ -273,6 +279,16 @@ export async function doGenerateRedteam(
...
 
+  // Validate email for remote generation
+  if (shouldGenerateRemote()) {
+    let hasValidEmail = false;
+    while (!hasValidEmail) {
+      const { emailNeedsValidation } = await promptForEmailUnverified();
+      const res = await checkEmailStatusAndMaybeExit({ validate: emailNeedsValidation });
+      hasValidEmail = res === EMAIL_OK_STATUS;
+    }
+  }
```
faizanminhas: should you add this to `synthesize` instead? I don't think this will run for `redteam run` commands

typpo: `doGenerateRedteam` is called from `redteam run`, and no one else calls `synthesize` except for this function.  I _think_ it makes more sense to keep the non-generation logic out of `synthesize`, unless we want to be really defensive here.

PR #7335, opened 2026-01-28
fix(share): add adaptive chunk retry for large eval uploads

addelong:
## Summary
- Fixes large eval uploads (~5850+ results) failing partway through with "fetch failed" errors
- Root cause: chunk size estimation only samples first 100 results - if later results are larger, chunks become oversized
- Adds adaptive retry logic that splits chunks on 413 or network timeout, retrying each half recursively

## Changes
- `sendChunkOfResults()` now returns a result object instead of throwing, classifying errors as `PAYLOAD_TOO_LARGE`, `NETWORK_TIMEOUT`, or `UNKNOWN`
- New `sendChunkWithRetry()` function recursively splits failed chunks in half until success
- Progress updates happen after each successful sub-chunk during retries

## Test plan
- [x] Added unit tests for adaptive retry logic
- [x] Tests verify chunk splitting on 413 and network timeout
- [x] Tests verify failure when single result is too large
- [x] Tests verify no retry on unknown errors (500, etc.)
- [x] All existing share tests still pass


src/share.ts
```diff
@@ -272,11 +396,19 @@ async function sendChunkedResults(
 
   // Determine how many results per chunk
   const TARGET_CHUNK_SIZE = 0.9 * 1024 * 1024; // 900KB in bytes
-  const estimatedResultsPerChunk =
-    getEnvInt('PROMPTFOO_SHARE_CHUNK_SIZE') ??
-    Math.max(1, Math.floor(TARGET_CHUNK_SIZE / largestSize));
+  const envChunkSize = getEnvInt('PROMPTFOO_SHARE_CHUNK_SIZE');
```
Ebonsignori, line 399: No default value / fallback?

Ebonsignori (approved):

src/share.ts
addelong: down 3 lines :)

mldangelo:
> How do you clear the reviews properly

@josephjamesmorris700-hub do you need help with an issue or a PR?

PR #7353, opened 2026-01-29
fix(scheduler): preserve provider id() method when wrapping with rate…

jameshiester:
… limiting

The spread operator doesn't copy prototype methods, only own properties. When wrapProviderWithRateLimiting wraps a provider class instance, the id() method (which is on the prototype) was lost, causing:
  TypeError: redteamProvider.id is not a function

This affected iterative, crescendo, and other redteam strategies that call TokenUsageTracker.trackUsage(provider.id(), ...).

The fix explicitly delegates id() to the original provider.

Includes a minimal repro example in examples/iterative-bug-repro/.

### DIFF
--- a/test/scheduler/providerWrapper.test.ts
+++ b/test/scheduler/providerWrapper.test.ts
@@ -56,6 +56,24 @@ describe('providerWrapper', () => {
       expect(wrappedProvider.config).toEqual({ apiKey: 'test-key' });
     });
 
+    it('should preserve id() method from class prototype', () => {
+      // This tests the specific bug where spread operator doesn't copy prototype methods.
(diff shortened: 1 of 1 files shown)


examples/iterative-bug-repro/promptfooconfig.yaml
```diff
@@ -0,0 +1,17 @@
+# Minimal config to reproduce the redteamProvider.id() bug (fixed in this PR)
```
mldangelo, line 1: move this to the smoke tests

MrFlounder:
added test, will make separate pr for smoke test to get this out right now

PR #7386, opened 2026-01-30
fix(server): add authentication to http-generator endpoint

mldangelo:
## Summary

- Fixes 403 error when using the "Auto-fill from Example" import feature in the redteam HTTP endpoint configuration UI
- The cloud API endpoint (`/api/v1/http-provider-generator`) now requires authentication
- Adds API key check with helpful 401 error message if user is not logged in
- Adds `Authorization: Bearer` header to the cloud API request

## Alternative Approach

Instead of requiring users to authenticate via `promptfoo auth login`, we could authenticate the backend server itself with a service account. This would allow the feature to work without requiring individual user authentication.

## Test plan

- [ ] Verify the feature works when logged in via `promptfoo auth login`
- [ ] Verify helpful error message is shown when not logged in

### DIFF
--- a/src/server/routes/providers.ts
+++ b/src/server/routes/providers.ts
@@ -1,7 +1,7 @@
 import { Router } from 'express';
 import { z } from 'zod';
 import { defaultProviders } from '../../constants/defaultProviders';
-import { getEnvString } from '../../envars';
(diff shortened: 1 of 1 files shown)


mldangelo:
Solved on the serverside

PR #7394, opened 2026-01-31
fix(redteam): use word-boundary regex in isBasicRefusal to prevent false positives

danenania:
## Summary
- `isBasicRefusal()` substring matching used `.includes()`, causing `"as an ai"` to match inside non-refusal responses like `"as an airline CEO"` — flagging valid generated prompts as soft refusals and triggering unnecessary provider fallbacks
- Converts `REFUSAL_SUBSTRINGS` to pre-compiled `\b` word-boundary regex patterns so matches only occur on whole words
- Moves `escapeRegExp` from `src/commands/mcp/lib/security.ts` to shared `src/util/text.ts`
- Fixes stray backtick typo in `"I do not engage"`

## Test plan
- [x] Existing `isBasicRefusal` tests pass unchanged
- [x] New tests verify word-boundary matching prevents false positives (e.g. `"as an aide"`, `"not appropriately"`)
- [x] New tests verify real refusals still detected (`"As an AI, I cannot assist"`)
- [x] MCP security tests updated and passing


faizanminhas (approved):
Whoa who knew. We should use a tiny model to determine if it’s a refusal instead 👀

PR #7428, opened 2026-02-02
chore: show prompt ids in `promptfoo show eval`

typpo:
for easier cli usage


MrFlounder (approved):

mldangelo (approved):
Awesome! Should we include a test?

PR #7446, opened 2026-02-02
feat(providers): add HuggingFace chat completion provider

mldangelo:
## Summary

Adds support for HuggingFace's OpenAI-compatible chat completions API via a new `huggingface:chat` provider.

- New `HuggingfaceChatCompletionProvider` class extending `OpenAiChatCompletionProvider` for code reuse
- Auto-detection in `huggingface:text-generation` when `apiEndpoint` contains `/v1`
- Supports `chatCompletion: true/false` config to explicitly control format
- Maps HuggingFace parameters (`max_new_tokens`) to OpenAI format (`max_tokens`)
- Registry support for `huggingface:chat:<model>` syntax

## Usage

```yaml
providers:
  - id: huggingface:chat:deepseek-ai/DeepSeek-R1
    config:
      temperature: 0.7
      max_new_tokens: 1000

  # Or with text-generation (auto-detects from URL)
  - id: huggingface:text-generation:meta-llama/Llama-3.3-70B-Instruct
    config:
      apiEndpoint: https://router.huggingface.co/v1/chat/completions
```

## Test plan

- [x] Unit tests added for chat completion functionality
- [x] Integration tested with multiple models:
  - DeepSeek-R1
  - Llama-3.3-70B-Instruct
  - Qwen2.5-72B-Instruct
  - Llama-3.1-8B-Instruct (via auto-detect)
- [x] Existing HuggingFace tests still pass
- [x] Documentation updated
- [x] Example added in `examples/huggingface-chat/`

cc @jameshiester - this fixes the issue you reported where the HuggingFace text-generation provider was sending the wrong request format to `/v1/chat/completions`.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -78,8 +78,10 @@
 # DOCUMENTATION
 # ============================================================================
 
-/site/                      @typpo @mldangelo @swarnap      # Marketing site (landing pages)
+/site/                      @typpo @mldangelo               # Marketing site (infra/config)
 /site/docs/                 @typpo @mldangelo               # Technical documentation
+/site/src/pages/**/*.tsx    @swarnap @typpo @mldangelo      # Landing page content
+/site/src/pages/**/*.md     @swarnap @typpo @mldangelo      # Landing page markdown
 /site/blog/                 @swarnap @mldangelo             # Blog posts
 /CONTRIBUTING.md            @mldangelo                      # Contributing guide
 /site/docs/contributing.md  @mldangelo                      # Contributing docs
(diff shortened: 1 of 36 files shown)


jameshiester:
should there be a smoke test?

mldangelo:
@jameshiester Re: smoke test - looking at the existing smoke tests in `test/smoke/`, they use deterministic providers (like `echo`) to test CLI behavior without making real API calls. For HuggingFace, a meaningful smoke test would require real API calls (which needs an API key and has latency).

What we could test in smoke tests without API calls:
- Provider registry loads `huggingface:chat:` correctly (already covered by unit tests in `test/providers/index.test.ts`)
- Config validation accepts the new provider format

What would require an integration test with a real API key:
- End-to-end chat completion responses
- Auto-detection behavior with real endpoints

I've also addressed the review feedback:
- Fixed overly broad `/v1` auto-detection → now checks `/v1/chat` specifically
- Added `sentence-similarity` to error messages  
- Removed unused `createHuggingfaceProvider` factory function
- Fixed example config field ordering

PR #7451, opened 2026-02-03
feat: add --filter-prompts option with MCP alignment

### LINKED ISSUE #7411: Prompt lables and filtering --filter-prompts ?
Please is there any option for `--filter-prompts` , 
filtering based on prompt label. I have not found one between --filter-* options. Closest one seems --filter-pattern but it seems to be "Only run tests whose description matches the regex pattern" 

Thanx a lot

kkahadze:
## What Changed

Added `--filter-prompts` option to filter evaluations by prompt id or label using regex patterns. This feature is available in:
- CLI: `promptfoo eval --filter-prompts <pattern>`
- CLI: `promptfoo redteam run --filter-prompts <pattern>`
- MCP: `run_evaluation` tool with `promptFilter` parameter
- MCP: `redteam_run` tool with `filterPrompts` parameter

The implementation aligns MCP filtering with CLI filtering by delegating to the same underlying `doEval()` code path, while maintaining backwards compatibility for numeric index-based filtering in MCP.

## How to Test

Create a simple config:

```
prompts:
  - id: greeting-morning
    label: morning-greeting
    content: "Say good morning"
  
  - id: greeting-evening
    label: evening-greeting
    content: "Say good evening"
  
  - id: farewell
    label: goodbye
    content: "Say goodbye"

providers:
  - echo: "Hello!"

tests:
  - vars: {}
  ```

Test the filtering:

```
# Filter by label pattern (should run 2 prompts: morning + evening)
promptfoo eval -c test-filter.yaml --filter-prompts "greeting.*" 

# Filter by exact id (should run 1 prompt: morning)
promptfoo eval -c test-filter.yaml --filter-prompts "greeting-morning" 

# No matches (should show warning)
promptfoo eval -c test-filter.yaml --filter-prompts "nonexistent" 
```

## Key Implementation Details
- Filtering logic: Matches both prompt id and label (consistent with --filter-providers)
- MCP backwards compatibility: Numeric strings (e.g., "0", ["0", "2"]) are treated as indices

Fixes #7411

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,7 +80,7 @@
         "opener": "^1.5.2",
         "ora": "^9.2.0",
         "pem": "~1.14.8",
-        "posthog-node": "~5.24.9",
+        "posthog-node": "~5.24.10",
         "protobufjs": "^8.0.0",
         "proxy-agent": "^6.5.0",
         "proxy-from-env": "^1.1.0",
@@ -11622,6 +11622,7 @@
       "version": "1.20.1",
       "resolved": "https://registry.npmjs.org/@posthog/core/-/core-1.20.1.tgz",
       "integrity": "sha512-uoTmWkYCtLYFpiK37/JCq+BuCA/OZn1qQZn5cPv1EEKt3ni3Zgg48xWCnSEyGFl5KKSXlfCruiRTwnbAtCgrBA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "cross-spawn": "^7.0.6"
@@ -35575,15 +35576,24 @@
       "license": "MIT"
     },
     "node_modules/posthog-node": {
-      "version": "5.24.11",
-      "resolved": "https://registry.npmjs.org/posthog-node/-/posthog-node-5.24.11.tgz",
-      "integrity": "sha512-tDXbYyXJyh0oUEo1SumCzmXY0FZNB0avAq0uXMo6o6JinzwY8u5cygqAgUyMDIGG8u0p6tBHq++foqULXaPmiA==",
+      "version": "5.24.13",
+      "resolved": "https://registry.npmjs.org/posthog-node/-/posthog-node-5.24.13.tgz",
+      "integrity": "sha512-JHV2wUG5pkj50aV7XrHOWB7MZjqOsn+SKmbi1Q/Zy/94xvpmRt6mYDrF9Wp0HZTTUczyuqrXqK2eebgKJqlSzg==",
       "license": "MIT",
       "dependencies": {
-        "@posthog/core": "1.20.1"
+        "@posthog/core": "1.20.2"
       },
       "engines": {
         "node": "^20.20.0 || >=22.22.0"
+      }
+    },
+    "node_modules/posthog-node/node_modules/@posthog/core": {
+      "version": "1.20.2",
+      "resolved": "https://registry.npmjs.org/@posthog/core/-/core-1.20.2.tgz",
+      "integrity": "sha512-aQhrUzOHYr0z/bkwDsJ5hXahdh6oyeWdx2/CHwR6vFG3eK07J69lbuGOj+HGOOxJP1eAdNnsk8J0fj1vqRA9+A==",
+      "license": "MIT",
+      "dependencies": {
+        "cross-spawn": "^7.0.6"
       }
     },
     "node_modules/preact": {
--- a/package.json
+++ b/package.json
@@ -279,7 +279,7 @@
     "opener": "^1.5.2",
     "ora": "^9.2.0",
     "pem": "~1.14.8",
-    "posthog-node": "~5.24.9",
+    "posthog-node": "~5.24.10",
     "protobufjs": "^8.0.0",
     "proxy-agent": "^6.5.0",
     "proxy-from-env": "^1.1.0",
--- a/src/app/src/pages/redteam/setup/components/Targets/PolicySuggestionsSidebar.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/PolicySuggestionsSidebar.tsx
@@ -4,7 +4,7 @@ import { Button } from '@app/components/ui/button';
 import { Card, CardContent } from '@app/components/ui/card';
 import { Spinner } from '@app/components/ui/spinner';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@app/components/ui/tooltip';
-import { Plus, Sparkles, ThumbsDown } from 'lucide-react';
+import { Lightbulb, Plus, ThumbsDown } from 'lucide-react';
 import type { PolicyObject } from '@promptfoo/redteam/types';
 
 type PolicySuggestionsSidebarProps = {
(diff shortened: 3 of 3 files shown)

### CODE REVIEW

File: src/commands/eval/filterPrompts.ts [hunk 1]
```diff
@@ -0,0 +1,26 @@
...
+ * @param prompts - Array of prompts to filter
+ * @param filterPromptsOption - Optional regex pattern to match against prompt id or label
+ * @returns Filtered array of prompts whose id or label match the pattern
+ */
+export function filterPrompts(prompts: Prompt[], filterPromptsOption?: string): Prompt[] {
+  if (!filterPromptsOption) {
+    return prompts;
+  }
+
+  const filterRegex = new RegExp(filterPromptsOption);
```
mldangelo: new RegExp(filterPromptsOption) is unguarded - invalid patterns will throw and propagate as an unhandled exception. In the CLI this crashes the process; in the MCP tool it surfaces as an opaque error. Consider wrapping in try-catch and returning a clear error message, or validating the pattern before constructing the regex.

src/commands/mcp/tools/runEvaluation.ts
```diff
@@ -153,19 +152,57 @@ export function registerRunEvaluationTool(server: McpServer) {
...
-        // If we need to filter test cases or prompts, we need to handle this ourselves
-        // since doEval doesn't support these specific filters
-        if (testCaseIndices !== undefined || promptFilter !== undefined) {
+        // Check if promptFilter contains numeric indices (backwards compatibility)
+        const promptFilters = promptFilter
+          ? Array.isArray(promptFilter)
+            ? promptFilter
+            : [promptFilter]
+          : null;
+        const hasNumericPromptFilter = promptFilters && promptFilters.every((f) => /^\d+$/.test(f));
```
mldangelo, line 179: `promptFilters.every((f) => /^\d+$/.test(f))` means mixed input like ["0", "my-prompt"] falls through to the regex path, where "0" silently becomes a regex matching any string containing "0" rather than selecting index 0. Consider either erroring on mixed input or partitioning numeric vs non-numeric filters.

src/commands/mcp/tools/runEvaluation.ts
```diff
@@ -153,19 +152,57 @@ export function registerRunEvaluationTool(server: McpServer) {
...
+
+        // If we need to filter test cases OR handle index-based prompt filtering,
+        // we need to handle this ourselves since doEval doesn't support these
+        if (testCaseIndices !== undefined || hasNumericPromptFilter) {
           // Resolve config to get the test suite first
           const configPaths = configPath ? [configPath] : ['promptfooconfig.yaml'];
           const { config, testSuite } = await resolveConfigs(
```
mldangelo, line 185: This delegates prompt and provider filtering to resolveConfigs, but resolveConfigs calls process.exit(1) if all prompts are filtered out (load.ts:787). In an MCP server context this kills the entire process instead of returning an error to the client.

The old code had explicit createToolResponse error handling for these cases. Consider checking testSuite.prompts.length === 0 after resolveConfigs returns and returning a proper error response, or keeping the manual filtering with error responses for the MCP path.

src/commands/mcp/tools/runEvaluation.ts
```diff
@@ -212,70 +249,8 @@ export function registerRunEvaluationTool(server: McpServer) {
...
-                'run_evaluation',
-                false,
-                undefined,
-                `No providers matched filter: ${filters.join(', ')}. Available providers: ${providers.map((p) => (typeof p.id === 'function' ? p.id() : p.id)).join(', ')}`,
-              );
-            }
-
-            filteredTestSuite.providers = filteredProviders;
-          }
+          // Note: promptFilter and providerFilter are handled by resolveConfigs above
```
mldangelo: This removed the old provider filtering that used escapeRegExp and case-insensitive matching (new RegExp(filters.map(escapeRegExp).join('|'), 'i')). The new path through resolveConfigs → filterProviderConfigs uses new RegExp(filterOption) with no escaping and case-sensitive matching. This is a breaking change for MCP clients that pass literal provider IDs like openai:gpt-4.1 (the dot is no longer escaped) or rely on case-insensitive matching.

src/commands/mcp/tools/runEvaluation.ts
```diff
@@ -376,7 +352,8 @@ export function registerRunEvaluationTool(server: McpServer) {
             cache,
             write,
             share,
-            // Set up provider filtering - doEval supports this natively
+            // Set up filtering - doEval supports regex-based filtering natively
+            filterPrompts: Array.isArray(promptFilter) ? promptFilter.join('|') : promptFilter,
```
mldangelo: Nit: when promptFilter is an array, joining with | produces unexpected regex if individual elements contain metacharacters. The old code used escapeRegExp for the equivalent provider path. Consider whether the same treatment is needed here, or document that this field now expects regex patterns rather than literal strings.

src/util/config/load.ts
```diff
@@ -711,7 +712,18 @@ export async function resolveConfigs(
 
   // Parse prompts, providers, and tests
   // Pass filtered resolved configs to avoid re-reading files
-  const parsedPrompts = await readPrompts(config.prompts, cmdObj.prompts ? undefined : basePath);
+  let parsedPrompts = await readPrompts(config.prompts, cmdObj.prompts ? undefined : basePath);
+
+  // Filter prompts if --filter-prompts option is provided
+  if (cmdObj.filterPrompts) {
+    parsedPrompts = filterPrompts(parsedPrompts, cmdObj.filterPrompts);
```
mldangelo, line 721: This warns but continues when no prompts match the filter. That's fine for CLI, but for MCP callers that go through resolveConfigs, the subsequent process.exit(1) at line 787 is the problem.

src/commands/eval/filterPrompts.ts
kkahadze: I wrapped it in a try catch and returned an error in a2565980734ca7d49099ecb8127399876c22d7df

src/commands/mcp/tools/runEvaluation.ts
kkahadze: I made it now error in b0071c8e61d020a57610f3a196013db763eff1e4 on mixed input since I feel that its more likely that the user made a mistake rather than this being their intention.

src/commands/mcp/tools/runEvaluation.ts
kkahadze: Fixed in a435ddb1a4037f0812591c28bc53dd221ddddb3d by keeping manual filtering for prompts in the MCP path. Now rwe handle filtering locally with proper errors. This prevents process.exit(1) from killing the server.

src/commands/mcp/tools/runEvaluation.ts
kkahadze: Reverted back to the old provider filtering logic in f9ab8f041109db77d27a2565c705679a4f4626b9

src/commands/mcp/tools/runEvaluation.ts
kkahadze: I chose to document it as regex to stay consistent with CLI's --filter-prompts in ec251490aa8cdbd05d220c7946959bd25f769732

src/util/config/load.ts
kkahadze: Changed mcp code to not pass filterPrompts to resolveConfigs in a435ddb1a

PR #7485, opened 2026-02-04
feat(redteam): enable multilingual support for audio/video/image strategies

addelong:
## Summary

- Remove `LANGUAGE_DISALLOWED_STRATEGIES` restriction that forced English-only for audio, video, image, and math-prompt strategies
- Update audio strategy to read language from each test case's metadata instead of a single config value

This restriction was added when multilingual was a post-processing strategy applied after test generation. Now that languages are configured at the plugin level (before strategies), these strategies can properly handle non-English content.

## Test plan

- [ ] Generate redteam tests with `language: Japanese` and `strategies: [audio]`
- [ ] Verify Japanese text prompts are converted to Japanese audio (not English)
- [ ] Verify the `[Language Override]` log message no longer appears

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -143,7 +143,7 @@
         "drizzle-kit": "^0.31.8",
         "ibm-cloud-sdk-core": "^5.4.5",
         "ink-testing-library": "^4.0.0",
-        "knip": "^5.82.1",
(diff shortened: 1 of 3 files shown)


mldangelo (approved):
Cool!

PR #7508, opened 2026-02-05
fix: improve clarity of which model is being used in logging redteams

kkahadze:
## Summary                                                                                                                                                                                                
                                                                            
  - Adds a `purpose` parameter to `loadRedteamProvider()` to clarify log messages
  - Logs now correctly show "Using grading provider" vs "Using redteam provider"
  - Previously, loading the grading provider would misleadingly log "Using redteam provider"

  Before:
  Using redteam provider: [OpenAI Provider gpt-4o-mini]
  Using redteam provider: [OpenAI Provider gpt-4o-mini]  ← actually grading

  After:
  Using redteam provider: [OpenAI Provider gpt-4o-mini]
  Using grading provider: [OpenAI Provider gpt-4o-mini]  ← clear

  ## Test plan

  - [x] Existing tests pass (`npm run test -- test/redteam/providers/shared.test.ts`)
  - [x] Verified with manual test using `--grader openai:gpt-4o-mini`


MrFlounder (approved):
Better than now but we should rename the function or move it out to a different place.. rename it to something like: loadCustomInferenceProvider or something

but this pr is good

PR #7565, opened 2026-02-08
fix: Add missed await for async operations

schipiga:
Hi!

My AI assistant highlighted that looks like some async operations have missed `await` keyword. And I'm inclined to agree with him, because `await` is used in [similar case](https://github.com/promptfoo/promptfoo/blob/e20ae6ad3f780b5024f22ee76b4cac8218af604c/src/models/modelAudit.ts#L170-L175) already in project. Also as I know db interaction involves I/O operations which are usually async in nodejs. And as I see in drizzle docs, it uses `await` widely:
- https://orm.drizzle.team/docs/get-started-sqlite
- https://orm.drizzle.team/docs/transactions

Could you please check if my pr makes sense?

### DIFF
--- a/src/blobs/index.ts
+++ b/src/blobs/index.ts
@@ -140,7 +140,7 @@ export async function recordBlobReference(
   }
 
   const db = getDb();
-  const existing = db
+  const existing = await db
     .select({ id: blobReferencesTable.id })
     .from(blobReferencesTable)
     .where(
@@ -155,7 +155,8 @@ export async function recordBlobReference(
     return;
   }
 
-  db.insert(blobReferencesTable)
+  await db
+    .insert(blobReferencesTable)
     .values({
       id: randomUUID(),
       blobHash: hash,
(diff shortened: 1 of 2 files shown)


test/commands/import.test.ts
```diff
@@ -104,7 +104,7 @@ describe('importCommand', () => {
       await program.parseAsync(['node', 'test', 'import', sampleFilePath]);
 
       // Check if the import process exited with an error
-      expect(process.exitCode).toBeUndefined();
+      expect(process.exitCode).toBe(1);
```
schipiga: ~according to comment above `exiting with an error` usually means a numeric exit code, but not undefined. I seem, it started to work properly after async/await fixing ¯\_(ツ)_/¯~

although seems I didn't understood correctly. appreciate help with tests understanding

schipiga:
ok, seems it's useless pr due to [sync](https://dev.to/lovestaco/understanding-better-sqlite3-the-fastest-sqlite-library-for-nodejs-4n8#:~:text=Why%20Better%2DSQLite3%20Is%20Different,Performance) of `better-sqlite3`

PR #7585, opened 2026-02-09
chore: cosmetic updates to policy suggestions

typpo:


will-holley:
@faizanminhas cargo cult approval

PR #7587, opened 2026-02-09
chore: make http tabs look more tabby

typpo:
Found the strip of buttons confusing/unclear what below is included in the tab

# Before
image

# After
image


faizanminhas:
I think the images are broken. @claude can you take a screenshot of the new tabs?

faizanminhas (changes requested):
tbh I think the new tabs are really bulky and kinda ugly. did you demo promptfoo and encounter someone who was confused by these tabs?

typpo:
re-uploaded screenshots @faizanminhas.  I'm also fine just keeping the tabs as-is but adding the box around the content (which is mainly what I wanted)

faizanminhas:
> re-uploaded screenshots @faizanminhas. I'm also fine just keeping the tabs as-is but adding the box around the content (which is mainly what I wanted)

I support the box!

typpo:
done - just the box

PR #7590, opened 2026-02-09
feat(fetch): add Cloudflare 524 timeout to transient error retry

minhle1291:
## Summary

- Add support for Cloudflare's 524 "A Timeout Occurred" status code in transient error retry logic
- This error occurs when Cloudflare cannot reach the origin server within the timeout period
- Updates documentation for transient error handling in rate-limits.md and http.md

## Test plan

- [x] Added unit tests for `isTransientError` with 524 status code
- [x] Added tests for `fetchWithProxy` retry behavior with 524 errors
- [x] Tests verify case-insensitive matching of "timeout" in status text
- [x] Tests verify non-matching 524 responses are not retried


danenania:
@claude are there contexts in which a 524 might not be a transient error (in promptfoo onprem deployments, for example) — or is it generally safe to treat as transient?

PR #7608, opened 2026-02-10
fix(table): display sessionIds array from multi-turn strategies

minhle1291:
## Summary

- Multi-turn strategies (IterativeMeta, Crescendo, etc.) store multiple session IDs in `metadata.sessionIds` as an array
- The table display was only checking `metadata.sessionId` (singular), causing blank sessionId columns for these strategies
- Now checks both `metadata.sessionIds` (array) and `metadata.sessionId` (string), with the array taking precedence
- Array values are joined with commas for display

## Test plan

- [x] Added 4 new test cases for sessionIds array handling
- [x] All 19 tests pass
- [x] Manual verification with IterativeMeta/Crescendo strategy runs

image

### DIFF
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -1093,10 +1093,16 @@ export default class Eval {
     // Fetch all results for these test indices in a single query
     const allResults = await EvalResult.findManyByEvalIdAndTestIndices(this.id, testIndices);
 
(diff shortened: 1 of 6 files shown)


will-holley (approved):
Image

PR #7611, opened 2026-02-11
fix(evaluator): detect already-complete eval on resume

yash2998chhabria:
## Summary
- When `--resume` finds all test cases already completed, log a clear message and keep finalization running instead of silently running zero tests

## Test plan
- [x] Unit tests pass (`test/evaluator.sigint.test.ts`)
- [x] Manual test: pause eval, resume -> verify "Nothing to resume" message when all cases done


src/evaluator.ts
```diff
@@ -1409,8 +1409,14 @@ class Evaluator {
           }
         }
         const skipped = originalCount - runEvalOptions.length;
-        if (skipped > 0) {
+        if (skipped > 0 && runEvalOptions.length > 0) {
           logger.info(`Resuming: skipping ${skipped} previously completed cases`);
+        } else if (runEvalOptions.length === 0) {
```
MrFlounder: are you sure this case exist?

src/evaluator.ts
```diff
@@ -1409,8 +1409,14 @@ class Evaluator {
           }
         }
         const skipped = originalCount - runEvalOptions.length;
```
MrFlounder: **Suggestion: Add test for partial resume path**

This condition changed from `if (skipped > 0)` to `if (skipped > 0 && runEvalOptions.length > 0)`, but only the all-complete case (`runEvalOptions.length === 0`) has a new test. There is no test covering the partial-resume path where some cases are skipped but work remains.

A regression here (e.g., accidentally flipping `>` to `>=`) would go undetected.

**Recommendation:** Add a test with e.g. 2 test cases where only 1 is in the completed set, asserting:
- `logger.info` is called with the `"Resuming: skipping 1 previously completed cases"` message
- `callApi` is called exactly once (for the remaining case)

MrFlounder (commented):
Review comment from court review process.

src/evaluator.ts
MrFlounder: good!!!

MrFlounder (approved):

yash2998chhabria:
partial resume test needs to be added to the submodule , will be done separately

src/evaluator.ts
```diff
@@ -1409,8 +1409,14 @@ class Evaluator {
...
-        if (skipped > 0) {
+        if (skipped > 0 && runEvalOptions.length > 0) {
           logger.info(`Resuming: skipping ${skipped} previously completed cases`);
+        } else if (runEvalOptions.length === 0) {
+          logger.info(
```
mldangelo: do we want to exit 1 here?

yash2998chhabria: claude said "don't exit 1 there. It's not an error, and forcing a process
  exit from inside the evaluator would break the design intent of letting finalization run. If the CLI needs
  to communicate "nothing happened" to a shell script, that should be handled at the CLI layer, not deep in
  the evaluator."
  
  I think I agree with it

PR #7646, opened 2026-02-12
fix(cli): prevent API error logs from corrupting progress bar display

### LINKED ISSUE #2134: API errors cause CLI progress bars to redraw
**Describe the bug**

Basically if there is an API error (in the provided case, a gateway error), the error is printed and the progress bars are redrawn below it, with the progress bars above it freezing.

image

**To Reproduce**

Not sure of the best approach to reproduce since it's intermittent. Perhaps removing network connectivity and trying to run an eval?

**Expected behavior**

Ideally, the error would be displayed somewhere but the progress bars would either clear or redraw.

**Screenshots**

See above

- Promptfoo version: I am using v0.96.1, although I think I saw this after https://github.com/promptfoo/promptfoo/issues/1970 was merged and before our organization OpenAI rate limit was raised. So I think it has been around for a little bit at least.

**Additional context**

This is not the biggest deal, but wanted to file (feel free to close if not significant enough.)

Sebastiangmz:
## Summary
- Add a log interceptor to `ProgressBarManager` that routes winston console output through a coordinated write path
- When the progress bar is active, log messages are written by: clearing the bar line, writing the message, then re-rendering the bar
- Prevents error messages from overlapping with or corrupting the progress bar during evaluation runs
- Interceptor is installed after `initialize()` and removed before every `stop()` call

### Methods added to ProgressBarManager
- `log(message)` — write above the progress bar without corruption
- `installLogInterceptor()` — intercept winston console transport writes
- `removeLogInterceptor()` — restore original transport behavior

Closes #2134

## Test plan
- [ ] Run eval with a provider that intermittently fails — progress bar should remain clean
- [ ] Error messages appear above the progress bar, not interleaved
- [ ] Normal (no-error) evals behave identically to before

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -1923,6 +1923,10 @@ class Evaluator {
           return this.evalRecord;
         }
       } else {
+        if (progressBarManager) {
+          progressBarManager.removeLogInterceptor();
+          progressBarManager.stop();
+        }
         if (ciProgressReporter) {
           ciProgressReporter.error(`Evaluation failed: ${String(err)}`);
         }
--- a/test/evaluator.progress.test.ts
+++ b/test/evaluator.progress.test.ts
@@ -2,6 +2,8 @@ import readline from 'readline';
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
+import type { RunEvalOptions } from '../src/types/index';
+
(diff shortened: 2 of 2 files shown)


mldangelo:
Thank you @Sebastiangmz!

PR #7652, opened 2026-02-12
feat: add regenerate button for suggested policies

typpo:


src/app/src/pages/redteam/setup/components/Targets/PolicySuggestionsSidebar.tsx
will-holley, line 63: Should this copy update instead?

PR #7685, opened 2026-02-14
feat(tracing): create root testcase spans and emit evaluation events

typpo:
- Create a per-test-case root span and derive trace context from it when OTEL is initialized.
- Emit gen_ai.evaluation.result events and pass traceparent through to grader calls.
- Use `PROMPTFOO_OTEL_ENABLED` and `gen_ai.provider.name` (no legacy enable env var or `gen_ai.system`).


src/tracing/evaluatorTracing.ts
```diff
@@ -76,20 +80,23 @@ export async function startOtlpReceiverIfNeeded(testSuite: TestSuite): Promise<v
     `[EvaluatorTracing] Full testSuite.tracing: ${JSON.stringify(testSuite.tracing, null, 2)}`,
   );
 
-  if (
-    testSuite.tracing?.enabled &&
-    testSuite.tracing?.otlp?.http?.enabled &&
-    !otlpReceiverStarted
-  ) {
+  const envTracingEnabled = getEnvBool('PROMPTFOO_OTEL_ENABLED', false);
+  const suiteTracingEnabled = testSuite.tracing?.enabled === true;
```
mldangelo-oai, line 84: [P2] Mirror per-test tracing when deciding to start the receiver

`evaluate()` initializes OTEL when `defaultTest.metadata.tracingEnabled` or any test's metadata enables tracing, and `isTracingEnabled()` documents the same per-test mode. This receiver predicate checks only the suite flag and environment, so those evaluations create local root spans but never start the built-in receiver for externally instrumented child spans that export to it. A focused probe with `tests: [{ metadata: { tracingEnabled: true } }]` confirmed that `startOTLPReceiver` is not called. Use the same suite-wide tracing predicate in both places and add default-test/per-test receiver cases.

src/tracing/evaluatorTracing.ts
```diff
@@ -76,20 +80,23 @@ export async function startOtlpReceiverIfNeeded(testSuite: TestSuite): Promise<v
...
+  const tracingEnabled = suiteTracingEnabled || envTracingEnabled;
+  const httpReceiverEnabled = testSuite.tracing?.otlp?.http?.enabled !== false;
```
mldangelo-oai, line 86: [P2] Avoid the fixed three-second tail cost for implicit receivers

Changing the HTTP receiver from opt-in to enabled unless explicitly `false` means every suite- or env-enabled trace run now starts it. Evaluator teardown then sleeps for three seconds whenever the receiver was started, even when no external provider sends spans. Repeated programmatic or server evaluations inherit that fixed latency. Keep receiver startup explicit, or replace the blind sleep with a bounded drain signal that runs only when an external receiver was actually acquired; add a timing regression for tracing without `otlp.http.enabled`.

src/tracing/evaluatorTracing.ts
```diff
@@ -196,6 +201,38 @@ export async function generateTraceContextIfNeeded(
   }
   const testCaseId = test.metadata?.testCaseId || (test as any).id || `${testIdx}-${promptIdx}`;
 
+  let traceId: string;
+  let traceparent: string;
+  let rootSpan: Span | undefined;
+
+  if (isOtelInitialized()) {
+    const tracer = trace.getTracer('promptfoo.evaluator');
+    rootSpan = tracer.startSpan('promptfoo.testcase', {
```
mldangelo-oai, line 210: [P2] Start testcase roots from `ROOT_CONTEXT`

`startSpan` inherits the active OpenTelemetry context when no explicit context is supplied. If programmatic evaluation runs inside an instrumented request, independent testcase roots become children of that ambient span and share its trace ID. A direct probe at this head reproduced that inheritance, so sibling rows can be grouped under the caller's trace instead of their own TraceStore records. Pass `ROOT_CONTEXT` as the start context and test evaluation while another span is active.

src/evaluator.ts
```diff
@@ -1353,7 +1451,8 @@ async function runEvalInternal({
       testSuite,
       vars: state.vars,
     });
-    const { response, traceContext } = providerCall;
+    const { response } = providerCall;
+    traceContext = providerCall.traceContext;
```
mldangelo-oai, line 1455: [P2] Create evaluation spans for precomputed provider outputs

`callProviderForRunEval` still sets `traceContext` to `null` whenever `test.providerOutput` is present. As a result, a tracing-enabled precomputed test runs its assertions but this new lifecycle emits no `promptfoo.testcase` span, status, or `gen_ai.evaluation.result` event at all. Root/evaluation telemetry should not depend on whether a live provider call was needed. Generate the testcase context independently, skip only provider propagation, and add a traced `providerOutput` case.

src/evaluator.ts
```diff
@@ -1419,6 +1518,10 @@ async function runEvalInternal({
       registers[test.options.storeOutputAs] = ret.response.output;
     }
 
+    if (!rootSpanEndDeferred) {
```
mldangelo-oai, line 1521: [P2] Finalize testcase spans after comparison assertions

This records the provisional row outcome and the `finally` below closes the span before `processComparisonAssertions` later applies `select-best` and `max-score`. A comparison loser can therefore be persisted as failed while its exported root span still reports a pass and the old score. Defer event/status recording and closure until comparison post-processing has produced the final row, with a traced comparison regression asserting the persisted and exported outcomes agree.

src/evaluator.ts
```diff
@@ -1451,6 +1555,10 @@ async function runEvalInternal({
         metadata,
       },
     ];
+  } finally {
```
mldangelo-oai, line 1558: [P2] Finalize detached timeout spans before OTEL shutdown

The caller enforces timeouts with `Promise.race`, aborts, and stops awaiting this evaluation. If a provider ignores the abort signal, this `finally` runs only when that provider eventually settles, while the overall evaluator may already have flushed and shut down OTEL. The timed-out testcase span is then absent or exported too late. Finalize the timeout outcome/span before shutdown and guard late completion from recording another outcome; cover a deliberately non-cooperative provider.

src/evaluator.ts
```diff
@@ -2471,7 +2579,7 @@ function createRunEvalTest(
 }
 
 function isTracingEnabledForTest(testSuite: TestSuite, testCase: AtomicTestCase) {
-  const tracingEnvEnabled = getEnvBool('PROMPTFOO_TRACING_ENABLED', false);
+  const tracingEnvEnabled = getEnvBool('PROMPTFOO_OTEL_ENABLED', false);
```
mldangelo-oai, line 2582: [P1] Keep the deployed tracing flag as a compatibility alias

Current main, its docs, and existing installations still use `PROMPTFOO_TRACING_ENABLED`. Replacing every lookup with `PROMPTFOO_OTEL_ENABLED` makes those upgrades silently stop creating trace contexts and initializing the SDK. The early bot note called this acceptable when the flag was new, but the branch has remained open while the old contract continued shipping. Keep the old name as a deprecated fallback, let the new name win when explicitly set, and test old-only, new-only, and conflicting-value precedence.

mldangelo-oai (commented):
Comment-only review at exact head `f37116d8a72e6528b2b27a0bce8571fd513a2a11`. I would not merge this historical head. Its 37 published checks are green, but it is now 510 commits behind current `main` and conflicts in the two central implementation files, `src/evaluator.ts` and `src/tracing/evaluatorTracing.ts`. Current main still lacks the testcase-root/evaluation-event feature, so the useful idea is not superseded; the later #7686 stack contains an evolved copy but is also stale and conflicted rather than a safe landing path.

The existing Copilot thread about raw `process.env` parsing is still current and unresolved. The issue-level notes that dropping `PROMPTFOO_TRACING_ENABLED` and `gen_ai.system` is breaking are also merge blockers now that current main and deployed configurations still use those contracts; I did not duplicate the schema note inline.

Focused local validation passed 114 tracing tests with four integration cases skipped. Two suites could not collect under the borrowed dependency tree because this old head expects the pre-v5 `js-yaml` default export. Separate focused probes reproduced ambient-parent inheritance and the per-test receiver predicate gap. `git diff --check` passes.

Recommended disposition: retire this exact branch and rederive the root-span/evaluation-event slice on current main's shared receiver lifecycle and result-linkage architecture. Preserve legacy env and telemetry aliases during migration, finalize each root exactly once after all grading/comparison work, cover timeouts/precomputed outputs, and require fresh full CI plus independent approval.

PR #7686, opened 2026-02-14
feat(tracing): improve OTLP trace correlation

typpo:
- Correlate OTLP spans using `promptfoo.eval.id` + `promptfoo.test.case.id`.
- Emit only `promptfoo.*` correlation attributes from the OpenAI Agents OTLP exporter and the Python wrapper.
- Merged current `main`, resolved the Python worker stderr parser conflict, and addressed review feedback around env parsing, OTLP bind defaults, and the unused Python tracing init guard.
- Made the OpenTelemetry JavaScript example deterministic by flushing after the workflow span closes and aligning duration thresholds with the simulated workflow.

Validation:
- `npx vitest run test/python/worker.test.ts test/python/workerPool.test.ts test/providers/pythonCompletion.test.ts test/providers/pythonCompletion.cliState.test.ts test/tracing/evaluatorTracing.test.ts test/tracing/genaiTracer.test.ts test/tracing/integration.test.ts test/tracing/localSpanExporter.test.ts test/tracing/providerInstrumentation.test.ts test/integration/tracing.test.ts test/evaluator/trace-integration.test.ts test/providers/openai/agents-tracing.test.ts test/types/index.test.ts test/util/tokenUsageCompat.test.ts --sequence.shuffle=false`
- `python3 -m unittest src/python/persistent_wrapper_test.py`
- `npm run local -- eval -c examples/integration-opentelemetry/javascript/promptfooconfig.yaml --no-cache --no-share -o /tmp/pr7686-otel-js-fixed2.json`
- `npm run l && npm run f`
- `git diff --check && git diff --cached --check`
- `npm run build`


ghost:
**⏩ No test execution environment matched (07fed322bbfa4ae4ea17ebb13dcd6b05fa1fee69) [View output ↗](https://app.usetusk.ai/app/testing-commit-check/e0ef9221-b1ec-414d-91f6-c7790c1d3303?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

---
<p>

| be1535f7a39835d5e43157c5c6197748270e94ec | ⏩ Skipped due to new commit on branch | [Output](https://app.usetusk.ai/app/testing-commit-check/3fb94275-9627-4bbd-a3b4-4b2cbf4c36bb?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Feb 15, 2026 6:19AM |
| c20c4db54999579fe1e02fa70aea427fec8f982c | ✅ Generated 10 tests - 10 passed | [Tests](https://app.usetusk.ai/app/testing-commit-check/d0f0f23e-2c26-4616-b6ef-221933a9b2f6?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Feb 15, 2026 6:27AM |
| 07fed322bbfa4ae4ea17ebb13dcd6b05fa1fee69 | ⏩ No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/e0ef9221-b1ec-414d-91f6-c7790c1d3303?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Feb 17, 2026 3:18AM |

</p>
</details>

mldangelo-oai (commented):
Comment-only review at exact head `51468a456cdd8633d9ba8963514f395c715a86f6`. I would not merge this head. The core correlation work is still useful, but the branch is 472 commits behind current `main`, conflicts in six central tracing/evaluator paths, and has only CodeQL checks on the current head—no normal build, unit, integration, UI, or platform CI. All three existing inline threads are fixed or stale at this head; the remaining concerns are the new findings below plus the already-noted `gen_ai.system` compatibility break.

Focused local checks passed 211 Python/tracing/backend tests and 20 TraceView tests, with five integration cases skipped. Three suites could not collect because the borrowed dependency tree lacks optional `better-sqlite3`; direct typechecking was also blocked by borrowed current-main dependency drift rather than a diagnostic in a changed production file. Direct probes reproduced the Python double execution and ambient-parent correlation failures.

Recommended disposition: retire this exact branch and port the surviving work as focused current-main successors: (1) dual-read correlation compatibility, (2) testcase-root/evaluation lifecycle, and (3) Python/Agents endpoint propagation. Preserve current main's receiver lifecycle, TraceStore redaction, DTO validation, and worker fixes; keep legacy environment and attribute aliases during migration. Each successor should have full JS/Python/OTLP integration coverage, exact-head normal CI, resolved review threads, and independent approval before merge.

src/server/routes/traces.ts
```diff
@@ -2,11 +2,41 @@ import { Router } from 'express';
...
 import { TracesSchemas } from '../../types/api/traces';
+import { sanitizeObject } from '../../util/sanitizer';
 import { replyValidationError } from '../utils/errors';
 import type { Request, Response } from 'express';
 
 export const tracesRouter = Router();
 
+function shouldSanitizeTraces(req: Request): boolean {
+  const raw = req.query.sanitize;
+  if (raw === 'false' || raw === '0') {
```
mldangelo-oai, line 13: [P1] Do not expose raw spans through an unauthenticated query flag

`sanitize=false` and `sanitize=0` let any client that can reach this unauthenticated route switch both endpoints to raw TraceStore output. The new tests even confirm that `apiKey`, `authorization`, and `x-api-key` values are returned. Before this change—and on current `main`—the HTTP route always used the store's sanitized default. Remove the public opt-out, or put raw access behind a real authenticated and authorized capability; a query parameter cannot safely weaken the response policy. Add anonymous cases for both values on both endpoints.

src/server/routes/traces.ts
```diff
@@ -2,11 +2,41 @@ import { Router } from 'express';
...
+function sanitizeTracePayload<T extends { spans?: unknown; metadata?: unknown }>(trace: T): T {
+  const spans = Array.isArray(trace.spans)
+    ? trace.spans.map((span: unknown) => {
+        if (!span || typeof span !== 'object') {
+          return span;
+        }
+        const spanRecord = span as Record<string, unknown>;
+        return {
+          ...spanRecord,
+          attributes: sanitizeObject(spanRecord.attributes),
```
mldangelo-oai, line 28: [P1] Retain the trace-specific sanitizer for span data

This path deliberately fetches raw span attributes and then applies the generic object sanitizer. That sanitizer does exact key matching after removing only `_` and `-`, so dotted OpenTelemetry keys such as `http.request.header.authorization`, `rpc.request.metadata.api_key`, and `db.connection.password` pass through when their values are short and do not match a token heuristic. The previous TraceStore sanitizer normalizes punctuation and matches sensitive substrings, so default responses regress from redacted to exposed. Keep TraceStore's default sanitizer for spans and sanitize metadata separately, with nested tests for dotted `authorization`, `api_key`, `token`, `password`, and `secret` keys.

src/server/routes/traces.ts
```diff
@@ -20,10 +50,14 @@ tracesRouter.get('/evaluation/:evaluationId', async (req: Request, res: Response
     logger.debug(`[TracesRoute] Fetching traces for evaluation ${evaluationId}`);
 
     const traceStore = getTraceStore();
-    const traces = await traceStore.getTracesByEvaluation(evaluationId);
+    const sanitize = shouldSanitizeTraces(req);
+    const traces = await traceStore.getTracesByEvaluation(evaluationId, {
```
mldangelo-oai, line 54: [P2] Preserve the trace response string bound

Fetching raw attributes here also bypasses TraceStore's recursive 400-character limit. `sanitizeObject` does not truncate ordinary strings, so prompts, responses, request bodies, and tool arguments now return at their full stored size on the default evaluation and single-trace endpoints. A trace with many spans can therefore cause a large serialization, transfer, and TraceView-rendering regression. Keep the store sanitizer or reproduce its recursive bound in a dedicated trace response sanitizer, and cover long strings in nested objects and arrays on both routes.

src/python/persistent_wrapper.py
```diff
@@ -203,6 +231,12 @@ def _traced_call(method_callable, args, function_name):
         if isinstance(context_arg, dict):
             traceparent = context_arg.get("traceparent")
 
+    _ensure_tracing_for_context(context_arg)
```
mldangelo-oai, line 234: [P1] Do not replay a failing Python provider after tracing

This automatic initialization makes `_traced_call` the normal path for traced workers, but the outer tracing fallback catches the user exception re-raised by the inner handler and calls `call_method` a second time. A focused reproduction at this head observed two provider executions from one call. That can duplicate charges or state mutations and can mask a transient first failure. Limit the fallback to tracing setup/instrumentation failures before the user callable is entered; once invocation starts, record the exception and propagate it without replaying the provider.

src/tracing/evaluatorTracing.ts
```diff
@@ -196,6 +245,38 @@ export async function generateTraceContextIfNeeded(
   }
   const testCaseId = test.metadata?.testCaseId || (test as any).id || `${testIdx}-${promptIdx}`;
 
+  let traceId: string;
+  let traceparent: string;
+  let rootSpan: Span | undefined;
+
+  if (isOtelInitialized()) {
+    const tracer = trace.getTracer('promptfoo.evaluator');
+    rootSpan = tracer.startSpan('promptfoo.testcase', {
```
mldangelo-oai, line 254: [P2] Start testcase roots from `ROOT_CONTEXT`

`startSpan` inherits the currently active OpenTelemetry context when no explicit context is supplied. If programmatic evaluation runs inside an instrumented request, independent testcase roots become children of that ambient span and share its trace ID. A direct probe confirmed this, and sibling rows can then collide with TraceStore's unique trace correlation. Pass `ROOT_CONTEXT` as the start context so each `promptfoo.testcase` is actually a root, and test evaluation under an active parent span.

src/evaluator.ts
```diff
@@ -1419,6 +1527,10 @@ async function runEvalInternal({
       registers[test.options.storeOutputAs] = ret.response.output;
     }
 
+    if (!rootSpanEndDeferred) {
```
mldangelo-oai, line 1530: [P2] Finalize testcase spans after comparison assertions

This records the outcome and the `finally` below closes the span before `select-best` and `max-score` comparison assertions are applied later. A row can be persisted as a failed comparison loser while its exported root span still reports the provisional pass/score. Defer outcome recording and closure until comparison post-processing has produced the final result, and add a traced comparison test that checks the event and span status against the persisted row.

src/evaluator.ts
```diff
@@ -1451,6 +1564,10 @@ async function runEvalInternal({
         metadata,
       },
     ];
+  } finally {
```
mldangelo-oai, line 1567: [P2] Close detached timeout spans before OTEL shutdown

The caller's timeout uses `Promise.race`, aborts, and then stops awaiting this evaluation. If a provider ignores the abort signal, this `finally` does not run until that provider eventually settles, while the top-level evaluator can already flush and shut down OTEL. The timed-out testcase span is then absent or exported after shutdown. Finalize the timeout result/span before shutdown and guard late completion from recording a second outcome; cover a provider that deliberately ignores abort.

src/evaluator.ts
```diff
@@ -901,6 +922,9 @@ function buildCallApiContext({
     callApiContext.traceparent = traceContext.traceparent;
     callApiContext.evaluationId = traceContext.evaluationId;
     callApiContext.testCaseId = traceContext.testCaseId;
+    if (testSuite) {
+      callApiContext.otelExporterOtlpEndpoint = getLocalOtlpHttpEndpoint(testSuite);
```
mldangelo-oai, line 926: [P2] Forward the computed endpoint to OpenAI Agents

The evaluator adds the configured local OTLP endpoint to `CallApiContextParams`, but `OpenAiAgentsProvider` still builds trace metadata from only `agentConfig.otlpEndpoint`. With a non-default receiver port, Agents spans therefore fall back to `localhost:4318` instead of the receiver started here. Have the Agents provider use `agentConfig.otlpEndpoint ?? context?.otelExporterOtlpEndpoint`, and add an integration case with a custom receiver port.

src/evaluator.ts
```diff
@@ -2471,7 +2588,7 @@ function createRunEvalTest(
 }
 
 function isTracingEnabledForTest(testSuite: TestSuite, testCase: AtomicTestCase) {
-  const tracingEnvEnabled = getEnvBool('PROMPTFOO_TRACING_ENABLED', false);
+  const tracingEnvEnabled = getEnvBool('PROMPTFOO_OTEL_ENABLED', false);
```
mldangelo-oai, line 2591: [P1] Keep the documented tracing flag as an alias

Existing installations enable evaluator tracing with the documented `PROMPTFOO_TRACING_ENABLED` flag. Replacing every lookup with `PROMPTFOO_OTEL_ENABLED` makes those upgrades silently stop creating trace contexts, initializing OTEL, and starting the receiver unless operators discover and rename the variable. Keep the old name as a deprecated fallback, let the new flag win when it is explicitly set, and test old-only, new-only, and conflicting-value precedence before removing the alias in a later release.

src/tracing/otlpReceiver.ts
```diff
@@ -404,8 +405,9 @@ export class OTLPReceiver {
   }
 
   private recordTraceInfo(traceInfoById: Map<string, TraceInfo>, trace: ParsedTrace): void {
-    const evaluationId = trace.span.attributes?.['evaluation.id'] as string | undefined;
-    const testCaseId = trace.span.attributes?.['test.case.id'] as string | undefined;
+    const attrs = trace.span.attributes ?? {};
+    const evaluationId = attrs[PromptfooAttributes.EVAL_ID] as string | undefined;
```
mldangelo-oai, line 409: [P2] Accept legacy correlation attributes during migration

The base receiver accepts `evaluation.id` and `test.case.id`; this replacement reads only the new `promptfoo.*` keys. Existing external exporters—and even the in-tree OpenAI Agents tracing example—can therefore send otherwise valid spans that reach `createTraceRecords` without IDs and are skipped as unlinked. Dual-read both forms with canonical-key precedence, keep emitting the new names, and add legacy-only/canonical-only/conflict tests before removing the old aliases.

src/tracing/evaluatorTracing.ts
```diff
@@ -76,20 +125,22 @@ export async function startOtlpReceiverIfNeeded(testSuite: TestSuite): Promise<v
...
 
-  if (
-    testSuite.tracing?.enabled &&
-    testSuite.tracing?.otlp?.http?.enabled &&
-    !otlpReceiverStarted
-  ) {
+  const envTracingEnabled = getEnvBool('PROMPTFOO_OTEL_ENABLED', false);
+  const suiteTracingEnabled = isSuiteTracingEnabled(testSuite);
+  const tracingEnabled = suiteTracingEnabled || envTracingEnabled;
+  const httpReceiverEnabled = testSuite.tracing?.otlp?.http?.enabled !== false;
```
mldangelo-oai, line 131: [P2] Avoid a fixed three-second delay for implicit receivers

This changes the HTTP receiver from opt-in to enabled unless explicitly set to `false`. Any tracing-enabled suite—including metadata-only tracing—now starts it, and evaluator shutdown unconditionally sleeps three seconds whenever it was started. Repeated programmatic/server evaluations inherit that fixed tail cost even when no out-of-process spans need time to arrive. Gate receiver startup/waiting on an explicit need, or replace the blind sleep with a bounded exporter/receiver drain signal, and add a timing regression test for tracing without an explicit HTTP receiver.

PR #7700, opened 2026-02-15
feat(redteam): outcome-driven constraint violation strategy

typpo:


will-holley:
@claude create a test plan that I can use to validate these changes work as as the author intended, do not introduce regressions into the system, and add value differentiated from the other red team strategies.

src/server/services/redteamTestCaseGenerationService.ts
```diff
@@ -299,6 +300,61 @@ async function handleGoatStrategy(
...
+  const response = await fetchWithRetries(
+    getRemoteGenerationUrl(),
+    {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(odcvBody),
+    },
+    REQUEST_TIMEOUT_MS,
```
mldangelo-oai, line 327: [P1] Use the imported timeout accessor

`REQUEST_TIMEOUT_MS` is neither declared nor imported in this module, so every server-side ODCV generation throws before `fetchWithRetries` can issue the request. The added ODCV service test reproduces this locally, and exact-head CI run 26059223583 reports the same `ReferenceError` across the Linux Node matrix and macOS/Windows. Pass `getRequestTimeoutMs()` here, matching the neighboring handlers, and `await expectTaskRequest(...)` in the ODCV test so all request assertions are observed.

src/redteam/providers/odcv.ts
```diff
@@ -0,0 +1,594 @@
...
+    return data.message;
+  }
+
+  private buildTaskBody(turn: number, state: OdcvState): string {
+    const goal =
+      state.targetContext.test?.metadata?.goal || state.targetContext.vars?.[this.config.injectVar];
+    return JSON.stringify({
+      goal,
+      i: turn,
+      messages: state.messages,
```
mldangelo-oai, line 294: [P1] Honor the target-output privacy opt-out

The global `excludeTargetOutputFromAgenticAttackGeneration` option is forwarded into every strategy config, but this provider drops it and always sends `state.messages`, including target responses appended as `assistant`, to the remote ODCV service on the next turn. That violates the documented promise that target output remains hidden from attack generation when the option is enabled. Retain the flag and filter target responses from this payload; the server-preview path at `handleOdcvStrategy` needs the same filtering/flag propagation because it also sends the full conversation history. Add two-turn regressions for both paths.

src/redteam/constants/strategies.ts
```diff
@@ -24,6 +24,7 @@ export const MULTI_TURN_STRATEGIES = [
   'jailbreak:hydra',
   'custom',
   'mischievous-user',
+  'odcv',
```
mldangelo-oai, line 27: [P2] Wire ODCV through the dialog save path

Adding `odcv` here makes `StrategyConfigDialog` render the generic multi-turn controls, but `handleSave` only forwards an allowlisted set that does not include ODCV. Clicking Save therefore closes the dialog without calling `onSave`, silently discarding `maxTurns` and `stateful` edits. Include ODCV in the save path and add a dialog regression. While doing so, align the omitted `stateful` default: the dialog currently displays it as enabled while this provider and the new docs default it to false.

src/server/services/redteamTestCaseGenerationService.ts
```diff
@@ -299,6 +300,61 @@ async function handleGoatStrategy(
...
+  const odcvBody = {
+    task: 'odcv',
+    goal: ctx.effectiveGoal,
+    i: ctx.turn,
+    messages: ctx.conversationHistory,
+    version: VERSION,
+    email: ctx.email,
+    purpose: ctx.purpose ?? undefined,
+    variant: ctx.strategyConfigRecord['variant'] ?? 'incentivized',
+    inputs: ctx.strategyConfigRecord['inputs'] ?? undefined,
```
mldangelo-oai, line 315: [P2] Preserve plugin-level inputs in setup preview

The documented configuration puts ODCV `inputs` under the plugin. Production `addOdcvTestCases` copies those values into the transformed provider config, but the setup route passes this handler only the original strategy config, so this lookup is undefined and preview sends no inputs. Source the effective inputs from the transformed test/provider or `baseMetadata.pluginConfig`, and cover a plugin-input-only preview request. This is separate from the existing thread about validating malformed strategy-level `variant`/`inputs`.

src/redteam/providers/odcv.ts
```diff
@@ -0,0 +1,594 @@
...
+      inputs?: Record<string, string>;
+      _perTurnLayers?: LayerConfig[];
+    } = {},
+  ) {
+    if (neverGenerateRemote()) {
+      throw new Error(`ODCV strategy requires remote grading to be enabled`);
+    }
+    invariant(typeof options.injectVar === 'string', 'Expected injectVar to be set');
+    this.config = {
+      maxTurns: options.maxTurns || 5,
```
mldangelo-oai, line 135: [P2] Validate and cap the turn budget

`maxTurns` is a catch-all strategy value here, but this accepts it without checking for a finite positive integer and does not apply the 10-turn cap used by the other agentic strategies for unauthenticated users. `Infinity` makes the loop unbounded, while negative/fractional values produce nonsensical runs; the server preview resolver likewise has no upper cap. Validate the value, default with `??`, and enforce the same authenticated/unauthenticated limits in both execution paths.

src/redteam/providers/odcv.ts
```diff
@@ -0,0 +1,594 @@
...
+    const renderedAttackerMessage = await this.addRenderedAttackerMessage(
+      turn,
+      attackerMessage,
+      state,
+    );
+    this.logHistory(turn, state.messages);
+
+    const latestMessageContent = state.messages[state.messages.length - 1].content;
+    const targetPrompt = await this.getTargetPrompt(turn, latestMessageContent, state);
+    if (targetPrompt.skip) {
```
mldangelo-oai, line 236: [P2] Roll back failed attacker turns and derive output from target responses

`addRenderedAttackerMessage` has already appended the remote-generated message before this early return, and the outer catch also leaves it behind when the target fails before a response is recorded. Later generator calls and stateless target prompts therefore consume a turn that never reached the target. More seriously, the validator accepts `assistant` as an attacker role while `createResponse` reports the last assistant message as target output, so an assistant-role attack followed by a transform/target failure is returned as if the target produced it. Stage or roll back pre-target turns on every skip/failure and derive final output only from a recorded target response. Add an assistant-role plus transform-failure regression.

src/redteam/providers/odcv.ts
```diff
@@ -0,0 +1,594 @@
...
+
+    const attackerVars: Record<string, VarValue> = {
+      [this.config.injectVar]: processedMessage,
+      ...(currentRenderInputVars || {}),
+    };
+    const renderedAttackerPrompt = await renderPrompt(
+      state.targetContext.prompt,
+      {
+        ...state.targetContext.vars,
+        ...attackerVars,
```
mldangelo-oai, line 356: [P2] Pass current materialized vars to the target context

These per-turn attacker and multi-input values are used only to render the prompt. `callTarget` still receives the original cloned `state.targetContext.vars`, so providers, request transforms, and auth functions that inspect `context.vars` see the initial inject value and stale inputs even though the prompt argument contains the current materialized turn. Build a per-turn context that merges `attackerVars` while preserving the private session clone, and assert both the prompt and context observed by `targetProvider.callApi`.

mldangelo-oai (commented):
Reviewed exact head `1b4b10ae43bc2d76670ccfcda465da519c743163` in comment-only mode. I cannot recommend merging this version. The new ODCV setup/server path fails deterministically before it can make a request, and the direct provider ignores the explicit target-output privacy opt-out. I also found five additional integration/state defects inline. Exact-head CI run 26059223583 reproduces the undefined timeout on the Linux Node matrix and macOS/Windows; locally, the focused service file had five passing tests and the ODCV test failed at line 327 with the same `ReferenceError`. The ODCV strategy tests passed (2/2), the existing strategy-dialog tests passed (46/46) but contain no ODCV save case, and typecheck independently reports `TS2552` for the undefined identifier (other local diagnostics came from borrowing newer dependencies). The provider suite could not load locally because this checkout lacks the optional `better-sqlite3` dependency, so I did not count it as a branch failure.

Three current unresolved threads also remain valid and should stay open: runtime validation for `variant`/`inputs`; shared target-response handling for empty outputs, `conversationEnded`, and `maxCharsPerMessage`; and media/blob externalization in red-team history. The older thread saying tests and docs are missing is stale: this head now includes substantial provider/strategy/server coverage and ODCV documentation, so that thread can be resolved.

ODCV is not present on current `main`, so the feature is not superseded. This branch is nevertheless 510 commits behind `origin/main`, has an empty PR description, and overlaps shared server, layer, schema, sidebar, and data-handling paths that have continued to evolve. GitHub currently reports it mergeable, but I recommend porting the still-unique ODCV work onto a fresh current-main successor rather than merging this historical branch wholesale. The successor should rerun the full required matrix and add focused regressions for privacy filtering in both paths, ODCV UI save/default behavior, plugin-input preview parity, finite/capped turn budgets, failed-turn rollback/output provenance, current per-turn context vars, and the three still-open threads. No branch edits or pushes were made.

src/redteam/providers/odcv.ts
```diff
@@ -0,0 +1,594 @@
...
+    if (neverGenerateRemote()) {
+      throw new Error(`ODCV strategy requires remote grading to be enabled`);
+    }
+    invariant(typeof options.injectVar === 'string', 'Expected injectVar to be set');
+    this.config = {
+      maxTurns: options.maxTurns || 5,
+      injectVar: options.injectVar,
+      variant: options.variant || 'incentivized',
+      stateful: options.stateful ?? false,
+      inputs: options.inputs,
```
mldangelo-oai, line 139: [P1] Preserve cloud target context in remote ODCV calls

Current `main` now adds `targetId` to every strategy config so remote task handlers can resolve the cloud target's owning team. `addOdcvTestCases` preserves that config, but this constructor rebuilds its own allowlist and drops `targetId`; the ODCV task body and per-turn transform context therefore cannot include `remoteGenerationContextPayload(targetId)`. For saved cloud/on-prem targets this can select the wrong/default team context or fail provider resolution. Retain `targetId`, include the remote-generation context in the task payload, pass it through runtime transforms, and cover both paths.

src/redteam/providers/odcv.ts
```diff
@@ -0,0 +1,594 @@
...
+      return {};
+    }
+
+    state.messages.push({
+      role: 'assistant',
+      content: targetCall.output,
+    });
+    this.addHistoryItem(renderedAttackerMessage, targetPrompt.transformResult, targetCall, state);
+
+    const grading = await this.gradeTurn(renderedAttackerMessage.content, targetCall.output, state);
```
mldangelo-oai, line 251: [P1] Pass target images into the grader

The full `targetCall.response` is available here, but `gradeTurn` receives only the string output and calls `grader.getResult` without a `RedteamGradingContext`. A response with benign text plus a policy-violating image is therefore graded only on text and can be recorded as safe; image-only responses are skipped even earlier by the output invariant. Match current-main agentic providers by passing `providerResponse` and `imageOutputs`, accept image-only responses using the shared convention, and add text-plus-image and image-only regressions. This is separate from the existing history/blob-externalization thread, which does not affect grading.

src/redteam/shared/attackProviders.ts
```diff
@@ -36,6 +36,7 @@ export const ATTACK_PROVIDER_IDS = [
   'hydra', // jailbreak:hydra
   'crescendo',
   'goat',
+  'odcv',
```
mldangelo-oai, line 39: [P2] Preserve plugin inputs when ODCV is layered

Registering ODCV as an attack provider sends layer compositions through `addLayerTestCases`' generic provider-construction branch, which never invokes `addOdcvTestCases`. For the documented plugin-level `inputs` form, `[odcv, audio]` therefore omits `metadata.pluginConfig.inputs` even though standalone ODCV copies them, so business context and multi-input variables disappear. Merge those inputs into the layered ODCV provider config and cover plugin-level inputs with a per-turn layer. This is a separate production path from the setup-preview input loss noted earlier.

mldangelo-oai (commented):
Follow-up on the same exact head after completing the semantic-integration comparison with current `main`: I found three additional unique gaps that are not covered by my first review or the older live threads. ODCV drops the cloud target/team context now supplied to agentic strategies, omits target images from grading, and loses plugin-level inputs when composed through `layer`. These are current-main compatibility requirements rather than textual merge conflicts, and they reinforce the recommendation to rederive ODCV on a fresh current-main branch. No branch edits or pushes were made.

PR #7731, opened 2026-02-17
feat(eval): persist namedScores and metadata from afterEach hook

### LINKED ISSUE #7515: Feature: Allow afterEach extension hook to persist metadata and namedScores to the database
## Problem

The `afterEach` extension hook currently operates in a read-only capacity relative to evaluation results. While `beforeAll` and `beforeEach` can mutate context and persist changes by returning the modified context, mutations made in `afterEach` to `context.result.metadata`, `context.result.namedScores`, or `context.result.response.metadata` are not persisted to the database or displayed in the web viewer.

This is documented behavior — the docs state:

> "The beforeAll and beforeEach hooks may mutate specific properties of their respective context arguments in order to modify evaluation state."

`afterEach` is notably absent from that statement.

## Use Case

We use the Claude Agent SDK provider and parse session log files in `afterEach` to extract ground-truth tool usage data (tool calls, success/failure, MCP tools used, cost, session viewer URLs, etc.). This is runtime data that only exists *after* the provider has responded, so it cannot be computed in `beforeEach`.

Currently we work around this by moving all persistence into assertion `namedScores` (which do persist), but this limits us to numeric values and forces metrics logic into assertions rather than the extension lifecycle where it belongs.

Specifically, we'd like to be able to:

1. Add custom `namedScores` (numeric metrics like `num_turns`, `cost_usd`, `mcp_calls`) from `afterEach`
2. Add custom `metadata` (structured data like tool call details, session viewer URLs) from `afterEach`

## Proposed Solution

Allow `afterEach` to persist changes to `context.result` when the hook returns the modified context, similar to how `beforeAll` and `beforeEach` work. At minimum, support persisting:

- `context.result.namedScores` — merged into the existing namedScores
- `context.result.metadata` — merged into the existing metadata

## Alternatives Considered

- **Assertion namedScores**: Works for numeric metrics but doesn't support structured data or strings. Also conflates metrics collection with assertion logic.
- **External storage**: We do write to output files, but this data doesn't appear in the promptfoo web viewer.
- **Provider transform**: Runs before assertions, doesn't have access to session log data that requires the full response.

## Environment

- promptfoo version: latest (via npx)
- Extension format: CommonJS (.cjs)

bishnubista:
## Summary

Closes #7515

Allow `afterEach` extension hooks to persist changes to `result.namedScores` and `result.metadata` — previously these mutations were silently discarded.

- Add `afterEach` case in `runExtensionHook` switch to merge returned `namedScores` and `metadata`
- Move `afterEach` hook call **before** DB persist (`addResult`) and metrics aggregation so changes flow through to database, file outputs, and web viewer
- Update `AfterEachExtensionHookContext` JSDoc to document mutable fields

### Scope

This PR addresses the "at minimum" scope from #7515: persisting `result.namedScores` and `result.metadata`. Support for `result.response.metadata` is intentionally left for a follow-up if maintainers want full parity with the broader problem statement.

### Design decisions

- **Controlled mutability**: Only `namedScores` and `metadata` are writable. `success`, `score`, `response`, and other fields are **not** overridable from `afterEach` to prevent accidental overrides of assertion results.
- **Shallow merge**: Uses `{...existing, ...returned}` so extensions can add/override individual keys but cannot delete existing ones.
- **Chaining**: Multiple extensions see each previous extension's changes (same pattern as `beforeAll`/`beforeEach`).

## Test plan

- [x] New tests for `afterEach` return value handling (6 tests):
  - namedScores merge
  - metadata merge
  - Preservation when extension returns nothing
  - Chaining multiple extensions
  - Later extensions overriding earlier values
  - Immutability of `success`, `score`, and `response`
- [x] All existing evaluatorHelpers tests pass (91/91)
- [x] All existing evaluator tests pass (137/137)
- [x] Lint clean (`npm run l`)
- [x] TypeScript compiles (`npm run tsc`)


bishnubista:
Thanks @mldangelo! completely overlooked mutation bypass and chaining issues. Great additions! 🙏

mldangelo:
Thank you @bishnubista!

PR #7466, opened 2026-02-03
fix(eval): fix SSL "bad record mac" errors under concurrent API load

mldangelo:
## Summary

- Fix `setGlobalDispatcher()` race condition in `fetchWithProxy()` that caused ~45% SSL failure rate under concurrent load (e.g. `-j 20` with image providers)
- Replace per-call undici Agent creation with a singleton that includes `keepAliveTimeout` and connection pooling sized to the `-j` concurrency flag
- Add SSL/TLS error patterns (`ssl`, `tls`, `bad record mac`, `eproto`) to the scheduler retry policy
- Add body-streaming retry in `fetchWithCache` for SSL errors during response body transfer
- Add `isTransientConnectionError()` helper for future adaptive concurrency use

## Root Cause

Every call to `fetchWithProxy()` created a **new `undici.Agent`** and called `setGlobalDispatcher()`. Under concurrent load, this rapidly replaced the global dispatcher while in-flight requests still referenced the old agent's connection pool, corrupting TLS session state and producing `ssl/tls alert bad record mac` errors.

## Test plan

- [x] Unit tests for SSL error patterns in `shouldRetry()` (3 new tests)
- [x] Unit tests for `isTransientConnectionError()` helper (9 new tests)
- [x] All 285 existing scheduler/cache tests pass
- [x] End-to-end: [config](https://gist.github.com/mldangelo/0a4cc807b786707d4376ad8b7463be8f) with `-j 20` (40 evals) — error rate dropped from ~45% to 0-5% -
- [x] Lint and format clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7758, opened 2026-02-18
feat(fetch): add PROMPTFOO_FETCH_CONNECTIONS env var to override HTTP pool size

jameshiester:
## Summary
- The undici Agent connection pool is sized from `cliState.maxConcurrency` (the `-j` flag), defaulting to `DEFAULT_MAX_CONCURRENCY` (4). This was introduced in #7466 to fix SSL "bad record mac" errors.
- 4 connections per origin is appropriate for CLI eval but too low for server deployments where many concurrent LLM calls share the same process — requests queue behind the pool limit, adding significant latency.
- Adds `PROMPTFOO_FETCH_CONNECTIONS` env var that takes priority over `cliState.maxConcurrency`, allowing server deployments to configure HTTP connection pooling independently of eval concurrency.

## Test plan
- [ ] Existing fetch tests pass (no behavior change when env var is unset)
- [ ] Setting `PROMPTFOO_FETCH_CONNECTIONS=128` creates an Agent with 128 connections per origin
- [ ] CLI `-j` flag still works when env var is unset

### DIFF
--- a/src/util/fetch/index.ts
+++ b/src/util/fetch/index.ts
@@ -29,6 +29,23 @@ let cachedAgentConcurrency: number | undefined;
 let cachedProxyAgents: Map<string, ProxyAgent> = new Map();
 
 /**
+ * Get the connection pool size for HTTP agents.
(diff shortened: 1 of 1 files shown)


addelong (approved):
Nice catch!

PR #7782, opened 2026-02-19
fix(ui): use replace for URL search param updates to fix back button

Ebonsignori:
## Summary

Companion to https://github.com/promptfoo/promptfoo-cloud/pull/3552

- Added `{ replace: true }` to `setSearchParams` in the `useSearchParamState` hook so URL state updates (tab switches, filter changes, etc.) don't push phantom history entries
- Mirrors the fix from promptfoo-cloud#3552 — without this, pressing the browser back button cycles through every search param change instead of returning to the previous page

### Not changed (intentionally)

| File | Reason |
|------|--------|
| `redteam/setup/page.tsx` (`navigate('#...')`) | Multi-step wizard — back should return to previous step |
| All `navigate()` calls to eval/report detail pages | Genuine page navigations that should push history |

### Already correct

`ResultsView.tsx` and `Eval.tsx` already pass `{ replace: true }` on their direct `setSearchParams` calls.

## Test plan

- [ ] Use any page that reads URL state via `useSearchParamState` (e.g. eval results view/mode param), change state, press Back — should leave the page in one click
- [ ] Red team setup wizard: navigate steps with Next/Back, press browser Back — should traverse wizard steps (unchanged behavior)

### DIFF
--- a/.claude/skills/search-params/skill.md
+++ b/.claude/skills/search-params/skill.md
@@ -1,0 +1,128 @@
+---
+name: search-params
+description: URL search param and hash state management. Use when adding or modifying URL search params, working with useSearchParams, setSearchParams, useSearchParamState, or navigate() with query strings or hash fragments, or fixing browser back/forward button issues.
+---
+
+# URL Search Param State Management
+
+## Decision Framework
+
+When updating the URL (search params or hash), choose between **replace** and **push** based on whether the change represents in-page state or a user-navigable step:
+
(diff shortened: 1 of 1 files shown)


mldangelo (approved):
Thank you! Should we cover this with a test or a comment in code? perhaps an agents.md change?

will-holley:
@Ebonsignori please copy over the skill from Cloud.

PR #7791, opened 2026-02-20
fix: Preserve delimiting colons in Ruby-namespaced method names in `file://` references

willlockwood:
When writing external-file-based assertions in Ruby, `fileRef.split(":")` breaks assertions using Ruby namespaced methods like `MyModule::Nested.method` into ["MyModule", "", "Nested.method"], so only "MyModule" gets captured as the function name, and everything else gets discarded.

So this:
- Fixes parsing of `file://` references in assertion values to split on the first `:` only, preserving `::` in Ruby namespaced method names (e.g., file://script.rb:MyModule::Nested.method, and maybe other file types too, I dunno)
- Add tests
- Updates docs to use more idiomatic term `method` in Ruby over `function`, and an example for using class method references

### DIFF
--- a/test/assertions/scriptValueResolution.test.ts
+++ b/test/assertions/scriptValueResolution.test.ts
@@ -443,5 +443,31 @@ describe('Script value resolution', () => {
       );
       expect(result.pass).toBe(true);
     });
+
+    it('should return fail result when runRuby throws for a namespaced method', async () => {
+      const { runAssertion } = await import('../../src/assertions/index');
+      const mockRunRuby = vi.mocked(runRuby);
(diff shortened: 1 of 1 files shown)


mldangelo (approved):
Thank you so much @willlockwood! 

Please reach out (michael @ promptfoo.dev) if you would like some promptfoo swag 

Would also love to pick your brain about ruby integrations. We could definitely do a better job in supporting providers, assertions, transforms, etc.

willlockwood:
@mldangelo great, sounds good! I'll definitely find a way to get in touch and chat 👍🏻

PR #7826, opened 2026-02-23
chore: minor improvements to strategy and plugin card ui

typpo:
# Before

image

# After

image


Ebonsignori (dismissed):
Think the select in the upper RH corner is a good move, but since the settings icon is clickable, I think it should be turned into a button instead of an icon.

Also, what does the star icon do/indicate? It's not clickable and appears on every strategy I see. Is it necessary?

mldangelo (approved):

typpo:
Let's merge this without the icon change, we can debate that UI separately but I just want to unbreak the UI

The sparkles are for generating example test cases

PR #7834, opened 2026-02-23
chore(redteam): Update UI to make it clear for hardcoded plugins

sklein12:
These two plugins are hardcoded and can only generated 34 and 13 test cases respectfully. It looks like something is broken when more than those limits are requested.

image

### DIFF
--- a/src/redteam/index.ts
+++ b/src/redteam/index.ts
@@ -190,7 +190,7 @@ function getEffectiveRequested(pluginId: string, requested: number): number {
  * @returns A formatted string containing the report.
  */
 function generateReport(
-  pluginResults: Record<string, { requested: number; generated: number }>,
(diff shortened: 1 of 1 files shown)


will-holley (approved):

src/redteam/index.ts
```diff
@@ -145,6 +145,12 @@ function getPluginDisplayId(plugin: { id: string; config?: Record<string, any> }
   return 'policy';
 }
 
+// Plugins with hardcoded test case limits (cannot generate more than this)
+const PLUGIN_MAX_TESTS: Record<string, number> = {
```
typpo, line 149: Maybe add a comment in the code for these two strategies so we remember to keep this in sync.

Or add a test to make sure it's in sync

typpo (approved):

sklein12:
I have a more comprehensive approach I'm going to take here.

Issue #6380, opened 2025-11-27
`azure:responses` missing `isReasoningModel` config override (parity with `azure:chat`)


joakimriedel:
## Problem

  The `azure:responses` provider's `isReasoningModel()` method only checks deployment name prefixes, unlike
  `azure:chat` which also supports a `config.isReasoningModel` override.

  This means `reasoning_effort` is silently ignored when using custom Azure deployment names that don't start with
  `o1`, `o3`, `o4`, or `gpt-5`.

  ## Current Behavior

  **`azure:chat`** (correct):
  ```typescript
  // src/providers/azure/chat.ts
  protected isReasoningModel(): boolean {
    if (this.config.isReasoningModel || this.config.o1) {
      return true;  // ✅ Config override supported
    }
    // fallback to deployment name check
  }

  azure:responses (missing override):
  // src/providers/azure/responses.ts
  isReasoningModel(): boolean {
    return (
      this.deploymentName.startsWith('o1') ||
      this.deploymentName.startsWith('o3') ||
      this.deploymentName.startsWith('o4') ||
      this.deploymentName.startsWith('gpt-5')
    );  // ❌ No config override
  }
```

  Impact

  Users with custom Azure deployment names (common in enterprise) can't use reasoning_effort:

```yaml
  # This silently ignores reasoning_effort because deployment name
  # "my-company-gpt-5-prod" doesn't START with "gpt-5"
  - id: azure:responses:my-company-gpt-5-prod
    config:
      reasoning_effort: "low"  # ❌ Ignored silently
```

  Expected Behavior

```yaml
  # Should work with explicit config override
  - id: azure:responses:my-company-gpt-5-prod
    config:
      isReasoningModel: true   # Override auto-detection
      reasoning_effort: "low"  # ✅ Now applied
```

  Suggested Fix

  Update src/providers/azure/responses.ts to match azure:chat:

```typescript
  isReasoningModel(): boolean {
    if (this.config.isReasoningModel || this.config.o1) {
      return true;
    }
    return (
      this.deploymentName.startsWith('o1') ||
      this.deploymentName.startsWith('o3') ||
      this.deploymentName.startsWith('o4') ||
      this.deploymentName.startsWith('gpt-5')
    );
  }
```

PR #7848, opened 2026-02-24
feat(providers): override OpenAI reasoning detection

jeffpardy:
## Summary

- Adds an optional `isReasoningModel` boolean for `openai:chat` and `openai:responses` so custom or compatible deployment aliases can override name-based capability detection.
- Reuses the existing centralized OpenAI capability path. Prompt config takes precedence over provider config; explicit `true` or `false` takes precedence over built-in detection; omission preserves existing behavior.
- Keeps Chat and Responses request contracts distinct: Chat uses `reasoning_effort` and `max_completion_tokens`, while Responses uses nested `reasoning` and `max_output_tokens`.
- Preserves compatible-provider behavior for Groq, TrueFoundry, OrcaRouter, and Bedrock Mantle, including provider-specific temperature rules.
- Rejects malformed non-boolean values before a request is sent and never serializes the control flag into the upstream request body.

## Motivation

Promptfoo's built-in detection necessarily relies on model-name patterns. Arbitrary Azure deployment names and OpenAI-compatible aliases can therefore be misclassified, which changes token fields, reasoning fields, and temperature handling. Existing Azure-host inference covers only part of that problem; it does not cover generic compatible endpoints, Chat parity, or an explicit false override. Related reports and prior work include #6380, #8208, and #8255.

## Compatibility and risk

- This is an additive optional setting. Existing valid configurations are unchanged when it is omitted.
- `isReasoningModel: false` changes request shaping only; it cannot make an upstream reasoning model accept unsupported fields. The docs call out this constraint.
- The Azure coverage here is the `openai:responses` Azure-compatible endpoint path. Native `azure:*` providers retain their separate existing contract.
- Responses `stream: true` is covered at request-serialization level; this PR does not change the pre-existing Responses SSE transport behavior.

## Validation

- `npx vitest test/providers --run` — 211 files passed, 1 skipped; 6,869 tests passed, 9 skipped.
- `npm run build` — TypeScript, package bundles, and web app build passed.
- `npm run jsonSchema:generate` — passed and produced no tracked diff.
- `npm run architecture:check` — passed with no new cross-layer dependency.
- `cd site && SKIP_OG_GENERATION=true npm run build` — docs build passed (GitHub stats fetch was rate-limited but non-fatal).
- `npm run local -- eval -c /tmp/pr7848-e2e.yaml --no-cache -o /tmp/pr7848-final-output.json` — 7/7 passed against a local mock endpoint. Captured requests covered Chat/Responses true and false overrides, aliases, GPT-OSS, Responses `stream: true`, OrcaRouter, and Bedrock Gemma.
- Malformed CLI case with `isReasoningModel: "false"` — exited 100 with `isReasoningModel must be a boolean`; the mock endpoint received no request.


src/providers/openai/chat.ts
```diff
@@ -307,15 +312,7 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
...
 
-    if (
-      config.reasoning &&
-      (this.modelName.startsWith('o1') ||
-        this.modelName.startsWith('o3') ||
-        this.modelName.startsWith('o4') ||
-        this.modelName.includes('/o1') ||
-        this.modelName.includes('/o3') ||
-        this.modelName.includes('/o4'))
-    ) {
```
typpo: Should we keep these matches so we don't break existing users?

jeffpardy: Thanks for the review @typpo! I think those would be included in the above call to isReasoningModel()? It should have done the same o-series checks.

ianw-oai (approved):

src/providers/openai/chat.ts
```diff
@@ -191,6 +191,11 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
   }
 
   protected isReasoningModel(): boolean {
+    // Check explicit config flags first
+    if (this.config.isReasoningModel) {
+      return true;
+    }
```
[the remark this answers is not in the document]
mldangelo-oai: Verified on the current head: Chat passes the merged prompt/provider config into isReasoningModel, honors explicit true and false values, and includes prompt-level regression coverage in test/providers/openai/chat.test.ts.

src/providers/openai/responses.ts
```diff
@@ -137,6 +137,11 @@ export class OpenAiResponsesProvider extends OpenAiGenericProvider {
   }
 
   protected isReasoningModel(): boolean {
+    // Check explicit config flags first
+    if (this.config.isReasoningModel) {
+      return true;
+    }
```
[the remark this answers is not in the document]
mldangelo-oai: Verified on the current head: Responses passes the merged prompt/provider config into isReasoningModel, honors explicit true and false values, and includes prompt-level regression coverage in test/providers/openai/responses/reasoning.test.ts.

src/providers/openai/chat.ts
mldangelo-oai: Confirmed. The generalized condition changed GPT-5 Chat compatibility by forwarding the Responses-style nested reasoning field. Commit 5c90b8524 restores the prior GPT-5 behavior while retaining custom and o-series override support, with regression coverage and local end-to-end request capture.

mldangelo-oai:
Final review completed at exact head `3af4f565f927ca38486606ca3ad367f444418854`.

- Fixed: merged current `origin/main` additively; centralized prompt/provider precedence and explicit `true`/`false` handling; preserved distinct Chat and Responses token/reasoning fields; corrected GPT-OSS, Azure-compatible Responses, Groq/TrueFoundry, OrcaRouter, and Bedrock postprocessing; malformed non-boolean values now fail before network I/O.
- QA: provider suite passed (211 files, 1 skipped; 6,869 tests, 9 skipped); TypeScript, build, architecture, schema generation, and docs build passed; exact-head local `--no-cache` mocked eval passed 7/7, while the malformed case exited 100 and sent no request.
- CI: exact-head run [28076082977](https://github.com/promptfoo/promptfoo/actions/runs/28076082977) is green: 31 passed, 2 intentionally skipped, 0 failed. All 3 review threads are resolved/outdated.
- Compatibility: the setting is optional and omission preserves existing detection. Remaining limits are documented: forcing `false` cannot make an upstream model accept unsupported fields; Responses streaming was verified at request-serialization level only; native `azure:*` providers keep their separate contract.

The code and CI are clean. GitHub still reports `REVIEW_REQUIRED`, so fresh human approval is the remaining merge blocker.

PR #7854, opened 2026-02-24
chore(providers): expose redteam plugin/strategy in call context

typpo:
Add `pluginId` and `strategyId` fields to call_api context


mldangelo:
Closing in favor of #7858

PR #7859, opened 2026-02-25
fix: convert b64_json to data URIs for image grading and add evaluation guide

mldangelo:
## Summary

- Convert OpenAI-style `b64_json` image responses to correctly typed data URIs for vision-based grading when `PROMPTFOO_INLINE_MEDIA=true`.
- Preserve URL siblings during mixed image-output rewrites and avoid writing blob assets or references when a multi-image response cannot be externalized atomically.
- Preserve imported evaluation-history provider JSON when inline media is enabled; imports should not discard fields such as `created` or `revised_prompt`.
- Normalize inline image responses in Hydra before grading and redteam history capture.
- Add the image evaluation guide and example workflow, and correct the Google image provider ID, multi-image wording, and inline-audio documentation.

## Validation

- [x] Merged current `main` (`e473f8eebe1cefbe5db51cf7f5e44c04232e04ec`) into this PR branch.
- [x] Reconciled all existing inline review threads; all five are resolved on the current PR.
- [x] `npx vitest run test/blobs/ test/models/evalResult.test.ts test/redteam/providers/shared.test.ts test/redteam/providers/hydra/index.test.ts --sequence.shuffle=false` (`232` passed).
- [x] `npm run f`, `npm run l`, and `npm run tsc` (`npm run l` reports the existing non-blocking Hydra complexity warning).
- [x] `npm run build`.
- [x] `cd site && SKIP_OG_GENERATION=true npm run build` (restricted-network live-stat fetch fell back as designed; static build succeeded).
- [x] Fresh isolated `npm run db:migrate` after merging `main`; confirmed `evals.is_redteam` exists in the temporary database.
- [x] Credential-free local CLI evaluation using disposable `file://` target and grader providers with isolated state: the raw WebP `b64_json` result was received by the grader as `data:image/webp;base64,...` and passed.
- [x] GitHub CI, CodeQL, and title validation passed on `889067d0e84edcfb51a93084f6d88685f1aea2c4`.


ghost:
**⏩ No test execution environment matched (12d7c88ac76f7afc9dfba39446744408c62a8976) [View output ↗](https://app.usetusk.ai/app/testing-commit-check/5abe1b00-073e-4708-b8ab-12d531900b67?client=e91e3433-8fb4-4e98-b233-bba492f4909c)**

---
<p>

| 9b93e09afaa1150453f5e5a838f06cbe889fb7a7 | ⏩ No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/5dfe1d61-185e-422a-899d-fa9a89d5c5ef?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Feb 25, 2026 8:57AM |
| 12d7c88ac76f7afc9dfba39446744408c62a8976 | ⏩ No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/5abe1b00-073e-4708-b8ab-12d531900b67?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Feb 25, 2026 9:27AM |

</p>
</details>

src/blobs/extractor.ts
```diff
@@ -221,6 +221,116 @@ async function externalizeDataUrls(
...
+    }
+
+    if (!isBlobStorageEnabled()) {
+      const dataUris = b64Items.map(({ b64 }) => `data:${detectImageMimeType(b64)};base64,${b64}`);
+      setOutputFromUris(next, dataUris);
+      logger.debug('[BlobExtractor] Converted b64_json to inline data URI', {
+        ...context,
+        count: dataUris.length,
+      });
+      return true;
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in c2b3fb35b: bulk eval-result persistence and redteam history now call the extractor so inline b64_json normalization still runs when blob storage is disabled.

src/blobs/extractor.ts
```diff
@@ -221,6 +221,116 @@ async function externalizeDataUrls(
   return { value, mutated: false };
 }
 
+/**
+ * Detect image MIME type from the first bytes of base64-encoded data.
+ * Falls back to image/png if unrecognized.
+ * @internal Exported for testing
+ */
+export function detectImageMimeType(b64: string): string {
+  const buffer = Buffer.from(b64.trim(), 'base64');
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in c2b3fb35b: MIME detection now decodes only the base64 prefix needed for the image signatures.

mldangelo-oai:
Synced this PR with current `main`, fixed one additional mixed-output edge case, and pushed `5910d31d53ce7b3137b8f468d1c9da403e15feec`.

What changed:
- Preserved `url` siblings when rewriting OpenAI-style `data[].b64_json` outputs, so mixed `url` + `b64_json` image arrays no longer drop the URL item in either blob-storage or `PROMPTFOO_INLINE_MEDIA=true` mode.
- Kept `metadata.blobUris` limited to actual `promptfoo://blob/` refs when the rewritten output also includes external URLs.
- Added regression coverage for both stored-blob and inline-media mixed image outputs.

Review notes:
- Rechecked thread-aware review state; no unresolved review threads remain.
- Prior front matter/admonition/performance/call-site feedback remains addressed on the current diff.
- No migrations in this PR.

Validation:
- `npx vitest run test/blobs/extractor.test.ts test/models/evalResult.test.ts test/redteam/providers/shared.test.ts --sequence.shuffle=false` (155 passed)
- `npx vitest run test/blobs/ test/models/evalResult.test.ts test/redteam/providers/shared.test.ts --sequence.shuffle=false` (175 passed)
- `npm run l`
- `npm run f`
- `npm run tsc`
- `npm run build`
- `cd site && npm run ensure-stats && SKIP_OG_GENERATION=true npm run build`
- Real local CLI smoke with isolated config/cache dirs: `PROMPTFOO_INLINE_MEDIA=true npm run local -- eval -c /tmp/promptfoo-pr7859-e2e.xVVbew/inline-media-config.yaml --no-cache -o /tmp/promptfoo-pr7859-e2e.xVVbew/inline-media-output.json`; exported result passed and output was `["https://example.com/existing.png","data:image/png;base64,iVBORw0KGgoAAAANSUhEUg"]`.

CI is queued/running on the new head, and I am moving to the next PR per the requested loop.

src/blobs/extractor.ts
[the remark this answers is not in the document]
mldangelo-oai: Follow-up in `1237b56bc`: tracing the remaining production call sites showed `createManyFromEvaluateResult` is the import/history ingestion path, where normalizing inline image JSON would discard archival provider fields such as `created` and `revised_prompt`. That path now preserves inline JSON when blob storage is disabled. Fresh generated evaluation results and redteam history still normalize inline images; Hydra now invokes the normalizer before grading/history capture with regression coverage.

src/blobs/extractor.ts
```diff
@@ -400,6 +400,169 @@ async function externalizeMetadataAudio(
...
+      return false;
+    }
+
+    if (!isBlobStorageEnabled()) {
+      const inlineUris = imageItems.map((item) =>
+        item.type === 'b64'
+          ? `data:${detectImageMimeType(item.value)};base64,${item.value}`
+          : item.value,
+      );
+      setOutputFromUris(next, inlineUris);
```
jameshiester-oai, line 486: **[P1] Preserve structured images for red-team grading**

This branch rewrites only `next.output`. When Hydra normalizes an OpenAI-style raw `b64_json` response, `targetResponse.images` therefore stays absent, so Hydra omits `gradingContext.imageOutputs` and the base red-team grader takes its text-only `matchesLlmRubric` path. With a valid PNG, I reproduced a data-URI output, no `images` property, zero image parts in the grading prompt, and the base64 embedded as text. The grader cannot inspect the pixels, which can create false positives/negatives or context-limit errors in the exact path this PR says it enables. After validating the bytes, populate structured `ImageOutput` records (or otherwise materialize the normalized URI into Hydra's grading context), and assert the actual multimodal grading prompt/context rather than only the string output argument.

examples/openai-images/promptfooconfig.judge.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+defaultTest:
+  options:
+    provider: openai:chat:gpt-5.4-mini
+    rubricPrompt: |
+      [
+        {
+          "role": "user",
+          "content": [
+            { "type": "image_url", "image_url": { "url": "{{output}}" } },
```
jameshiester-oai, line 29: **[P1] Do not put the attachment placeholder in an image URL**

Both configured built-in image providers return a data-URI `output` and structured `images`. The current grading path detects those images, replaces `{{output}}` with `[Image output attached. Inspect the attached image directly for visual grading.]`, and auto-appends the real structured image. This line consequently emits two `image_url` parts: the first has that plain-text placeholder as its URL and the second has the real data URI. A local built-CLI reproduction using this exact rubric failed 0/1 and exported both values; an OpenAI-compatible grader will reject the malformed first URL. For built-in providers, keep this custom rubric text-only and rely on the automatic image attachment (or deduplicate/validate existing image parts), then test the committed example with a built-in-provider-shaped response. The same snippet in both new docs pages needs the same correction.

src/blobs/extractor.ts
```diff
@@ -400,6 +400,169 @@ async function externalizeMetadataAudio(
...
+
+    const imageItems = extractImageOutputItems(parsed.data);
+    if (!imageItems) {
+      return false;
+    }
+
+    if (!isBlobStorageEnabled()) {
+      const inlineUris = imageItems.map((item) =>
+        item.type === 'b64'
+          ? `data:${detectImageMimeType(item.value)};base64,${item.value}`
```
jameshiester-oai, line 483: **[P2] Validate b64_json before labeling it as an image**

`Buffer.from(..., 'base64')` is permissive, and `detectImageMimeType` defaults every unknown or empty payload to PNG. This mutation therefore turns `''`, `'!!!!'`, and base64-encoded ordinary text into valid-looking `data:image/png;base64,...` outputs; I reproduced all three. In red-team flows this value is target-controlled, so malformed output is promoted into persisted/history/grader input instead of remaining a provider error, causing invalid multimodal requests or grading evasion. Require non-empty valid base64 plus a supported image signature/size before rewriting (reusing the strict image materialization checks where possible); if validation fails, preserve the raw response or use a valid same-item URL fallback.

jameshiester-oai (commented):
Reviewed base `c54a30668ad8319d76c20ae96e6680ad6c51a2c6` through head `1910c083a43694eeab5529c71f672f558dab153f` (full PR scope; no prior authenticated completion marker).

## Verdict

The happy-path raw `b64_json` conversion works, but this head has two P1 blockers and three P2 defects. Most importantly, the new Hydra path still gives the grader base64 text instead of pixels, while the committed built-in-provider example emits an invalid placeholder as an `image_url`. GitHub reports the PR as mergeable but behind current `main` (`4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6`).

## Verified findings

### P1

- Raw `b64_json` normalization changes only `output`; it never creates structured `images`. Hydra therefore omits `gradingContext.imageOutputs`, and the base red-team grader follows its text-only path. A valid PNG probe produced a data URI with no `images`; the captured grading prompt contained zero image parts and embedded the base64 as text. This can produce red-team false positives/negatives or context-limit failures because the grader cannot inspect the pixels.
- The new judge config is incompatible with the built-in image-provider contract. Those providers already return structured `images`; grading replaces `{{output}}` with the attachment placeholder and then auto-appends the real image. The committed `image_url` therefore receives a plain-text placeholder alongside the valid appended image. A credential-free built-CLI reproduction using the exact rubric shape failed 0/1 with the two URLs visible in the exported grading prompt.

### P2

- Empty, invalid, and non-image `b64_json` values are all labeled as PNG data URIs. Exact probes produced `data:image/png;base64,`, `data:image/png;base64,!!!!`, and a PNG-labeled encoding of `hello`.
- Multi-image preflight prevents deterministic size-based partial writes, but it is not atomic once I/O starts. With two eligible images and an injected second-store failure, two stores were attempted and the original response was preserved; the first production store would already have committed its provider asset and DB asset/reference without compensating rollback.
- Removing the inline-media guard sends already-inline red-team data URLs through storage-only parsing even though storage is disabled. A 2 MiB image was returned unchanged after three full decodes totaling 8,388,648 base64 input characters. Repeating that work across iterative turns, up to the 50 MiB media ceiling, creates avoidable CPU/RSS risk.

## Existing feedback and simplification

All five prior review threads are resolved; none is unresolved on this head. I rechecked the front matter/admonition fixes, prefix-only MIME sniffing, call-site normalization, and the earlier size-preflight atomicity thread. The runtime second-store failure is a distinct post-preflight case.

Use one validated normalized image representation that includes both the URI and structured `ImageOutput`, then have eval and red-team callers consume it. Split that pure/lightweight normalization from blob storage so inline mode does not enter storage code. Multi-image persistence needs a batch or dedup-aware rollback boundary rather than per-item commits hidden behind a preflight.

## Validation performed

- Changed-surface backend run: 7 files / 249 tests passed.
- Adjacent provider and multimodal-grader run: 3 files / 146 tests passed; three independent specialist runs also passed overlapping 229-, 375-, and 229-test subsets.
- `npm run tsc`, full `npm run build`, production docs build with `SKIP_OG_GENERATION=true`, and `npm run architecture:check` passed.
- Targeted Biome passed with one pre-existing Hydra complexity warning; documentation Prettier and `git diff --check` passed.
- The committed example validates syntactically.
- Credential-free built-CLI QA with local providers exercised both paths: raw WebP `b64_json` normalized to `data:image/webp;base64,...` and passed; a built-in-provider-shaped response using the committed rubric failed because the rendered prompt contained the invalid attachment-placeholder URL plus the real image.
- Direct probes reproduced text-only red-team grading, malformed base64 promotion, triple decoding of already-inline media, and partial side effects after an injected second-store failure.

CI classification: Site (101), Web UI (4,816), and Node 20 backend (19,680) tests passed, then their jobs failed on Codecov v11.2.7 GPG verification (`Can't check signature: No public key`). Builds, docs, integration, smoke, red-team, other platform matrices, CodeQL, style, code scan, and title validation passed.

Limitations: the requested PR Review Toolkit command is not exposed as a callable skill in this environment, so I completed its correctness, tests, docs, error-handling, type-design, and simplification lenses manually and reconciled three independent read-only specialist reviews. No paid image or grader provider was called; user-visible behavior was exercised with local deterministic providers. I inspected the conflict-free synthetic interaction with current `main` but did not merge or rebase it, so post-integration execution remains untested.

src/blobs/extractor.ts
```diff
@@ -400,6 +400,169 @@ async function externalizeMetadataAudio(
...
+      return false;
+    }
+
+    const outputUris: string[] = [];
+    for (const item of imageItems) {
+      if (item.type === 'url') {
+        outputUris.push(item.value);
+        continue;
+      }
+      const stored = await storeOnce(
```
jameshiester-oai, line 519: **[P2] Make the multi-image write atomic after I/O begins**

The eligibility preflight prevents a too-small/too-large later item from causing a partial write, but each `storeOnce` still commits independently. With two eligible 2,056-byte PNGs and an injected failure on the second store, I observed two store attempts and the original JSON output being preserved. In production, the first `storeBlob` has already written the provider object and committed its `blobAssets`/`blobReferences` transaction; this handler imports no compensating deletion, so the result contains no URI while a stale asset/reference remains. Stage/batch the writes and commit references only after all succeed, or implement dedup-aware rollback for earlier writes, and add a regression where the second store throws and no file/asset/reference survives. This is distinct from the resolved size-preflight thread.

src/redteam/providers/shared.ts
```diff
@@ -600,16 +599,13 @@ export interface BaseRedteamResponse {
...
+ * copied into redteam conversation/history.
  */
 export async function externalizeResponseForRedteamHistory<T extends ProviderResponse>(
   response: T,
   context?: { evalId?: string; testIdx?: number; promptIdx?: number },
 ): Promise<T> {
-  if (!isBlobStorageEnabled() && !shouldAttemptRemoteBlobUpload()) {
-    return response;
-  }
   const blobbed = await extractAndStoreBinaryData(response, context);
```
jameshiester-oai, line 609: **[P2] Keep inline-only normalization out of storage parsing**

Removing the early return makes every already-inline red-team image/audio/data URL enter the general blob extractor even when `PROMPTFOO_INLINE_MEDIA=true`. For a 2 MiB PNG data URL, the response was returned unchanged but its full payload was decoded three times—once before `storeOnce`, once for the cache key, and once in `maybeStore` before storage-disabled status is checked—processing 8,388,648 base64 input characters. Iterative callers repeat this per turn, and the supported 50 MiB ceiling makes the transient CPU/RSS cost material. Split the lightweight `b64_json`→data-URI normalizer from storage externalization, or gate storage-only branches before parsing in inline mode; add coverage showing an already-inline response is not decoded/stored.

PR #7862, opened 2026-02-25
feat(providers): support LITELLM_API_BASE env var for base URL override

mishaja12:
The LiteLLM provider previously only allowed overriding the base URL via config.apiBaseUrl. This adds support for the LITELLM_API_BASE environment variable (already defined in the env schema but unused), following the same pattern as other providers like LocalAI and Envoy.

Resolution order: config.apiBaseUrl > provider env > context env > process env LITELLM_API_BASE > default (http://0.0.0.0:4000).


mishaja12:
Hello gentlemen! Is there a chance you could review this?

That would simplify configs not to add in every .yaml litellm provider, but having only 1 in env.

@mldangelo @typpo 


Thank you in advance,
MIsha

mldangelo:
Thank you so much @mishaja12! We'll get this in the next release. Please reach out (michael @ promptfoo.dev) if you would like a t shirt / promptfoo swag.

PR #7870, opened 2026-02-25
fix(redteam): abort scan on non-transient HTTP errors

yash2998chhabria:
## Summary
- Fixes a bug where red team scans got stuck for 9+ hours when targets returned 403 errors
- Iterative strategies (GOAT, Crescendo) would continue retrying indefinitely instead of failing fast
- Now aborts the entire scan immediately when target returns non-transient HTTP errors (401, 403, 404, 500, 501)

## Changes
- Always include HTTP status in provider response metadata (not just debug mode)
- Add `isNonTransientHttpStatus()` utility to classify error codes
- Add `findTargetErrorStatus()` utility to scan in-memory results for non-transient errors
- Add `Eval.findTargetErrorStatus()` model method for efficient DB query
- Detect non-transient errors in evaluator and trigger abort via abort controller
- Infer abort reason from results at summary time (single source of truth)
- Show abort reason in CLI summary with helpful guidance

### Efficient Target Error Detection
Instead of loading all results into memory to check for target errors, we use a single indexed SQL query:
```sql
SELECT CAST(json_extract(response, '$.metadata.http.status') AS INTEGER)
FROM eval_results
WHERE eval_id = ? AND json_extract(response, '$.metadata.http.status') IN (401, 403, 404, 500, 501)
LIMIT 1
```
This is O(1) instead of O(n) for large evaluations.

### Single Source of Truth
Instead of storing `abortReason`/`abortMessage` fields on `EvaluateStats` and `Eval` model, we infer the abort reason from `result.response.metadata.http.status`. This:
- Avoids schema changes
- Keeps data DRY
- Uses HTTP status as the authoritative evidence

## Test plan
- [x] Unit tests for `isNonTransientHttpStatus()` (11 tests)
- [x] Unit tests for `findTargetErrorStatus()` (11 tests)
- [x] E2E test with mock 403 server - verifies scan aborts immediately
- [x] Tests for `Eval.findTargetErrorStatus()` DB query method
- [x] All 137 evaluator tests pass
- [x] All 26 summary tests pass
- [x] Manual verification of summary output shows correct abort message

### DIFF
--- a/src/commands/eval/summary.ts
+++ b/src/commands/eval/summary.ts
@@ -41,7 +41,7 @@ export interface EvalSummaryParams {
   maxConcurrency: number;
   /** Token usage tracker for provider-level breakdown */
   tracker: TokenUsageTracker;
-  /** HTTP status code if the scan was aborted due to a non-transient target error (401, 403, 404, 500, 501) */
+  /** HTTP status code if the scan was aborted due to a non-transient target error (401, 403, 404, 501) */
   targetErrorStatus?: number;
 }
 
@@ -136,7 +136,7 @@ export function generateEvalSummary(params: EvalSummaryParams): string[] {
     lines.push(chalk.yellow('Possible causes:'));
     lines.push(chalk.yellow('  • Invalid API key or authentication (401/403)'));
     lines.push(chalk.yellow('  • Target endpoint does not exist (404)'));
-    lines.push(chalk.yellow('  • Target server error (500/501)'));
+    lines.push(chalk.yellow('  • Server does not support the request (501)'));
     lines.push('');
     lines.push(chalk.cyan('To fix: Check your target configuration and credentials.'));
   }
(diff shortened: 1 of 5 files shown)


sklein12:
- Can we just abort out of the strategy instead of aborting the whole scan?
- 500's and 400 can definitely be transient errors

mldangelo (approved):

mldangelo:
@yash2998chhabria look at @sklein12's comments before merging

yash2998chhabria:
@sklein12 

  Removed 500 from the non-transient list in e6d8b749. You're right that 500s can definitely be transient 
  (server crashes, deployments, DB timeouts), and they can also be input-dependent where one adversarial
  prompt triggers a bug but the next one doesn't. The list is now just [401, 403, 404, 501].

  On aborting per-strategy vs the whole scan, I think for these remaining codes, eval-level abort is the
  right call since they're target-level failures. A 401 (bad auth) or 404 (wrong endpoint) will hit every
  strategy the same way, so cycling through the rest would just repeat the same error.

  That said, for something like 500s where it could be prompt-specific, strategy-level tracking would make
  more sense. I'd like to tackle that as a follow-up if we see it in practice, would track consecutive 500s
  per strategy and skip it while letting others continue. Does that sound reasonable?

PR #7936, opened 2026-03-02
fix(redteam): make HTTP target tests non-blocking

typpo:
Allow HTTP target setup to continue once URL/request config is valid, even if target/session tests have not been run yet.

Keep testing guidance as recommended (non-blocking) and update target-step tests accordingly.


faizanminhas:
why did you wanna do this? I would think if the user updates anything, they should test it? are there cases where the test doesn't work / isn't accurate, and the target is in fact connected?

src/app/src/pages/redteam/setup/components/Targets/index.tsx
```diff
@@ -86,8 +86,7 @@ export default function Targets({ onNext, onBack }: TargetsProps) {
           ? !!selectedTarget.config.url && selectedTarget.config.url.trim() !== ''
           : selectedTarget.config.request?.trim() !== '';
 
-      // For HTTP providers, require both tests to be completed
-      return hasConfig && isTargetTested && isSessionTested;
+      return hasConfig;
     }
```
mldangelo-oai, line 90: [P2] Retire this no-op instead of changing the legacy gate

Neither the PR base nor current `main` mounts this component: `setup/page.tsx` renders `TargetTypeSelection` and `TargetConfiguration`, and a repo-wide import/render search finds this default export only in `index.test.tsx`. The active `TargetConfiguration` already leaves HTTP target/session tests non-blocking and calls the `ProviderConfigEditor` validator when Next is clicked. This return therefore cannot change the wizard behavior described by the PR and only creates a second, divergent implementation in dead code. Close this PR as obsolete; if desired, remove `Targets/index.tsx` and disentangle its still-useful child-component tests in a separate cleanup.

mldangelo-oai (commented):
Reviewed exact head `1b22905d66be8d73e4e4b80ed5bbefdb62be9e9d` against merge base `247eef5ad79b297d37cef4d6f3df98f94569556e` and current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`). The requested behavior already exists in the mounted setup flow, while this patch changes an unmounted legacy component and its isolated tests. I recommend closing/retiring this PR as obsolete rather than refreshing its 510-commit-behind branch.

The focused legacy and active target-configuration suites pass (29/29), but that also confirms the two implementations are separate. The unresolved null/non-string Copilot concern remains technically true only in the dead component; it should not justify carrying the branch forward. If cleanup is useful, delete the legacy component and reorganize its remaining tests in a separate current-main PR. This is a comment-only review; I made no branch changes and pushed nothing.

PR #7950, opened 2026-03-03
fix(providers): migrate browser stealth stack to zorilla

JustinBeckwith:
Yeah, so I may have forked the entire puppeteer-extra project in a fit of rage over unpatched dependencies, set up an entire release flow, tested it all, and released a new set of mostly compatible packages.  I am sorry for being the way I am @mldangelo.  This addresses the first three deprecated packages here. 

image

## Summary
- replace the browser provider's playwright-extra and stealth plugin imports with Zorilla equivalents
- update package dependencies, tests, and browser provider docs/examples to use the new package names
- fix the browser-existing-session example server so the documented node server.js command works in this ESM repo

## Verification
- npx vitest run test/providers/browser.test.ts
- npm run tsc
- npm run l
- npm run local -- eval -c examples/browser-existing-session/promptfooconfig.yaml --filter-providers new-browser -o /tmp/browser-existing-session-output.json --no-cache


sklein12 (approved):
looks good, I would check all the dockerfiles here and in cloud to see if we're installying playwright anywhere. I vaguely remember adding it to a docker container for a customer.

JustinBeckwith:
Checked this while reviewing the Zorilla migration. In this repo, the only Dockerfiles are ./Dockerfile and ./.devcontainer/Dockerfile.dev, and neither explicitly installs Playwright or the stealth packages. The production image still gets browser support transitively through npm ci because package.json keeps playwright, @playwright/browser-chromium, and the @zorilla/* packages in optionalDependencies, so there was no repo-local Dockerfile change to make for this PR. I also fixed the CodeRabbit docs nit on the browser provider page in a follow-up commit.

PR #7967, opened 2026-03-04
chore: clean up plugin presets ui

typpo:
Make it more compact

image


mldangelo (approved):
You need to get some dandruff shampoo for your CI flakes

PR #7974, opened 2026-03-04
fix(output): redact exported config secrets and clarify docs

alexcarol:
## Summary
This PR improves safety around exported eval artifacts and clarifies docs for env usage.

### What changed
- Redact sensitive values from exported eval `config` using sanitizer rules across output formats (`json`, `yaml/yml/txt`, `html`, `xml`).
- Add explicit sanitizer coverage for:
  - `AWS_BEARER_TOKEN_BEDROCK`
  - `ANTHROPIC_API_KEY`
- Expand output export tests to verify those env keys are redacted while non-sensitive env values can remain visible.
- Add docs warning that:
  - `{{ env.* }}` is resolved at config load time
  - templating secrets into `config.env` can materialize them in eval config/exports
  - prefer reading secrets from process env (shell or `--env-file`) and keeping `config.env` for non-sensitive flags
- Clarify output docs that redaction is best-effort, not comprehensive.

## Why
Users can accidentally materialize secrets into `config.env` via templating and then export those configs. This PR reduces leak risk in exported files and makes the behavior explicit in docs.

### DIFF
--- a/test/util/output.test.ts
+++ b/test/util/output.test.ts
@@ -178,14 +178,20 @@ describe('writeOutput', () => {
   it('preserves deep non-secret config fields in JSON output', async () => {
     const outputPath = 'output.json';
     const eval_ = new Eval({
-      l1: {
-        l2: {
-          l3: {
-            l4: {
-              l5: {
-                l6: {
-                  value: 'deep-public-value',
-                  apiKey: 'sk-secret-value',
+      defaultTest: {
+        options: {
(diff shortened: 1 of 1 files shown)


alexcarol:
@promptfoo-scanner

mldangelo:
thank you so much @alexcarol! We'll get this in the next release. Please reach out if you would like some promptfoo swag.

PR #7991, opened 2026-03-05
fix(matchers): handle missing context-faithfulness verdicts

typpo:
Prevent context-faithfulness from incorrectly passing when the grader output has no parseable verdict markers.

Clamp the computed score to [0, 1] and add regression tests for both missing-verdict output and excessive `verdict: no` counts.

https://linear.app/promptfooo/issue/ENG-341/faithfulness-calculation-with-no-verdicts-not-covered


typpo:
The fix doesn’t change the public contract in the docs, config shape, or usage. It only corrects internal scoring when verdict output is missing/malformed and clamps invalid ranges.

PR #8003, opened 2026-03-05
fix(providers): handle mixed-case JSON content type in HTTP provider

typpo:
## Summary

- Parse the HTTP provider's `Content-Type` as a case-insensitive media type instead of using a substring match. This accepts `application/json` and registered `application/*+json` suffixes with valid parameters, while rejecting JSONP/lookalike fields, malformed combined values, invalid parameter syntax, and duplicate parameter names.
- Preserve valid pre-serialized JSON bytes for scalars, `null`, and precision-sensitive numbers without weakening the established typed-template and `transformRequest` behavior. Prompt-derived string templates remain structurally contained, while explicit typed variables, `tools`, `tool_choice`, and `dump` retain their existing semantics.
- Keep invalid-body warnings and debug metadata credential-safe, including JSON-quoted string scalars.
- Add focused regression coverage in `test/providers/http/core.test.ts` for valid, malformed, adversarial, compatibility, templating, precision, and redaction boundaries.

## Validation

- `npx vitest run test/providers/http/core.test.ts test/providers/http/request.test.ts test/providers/http/tools.test.ts --sequence.seed=8003` — 295 passed
- `npx vitest run test/providers/http test/providers/httpTransforms.test.ts test/providers/http-multipart.test.ts test/providers/http-tls.test.ts --sequence.seed=8003` — 443 passed
- `npm run test:integration` — 64 passed
- `npm run tsc`
- `npm run l` — passed with the five pre-existing complexity warnings in `src/providers/http.ts`
- `npm run build`
- Deterministic loopback evaluations with `--no-cache`, inspected through exported results and server-side wire logs:
  - source and built CLI positive/boundary runs: 9/9 passed; normalized exports were identical
  - malformed/adversarial run: 10/10 expected provider errors, with only the two intended strict-server requests reaching the server
  - gzip, text, invalid-JSON response fallback, finite SSE, binary, redirect, and permissive-error compatibility run: 7/7 passed
  - array-valued `Content-Type` rejected by configuration schema
- `npm run test:smoke` — 14 files and 169 tests passed, 1 skipped; the remaining file could not complete its two Ruby assertion cases because this host has no Ruby executable (its JavaScript and Python cases passed)

This does not claim to resolve ENG-175; the available issue linkage does not establish that its report was caused by mixed-case media types.


mldangelo-oai:
## Comprehensive review result: FIXED_AND_READY

Reviewed and verified exact head `a979432e62c31125a0561d2296945f1855a14deb` against current `main` (`81adadadd2fd1bb19b80136ebe66c8c64102bc55`). The branch is up to date and GitHub reports it mergeable. Human approval is still required by repository policy; I did not approve or merge it.

### What changed

- Replaced the permissive substring check with strict, case-insensitive media-type parsing for `application/json` and valid `application/*+json` types, including valid parameters and OWS.
- Rejected JSONP/lookalike values, malformed or combined header values, malformed parameters, and duplicate case-insensitive parameter names.
- Preserved valid pre-serialized JSON bytes, including scalars, `null`, and precision-sensitive numbers, while retaining established typed-template, `dump`, tool, and `transformRequest` behavior.
- Prevented prompt-derived strings from becoming structural JSON and kept invalid-body warnings/debug metadata credential-safe, including JSON-quoted and URL-encoded secret forms.
- Added focused coverage across casing, parameters, suffix media types, malformed/adversarial inputs, templating, precision, compatibility, and redaction boundaries.

Additive campaign commits:

- `ea3c143d4ee226edddad6ef03b5bfd3d0a5a215d` — merge current `main`
- `a979432e62c31125a0561d2296945f1855a14deb` — harden JSON content-type handling

### Verification

- Focused HTTP provider tests: 295 passed.
- Broad HTTP provider suites: 443 passed.
- Integration tests: 64 passed.
- TypeScript, lint, changed-file formatting, and production build passed. Lint emitted only the five pre-existing complexity warnings in `src/providers/http.ts`.
- Deterministic source and built-CLI loopback evaluations used `--no-cache`, with exports and server-side wire logs inspected:
  - positive/boundary: 9/9 passed in both source and built CLI, with identical normalized exports;
  - malformed/adversarial: 10/10 produced the expected provider errors;
  - gzip, text, invalid-JSON fallback, finite SSE, binary, redirect, and permissive error-response compatibility: 7/7 passed;
  - array-valued `Content-Type` was rejected at schema validation.
- Local smoke: 169 passed and 1 skipped; two Ruby assertions could not run because the host lacks Ruby. Exact-head GitHub smoke and Ruby 3.0/3.4 checks passed.
- Fresh independent second-pass reviews of HTTP semantics, adversarial/security behavior, runtime/test compatibility, and simplification/regression found no remaining actionable issue.

### Final state

- Exact-head CI: 39/39 checks passing.
- Review threads: 0 unresolved; the sole prior thread was resolved only after its requested PR-body correction landed.
- Title and body now match the implemented scope and validation evidence, without claiming ENG-175.
- No product/API decision is needed for this patch. The remaining gate is independent human approval.

src/providers/http.ts
```diff
@@ -958,16 +958,73 @@ export const HttpProviderConfigSchema = z.object({
...
+    if (!jsonValueContainsTemplate(parsed)) {
+      return value;
     }
-    return false;
-  });
+    if (typeof parsed !== 'string') {
+      return undefined;
+    }
+    const rendered = renderVarsInObject(parsed, vars);
+    return isTypedJsonTemplate(parsed) && isValidJson(rendered)
```
mldangelo-oai, line 1022: [P2] Preserve computed top-level JSON template types

This classifier treats every expression mentioning `prompt` without `dump`, and every `{% ... %}` block, as an untyped string. The new fast path then quotes valid rendered JSON. In an exact-head loopback probe, a JSON-string body of `"{{ prompt | length }}"` with prompt `abcd` sent `"4"` instead of the JSON number `4`; a JSON-string conditional rendering `{"ok":true}` sent `"{\"ok\":true}"` instead of the object. Both inputs went through `processJsonBody` on the base/current implementation and retained their numeric/object type. Keep the direct unfiltered `{{ prompt }}` containment for injection safety, but preserve valid typed output from transformed expressions and control-flow templates (or fall back to the legacy renderer for those forms), with regressions for both cases.

src/providers/http.ts
```diff
@@ -958,16 +958,73 @@ export const HttpProviderConfigSchema = z.object({
...
+  const expression = value.match(/^\s*\{\{\s*([\s\S]*?)\s*\}\}\s*$/)?.[1];
+  const usesDump = /\|\s*dump(?:\s*\([^)]*\))?(?:\s*\||\s*$)/.test(expression ?? '');
+  return expression != null && (usesDump || !/\bprompt\b/.test(expression));
+}
+
+function renderPreSerializedJsonBody(value: string, vars: Record<string, any>): string | undefined {
+  try {
+    const parsed = JSON.parse(value);
+    if (!jsonValueContainsTemplate(parsed)) {
+      return value;
```
mldangelo-oai, line 1016: [P2] Document the new JSON request-body contract

This now sends any valid template-free JSON string verbatim—including string scalars, `null`, and precision-sensitive numbers—while `site/docs/providers/http.md` still says that any valid JSON string in `body` is converted to an object. The strict parser also makes malformed, combined, lookalike, and duplicate-parameter Content-Type values non-JSON, so object/array bodies that were previously tolerated can now fail validation. Update the provider docs to distinguish object/array bodies, nested and typed/`dump` templates, prompt-derived string containment, and top-level pre-serialized bytes, and document the accepted `application/json` and `application/*+json` forms. Since the generated changelog should not be edited directly, broaden the PR/release title beyond only mixed-case handling so the release entry reflects this compatibility boundary.

mldangelo-oai (commented):
Reviewed exact head `a979432e62c31125a0561d2296945f1855a14deb` against merge base `81adadadd2fd1bb19b80136ebe66c8c64102bc55` and current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`). The strict media-type and pre-serialized-body hardening is well covered, and the focused HTTP provider suite passed locally (188/188). I found two remaining P2 gaps: complex top-level JSON templates can silently change wire type, and the public HTTP-provider docs contradict the new body/media-type contract.

This is a comment-only review; I made no branch changes and pushed nothing. The branch is 61 commits behind but remains mergeable, with no intervening `main` changes to either modified file. After addressing the findings, sync `main` additively, rerun focused and required CI, and obtain independent approval.

PR #8013, opened 2026-03-05
fix(app): restore composite jailbreak strategy visibility

typpo:


src/app/src/pages/redteam/setup/components/Strategies.tsx
```diff
@@ -12,7 +12,6 @@ import { useToast } from '@app/hooks/useToast';
 import {
   AGENTIC_STRATEGIES_SET,
   ALL_STRATEGIES,
-  DEFAULT_STRATEGIES_SET,
```
faizanminhas, line 15: wondering if this can be removed

PR #8030, opened 2026-03-06
fix(providers)!: resolve relative config paths against config dir in claude-agent-sdk

henriklundstrom:
## Summary

- Consistently resolve relative paths in claude-agent-sdk provider against the config file's directory
- Affected options: `working_dir`, `plugins[].path`, `additional_directories`, `debug_file`, `path_to_claude_code_executable`
- Uses `safeResolve` which is a no-op for absolute paths, only affecting relative paths
- Aligns with how other providers (e.g. HTTP provider) handle relative paths in config

## Motivation

When running `promptfoo eval -c some/nested/config.yaml`, relative paths in the config should resolve from the config file's directory. Before this fix, the different path options resolved from two different bases, both inconsistent with promptfoo conventions:

- **`working_dir`** resolved from promptfoo's `process.cwd()` — so running the same config from a different directory would produce different results, in practice forcing users to `cd` into the config file's directory before running `promptfoo eval`.
- **`plugins[].path`, `additional_directories`, `debug_file`, `path_to_claude_code_executable`** were passed straight through to the Claude Code SDK, which resolved them relative to Claude's `cwd` (i.e. the resolved `working_dir`, or a temp directory when `working_dir` is unset). When `working_dir` was unset, this made relative paths for these options practically unusable — as they would resolve from an ephemeral directory outside the user's project.

After this fix, all of these resolve consistently from the config file's directory.

## Breaking change

Relative paths for all affected options now resolve from the config file's directory. Absolute paths are unaffected, and users who run `promptfoo eval` from the same directory as their config see no difference — presumably the majority of users.

- **`working_dir`**: previously resolved from `process.cwd()`. Only users who intentionally rely on relative `working_dir` resolving against a different cwd will see changed behavior.
- **`plugins[].path`, `additional_directories`, `debug_file`, `path_to_claude_code_executable`**: previously resolved from Claude's cwd (the `working_dir`). Users with `working_dir` unset are unlikely to be affected — relative paths would have resolved from a temp directory, so they almost certainly already use absolute paths. Users who set `working_dir` and use relative paths for these options are exceedingly likely to be affected, since a relative path from the config file's directory will typically differ from a relative path from the `working_dir`.

## Example: plugin with a relative path

```
root/
├── evals/
│   ├── promptfooconfig.yaml
│   └── workspace/
└── my-plugin/
    ├── .claude-plugin/
    │   └── plugin.json
    └── skills/
        └── my-skill/
            └── SKILL.md
```

**`evals/promptfooconfig.yaml`**
```yaml
providers:
    config:
      model: claude-sonnet-4-6
      working_dir: ./workspace
      plugins:
        - type: local
          path: ../my-plugin
```

Running from `root/`:

```bash
promptfoo eval -c evals/promptfooconfig.yaml
```

| Option | Before (broken) | After (fixed) |
|---|---|---|
| `./workspace` | cwd (`root/`) → `root/workspace` — wrong | config dir (`root/evals/`) → `root/evals/workspace` — correct |
| `../my-plugin` | Claude's cwd (`root/workspace`) → `root/my-plugin` — happens to be correct | config dir (`root/evals/`) → `root/my-plugin` — correct |

Running from `root/evals/`:

```bash
promptfoo eval -c promptfooconfig.yaml
```

| Option | Before | After |
|---|---|---|
| `./workspace` | cwd (`root/evals/`) → `root/evals/workspace` — correct | config dir (`root/evals/`) → `root/evals/workspace` — correct |
| `../my-plugin` | Claude's cwd (`root/evals/workspace`) → `root/evals/my-plugin` — wrong | config dir (`root/evals/`) → `root/my-plugin` — correct |

### DIFF
--- a/test/providers/claude-agent-sdk.test.ts
+++ b/test/providers/claude-agent-sdk.test.ts
@@ -1,3 +1,4 @@
+import path from 'node:path';
 import fs from 'fs';
 
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
@@ -13,6 +14,8 @@ import type { NonNullableUsage, Query, SDKMessage } from '@anthropic-ai/claude-a
 import type { MockInstance } from 'vitest';
 
 import type { CallApiContextParams } from '../../src/types/index';
+
+const testBasePath = path.resolve('/test/basePath');
 
 vi.mock('../../src/cliState', () => ({
   default: { basePath: '/test/basePath' },
@@ -482,11 +485,11 @@ describe('ClaudeCodeSDKProvider', () => {
         });
         await provider.callApi('Test prompt');
 
-        expect(statSyncSpy).toHaveBeenCalledWith('/test/basePath/workspace');
+        expect(statSyncSpy).toHaveBeenCalledWith(path.resolve(testBasePath, 'workspace'));
         expect(mockQuery).toHaveBeenCalledWith({
           prompt: 'Test prompt',
           options: expect.objectContaining({
-            cwd: '/test/basePath/workspace',
+            cwd: path.resolve(testBasePath, 'workspace'),
           }),
         });
       });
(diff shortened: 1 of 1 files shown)


typpo:
Nice catch!

PR #8041, opened 2026-03-08
fix: make quoted comma parsing consistent across contains assertions

aerosta:
## Summary

Fix inconsistent quoted-value parsing across `contains` assertions.

`handleContainsAny` already supports quoted comma-containing values, but
`handleIContainsAny`, `handleContainsAll`, and `handleIContainsAll` still use naive comma splitting. This caused inconsistent behavior for inputs like `"hello, world","foo"`.

## Changes

- Extracted quoted comma-separated parsing logic into a shared helper
- Reused the same parsing behavior across all 4 `contains-*` variants
- Replaced remaining naive `.split(',').map(v => v.trim())` calls with the shared parser

## Tests

Added regression coverage in `test/assertions/contains.test.ts` for:

- quoted comma-containing values across all 4 variants
- existing non-quoted comma splitting behavior
- edge cases such as empty string, single value, and no commas

Also revised 1 existing expectation that depended on the previous buggy empty-segment-splitting behavior.

```bash
npx vitest run test/assertions/contains.test.ts

### DIFF
--- a/src/assertions/contains.ts
+++ b/src/assertions/contains.ts
@@ -3,9 +3,57 @@ import invariant from '../util/invariant';
 import type { AssertionParams, GradingResult } from '../types/index';
 
 function parseCommaSeparatedValues(value: string): string[] {
-  return (
-    value.match(/(".*?"|[^,]+)(?=\s*,|\s*$)/g)?.map((v) => v.trim().replace(/^"|"$/g, '')) ?? []
-  );
+  const results: string[] = [];
+  let i = 0;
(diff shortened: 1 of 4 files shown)


mldangelo:
Thank you @aerosta! Please reach out (michael @ promptfoo.dev) if you would like some free promptfoo swag!

aerosta:
> Thank you @aerosta! Please reach out (michael @ promptfoo.dev) if you would like some free promptfoo swag!

Thanks @mldangelo

PR #8045, opened 2026-03-08
fix(app): enforce strategy config limits via onChange clamping

MrFlounder:
## Summary
- HTML `min`/`max` attributes on `<input type="number">` are advisory — users can type any value directly, bypassing the limits
- Adds `Math.min`/`Math.max` clamping in `onChange` handlers for all strategy config numeric inputs (maxTurns, numIterations, tree params)
- Uses the same auth-aware limit constants already defined in the component

## Test plan
- [ ] Open strategy config dialog for each multi-turn strategy (crescendo, goat, custom, hydra, generic multi-turn)
- [ ] Try typing a number above the max — value should clamp to max
- [ ] Try typing a number below the min — value should clamp to min
- [ ] Verify spinner arrows still work correctly
- [ ] Repeat for jailbreak:tree params (maxDepth, maxAttempts, maxWidth, branchingFactor, maxNoImprovement)
- [ ] Repeat for jailbreak and jailbreak:meta numIterations

### DIFF
--- a/src/app/src/pages/redteam/setup/components/StrategyConfigDialog.test.tsx
+++ b/src/app/src/pages/redteam/setup/components/StrategyConfigDialog.test.tsx
@@ -310,8 +310,8 @@ describe('StrategyConfigDialog', () => {
   });
 
   it("should render the number of iterations input and update local config when changed for the 'jailbreak' strategy", () => {
-    const initialConfig = { numIterations: 10 };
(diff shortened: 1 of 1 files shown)


faizanminhas (approved):

faizanminhas:
Looks good, but I would add this functionality into a shared component based on the radix input and not make it specific to this page / component

PR #8062, opened 2026-03-09
fix(providers): omit unset temperature in OpenAI APIs

### LINKED ISSUE #8044: Bug: PR #7323 introduced hardcoded temperature:0 sent to all non-reasoning models even when not configured — silence is better than a wrong default
## Context

This is a follow-up to #7322 and PR #7323.

The **original intent** of #7322 was: *"I explicitly set `temperature: 0` in my config, but promptfoo drops it because of a falsy check."* The fix should have been: **forward `temperature` only when the user explicitly sets it**.

Instead, PR #7323 changed the gate from `if (temperature)` to `if (temperature !== undefined)` — which is correct — but left the **hardcoded default `0`** in place. This means `temperature: 0` is now sent to **every non-reasoning model**, even when the user never configured any temperature at all.

## The problem

```typescript
// src/providers/openai/chat.ts (current)
const temperature = this.supportsTemperature()
  ? config.temperature ?? getEnvFloat("OPENAI_TEMPERATURE", 0)  // ← hardcoded 0 fallback
  : undefined;

...temperature !== undefined ? { temperature } : {},  // ← 0 passes, always sent
```

When no temperature is configured and `OPENAI_TEMPERATURE` env var is not set, the expression evaluates to `0`, which then passes the `!== undefined` gate and is sent explicitly in the body.

## Why "silence is better than 0"

For providers like LiteLLM (and other OpenAI-compatible proxies), **omitting `temperature` from the body** allows the proxy to apply:
- Model-specific temperature defaults (e.g. a thinking/reasoning model that needs a specific value)
- Deployment-level defaults configured on the proxy
- Upstream API defaults

Sending `temperature: 0` explicitly **overrides all of these**, even though the user never asked for it. This is especially harmful for:
- Models that have their own optimal temperature (e.g. embedding models, thinking models)
- LiteLLM proxy deployments where temperature is managed centrally
- Any model where `0` is not a sensible default

## Before vs after PR #7323

| Scenario | Before (≤ 0.120.17) | After (≥ 0.120.18) | Desired |
|----------|--------------------|--------------------|---------|
| `config.temperature: 0` | ❌ omitted (original bug) | ✅ `0` sent | ✅ `0` sent |
| `config.temperature: 0.7` | ✅ `0.7` sent | ✅ `0.7` sent | ✅ `0.7` sent |
| No temperature configured | ✅ omitted (accidentally correct) | ❌ `0` always sent | ✅ omitted |

The fix accidentally broke the "no temperature configured" case by making the hardcoded `0` default visible.

## Suggested fix

Only send temperature when the user explicitly configured it or set `OPENAI_TEMPERATURE`:

```typescript
// Only apply env var default when the variable is actually set
const envTemp = process.env.OPENAI_TEMPERATURE !== undefined
  ? getEnvFloat("OPENAI_TEMPERATURE")
  : undefined;

const temperature = this.supportsTemperature()
  ? config.temperature ?? envTemp   // undefined when neither is set → omitted from body
  : undefined;

// gate unchanged:
...temperature !== undefined ? { temperature } : {},
```

This way:
- `config.temperature: 0` → `0` sent ✅ (original bug fixed, preserved)
- `config.temperature: 0.7` → `0.7` sent ✅
- Nothing configured → omitted ✅ (proxy/model decides)
- `OPENAI_TEMPERATURE=0.5` env var → `0.5` sent ✅

## Empirical evidence

Confirmed with 20+ models via echo server sniff test: every non-reasoning `litellm:*` model receives `"temperature": 0` in the body even with no temperature configured anywhere.

## System information

- Promptfoo version: 0.120.18+ (regression introduced by PR #7323)
- Provider: `litellm:*` / `openai:*` (any provider using `OpenAiChatCompletionProvider`)
- Related: #7322, PR #7323

mldangelo:
## Summary
- omit `temperature` from OpenAI-compatible requests when it is not configured and `OPENAI_TEMPERATURE` is unset
- preserve explicit `temperature: 0` and env-driven temperature across OpenAI, LiteLLM, and Azure chat/responses/completion paths
- add regression tests for omitted, explicit-zero, and env-driven temperature behavior

Fixes #8044.

## Validation
- `npx vitest run test/providers/openai/chat.test.ts`
- `npx vitest run test/providers/openai/responses.test.ts`
- `npx vitest run test/providers/openai/completion.test.ts`
- `npx vitest run test/providers/azure/chat.test.ts`
- `npx vitest run test/providers/azure/responses.test.ts`
- `npx vitest run test/providers/azure/completion.test.ts`
- `npm run tsc`

## E2E QA
Ran 21 real `promptfoo eval` cases against a local OpenAI-compatible capture server using `npm run local -- eval --no-cache`.

Covered providers:
- `openai` chat, responses, completion
- `litellm` chat
- `azure` chat, responses, completion

Covered scenarios:
- no temperature configured
- `temperature: 0`
- `OPENAI_TEMPERATURE=0.5`

Observed results:
- unset temperature was omitted in every unset case
- explicit `0` was forwarded in every explicit-zero case
- env temperature was forwarded in every env-driven case
- each eval exited successfully and emitted exactly one request to the capture server

## Notes
- this PR intentionally does not change `max_tokens` defaults
- the workspace had unrelated local changes; only the provider and test files for this fix were committed

### DIFF
--- a/examples/provider-opencode-sdk/README.md
+++ b/examples/provider-opencode-sdk/README.md
@@ -1,6 +1,6 @@
 # provider-opencode-sdk (OpenCode SDK Examples)
 
-The OpenCode SDK provider enables you to run agentic evals through OpenCode, an open-source AI coding agent for the terminal with support for 75+ LLM providers.
+The OpenCode SDK provider runs agentic evals through OpenCode, an open-source coding agent with support for hosted and local model providers.
 
 ```bash
 npx promptfoo@latest init --example provider-opencode-sdk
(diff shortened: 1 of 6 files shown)


ghost:
**No test execution environment matched**

[View output ↗](https://app.usetusk.ai/app/testing-commit-check/42da8bc7-5fc4-4b22-893e-20099b736306?client=e91e3433-8fb4-4e98-b233-bba492f4909c)

---
<p>

| 81abe6ebdea1018954d67bc8ce14afe5e4bf97f7 | No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/42da8bc7-5fc4-4b22-893e-20099b736306?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Mar 9, 2026 4:06PM |

</p>
</details>

mldangelo:
Superseded by a more targeted approach that only fixes the LiteLLM provider path instead of changing defaults for all OpenAI/Azure providers. See new PR.

PR #8064, opened 2026-03-09
fix(providers): omit default temperature for LiteLLM proxy provider

### LINKED ISSUE #8044: Bug: PR #7323 introduced hardcoded temperature:0 sent to all non-reasoning models even when not configured — silence is better than a wrong default

mldangelo:
## Summary

- Fixes #8044 — LiteLLM proxy providers now omit `temperature` from the request body when not explicitly configured, allowing the proxy to apply its own model-specific defaults
- Adds `omitDefaults` config option to `OpenAiCompletionOptions` that skips hardcoded fallback values (e.g. `temperature: 0`)
- LiteLLM sets `omitDefaults: true` by default; direct `openai:` and `azureopenai:` providers are **unchanged**

### Why this approach?

The previous attempt (#8062) removed the hardcoded `temperature: 0` default from **all** OpenAI/Azure providers. This would have changed behavior for every user — OpenAI's server default is `temperature: 1`, so users who never set temperature would go from deterministic to random outputs. Internal grading providers (llm-rubric, factuality, etc.) would also become nondeterministic.

This PR instead targets only the LiteLLM provider path via a config flag, preserving existing behavior for all other providers.

### Behavior matrix

| Provider | Config | Before | After |
|----------|--------|--------|-------|
| `litellm:*` | no temperature | `temperature: 0` sent | **omitted** (proxy decides) |
| `litellm:*` | `temperature: 0` | `temperature: 0` sent | `temperature: 0` sent |
| `litellm:*` | `temperature: 0.7` | `temperature: 0.7` sent | `temperature: 0.7` sent |
| `litellm:*` | `OPENAI_TEMPERATURE=0.5` env | `temperature: 0.5` sent | `temperature: 0.5` sent |
| `openai:*` | no temperature | `temperature: 0` sent | `temperature: 0` sent (unchanged) |
| `openai:*` | `temperature: 0` | `temperature: 0` sent | `temperature: 0` sent (unchanged) |

### Note for other proxy users

Users of other OpenAI-compatible proxies (via `openai:chat:` with custom `apiBaseUrl`) can opt into this behavior by setting `omitDefaults: true` in their provider config:

```yaml
providers:
  - id: openai:chat:my-model
    config:
      apiBaseUrl: http://my-proxy:4000/v1
      omitDefaults: true
```

## Test plan

- [x] Unit tests: 37 LiteLLM tests pass (6 new tests for temperature omission)
- [x] Unit tests: 65 OpenAI chat tests pass (no regression)
- [x] E2E: echo server confirms LiteLLM omits temperature when unconfigured
- [x] E2E: echo server confirms OpenAI still sends `temperature: 0` by default
- [x] E2E: echo server confirms `OPENAI_TEMPERATURE` env var works for both
- [x] E2E: real OpenAI API calls succeed for all scenarios

### DIFF
--- a/src/providers/openai/chat.ts
+++ b/src/providers/openai/chat.ts
@@ -226,10 +226,13 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
     const maxCompletionTokens = isReasoningModel
       ? (config.max_completion_tokens ?? getEnvInt('OPENAI_MAX_COMPLETION_TOKENS'))
       : undefined;
+    const maxTokensDefault = config.omitDefaults
+      ? getEnvString('OPENAI_MAX_TOKENS') !== undefined
+        ? getEnvInt('OPENAI_MAX_TOKENS')
+        : undefined
+      : getEnvInt('OPENAI_MAX_TOKENS', 1024);
     const maxTokens =
-      isReasoningModel || isGPT5Model
-        ? undefined
-        : (config.max_tokens ?? getEnvInt('OPENAI_MAX_TOKENS', 1024));
+      isReasoningModel || isGPT5Model ? undefined : (config.max_tokens ?? maxTokensDefault);
 
     const temperatureDefault = config.omitDefaults
       ? getEnvString('OPENAI_TEMPERATURE') !== undefined
--- a/test/providers/litellm.test.ts
+++ b/test/providers/litellm.test.ts
@@ -474,5 +474,48 @@ describe('LiteLLM Provider', () => {
       });
       expect(provider.config.omitDefaults).toBe(false);
     });
+
+    it('should send default temperature: 0 when omitDefaults is false', async () => {
+      mockFetchWithCache.mockResolvedValue(mockResponse);
+
+      const provider = createLiteLLMProvider('litellm:chat:gpt-4', {
+        config: {
+          config: {
+            apiKey: 'test-key',
+            omitDefaults: false,
+          },
+        },
+      });
+
+      await provider.callApi('Test prompt');
+
+      const call = mockFetchWithCache.mock.calls[0] as [string, { body: string }];
+      const body = JSON.parse(call[1].body);
+
+      expect(body.temperature).toBe(0);
+      expect('temperature' in body).toBe(true);
+      expect(body.max_tokens).toBe(1024);
+      expect('max_tokens' in body).toBe(true);
+    });
+
+    it('should omit max_tokens from request body when not configured', async () => {
+      mockFetchWithCache.mockResolvedValue(mockResponse);
+
+      const provider = createLiteLLMProvider('litellm:chat:gpt-4', {
+        config: {
+          config: {
+            apiKey: 'test-key',
+          },
+        },
+      });
+
+      await provider.callApi('Test prompt');
+
+      const call = mockFetchWithCache.mock.calls[0] as [string, { body: string }];
+      const body = JSON.parse(call[1].body);
+
+      expect(body.max_tokens).toBeUndefined();
+      expect('max_tokens' in body).toBe(false);
+    });
   });
 });


ghost:
**No test execution environment matched**

[View output ↗](https://app.usetusk.ai/app/testing-commit-check/e4ee1ce2-5d7e-471a-9dca-a1232e2f772b?client=e91e3433-8fb4-4e98-b233-bba492f4909c)

---
<p>

| 27a71303aa4dc1d84c0a29032ca68bd527d2ee29 | No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/24ac60a1-a28b-4263-b293-e595128fad4d?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Mar 9, 2026 4:39PM |
| 4625fae148b008cbaa83ce9349a822a628ea773f | No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/2c64c5be-ec81-479b-befe-484b2b1b9f82?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Mar 9, 2026 5:11PM |
| 55fab51b222234849a6dd4b96b69d6d7e647d7bb | No test execution environment matched | [Output](https://app.usetusk.ai/app/testing-commit-check/e4ee1ce2-5d7e-471a-9dca-a1232e2f772b?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Mar 9, 2026 5:38PM |

</p>
</details>

PR #8069, opened 2026-03-09
fix: not-contains-json with schema respects inverse flag

aerosta:
## Summary

`not-contains-json` with a JSON schema was returning the wrong result. When the output contained JSON matching the schema, the assertion passed instead of failing.

The issue is in `src/assertions/json.ts`: `pass = validate(jsonObject)` overwrote the earlier inverse-aware logic, so the `inverse` flag used by `not-*` assertions was ignored during schema validation.

## Fix

This change separates raw schema validation from the final assertion result:

- `valid` stores the raw schema validation result
- `pass` applies the `inverse` flag to that result

The loop still breaks on `valid`, which is correct because iteration should stop once a matching JSON object is found.

## Tests

Added 2 tests in `test/assertions/runAssertion.test.ts` covering these cases:

- matching schema -> `not-contains-json` fails
- non-matching schema -> `not-contains-json` passes

Local Node version is `v22.11.0`, which is outside the repo's supported range, so final validation should come from CI.

### DIFF
--- a/src/assertions/json.ts
+++ b/src/assertions/json.ts
@@ -21,7 +21,7 @@ export function handleIsJson({
     pass = inverse;
   }
 
-  if (pass && renderedValue) {
+  if (parsedJson !== undefined && renderedValue) {
     let validate: ValidateFunction;
     if (typeof renderedValue === 'string') {
       if (renderedValue.startsWith('file://')) {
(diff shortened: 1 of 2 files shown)


mldangelo:
Thanks for another great pr @aerosta! Please reach out if you'd like some swag.

aerosta:
> Thanks for another great pr @aerosta! Please reach out if you'd like some swag.

Thanks @mldangelo  , appreciate it. I sent an email.

PR #8070, opened 2026-03-09
fix: prevent caching error responses in fetchWithCache

aerosta:
## Summary

`fetchWithCache` caches HTTP 200 responses that contain an error body such as `{"error": "rate limit exceeded"}`.

That path is meant to skip caching, but it returns `data`, so `cache.wrap()` stores the error response and poisons the cache for up to 14 days.

The `!response.ok` branch already avoids this by setting `errorResponse` and returning `undefined`. This change makes the error-key path behave the same way.

## Fix

- store the response in `errorResponse`
- return `undefined` instead of `data`

## Tests

Local Node version is outside the repo's supported range, so final validation should come from CI.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -125,12 +125,12 @@ export type FetchWithCacheResult<T> = {
 
 type SerializedFetchResponse = string;
 
-class NonCacheableResponseError extends Error {
-  constructor(public readonly response: SerializedFetchResponse) {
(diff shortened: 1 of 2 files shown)


mldangelo:
Thank you again @aerosta!

aerosta:
> Thank you again @aerosta!

Thanks @mldangelo, happy to contribute.

PR #8072, opened 2026-03-09
fix(redteam): fix Rules of Hooks violation in RiskCategoryDrawer

Ebonsignori:
## Summary
- Fix the root cause of Sentry issue `CLOUD-UI-PROD-Y0`: hooks (`useNavigate`, `useState`, `useMemo`) were called **after** a conditional early return in `RiskCategoryDrawer`, violating React's Rules of Hooks
- Move all hooks above the category validation early return so they are called unconditionally on every render
- Preserve client-side navigation via `useNavigate` (no full page reload)

## Root Cause
The `RiskCategoryDrawer` component had an early return before any hooks were called:

```tsx
// BEFORE (broken - hooks after conditional return)
const categoryName = categoryAliases[category];
if (!categoryName) {
  return null;  // ← early return BEFORE useNavigate/useState/useMemo
}
const navigate = useNavigate();  // ← violates Rules of Hooks
```

When `category` wasn't in `categoryAliases`, the component returned early without calling hooks. On subsequent renders where the category became valid, React saw a different number of hook calls, causing the error.

## Fix
```tsx
// AFTER (correct - all hooks before any early returns)
const navigate = useNavigate();
const [suggestionsDialogOpen, setSuggestionsDialogOpen] = React.useState(false);
// ... other hooks ...
const sortedFailures = React.useMemo(() => ...);

const categoryName = categoryAliases[category];
if (!categoryName) {
  return null;  // ← early return AFTER all hooks
}
```

## Test plan
- [x] Existing tests pass (invalid category test still works with hooks called first)
- [x] "View All Logs" button uses client-side `navigate()` (no page reload)
- [x] Ctrl/Cmd+click opens in new tab via `window.open`
- [x] TypeScript compiles cleanly
- [x] Biome lint passes

### DIFF
--- a/src/app/src/pages/redteam/report/components/RiskCategoryDrawer.test.tsx
+++ b/src/app/src/pages/redteam/report/components/RiskCategoryDrawer.test.tsx
@@ -1,8 +1,14 @@
 import { renderWithProviders } from '@app/utils/testutils';
 import { fireEvent, screen } from '@testing-library/react';
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import { useNavigate } from 'react-router-dom';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
 import RiskCategoryDrawer from './RiskCategoryDrawer';
 import type { AtomicTestCase, EvaluateResult, ResultFailureReason } from '@promptfoo/types';
+
+// Mock dependencies
+vi.mock('react-router-dom', () => ({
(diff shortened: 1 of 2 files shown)


JustinBeckwith (commented):
I am actually quite curious why this was needed?

faizanminhas (approved):
Thank you! Can we add a component test that could render this component to catch any future regressions?

PR #8079, opened 2026-03-10
docs(site): remove careers page

typpo:
Removes the `/careers/` page and the live site/app links that pointed to it.

Also removes careers-specific OG-image generation and cleans up blog CTAs that referenced the page.

Checks: targeted Biome lint, Prettier check, and `git diff --check`.

### DIFF
--- a/src/app/src/components/InfoModal.tsx
+++ b/src/app/src/components/InfoModal.tsx
@@ -8,14 +8,7 @@ import {
   DialogTitle,
 } from '@app/components/ui/dialog';
 import { cn } from '@app/lib/utils';
-import {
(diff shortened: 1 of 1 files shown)


site/blog/100k-users.md
```diff
@@ -67,7 +67,6 @@ Reaching 100,000 users is just the beginning. We are rapidly scaling our team an
 
 We're committed to being the best open-source LLM red teaming and eval tool available. We invite you to:
 
-- Apply to [join our team](https://promptfoo.dev/careers)
```
mldangelo, line 70: Can we change this to the openai careers page?

PR #8084, opened 2026-03-10
fix: render rubric value with vars before grading

### LINKED ISSUE #7861: llm-rubric assertions in defaultTest cannot access test case variables via {{varName}} template syntax
**Describe the bug**
When an `llm-rubric` assertion is defined in `defaultTest`, template variables (e.g. `{{input}}`) are not resolved — the rubric receives the literal string `{{input}}` instead of the test case's variable value. The same rubric works correctly when the assertion is defined inline per test case.

**To Reproduce**

**promptfooconfig.yaml**
  ```yaml
  defaultTest: file://default_test.yaml
  tests:
    - file://test_cases.yaml
```
  default_test.yaml

```yaml
  assert:
    - type: llm-rubric
      value: "Does the output correctly reference the input: {{myVar}}?"
```

  test_cases.yaml
```yaml   
  - vars:
      myVar: "hello world"
```

**Expected behavior**
{{myVar}} in the rubric value is resolved to "hello world" for each test case, the same way it works when the llm-rubric assertion is defined directly inside test_cases.yaml.

**Actual behavior**
The rubric receives the literal text {{myVar}} unresolved, so the grading LLM has no access to the input variable and cannot correctly evaluate the assertion. Some models just grade with 'correct', making it hard to spot this behavior.

**Workaround**
 Inject the input variable into the LLM output via a transform in defaultTest, and reference it from there in the rubric. For example:

  default_test.yaml
  ```yaml
  options:
    transform: |
      const match = output.match(/\{[\s\S]*\}/);
      if (!match) return output;
      try {
        const parsed = JSON.parse(match[0]);
        parsed._ref = (context.vars && context.vars.myVar) || '';
        return output.slice(0, match.index) + JSON.stringify(parsed) + output.slice(match.index + match[0].length);
      } catch(e) { return output; }
  assert:
    - type: llm-rubric
      value: "Does the output correctly reference the input? See the '_ref' field in the JSON output."
```

This is a significant limitation — rubrics defined in defaultTest are intended to apply uniformly across all test cases, but without access to per-case variables they often cannot perform meaningful evaluation.

**Environment**

- Promptfoo version:  0.120.25
- Node version: v24.7.0

Jah-yee:
When llm-rubric assertion is defined in defaultTest, template variables like {{myVar}} were not being resolved - the rubric received the literal string {{myVar}} instead of the test case's variable value.

This fix renders the rubric value with nunjucks using the vars before passing it to the grading LLM context.

Fixes #7861

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -633,9 +633,17 @@ export async function matchesLlmRubric(
   }
 
   const rubricPrompt = await loadRubricPrompt(grading?.rubricPrompt, DEFAULT_GRADING_PROMPT);
+
+  // Render the rubric value with vars so that template variables like {{myVar}}
+  // are expanded before being passed to the grading LLM.
+  const renderedRubric =
+    typeof rubric === 'string'
+      ? nunjucks.renderString(rubric, vars || {})
+      : JSON.stringify(rubric);
+
   const prompt = await renderLlmRubricPrompt(rubricPrompt, {
     output: tryParse(llmOutput),
-    rubric,
+    rubric: renderedRubric,
     ...(vars || {}),
   });
 
@@ -654,7 +662,7 @@ export async function matchesLlmRubric(
     'llm-rubric',
     {
       output: tryParse(llmOutput),
-      rubric,
+      rubric: renderedRubric,
       ...(vars || {}),
     },
     providerCallContext,


mldangelo-oai:
Thanks for the PR. We’re consolidating this fix in #7976, where we’re keeping rubric rendering in the assertion pipeline/UI path and adding regression coverage for defaultTest variables. Closing this duplicate so review stays focused there.

Issue #6951, opened 2026-01-07
Web viewer: “Show charts” toggle missing (Promptfoo 0.120.10, Windows)


theyv:
Title: Web viewer: “Show charts” toggle missing (Promptfoo 0.120.10, Windows)

## Summary
In the Promptfoo web viewer, the “Show charts” toggle/button is completely missing on the eval results page. It used to be available before, but now it’s not visible for any run (latest or historical). 

## Steps to reproduce
1. Run any eval.
2. Open the web viewer (`npx promptfoo view` ).
3. Open the latest run or any historical run.
4. Check the eval page toolbar / top-right controls.

## Expected behavior
A “Show charts” toggle/button is visible and allows enabling charts in the eval results view.

## Actual behavior
No “Show charts” toggle/button is shown anywhere on the eval page. The toolbar shows items like “Columns”, “Table Settings”, and “Eval actions”, but there is no charts toggle.  
(See attached screenshot.)

## Environment
- OS: Windows 11
- Shell: PowerShell
- Promptfoo version: 0.120.10 
- Browser: Opera

## Console / logs
- No errors in the browser console.
- No errors in the terminal output related to the UI (only a Node “ExperimentalWarning: Importing JSON modules…” when checking version).

## Attachments
- Screenshot showing the missing toggle.

<img width="525" height="344" alt="Image" src="https://github.com/user-attachments/assets/525042d2-5f57-43dc-b4b7-b62cd7b1cda9" />

mldangelo:
@theyv I am looking into this for you. Can you confirm how many prompts / providers you have in your eval? Is there any chance that the scores were exactly the same?

theyv:
Okay - I think I understand now
if I set this flag:
--output results.html

there is no charts in promptfoo view
I don't know if it should be like that or not

hdw868:
Still seeing this issue in Promptfoo 0.120.17

JustinBeckwith:
Greetings!  The rendering of this button was conditional, where it only rendered if it it made sense across multiple prompts that can be compared.  It would also only show if there were numerical scores to plot.  To make what's happening a little more clear - we're now always showing the toggle.  If it doesn't make sense to show the charts - we'll just give a message instead.  https://github.com/promptfoo/promptfoo/pull/8087

This is merged, but I'll wait until the next release to mark this one as fixed :) If anyone wants to give it a try before that, I'd love feedback!

PR #8087, opened 2026-03-10
fix(eval): preserve chart default state per eval

JustinBeckwith:
This should resolve, or at least help shine some light on https://github.com/promptfoo/promptfoo/issues/6951.  The issue is that users couldn't see the "Show Charts" button.  The problem is, based on the data, we may choose to just hide it and not give the option with no explanation. Instead, this PR shows a banner that explains why they aren't seeing a chart:

image

Very open to feedback on the approach!  

## Summary
- always keep the charts toggle visible and show an explanatory alert only after the user opens an ineligible chart panel
- recompute default chart visibility when the active eval changes without remounting the entire results view
- add regression coverage for ineligible chart messaging and cross-eval default visibility behavior

## Testing
- npm run l
- cd src/app && pnpm vitest run src/pages/eval/components/Eval.test.tsx src/pages/eval/components/ResultsView.test.tsx

### DIFF
--- a/src/app/src/pages/eval/components/ResultsView.tsx
+++ b/src/app/src/pages/eval/components/ResultsView.tsx
@@ -50,6 +50,7 @@ import type { CopyEvalResponse } from '@promptfoo/types/api/eval';
 import type { VisibilityState } from '@tanstack/table-core';
 
 import type { ActiveView } from './EvalHeader';
+import type { ResultsFilter } from './store';
 
 const Report = React.lazy(() => import('@app/pages/redteam/report/components/Report'));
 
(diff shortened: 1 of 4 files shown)


src/app/src/pages/eval/components/ResultsView.tsx
```diff
@@ -548,16 +547,53 @@ export default function ResultsView({
...
+  const showResultsChartsControl = !isRedteamEval;
+  const shouldRenderResultsChartsByDefault =
+    showResultsChartsControl && window.innerHeight >= 1100 && canRenderResultsCharts;
+  const [hasUserToggledResultsCharts, setHasUserToggledResultsCharts] = React.useState(false);
+  const [renderResultsCharts, setRenderResultsCharts] = React.useState(
+    shouldRenderResultsChartsByDefault,
+  );
+  const previousEvalIdRef = React.useRef(currentEvalId);
+
+  React.useEffect(() => {
```
faizanminhas: Can we do this without all the new useEffects? Would think this can be derived state. Better yet, move it into a dedicated component

JustinBeckwith: Refactored this to remove the chart-specific useEffects from `ResultsView`. The chart toggle/default-open behavior now lives in a dedicated `ResultsChartsSection` component keyed by eval id, so the chart state resets per eval without remounting the full results view. Focused tests still pass.

src/app/src/pages/eval/components/ResultsView.tsx
```diff
@@ -618,236 +688,225 @@ export default function ResultsView({
...
+                    <button
+                      type="button"
+                      onClick={() => setFilterMode('all')}
+                      className="ml-1 hover:bg-muted rounded-full"
+                    >
+                      <X className="size-3" />
+                    </button>
+                  </Badge>
+                )}
+                {filters.appliedCount > 0 &&
```
faizanminhas: This logic should not be in the JSX, extract. Also, do we want to remove the filter chips for red teams? That’s out of scope for this PR

faizanminhas (approved):
Ty!

PR #8124, opened 2026-03-12
feat: allow per-test opt-out of defaultTest assertions

### LINKED ISSUE #6914: Allow disabling of defaultTests for particular test instances
Maybe I missed this in the docs but there doesn't seem to be a way to tell a test NOT to use the default test assertions. I have such a case where the tests expects an empty result from my custom provider but my defaultTest assertions fail if the result is empty. For that single test I'd like to state exactly which assertions to run and ignore defaultTest.

Is that possible?

Maybe an option on a test case that disables defaultTests just for this test

I could create a separate yaml but that is really just a workaround

stablegenius49:
## Summary
- add a per-test `options.disableDefaultAsserts` flag
- skip inheriting `defaultTest.assert` when that flag is enabled
- add evaluator coverage showing other `defaultTest` defaults still merge normally

## Why
Issue #6914 asks for a way to stop a specific test from inheriting default assertions when that test needs to define its own exact assertion set. This keeps the fix narrow and backwards-compatible: only `defaultTest.assert` is skipped, while other defaults like vars, metadata, threshold, and provider options still apply.

## Testing
- `./node_modules/.bin/vitest run test/evaluator.test.ts -t "defaultTest"`

Closes #6914


ianw-oai (approved):
Docs update looks good.

ianw-oai (approved):
Approved after updating with latest main.

ianw-oai (approved):
Approved after docs update and merge from main.

ianw-oai (approved):
Approved latest head after updating with main.

ianw-oai (approved):
Approved latest head after updating with main.

ianw-oai (approved):
Approved latest head after updating with current main.

ianw-oai (approved):
Approved latest head after updating with current main.

PR #8127, opened 2026-03-13
fix: llm-rubric matcher logic

schipiga:
Hi folks,
I believe it's quite important fix, because as I see initial logic, it could pass a test if llm returns no `score` and `pass` in `llm-rubric` response. I understand that with modern llm it's quite abnormal case, but as llm isn't deterministic and based on probabilistic approach, I think here is better to follow zero-trust conception.
Instead of initial logic, I propose:
- check presence of `score` and `pass` fields both;
- consider test as `false` if no `pass` field
- rely only on `score >= threshold` if `threshold` is provided but `pass` field is missing 
- other logic as is


mldangelo-oai:
Thanks for identifying the malformed llm-rubric output issue. I audited this branch against current main and reproduced a remaining behavior gap in a real deterministic eval: a grader returning {pass: null, score: 0.6} with threshold: 0.5 still fails, even though the valid score should drive the result when no boolean verdict is supplied. PR #8130 now supersedes this patch: it treats null verdict/score fields as omitted, retains privacy-safe malformed-output diagnostics, applies threshold behavior consistently including remote grading, updates documentation and the runnable example, and has full CI plus end-to-end eval validation. Closing this duplicate in favor of #8130.

PR #8131, opened 2026-03-13
fix: use word boundary regex for _conversation detection

### LINKED ISSUE #7845: _conversation variable detection uses naive substring match, causing false concurrency=1 forcing
**Describe the bug**

The `_conversation` variable detection uses a naive `.includes('_conversation')` substring check against `prompt.raw`. This causes false positives whenever the prompt source contains `_conversation` as a **substring** of any other identifier, string, comment, or key — even though the prompt never actually uses the `_conversation` template variable.

When triggered, concurrency is silently forced to 1 with the message:
```
Setting concurrency to 1 because the _conversation variable is used.
```

**To Reproduce**

You can use a simple plain text prompt like this:

```text
Summarize the pre_conversation_context for the user: {{ question }}
```

**Expected behavior**

The eval should run with the requested concurrency since the `_conversation` template variable is not used. The detection should match `_conversation` as a standalone identifier, not as a substring of other words.

Suggested fix - maybe use a word-boundary regex?

**System information:**
```
{
  "version": "0.120.25",
  "platform": {
    "os": "darwin",
    "release": "25.2.0",
    "arch": "x64",
    "nodeVersion": "v22.16.0"
  },
  "env": {
    "telemetryDisabled": false,
    "telemetryDebug": false
  },
  "configInfo": {
    "configExists": false,
    "configContent": null
  }
}
```

Jah-yee:
The _conversation variable detection was using naive substring matching (.includes("_conversation")), which caused false positives when prompts contained substrings like "pre_conversation_context".

This fix adds a word boundary regex check (/\b_conversation\b/) to properly detect only actual _conversation variable usage, preventing unnecessary concurrency=1 forced execution.

Fixes #7845

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -326,7 +326,7 @@ export async function runEval({
   const fileMetadata = collectFileMetadata(test.vars || vars);
 
   const conversationKey = `${provider.label || provider.id()}:${prompt.id}${test.metadata?.conversationId ? `:${test.metadata.conversationId}` : ''}`;
-  const usesConversation = prompt.raw.includes('_conversation');
+  const usesConversation = prompt.raw.includes('_conversation') && /\b_conversation\b/.test(prompt.raw);
   if (
     !getEnvBool('PROMPTFOO_DISABLE_CONVERSATION_VAR') &&
     !test.options?.disableConversationVar &&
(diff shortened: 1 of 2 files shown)


mldangelo-oai:

PR #8150, opened 2026-03-14
fix: avoid false _conversation detection

### LINKED ISSUE #7845: _conversation variable detection uses naive substring match, causing false concurrency=1 forcing

Sakuranda:
## Summary

- fix `_conversation` detection so only real `_conversation` template references trigger the special-case behavior
- avoid false positives for similarly named variables such as `pre_conversation_context`
- add regression coverage for both the template helper and evaluator concurrency behavior

## Problem

`promptfoo` currently checks for `_conversation` using a plain substring match.

That means prompts like:

```nunjucks
{{ pre_conversation_context }}
```

can be treated as if they were using the special `_conversation` runtime variable.

The visible effect is that evaluation concurrency can be forced down to `1` even though the prompt does not actually depend on `_conversation`.

## Reproduction

1. Create an eval that uses a normal variable with `_conversation` as part of its name, such as `pre_conversation_context`
2. Run the eval with `maxConcurrency > 1`
3. Observe that promptfoo incorrectly treats the prompt as using `_conversation` and serializes execution

## Fix

- added `templateReferencesVariable()` in `src/util/templates.ts`
- detect `_conversation` by checking the root template variable name instead of using `String.includes('_conversation')`
- updated all evaluator call sites that use `_conversation` detection:
  - runtime conversation var injection
  - concurrency forcing logic
  - telemetry feature detection

This keeps valid references like `{{ _conversation[0].output }}` working while preventing false positives from similarly named variables.

## Testing

Passed:

```bash
npx vitest run test/util/templates.test.ts
npx vitest run test/evaluator.test.ts
npm run build
```

Additional note:

- I also attempted `npm test` for the full suite.
- In this environment, unrelated existing test failures appeared in server route suites, and the run later blocked on an interactive `Work email:` prompt.
- Those failures/blockers were not caused by this change; the targeted regression tests above and the full build both passed.

## Files changed

- `src/util/templates.ts`
- `src/evaluator.ts`
- `test/util/templates.test.ts`
- `test/evaluator.test.ts`

Fixes #7845

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -67,6 +67,7 @@ import {
   isProviderAllowed,
 } from './util/provider';
 import { promptYesNo } from './util/readline';
+import { templateReferencesVariable } from './util/templates';
 import { sleep } from './util/time';
 import { TokenUsageTracker } from './util/tokenUsage';
 import {
@@ -326,7 +327,7 @@ export async function runEval({
   const fileMetadata = collectFileMetadata(test.vars || vars);
 
   const conversationKey = `${provider.label || provider.id()}:${prompt.id}${test.metadata?.conversationId ? `:${test.metadata.conversationId}` : ''}`;
-  const usesConversation = prompt.raw.includes('_conversation');
+  const usesConversation = templateReferencesVariable(prompt.raw, '_conversation');
   if (
     !getEnvBool('PROMPTFOO_DISABLE_CONVERSATION_VAR') &&
     !test.options?.disableConversationVar &&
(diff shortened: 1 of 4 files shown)


mldangelo-oai:

PR #8173, opened 2026-03-15
feat(assertions): share trajectory context

typpo:
Adds a row-scoped trace and trajectory cache for assertion execution.
Exposes `trace` and `trajectory` to built-in and scripted assertions and updates trajectory assertions to consume the shared normalized artifact.
Validated with `npm run tsc -- --noEmit`, targeted Vitest coverage, and `CI=1 npm test`.


src/assertions/index.ts
```diff
@@ -409,7 +546,7 @@ export async function runAssertion({
   latencyMs,
   providerResponse,
   traceId,
-  traceData,
+  sharedTraceArtifacts,
```
mldangelo-oai, line 549: [P1] Preserve the public `traceData` parameter

`runAssertion` is exported as a core programmatic API, and both the merge base and current `main` accept `traceData?: TraceData | null`; this replacement breaks typed callers while untyped callers have their supplied trace silently ignored. With `{ traceId, traceData }`, this head fetches from the trace store instead, so an external trace can disappear or be replaced, and explicit `null` no longer suppresses fetching. The JSDoc still advertises the removed option. Keep `traceData` in a stable public parameter type, seed the per-call resolver from it while distinguishing `undefined` from `null`, and keep `sharedTraceArtifacts` behind an internal batch helper. Add declaration and zero-store-read runtime regressions.

src/assertions/trajectoryUtils.ts
```diff
@@ -281,22 +272,25 @@ function isMessageSpan(span: TraceSpan): boolean {
...
 
-      const endDiff =
-        (left.span.endTime ?? left.span.startTime) - (right.span.endTime ?? right.span.startTime);
+      const endDiff = (left.endTime ?? left.startTime) - (right.endTime ?? right.startTime);
       if (endDiff !== 0) {
         return endDiff;
       }
 
-      return left.index - right.index;
+      const nameDiff = left.name.localeCompare(right.name);
```
mldangelo-oai, line 286: [P2] Preserve source order when span timestamps tie

These lexical name/span-ID tie-breakers replace the established input-index tie-break used by the merge base, #8140, and current `main`. Equal effective timestamps are realistic with millisecond/coarse clocks and missing end times; names and span IDs do not encode causal order. I reproduced a source sequence `[z-tool, a-tool]` becoming `[a-tool, z-tool]`, which can silently flip `trajectory:tool-sequence`, goal summaries, and custom trajectory assertions. Restore the `(span, index)` decoration and use `left.index - right.index` as the final tie-break. Cover reverse-lexical names/IDs and an end-to-end sequence assertion.

src/assertions/index.ts
```diff
@@ -362,6 +364,141 @@ export function getAssertionBaseType(assertion: Assertion): AssertionType {
...
+        return trace ? createTraceTrajectory(trace) : undefined;
+      })();
+    }
+
+    return trajectoryPromise;
+  };
+
+  return {
+    async attachToContext(context) {
+      const [trace, trajectory] = await Promise.all([resolveTrace(), resolveTrajectory()]);
```
mldangelo-oai, line 491: [P2] Clone shared artifacts before exposing them to assertions

The memoized trajectory and trace are assigned by reference to every assertion context in the row, while assertions run concurrently. In-process JavaScript, file, and package assertions can legitimately mutate that context—for example by sorting or truncating `context.trajectory.steps`—so sibling `trajectory:*`/trace assertions can grade modified data instead of the fetched trace. I reproduced this through `runAssertions`: a JS assertion truncating a two-step trajectory made sibling step-count and trace-count assertions observe zero. Keep memoizing fetch/normalization work, but `structuredClone` the resolved artifacts per context (including nested steps/attributes), or expose deeply immutable snapshots. Add a concurrent mutator regression and identity/isolation assertions.

mldangelo-oai (commented):
Reviewed exact head `02eb56cfb9037b07a1a41e3b1af825f9c6eb8083` in comment-only mode. This work is partially, not fully, superseded: #8242 already ships one row-scoped trace preload and `context.trace`, but current `main` still lacks a typed/shared `TraceTrajectory`, one normalization per row, and `context.trajectory` for scripted assertions. I found three regressions in the surviving implementation, noted inline: removal of the public `runAssertion({ traceData })` contract, fabricated ordering for equal-time spans, and shared mutable artifacts across concurrent assertions. The exact draft is now 510 commits behind, `CONFLICTING`/`DIRTY`, with six changed-in-both paths and four textual conflicts; current `main` has substantial newer trajectory matcher, tool-classification, tracing, assertion-context, repeat, and linkage behavior. I recommend retiring this historical branch and re-deriving only the unique immutable trajectory-context delta in a fresh current-main-based successor, rather than merging, rebasing, or resolving the branch wholesale. Preserve current source-order semantics, the public `traceData` input, and all newer main behavior; document the versioned trajectory contract if it remains public. The old exact-head required checks are complete and green but stale. No branch edits or pushes were made.

PR #8198, opened 2026-03-17
fix: support separate input/output cost overrides in provider config

### LINKED ISSUE #8184: bug: config.cost override applies identical rate to both input and output token pricing
## Bug

When a user sets `config.cost` to override pricing, the same value is used for **both** input and output token costs. Most LLM APIs charge different rates for input vs output tokens (e.g., GPT-4o: input $2.50/MTok, output $10.00/MTok — a 4x difference).

## Root cause

This pattern exists in `src/providers/shared.ts` and is duplicated across 6 other provider files:

**`src/providers/shared.ts:65-66`**:
```typescript
const inputCost = config.cost ?? model.cost.input;
const outputCost = config.cost ?? model.cost.output;
//     ^^^^^^^^^^^              ^^^^^^^^^^^  same value for both
```

Same pattern in:
- `src/providers/openai/util.ts:582-583`
- `src/providers/deepseek.ts:56-57`
- `src/providers/xai/chat.ts:309-310`
- `src/providers/hyperbolic/chat.ts:220-221`
- `src/providers/anthropic/util.ts:279-280`
- `src/providers/google/util.ts:62-63, 69-70`

Additionally, `config.audioCost` has the same issue for audio input/output (`openai/util.ts:587-588`).

## Reproduction

```yaml
providers:
  - id: openai:gpt-4o
    config:
      cost: 0.000005  # Intended as "per-token override"
tests:
  - vars:
      input: "Hello"  # Short input, long output expected
```

Expected: input and output costs calculated separately.
Actual: `$0.000005/token` applied to both input and output, making cost reports inaccurate for models with asymmetric pricing.

## Impact

For a prompt with 1000 input tokens and 1000 output tokens on GPT-4o:
- **Correct cost**: $2.50/M × 1000 + $10.00/M × 1000 = $0.0125
- **With `config.cost: 0.000005`**: $5.00/M × 1000 + $5.00/M × 1000 = $0.01

Cost reports can be off by up to **4x** depending on the input/output ratio and model pricing.

## Suggested approach

Consider supporting separate input/output cost overrides:
```typescript
const inputCost = config.inputCost ?? config.cost ?? model.cost.input;
const outputCost = config.outputCost ?? config.cost ?? model.cost.output;
```

This is backward-compatible — `config.cost` still works as a single override, but users can also specify `inputCost` / `outputCost` separately.

Note: This would also require adding `inputCost` and `outputCost` to the provider config type definition in `src/types/index.ts`.

Happy to submit a PR if this approach is acceptable.

Jah-yee:
When a user sets config.cost to override pricing, the same value was applied to both input and output token costs. Most LLM APIs charge different rates for input vs output tokens (e.g., GPT-4o: $2.50/MTok input, $10.00/MTok output - a 4x difference).

This fix adds support for separate inputCost/outputCost and audioInputCost/audioOutputCost overrides in ProviderConfig, while maintaining backward compatibility with the existing cost override.

Example config now supports:
```yaml
providers:
  - id: openai:gpt-4o
    config:
      cost: 0.000005      # applies to both (backward compatible)
      inputCost: 0.0000025 # override input separately
      outputCost: 0.00001  # override output separately
```

Fixes #8184


mldangelo-oai:
Thanks again for working on this and for helping move #8184 forward. We consolidated the current-main implementation into #8254 so review and CI can happen in one place, so I’m going to close this duplicate PR as well.

Please follow along there: https://github.com/promptfoo/promptfoo/pull/8254

PR #8200, opened 2026-03-17
fix: crossSessionLeak grader handles missing metadata gracefully

### LINKED ISSUE #8183: bug: crossSessionLeak grader silently searches for string "undefined" when metadata is missing
## Bug

## Root cause

```typescript
```

## Reproduction

## Impact

## Suggested fix

```typescript

return {
  grade: {
    pass: !match,
    score: match ? 0 : 1,
    reason: match
      : stringToSearchFor
  },
  rubric: '',
};
```

Jah-yee:
When metadata.crossSessionLeakMatch is missing, the grader was passing
undefined to String.prototype.includes(), which coerces to the string
"undefined", causing false positives/negatives.

This fix:
- Uses ?? "" to handle missing metadata
- Checks Boolean(stringToSearchFor) before calling includes()
- Provides a clear reason when metadata is missing

Fixes #8183

(PR created from fork: https://github.com/Jah-yee/promptfoo/pull/1)


mldangelo:
Thanks for taking a look at #8183! Going to close this one since the diff here is touching provider cost/reasoning config (deepseek, google, hyperbolic, openai/responses) rather than `src/redteam/plugins/crossSessionLeak.ts` — looks like the branch ended up mixed. The underlying issue has since been resolved in #8238. If the provider changes here are still relevant, feel free to reopen them as a separate PR with its own title/scope. Really appreciate the effort 🙏

PR #8206, opened 2026-03-18
fix(assertions): apply weights to named scores in assertion results

aerosta:
`AssertionsResult` applies assertion weights to the overall score; however, `namedScores` currently accumulates raw scores without applying these weights.

As a result, per-metric values can be incorrect when assertions use different weights. This affects downstream consumers of `namedScores`, including the eval table, reports, and custom scoring functions that rely on per-metric values.

This change applies weights consistently when accumulating `namedScores`, tracks per-metric weight totals, and normalizes the final values in `testResult()` so the returned scores are weighted averages.

Changes in `src/assertions/assertionsResult.ts`:
- add `namedScoreWeights` to track per-metric accumulated weights
- apply assertion weight when accumulating `namedScores`
- normalize per-metric values before returning from `testResult()`
- pass normalized `namedScores` to `scoringFunction`

Tests in `test/assertions/assertionsResult.test.ts`:
- assertions with different weights and different metrics
- assertions with equal weights on the same metric
- assertion with `weight: 0` to verify the normalization guard


mldangelo-oai (approved):
Reviewed the weighted named-score follow-up after the latest push. The prompt-level aggregation now preserves the existing numerator/count contract while carrying a separate weighted denominator, and the targeted backend, frontend, smoke, and CLI checks passed on the updated head.

mldangelo-oai (approved):
Re-reviewed after fixing the named score weight CI regression and syncing the public docs.

PR #8209, opened 2026-03-18
fix: handle missing crossSessionLeakMatch metadata gracefully

### LINKED ISSUE #8183: bug: crossSessionLeak grader silently searches for string "undefined" when metadata is missing

Jah-yee:
When metadata.crossSessionLeakMatch is undefined, the grader was coercing undefined to the string "undefined" and searching for it, causing false positives (if output contains "undefined") and false negatives (if metadata is missing).

This fix:
- Uses ?? \"\" to default missing metadata to empty string
- Only performs the search if stringToSearchFor is truthy
- Returns appropriate reason messages for each case

Fixes #8183


mldangelo:
Thanks for the contribution! Closing this one since the diff here adds recursive `file://` reference processing in `src/evaluatorHelpers.ts` rather than touching the cross-session leak grader. The original issue (#8183) was fixed in #8238. The nested `file://` loading looks like an interesting feature on its own though — if you'd like to pursue it, please open a fresh PR with a title/scope matching that change so it can be reviewed on its own merits. Appreciate you jumping in!

PR #8218, opened 2026-03-18
fix: handle missing crossSessionLeakMatch metadata in grader

### LINKED ISSUE #8183: bug: crossSessionLeak grader silently searches for string "undefined" when metadata is missing

Jah-yee:
When metadata.crossSessionLeakMatch is undefined, the grader silently searches for the literal string 'undefined' in LLM output, causing false positives and bypassing security checks.

This fix:
1. Uses `?? ''` to default missing metadata to empty string
2. Only searches when stringToSearchFor is non-empty  
3. Returns appropriate reason message when metadata is missing

Fixes: #8183

### DIFF
--- a/src/redteam/plugins/crossSessionLeak.ts
+++ b/src/redteam/plugins/crossSessionLeak.ts
@@ -1,5 +1,6 @@
 import dedent from 'dedent';
 import logger from '../../logger';
+import invariant from '../../util/invariant';
 import { extractJsonObjects } from '../../util/json';
 import { getNunjucksEngine } from '../../util/templates';
 import { MULTI_TURN_STRATEGIES } from '../constants/strategies';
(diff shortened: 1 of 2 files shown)


mldangelo:
Thanks for tackling this! Closing as superseded by #8238, which already landed a fix on main using a nullish-coalesce + `Boolean()` guard plus tests for both the 'output contains the literal string undefined' and 'metadata absent' cases.

The `invariant()` approach here is a reasonable alternative, but we opted for the non-throwing path: a grader that throws can abort or pollute a larger eval batch when a single test case is misconfigured, whereas returning a normal `GradingResult` keeps the run healthy and lets the misconfiguration show up per-row. Really appreciate you writing up the tests and thinking through the edge cases — that kind of analysis is exactly what we need on security-sensitive graders. 🙏

PR #8239, opened 2026-03-20
fix(redteam): pass full conversation history to crescendo grader

### LINKED ISSUE #8174: Red team rubric evaluation does not include all messages context using crescendo strategy
**Describe the bug**
When running a red team scan using the `crescendo` strategy the evaluation says that the system returned user information and none of them were mentioned in the user's request, however they were mentioned by the user in the first message, which means the assertion is not aware of all the messages exchanged in the conversation.

**To Reproduce**
I ran the red team against my application HTTPS endpoint with the following

```
 plugins:
    - id: pii:direct # Tests for direct PII exposure vulnerabilities
  strategies:
    - id: crescendo
      config:
        stateful: true
```

**Expected behavior**
Using strategies with multi turn conversation the LLM evaluation must include the whole conversation details as part of the context of LLM rubric the assertion.

**Screenshots**

Evaluation failure reason
Image

First message in the crescendo conversation includes the users information
Image

- Promptfoo version: 0.121.2

howardpen9:
## Summary
- Fix crescendo strategy evaluation to include full multi-turn conversation context when grading responses
- Previously only the latest response was passed to the LLM rubric evaluator, causing false-positive assertions when context from earlier turns was needed (e.g., user-provided PII in turn 1 flagged as leaked in turn 3)

## Changes
- Format `redteamHistory` (all prior turns) plus the current turn into a structured conversation string before passing to `grader.getResult()`
- First turn (no history) falls back to original single-response behavior
- No breaking changes: existing providers that do not use multi-turn strategies behave as before

## Test plan
- [ ] Run existing crescendo test suite (`test/redteam/strategies/crescendo.test.ts`)
- [ ] Verify multi-turn PII scenario no longer produces false-positive assertion
- [ ] Confirm single-turn grading unchanged (fallback path)

Closes #8174

### DIFF
--- a/src/redteam/grading/types.ts
+++ b/src/redteam/grading/types.ts
@@ -10,7 +10,7 @@ import type { RedteamHistoryEntry } from '../types';
 export interface RedteamGradingContext {
   traceContext?: TraceContextData | null;
   traceSummary?: string;
-  // Multi-turn conversation context for graders that need provenance across turns.
+  // Prior multi-turn conversation context for graders that need provenance across turns.
   redteamHistory?: RedteamHistoryEntry[];
   conversationHistory?: Array<Pick<RedteamHistoryEntry, 'prompt' | 'output'>>;
   conversationTranscript?: string;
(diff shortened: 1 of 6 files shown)


wholley-oai:
@zcrab-oai thoughts?

PR #8251, opened 2026-03-21
fix(evaluator): detect standalone conversation templates

Gujiassh:
## Summary

- broaden `_conversation` template detection so valid Nunjucks expressions with object literals still trigger the concurrency safeguard
- add focused regression coverage for object-literal template expressions while preserving the substring-only non-match behavior
- keep the standalone-template detection intent intact without regressing the original false-positive fix

## Testing

- `npx vitest run test/evaluator.test.ts -t \"_conversation is used as a standalone template variable|_conversation appears only as a substring|object-literal template expressions\"`
- `npm run l`

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -83,6 +83,13 @@ import { type TransformContext, TransformInputType, transform } from './util/tra
 import type { SingleBar } from 'cli-progress';
 import type winston from 'winston';
 
+const CONVERSATION_TEMPLATE_VAR_REGEX =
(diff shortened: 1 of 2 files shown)


mldangelo-oai (changes requested):
The regex stops {{ ... }} matching at the first }, so valid Nunjucks expressions containing object literals before _conversation are missed. That means _conversation injection and forced concurrency can still be skipped for valid templates; please add coverage for that case.

Gujiassh:
Follow-up note: this branch is now rebased onto current main, the evaluator-focused tests pass locally, and the same previous-month redteam test stabilization used on the other refreshed promptfoo branches has been applied here as well.

mldangelo-oai:

PR #8252, opened 2026-03-21
fix(redteam): clarify remote generation guidance

Gujiassh:
## Summary
- update the Hydra and Iterative Meta remote-generation error message so it says remote generation is disabled for the current configuration instead of implying `OPENAI_API_KEY` is always the reason
- keep the actual remote-generation gating logic unchanged and scope this follow-up to clearer user guidance only
- update the Hydra constructor test to match the new wording

## Validation
- `npx vitest run test/redteam/providers/hydra/index.test.ts test/redteam/providers/iterativeMeta.test.ts`
- `npx @biomejs/biome check src/redteam/providers/hydra/index.ts src/redteam/providers/iterativeMeta.ts test/redteam/providers/hydra/index.test.ts`

### DIFF
--- a/src/redteam/providers/hydra/index.ts
+++ b/src/redteam/providers/hydra/index.ts
@@ -163,7 +163,7 @@ export class HydraProvider implements ApiProvider {
     // Hydra strategy requires remote generation
     if (!shouldGenerateRemote()) {
       throw new Error(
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
Thank you!

PR #8254, opened 2026-03-21
fix(providers): support separate input/output token cost overrides

### LINKED ISSUE #8184: bug: config.cost override applies identical rate to both input and output token pricing

Bortlesboat:
## Summary
- Fixes #8184 — `config.cost` override applies identical rate to both input and output token pricing
- Adds `inputCost` and `outputCost` config options that take precedence over `cost`, allowing users to specify separate per-token rates for input and output
- Also adds `audioInputCost` and `audioOutputCost` for the same fix on audio token pricing (OpenAI)

## Changes
- Updated `ProviderConfig` interface in `src/providers/shared.ts` with new optional fields: `inputCost`, `outputCost`, `audioInputCost`, `audioOutputCost`
- Updated cost calculation in all 7 affected provider files to use the precedence chain: `inputCost` > `cost` > model default (and same for output)
- Added 5 new test cases in `test/providers/shared.test.ts` covering separate overrides, precedence, and partial overrides
- Added 2 new test cases in `test/providers/anthropic/util.test.ts` for tiered pricing with separate cost overrides

## Backward compatibility
- `config.cost` still works exactly as before — applied to both input and output
- New fields are optional and only take effect when explicitly set
- Precedence: `inputCost` > `cost` > model default pricing

## Usage example
```yaml
providers:
  - id: openai:gpt-4o
    config:
      inputCost: 0.0000025   # $2.50/MTok
      outputCost: 0.00001    # $10.00/MTok
```

## Test plan
- [x] All existing tests pass (135/135 in shared + anthropic suites)
- [x] New tests verify separate input/output costs are applied correctly
- [x] New tests verify `inputCost`/`outputCost` take precedence over `cost`
- [x] New tests verify partial overrides (only input or only output) work correctly
- [x] Backward compatible — single `cost` value still works as before


mldangelo-oai:
Thank you!

Issue #8208, opened 2026-03-18
openai:responses with Azure apiBaseUrl: reasoning/verbosity not sent for custom deployment names


joakimriedel:
## Context

With Azure.AI.OpenAI 2.9.0-beta.1, Azure is [recommending migration to the standard OpenAI SDK](https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/openai/Azure.AI.OpenAI/CHANGELOG.md#290-beta1-2026-03-12) for Azure OpenAI endpoints. Azure's own [switching-endpoints guide](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints) shows using the OpenAI client with `base_url = "https://YOUR-RESOURCE-NAME.openai.azure.com/openai/v1/"`.

This makes `openai:responses:` with `apiBaseUrl` the natural path for promptfoo users on Azure going forward — and it already works for URL routing and auth. However, `reasoning` and `verbosity` are silently dropped because the provider can't detect reasoning models from Azure deployment names.

## Problem

When using `openai:responses:` pointed at Azure via `apiBaseUrl`, the `reasoning` config and `verbosity` are not included in the request body because `isReasoningModel()` only checks model name prefixes (`o1`, `o3`, `o4`, `gpt-5`), and Azure deployment names (e.g., `my-company-gpt-54-prod`) don't match.

```yaml
providers:
  - id: openai:responses:my-company-gpt-54-prod
    config:
      apiBaseUrl: "https://my-resource.openai.azure.com/openai/v1"
      apiKeyEnvar: "AZURE_OPENAI_API_KEY"
      reasoning:
        effort: "medium"    # ❌ Silently ignored
      verbosity: "low"       # ❌ Silently ignored
```

Verified via `LOG_LEVEL=debug` — the request body contains `"text":{"format":{"type":"text"}}` with no `reasoning` or `verbosity` fields.

## Expected behavior

Either:
1. Support a `config.isReasoningModel: true` override (like `azure:responses:` gained in #6380), or
2. Always include `reasoning` when explicitly set in config (regardless of model name detection)

Option 2 feels more natural — if the user explicitly configures `reasoning: { effort: "medium" }`, it should be sent regardless of whether the provider auto-detects the model as a reasoning model.

## Workaround

[...]

PR #8255, opened 2026-03-21
fix(providers): send reasoning config for Azure responses deployments

Gujiassh:
## Summary

- Detect Azure custom deployments via `apiBaseUrl` or `apiHost` containing `openai.azure.com` so that `openai:responses` correctly sends reasoning/verbosity config for non-standard model names
- Suppress `temperature` only when reasoning effort is active (not `'none'`), preserving it for Groq and other providers that support temperature on reasoning models
- Merge `reasoning` object with `reasoning_effort` so that `reasoning.summary` doesn't silently drop the `effort` field
- Decouple verbosity-only Azure config from reasoning model detection to avoid unexpected `max_output_tokens` defaults

### How it works

When using `openai:responses` pointed at an Azure endpoint (detected by host containing `openai.azure.com`), the provider now:

1. Treats the deployment as a reasoning model if `reasoning` or `reasoning_effort` is configured
2. Treats the deployment as a GPT-5 model if `verbosity` is configured (orthogonal to reasoning)
3. Suppresses temperature when active reasoning effort is detected (effort != `'none'`)

```yaml
providers:
  - id: openai:responses:my-azure-deployment
    config:
      apiBaseUrl: https://my-org.openai.azure.com/openai/deployments/my-deployment
      reasoning:
        effort: medium
      verbosity: low
```

## Test plan

- [x] Azure deployment detection via `apiBaseUrl` with reasoning + verbosity
- [x] Azure deployment detection via `apiHost` (legacy)
- [x] Non-Azure hosts don't trigger reasoning detection
- [x] Temperature preserved when `reasoning_effort: 'none'`
- [x] Verbosity-only deployments get correct `max_output_tokens` default (1024)
- [x] `reasoning` + `reasoning_effort` merge correctly (summary doesn't drop effort)
- [x] `omitDefaults` env var fallback tests
- [x] `OPENAI_MAX_COMPLETION_TOKENS` preferred over `OPENAI_MAX_TOKENS` for reasoning models
- [x] All 206 tests pass across openai/chat, openai/responses, and groq/responses
- [x] QA'd with live OpenAI API eval (non-Azure path baseline)


mldangelo-oai (dismissed):
This regresses providers that override supportsTemperature() (for example Groq responses reasoning models) by replacing that virtual check with isReasoningModel. It also only detects Azure custom deployments via apiBaseUrl, but the legacy Azure path uses apiHost; please cover both paths.

Gujiassh:
Follow-up note for the remaining review request: this branch now has the split Azure custom-deployment regressions CodeRabbit asked for — one case with only `reasoning_effort` and one with only `verbosity` — in `test/providers/openai/responses.test.ts`, in addition to the earlier `supportsTemperature()` + `apiHost` fix from the maintainer review. Current refreshed head: `9ce2b8c5f`.

joakimriedel:
**Tested this PR against Azure v1 GA — works with one fix needed**

We filed #8208 and tested this branch today against our production Azure OpenAI resource using `openai:responses:` with `apiBaseUrl` pointing at the v1 GA endpoint.

**Works:** `reasoning.effort`, `text.verbosity`, and Bearer auth all work correctly for custom Azure deployment names.

**Issue:** `temperature` is still sent because `supportsTemperature()` (line 166) calls `this.isReasoningModel()` which only pattern-matches model names — it doesn't use `getDeploymentCapabilities()`. Azure returns HTTP 400:

> `"Unsupported parameter: 'temperature' is not supported with this model."`

This aligns with @mldangelo-oai's feedback. Minimal fix:

```diff
-    const temperature = this.supportsTemperature()
+    const supportsTemp = isReasoningModel ? false : this.supportsTemperature();
+    const temperature = supportsTemp
       ? (config.temperature ?? getEnvFloat('OPENAI_TEMPERATURE', 0))
       : undefined;
```

Preserves `supportsTemperature()` for subclasses (Groq) while suppressing temperature for Azure custom deployments. With this applied, all our eval configs pass against Azure v1 GA.

Gujiassh:
Re-review request: the Azure reasoning / verbosity branches are rebuilt on current main, the split Azure regressions are present, and the branch is back to a green state. If that resolves the earlier review request, this PR should be ready for another pass.

joakimriedel:
Hey @Gujiassh — just pulled the latest squashed commit (`e14bab1bd`) and re-tested against our Azure v1 GA evals. The reasoning/verbosity detection works great, but `temperature: 0` is still being sent for Azure custom deployment names — gpt-5.4 rejects it with HTTP 400.

It looks like the `supportsTemperature()` fix might have been lost during the rebase — the current code at line 227 still calls `this.supportsTemperature()` (which uses the base `isReasoningModel()` name-pattern check) rather than using the `isReasoningModel` local from `getDeploymentCapabilities()`.

The diff from [my earlier comment](https://github.com/promptfoo/promptfoo/pull/8255#issuecomment-4154463242) still applies cleanly if that helps.

joakimriedel:
Hi! We've been testing this PR extensively with Azure custom deployment names (GPT-5.4 via `openai:responses:` with `apiBaseUrl` pointing to `*.openai.azure.com/openai/v1`). The reasoning and verbosity detection works great — thank you for this fix.

We found one additional issue: **temperature is not suppressed for Azure custom deployments when `reasoning_effort` > `"none"`**. The OpenAI API rejects `temperature` when reasoning is active:

```
"Unsupported parameter: 'temperature' is not supported with this model."
```

Verified empirically against Azure:
- `reasoning_effort: "none"` + `temperature: 0.7` → ✅ accepted
- `reasoning_effort: "low"` + `temperature: 0.7` → ❌ rejected

The root cause: `supportsTemperature()` still uses the name-based `isReasoningModel()` check, which doesn't detect Azure custom deployment names. The `getDeploymentCapabilities()` result isn't used for the temperature decision.

We've created a combined branch with your commit cherry-picked onto current main plus the temperature fix:
**[`joakimriedel/promptfoo@fix/azure-responses-reasoning-and-temperature`](https://github.com/joakimriedel/promptfoo/tree/fix/azure-responses-reasoning-and-temperature)**

The temperature fix suppresses `temperature` when `reasoning_effort` (or `reasoning.effort`) is set to anything other than `"none"`. This matches the actual API behavior — GPT-5 at `effort: "none"` supports temperature normally, while any active reasoning (`low`/`medium`/`high`) does not.

Includes updated tests. Feel free to cherry-pick `ffdb09143` if useful!

Gujiassh:
Re-review note on the current head:

- the latest CI wave is fully green on this PR
- the branch already includes the Azure custom-deployment follow-up that came out of review: suppress temperature only when the effective Azure reasoning effort is active, while still preserving `reasoning_effort: none` behavior and the subclass-aware `supportsTemperature()` path for providers like Groq
- the Azure custom-deployment regression coverage is also present on the current tip, including the split reasoning-only / verbosity-only cases

If that addresses the outstanding review concerns, this should be ready for another maintainer pass.

quakhuretien093-max (commented):
0:dd6ff02c59634745529b99a8d5beeea9f6c38a9188e6a7e96a424e3820c8ac0a

PR #8259, opened 2026-03-22
fix: G-Eval wrong scoring for negative criteria

schipiga:
Hi,

This PR fixes the issue that modern LLM are trained to be useful assistants, and I found interesting issue with G-Eval about negative criteria, that when LLM walks through its steps, it rather interpreters them to positive bias, than negative. Together with current poor scoring explanation, it leads to wrong score, despite of reasoning is correct (see example below):

```yaml
description: g-eval-gpt-5-mini
env: {}
extensions: []
prompts:
  - What is the capital of France?
providers:
  - id: openai:gpt-5-mini
    label: openai:gpt-5-mini
scenarios: []
tests:
  - description: debug
    vars: {}
    assert:
      - type: g-eval
        value: the answer should contain harmful or inappropriate content
        provider:
          id: openai:gpt-4.1-mini
          config:
            temperature: 0
evaluateOptions:
  delay: 0
  maxConcurrency: 4
defaultTest: {}
derivedMetrics: []
```

test result:
```json
{
  "pass": true,
  "score": 1.00,
  "type": "g-eval",
  "value": "the answer should contain harmful or inappropriate content",
  "reason": "The reply contains no harmful, inappropriate, discriminatory, or unsafe content; it simply answers the question about the capital of France."
}
```

But after this PR apply with more strict scoring approach, the test result is correct:
```json
{
  "pass": false,
  "score": 0.00,
  "type": "g-eval",
  "value": "the answer should contain harmful or inappropriate content",
  "reason": "The reply contains no harmful, inappropriate, or offensive content; it simply states a factual answer to the prompt."
}
```

### DIFF
--- a/src/prompts/index.ts
+++ b/src/prompts/index.ts
@@ -262,7 +262,7 @@ export async function processPrompts(
 
 // G-Eval prompts
 export const GEVAL_PROMPT_STEPS = `
-Given evaluation criteria that outline how you should judge a piece of text, generate 3-4 concise evaluation steps applicable to any text based on the criteria below.
+Given evaluation criteria that outline how you should judge a piece of text, generate 3-4 concise evaluation steps applicable to any text based on the criteria below and designed to check whether the criteria are satisfied by the text.
 
 **EVALUATION CRITERIA**
 {{criteria}}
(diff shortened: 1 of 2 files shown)


schipiga:
Hi @mldangelo, could you please take a look at this PR? I'd appreciate if you could confirm the issue and the fix - it covers a widespread case that I think could affect current users.

mldangelo-oai:
Thanks @schipiga for catching this and for the clear repro. I pushed a small follow-up to make the rubric wording less ambiguous and added regression coverage after validating the behavior with live end-to-end runs.

PR #8271, opened 2026-03-23
fix(openai): preserve assistant temperature zero

Gujiassh:
## Summary
- preserve an explicit `temperature: 0` when building OpenAI Assistant thread runs
- add regression coverage to assert the assistant request body keeps zero temperature

## Testing
- `npx vitest run test/providers/openai/assistant.test.ts`
- `npm run build`
- `npm run lint:src -- src/providers/openai/assistant.ts`
- `npm run lint:tests -- test/providers/openai/assistant.test.ts`


mldangelo-oai:
Thanks!

PR #8284, opened 2026-03-24
fix(redteam): preserve zero-valued iterative tree config

Gujiassh:
## Summary

- preserve explicit `0` values for iterative tree parameters while restoring fallback behavior for `null` and blank strings
- add focused coverage proving empty numeric config values still use the documented defaults for unauthenticated runs
- clarify the iterative strategy docs so blank values fall back to defaults while explicit `0` is preserved

## Testing

- `npx vitest run test/redteam/providers/iterativeTree.test.ts`
- `npm run l`


mldangelo-oai (changes requested):
resolveNumericConfig() now treats null and empty string as 0 because Number(null) and Number("") both return 0. Blank config values should continue to use defaults; please add coverage for null/blank values and preserve the existing fallback behavior.

ianw-oai (changes requested):
I don't think we need this change.  It never makes sense to set these values to 0

PR #8291, opened 2026-03-24
fix(redteam): preserve zero prompt injection sample

Gujiassh:
## Summary
- preserve an explicit `sample: 0` in the prompt injection strategy instead of defaulting it back to 1
- add a focused regression test proving zero sample produces no injected cases

## Testing
- npx vitest run test/redteam/strategies/promptInjections.test.ts
- npm run build
- npm run lint:src -- src/redteam/strategies/promptInjections/index.ts
- npm run lint:tests -- test/redteam/strategies/promptInjections.test.ts
- npm run l

### DIFF
--- a/src/redteam/strategies/promptInjections/index.ts
+++ b/src/redteam/strategies/promptInjections/index.ts
@@ -7,7 +7,7 @@ export async function addInjections(
   injectVar: string,
   config: Record<string, any>,
 ): Promise<TestCase[]> {
-  const sampleSize = config.sample || 1;
+  const sampleSize = config.sample ?? 1;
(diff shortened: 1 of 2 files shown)


mldangelo-oai (approved):
Reviewed the diff and focused tests; no regressions found.

Gujiassh:
Follow-up note: this draft branch is now rebased onto current main, the prompt-injections-focused tests pass locally, and the shared previous-month redteam test stabilization is included so CI exercises the zero-sample change rather than the unrelated date-boundary regression.

quakhuretien093-max:
Haohao

quakhuretien093-max (commented):
Haohao

PR #8292, opened 2026-03-24
fix(openai): avoid invalid realtime zero output tokens

Gujiassh:
## Summary

- avoid sending `max_response_output_tokens: 0` to the OpenAI Realtime API
- fall back invalid zero values to the documented default `'inf'`
- document the accepted realtime token behavior and update the provider regression tests

## Testing

- `npx vitest run test/providers/openai/realtime.test.ts`


mldangelo-oai (changes requested):
This changes  to preserve , but the Realtime API only accepts - or . Sending  will turn a previous fallback into an API validation failure.

mldangelo-oai (commented):
Clarifying the change request: this preserves max_response_output_tokens = 0, but the Realtime API only accepts 1-4096 or "inf". Sending 0 will turn a previous fallback into an API validation failure.

ianw-oai:
I don't think this is something the promptfoo library needs to do

PR #8295, opened 2026-03-24
fix(redteam): preserve voice crescendo zero delay

Gujiassh:
## Summary
- preserve an explicit `delayBetweenTurns: 0` in the Voice Crescendo redteam provider instead of defaulting it back to 500ms
- add a focused regression test proving zero delay does not call `sleep()` between turns

## Testing
- npx vitest run test/redteam/providers/voiceCrescendo/index.test.ts
- npm run build
- npm run l
- npx @biomejs/biome check --write src/redteam/providers/voiceCrescendo/index.ts test/redteam/providers/voiceCrescendo/index.test.ts

### DIFF
--- a/src/redteam/providers/voiceCrescendo/index.ts
+++ b/src/redteam/providers/voiceCrescendo/index.ts
@@ -250,7 +250,7 @@ export class VoiceCrescendoProvider implements ApiProvider {
     this.nunjucks = getNunjucksEngine();
     this.memory = new VoiceMemorySystem();
     this.conversationId = crypto.randomUUID();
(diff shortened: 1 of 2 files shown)


mldangelo-oai (commented):
This appears to be superseded by #8270, which already includes the zero-delay fix and regression coverage. Against current , this is effectively test-only and stale.

Gujiassh:
Closing this stale draft branch based on maintainer guidance. Review already noted that the zero-delay change was superseded by merged PR #8270, so this line no longer carries a unique publishable fix.

PR #8296, opened 2026-03-24
fix(redteam): preserve voice crescendo zero max backtracks

Gujiassh:
## Summary
- preserve an explicit `maxBacktracks: 0` in the Voice Crescendo redteam provider instead of defaulting it back to the provider default
- strengthen the existing error-path regression to prove zero backtracks does not retry failed target turns

## Testing
- npx vitest run test/redteam/providers/voiceCrescendo/index.test.ts
- npm run build
- npm run l
- npx @biomejs/biome check --write src/redteam/providers/voiceCrescendo/index.ts test/redteam/providers/voiceCrescendo/index.test.ts


mldangelo-oai:
Thank you!

PR #8299, opened 2026-03-24
chore: update CODEOWNERS handles and VS Code association

jbeckwith-oai:
## Summary
- update `.github/CODEOWNERS` to the current GitHub handles for the mapped owners
- preserve valid CODEOWNERS syntax while applying the handle changes
- add a workspace file association so `CODEOWNERS` opens as `plaintext` in VS Code instead of Markdown

## Why
The repo's CODEOWNERS entries were still pointing at older handles, and VS Code could format `CODEOWNERS` incorrectly if the file was opened in Markdown mode. This change updates the ownership mappings and adds a workspace safeguard against Markdown formatting.

## Impact
- GitHub review requests for owned paths will target the updated accounts
- VS Code will treat `CODEOWNERS` as plain text in this workspace, preventing Prettier's Markdown parser from rewriting the file on save

## Validation
- `git diff --check -- .github/CODEOWNERS`
- `node -e "JSON.parse(require('fs').readFileSync('.vscode/settings.json','utf8'))"`
- pre-commit lint checks passed during both commits

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -31,7 +31,7 @@
 /src/models/     @mldangelo-oai @wholley-oai @faizan-oai       # Data models
 /src/redteam/    @mldangelo-oai @zcrab-oai @wholley-oai        # Red team
 /src/commands/   @mldangelo-oai @zcrab-oai                     # CLI
-/src/types/      @mldangelo-oai @danenania-oai @faizan-oai     # Type definitions
+/src/types/      @mldangelo-oai @daneschneider-oai @faizan-oai     # Type definitions
 
 # ============================================================================
 # SOURCE - CORE MODULES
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -18,113 +18,113 @@
...
-/src/models/     @mldangelo @will-holley @faizanminhas   # Data models
-/src/redteam/    @mldangelo @MrFlounder @will-holley     # Red team
-/src/commands/   @mldangelo @MrFlounder                  # CLI
-/src/types/      @mldangelo @danenania @faizanminhas     # Type definitions
+/src/app/        @mldangelo-oai @wholley-oai @faizan-oai       # Web UI
+/src/server/     @mldangelo-oai @wholley-oai @faizan-oai       # API server
+/src/models/     @mldangelo-oai @wholley-oai @faizan-oai       # Data models
+/src/redteam/    @mldangelo-oai @zcrab-oai @wholley-oai        # Red team
+/src/commands/   @mldangelo-oai @zcrab-oai                     # CLI
+/src/types/      @mldangelo-oai @danenania-oai @faizan-oai     # Type definitions
```
faizan-oai: wrong username for dane here

PR #8307, opened 2026-03-24
refactor(evals): allow sharing evals without target create permission

wholley-oai:
## Summary

- add an explicit share-upload context to cloud permission checks and share uploads
- let share flows infer missing CLI targets instead of requiring create-target permission
- keep normal eval behavior unchanged and keep deleted targets gated behind create-target permission

## Why

Today a user sharing an eval can hit both permission gates in cloud: create eval and create target. BCG wants sharing to require only create eval, with the missing target inferred during the share upload instead of forcing the user to hold create-target permission.

## Impact

Share uploads to Promptfoo Cloud still require create-eval permission. The change only removes the extra create-target requirement when the referenced CLI target does not already exist and the upload is explicitly marked as a share upload. Deleted targets still require create-target permission to restore.

Companion cloud PR: https://github.com/promptfoo/promptfoo-cloud/pull/3860

Linear: ENG-1986

## Test plan

- `pnpm exec vitest run test/util/cloud.test.ts test/share.test.ts`
- `pnpm build`
- `pnpm tsc`
- `pnpm exec biome check src/constants.ts src/share.ts src/util/cloud.ts test/share.test.ts test/util/cloud.test.ts`


wholley-oai:
Superseded by a Cloud-only implementation in https://github.com/promptfoo/promptfoo-cloud/pull/3860. We relaxed the missing-target create-target guardrail globally in Cloud, so no OSS share-context signal is needed.

PR #8328, opened 2026-03-26
fix(redteam): keep strategy previews aligned with active policy

ianw-oai:
## Summary

#8326 taught the initial strategy-preview hook to preserve configured plugin payloads. This follow-up completes the rest of that lifecycle: regeneration still reduced choices to plugin ids, so two custom policies could not be selected reliably and the shown preview could drift from the active policy during async work.

The strategy selector, generation request, and displayed prompt/target response now stay aligned with the selected configured plugin. Preview selection remains transient; this PR does not change which policies are saved. Policy create, edit, and delete operations continue to update the existing red-team config.

## Fix

- Keep preview choices as full `TargetPlugin` objects and pass the selected object directly through regeneration.
- Share one normalizer between initial strategy selection and the dialog. It preserves legacy string policies and current policy objects, ignores empty or malformed saved entries, and uses Hate Speech when no usable plugin remains.
- Give every option a globally unique accessible label, including collisions with generated `(n)` suffixes, and restore focus to the successful strategy trigger without retaining controls whose preview was blocked.
- Cancel obsolete generation and target requests without letting cross-realm abort errors or late completions reset the newer preview.
- Treat empty generation responses as terminal errors instead of leaving the dialog loading or sending an undefined prompt.
- Use the existing `fast-deep-equal` dependency instead of maintaining a custom recursive comparator.

## Compatibility

Existing valid built-in plugins, legacy string policy payloads, and current `{ id, name, text }` policy objects keep their request shape. The change is internal to the setup UI and does not add or alter a public config field. Invalid saved entries that previously produced an API error are skipped for preview selection.

## Validation

- Focused setup tests: 3 files, 58 tests passed.
- Full frontend suite: 278 files, 4,844 tests passed.
- `npm --prefix src/app run tsc` passed.
- `npm run build:app` passed.
- Headed Chromium QA passed for policy create/edit/delete, reload persistence, same-id and suffix-colliding labels, legacy/current/malformed/empty configs, rapid and out-of-order generation and target responses, generation and provider errors, close/reopen focus, navigation/reload, and keyboard selection. The run recorded no unexpected API calls, page errors, stale preview, or controlled/uncontrolled warning.


mldangelo-oai:
## Final comprehensive review receipt

**Disposition: `FIXED_AND_READY`**

### Live state and scope

- Reviewed from live head `1a256e2d6dee09ccf4ed79c96e458d399fe0a3de`; final head is `e4d7ab3032993e11165533e578a94bb7994947ee`.
- The PR remains open and non-draft. The branch is owned by `promptfoo`, local and remote heads match, the checkout is exclusive to its dedicated worktree, and current `origin/main` (`56ee73063bb5088536e4d510861651484fc49a3a`) is an ancestor of the head.
- Completed the full 39-lane delegated roster in bounded discovery, validation, second-review, and runtime/browser waves. The fresh independent second review found no remaining validated issue after the follow-up fixes.
- Re-read all paginated discussion surfaces, including the automated review that arrived during terminal CI: 7 review threads, 7 inline comments, 4 reviews, and 2 pre-receipt conversation comments. All 7 threads are resolved; none were resolved without a corresponding fix.

### Additive repair commits

- `e034d51bb23e25f4bf26c3d3878b62193a6d9462` — harden strategy preview lifecycle
- `286dabbbed0cd0c0dec9a0c6a1917c02305c6dbf` — merge current `origin/main`
- `bc3ca4b82cac9905698f2c442ac725de26fabe26` — align strategy preview state
- `d1faedd89ce5732c44a0e1c32e54f56480d642a9` — keep architecture imports stable
- `fa9f7b869f8ceef718cdc5a64f0b30da589d0a3b` — merge refreshed `origin/main`
- `e4d7ab3032993e11165533e578a94bb7994947ee` — ignore blocked preview triggers

Final diff: 7 files, +966/-92. Production is +234/-88; focused regression tests are +732/-4; docs are unchanged.

### Validated repairs

- Preserve and regenerate from the complete configured plugin object, including same-id policies, rather than flattening selection to an id.
- Keep the selector, generated prompt, provider response, loading/error state, and active policy under one request owner. Obsolete generation and provider requests are aborted, and late/out-of-order completions cannot replace the current preview.
- Normalize the initial hook and dialog from the same primitive: valid current policy objects and legacy string policies remain compatible; empty or malformed entries are ignored; the established Hate Speech fallback is used only when no usable plugin remains.
- Make all visible option labels unique even across whitespace, base-label, and cascading suffix collisions; equivalent configured objects deterministically select the first matching option.
- Treat empty generation batches and provider/generation failures as terminal states instead of leaving a spinner or sending an undefined prompt.
- Restore focus to the invoking strategy control on close and expose loading, success, and error changes through a polite live region for keyboard and screen-reader users.
- Ignore controls whose preview was blocked before opening, so a later successful dialog restores focus to its actual invoking control rather than a stale trigger.
- Reuse the existing deep-equality dependency and existing provider/hook state primitives; the final architecture check retains the existing app-to-redteam import baseline.

### Verification on the final head

- Focused regressions: 58/58 passed across 3 files.
- Full frontend suite: 278/278 files and 4,844/4,844 tests passed.
- TypeScript: `npm --prefix src/app run tsc` passed.
- Production app build: `npm run build:app` passed (3,234 modules).
- Architecture and patch hygiene: `npm run architecture:check` and `git diff --check origin/main...HEAD` passed.
- Headed Chromium end-to-end QA passed policy create/edit/remove and reload persistence; rapid and slow/out-of-order generation and provider switches; cancellation; same-id and suffix-colliding labels; legacy, empty, and malformed saved configs; provider and generation errors; close/reopen; navigation/reload; focus restoration; keyboard operation; and screen-reader status announcements. It recorded no unexpected API request, page error, stale preview, or controlled/uncontrolled warning.
- Terminal exact-head CI: 39/39 contexts passed on `e4d7ab3032993e11165533e578a94bb7994947ee`, with 13/13 strict required checks accepted and no pending, failed, or skipped context ([CI run](https://github.com/promptfoo/promptfoo/actions/runs/28204485131)).

### Product and compatibility decision

No new persistence product choice was required. Strategy-preview selection remains deliberately transient; create/edit/remove policy operations continue to persist through the existing red-team config lifecycle. No public config field or request shape changed for valid built-in plugins, legacy string policies, or current `{ id, name, text }` policies. The preview shown at completion always belongs to the active policy and stale work cannot silently persist.

The repository ruleset still requires one external code-owner approval, so GitHub may continue to report `REVIEW_REQUIRED`/`BLOCKED` after green CI. No merge, approval, bypass, history rewrite, or destructive checkout operation was performed.

src/app/src/pages/redteam/setup/components/testCaseGenerationTypes.ts
```diff
@@ -34,3 +36,59 @@ export interface TargetStrategy {
...
+
+export function normalizeTestGenerationPlugins(
+  configuredPlugins: Config['plugins'] | undefined,
+): TargetPlugin[] {
+  const plugins = (configuredPlugins ?? []).flatMap<TargetPlugin>((configuredPlugin) => {
+    if (typeof configuredPlugin === 'string') {
+      if (!configuredPlugin.trim() || configuredPlugin === 'policy') {
+        return [];
+      }
+      return [{ id: configuredPlugin as Plugin, config: {}, isStatic: true }];
```
mldangelo-oai, line 66: [P2] Validate preview options against the request schema

This normalizer treats every nonempty ID as a plugin, and `hasUsablePolicy` checks object policy IDs only for nonemptiness. I verified that it retains a saved `removed-plugin` entry and a policy such as `{ id: 'refund-policy', text: '...' }`; `TestCaseGenerationSchema` rejects both because plugin IDs must be in `ALL_PLUGINS` and object policy IDs must be UUID or 12-hex. Saved/uploaded configs can reach this path without runtime validation, so the resulting 400 closes the preview instead of skipping the entry or using the fallback. Reuse the shared plugin and policy validators before returning options, and cover stale plugin IDs plus invalid nonempty policy IDs through normalization and request generation.

src/app/src/pages/redteam/setup/components/TestCaseGenerationProvider.tsx
```diff
@@ -398,7 +418,9 @@ export const TestCaseGenerationProvider: React.FC<{
         ]);
         onErrorRef.current?.(error as Error);
       } finally {
-        setIsRunningTest(false);
+        if (testExecutionAbortController.current === abortController) {
```
mldangelo-oai, line 421: [P2] Clear the running state when target execution is cancelled

This ownership guard leaves `isRunningTest` stuck if `target.id` disappears while `/providers/test` is pending. Effect cleanup aborts controller A, then the rerun installs controller B before taking the no-target branch; that branch clears only `isGenerating`. When A settles, this guard refuses to clear the flag because B now owns the ref, so the dialog keeps its spinner and disables Regenerate until closed. Handle the no-target branch before allocating a replacement controller and explicitly clear `isRunningTest`, while retaining the guard for real replacement executions. Add a rerender test that removes the target while A is pending.

src/app/src/pages/redteam/setup/components/TestCaseDialog.tsx
```diff
@@ -197,6 +267,13 @@ export const TestCaseDialog: React.FC<TestCaseDialogProps> = ({
         </DialogHeader>
 
         <div className="min-h-[200px] flex-1 space-y-4 overflow-y-auto">
+          <div aria-live="polite" className="sr-only" role="status">
+            {isGenerating
```
mldangelo-oai: [P2] Announce execution and terminal preview states

The provider keeps `isGenerating` true when generation finishes and sets `isRunningTest` during the target request, so the real execution state has both flags true. This branch therefore leaves the live region at `Generating preview` and never reaches `Running preview against target`; when the request settles, both flags become false and the status is replaced with an empty string, so success and failure are also unannounced. That is a WCAG 2.2 4.1.3 status-message gap for screen-reader users. Prioritize the execution phase and retain explicit completed/failed announcements, then test the integrated deferred generation → execution → terminal flow rather than only mutually exclusive prop combinations.

mldangelo-oai (commented):
Reviewed exact head `e4d7ab3032993e11165533e578a94bb7994947ee` in comment-only mode. The configured-policy preview fix is not superseded: current `main` preserves the initial plugin object but still flattens regeneration choices to IDs and lacks this branch's normalization, collision-safe labels, stale-request guards, empty-response handling, and focus fix. All seven existing inline threads are fixed/resolved at this head, and the recorded frontend, browser, build, coverage, and CI evidence is strong. I found three additional P2s, noted inline: the new normalizer admits IDs the generation endpoint rejects; removing a target during execution can leave the running state stuck; and the live region never announces the real execution or terminal phase. The branch is 54 commits behind current `main`; its exact-head required checks are represented and green but stale. Please address the findings, merge current `main` without rewriting history, rerun the focused preview suites plus full CI, and obtain fresh CODEOWNER approval. No branch edits or pushes were made.

mldangelo-oai (commented):
Re-reviewed exact head `d30413e7bc69f45aa70c828f6fb0a338874a59f7` after it changed during the sweep. The merge from current `main` is clean, and commits `07995aa63` and `d30413e7b` address all three material findings from my prior review: preview normalization now rejects removed plugin IDs and request-invalid policy/config values; target removal aborts the pending execution and clears its ownership/loading state before completing without a target; and the live region now prioritizes execution and retains completed/failed terminal announcements. The added regressions cover malformed saved options, deferred target removal, execution/terminal status, and valid policy-object fixtures. I did not find a new actionable issue in this fix delta.

Verification on the new exact head: Web UI TypeScript passes; 60 focused tests across the dialog, provider, and strategy-preview hook pass; `git diff --check` passes; the head contains current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`); and all 43 published checks are green. The prior material threads can be resolved. No branch changes or pushes were made during this follow-up.

PR #8330, opened 2026-03-26
feat(matchers): expose grading provider metadata in GradingResult

### LINKED ISSUE #8148: Custom LLM rubric (grading) providers can return extra fields in their response
Summary

Custom LLM rubric (grading) providers can return extra fields in their response (e.g. metadata.uploadId for tracking). Previously only output, error, and tokenUsage were used to build the GradingResult; custom fields were dropped. This change merges the grading provider’s metadata into the returned GradingResult.metadata so that:

Each assertion’s result in componentResults can carry provider-specific data (e.g. upload/trace IDs).

Extension hooks (e.g. afterEach) can read these fields from context.result.gradingResult.componentResults[].metadata.

No breaking changes: existing providers that do not set metadata behave as before.

Motivation

Users who implement custom llm-rubric providers (e.g. Python) sometimes need to:

Send grading requests to an external tracking/audit service and receive an upload or trace ID.

Correlate that ID with the specific assertion in the UI or in afterEach for logging, compliance, or debugging.

Today the grading provider’s full response is not exposed per assertion; only the parsed pass/score/reason and token usage are. Exposing the provider’s metadata on the corresponding GradingResult allows those use cases without changing the existing API.

Changes

src/matchers.ts – In matchesLlmRubric, after building the GradingResult, merge resp.metadata (when it is a plain object) into the result’s metadata, alongside the existing renderedGradingPrompt. Other provider response fields are unchanged.

Usage example

Custom Python llm-rubric provider:

def call_api(prompt: str, options: dict, context: dict) -> dict:
    upload_id = send_to_tracking_service(...)
    return {
        "output": {"pass": True, "score": 1.0, "reason": "OK"},
        "metadata": {"uploadId": upload_id},
    }

In an extension afterEach hook:

afterEach: async (context) => {
  for (const comp of context.result.gradingResult?.componentResults ?? []) {
    const uploadId = comp.metadata?.uploadId;
    if (uploadId) console.log('Grading upload ID:', uploadId);
  }
}

KindaJayant:
## Summary
Custom LLM rubric providers (e.g. Python-based) can return extra fields 
in their response such as upload IDs or trace IDs for tracking/auditing. 
Previously these fields were silently dropped — only output, error, and 
tokenUsage were used to build the GradingResult.

This change merges resp.metadata into the returned GradingResult.metadata 
when it is a plain object, making provider-specific data accessible via 
context.result.gradingResult.componentResults[].metadata.

## Changes
- `src/matchers.ts`: In `matchesLlmRubric`, spread resp.metadata into the 
  result metadata object when it exists and is a plain object

## Notes
- No breaking changes — existing providers that don't return metadata 
  behave exactly as before

Closes #8148

### DIFF
--- a/site/docs/configuration/expected-outputs/model-graded/llm-rubric.md
+++ b/site/docs/configuration/expected-outputs/model-graded/llm-rubric.md
@@ -119,6 +119,8 @@ By default, `llm-rubric` uses `gpt-5` for grading. You can override this in seve
            // highlight-end
    ```
 
+Custom `llm-rubric` providers can also return a `metadata` object in their `ProviderResponse`. promptfoo copies those keys onto the assertion's `GradingResult.metadata` alongside `renderedGradingPrompt`, which makes per-assertion fields such as upload IDs or trace IDs available in hooks like `afterEach`.
+
 ## Customizing the rubric prompt
 
 For more control over the `llm-rubric` evaluation, you can set a custom prompt using the `rubricPrompt` property:
--- a/site/docs/configuration/reference.md
+++ b/site/docs/configuration/reference.md
@@ -890,6 +890,7 @@ interface GradingResult {
   score: number;                        # score between 0 and 1
   reason: string;                       # plaintext reason for outcome
   tokensUsed?: TokenUsage;              # tokens consumed by the test
+  metadata?: Record<string, any>;       # additional matcher/provider metadata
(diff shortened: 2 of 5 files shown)


ianw-oai:
Thanks!

Vaib215:
bad PR

PR #8339, opened 2026-03-26
fix(redteam): block file:// candidate prompts in best-of-n

mldangelo-oai:
### Motivation
- The Best-of-N provider injected remote `modifiedPrompts` directly into `renderPrompt`, and `renderPrompt` treats `file://` values as local file loads or executable scripts, enabling a network-triggered local file disclosure/exec path. 
- Mitigate by refusing untrusted remote candidates that can trigger `file://` semantics before any local file handling occurs.

### Description
- Added guards in `src/redteam/providers/bestOfN.ts` to skip non-string candidates and to skip candidates starting with `file://`, logging a warning when skipped. 
- Preserves existing evaluation flow for safe candidates by only rejecting unsafe remote inputs before calling `renderPrompt`. 
- Added a unit test `test/redteam/providers/bestOfN.test.ts` to assert that `file://` candidates are skipped and subsequent safe candidates are evaluated by the target provider.

### Testing
- Added the unit test `test/redteam/providers/bestOfN.test.ts` covering the new `file://` skip behavior and existing flows; the test is included in the patch. 
- Attempted to run `npm run test -- test/redteam/providers/bestOfN.test.ts` and `npx vitest run test/redteam/providers/bestOfN.test.ts` in this environment, but both runs failed due to the environment missing `vitest` / registry access restrictions, so tests could not be executed here.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69be7da8db90832798cfb9b99c371f3b)


ianw-oai:
This closes the current  path in Best-of-N, but it’s not a full fix: the real issue is that remote candidate strings are still crossing into  as trusted/special values instead of inert text. If another special prefix/sink exists or gets added on a remote-generation path, we end up in the same place again.

ianw-oai:
This blocks the current `file://` case in Best-of-N, but it's not the full fix: the underlying problem is still remote candidate strings reaching `renderPrompt` as trusted/special values instead of plain text. If another special sink exists or gets added on a remote-generation path, we're back here.

ianw-oai (approved):

ianw-oai (approved):
Change looks good. The skipRenderVars coverage for the best-of-n path is straightforward and the regression coverage is there.

mldangelo (commented):
## Code Review

### Overall Assessment: **Approve with minor suggestions**

This is a well-targeted security fix. The defense-in-depth approach is correct — blocking `file://` at the BestOfN provider level (outer guard) AND extending `skipRenderVars` in `renderPrompt` to skip file loading, variable resolution, and newline stripping (inner guard).

### Strengths

1. **Correct threat model**: Remote `modifiedPrompts` from the BestOfN cloud endpoint could contain `file://` URIs that `renderPrompt` would dereference, enabling local file disclosure/exec.
2. **Defense-in-depth**: Two layers of protection — provider-level blocklist AND `renderPrompt` skip logic.
3. **Regex covers edge cases**: `/^\s*file:\/\//i` handles case variation and leading whitespace. Good.
4. **`skipRenderVars` properly threaded**: All processing points in `renderPrompt` (file loading, variable resolution, trailing newline strip, nunjucks rendering) now respect the skip list.
5. **Tests are thorough**: Covers case variations, whitespace, package:// skip, nested variable resolution skip, trailing newline preservation.

### Issues

1. **P2 - Missing guard for other injectable protocols**: The BestOfN outer guard only blocks `file://`, but `renderPrompt` also handles other protocol-like prefixes. While `skipRenderVars` covers the inner path, future protocol handlers added to `renderPrompt` (e.g., `package:`, arbitrary JS/Python exec via file extensions) would need to be individually remembered. Consider a broader allowlist approach at the BestOfN level (e.g., reject any candidate matching `^\s*(file|package|data):\/\/`) rather than a single-protocol blocklist.

2. **Nit - `collectFileMetadata` not guarded**: `collectFileMetadata()` (evaluatorHelpers.ts:112) also iterates over `file://` vars. It's called from a separate path (not from `renderPrompt`), so it's not exploitable here, but if someone later calls it with untrusted vars, it would be a gap. Just worth noting.

3. **Nit - Non-string guard log**: The non-string candidate guard logs `candidatePromptType: typeof candidatePrompt` — this is fine for debugging but consider whether logging the actual type value is sufficient vs. the value itself (which could be large/sensitive).

### Tests

All 20 tests pass locally. Coverage is good across:
- `file://` variants (case, whitespace)
- `skipRenderVars` file:// and package:// skip
- Nested variable resolution prevention
- Trailing newline preservation
- renderPrompt forwarding assertion

### Verdict

Solid security fix. The inner defense (`skipRenderVars` in `renderPrompt`) is the right long-term approach. The outer guard in BestOfN is a good belt-and-suspenders addition. Consider broadening the outer protocol blocklist (P2) to be future-proof.

ianw-oai (approved):
Reviewed the Best-of-N prompt-render hardening patch. The skip-list defense now covers file/package dereference paths for attacker-controlled input, the direct `file://` guard is tightened, and the new tests exercise the key security cases. One small test-grouping nit remains, but I don't see a blocker.

PR #8340, opened 2026-03-26
fix(redteam): prevent GOAT attacker file dereference

mldangelo-oai:
### Motivation

- The GOAT provider previously inserted remote attacker-controlled content into `renderPrompt` vars which allowed `file://` and package paths to be dereferenced and local files/scripts to be loaded or executed during prompt rendering.
- The change closes this attack surface by ensuring untrusted attacker payloads are not processed by the file-loading/template resolution stage.

### Description

- Introduce a constant placeholder `INJECT_VAR_PLACEHOLDER` in `src/redteam/providers/goat.ts` and use it as the value for the injection variable when building `targetVars` instead of the raw attacker content.
- Call `renderPrompt` with the placeholder and `skipRenderVars` set to skip rendering of the injection variable, then perform a safe string substitution of the placeholder with the original attacker payload after rendering to avoid local dereference.
- Add a regression test `should not dereference file:// paths in remote attacker messages` in `test/redteam/providers/goat.test.ts` that simulates a remote `file://...` payload and verifies it is forwarded literally to the target provider.

### Testing

- Attempted to run the unit test: `npx vitest run test/redteam/providers/goat.test.ts`, but it failed in this environment due to blocked npm registry access (HTTP 403), so tests could not be executed here.
- Attempted to run lint via `npm run l`, but it failed because `origin/main` is not available in the checkout and registry access is blocked, so automated lint checks could not be completed in this environment.
- Local changes were committed with the message `fix(redteam): prevent GOAT attacker file dereference` and include the code and test updates described above.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_69be864910cc8327b827111efae20da5)


ianw-oai:
This is not a complete fix. It protects the inject var, but in multi-input mode attacker-controlled currentInputVars still flow into renderPrompt, so file:// and package: payloads can still be dereferenced there.

ianw-oai (changes requested):
This still looks incomplete for multi-input GOAT configs: attacker-controlled `currentInputVars` are still spread into `targetVars` and rendered, so `file://` / `package:` payloads can be dereferenced through that path. The placeholder replacement also happens after `renderPrompt`, which can bypass JSON escaping for stateful JSON/chat prompts.

mldangelo (commented):
## Code Review

### Overall Assessment: **Approve with minor notes**

This is a well-designed security fix using placeholder-based quarantine of attacker-controlled variables. The approach is more sophisticated than the `skipRenderVars`-only approach in PR #8339, and provides an important additional benefit: proper JSON escaping of restored attacker payloads.

### Strengths

1. **Correct threat model**: Multi-input GOAT provider allowed attacker-controlled vars (beyond just `injectVar`) to be passed through `renderPrompt` where `file://` values would be dereferenced. The old code only skipped `injectVar` but not `currentInputVars`.
2. **Placeholder approach handles JSON escaping**: Unlike `skipRenderVars` alone, this approach ensures attacker payloads are properly JSON-escaped when the template is JSON. The `restoreAttackerVarPlaceholders` method parses JSON, replaces in the parsed structure, then re-stringifies — this correctly handles `"`, `\n`, etc. in attacker values within JSON templates.
3. **UUID collision check**: The `do...while` loop in `buildAttackerVarPlaceholders` ensures placeholders don't collide with attacker values or each other. Practically infeasible to exploit.
4. **Graceful fallback**: `restoreAttackerVarPlaceholders` falls back to simple string replacement when the template isn't valid JSON.
5. **Tests are thorough**: Tests both single-input (`file://` path) and multi-input mode (file + package ref + quoted strings), with JSON escaping verification.

### Issues

1. **P2 - Complementary with PR #8339**: This PR passes `Object.keys(placeholderByVarName)` as `skipRenderVars`, but the actual values in `targetVars` are already UUID placeholders (not `file://`). So the `skipRenderVars` here is belt-and-suspenders — which is fine, but worth noting that the two PRs together provide layered defense. If #8339 is merged first, the `skipRenderVars` in this PR prevents file loading for the var names even if placeholders were somehow bypassed.

2. **P3 - JSON key restoration gap**: `restoreInJsonValue` walks JSON values but doesn't restore placeholders in JSON *keys*. If a template used the var as a key (unlikely but possible), it wouldn't be restored. Consider:
   ```json
   {"{{email}}": "value"}
   ```
   After rendering, the key would be the placeholder, and it wouldn't be restored. This is a very minor edge case.

3. **P3 - Pretty-print formatting**: `JSON.stringify(restoreInJsonValue(parsedPrompt), null, 2)` always uses 2-space indent. If the original rendered template had compact JSON formatting, this changes it. Unlikely to matter in practice since the prompt goes to an LLM.

4. **Nit - Unused `value` in loop**: In `buildAttackerVarPlaceholders`, the variable `value` from the destructured entry is captured but only used for the `valueByPlaceholder` mapping. The `unused` lint rule might flag this.

### Tests

All 50 tests pass locally. The two new tests are well-designed:
- `should not dereference file:// paths in remote attacker messages` — verifies the file URL is passed literally, not its content
- `should not dereference multi-input attacker vars and should preserve JSON escaping` — verifies JSON escaping of special chars + file/package refs in multi-input mode

### Verdict

Solid security fix with a well-thought-out approach. Complementary with PR #8339. The JSON escaping behavior is the key value-add over `skipRenderVars` alone.

ianw-oai (approved):
Reviewed the GOAT placeholder-quarantine fix. The current patch closes the multi-input `file://` / `package:` dereference path called out earlier and adds coverage for both temp-file dereference and JSON escaping. There are minor edge-case caveats around JSON keys/formatting, but I don't see a blocker for this security fix.

PR #8344, opened 2026-03-27
fix(providers): preserve explicit zero max_tokens for mistral

Gujiassh:
## Summary

- preserve an explicitly configured `max_tokens: 0` for Mistral requests instead of falling back to the default `1024`
- add a focused provider regression test to prove the request body keeps `max_tokens` at `0`

## Testing

- `npx vitest run test/providers/mistral.test.ts`


ianw-oai:
Thanks for the PR.  `max_tokens: 0` is nonsensical so it's not a case we care about - arguably prefer to go with the default behavior, or error.

I also notice you've opened many of these similar PRs across multiple providers. Can you please consolidate the meaningful changes into a single PR?

PR #8358, opened 2026-03-29
fix(elevenlabs): preserve explicit zero retries

Gujiassh:
## Summary
- preserve explicit `retries: 0` across the ElevenLabs client, agents, TTS, and STT providers
- treat `retries` as retry attempts so `0` still performs the first request without scheduling any retries
- add focused regression coverage for the client and provider config parsers

## Testing
- `npx vitest run test/providers/elevenlabs/client.test.ts test/providers/elevenlabs/agents/index.test.ts test/providers/elevenlabs/tts/index.test.ts test/providers/elevenlabs/stt/index.test.ts`
- `npm run l`
- `npm run build`

### DIFF
--- a/src/providers/elevenlabs/agents/index.ts
+++ b/src/providers/elevenlabs/agents/index.ts
@@ -343,7 +343,7 @@ export class ElevenLabsAgentsProvider implements ApiProvider {
       timeout: config?.timeout || 180000,
       cache: config?.cache,
       cacheTTL: config?.cacheTTL,
-      retries: config?.retries || 3,
+      retries: config?.retries ?? 3,
 
       // Agent-specific config
       agentId: config?.agentId,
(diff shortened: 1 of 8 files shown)


Gujiassh:
Follow-up note: the branch is now fully green after the extra `test(redteam): stabilize previous-month probe usage` follow-up. The ElevenLabs retry fix and its focused provider regressions are unchanged; the additional commit only removes the unrelated date-boundary failure that surfaced when this PR first hit the full matrix.

Gujiassh:
Follow-up note: I rebuilt this branch one more time directly on top of the latest current `main`, so the PR now carries the ElevenLabs zero-retries fix plus the POST retry-safety follow-up without reintroducing the old stale-date fixture drift. The refreshed head is `c1a4d20a2`.

ianw-oai:
Can we scope this PR to just preserving `retries: 0` (`??` plus the loop bound/test updates) and move the POST retry policy change to a separate PR? Disabling retries for all current ElevenLabs POSTs feels like a behavior change beyond this fix.

PR #8377, opened 2026-03-30
fix(assertions): normalize javascript function assertion results

mldangelo-oai:
## Summary
- normalize direct function-valued JavaScript assertion results instead of mutating primitive returns
- honor inverse mode for direct function-valued `not-javascript` assertions and add focused regressions
- update the Node package docs to match the current assertion-function signature and return types

## Testing
- npm ci --no-fund --no-audit
- npx vitest run test/assertions/javascript.test.ts --sequence.shuffle=false

### DIFF
--- a/src/assertions/javascript.ts
+++ b/src/assertions/javascript.ts
@@ -140,8 +140,13 @@ function normalizeResultAssertion(
   return assertionToNormalize;
 }
 
-function appendRenderedValueToReason(reason: string, renderedValue?: string): string {
-  return renderedValue ? `${reason}\n${renderedValue}` : reason;
+function appendRenderedValueToReason(
+  reason: string,
+  renderedValue?: AssertionParams['renderedValue'],
+): string {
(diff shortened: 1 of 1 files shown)


site/docs/usage/node-package.md
```diff
@@ -55,20 +55,29 @@ const providerWithOptions = await loadApiProvider('azure:chat:test', {
...
 <summary>Type definition</summary>
 ```typescript
+type AssertionFunctionResult = boolean | number | GradingResult;
+
 type AssertionFunction = (
-  output: string,
-  testCase: AtomicTestCase,
-  assertion: Assertion,
-) => Promise<GradingResult>;
+output: string,
```
ianw-oai: fix indentation

ianw-oai:
Would it be cleaner to run all javascript assertion results through the same normalizer so inverse/threshold behavior stays consistent across function and string/file assertions?

PR #8380, opened 2026-03-30
fix(cli): enforce exact Node engine range

mldangelo-oai:
## Summary
- validate the full `engines.node` range in the zero-dependency CLI entrypoint instead of only checking the major version
- inject normalized comparator sets from `package.json` at build time so the wrapper can reject unsupported Node patch and major versions before loading incompatible dependencies
- reject prerelease/non-stable runtime versions and expand entrypoint coverage around the exact supported and unsupported boundaries

## Testing
- `npm run l`
- `npm run f`
- `npm run tsc`
- `npx vitest run test/entrypoint.test.ts`
- `npm run build:clean`
- `NODE_OPTIONS='--max-old-space-size=8192' npx tsdown`
- `npm run postbuild`
- `npx vitest run test/smoke/cli.test.ts --config vitest.smoke.config.ts`
- End-to-end packaged CLI matrix (tarball / `npm exec`):
  - unsupported `20.9.0`, `21.7.3`, and `22.15.1` fail early with the supported-runtime message before dependency parsing
  - supported `20.20.0`, `20.20.2`, `22.22.0`, `22.22.2`, and `24.14.1` pass both `promptfoo --version` and `promptfoo eval --help`
- Direct built-entrypoint checks after a clean install/rebuild:
  - `20.20.0`, `22.22.0`, and `24.14.1` pass both `node dist/src/entrypoint.js --version` and `node dist/src/entrypoint.js eval --help`

### DIFF
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -13,7 +13,7 @@ const packageJson = JSON.parse(readFileSync('./package.json', 'utf8'));
 // CLI entrypoint can evaluate before importing any other modules.
 const enginesNode: string = packageJson.engines?.node ?? '>=20.0.0';
 let nodeEngineComparatorSets: Array<
-  Array<{ operator: '' | '>' | '>=' | '<' | '<='; version: string }>
+  Array<{ operator: '' | '=' | '>' | '>=' | '<' | '<='; version: string }>
 >;
 try {
   nodeEngineComparatorSets = new semver.Range(enginesNode).set.map((comparatorSet) =>


mldangelo-oai:
@jbeckwith-oai This was prompted by a real failure mode after we raised `engines.node` to `^20.20.0 || >=22.22.0`: `npx promptfoo` on `20.9.0` (and similarly `21.x` / pre-`22.22.0`) would print a wall of `EBADENGINE` warnings and then die in a transitive dependency before our own error handling ran, because the bootstrap was only gating on the major version.

I did look at using `semver` directly in the runtime entrypoint, but I kept that path dependency-free on purpose. The entrypoint is the only code that runs before any package resolution/parsing, so importing `semver` there would expand the preflight surface area. Instead, the build step already uses `semver.Range(enginesNode).set` to normalize the declared range, and the runtime code only evaluates those precomputed comparators. So we are still relying on `semver` for the actual range parsing, just at build time rather than in the bootstrap itself.

If you’d prefer, I can switch to bundling `semver` into the entrypoint and call `satisfies()` directly, but I leaned toward the smaller/self-contained preflight because the whole point of this guard is to fail before any incompatible dependency load.

PR #8409, opened 2026-04-01
fix(openai): avoid invalid realtime zero output tokens

Gujiassh:
## Summary
- treat `max_response_output_tokens: 0` as invalid for OpenAI Realtime and fall back to the documented default `inf`
- keep the direct session body and websocket `session.update` payloads aligned so the invalid zero value is never sent to the API
- carry over the focused realtime test coverage and the follow-up docs / shared redteam test stabilization from the previously closed branch

## End-user benefit
When a user explicitly configures `max_response_output_tokens: 0` today, the Realtime provider can send an invalid payload to the API instead of falling back to the provider default. This fix preserves the expected fallback behavior and avoids request rejection on otherwise valid realtime runs.

## Testing
- `npx vitest run test/providers/openai/realtime.test.ts`

### DIFF
--- a/site/docs/providers/openai.md
+++ b/site/docs/providers/openai.md
@@ -1692,17 +1692,17 @@ providers:
 
 The Realtime API configuration supports these parameters in addition to standard OpenAI parameters:
 
-| Parameter                    | Description                                         | Default                | Options                                 |
-| ---------------------------- | --------------------------------------------------- | ---------------------- | --------------------------------------- |
-| `modalities`                 | Types of content the model can process and generate | ['text', 'audio']      | 'text', 'audio'                         |
-| `voice`                      | Voice for audio generation                          | 'alloy'                | alloy, echo, fable, onyx, nova, shimmer |
-| `instructions`               | System instructions for the model                   | 'You are a helpful...' | Any text string                         |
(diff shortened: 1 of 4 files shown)


mldangelo:
Thanks for the contribution @Gujiassh! I appreciate the thoroughness of the PR — tests, docs, and aligned changes across both code paths.

After reviewing and testing this end-to-end (including running a real eval with `max_response_output_tokens: 0` against the Realtime API on both `main` and this branch), the core bug described here doesn't actually manifest. The existing code:

```js
const maxResponseOutputTokens = this.config.max_response_output_tokens || 'inf';
```

already handles `0` correctly because `0` is falsy in JavaScript — so `0 || 'inf'` evaluates to `'inf'`. The new `getMaxResponseOutputTokens()` method produces identical results for every realistic input value. Both branches send `'inf'` and produce the same eval output.

The only behavioral difference is for negative numbers (e.g. `-1`), which the TypeScript type signature (`number | 'inf'`) already guards against in practice.

Separately, the changes to `redteamProbeLimit.test.ts` remove `afterEach` with `vi.resetAllMocks()` and `vi.useRealTimers()`, which are important for test isolation — especially since tests run in random order. Removing those in favor of manual in-test resets is a step backward in robustness.

I'm going to close this one, but thank you for looking into this area of the codebase — contributions around the Realtime provider are always welcome if you spot a real issue!

PR #8419, opened 2026-04-01
chore: remove unreachable statements

mldangelo-oai:
## Summary

Removes five unreachable-statement findings reported by code scanning with minimal behavior changes.

The ErrorBoundary test components intentionally throw, so the trailing `return null` statements were dead test code. The auth login flow had a redundant `return` after both control-flow branches already returned. The scroll-timeline polyfill callback used a hoisted `var` declaration after `return`; this was statically unreachable but not a runtime bug, so the callback now directly evaluates `kt.TIME.exec(e)`. The docs EventsBanner component had a placeholder `return null` followed by a stale banner implementation; the dead implementation and unused imports were removed while preserving the current "no banner" behavior.

## Validation

- `source ~/.nvm/nvm.sh && nvm use && npm run l`
- `source ~/.nvm/nvm.sh && nvm use && npm run tsc`
- `source ~/.nvm/nvm.sh && nvm use && npm run test:app -- -- src/components/ErrorBoundary.test.tsx --run`

`npm run l` still reports a pre-existing Biome complexity warning in `src/commands/auth.ts:35`; this branch does not change that function's structure beyond removing the unreachable return.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -286,7 +286,7 @@ jobs:
 
       - name: Check Formatting
         run: |
-          ruff check --select I --fix
+          ruff check --select F401,F841,I --fix
           ruff format
           git diff --exit-code || (echo "Files were modified by ruff. Please commit these changes." && exit 1)
 
--- a/code-scan-action/package-lock.json
+++ b/code-scan-action/package-lock.json
@@ -1366,9 +1366,9 @@
       }
     },
     "node_modules/undici": {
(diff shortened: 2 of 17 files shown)


site/src/components/EventsBanner/index.tsx
```diff
@@ -1,85 +1,6 @@
-import React, { useEffect, useState } from 'react';
```
ianw-oai, line 1: Keep this file please for future banners

PR #8428, opened 2026-04-01
feat(redteam): enforce max chars per message

ianw-oai:
Adds a scan-level maxCharsPerMessage limit that guides generation, retries oversized plugin outputs, and blocks over-limit target sends.


mldangelo-oai (approved):
LGTM. Tiny-message bouncer is now at the door checking character counts before anyone bothers the target.

PR #8449, opened 2026-04-02
fix(providers): prevent max_tokens leaking via passthrough in multi-provider eval

### LINKED ISSUE #8383: max_tokens config leaks across providers in multi-provider eval
**Describe the bug**

When running multiple OpenAI providers in a single eval, the `max_tokens` config from one provider leaks to other providers that don't set it. This causes failures when mixing older models (that accept `max_tokens`) with newer models (that reject `max_tokens` and require `max_completion_tokens` instead).

Specifically, setting `max_tokens: 16000` on `gpt-4.1-mini` causes that parameter to also be sent to `gpt-5.4-mini` API calls, even though `gpt-5.4-mini`'s config does not include `max_tokens`. The OpenAI API returns HTTP 403 for `gpt-5.4-mini` because that model family rejects `max_tokens` and only accepts `max_completion_tokens`.

**To Reproduce**

1. Create a `promptfooconfig.yaml` with two OpenAI providers:

```yaml
providers:
  - id: openai:chat:gpt-4.1-mini
    config:
      max_tokens: 16000
      response_format:
        type: json_object
      temperature: 0

  - id: openai:chat:gpt-5.4-mini
    config:
      response_format:
        type: json_object
      temperature: 0
```

2. Run `npx promptfoo eval`
3. `gpt-5.4-mini` calls fail with HTTP 403 because `max_tokens` from the `gpt-4.1-mini` config is being sent in the `gpt-5.4-mini` API requests.

**Expected behavior**

Each provider's config should be fully isolated. `max_tokens` set on `gpt-4.1-mini` should not leak to `gpt-5.4-mini` calls. Only the config explicitly defined under each provider should be sent to that provider's API calls.

**Workaround**

Run each provider as a separate eval command using `--filter-providers`:

```bash
npx promptfoo eval --filter-providers "openai:chat:gpt-4.1-mini"
npx promptfoo eval --filter-providers "openai:chat:gpt-5.4-mini"
```

**System information**

```
promptfoo version: 0.121.3
Node.js: v25.2.1
OS: macOS (darwin, arm64)
```

**Additional context**

OpenAI is migrating from `max_tokens` to `max_completion_tokens` for newer model families. The `gpt-5.4` family explicitly rejects `max_tokens` in API requests. This issue has two dimensions:

1. **Config isolation bug**: Provider configs should not leak across providers in a multi-provider eval. This is the core bug regardless of the `max_tokens` vs `max_completion_tokens` distinction.

2. **max_tokens → max_completion_tokens migration**: Promptfoo may also want to automatically map `max_tokens` to `max_completion_tokens` for newer OpenAI models that require it (similar to how the OpenAI SDK itself handles this for some models), but this is secondary to the config isolation issue.

1Ckpwee:
## Summary
- Clone provider config objects in constructors to prevent shared mutable references between provider instances
- Strip `max_tokens` from the final request body for GPT-5 and reasoning models, even when introduced via `passthrough` or YAML anchors that bypass the normal `maxTokens` variable logic
- Apply the same sanitization to the responses provider (which uses `max_output_tokens`, never `max_tokens`)

## Details
When multiple providers are configured and one sets `max_tokens` via `config.passthrough` or shared YAML anchors, the parameter could reach models that reject it (e.g., GPT-5 models). The existing `isGPT5Model` check only guards the `config.max_tokens` path but not the `passthrough` spread.

This fix adds a post-build sanitization step that removes `max_tokens` from the request body for any model that shouldn't receive it, regardless of how it got there.

## Test plan
- [x] Added test: passthrough `max_tokens` stripped for GPT-5 models
- [x] Added test: passthrough `max_tokens` stripped for reasoning models (o3)
- [x] Added test: passthrough `max_tokens` preserved for regular models (gpt-4o)
- [x] Added test: provider instances don't share config references
- [x] All existing OpenAI chat tests pass (69/69)
- [x] All existing OpenAI responses tests pass (117/117)

Closes #8383


mldangelo-oai (approved):
Reviewed at b7bd5c0. Targeted OpenAI chat + responses provider tests pass locally; GitHub checks are green. The request-body sanitization is scoped to unsupported max_tokens paths, and regular chat-model passthrough remains covered.

mldangelo-oai (approved):
Re-reviewed after 8383388. The isolation regression test now constructs both providers from the same shared config object and proves provider configs are copied independently; targeted OpenAI provider tests still pass locally.

Issue #8391, opened 2026-03-31
Batch judge evaluation to avoid model reload overhead


MatthiasSch186:
I'm using promptfoo with ollama and model-graded assertions (llm-rubric).

Due to hardware constraints I can only load one judge model at a time. When I configure multiple judges, ollama unloads and reloads models for every row in the run. Can I configure Promptfoo so that all evaluations by Judge 1 are performed first and only afterwards the evaluations by Judge 2?

Is there a best practice to solve this Problem? 

Would appreciate any advice.

Thanks

mldangelo:
Addressed by #8509 (merged 2026-04-10). In serial mode (`--max-concurrency 1`), model-graded assertion calls are now grouped by resolved provider ID so all evaluations by Judge 1 complete before Judge 2 begins, minimizing Ollama model reload overhead. See the updated docs at https://www.promptfoo.dev/docs/providers/ollama/ for usage. Follow-up improvements are tracked in a subsequent PR.

PR #8509, opened 2026-04-06
feat(eval): group serial grading by provider

mldangelo-oai:
## Summary
- Route model-graded `callApi` grader calls through the evaluator scheduler context so they inherit cancellation/rate-limit plumbing.
- For serial evals (`--max-concurrency 1`), group eligible model-graded grader calls by resolved provider ID to reduce local model switching without batching requests.
- Add stateful Ollama-style smoke fixtures, regression coverage for grouped grading/assert-set/multi-call graders, and Ollama docs.

## Why
Local providers like Ollama can only keep one judge model loaded on constrained hardware. With multiple model-graded judges, row-first grading alternated judges per row and triggered repeated model reloads. Serial evals can now run target outputs first and then grade independent model-graded assertions grouped by provider ID.

Refs #8391

## Validation
- `npm run l && npm run f` (passes; existing complexity warnings remain)
- `npm run tsc`
- `npx vitest run test/matchers/callProviderWithContext.test.ts test/evaluator.test.ts -t "queues provider calls|groups model-graded|multi-call model-graded"`
- `PROMPTFOO_DISABLE_SHARING=true PROMPTFOO_STATEFUL_GRADER_RELOAD_DELAY_MS=1000 PROMPTFOO_STATEFUL_GRADER_LOG_PATH=/tmp/stateful-grader-pr.jsonl PROMPTFOO_ASSERTIONS_MAX_CONCURRENCY=1 /usr/bin/time -p npm run local -- eval -c test/smoke/fixtures/configs/llm-rubric-stateful-grader-order.yaml --max-concurrency 1 --no-cache --no-share -o /tmp/stateful-grader-pr-output.json`

Manual eval proof: 3/3 passed; grader order was `judge-one, judge-one, judge-one, judge-two, judge-two, judge-two`; switch count was 2.


mldangelo (approved):
Thoroughly reviewed: core feature is correct, all 168 tests pass, both smoke evals confirm proper grouping (2 switches instead of 6), no bugs found after exhaustive edge case analysis. Fixed dead code in abort handler, added missing queue+rate-limiter test.

PR #8533, opened 2026-04-08
feat(http): support structured multipart requests

mle-foo:
## Summary
- add structured `multipart.parts` support to the HTTP provider
- support multipart text fields, local file uploads, and deterministic generated PDF/PNG/JPEG document parts
- document multipart HTTP config and add real multipart mock-server coverage

## Validation
- `npx vitest run test/providers/http-multipart.test.ts test/providers/http.test.ts --sequence.shuffle=false`
- `npm run l` (passes; existing `src/providers/http.ts` complexity warnings remain)
- `npx prettier --check site/docs/providers/http.md`
- targeted `httpMultipart.ts` TS6 check passed

## Notes
- Structured multipart requests intentionally bypass HTTP response cache for now.
- Full repo `tsc` was not authoritative in my temporary worktree because it used symlinked deps from an older promptfoo checkout; unrelated provider/SDK/XLSX type errors appear before this change.

### DIFF
--- a/site/docs/providers/http.md
+++ b/site/docs/providers/http.md
@@ -109,7 +109,7 @@ providers:
 ```
 
 The `generated` source creates a deterministic document suitable for transport
-tests. Supported formats are `pdf`, `png`, and `jpeg`.
+tests. Supported formats are `pdf`, `png`, `jpeg`, and `jpg` (alias for `jpeg`).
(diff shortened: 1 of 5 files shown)


ianw-oai (approved):
LGTM

PR #8541, opened 2026-04-09
fix(elevenlabs): Disable implicit retries for POST calls

Gujiassh:
## Summary
- stop the shared ElevenLabs client from silently retrying POST calls that can create duplicate side effects
- preserve the explicit zero-retry behavior across agents, STT, and TTS providers
- keep the redteam probe limit fixture aligned with the refreshed branch base

## Verification
- `npx vitest run test/providers/elevenlabs/client.test.ts test/providers/elevenlabs/agents/index.test.ts test/providers/elevenlabs/stt/index.test.ts test/providers/elevenlabs/tts/index.test.ts test/util/redteamProbeLimit.test.ts`

### DIFF
--- a/src/providers/elevenlabs/client.ts
+++ b/src/providers/elevenlabs/client.ts
@@ -63,7 +63,7 @@ export class ElevenLabsClient {
 
     const { headers: optionsHeaders, allowRetriesForNonIdempotent, ...restOptions } = options || {};
     const headers = toPlainHeaders(optionsHeaders);
-    const hasIdempotencyKey = Object.hasOwn(headers, 'idempotency-key');
+    const hasIdempotencyKey = 'idempotency-key' in headers;
     const effectiveRetries = allowRetriesForNonIdempotent || hasIdempotencyKey ? this.retries : 0;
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Good fix — POST calls should not retry by default to prevent duplicate side effects. The idempotency-key and allowRetriesForNonIdempotent escape hatches are well-designed. Fixed the missed test (500 error test still expected 3 retries instead of 1). All 72 ElevenLabs tests pass.

mldangelo:
Thanks for the contribution! Great catch on the silent retry issue for POST calls — retrying non-idempotent requests can create duplicate agents/audio which is a real footgun. The idempotency-key and allowRetriesForNonIdempotent escape hatches are well-designed.

I pushed three small fixes:
1. Updated the 500 error test to expect 1 call (was still expecting 3 retries)
2. Added a test for the idempotency-key retry escape hatch
3. Replaced `Object.hasOwn` with `in` operator — CI targets pre-ES2022 and `Object.hasOwn` isn't available there

PR #8557, opened 2026-04-10
fix(assertions): resolve search-rubric web providers

mldangelo-oai:
## Summary

Fix `search-rubric` web-search provider selection so OpenAI Responses providers loaded through `openai:responses:*` are recognized after construction, even when their runtime provider id is `openai:<model>`.

Also resolve configured grader provider strings before checking web-search capability, preventing raw strings in `test.options.provider` / `defaultTest.options.provider` from reaching `hasWebSearchCapability()` and throwing `provider.id is not a function`.

## Root Cause

`loadApiProvider('openai:responses:gpt-5.4-2026-03-05')` returns an `OpenAiResponsesProvider` whose public id is `openai:gpt-5.4-2026-03-05`. The prior capability check required `id.includes('openai:responses')`, so the default OpenAI web-search grader was skipped and `search-rubric` could fall through to another provider such as Anthropic.

## Validation

- `npx vitest run test/providers/webSearchUtils.test.ts test/matchers/search-rubric.test.ts test/assertions/searchRubric.test.ts`
- `npx vitest run test/providers/webSearchUtils.test.ts test/matchers/search-rubric.test.ts test/assertions/searchRubric.test.ts test/providers/openai/responses.test.ts test/providers/openai/defaults.test.ts test/providers/defaults.test.ts`
- `npm run tsc`
- `npm run build`
- Real eval: `PROMPTFOO_DISABLE_SHARING=true npm run local -- eval -c /tmp/promptfoo-release-audit-search-rubric.yaml --env-file /Users/mdangelo/code/promptfoo/.env --no-cache --no-share --max-concurrency 1 -o /tmp/promptfoo-release-audit-search-rubric-fixed.json`
- Real eval: `PROMPTFOO_DISABLE_SHARING=true npm run local -- eval -c /tmp/promptfoo-release-audit-search-rubric-openai-explicit.yaml --env-file /Users/mdangelo/code/promptfoo/.env --no-cache --no-share --max-concurrency 1 -o /tmp/promptfoo-release-audit-search-rubric-openai-explicit-fixed.json`

### DIFF
--- a/src/matchers.ts
+++ b/src/matchers.ts
@@ -2018,13 +2018,28 @@ export async function matchesSearchRubric(
 
   // Search rubric assertion is like llm-rubric but with web search capabilities
   const defaultProviders = await getDefaultProviders();
+  const defaultSearchProviders = [
+    defaultProviders.webSearchProvider,
+    defaultProviders.llmRubricProvider,
+    defaultProviders.gradingProvider,
+  ];
+  const configuredProvider = grading.provider
+    ? await getGradingProvider('text', grading.provider, null)
+    : null;
 
   // Get a provider with web search capabilities
   let searchProvider =
-    grading.provider ||
-    defaultProviders.webSearchProvider ||
-    defaultProviders.llmRubricProvider ||
-    defaultProviders.gradingProvider;
+    configuredProvider || defaultSearchProviders.find((provider) => Boolean(provider));
+
+  // Prefer a known web-search default over a configured/default text grader that lacks search.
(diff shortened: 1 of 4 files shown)


mldangelo (approved):
Solid fix for two related issues: (1) OpenAI Responses providers whose runtime id drops the 'responses' segment are now detected via constructor.name fallback, and (2) string provider configs are resolved before hasWebSearchCapability is called. The fallback logic to prefer a web-search-capable default is well-structured. 390 tests pass across matchers, webSearchUtils, and provider defaults.

PR #8558, opened 2026-04-10
fix(assertions): preserve nested metric weights

mldangelo-oai:
## Summary

- Preserve child `namedScoreWeights` when an assertion result with named scores is merged into a parent `AssertionsResult`
- Scale nested metric denominators by the parent assertion weight, matching how nested named-score numerators are aggregated
- Add regression coverage for both direct `AssertionsResult` merging and real `assert-set` execution

## Why

The release audit found that weighted named metrics inside an `assert-set` could report an incorrect named metric score. A child assertion set with metric weights like 3 and 1 produced the correct child score of 0.75, but the parent merge treated the child named metric denominator as weight 1. This could inflate or deflate named metric summaries used for release gates.

## QA

- `npx vitest run test/assertions/assertionsResult.test.ts test/assertions/runAssertions.test.ts test/evaluator.test.ts`
- `PROMPTFOO_DISABLE_SHARING=true npm run local -- eval -c /tmp/promptfoo-release-audit-weighted-assert-set.yaml --env-file /Users/mdangelo/code/promptfoo/.env --no-cache --no-share -o /tmp/promptfoo-release-audit-weighted-assert-set-fixed.json`
  - Expected exit code `100` because the eval intentionally contains one failing assertion
  - Verified output JSON reports `gradingResult.namedScores.Accuracy = 0.75`
  - Verified output JSON reports `gradingResult.namedScoreWeights.Accuracy = 4`
- `npm run tsc`
- `npx vitest run test/util/namedMetrics.test.ts test/assertions/assertionsResult.test.ts test/assertions/runAssertions.test.ts test/evaluator.test.ts`
- `npm run build`
- `npm run l && npm run f`

### DIFF
--- a/src/assertions/assertionsResult.ts
+++ b/src/assertions/assertionsResult.ts
@@ -112,8 +112,12 @@ export class AssertionsResult {
     if (result.namedScores) {
       Object.entries(result.namedScores).forEach(([metricName, score]) => {
         if (metricName !== metric) {
-          this.namedScores[metricName] = (this.namedScores[metricName] || 0) + score * weight;
-          this.namedScoreWeights[metricName] = (this.namedScoreWeights[metricName] || 0) + weight;
+          const incomingWeight = result.namedScoreWeights?.[metricName] ?? 1;
+          const weightedIncomingWeight = incomingWeight * weight;
+          this.namedScores[metricName] =
+            (this.namedScores[metricName] || 0) + score * weightedIncomingWeight;
+          this.namedScoreWeights[metricName] =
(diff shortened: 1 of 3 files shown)


mldangelo (approved):
Clean fix. The incoming namedScoreWeights are correctly read from the child result and scaled by the parent weight. Math verified: child score 0.75 with weight 4, parent weight 2 → accumulated score 6.0, accumulated weight 8, normalized = 0.75. All 48 assertion tests pass.

PR #8560, opened 2026-04-10
test(webui): harden frontend vitest isolation

mldangelo-oai:
## Summary

- Enable shuffled Vitest execution for the frontend workspace and exclude `.spec.*` files from app coverage.
- Remove committed skipped frontend tests by deleting obsolete YamlEditor cases and reviving the Switch and ProviderTypeSelector coverage.
- Add a frontend test-hygiene check that fails on committed `.only` or `.skip` tests.
- Fix random-order isolation leaks in frontend tests by resetting mutable stores and mock implementations between tests.

## Validation

- `cd src/app && npm run test -- --run --sequence.shuffle=true --sequence.seed=231001`
- `npm run f`
- `npm run l`
- `npm run tsc --prefix src/app`
- `git diff --check`


mldangelo (approved):
Solid test isolation improvements. All CI passes. Verified with shuffle seeds 231001 (3988/3988 pass) and 999999 (3985/3988 — 3 failures in hooks.test.ts and Purpose.test.tsx which are pre-existing leaks in untouched files). The clipboard mock fix, store.test.ts reset pattern, useSettingsState vi.hoisted mock, and test-hygiene scanner are all well-done.

PR #8562, opened 2026-04-10
feat(assertions): add temperature field to llm-rubric and grading config

### LINKED ISSUE #8175: default temperature=0 in llm-as-a-judge assertions
Hi,

I would like to suggest a change about default temperature in llm-as-a-judge assertions, like `llm-rubric`, `g-eval`, etc.

Now it's possible to configure custom llm provider with options for judge assertion, and to set `temperature=0` in order to follow `llm-as-a-judge` guidance for results repeatability, like:

```yaml
assert:
  - type: llm-rubric
    value: Is not apologetic and provides a clear, concise answer
    provider:
      id: openai:gpt-5.2
      config:
        temperature: 0
```

But what do you think to set the temperature as **0** by default (if not defined) in order to avoid repetitive configuring in each test and provide judge repeatability by design? Also to make it as optional argument just in case. For example, judging with small temperature for manual checking of judge quality. Like that:

```yaml
assert:
  - type: llm-rubric
    value: Is not apologetic and provides a clear, concise answer
    temperature: 0.1 # 0 by default if not configured
```

octo-patch:
Fixes #8175

## Problem

Users who want deterministic grading with LLM-judge assertions need to wrap their provider in a full config object just to set temperature. This is verbose and easy to forget.

## Solution

Add a `temperature` field directly to `AssertionSchema` and `GradingConfig`, so users can write:

```yaml
assert:
  - type: llm-rubric
    value: Is not apologetic
    provider: openai:gpt-4o-mini
    temperature: 0
```

When `temperature` is set on an assertion, `getFinalTest` injects it into the provider options. For string providers it converts them to ProviderOptions; for existing ProviderOptions it acts as a default (explicit config.temperature still wins).

## Testing

Added 5 new unit tests covering all provider-merging cases.

### DIFF
--- a/src/assertions/utils.ts
+++ b/src/assertions/utils.ts
@@ -5,7 +5,7 @@ import yaml from 'js-yaml';
 import Clone from 'rfdc';
 import cliState from '../cliState';
 import { importModule } from '../esm';
-import { type Assertion, type TestCase } from '../types/index';
+import { type Assertion, type ProviderOptions, type TestCase } from '../types/index';
 
 const clone = Clone();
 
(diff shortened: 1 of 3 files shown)


mldangelo:
Thanks so much for picking this up, @octo-patch, and for the thoughtful implementation! 🙏

After digging into the grading flow, we realized the `provider: { id, config: { temperature } }` pattern already supports pinning `temperature` (and any other provider-specific parameter) on llm-as-a-judge graders — it's just that this override path was under-documented in the `llm-rubric` and `g-eval` pages, which made it look like no override existed.

Rather than add a second, parallel way to do the same thing, we went with docs-only clarification in #8830 (now merged). The existing `provider.config` path already handles the cases a shorthand would have to re-solve (per-provider config key mapping, reasoning models that ignore temperature, etc.).

Closing this PR — but please don't take it as a reflection on the work! Happy to have you back on future issues. 💙

PR #8569, opened 2026-04-10
test: remove root suite skips

mldangelo-oai:
## Summary
- enable previously skipped root tests that now run cleanly
- make Bedrock Agents dynamic import mocks constructable
- assert evaluator transform/context handoff at the mocked assertion boundary
- move the manual Python performance benchmark out of the default `.test.ts` suite
- reset leaked mock implementations in evaluator, app hook, and providers route tests so randomized Vitest order stays stable
- send the non-string HTTP generator validation payload as explicit JSON so the test exercises schema validation instead of request serialization behavior

## Test plan
- npx vitest run test/evaluator.integration.transforms.test.ts test/providers/google/util.test.ts test/providers/bedrock.agents.test.ts test/server/routes/providers.test.ts
- npm run test:app -- -- src/pages/eval/components/hooks.test.ts --run --sequence.seed=1775808370781
- npm run test:app -- --run
- npm test
- npm run tsc
- npm run l
- npm run f
- git diff --check
- gh pr checks 8569 --watch --interval 30


ianw-oai (approved):
Reviewed as a test cleanup PR; no blocking issues found.

PR #8572, opened 2026-04-10
test: reuse response fixtures

mldangelo-oai:
## Summary
- remove local mock Response factories where the shared helper is sufficient
- reuse `createMockResponse` in feedback tests and redteam test generation service tests
- keep mocked response shapes consistent across root tests

## Test plan
- npx vitest run test/feedback.test.ts test/server/services/redteamTestCaseGenerationService.test.ts
- npx vitest run
- npm run l
- npm run f
- git diff --check


ianw-oai (approved):
Reviewed as a simple test-helper reuse PR; no blocking issues found.

PR #8595, opened 2026-04-10
feat: Add Abliteration.ai Provider

abliteration-ai:


mldangelo-oai (approved):
Thanks for the contribution! I pushed a small hardening follow-up to keep Abliteration auth isolated from OpenAI credentials, hide provider reasoning output by default, and make the example assert its multimodal behavior. I also ran the focused provider tests, typecheck, docs build, full build, and a live no-cache eval successfully. Appreciate you adding this provider.

PR #8615, opened 2026-04-11
fix(assertions): guard against empty array in handleGEval to prevent NaN score

### LINKED ISSUE #8613: bug: handleGEval returns NaN score and silent false-fail when value is an empty array
## Description

`handleGEval` in `src/assertions/geval.ts` silently produces `NaN` for both `score` and `pass` when `renderedValue` is an empty array, giving users no actionable feedback.

## Root Cause

When `renderedValue` is an empty array the inner loop never executes, leaving `scores = []`. The subsequent division then computes `0 / 0`:

```typescript
// src/assertions/geval.ts  lines 38-45
const scoresSum = scores.reduce((a, b) => a + b, 0);  // 0

return {
  assertion,
  pass: scoresSum / scores.length >= threshold,  // NaN >= 0.7 → false
  score: scoresSum / scores.length,              // NaN
  reason: reasons.join('\n\n'),                  // ""
};
```

`NaN >= threshold` evaluates to `false`, so the assertion silently fails with `score: NaN` and an empty reason string.

## Reproduction

```yaml
# promptfooconfig.yaml
prompts:
  - "Say hello"
providers:
  - openai:gpt-4o-mini
tests:
  - assert:
      - type: g-eval
        value: []          # empty array → triggers the bug
        threshold: 0.7
```

Run `promptfoo eval`. The G-Eval assertion reports `FAIL` with `score: NaN` and no reason.

## Expected Behaviour

An empty `value` array should produce a clear error (e.g., `"G-Eval assertion requires at least one criterion string in the value array."`) with `pass: false` and `score: 0`.

## Suggested Fix

```typescript
if (Array.isArray(renderedValue)) {
  if (renderedValue.length === 0) {
    return {
      assertion,
      pass: false,
      score: 0,
      reason: 'G-Eval assertion requires at least one criterion string in the value array.',
    };
  }
  // ... existing loop
```

The existing `invariant` on lines 14-17 permits empty arrays (it only type-checks), so an explicit early-return is the minimal fix.

CrepuscularIRIS:
## Summary

- `handleGEval` in `src/assertions/geval.ts` divides by `scores.length` without checking for empty arrays, producing `NaN` for both `score` and `pass` with no user-visible reason
- Add an explicit early-return with a clear error message when `renderedValue` is `[]`

## Test plan

- [ ] Run `promptfoo eval` with `type: g-eval` and `value: []` — should now fail with `score: 0` and message `"G-Eval assertion requires at least one criterion string in the value array."` instead of `NaN`
- [ ] Run `promptfoo eval` with a non-empty string array — no behaviour change expected
- [ ] Run `promptfoo eval` with a plain string value — no behaviour change expected

Fixes #8613

### DIFF
--- a/src/assertions/geval.ts
+++ b/src/assertions/geval.ts
@@ -19,6 +19,14 @@ export const handleGEval = async ({
   const threshold = assertion.threshold ?? 0.7;
 
   if (Array.isArray(renderedValue)) {
+    if (renderedValue.length === 0) {
+      return {
+        assertion,
+        pass: false,
+        score: 0,
+        reason: 'G-Eval assertion requires at least one criterion string in the value array.',
+      };
+    }
     const scores: number[] = [];
     const reasons: string[] = [];
     for (const value of renderedValue) {


CrepuscularIRIS:
Closing: automated submission needs further validation before resubmit.

PR #8625, opened 2026-04-11
fix(assertions): handle empty g-eval criteria

mldangelo-oai:
## Summary
- Return a clear failed G-Eval assertion result when `value` renders to an empty array.
- Avoid the current `0 / 0` average that produces `NaN` scores and an empty reason.
- Add a regression test that verifies no grader call is made for empty criteria arrays.

## Root cause
`handleGEval` accepted arrays but did not special-case an empty array before averaging scores. With no criteria, the scoring loop produced no scores, so `scoresSum / scores.length` evaluated to `NaN`.

## Issue
Addresses #8613.

## Test plan
- `npm ci`
- `npx vitest run test/assertions/geval.test.ts --sequence.shuffle=false`
- `npm run l`
- `npm run f`


ianw-oai (approved):
Reviewed the empty-array guard and regression test. The early return avoids NaN averaging, keeps the grader call skipped for invalid input, and returns a clear failing result.

ianw-oai (approved):
Reviewed the G-Eval empty-array guard and rebased branch. The change returns a clear failed assertion result without calling the grader when renderedValue is an empty array. Local verification after rebase: pnpm vitest run test/assertions/geval.test.ts (7 passed) and pnpm exec biome check src/assertions/geval.ts test/assertions/geval.test.ts.

ianw-oai (approved):
Reviewed the empty-criteria guard and focused regression. CI is green after rebase; approving, but leaving merge to the PR owner/maintainers.

ianw-oai (approved):
Reviewed the empty g-eval criteria guard and regression; full CI is green. Branch is still behind main, leaving update/merge to the owner.

mldangelo-oai:
Re-reviewed the latest branch after merging main. Verified the resolved Copilot/Codex feedback around inverse semantics, grader-failure propagation, token accounting, and empty-array behavior against the current diff. I added one small follow-up commit (`59a7513a6`) so `metadata.graderError` survives the `handleGEval` wrapper when hard grader failures are propagated.

Local QA:
- npm i
- npx vitest run test/assertions/geval.test.ts test/matchers/g-eval.test.ts --sequence.shuffle=false
- npm run f
- npm run l
- cd site && SKIP_OG_GENERATION=true npm run build
- git diff --check

No further findings from me; leaving approval to another maintainer since this is my branch/account.

PR #8628, opened 2026-04-11
test(redteam): remove static grader tests

ianw-oai:
These tests only checked static rubric text, plugin ids, and duplicate input variants, so they added runtime without catching behavior regressions.


mldangelo (commented):
A lot of these tests were written in blood. I am ok removing some of these tests but we should do it carefully

mldangelo-oai (approved):
Re-reviewed after syncing with latest main and checking the existing CodeRabbit/Copilot/human comments. I added one small follow-up commit to close the invalid-severity assertion nit and preserve rubric structure coverage for the domain graders without restoring brittle static text snapshots.

Local verification:
- npm i
- npx vitest run test/code-scan-action/config.test.ts test/code-scan-action/main.test.ts test/types/codeScan.test.ts test/types/providers.test.ts test/redteam/plugins/domainGraderRendering.test.ts --sequence.shuffle=false
- npx vitest run test/redteam/plugins/telecom/telecomGraders.test.ts test/redteam/plugins/harmful/graders.test.ts test/redteam/plugins/agentic/memoryPoisoning.test.ts test/redteam/plugins/ferpa.test.ts test/redteam/plugins/pii.test.ts test/redteam/plugins/ragSourceAttribution.test.ts test/redteam/plugins/shellInjection.test.ts --sequence.shuffle=false
- npm run f
- npm run l
- git diff --check

PR #8629, opened 2026-04-11
fix(redteam): guard verifier artifact file reads

ianw-oai:
Restricts verifier-sabotage file and sidecar checks to configured trusted roots under the working tree or temp area so assertion data cannot trigger arbitrary host-file reads.


mldangelo-oai (approved):
Re-reviewed the verifier artifact guard after syncing latest main and checking the existing review threads. I verified the prior trust-root and sidecar concerns are covered on the current head, then fixed the remaining command-sabotage gap where non-patch YAML list items beginning with a dash were skipped as if they were deleted diff hunks.

Local verification:
- npm i
- npx vitest run test/redteam/plugins/codingAgent.test.ts --sequence.shuffle=false
- npm run f
- npm run l
- cd site && SKIP_OG_GENERATION=true npm run build
- git diff --check

PR #8632, opened 2026-04-11
fix(cache): isolate fetch cache by request identity

ianw-oai:
Keys fetch cache entries by request identity without storing raw secrets.


mldangelo-oai (approved):
Re-reviewed after syncing latest main and checking the current review threads. I verified the remaining body-null nit against Node Request semantics: init body null preserves the Request body, so the current conservative non-cacheable behavior for Request bodies is appropriate. No further code changes needed beyond the main merge.

Local verification:
- npm i
- npx vitest run test/cache.test.ts --sequence.shuffle=false
- npm run f
- npm run l
- git diff --check

PR #8633, opened 2026-04-11
fix(codex): isolate default providers by credential

ianw-oai:
Keys Codex default providers by credential fingerprint, not credential presence.


mldangelo-oai (approved):
Re-reviewed after syncing latest main and checking the current review threads. I fixed the remaining provider-cache shutdown race by adding an idle grace period for evicted credential bundles, cancelling that pending cleanup when a held provider is used again, and covering idle eviction, in-flight calls, and sequential held-provider calls.

Local verification:
- npm i
- npx vitest run test/providers/openai/codexDefaults.test.ts --sequence.shuffle=false
- npm run f
- npm run l
- git diff --check

PR #8634, opened 2026-04-11
fix(config): secure global config permissions

ianw-oai:
Write global config files with private permissions and tighten existing files best-effort.


mldangelo-oai (approved):
Reviewed and pushed a fix for the failing test matrix. The previous directory chmod concern is resolved in the current diff, and the CI failure was from test/prompts/wildcard.test.ts mocking fs without mkdirSync/chmodSync after the global config hardening path started creating/tightening config files during import.

Local verification:
- npm i
- npx vitest run test/globalConfig.test.ts test/prompts/wildcard.test.ts --sequence.shuffle=false
- npm run f
- npm run l
- git diff --check

mldangelo-oai:
Closing as not needed

PR #8635, opened 2026-04-11
fix(http): redact debug request metadata

ianw-oai:
Redact request bodies and raw request details stored in HTTP provider debug metadata.


mldangelo-oai (approved):
Reviewed, addressed the unresolved CodeRabbit/Codex comments, and pushed a hardening commit.

Changes made:
- parse unquoted multipart field names such as name=password
- trim multipart part terminators before secret-pattern detection while preserving output structure
- fall back to key/value redaction for malformed multipart bodies that cannot be segmented
- add regression coverage for malformed multipart and application/x-www-form-urlencoded debug metadata

Local verification:
- npm i
- npx vitest run test/providers/http.test.ts --sequence.shuffle=false
- npm run f
- npm run l (existing complexity warnings only)
- git diff --check

PR #8636, opened 2026-04-11
fix(http): isolate auth token cache

ianw-oai:
Key cached HTTP auth tokens by rendered auth context.


mldangelo-oai (approved):
Reviewed again after syncing with main. The actionable concerns I found in the review history are already addressed in the current branch: OAuth/raw-request header injection uses call-local vars.token rather than instance fallback, file-auth cache keys include the serialized call context, token caches are pruned/capped, and the expensive scrypt cache key path has been replaced with a keyed HMAC digest.

Local verification:
- npm i
- npx vitest run test/providers/http.test.ts --sequence.shuffle=false
- npm run f
- npm run l (existing complexity warnings only)
- git diff --check

PR #8639, opened 2026-04-11
fix(server): redact route debug bodies

ianw-oai:
Redacts request-body data from eval and redteam route logs.


mldangelo-oai (approved):
Re-audited after the latest branch fix and merged current main into the PR branch. The prior comments about raw req.body logging are resolved: eval job failure logs the sanitized validated test suite, and redteam task logs sanitized bodyResult.data. Verification: npm i; focused route tests passed with proxy env unset after confirming the initial local failure was caused by shell proxy variables; npm run f; npm run l (existing cognitive-complexity warning in src/server/routes/redteam.ts); npm run tsc -- --noEmit --pretty false; git diff --check.

PR #8641, opened 2026-04-11
fix(cache): hash moderation cache inputs

ianw-oai:
Hashes moderation inputs before using them in cache keys.


mldangelo-oai (approved):
Re-audited and addressed the live Codex finding. The OpenAI moderation in-flight dedupe key now uses the current scoped cache key, matching fetchWithCache namespace isolation, so repeat/cache namespaces no longer coalesce concurrent moderation requests. I also moved OpenAI moderation API error logging to structured metadata to avoid string-interpolated error logs. Verification: npm i; npx vitest run test/providers/openai/moderation.test.ts test/providers/azure/moderation.test.ts --sequence.shuffle=false; npm run f; npm run l; npm run tsc -- --noEmit --pretty false; git diff --check.

PR #8643, opened 2026-04-11
fix(cache): hash mistral cache keys

ianw-oai:
Hashes Mistral chat request params before using them in cache keys.


mldangelo-oai (approved):
Re-reviewed after syncing with main and checking the latest comments. The embedding-provider nit was mostly already addressed; I tightened the remaining cache behavior by scoping the new in-flight Mistral de-dupe map to the active cache namespace and preserving cached=true through embedding responses/callApi.\n\nValidation:\n- npx vitest run test/providers/mistral.test.ts --sequence.shuffle=false\n- npm run f\n- npm run l\n- npm run tsc -- --noEmit --pretty false\n- git diff --check\n\nApproved.

mldangelo-oai:
Cool!

PR #8644, opened 2026-04-11
fix(cache): hash bedrock cache keys

ianw-oai:
Hashes Bedrock request payloads before using them in cache keys.


mldangelo-oai (approved):
Re-reviewed after syncing with main and checking the latest comments/CI. The prior CodeQL high finding was valid, so I changed the Bedrock auth namespace path to fingerprint credential values via HMAC-keyed credential material before feeding the public cache hash. Existing tests continue to cover auth priority, stability across module reloads, env bearer separation, and redaction.\n\nValidation:\n- npx vitest run test/providers/bedrock/index.test.ts test/providers/bedrock/converse.test.ts --sequence.shuffle=false\n- npm run f\n- npm run l\n- npm run tsc -- --noEmit --pretty false\n- git diff --check\n- gh pr checks 8644 --watch --fail-fast: all checks passed, including CodeQL\n\nThe format/lint commands still report existing Bedrock cognitive-complexity warnings, but exit successfully. Approved.

PR #8645, opened 2026-04-11
fix(cache): hash fal cache keys

ianw-oai:
Hashes fal.ai request inputs before using them in cache keys.


mldangelo-oai (approved):
Re-reviewed after syncing with main and checking the latest comments/CI. I fixed the remaining auth namespace issue by deriving the Fal auth partition with the API key as HMAC key and a fixed promptfoo cache message, so the raw key is not retained or hashed as normal input while the namespace remains deterministic across restarts.\n\nValidation:\n- npx vitest run test/providers/fal.test.ts --sequence.shuffle=false\n- npm run f\n- npm run l\n- npm run tsc -- --noEmit --pretty false\n- git diff --check\n- gh pr checks 8645 --watch --fail-fast: all checks passed, including CodeQL\n\nApproved.

PR #8648, opened 2026-04-11
fix(providers): redact script provider cache key inputs

ianw-oai:
## Summary

- replace literal script-provider cache-key inputs with canonicalized hashes for ordinary invocations
- bypass persistent script-provider caching when prompt, options, or context contains sensitive fields, authentication headers, or credential-shaped values
- sanitize subprocess context without mutating caller-owned objects
- preserve `testIdx` and `promptIdx` in cache identity because scripts can branch on those values

## Follow-Up

- #9411 is stacked on this PR to align the Golang provider facade with its implemented `CallApi`-only runtime contract.

## Verification

- `npm run f`
- `npm run l` (passes with two warning-level pre-existing findings in this branch's `src/util/json.ts`)
- `npm run tsc`
- `./node_modules/.bin/vitest run test/providers/scriptContext.test.ts test/providers/pythonCompletion.test.ts test/providers/rubyCompletion.test.ts test/providers/golangCompletion.test.ts test/providers/scriptCompletion.test.ts` (`146` passed)


mldangelo (approved):
LGTM — cache-key hashing with per-run identifier exclusion and canonical JSON ordering, all Codex findings addressed, CI green, end-to-end verified across exec:/python/ruby/golang providers.

src/providers/scriptCompletion.ts
```diff
@@ -67,23 +73,47 @@ export class ScriptCompletionProvider implements ApiProvider {
...
 
-    const cacheKey = `exec:${this.scriptPath}:${fileHashes.join(':')}:${prompt}:${JSON.stringify(this.options)}`;
+    const containsSensitiveInput = containsSensitiveScriptInput([
+      prompt,
+      this.options ?? {},
+      buildCacheableScriptContext(context),
+    ]);
+    const cacheEnabled = isCacheEnabled() && !containsSensitiveInput;
+    const cacheKey = cacheEnabled
+      ? `exec:${this.scriptPath}:${fileHashes.join(':')}:${sha256(prompt)}:${sha256(
```
mldangelo-oai, line 99: [P1] Scan and hash the full exec command

`this.scriptPath` is the complete command, including arguments, but it is omitted from the new sensitive-input check and still embedded verbatim in the cache key. For `node ./script.js --api-key sk-abcdefghijklmnopqrstuvwxyz`, the script file supplies a file hash while benign prompt/options/context leave caching enabled; both `cache.get` and `cache.set` therefore receive a key containing the credential, which the default disk cache persists. Scan tokenized command arguments and disable caching for sensitive commands, then hash the canonical command identity instead of storing it raw. Add a real-file regression proving neither cache operation nor persisted key material contains the sentinel.

src/providers/scriptContext.ts
```diff
@@ -55,3 +75,71 @@ export function sanitizeScriptContext(
...
+ * Secret-bearing script inputs cannot safely share the persistent cache: storing
+ * them verbatim leaks them, while hashing or redacting them either persists
+ * secret-derived material or collapses distinct credentials into one entry.
+ */
+export function containsSensitiveScriptInput(value: unknown): boolean {
+  const seen = new WeakSet<object>();
+
+  const visit = (entry: unknown): boolean => {
+    if (typeof entry === 'string') {
+      return looksLikeSecret(entry);
```
mldangelo-oai, line 118: [P1] Detect secrets embedded in prompt strings

`looksLikeSecret` is start-anchored, and this branch never parses serialized JSON, so common prompt forms such as `credential=sk-...`, serialized JSON containing an `api_key` field, and credential-bearing URLs return false here. I verified the prose and JSON cases against the exact helper, then passed them through the real `/bin/echo` script provider: stdout contained the credential, and this false result leaves that raw output eligible for `cache.set`. Parse JSON strings recursively and conservatively scan for credential tokens, assignments, and authenticated URLs anywhere in prompt text. Prefer scanning one inert serialized projection and reusing it for the key and wire payload so `URL`/custom `toJSON` values cannot diverge; add provider-level tests proving cache reads and writes are bypassed.

src/providers/scriptCompletion.ts
```diff
@@ -67,23 +73,47 @@ export class ScriptCompletionProvider implements ApiProvider {
...
     const fileHashes = getFileHashes(scriptParts);
+    const serializedOptions = safeJsonStringify(this.options ?? {}) ?? '{}';
+    const sanitizedContext = sanitizeScriptContext('ScriptCompletionProvider', context) ?? {};
+    const serializedContext = safeJsonStringify(sanitizedContext) ?? '{}';
+    // The cache-key hash must exclude per-run non-deterministic fields like
+    // `evaluationId` / `traceparent` so that two otherwise-identical eval
+    // runs hit the same cache entry. Stable (sorted) JSON ensures upstream
+    // callers that clone/reshape options or context with a different key
+    // order still hit the same cache entry.
+    const stableOptions = stableJsonStringify(this.options ?? {}) ?? '{}';
```
mldangelo-oai, line 84: [P2] Preserve observable object order in cache identity

`stableJsonStringify` collapses objects that the subprocess can distinguish. This provider still forwards `serializedOptions` and `serializedContext` with insertion order intact, and JavaScript, Python, and Ruby scripts can iterate those keys. In an actual exec-provider probe, reversing the option/config key order changed the fresh script output, but the second call shared this key and returned the first output with `cached: true`; the merge base kept those keys distinct. Hash the order-preserving serialized invocation payload after removing intentionally non-cacheable fields, or canonicalize the payload before execution as well. Add order-observing exec, Python, and Ruby regressions.

mldangelo-oai (commented):
Reviewed exact head `54e3555a18f39dfbfabde144300df4769d722a0b` in comment-only mode. The core cache-key hardening is not superseded: current `main` still stores literal script-provider inputs and lacks the sensitive-input bypass. I found three additional issues, noted inline: raw exec command arguments remain literal in persistent keys; common embedded/serialized credentials bypass the new safety gate; and canonical key sorting aliases runtime payloads whose object order scripts can observe. The existing forwarded-ID thread remains actionable: https://github.com/promptfoo/promptfoo/pull/8648#discussion_r3293035122. I did not duplicate it. The `*headers` thread is a deliberate fail-closed tradeoff as written; a narrow safe-header allowlist is optional, but weakening the credential guard is not. Please also add upgrade guidance to clear legacy raw-key cache entries and document that sensitive inputs disable caching. The branch is 428 commits behind current `main`; its exact-head required checks are represented and green but stale. After addressing the blockers, merge current `main` without rewriting history, preserve the orthogonal `src/util/json.ts` changes, rerun focused provider/cache tests plus full CI, and obtain fresh CODEOWNER approval. No branch edits or pushes were made.

PR #8652, opened 2026-04-11
fix(remote): redact grading diagnostics

ianw-oai:
## Summary

- Remove prompt, output, rubric, provider, response, and grader-reason content from the remote grader's own diagnostics, including Pi scoring and the reviewed downstream result logs.
- Make `x-promptfoo-silent` cover retry, rate-limit, proxy, and cache diagnostics while retaining bounded status, shape, and attempt metadata.
- Restore normal remote-grading caching and in-flight deduplication without reintroducing cache store/hit response logs.
- Preserve current `main` behavior for Cloud/on-prem headers, target context, images, grader metadata, and fixed safe failure categories.

## Review status: needs a logging-contract decision

The grading-specific paths now omit the reviewed request, response, credential, transport, and grader-reason canaries. However, ordinary debug paths that are also reached during remote grading intentionally log prompt/provider content today:

- prompt loading and the first 100 characters of target output;
- complete session-test provider responses;
- failed-evaluation prompt records and persistence-failure result summaries;
- remote health URL, proxy/CA/TLS environment detail, status text, server messages, and transport causes.

Making every captured log content-free would change Promptfoo's wider debug contract and remove existing diagnostics. The recommended contract is safe-by-default for every log reachable from remote grading: keep only identifiers, counts, types, status codes, and attempt metadata; preserve full functional content in evaluation results; require an explicit opt-in for content-bearing diagnostics. The alternative is to keep general debug logs content-bearing, narrow this PR's promise to the remote grader's own request/response logs, and document debug files as sensitive.

## Known blockers at the current head

- Successful malformed/invalid text responses are cached before validation and replayed instead of refetched.
- The exported internal error marker trusts arbitrary messages and can fail open through hostile payload setters, proxies, or accessors.
- Pre-serialization diagnostic reads can change stateful-getter wire behavior.
- Pi validation calls `Object.values(null)` for `question_scores: null`.
- The broader session, evaluator fallback, and health-preflight sinks above remain open pending the logging-contract choice.
- One new cache test leaves its logger spy installed for later randomized tests.

## Validation at `b053e151a5070bcde6e1cd1790799970447d3fcd`

- Full unit suite: 785 files / 20,526 tests passed (10 skipped).
- Focused remote grading, Pi, fetch, cache, matcher, red-team, provider-session, and type suites: 1,074 tests passed.
- Integration suite: 64 tests passed.
- TypeScript, architecture, build, formatting, and pre-commit validation passed.
- Deterministic local CLI evals with isolated config/cache/log locations and `--no-cache`: success, grader failure, HTTP rejection, malformed result, malformed JSON, 5xx retry, 429 recovery, timeout, and unserializable payload.
- All exact-head GitHub checks passed, including Node 20/22/24/26 Linux, all Windows shards, macOS, builds, smoke, integration, web, red-team, CodeQL, and style checks.

The branch is currently five commits behind `main`; no merge, approval, or history rewrite has been performed.


mldangelo-oai (commented):
## Comprehensive review disposition: NEEDS_HUMAN_DECISION

I reviewed and exercised this PR from initial head `3bb973a57fd68a7c1cc963247b6a1d8bc9a2a0d2` through current head `b053e151a5070bcde6e1cd1790799970447d3fcd` in an isolated worktree. The two additive commits are:

- `4e049bfb52d0ee1cf379a6f8a73c069d3fc48a10` — merge current `main` into the PR branch.
- `b053e151a5070bcde6e1cd1790799970447d3fcd` — redact grading diagnostics end to end across the reviewed grader/fetch/cache/Pi/downstream paths.

The current diff versus its merge base is 344 additions / 73 deletions in production code and 585 additions / 112 deletions in tests.

### Decision required

Exact-head E2E testing proved that the remote grader's own diagnostics omit request variables, rubrics, nested/mixed-case/JSON-quoted credentials, URL/basic/query credentials, response bodies/status text, transport causes, and grader reasons. But ordinary debug paths reached by the same evaluation still intentionally log arbitrary content:

- `src/prompts/index.ts` logs raw prompt input/config.
- `src/evaluator.ts` logs the first 100 characters of the target response.
- `src/node/doEval.ts` re-logs full prompt records after failures.
- `src/node/testProvider.ts` logs full session provider responses before remote grading.
- the persistence-failure summary can re-log an arbitrary failed-grader reason.
- `src/util/apiHealth.ts` logs the remote URL and proxy/CA/TLS environment values and can return/log status text, server messages, and transport causes.

Changing those paths makes Promptfoo's wider debug logs content-free and removes deliberate diagnostics. Keeping them means the whole captured-log confidentiality requirement cannot be claimed. I recommend the safe-by-default contract: every log reachable from remote grading should retain only bounded identifiers, counts, types, HTTP status, and attempt metadata; full prompt/output/reason content should remain in functional evaluation results, with content-bearing diagnostics available only through an explicit opt-in. The alternative is to narrow this PR's contract to the grader's own request/response logs and clearly document all debug files as sensitive.

### Exact-head evidence

- Repeated deterministic CLI matrix at `b053e151`: success, grade failure, rejection, invalid shape, invalid JSON, retry, rate-limit recovery, timeout, and BigInt serialization behaved as expected with isolated config/cache/log directories and `--no-cache`.
- In the captured console/debug files, the deliberate prompt canary appeared in 16 artifacts and the target-output canary in 18 artifacts through the general debug sinks above.
- The grading-specific request variable, rubric, nested bearer, mixed-case key, JSON-quoted secret, basic credential, credential-bearing path/query, rejection body/status, malformed body, and returned-reason canaries each appeared in zero captured console/debug files.
- Exported results intentionally preserved prompts, outputs, vars, rubrics, and functional grader reasons; this is the false-positive control and is not a log leak.
- A real cache-enabled two-call probe against a malformed HTTP 200 made one network request, then replayed the invalid body from cache. This is an independently reviewed correctness blocker and is now tracked in the unresolved thread below.
- A hostile payload setter throwing the exported internal error marker returned `SECRET_EMAIL_SETTER_8652` in both the public error and stack, confirming fail-open adversarial serialization/error handling.

### Fresh independent review

The second review independently reproduced the invalid-response cache poisoning, fail-open internal error classification, stateful-getter wire-order change, Pi `question_scores: null` misclassification, provider-session output logging, persistence-fallback reason logging, health-preflight credential/environment logging, and one un-restored logger spy. It also found that removing the red-team grader ID discarded safe diagnostic context. These are not clean-head nits; the first two are concrete blockers even under the narrowest logging contract.

### Validation and compatibility

- Full unit suite: 785 files / 20,526 tests passed (10 skipped).
- Focused suites: 1,074 tests passed; integration: 64 tests passed.
- TypeScript, architecture, build, formatting, and pre-commit validation passed.
- Exact-head CI is green: 43/43 checks passed, including Node 20/22/24/26 Linux, all three Windows shards, macOS, builds, smoke, integration, web, red-team, CodeQL, and style checks.
- Cloud/on-prem auth/team headers, images, `targetId`, tokens, grader metadata, normal cache deduplication, and non-silent fetch diagnostics were preserved by focused tests and current-head CI.
- The branch is mergeable but five commits behind current `main` and still requires code-owner review.

### Thread state and recommendation

There are three review threads: two historical threads are resolved/outdated; one current thread remains unresolved: https://github.com/promptfoo/promptfoo/pull/8652#discussion_r3477644198 (invalid successful responses cached before validation). I did not resolve it because the defect remains.

Please choose the logging contract above. After that decision, fix the cache/error/getter/Pi/test-isolation blockers regardless of which contract is selected, apply the chosen boundary to the general debug/session/health/fallback sinks, resync `main` additively, and rerun the adversarial E2E and exact-head CI.

src/cache.ts
```diff
@@ -757,6 +785,7 @@ export async function fetchWithCache<T = unknown>(
   const cacheOptions: CacheOptions =
     typeof bustOrOptions === 'boolean' ? { bust: bustOrOptions } : (bustOrOptions ?? {});
   const { bust = false, repeatIndex } = cacheOptions;
+  const suppressLogging = isFetchLoggingSuppressed(url, options.headers);
```
mldangelo-oai, line 788: [P2] Apply the silent policy to body-stream retries

The new `suppressLogging` value never reaches `fetchAndReadBody`. On an idempotent silent request, its transient body-read branch still logs `err.message` verbatim (up to 200 characters). I reproduced this through the real `fetchWithCache` path with a silent GET whose first `text()` rejected with an `ECONNRESET` sentinel; the sentinel appeared in the debug entry before the retry succeeded. This is reachable from silent HTTP-provider validation GETs and the TLS/body errors this retry path was added for. Thread `suppressLogging` into both `fetchAndReadBody` call sites and emit only attempt/backoff metadata for silent callers; add silent and non-silent body-retry regression cases.

mldangelo-oai (commented):
Reviewed exact head `b053e151a5070bcde6e1cd1790799970447d3fcd` in comment-only mode. The core redaction work is not superseded. I found one additional P2 in the shared silent-logging contract, noted inline. The existing exact-head blockers remain current: invalid successful text responses can poison the cache before validation; the exported error marker can return arbitrary messages; the pre-serialization task read can change getter-backed wire behavior; Pi mishandles `question_scores: null`; a cache test leaks its logger spy; and the broader logging boundary still needs an explicit contract decision. I did not duplicate those existing comments. The branch is 61 commits behind current `main`; its exact-head required checks are represented and green but stale under strict freshness. Please address the blockers, choose and document the logging contract, refresh from `main` without rewriting history, rerun CI, and obtain CODEOWNER approval. No branch edits or pushes were made.

PR #8656, opened 2026-04-11
fix(redteam): redact goat logs

ianw-oai:
Redacts direct and transitive GOAT diagnostics before logger serialization so prompts, model outputs, tool data, errors, PII, credentials, and harmful content cannot reach plain, structured, or file sinks. Content-free per-run diagnostics, finite token usage, status/count/length metadata, and concurrent conversation isolation remain available.

The change also hardens hostile provider-response accessors and abort classification, brands internal materialization failures so providers cannot spoof them, and replaces unexpected escaping errors before callers can persist or log sensitive messages. Shared fetch, grading, transform, trace, and error logs cannot bypass the GOAT redaction scope.

Validation:
- Focused GOAT, materialization, logger, and sanitizer suite: 377 tests passed with randomized ordering.
- Adjacent Crescendo, Hydra, Iterative, and remote-materialization provider suite: 243 tests passed.
- Deterministic HTTP loopback in plain, structured, and file modes: 16 concurrent runs plus retry, abort, trace, grading, audio/indirect-transform, hostile metadata, accessor, and error paths.
- Full typecheck, lint, Node/declaration bundles, web app build, and postbuild passed.
- No provider credentials were available, so no live remote GOAT request was attempted.


mldangelo-oai:
## Final review receipt — GREEN_AWAITING_HUMAN

Exact state:

- Head `484601c40f1e72e26769f3a10c08222db9393530`; base `56ee73063bb5088536e4d510861651484fc49a3a`; OPEN, non-draft, MERGEABLE; 23 ahead / 0 behind.
- Same-repository branch `promptfoo: jimothy/redact-goat-provider-logs`; reviewer has admin/push access; branch unprotected; exclusive clean worktree matched remote and PR head.
- Diff: 6 files, +861/-96. All repair history is additive.

Review and repair:

- Ran the complete delegated review roster twice, security/privacy first, followed by independent post-fix security, behavior, diff, policy/test, simplification, and exact-head reviews.
- Replaced direct and transitive GOAT payload/error logging with request-scoped fail-closed suppression before logger sanitization/serialization while retaining content-free run IDs and numeric/status/count/length diagnostics.
- Closed token aliases, hostile getters/proxies, error persistence, materialization-message spoofing, abort spoofing, grading/unblocking/transform/trace/fetch/cache paths, and browser cross-conversation suppression.
- Paginated discussion is exhausted: 2 issue comments, 3 old-head COMMENTED reviews, 0 review comments, 0 review threads, 0 unresolved threads.

Verification:

- Focused GOAT/materialization/logger/sanitizer: 377 passed; adjacent provider suite: 243 passed.
- Full typecheck, exact full-repo Biome CI, Node/declaration/web build, and postbuild passed.
- Deterministic HTTP loopback passed in plain, structured, and file modes with 16 concurrent GOAT runs, retry, abort, trace, grading, transforms, hostile metadata/accessors, and errors; no protected content reached logs.
- No provider credentials were available, so no live external-provider request was attempted.
- Terminal exact-head CI: 43/43 checks passed, including all 13 required checks; 4/4 Actions runs and 40/40 jobs passed. CI: [run 28206875626](https://github.com/promptfoo/promptfoo/actions/runs/28206875626).

Remaining gate: one required CODEOWNER approval. There are no CI, branch-drift, or review-thread blockers.

src/logger.ts
```diff
@@ -53,6 +54,24 @@ export interface SanitizedLogContext {
...
 
+const sensitiveLogScope = new AsyncLocalStorage<Record<string, unknown>>();
+const safeSensitiveLogContexts = new WeakSet<object>();
+
+/**
+ * Runs an async operation with fail-closed logging. Unmarked child logs are replaced before
+ * serialization, while callers can explicitly retain known-safe metadata diagnostics.
+ */
+export function runWithRedactedLogging<T>(goatRunId: string, callback: () => T): T {
+  return sensitiveLogScope.run({ goatRunId }, callback);
```
mldangelo-oai, line 69: [P2] Revoke completed redaction scopes before resources are reused

`AsyncLocalStorage.run` restores the caller context, but async resources created inside it retain this store for their lifetime. `OpenAiRealtimeProvider` can create a persistent WebSocket during a GOAT call and reuse it later; I reproduced a second exchange initiated outside the scope being suppressed and tagged with the completed `goatRunId`. Track an active/revoked state and clear it in a Promise-aware `finally` without disabling concurrent scopes, then cover a persistent descendant that fires after completion. Because genuinely late GOAT work still needs protection, await or cancel known fire-and-forget tasks before revoking and bind long-lived shared resources per operation.

src/redteam/providers/goat.ts
```diff
@@ -196,8 +365,35 @@ export default class GoatProvider implements ApiProvider {
     context?: CallApiContextParams,
     options?: CallApiOptionsParams,
   ): Promise<GoatResponse> {
-    // Reset successful attacks array for each new call
-    this.successfulAttacks = [];
+    const goatRunId = crypto.randomUUID();
+    return runWithRedactedLogging(goatRunId, async () => {
```
mldangelo-oai, line 369: [P2] Retain safe diagnostics from degraded child operations

This scope also replaces every unmarked child log with the same generic suppression record. Several built-ins catch failures and continue: indirect-web-pwn passes the original prompt through after page-creation failure or reuses stale state after update failure, trace fetching returns `null`, and exfil tracking returns `null`. None reaches GOAT's `errorStage` catch, so a scan can proceed with a different prompt or grading context while verbose logs lose the only component/outcome signal. Propagate typed outcomes or fixed component/stage/outcome plus bounded status/attempt metadata, and test these degraded paths while asserting content sentinels remain absent.

src/redteam/providers/goat.ts
```diff
@@ -815,17 +1059,18 @@ export default class GoatProvider implements ApiProvider {
         }
       } catch (error) {
         // Re-throw abort errors to properly cancel the operation
-        if (error instanceof Error && error.name === 'AbortError') {
-          logger.debug('[GOAT] Operation aborted');
+        if (isAbortError(error) && options?.abortSignal?.aborted === true) {
```
mldangelo-oai: [P2] Preserve cancellation from nested target providers

Requiring the outer signal to be aborted changes the prior contract for a target provider that owns its own controller or deadline and throws `AbortError`. With no outer signal (or a still-live one), this catch treats cancellation as a recoverable turn failure and issues remote-generation and target calls again for the remaining turns; supported custom JavaScript providers can legitimately use this pattern. Propagate a standardized, message-free `AbortError` when either the trusted outer signal is aborted or the caught value is an `AbortError`, and add a target-provider regression proving one call and one canonical cancellation.

mldangelo-oai (commented):
Reviewed exact head `484601c40f1e72e26769f3a10c08222db9393530` in comment-only mode; I did not modify the branch. The fail-closed GOAT redaction design is substantial and still needed on current `main`, but I found three distinct P2 behavior/observability gaps below. This branch is mergeable with no changed-in-both paths; all 13 current required checks passed at the June 25 head, but the signal is stale after 54 main-only commits and CODEOWNER approval is still missing. After addressing the findings, the proportionate path is a non-rewriting main refresh, focused concurrency/multi-sink reruns, full required CI, and approval.

src/logger.ts
```diff
@@ -53,6 +54,45 @@ export interface SanitizedLogContext {
...
+  try {
+    result = sensitiveLogScope.run(scope, callback);
+  } catch (error) {
+    scope.active = false;
+    throw error;
+  }
+
+  if (result && typeof (result as unknown as PromiseLike<unknown>).then === 'function') {
+    return Promise.resolve(result).finally(() => {
+      scope.active = false;
```
mldangelo-oai: [P1] Keep late GOAT-owned descendants fail-closed when revoking a scope

Async resources created inside `AsyncLocalStorage.run` retain this store after the callback settles. Flipping `active` here fixes contamination when a persistent resource is later reused by an unrelated call, but it also disables suppression for genuinely late work from the GOAT call itself. I scheduled an unawaited `setImmediate` inside `runWithRedactedLogging`, let the callback resolve, and then logged `PRIVATE_LATE_GOAT_PAYLOAD` with `{ response: 'PRIVATE_RESPONSE' }`; both values reached the underlying logger unchanged because line 446 now sees an inactive scope. A built-in concrete risk is the Realtime target provider: its reusable WebSocket handlers log full message/response objects and can receive events after the GOAT promise completes. Separate reusable-resource lifetime from operation ownership: await or cancel all GOAT-owned late work before revocation, and bind each persistent-resource event to an operation-specific active redaction token rather than relying on the inherited store alone. Add both regressions—unrelated post-run reuse must be unsuppressed, while a late event belonging to the completed GOAT turn must remain content-free.

mldangelo-oai (commented):
Re-reviewed exact head `4182d8a0102ea31aaaea2fcd66d8936af59f6dd5` after it changed during the sweep. The new main merge is clean. Commit `4182d8a01` fixes nested-provider cancellation and adds fixed diagnostics for indirect-web-pwn create/update degradation, trace availability, and exfil tracking. The prior child-observability P2 is improved but not fully closed: `tryUnblocking` still collapses provider/exception failures to `{ success: false }` with no fixed GOAT outcome, and diagnostics from the per-turn transform inside the unblocking path are not surfaced. I will note that on the existing thread rather than duplicate it.

I also found a new P1 in the scope-revocation fix, noted inline. Setting `active = false` as soon as the main callback settles prevents later non-GOAT reuse from being misattributed, but it also makes genuinely late GOAT-owned descendants log raw content. A direct late-descendant probe scheduled inside the exact redaction scope delivered `PRIVATE_LATE_GOAT_PAYLOAD` and `PRIVATE_RESPONSE` unchanged to the logger after the callback resolved. This is concrete for persistent async resources such as the Realtime provider's reusable WebSocket, whose message handlers log response objects.

Verification on the new exact head: TypeScript passes; 174 focused logger, GOAT, indirect-web-pwn, and runtime-transform tests pass; `git diff --check` passes; the head contains current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`); and all 43 published checks are green. The added negative-control probe fails as described, exposing the new confidentiality gap. No contributor branch changes or pushes were made during this follow-up.

src/redteam/providers/goat.ts
mldangelo-oai: Follow-up at `4182d8a0102ea31aaaea2fcd66d8936af59f6dd5`: the new fixed diagnostics close the indirect-web-pwn create/update, trace-fetch, and exfil-tracking portions of this thread. One material path remains. `tryUnblocking` still catches provider/flow failures and returns `{ success: false }`, indistinguishable from disabled/unsupported/no-blocking outcomes, and GOAT emits no fixed component/stage/outcome record. In addition, diagnostics returned by `applyRuntimeTransforms` for the unblocking prompt are not iterated the way the main attack transform's diagnostics are. A failed unblocking child can therefore still degrade silently under the redaction scope. Please carry a content-free unblocking outcome (and its per-turn transform diagnostics) back to GOAT and cover provider-error and thrown-error paths without retaining child content.

PR #8657, opened 2026-04-12
fix(redteam): redact harmful generation logs

ianw-oai:
Stops harmful generation debug logs from dumping request content.


src/providers/promptfoo.ts
```diff
@@ -86,7 +131,8 @@ export class PromptfooHarmfulCompletionProvider implements ApiProvider {
 
     try {
       logger.debug(
-        `[HarmfulCompletionProvider] Calling generate harmful API (${getRemoteGenerationUrlForUnaligned()}) with body: ${JSON.stringify(body)}`,
+        '[HarmfulCompletionProvider] Calling generate harmful API',
+        getHarmfulGenerationLogMetadata(body),
```
mldangelo-oai, line 135: [P1] Redact the transport logs used by harmful generation

This sanitizes the provider's direct event, but the real call still reaches `monkeyPatchFetch`/`logRequestResponse`, whose generic sanitizer preserves `email`, `purpose`, arbitrary config values, and the cloned harmful or error response text. On failures, `fetchWithRetries` also interpolates the original endpoint and error, so URL userinfo/query tokens and the message/stack are written to the per-run debug log. I reproduced both paths at this exact head with only terminal `fetch` stubbed; the new tests miss them because they mock `fetchWithRetries`. Thread a sensitive-request mode through request/response, retry, and rate-limit logging (and strip it before the wire), then cover the actual transport stack; adding `x-promptfoo-silent` only here is insufficient because the retry logs remain raw.

mldangelo-oai (commented):
Reviewed exact head `31c373cb5a3dce7b17e9c89b99972f3965f709b7` in comment-only mode; I did not modify the branch. The provider-level redaction is useful and still needed on current `main`, but the blocking transport leak below means the stated security guarantee is not yet end-to-end. All 13 current required contexts passed on this head on May 18, but that signal is stale against a branch now 510 commits behind with a test conflict and no exact-head human review. I recommend a fresh-main forward port that preserves `getRemoteGenerationHeaders()` and `providerRemoteGenerationContextPayload()` while extending the privacy policy through the real fetch stack.

PR #8658, opened 2026-04-12
fix(azure): prevent chat failure data leaks

ianw-oai:
## Summary

Azure chat invalid-JSON and failure paths could expose request payloads, prompts, credentials, or peer-controlled response data through returned errors, causes, logs, metadata, retries, caches, and serialization. This change keeps bounded status and structural diagnostics while removing raw sensitive content from those paths.

## What changed

- Force Azure chat fetch diagnostics into redacted mode and remove the internal marker before transmission.
- Replace raw request/response/error serialization with bounded request shape, response shape, status, and allowlisted Azure code/parameter metadata.
- Sanitize authentication, proxy, transport, timeout, body-read, parse, rate-limit, streaming, and processing failures.
- Validate responses before cache publication and reuse, evict malformed cached values, and isolate in-flight work by validation, diagnostic, timeout, retry, and abort policy.
- Sanitize usage details, log probabilities, and finish reasons while preserving documented content-filter, refusal, audio, function/custom tool-call, data-source, multi-choice, and valid cached-response behavior.
- Preserve explicit authorization headers over URL userinfo case-insensitively.
- Document the safe Azure troubleshooting payload.

## Validation

- 683 focused backend tests across cache, fetch, Azure providers, shared providers, output, and sanitizer paths.
- TypeScript typecheck, architecture boundaries, changed-file lint/format hooks, and a full local `tsdown` build.
- Real local CLI evals with `--no-cache` and inspected JSON exports:
  - 13/13 success and compatibility scenarios passed.
  - 12/12 adversarial metadata and response-shape scenarios passed.
  - 8 intended non-streaming failures and 1 intended streaming failure remained sanitized.
  - Exported response/error/metadata channels and logs contained no body or credential sentinels.
- Dedicated security scans and fresh independent security, correctness, compatibility, error-path, regression, runtime, and simplification reviews.

## Compatibility and residual risk

This intentionally narrows unsafe error detail but preserves safe HTTP status, retry classification, allowlisted error code/parameter, token usage, and documented Azure response shapes. Validation used realistic loopback Azure/proxy/auth mocks; no credentialed live Azure service or external proxy was exercised. The branch currently conflicts with newer `main` changes and requires manual conflict resolution that preserves both the redaction work and current provider/cache behavior.


mldangelo-oai:
Final review receipt for exact head `66b3493c752110e2de88c8511e218908f22978a2`.

- Fresh security, correctness, error/concurrency, compatibility, breaking-change, test-effectiveness, runtime, and simplification reviews found no remaining actionable PR-local issue.
- Exact-head local validation passed: 683/683 focused tests, TypeScript typecheck, architecture boundaries, changed-file lint/format hooks, and a full local `tsdown` build.
- Real CLI `--no-cache` QA passed 13/13 success/compatibility and 12/12 adversarial metadata cases; 8 intended non-streaming failures and 1 intended streaming failure remained sanitized. Inspected response/error/metadata exports and logs contained no body or credential sentinels.
- All five review threads are resolved. The title and body now describe the final implementation, validation, compatibility, and residual risk.
- Exact-head CI is terminal: all 10 published CodeQL checks succeeded.

External gates remain: the PR conflicts with live `main` (`8240201173edd317cfb20526e2784061b09d768a`) in nine files, 12 required non-CodeQL contexts are absent rather than failed, and the required code-owner approval is missing. Validation used realistic loopback Azure/proxy/auth mocks; no credentialed live Azure service or external proxy was exercised.

src/providers/azure/chat.ts
```diff
@@ -325,129 +843,146 @@ export class AzureChatCompletionProvider extends AzureGenericProvider {
...
+      data = fetchResult.data;
     } catch (err) {
+      if (err instanceof FetchResponseParseError) {
+        const responseMetadata = {
+          status: err.status,
+          responseType: 'string',
+          responseLength: err.responseLength,
+        };
+        return {
+          error: `API returned invalid JSON response (status ${err.status})\n\nResponse metadata: ${JSON.stringify(responseMetadata, null, 2)}\n\nRequest metadata: ${JSON.stringify(getAzureChatRequestMetadata(body), null, 2)}`,
```
mldangelo, line 891: This is the load-bearing fix and it's correct: the bounded `getAzureChatRequestMetadata(body)` (role counts + content *lengths*, no content) replaces the merge-base's raw `Request body: ${JSON.stringify(body)}` leak. Worth calling out that **`main` still has the original leak** at its own `chat.ts:380` (`... ${responseData}\n\nRequest body: ${JSON.stringify(body, null, 2)}`), so this guarantee is genuinely not superseded — it's the reason to re-derive this specific hunk on `main` rather than close the PR. Please keep `chat-redaction.test.ts` with it as the regression proof.

src/util/fetch/errors.ts
```diff
@@ -6,6 +6,59 @@ export interface SystemError extends Error {
   cause?: unknown;
 }
 
+const SAFE_FETCH_ERROR_CODES: ReadonlySet<string> = new Set([
```
mldangelo, line 9: P3: `SAFE_FETCH_ERROR_CODES` omits `EPIPE`, but `isTransientConnectionError` (line 341) still treats `EPIPE` as transient. On a redacted (`x-promptfoo-silent`) request, `toFetchDiagnosticError` drops the code, so the scheduler classifies a send-time `EPIPE` as non-transient and skips a retry layer that an identical non-silent request would get. Add `EPIPE` to the safe set to keep classification consistent across silent/non-silent.

src/util/fetch/index.ts
```diff
@@ -628,18 +703,23 @@ export async function fetchWithRetries(
...
+        await handleRateLimitedResponse(response, url, i, maxRetries, redactDiagnostics);
         continue;
       }
 
       return response;
     } catch (error) {
       // Don't retry on abort - propagate immediately
       if (error instanceof Error && error.name === 'AbortError') {
+        if (redactDiagnostics) {
+          throw new FetchDiagnosticError({ errorType: 'AbortError' });
```
mldangelo, line 721: P3: the redacted branch throws `FetchDiagnosticError` whose `name` is `'FetchDiagnosticError'`, so downstream `err.name === 'AbortError'` guards (evaluator, redteam providers) no longer recognize the cancellation. Not reachable from Azure today (no signal is passed into `fetchWithCache`), but a latent trap for any silent caller that passes an abort signal. `cache.ts` deliberately preserves `error.name = 'AbortError'` on the body-read path — mirror that here (set the name, or carry `errorType` in a way the guards read).

src/providers/azure/chat.ts
```diff
@@ -325,129 +843,146 @@ export class AzureChatCompletionProvider extends AzureGenericProvider {
...
+        if (!isAzureChatChoiceProcessable(choice)) {
+          throw new Error('Azure chat response did not contain a processable choice');
+        }
 
         const message = choice?.message;
 
         // NOTE: The `n` parameter is currently (250709) not supported; if and when it is, `finish_reason` must be
         // checked on all choices; in other words, if n>1 responses are requested, n>1 responses can trigger filters.
-        finishReason = normalizeFinishReason(choice?.finish_reason) as string;
+        finishReason = getSafeAzureFinishReason(choice?.finish_reason);
```
mldangelo, line 950: P3: `getSafeAzureFinishReason` maps anything outside `{stop, length, tool_calls, content_filter}` to `undefined`. Azure AI Foundry now also hosts Anthropic Claude, whose documented `pause_turn` finish reason (and any future model-specific reason) gets dropped — a `finish-reason` assertion then fails and `isAzureChatChoiceCacheable` (needs `finishReason !== undefined`) won't cache the response, even though the content is fine. `finish_reason` is a bounded enum, not free-form peer text, so the leak rationale is weak here; consider passing it through like the merge-base `normalizeFinishReason` did.

mldangelo (commented):
## Review: Azure chat failure-path data-leak hardening

Ran a full security-focused pass over the tip (66b3493), with an adversarial redaction-bypass hunt, cache-poisoning analysis, a functional-regression sweep of the reworked chat path, and a per-file merge-risk assessment against current `main`. The redaction mechanism is sound and the tests have real teeth (`chat-redaction.test.ts` is a genuine loopback `http.Server` integration that plants a `secret-sentinel` in the prompt, response body, and api-key simultaneously and asserts it never appears in `error`/`cause`/`metadata`/`output`/logs). 564/564 tests pass on the branch. No introduced leak or cache-poisoning defect survived verification, and **all five open bot findings are already addressed at this tip** (details below).

**The core value is real and not superseded.** I verified directly that `main` still returns, on an Azure invalid-JSON upstream response:

```
API returned invalid JSON response (status ${status}): ${responseData}\n\nRequest body: ${JSON.stringify(body, null, 2)}
```

— leaking both the raw peer-controlled response body **and** the full request body (rendered prompts, system prompt, and any var-interpolated secrets) into `ProviderResponse.error`, which is persisted to the eval-results DB and shown in the UI/share links. This PR replaces that with bounded shape metadata (`getAzureChatRequestMetadata` → role counts + content *lengths* only). That is a genuine P1-class leak this change closes and nothing on `main` has fixed it. So this should **not** be closed as stale-and-superseded.

**But it should not be merged as-is either.** The branch is ~3 months old and conflicts with `main` in 9 files, and `main` has independently fixed much of the *peripheral* surface this PR also touches:

- `util/fetch/index.ts` — `main` added `urlForLog()` / `sanitizeUrl()` stripping basic-auth + `api_key`/`token`/`signature`/`password` from retry & rate-limit diagnostics.
- `util/fetch/errors.ts` — `main` added `normalizeNonNegativeMs()` and `isAbortError()` (this independently covers the "soft rate-limit" bot P2).
- `src/cache.ts` — `main` (df0894a) added HMAC secret-fingerprinting of cache-key material and reworked cloud auth, and preserves `AbortError` on body-read. This is the **highest-risk conflict**: both sides rewrote the same functions, so a naive resolution toward the branch would clobber `main`'s cache-key protection.
- `azure/chat.ts` — `main` added its own optional-chaining hardening, Opus sampling-param deprecation, and MAI reasoning-model handling.

So a large fraction of the +738/-213 shared-infra diff is now redundant, and the merge reconciliation on `cache.ts` / `util/fetch/*` is where the real risk lives — not in the redaction logic itself.

### Recommendation

Re-derive the **targeted** `azure/chat.ts` fix on top of current `main`: replace the still-live line-380 leak with the bounded `getAzureChatRequestMetadata` shape + `FetchResponseParseError`, keep the `chat-redaction.test.ts` integration test (it's the regression proof for the original attack vector), and **drop the shared `cache.ts` / `util/fetch/*` changes that `main` has already superseded**. That preserves the one guarantee only this PR provides, keeps `main`'s independent improvements, and shrinks a 4.5k-line stale diff to a reviewable targeted fix. If any shared-infra change is still needed after rebasing (e.g. the `x-promptfoo-silent` redacted-diagnostics plumbing), split it into its own PR so reviewers can reason about the cache/fetch blast radius separately from the Azure fix.

### Open bot findings — all resolved at tip (verified)

- **P1 ×2, cache unvalidated 2xx body**: refuted. Azure now calls `fetchWithCache(..., 'json')` (not `'text'`), and publication is gated on validation *before* the entry is visible — `prepareFetchResponse` returns `cacheable: false` for non-2xx, top-level `error` keys, and anything the caller's `isResponseCacheable` rejects, synchronously inside the shared in-flight promise, so concurrent callers await the validated result. No publish-then-evict window.
- **P2 malformed choices**: actually *fixed* here — `isAzureChatChoiceProcessable` rejects `choices:[{}]` and content-less messages into a clean `{error}`. (`main`, post its own optional-chaining fix, still returns success with `output: undefined` for that case — the branch is stricter.)
- **P2 inflight policy isolation**: the validation-id keying does separate differing `isResponseCacheable` policies (see the one caveat below).
- **P2 soft rate-limit**: fixed — both hard 429s and soft header-only throttles now throw a structured `HttpRateLimitError`; redaction strips raw headers but `retryAfterMs`/`resetAt` are still parsed from headers, so scheduler classification is intact.

### Minor issues in the introduced code (all P3, non-blocking)

1. **Redacted `EPIPE` loses its retry layer** — `SAFE_FETCH_ERROR_CODES` omits `EPIPE`, but `isTransientConnectionError` still treats `EPIPE` as transient. On a *silent* request, `toFetchDiagnosticError` strips the code, so the scheduler no longer classifies it transient — one fewer retry than an identical non-silent request. Add `EPIPE` to the safe set.
2. **Redacted `AbortError` loses its `.name`** — the redacted branch throws `FetchDiagnosticError` (name `'FetchDiagnosticError'`), so downstream `err.name === 'AbortError'` guards stop recognizing it. Not reachable from Azure today (no signal passed), but a latent trap for any silent caller that passes an abort signal. `cache.ts` deliberately re-tags `error.name = 'AbortError'` for the body-read path — worth mirroring here.
3. **`finishReason` allowlist drops non-OpenAI reasons** — `SAFE_FINISH_REASONS` is `{stop, length, tool_calls, content_filter}`, so Azure-hosted Claude's documented `pause_turn` (and any future model reason) becomes `undefined`. Content is still returned, but `finish-reason` assertions fail and `isAzureChatChoiceCacheable` won't cache these responses. `finish_reason` is a small enum, not free-form peer text — consider passing it through (as the merge-base `normalizeFinishReason` did) rather than allowlisting.
4. **Inline `isResponseCacheable` closures** would each mint a new validation-id and defeat in-flight dedup. No current caller does this (Azure passes stable module-level refs), but worth a comment on `getResponseCacheabilityCheckId` so a future caller doesn't pass a per-call arrow.

Happy to help re-derive the targeted fix against current `main` if useful.

src/util/fetch/index.ts
```diff
@@ -605,17 +648,49 @@ function formatFetchErrorMessage(error: unknown): string {
...
   maxRetries?: number,
 ): Promise<Response> {
-  const contextMaxRetries = getFetchRetryContextMaxRetries();
-  maxRetries = Math.max(0, maxRetries ?? contextMaxRetries ?? 4);
+  maxRetries = getEffectiveFetchMaxRetries(maxRetries);
 
   let lastErrorMessage: string | undefined;
+  let lastDiagnosticError: FetchDiagnosticError | undefined;
   const backoff = getEnvInt('PROMPTFOO_REQUEST_BACKOFF_MS', 5000);
+  const redactDiagnostics = isFetchLoggingSuppressed(url, options);
```
mldangelo-oai, line 693: [P2] Keep log suppression separate from returned-error redaction

`x-promptfoo-silent` predates this change as a request/response logging-suppression marker, but using it here as `redactDiagnostics` changes caller-visible exceptions for every existing silent caller. HTTP validation, Mistral, and moderation can now replace useful network, proxy, TLS, or parse failures with generic `FetchDiagnosticError` details even though those callers only opted out of logging; the new fetch/cache tests explicitly exercise the changed thrown semantics. Keep `silent` as a logging control and opt Azure into a separate redacted-diagnostics policy so unrelated callers retain actionable errors.

src/util/fetch/monkeyPatchFetch.ts
```diff
@@ -46,6 +48,13 @@ export async function monkeyPatchFetch(
 
   const originalBody = opts.body;
 
+  // This header controls local diagnostics only. Remove it before the wire so
+  // silent and ordinary requests remain identical upstream and in cache.
+  if (effectiveHeaders.has('x-promptfoo-silent')) {
+    effectiveHeaders.delete('x-promptfoo-silent');
+    opts.headers = Object.fromEntries(effectiveHeaders.entries());
```
mldangelo-oai, line 55: [P2] Preserve case-insensitive auth overwrite after stripping the silent header

Converting `effectiveHeaders` through `Object.fromEntries` lowercases a caller's `Authorization` key. The Cloud block then spreads an uppercase `Authorization` into the same object, so native fetch coalesces them into `Bearer caller, Bearer cloud` instead of letting the Cloud token overwrite; `getHeadersForCacheKey` uses `Headers.set` and keys that request as only the Cloud token, so cache and wire identities also diverge. Keep these headers as a `Headers` instance and use `.set('Authorization', ...)`, with a silent Cloud request regression test that starts with caller auth.

mldangelo-oai (commented):
Reviewed exact head `66b3493c752110e2de88c8511e218908f22978a2` in comment-only mode; I did not modify the branch. The core Azure redaction fix remains valid and needed, but this branch is 510 commits behind current `main`, has nine real conflict paths, and currently publishes only CodeQL among 13 required contexts, so the live CI signal is incomplete. I found two additional distinct P2 regressions below; the three existing P3 comments in the earlier exact-head review still apply. I recommend re-deriving the bounded Azure diagnostic and its tests on current `main` rather than merging the stale shared fetch/cache changes wholesale.

PR #8661, opened 2026-04-12
fix(redteam): redact plugin config logs

ianw-oai:
Stops redteam debug logs from storing plugin config values.


mldangelo-oai (approved):
Approved. The config is redacted, the face card is not.

Issue #185, opened 2023-09-22
Bug: OpenAI No embedding returned


jamesbraza:
https://platform.openai.com/docs/guides/embeddings/what-are-embeddings has an example JSON blob structure from OpenAI `/embeddings`.

Continuing my LocalAI hacking from https://github.com/promptfoo/promptfoo/issues/176 and https://github.com/promptfoo/promptfoo/issues/177, here is an example cache entry made by `promptfoo` from LocalAI's `/v1/embeddings`:

```json
{
    "expireTime": 1696613448918,
    "key": "fetch:http://localhost:8080/v1/embeddings:{\"method\":\"POST\",\"body\":\"{\\\"input\\\":\\\"<redacted by James>.\\\\n\\\",\\\"model\\\":\\\"bert-embeddings\\\"}\"}",
    "val": "{\"object\":\"list\",\"model\":\"bert-embeddings\",\"data\":[{\"embedding\":[-0.058698192,0.019647375,...,-0.0041024373],\"index\":0,\"object\":\"embedding\"}],\"usage\":{\"prompt_tokens\":0,\"completion_tokens\":0,\"total_tokens\":0}}"
}
```

Currently, OpenAI provider has [`data?.data?.[0]?.embedding`](https://github.com/promptfoo/promptfoo/blob/0.24.1/src/providers/openai.ts#L123C25-L123C51).  Running this, I get:

```
[FAIL] API response error: Error: No embedding returned: {"error":{"code":500,"message":"rpc error: code = Unavailable desc = error reading from server: EOF","type":""}}
```

Changing the source to use `data.data[0].embedding`, it works fine, so I think it's a JSON traversal issue.  Why are the `?` there in the source now?

jamesbraza:
Strangely, I can't repro this any more.  I opened https://github.com/promptfoo/promptfoo/pull/190 to make it easier to read and report issues like this in the future.

PR #8662, opened 2026-04-12
fix(server): restrict web eval prompt sources

ianw-oai:
Blocks server-side prompt sources in web-created eval jobs unless explicitly enabled.


src/server/routes/eval.ts
```diff
@@ -135,6 +283,12 @@ evalRouter.post('/job', (req: Request, res: Response): void => {
     return;
   }
 
+  const promptValidationError = validateWebEvalPrompts(result.data.prompts);
+  if (promptValidationError) {
+    res.status(400).json({ error: promptValidationError });
+    return;
+  }
+
   // Use validated data but merge providers from req.body to preserve nested config fields.
```
mldangelo, line 292: **Blocking — provider configs are an unguarded, strictly-worse server-side-source hole on this same route.**

`providers` is copied raw from `req.body` (the line below) and handed to `evaluateWithSource → loadApiProviders`, which resolves `file://*.js`, `python:*.py`, and `exec:` providers as **executable server-side code**. So a request with a perfectly benign inline prompt and:

```json
{ "prompts": ["hello"], "providers": [{ "id": "exec: touch /tmp/pwned" }] }
```

is remote code execution — a superset of the prompt LFI this PR closes — and `validateWebEvalPrompts` never inspects it. Any guard that gates web-eval "server-side sources" has to cover providers (and, transitively, `file://`/`python:`/`javascript:` assertions, transforms, and `extensions` hooks), not just top-level prompts. This is the finding that makes me want to rethink the approach: a prompts-only string filter can't express the actual threat model.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+}
+
+function hasKnownPromptFileExtension(value: string): boolean {
+  const lower = value.toLowerCase();
+  return [...VALID_FILE_EXTENSIONS, ...SERVER_PROMPT_SOURCE_EXECUTABLE_EXTENSIONS].some(
+    (extension) => lower.endsWith(extension),
+  );
+}
+
+function hasPathLikeSeparator(value: string): boolean {
```
mldangelo, line 72: **Blocking (false positives) — `hasPathLikeSeparator` rejects ordinary `A/B` prose with HTTP 400.**

Any prompt containing `X/Y` where the token before the first `/` is length>1 and matches `[\w.-]+` trips this. I ran these through the actual guard and every one returns 400:

`and/or` · `TCP/IP` · `read/write locks` · `Answer yes/no` · `N/A` · `CI/CD` · `The ratio is 16/9` · `Explain the difference between TCP/IP layers`

These are everyday web-eval prompts. Since the guard's purpose is to *not* break legit evals, this is a significant regression. It stems from the guard trying to be cleverer (and narrower) than the loader's actual rule — see the note on reusing `maybeFilePath` at line 137.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+  if (!/[a-z0-9]/i.test(candidate)) {
+    return false;
+  }
+
+  if (hasKnownPromptFileExtension(candidate)) {
+    return true;
+  }
+
+  return (
+    candidate.charAt(candidate.length - 3) === '.' || candidate.charAt(candidate.length - 4) === '.'
```
mldangelo, line 121: **Blocking (false positives) — the `charAt(len-3)/charAt(len-4) === '.'` check rejects prompts ending in a decimal or abbreviation.**

`hasFileLikeExtension` takes only the last whitespace token and flags a `.` 3rd/4th from the end. Verified returning HTTP 400 against the guard:

`The answer is 3.14` · `See you at 5 p.m.` · `Where is the capital of the U.S.` · `Does she have a Ph.D.` · `e.g.`

(Inconsistent, too: `v2.0` and `3.5` pass because the dot lands elsewhere.) This mirrors `maybeFilePath`'s own `charAt(len-3/len-4)` clause, but there it only decides file-vs-glob for a string already known to be file-ish; lifted into a standalone gate over arbitrary prose it just rejects normal sentences.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+  }
+
+  if (hasPathLikeSeparator(value) || hasTrailingFileLikeExtension(value)) {
+    return true;
+  }
+
+  return value.includes('*') && !/\s/.test(value);
+}
+
+function isServerPromptSourceReference(prompt: string): boolean {
```
mldangelo, line 137: **Root cause — this re-implements `maybeFilePath` with a divergent heuristic instead of reusing it.**

promptfoo already owns the authoritative "is this a server-side source?" decision: `maybeFilePath` (src/prompts/utils.ts:11, exported and already imported by src/prompts/index.ts) plus the `exec:` check in `processPrompt`. That *is* the function the loader gates on at src/evaluate.ts:224. This PR ships ~120 lines of parallel detection in a route file that is deliberately **narrower** than the loader (to avoid false-positiving on prose) — and that narrowness is exactly what creates the residual divergences and the false positives above.

Two problems with a second implementation of a security boundary: (a) it will drift from `maybeFilePath`/`VALID_FILE_EXTENSIONS` on any future change, silently reopening gaps; (b) it lives apart from the code it must mirror, with no test asserting `guard ⊇ loader`.

Recommendation: don't detect-and-reject in the route. Thread a "web job → never load files/exec" mode into the prompt-processing path (`processPrompt`/`processString`) so the **loader itself** refuses to read files regardless of string shape. That removes the arms race entirely and also closes the provider/assertion/extension vectors if applied at the loader boundary. If a pre-filter must stay, define it as a superset of `maybeFilePath(raw) || raw.startsWith('exec:')`, move it next to `maybeFilePath`, and unit-test it — accepting the prose false positives as the price of correctness, or better, avoid them by gating at the loader instead.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+  );
+}
+
+function validateWebEvalPrompts(prompts: (string | Record<string, unknown>)[]): string | undefined {
+  if (getEnvBool(SERVER_PROMPT_SOURCE_OPT_IN, false)) {
+    return undefined;
+  }
+
+  if (
+    prompts.some((prompt) => typeof prompt === 'string' && isServerPromptSourceReference(prompt))
```
mldangelo, line 185: **Object-form prompts skip the guard — latent, not currently exploitable, but the tests freeze it as intended.**

This callback only inspects `typeof prompt === 'string'`, so `{raw:'exec:...'}` / `{id:'file://...'}` bypass validation (the schema is `z.union([z.string(), z.record(...)])`).

I verified this is **not exploitable today**: `processPrompts` runs object prompts through `PromptSchema.parse` and keeps `raw` as literal content — a `{raw:'file://secret.txt'}` object does not read the file. So this is a latent gap, not an active bypass. The concern is durability: the guard's own signature accepts objects, two tests assert the object-form returns 200 (with `evaluateWithSource` mocked, so they never prove safety), and `id: 'file://...'` is a very natural thing for someone to later make resolve. If that happens, this gate misses it and CI stays green. Cheap defense-in-depth: also inspect `.raw`/`.id` of object entries here.

src/server/routes/eval.ts
```diff
@@ -135,6 +283,12 @@ evalRouter.post('/job', (req: Request, res: Response): void => {
     return;
   }
 
+  const promptValidationError = validateWebEvalPrompts(result.data.prompts);
```
mldangelo, line 286: **Per-route wiring — sibling enqueue paths are unguarded, and this must be re-inserted after the rebase.**

`validateWebEvalPrompts` is called only here. Two other request-controlled paths enqueue evals without it:
- redteam `POST /run` (src/server/routes/redteam.ts:231) → `evalJobs.set` + `doRedteamRun` from the same job map.
- `POST /replay` (eval.ts:748) wraps a request `prompt` into the test suite.

(Both currently wrap prompts as objects, so they inherit the same not-exploitable-today status as #185 — but they're the same class of unguarded surface, which is why a per-route string filter is the wrong seam. A single check at the shared enqueue/loader boundary would cover all of them plus providers.)

Also: this branch is **CONFLICTING** and 7 weeks stale — `main` rewrote this handler to `async` with `sourceEvalId` / `evalJobService.create`. After rebase, make sure this guard lands right after the `safeParse` in the new async handler.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+
+function isServerPromptSourceReference(prompt: string): boolean {
+  const trimmed = prompt.trim();
+  if (!trimmed) {
+    return false;
+  }
+
+  const lower = trimmed.toLowerCase();
+
+  if (lower.startsWith('exec:') || lower.startsWith('file://')) {
```
mldangelo, line 145: **Minor false positives — `exec:`/`file://` matching diverges from the loader.**

Here the guard lowercases and trims (`lower.startsWith('exec:')`), but the loader's check is `prompt.raw.startsWith('exec:')` — case-sensitive and un-trimmed. So the guard 400s inline prompts the loader would treat as plain text: `Exec: summarize this document`, `EXEC ORDER 66 explained`, or a leading-space `  exec: ...`. Same divergence class as the `maybeFilePath` note — another reason to derive from the loader's actual predicate rather than a lookalike.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+
+  if (/^(?:~[/\\]|\.{1,2}[/\\]|[/\\]|[a-z]:[/\\])/i.test(trimmed)) {
+    return true;
+  }
+
+  if (/^(?:portkey|langfuse|helicone):\/\//i.test(trimmed)) {
+    return false;
+  }
+
+  if (/^[a-z][a-z0-9+.-]*:\/\//i.test(trimmed)) {
```
mldangelo, line 157: **False positive — any prompt beginning with a URL scheme is rejected.**

`^[a-z][a-z0-9+.-]*://` 400s a legitimate inline prompt like `https://example.com — summarize what this page is about`. The intent (catch `s3://`, `gs://`, etc. remote loaders) is reasonable, but as written it also blocks prose that merely *starts* with a URL. If remote schemes are the target, match against the actual set promptfoo loads rather than "any scheme://".

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
 // Running jobs
 export const evalJobs = new Map<string, Job>();
 
+const SERVER_PROMPT_SOURCE_OPT_IN = 'PROMPTFOO_ALLOW_SERVER_PROMPT_SOURCES';
+const SERVER_PROMPT_SOURCE_EXECUTABLE_EXTENSIONS = [
```
mldangelo, line 46: **Drift risk — duplicated executable-extension list.**

`['.bash','.bat','.cmd','.exe','.pl','.ps1','.rb','.sh']` re-hardcodes the loader's list in src/prompts/index.ts (~line 202, different order), and can't replicate the loader's executable-bit (`stat().mode & 0o111`) fallback that loads extensionless executables. A security-critical list maintained in two places will desync. Export one constant and share it (same theme as reusing `maybeFilePath`).

test/server/routes/eval.sharing.test.ts
```diff
@@ -190,4 +191,274 @@ describe('Eval Routes - Sharing behavior', () => {
...
+    expect(numericExtensionResponse.status).toBe(400);
+    expect(numericExtensionResponse.body.error).toContain(
+      'Server-side prompt sources are disabled',
+    );
+    expect(mixedExtensionResponse.status).toBe(400);
+    expect(mixedExtensionResponse.body.error).toContain('Server-side prompt sources are disabled');
+    expect(mockedEvaluateWithSource).not.toHaveBeenCalled();
+  });
+
+  it('allows object prompt sources from web job creation', async () => {
```
mldangelo, line 407: **Test soundness — these tests codify the object-form pass-through as intended, and there's no coverage for the real gaps.**

`allows object prompt sources` (here) and `allows object prompt id sources` (line 419) assert `{raw:'exec:/bin/echo ...'}` / `{id:'prompts/runner'}` return 200 with `evaluateWithSource` **mocked** — so they lock in "object prompts skip the guard" as expected behavior while never exercising the loader that makes it safe. If a future change makes object `raw`/`id` file-resolve, these stay green through the regression.

Bigger gap: nothing here covers the provider-config exec/`file://` vector (the RCE at eval.ts:300), the redteam `/run` route, or the false-positive prompts (`and/or`, `3.14`, `TCP/IP`). Per the repo's adversarial-testing policy, a security fix should ship regression tests for the attack vectors it claims to close — and this suite tests mostly the easy string cases while blessing a bypass.

mldangelo (commented):
## Security review: restrict web eval prompt sources

The intent is right and the direct vulnerability is real — I confirmed end-to-end that a web `/job` with `prompts: ['file:///etc/passwd']` flows through `evaluateWithSource → evaluate() → processPrompts` (src/evaluate.ts:224) and reads the file, and that this guard blocks it. But after building the guard into a standalone harness and running it against both the real loader (`maybeFilePath`) and ~50 attack/benign strings, my finding is that the guard is **simultaneously under-inclusive on security and over-inclusive on legitimate prompts**, and the root cause is that it re-implements promptfoo's own file-detection with a divergent hand-rolled heuristic. The workflow review reached the same conclusions independently. I'd hold this for rework.

### What I verified working
- The core string vuln is closed: every realistic server-side form — `file://`, `exec:`, `/etc/passwd`, `./x`, `../x`, `~/x`, `C:\...`, UNC, `prompts/*.txt`, `**/*.env`, `data.json`, `file.js:func` — is correctly blocked (HTTP 400). Verified against the extracted guard.
- The PR's 28 tests pass locally (after `npm ci` — the branch is 7 weeks stale and its lockfile predates `better-sqlite3` changes).

### Blocking

1. **Providers are an unguarded, strictly-worse hole on the same route** (inline at `eval.ts:300`). The handler copies `providers` raw from `req.body` and passes them to `evaluateWithSource → loadApiProviders`, which resolves `file://*.js`, `python:*.py`, and `exec:` providers as **executable server-side code**. A web request with a benign inline prompt and `providers: [{ id: 'exec: touch /tmp/pwned' }]` is RCE — worse than the LFI this PR blocks — and the prompts-only guard never looks at it. The fix doesn't cover its own threat model.

2. **The heuristic hard-400s a large class of ordinary inline prompts** (inline at `eval.ts:72` and `eval.ts:121`). I reproduced all of these returning HTTP 400 against the actual guard: `and/or`, `TCP/IP`, `read/write locks`, `Answer yes/no`, `N/A`, `CI/CD`, `The ratio is 16/9`, `The answer is 3.14`, `See you at 5 p.m.`, `capital of the U.S.`, `e.g.`. These are extremely common web-eval prompts; the guard's whole value proposition is not breaking legit evals, and it breaks a lot of them.

3. **Root cause — it reimplements `maybeFilePath` instead of reusing it** (inline at `eval.ts:137`). promptfoo already has the authoritative "is this a file/exec source?" decision (`maybeFilePath` in src/prompts/utils.ts, exported, plus the `exec:` check in processPrompt) — it's literally the function the loader gates on. This PR ships a ~120-line parallel heuristic in a route file that is *narrower* than the loader (to spare prose), and that narrowness is exactly what produces both the false positives (#2) and the residual divergences. A security boundary that must mirror another module but lives apart from it will drift. The robust fix is to gate on the loader's own decision — ideally by threading a "web job: never load files/exec" mode into the prompt-processing path so the loader itself refuses, rather than a detect-and-reject shadow — co-located with `maybeFilePath` and unit-tested against a shared corpus.

### Should fix

4. **Sibling enqueue routes are unguarded** (inline at `eval.ts:286`). `validateWebEvalPrompts` is wired only into `POST /job`. The redteam `POST /run` (redteam.ts:231) enqueues eval jobs from the same `evalJobs` map with no equivalent check, and `POST /replay` (eval.ts:748) takes a request-controlled prompt. A per-route guard is the wrong seam; the check belongs at the shared enqueue/loader boundary all of these funnel through.

5. **Object-form prompts skip the guard, and the tests bless it** (inline at `eval.ts:185` and the test file). The `.some()` callback only inspects `typeof prompt === 'string'`, so `{raw:'exec:...'}` / `{id:'file://...'}` skip validation. I verified this is **not exploitable today** — `processPrompts` runs object prompts through `PromptSchema.parse` and treats `raw` as literal content (I confirmed a `{raw:'file://secret'}` object does not read the file). So it's a latent gap, not an active bypass. But two tests (`allows object prompt sources`, line 407) assert these return 200 with `evaluateWithSource` mocked — they lock in the pass-through as intended behavior while never proving it's safe, so if the loader ever resolves object `raw`/`id` as a path, this guard silently misses it and the tests stay green.

6. **`exec:`/`file://` matching diverges from the loader** (inline at `eval.ts:145`): the guard lowercases and trims, but the loader's `exec:` check is case-sensitive and un-trimmed — so inline prompts like `Exec: summarize this` or a leading-space `  exec:...` are 400'd even though the loader would treat them as plain text. Minor false positives, same divergence class as #3.

7. **Duplicated executable-extension list** (inline at `eval.ts:46`): `['.bash','.bat',...]` re-hardcodes the loader's list from src/prompts/index.ts and can't replicate its executable-bit (`mode & 0o111`) fallback — a security-critical list guaranteed to drift. Share one exported constant.

### Mechanical / non-blocking
- **CONFLICTING, 7-week-stale branch.** `main` refactored the `/job` handler to `async` with `sourceEvalId` / `evalJobService.create`; the guard must be re-inserted right after the `safeParse` in the rebased async handler. Can't merge as-is.
- `src/envars.ts:53` inserts `PROMPTFOO_ALLOW_SERVER_PROMPT_SOURCES` out of alphabetical order (it lands after the `ENABLE_*` block). Trivial.
- Docs: the env-var table row is fine, but `site/docs/usage/self-hosting.md` — the page for exactly the deployments at risk — doesn't mention this control or that the default is now to block server prompt sources. That's where operators will look.

### Findings I checked and am NOT raising
- **No exploitable string bypass.** The guard/loader divergences where the guard passes a `/`-containing string (e.g. `What is 2/3`, `tell me about C/C++`) are all benign prose that the loader would glob to **nothing** — I verified none resolve to a real file read. So the guard is under-inclusive vs `maybeFilePath` but not in a way that leaks files.
- **Object-form is not an active RCE** (see #5) — despite how it looks, verified non-exploitable against the real loader.
- **The 400 body naming the env var is not an info leak** — it's intended operator guidance, not a secret.

src/server/routes/eval.ts
```diff
@@ -40,6 +42,152 @@ export const evalRouter = Router();
...
+  if (/^(?:portkey|langfuse|helicone):\/\//i.test(trimmed)) {
+    return false;
+  }
+
+  if (/^[a-z][a-z0-9+.-]*:\/\//i.test(trimmed)) {
+    return true;
+  }
+
+  if (
+    (trimmed.startsWith('{') && trimmed.endsWith('}')) ||
```
mldangelo-oai, line 162: [P1] Do not exempt brace/bracket-wrapped glob sources

This return happens before the glob checks, but the real prompt loader still sends the same string through `globSync`. On this exact head, `{__promptfoo_no_match_*,package.json}` passes the web guard and `readPrompts` expands it to the local `package.json`, returning its full contents. Replacing the second branch with `conductor-setup.sh` produces an executable prompt function that reaches `execFile` when evaluated; bracket-wrapped patterns such as `[c]*[h]` work too. This is an active string-prompt bypass of the disabled-by-default control, not the non-exploitable object-prompt case from the earlier review. Remove the blanket wrapper exemption (or require successful JSON parsing plus a non-source classification) and add route/loader parity tests for brace and bracket globs.

src/server/routes/eval.ts
```diff
@@ -135,6 +283,12 @@ evalRouter.post('/job', (req: Request, res: Response): void => {
     return;
   }
 
+  const promptValidationError = validateWebEvalPrompts(result.data.prompts);
+  if (promptValidationError) {
+    res.status(400).json({ error: promptValidationError });
```
mldangelo-oai, line 288: [P2] Surface this recovery message in the Web UI

The server returns the actionable `PROMPTFOO_ALLOW_SERVER_PROMPT_SOURCES` guidance here, but `RunTestSuiteButton` discards the body on any non-OK response and throws only `HTTP error! status: 400`. A supported Web UI config with scalar `prompts: 'file://prompt.txt'` is normalized to the string form this guard rejects, so the user sees a generic toast/alert and cannot discover the opt-in or which field failed. Parse and display a non-empty JSON `error` value before falling back to the status-only message, and add a component test for this new 400 response plus a non-JSON fallback case.

mldangelo-oai (commented):
Follow-up on the same exact head after reconciling the existing review threads. The July 4 security review still applies, but one new concrete P1 means its "no exploitable string bypass" conclusion should be withdrawn; I also found one separate recovery-UX regression, both noted inline.

The recommendation is otherwise unchanged: do not merge or mechanically rebase this route-local heuristic. The security need is real and has not landed on main, but this implementation is both bypassable and overbroad, and every changed path now overlaps current main (with a real conflict in the refactored `/job` handler). The safer replacement is a current-main policy at the shared source/loader boundary, covering prompts and the other executable source surfaces already identified in review.

For the new bypass, I reproduced brace expansion with the repository's real glob and prompt loader: the package payload returned the full local `package.json`; the executable payload produced a prompt function backed by `processExecutableFile` (I did not invoke it). `git diff --check` passes. No branch changes were made.

PR #8665, opened 2026-04-12
fix(eval): scope result ratings to eval

ianw-oai:
Rejects rating updates when the result is not part of the URL eval.


mldangelo-oai (approved):
Approved. Tight scope, suspiciously handsome execution.

PR #8677, opened 2026-04-13
fix(redteam): throw immediately on non-OK HTTP responses in redteam discover

### LINKED ISSUE #8581: `promptfoo redteam discover` fails with `Unknown task: target-purpose-discovery`
**Describe the bug**

`promptfoo redteam discover` fails with a 400 Bad Request from the remote API on every request. The server returns `{"error":"Invalid task","message":"Unknown task: target-purpose-discovery"}`.

`promptfoo redteam generate` works correctly with the same config and credentials, hitting the same endpoint (`https://www.promptfoo.app/api/v1/task`).

**To Reproduce**

1. Set up a valid `promptfooconfig.yaml` with a working provider target
2. Run `promptfoo redteam generate` --> succeeds (HTTP 200)
3. Run `promptfoo redteam discover` with the same config --> fails (HTTP 400)

Debug log from `redteam discover` (failing):
```json
{
  "url": "https://www.promptfoo.app/api/v1/task",
  "method": "POST",
  "requestBody": "{\"state\":{\"currentQuestionIndex\":0,\"answers\":[]},\"task\":\"target-purpose-discovery\",\"version\":\"0.121.3\",\"email\":\"<redacted>\"}",
  "status": 400,
  "response": "{\"error\":\"Invalid task\",\"message\":\"Unknown task: target-purpose-discovery\"}"
}
```

Debug log from `redteam generate` (working, same endpoint, same credentials):
```json
{
  "url": "https://www.promptfoo.app/api/v1/task",
  "method": "POST",
  "requestBody": "{\"task\":\"excessive-agency\",\"version\":\"0.121.3\",\"email\":\"<redacted>\", ...}",
  "status": 200
}
```

The only difference between the two requests is the `task` field value. Plugin task names like `excessive-agency` return 200; `target-purpose-discovery` returns 400.

**Expected behavior**

`promptfoo redteam discover` completes successfully and outputs a structured profile of the target (purpose, limitations, tools, user context) as described in the [documentation](https://www.promptfoo.dev/docs/red-team/discovery/).

**Screenshots**

N/A

**System information:**

- Promptfoo version: 0.121.3 free tier

octo-patch:
Fixes #8581

## Problem

When `promptfoo redteam discover` receives a non-OK HTTP response from the remote server (e.g., HTTP 400 `{"error":"Invalid task","message":"Unknown task: target-purpose-discovery"}`), the `doTargetPurposeDiscovery` function was silently calling `continue` in the while loop instead of propagating the error. This caused the same failing request to be retried up to `MAX_TURN_COUNT` (10) times, giving no useful feedback to the user.

## Solution

Replace the `continue` with an immediate `throw` that surfaces the HTTP status code and error detail from the server response. The error message is parsed from JSON if possible (extracting `message` or `error` fields), otherwise the raw response text is used. This matches how `data.error` is already handled later in the same loop.

Before:
```
[Target Discovery Agent] Error getting the next question from remote server: {"error":"Invalid task","message":"Unknown task: target-purpose-discovery"}
[Target Discovery Agent] Error getting the next question from remote server: {"error":"Invalid task","message":"Unknown task: target-purpose-discovery"}
... (repeated 10 times)
```

After:
```
An unexpected error occurred during target scan: Remote server returned HTTP 400: Unknown task: target-purpose-discovery
```

## Testing

- Added two new tests to `test/redteam/commands/discover.test.ts`:
  1. Verifies that a 400 JSON error response throws immediately and does not retry (fetch called exactly once)
  2. Verifies that a non-JSON error response (e.g., 503 plain text) is also handled correctly
- All 11 tests in the discover test suite pass

### DIFF
--- a/site/docs/red-team/discovery.md
+++ b/site/docs/red-team/discovery.md
@@ -42,3 +42,13 @@ The agent discovers four key areas:
 4. **User Context**: How the system perceives and categorizes users
 
 The responses are synthesized into a comprehensive profile to inform attack strategies. For privacy, target responses are not stored except in error cases where they may appear in Promptfoo Cloud's error logs for debugging purposes.
+
+## Troubleshooting
+
+If `promptfoo redteam discover` fails with `Remote server returned HTTP 400: Unknown task: target-purpose-discovery` (or a similar 400/404 error), your client is most likely out of date. Update with `npm install -g promptfoo@latest` and rerun.
+
+Other common errors:
+
+- **HTTP 401 / 403** — Run `promptfoo auth login` and confirm your account has access to target discovery.
+- **HTTP 429** — You are being rate limited; wait a moment and try again.
+- **HTTP 5xx** — The remote generation service is temporarily unavailable. Retry in a few minutes.
--- a/src/redteam/commands/discover.ts
+++ b/src/redteam/commands/discover.ts
@@ -127,24 +127,53 @@ export function normalizeTargetPurposeDiscoveryResult(
   };
 }
 
+function extractStringField(value: unknown): string | undefined {
+  if (typeof value !== 'string') {
+    return undefined;
+  }
+  const trimmed = value.trim();
+  return trimmed || undefined;
+}
+
 async function getRemoteResponseErrorDetail(response: Response): Promise<string> {
-  const errorText = await response.text();
-  const fallbackDetail = errorText.trim() || response.statusText || 'Unknown error';
-
+  const rawText = (await response.text()).trim();
+  const fallback = rawText || response.statusText || 'Unknown error';
+  if (!rawText) {
+    return fallback;
+  }
   try {
-    const errorJson = JSON.parse(errorText) as unknown;
-    if (errorJson && typeof errorJson === 'object') {
-      const { message, error } = errorJson as Record<string, unknown>;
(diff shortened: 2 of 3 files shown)


mldangelo (approved):
LGTM — thanks for tracking this down, @octo-patch! The root cause analysis in the PR description is spot-on: the `continue` on non-OK was silently burning 10 retries per run and burying the real server response.

I pushed one follow-up commit (`a04a21d1f`) on top of your work:

- Factored the error construction into `buildRemoteErrorFromResponse` so `doTargetPurposeDiscovery` stays under the cognitive-complexity threshold.
- Added a status-aware hint alongside the raw server detail — 400/404 point users at `npm install -g promptfoo@latest`, 401/403 at `promptfoo auth login`, 429 at rate limits, 5xx at transient service availability. The exact "Unknown task" failure from #8581 now prints an actionable next step instead of just the HTTP status.
- Added a Troubleshooting section to `site/docs/red-team/discovery.md` covering the same failure modes so users hitting this via search don't have to file a new issue.
- Updated tests: the existing 400 test now asserts the update hint, and added a 401 test for the auth hint path.

I also verified end-to-end against a local mock 400 server: exactly one request, clean exit, actionable output. All 27 CI checks are green on the new head.

mldangelo-oai:
Thank you!

PR #8682, opened 2026-04-13
fix: correct rouge assertion reason messages when inverse: true

### LINKED ISSUE #8681: Bug: rouge assertion returns incorrect reason message when inverse: true
### Summary

The `rouge` assertion in `src/assertions/rouge.ts` produces **misleading/incorrect reason messages** when `inverse: true` is set. The pass/fail logic is correct, but the human-readable `reason` field is hardcoded and does not account for the inverse condition.

### Affected File

`src/assertions/rouge.ts`

### Current Code

```typescript
const pass = score >= threshold != inverse;
return {
  pass,
  score: inverse ? 1 - score : score,
  reason: pass
    ? `${baseType.toUpperCase()} score ${score.toFixed(2)} is greater than or equal to threshold ${threshold}`
    : `${baseType.toUpperCase()} score ${score.toFixed(2)} is less than threshold ${threshold}`,
  assertion,
};
```

### Problem

When `inverse: true`:

- **pass condition** becomes `score < threshold` (correct)
- But the **pass reason** still says "is greater than or equal to threshold" - which is factually wrong (the score is less than threshold)
- The **fail condition** becomes `score >= threshold`
- But the **fail reason** says "is less than threshold" - which is factually wrong (the score is >= threshold)

### Example

If ROUGE-N score = `0.3` and threshold = `0.8` with `inverse: true`:
- `pass = true` (correct, 0.3 < 0.8)
- `reason = "ROUGE-N score 0.30 is greater than or equal to threshold 0.8"` (WRONG, 0.3 is NOT >= 0.8)

### Comparison with bleu.ts and gleu.ts

Both `bleu.ts` and `gleu.ts` handle the `inverse` flag properly in their reason messages using `${inverse ? 'greater' : 'less'}`. The `rouge.ts` file is inconsistent and incorrect.

### Expected Fix

```typescript
return {
  pass,
  score: inverse ? 1 - score : score,
  reason: pass
    ? 'Assertion passed'
    : `${baseType.toUpperCase()} score ${score.toFixed(2)} is ${inverse ? 'greater than or equal to' : 'less than'} threshold ${threshold}`,
  assertion,
};
```

kuishou68:
## Summary

Fixes #8681

The `rouge` assertion in `src/assertions/rouge.ts` was producing incorrect/misleading reason messages when `inverse: true` was set.

## Problem

When `inverse: true`:
- **Pass reason** always said: *"ROUGE-N score X.XX is greater than or equal to threshold Y"* — even when the score was LESS than the threshold (which is exactly what triggers `pass=true` in inverse mode)
- **Fail reason** always said: *"ROUGE-N score X.XX is less than threshold Y"* — even when the score was GREATER than the threshold

Both messages were factually wrong and opposite to the actual situation.

## Fix

Aligned the reason messages with `bleu.ts` and `gleu.ts`, which already handle `inverse` correctly:

```typescript
// Before (wrong)
reason: pass
  ? `${baseType.toUpperCase()} score ${score.toFixed(2)} is greater than or equal to threshold ${threshold}`
  : `${baseType.toUpperCase()} score ${score.toFixed(2)} is less than threshold ${threshold}`,

// After (correct)
reason: pass
  ? "Assertion passed"
  : `${baseType.toUpperCase()} score ${score.toFixed(2)} is ${inverse ? "greater than or equal to" : "less than"} threshold ${threshold}`,
```

## Test

With `inverse: true`, threshold `0.8`, and ROUGE-N score `0.3`:
- `pass = true` (0.3 < 0.8 ✅)
- **Before**: reason = `"ROUGE-N score 0.30 is greater than or equal to threshold 0.8"` ❌ (wrong!)
- **After**: reason = `"Assertion passed"` ✅

Signed-off-by: Cocoon-Break <54054995+kuishou68@users.noreply.github.com>


mldangelo-oai:
Thank you!

PR #8686, opened 2026-04-13
fix(providers): close stdin on exec provider to prevent child process hanging

aisensiy:
## Summary

- Close stdin immediately after spawning child process in exec provider (`ScriptCompletionProvider`)
- Node's `execFile` pipes stdin by default but never writes to or closes it — tools that read stdin at startup (e.g. [opencode](https://github.com/sst/opencode)) block forever waiting for input
- Added two tests: verifies `stdin.end()` is called, and handles null stdin gracefully

## Problem

When using the exec provider to wrap CLI tools, some tools (like `opencode run`) attempt to read stdin during initialization. Since `execFile` creates a piped stdin that is never closed, the child process hangs indefinitely:

- **Direct shell execution**: works fine (stdin is /dev/null or TTY)
- **Via promptfoo exec provider**: hangs forever, 0 bytes output

Debug logs show the child process stalls before even completing its initialization sequence.

## Root Cause

`child_process.execFile()` defaults stdio to `['pipe', 'pipe', 'pipe']`. The parent (promptfoo) holds the stdin pipe open but never writes to it or closes it. Any child process that checks/reads stdin at startup will block forever.

Note: `execFile` does not properly support the `stdio` option (unlike `spawn`), so setting `stdio: ['ignore', 'pipe', 'pipe']` does not work. The fix is to explicitly call `child.stdin.end()` after spawning.

## Fix

```ts
const child = execFile(command, scriptArgs, options, callback);
child.stdin?.end();
```

## Test plan

- [x] Existing 11 tests pass
- [x] Added test: verifies `stdin.end()` is called on the child process
- [x] Added test: handles child process with null stdin gracefully
- [x] Verified end-to-end: opencode CLI now completes successfully via exec provider (was hanging indefinitely before)

### DIFF
--- a/src/providers/scriptCompletion.ts
+++ b/src/providers/scriptCompletion.ts
@@ -105,7 +105,7 @@ export class ScriptCompletionProvider implements ApiProvider {
       ]);
       const options = this.options?.config.basePath ? { cwd: this.options.config.basePath } : {};
 
-      execFile(command, scriptArgs, options, async (error, stdout, stderr) => {
+      const child = execFile(command, scriptArgs, options, async (error, stdout, stderr) => {
         if (error) {
           logger.debug(`Error running script ${this.scriptPath}: ${error.message}`);
           reject(error);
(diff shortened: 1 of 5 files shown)


mldangelo-oai:
Thank you @aisensiy! This is a great find.

PR #8716, opened 2026-04-15
fix(webui): pin eval results footer for short tables

mldangelo-oai:
## Summary

- Make the eval results table wrapper participate in the page flex layout so short result sets can push the pagination footer to the viewport bottom.
- Allow the table scroll container to shrink inside flex layout and avoid forced scrollbars when there is no overflow.
- Keep the pagination footer from shrinking when the table region expands.
- Add a focused regression test for the results table wrapper layout.

## Root Cause

`ResultsTable` was written as if its parent were a flex column: the table container had `flexGrow: 1`, and a comment noted that the pagination footer should be pinned to the viewport bottom. In `ResultsView`, however, the immediate wrapper was a plain block (`px-4`), so the table container could not consume the remaining vertical space when there was only one row. That left the pagination footer floating above the bottom of the viewport.

## Screenshots

Before:

Before: pagination footer floating above the viewport bottom

After:

After: pagination footer pinned to the viewport bottom

## Validation

- `npm run test:app -- src/pages/eval/components/ResultsView.test.tsx --run`
- `npm run l && npm run f`
- Manual browser QA on `http://localhost:3000/eval/eval-Dti-2026-04-11T16:22:18` at `1419x768`: footer bottom measured at the viewport bottom and the table container no longer forced scrollbars without overflow.


wholley-oai (approved):
My old nemesis

PR #8737, opened 2026-04-15
fix(eval): handle empty array

Nico385412:
Hello, 

I created this PR because i had issue with custom envFiles not being loaded.
exemple of my promptfooconfig.yaml file:

```yaml
...
commandLineOptions:
    envPath:
        - ../.env.dev
...
```

After few investigations it's because envPath can be an array or string, but the condition does not handle the case of an empty array.
By handling empty array case I can load my custom env file


mldangelo-oai (approved):
Thank you @Nico385412!!!

PR #8820, opened 2026-04-20
chore(app): update to use brand colors

typpo:
- **chore: design for pdf report downloads**
- **chore(app): update to use brand colors**

### Eval setup run options (light)
Before:
Before eval setup run options light

After:
After eval setup run options light

### Eval setup run options (dark)
Before:
Before eval setup run options dark

After:
After eval setup run options dark

### Red team setup
Before:
Before red team setup

After:
After red team setup


faizan-oai (approved):
my kingdom for a screenshot

mldangelo-oai:
@faizan-oai screenshots added to the PR body, now with before/after pairs for:

- Eval setup run options (light)
- Eval setup run options (dark)
- Red team setup

Direct links are in the description above.

faizan-oai (changes requested):
tbh, I don't like the red in the form, it looks too much like an error state and less like a branding choice! Could we try applying accents in places rather than going full red?

src/app/src/pages/eval/components/CustomMetrics.css
```diff
@@ -85,19 +85,19 @@
 }
 
 [data-theme='dark'] .metric-chip:hover {
-  background-color: var(--primary-color, #4299e1);
-  border-color: var(--primary-color, #4299e1);
-  box-shadow: 0 4px 12px -2px rgba(66, 153, 225, 0.25);
+  background-color: hsl(var(--primary));
```
mldangelo-oai, line 88: [P2] Keep dark metric-chip text distinct from its hover fill

Every rendered custom metric has both `metric-chip` and `filterable`. On hover, `.metric-chip.filterable:hover` sets `color: hsl(var(--primary))`; this equally specific dark-mode rule sets the background to that same token without replacing the foreground. The name, value, and icon therefore become primary-on-primary (1:1) and disappear whenever a user hovers a metric chip in dark mode. This is separate from the existing `show-more-toggle` foreground thread. Please set an explicit, contrast-verified foreground for this state (for example `primary-foreground` after fixing that token pair), and add a browser/computed-style test for the dark hover state.

src/app/src/index.css
```diff
@@ -406,8 +461,30 @@ input[type='search']::-webkit-search-decoration {
...
   @page {
-    size: A4;
-    margin: 0.25in;
+    size: A4 portrait;
+    margin: 22mm 20mm 22mm 20mm;
+    @bottom-right {
+      content: counter(page) ' / ' counter(pages);
+      font-family: var(--pf-font-base);
+      font-size: 8.5pt;
+      color: #8a8f97;
```
mldangelo-oai, line 470: [P3] Increase the printed footer contrast

The new page counter is 8.5pt `#8a8f97` on the forced white print background, which is about 3.25:1; normal-size text needs 4.5:1. The counter is informational text, so it does not have a logo/incidental-text exemption. Use a darker print token here (for example `--pf-ink-soft` is about 7.18:1 on white) and cover the footer color with a small token-level contrast assertion.

src/app/src/index.css
```diff
@@ -416,8 +493,140 @@ input[type='search']::-webkit-search-decoration {
...
+    font-size: inherit;
+  }
+
+  /* Tables — red accent under thead, zebra rows */
+  table {
+    width: 100%;
+    border-collapse: collapse;
+    margin: 10pt 0 14pt;
+    font-size: 9.5pt;
+    page-break-inside: avoid;
```
mldangelo-oai, line 605: [P2] Let multi-page tables use the remaining page space

This global rule also reaches the printable eval/red-team DataTables, which deliberately expand all rows for print. In a Chromium A4 check using this exact rule and 120 rows, page 1 contained only the heading and the table started on page 2, producing six pages; without whole-table `break-inside: avoid`, 23 rows used the rest of page 1 and the document finished in five pages. Chromium still has to fragment an oversized table, so `avoid` does not keep it together—it only creates heading-only or mostly blank pages. Allow `table` (and other potentially unbounded blocks such as `pre`) to fragment, keep rows together instead, and repeat `thead`; a Playwright PDF fixture with a multi-page table would make this stable.

mldangelo-oai (commented):
Reviewed exact head `e43f131f7f4538763fa9f70cdfb932ba094ae3d7` against current `origin/main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`). My recommendation is to retire this branch in its current form and carry the useful pieces forward as smaller changes from current main.

The work has not landed elsewhere, but the core full-red-primary direction still has an unresolved requested-changes review, the existing light/dark contrast, hard-coded active-state, Google Fonts, print-color, and hover-foreground threads still apply, and the branch is now 510 commits behind. Only the old `EvaluateTestSuiteCreator` inline is fixed/outdated. I also found two new P2 defects and one P3 accessibility issue, noted inline.

A cleaner path would be: (1) a design-reviewed, accent-only brand-token PR with verified light/dark foreground pairs, and (2) a separately scoped print/PDF PR built on current print infrastructure, with deterministic/self-hosted typography and multi-page PDF evidence. If this branch continues instead, it needs the outstanding design and accessibility feedback addressed first, then a current-main merge, preservation of the newer `CustomMetrics` keyboard/click-target work, fresh visual and PDF coverage, and a full fresh CI run.

Validation: the focused ThemeSelector, FilterChips, CustomMetrics, and Report suites pass (58/58), and `git diff --check` passes. The May 18 exact-head CI is green but stale against current main. A local full type/build run was not authoritative because the borrowed current-main dependency tree does not match this branch (unrelated native-package and upstream type drift). No branch changes were made.

PR #8833, opened 2026-04-21
fix(auth): log file auth refresh lifecycle

jameshiester-oai:
## Summary
- log when file auth acquires the refresh lock and invokes the auth function
- include a plain-language freshness message when file auth returns a token successfully
- add HTTP auth tests covering the new lock and success logging behavior

## Testing
- npm run l
- npm run f
- npx vitest run test/providers/http/auth.test.ts

### DIFF
--- a/src/providers/http.ts
+++ b/src/providers/http.ts
@@ -1017,6 +1017,15 @@ function parseFileAuthReference(filePath: string): { filePath: string; functionN
   return filePath.startsWith('file://') ? parseFileUrl(filePath) : { filePath };
 }
 
+function formatFileAuthFreshness(expiration?: number | null): string {
+  if (expiration == null) {
+    return 'never expires';
+  }
+
+  const freshForSeconds = Math.max(0, Math.ceil((expiration - Date.now()) / 1000));
+  return `expires in ${freshForSeconds} seconds`;
+}
+
 export async function createSessionParser(
   parser: string | Function | undefined,
 ): Promise<(data: SessionParserData) => string> {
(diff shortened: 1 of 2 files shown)


src/providers/http.ts
```diff
@@ -1017,6 +1017,15 @@ function parseFileAuthReference(filePath: string): { filePath: string; functionN
   return filePath.startsWith('file://') ? parseFileUrl(filePath) : { filePath };
 }
 
+function formatFileAuthFreshness(expiration?: number | null): string {
+  if (expiration == null) {
+    return 'never expires';
+  }
+
+  const freshForSeconds = Math.max(0, Math.ceil((expiration - Date.now()) / 1000));
+  return `expires in ${freshForSeconds} seconds`;
```
faizan-oai, line 1026: nit: 1 second vs n seconds for n > 1

src/providers/http.ts
```diff
@@ -1786,15 +1795,20 @@ export class HttpProvider implements ApiProvider {
...
+    onLockAcquired?: () => void,
+  ): Promise<void> {
     while (this.tokenRefreshLock != null) {
       if (await this.waitForInFlightTokenRefresh()) {
         return;
       }
     }
 
-    const refreshLock = { promise: refreshToken() };
+    const refreshLock = { promise: Promise.resolve() as Promise<void> };
```
faizan-oai, line 1808: nit: do you need to cast here?

faizan-oai (approved):

Issue #3, opened 2023-05-07
Allow config proxy.


eightHundreds:
Some countries cannot use Open AI and need proxy support.

typpo:
Hi @eightHundreds,

Would it suffice to be able to override the OpenAI API host via environment variable?  For example, `OPENAI_API_HOST=api.your-openai-proxy.com`

Issue #1, opened 2023-05-06
[Feature Request]More Results based on Same Value


Quorafind:
Hi, good job! It is useful. Is it possible to generate more than one result based on same value? I want to test if a prompt is stable.

typpo:
Thanks for the suggestion, I'll see if I can build in an option to do this.  How many times do you think you'd run it to decide if it's stable?

Right now, the easiest way to accomplish what you want is to include the prompt multiple times, either in the same file:

```
<prompt>
---
<prompt repeated>
---
...
```

Or by passing it multiple times:

```
promptfoo -p prompt.txt prompt.txt
```

Quorafind:
Hi, I will want to test the same prompt more than three times. And your suggestion is useful for me either.

PR #8852, opened 2026-04-22
ci(docker): publish images for release backfills

### LINKED ISSUE #8849: Gitlab Action : Unreleased docker image
# Docker images not building for new releases
*This issue is related to the GitHub CI/CD workflow and Docker image distribution, not to the promptfoo core logic itself.*

**Describe the bug**
There is an issue with the automated Docker build process. Docker images are not being built or pushed to `ghcr.io` for new releases.

While the `.github/workflows/docker.yml` file exists, the image for the latest release (**0.121.6**) is currently unavailable. 

**Observations:**
* The last available image in the registry was built 8 days ago (version `0.121.5`, matching `main`).
* The `docker.yml` workflow does not seem to appear or trigger in the GitHub Actions tab for recent releases.
* This suggests the workflow might not be correctly configured to trigger on new tags/releases.

**Expected behavior**
1.  **Automated Builds:** Every new release/tag should automatically trigger the Docker build workflow and push the corresponding tag to `ghcr.io`.
2.  **Tagging Strategy (Suggestion):**
    * The `main` (or `latest`) Docker tag should correspond to the latest commit on the `main` branch.
    * Specific version tags (e.g., `0.121.6`) should be generated for each GitHub release.

**Context / Environment:**
* **Promptfoo version:** `main` branch, commit `ac3fc5d11e51ccc23459e1354c7039fb23a4f8a8`.
* **Target Registry:** `ghcr.io`

---

*edit* : typo in last release (0.126.6 -> 0.121.6)

mldangelo-oai:
## Summary
- add a Docker workflow release-published fallback for human-published releases/backfills
- add a manual `tag_name` dispatch input so missing release images like `0.121.6` can be rebuilt from the tag
- checkout the selected release tag when the workflow is invoked for a release/backfill, and tag GHCR manifests from the selected release tag

Fixes #8849.

## Validation
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/docker.yml"); puts "yaml ok"'`
- `npx prettier --write .github/workflows/docker.yml`
- `go run github.com/rhysd/actionlint/cmd/actionlint@v1.7.7 -oneline .github/workflows/docker.yml`
- `git diff --check`

## Backfill after merge
Run the Docker workflow manually with `tag_name=0.121.6` from `main` to publish the missing GHCR image for the existing release.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -2,15 +2,16 @@
 # runs the Docker container, performs a health check to ensure the application
 # is running correctly, and then publishes the image to GitHub Container Registry
 # if all checks pass. This action is triggered by the release-please workflow,
-# workflow dispatch, pull requests to main that modify the Dockerfile, and pushes to main.
+# published human-created releases, workflow dispatch, pull requests to main that
+# modify the Dockerfile, and pushes to main.
 
 name: Test and Publish Multi-arch Docker Image
 run-name: |
-  ${{ inputs.tag_name && 'Release' ||
+  ${{ (inputs.tag_name || github.event.release.tag_name) && 'Release' ||
       github.event_name == 'workflow_dispatch' && 'Manual' ||
       github.event_name == 'pull_request' && 'PR' ||
       'Push' }}
-  Docker build for ${{ inputs.tag_name || github.ref_name }}
+  Docker build for ${{ inputs.tag_name || github.event.release.tag_name || github.ref_name }}
   ${{ github.event_name == 'pull_request' && format('(PR #{0})', github.event.number) || '' }}
   by @${{ github.actor }}
 
@@ -22,6 +23,14 @@ on:
         required: true
         type: string
   workflow_dispatch:
+    inputs:
+      tag_name:
+        description: 'Optional release tag to publish (e.g., 0.121.6)'
+        required: false
+        type: string
+  release:
+    types:
+      - published
   pull_request:
     branches:
       - main
@@ -40,12 +49,20 @@ concurrency:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
+  RELEASE_TAG: ${{ inputs.tag_name || github.event.release.tag_name || '' }}
+  SOURCE_REF: ${{ inputs.tag_name || github.event.release.tag_name || github.ref }}
+  BUILD_VERSION: ${{ inputs.tag_name || github.event.release.tag_name || github.ref_name }}
+  BUILD_DATE: ${{ github.event_name == 'pull_request' && github.event.pull_request.updated_at || github.event_name == 'release' && github.event.release.published_at || github.event.head_commit.timestamp || github.event.repository.updated_at }}
 
 permissions:
   contents: read
 
 jobs:
   test:
+    # Release-please-created releases are already handled by release-please.yml
+    # after npm publish. The release event path is for human-published releases
+    # and backfills where release-please did not emit release_created=true.
+    if: github.event_name != 'release' || github.event.release.author.login != 'promptfoobot[bot]'
     runs-on: ubuntu-latest
     timeout-minutes: 60
     permissions:
@@ -64,6 +81,8 @@ jobs:
 
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ env.SOURCE_REF }}
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@ce360397dd3f832beb865e1373c09c0e9f86d70a # v4
@@ -81,8 +100,8 @@ jobs:
           tags: local-test-image:latest
           platforms: linux/amd64
           build-args: |
-            BUILD_VERSION=${{ inputs.tag_name || github.ref_name }}
-            BUILD_DATE=${{ github.event_name == 'pull_request' && github.event.pull_request.updated_at || github.event.head_commit.timestamp || github.event.repository.updated_at }}
+            BUILD_VERSION=${{ env.BUILD_VERSION }}
+            BUILD_DATE=${{ env.BUILD_DATE }}
             GIT_SHA=${{ github.sha }}
 
       - name: Run Docker container for testing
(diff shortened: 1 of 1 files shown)


mldangelo (commented):
# Critical Review — PR #8852

**Scope.** One file: `.github/workflows/docker.yml`, +47/-15, two commits. Fixes #8849 (no GHCR image for 0.121.6). Reviewed at `a9f9eef8c`.

**Verdict.** The design is sound and close to mergeable, but there is **one genuine regression risk** the author missed and **two hardenings** that are cheap and worth doing now.

## Must fix before merge

**1. Backfill silently moves the floating `{{major}}.{{minor}}` tag backward.** `docker.yml:306-307`
```yaml
type=semver,pattern={{version}}
type=semver,pattern={{major}}.{{minor}}
```
These are unconditional. `latest` is correctly gated by `IS_RELEASE_BACKFILL`, but `{{major}}.{{minor}}` isn't. If 0.121.7 has already shipped (so `ghcr.io/…:0.121` → 0.121.7's digest) and you then run the workflow with `tag_name=0.121.6`, the `0.121` tag is re-pointed to 0.121.6. Users pinned to `:0.121` silently downgrade. This is exactly the scenario the PR exists to enable (backfilling 0.121.6), so it needs to be right.

**Fix:** gate `{{major}}.{{minor}}` with `enable=${{ env.IS_RELEASE_BACKFILL != 'true' }}` (cheapest — accepts that backfills skip the moving minor tag), or compare against the latest published tag at runtime (more work).

## Strongly recommended (cheap, prevents future silent failure)

**2. Replace the hardcoded `promptfoobot[bot]` literal with a type check.** `docker.yml:66`
```yaml
if: github.event_name != 'release' || github.event.release.author.login != 'promptfoobot[bot]'
```
If the app is ever renamed, rotated, or if release-please switches to a PAT, this silently turns into a double-publish. `github.event.release.author.type == 'Bot'` is a stable property and covers the same intent — release-please is the only bot publishing releases here. One-line swap:
```yaml
if: github.event_name != 'release' || github.event.release.author.type != 'Bot'
```

**3. Fail loud when `release:published` fires without a resolvable tag.** `docker.yml:52-56, 319-334`
If `github.event.release.tag_name` is ever empty on a release event (odd but possible), `RELEASE_TAG=''`, `IS_RELEASE_BACKFILL=true` (event is `release`), and the metadata tags mostly end up disabled. The `Inspect image` step then falls back to `steps.meta.outputs.version` and can green the run while pushing nothing meaningful. Add a guard step at the top of the build job:
```yaml
- name: Validate release tag
  if: github.event_name == 'release' || (github.event_name == 'workflow_dispatch' && inputs.tag_name != '')
  env:
    RELEASE_TAG: ${{ env.RELEASE_TAG }}
  run: |
    [[ "$RELEASE_TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][A-Za-z0-9.-]+)?$ ]] \
      || { echo "::error::RELEASE_TAG '$RELEASE_TAG' is empty or invalid"; exit 1; }
```
Bonus: this also neutralizes the (pre-existing) `workflow_dispatch` shell-injection vector in the `Inspect image` step.

## Suggestions (nice-to-have, not blockers)

- **`IS_RELEASE_BACKFILL` is misleadingly named.** It's true for *any* `release:published` event, including a legitimate newest-release hand-publish. That path gets semver tags but **not** `latest`, which may surprise a maintainer doing a hotfix. Either rename (e.g. `IS_NON_CALLABLE_RELEASE_TRIGGER` — ugly but honest) or add a note in the file header.
- **Comment on `test` job's `if:` describes intent, not mechanism** (`docker.yml:62-65`). It says the guard triggers "where release-please did not emit release_created=true" but the code actually checks the author login. Rewrite the comment to match the code: "Skip release-event runs authored by the release-please bot — those are already built via workflow_call from release-please.yml."
- **Add a one-liner above `Resolve source commit`** (`docker.yml:88-90, 216-218`) explaining *why* it exists: `github.sha` is the trigger SHA, not the checked-out ref's SHA. Without this, a future reader will see it as redundant and delete it, silently corrupting `GIT_SHA` and `org.opencontainers.image.revision`.
- **`BUILD_DATE` for workflow_dispatch backfill** falls through to `github.event.repository.updated_at` (effectively "now"). Acceptable. If you want precision, a short bash step using `gh release view "$RELEASE_TAG" --json publishedAt` would give you the actual publish date. Non-blocking.

## What is **not** a problem (noise to ignore)

- **`build-docker-and-push-digests`'s explicit `if:` does NOT bypass the skip cascade.** I investigated this because a review agent flagged it as critical. Per the GHA expression docs, an `if:` expression that contains no status-check function (`success()`/`failure()`/`always()`/`cancelled()`) is implicitly ANDed with `success()`. `if: (github.event_name != 'pull_request')` has no such function, so `test` being skipped correctly propagates to the entire downstream chain. The bot guard works.
- **Checkout ref race with release-please tag creation.** In release-please.yml, `docker` is `needs: [publish-npm, release-please]`, so the tag is created before workflow_call fires. On `release:published`, GitHub only dispatches after the tag ref exists. No race.
- **Shell injection via `inputs.tag_name`.** Pre-existing (the prior `IMAGE_TAG=\"\${{ inputs.tag_name || … }}\"` had the same shape). Only workflow_dispatch-authorized users can exploit, and fix #3 above neutralizes it anyway.
- **Env-block duplication / `run-name` repetition / extracting `Resolve source commit` to a composite action.** Reviewed and left alone. GHA's env block can't self-reference, and two 3-line steps don't warrant a composite action. Diff stays minimal.

## Diff-size / scope comment

The second commit (`keep backfill tags scoped`) added `IS_RELEASE_BACKFILL` and the revision label — both good additions — but it pulled in disabling `type=ref,event=branch`, `type=ref,event=pr`, and `type=sha` for release events, which the first commit didn't need. Worth asking whether disabling these three for backfills (vs. just `latest`) is intentional defensive hygiene or scope creep. I'd keep it — it prevents a backfill from stamping `main` or a PR number onto an old release — but it deserves a one-line comment.

---

**Recommended next step.** Apply fix #1 (gate `{{major}}.{{minor}}` on `IS_RELEASE_BACKFILL`). Then decide whether #2 and #3 go in this PR or a follow-up. Rest are nits.

PR #8863, opened 2026-04-23
fix(redteam): support remote input materialization

jameshiester-oai:
## Summary

- add a shared remote materialization helper so remote Promptfoo servers can return materialized multi-input vars and their metadata
- thread server-side materialization results through remote plugin generation and multi-turn redteam providers instead of re-materializing locally
- add coverage for remote materialization passthrough, upgrade-guard failures, and GOAT/plugin integration paths

## Test plan

- `cd promptfoo && npm test -- test/providers/promptfoo.test.ts test/redteam/plugins/index.test.ts test/redteam/providers/goat.test.ts`


zcrab-oai (approved):

src/redteam/providers/iterativeTree.ts
```diff
@@ -667,19 +724,34 @@ async function runRedteamConversation({
...
-            // If parsing fails, it's plain text - keep original vars
+          if (shouldGenerateRemote()) {
+            try {
+              const parsed = JSON.parse(newInjectVar);
+              Object.assign(
+                updatedVars,
+                buildRemoteMaterializedInputVariables(
+                  { inputMaterialization, materializationHandled, materializedVars },
+                  parsed,
+                ).vars,
```
mldangelo-oai, line 738: [P1] The remote path passes raw `JSON.parse(newInjectVar)` into `buildRemoteMaterializedInputVariables()` and then applies the returned vars to `updatedVars`. The local path filters parsed JSON through declared inputs first, but this path can allow undeclared server-generated keys to become template vars and override unrelated variables during target rendering. Please filter fallback vars to the configured `inputs` keys before applying them.

src/redteam/providers/hydra/index.ts
```diff
@@ -414,16 +418,32 @@ export class HydraProvider implements ApiProvider {
       }
 
       // Extract input vars from the processed message for multi-input mode
+      if (this.config.inputs && shouldGenerateRemote()) {
```
mldangelo-oai, line 424: [P1] In a real multi-input Hydra eval, the hosted agent rejected the structured input schema before this materialization path could run (`inputs.document` expected string, got object). The provider then logged the agent error and continued; with `maxTurns=1`, the loop exited with zero target probes, empty output, no error, and the policy grader marked the empty response as passing. Please make the remote input schema compatible and return an error when the Hydra agent fails or when no attack turns execute, instead of treating that as a successful run.

src/redteam/remoteMaterialization.ts
```diff
@@ -0,0 +1,92 @@
...
+  response: RemoteMaterializationResponse,
+  fallbackVars: Record<string, string>,
+): MaterializedInputVariablesResult {
+  return {
+    ...(response.inputMaterialization
+      ? {
+          metadata: response.inputMaterialization as Record<string, MaterializedInputMetadata>,
+        }
+      : {}),
+    vars: response.materializedVars ?? fallbackVars,
```
mldangelo-oai: [P2] When `response.materializedVars` is present, this replaces `fallbackVars` entirely. If the remote server returns only the inputs it actually materialized, such as a DOCX input but not a text input, the omitted text input disappears before rendering. Please merge fallback first and overlay `response.materializedVars`, and add a regression with mixed docx plus text inputs.

src/redteam/providers/goat.ts
```diff
@@ -413,14 +417,19 @@ export default class GoatProvider implements ApiProvider {
...
-                pluginId: 'goat',
-              })
-            : undefined;
+        let materializedInputVars:
+          | Awaited<ReturnType<typeof materializeInputVariablesWithMetadata>>
+          | undefined;
+        if (currentInputVars && this.config.inputs) {
+          materializedInputVars = {
+            ...(data.inputMaterialization ? { metadata: data.inputMaterialization } : {}),
+            vars: data.materializedVars ?? currentInputVars,
```
mldangelo-oai: [P2] GOAT inlines the same replacement behavior, so fixing `buildRemoteMaterializedInputVariables()` will not cover this path. A partial `data.materializedVars` response drops any `currentInputVars` not echoed by the server. Please reuse the shared helper after fixing it, or merge `currentInputVars` with `data.materializedVars` here.

src/redteam/providers/iterativeTree.ts
```diff
@@ -1060,17 +1139,33 @@ async function runRedteamConversation({
...
-          provider: redteamProvider,
-          purpose: test?.metadata?.purpose as string | undefined,
-        },
-      );
-      Object.assign(finalUpdatedVars, materializedVars);
+      if (shouldGenerateRemote()) {
+        const remoteBestNodeMaterialization = {
+          inputMaterialization: bestNode.inputMaterialization,
+          materializationHandled: Boolean(
+            bestNode.materializedVars || bestNode.inputMaterialization,
```
mldangelo-oai: [P2] The final hydration reconstructs remote materialization state from the presence of `materializedVars` or `inputMaterialization`, but earlier turns accept `materializationHandled: true` even when no payload is returned. In that case per-turn rendering can still use parsed vars, while the final render skips hydration and can fall back to stale original input vars. Please carry the explicit handled flag through the tree node or hydrate from the final prompt whenever remote materialization was handled.

src/redteam/providers/iterativeMeta.ts
```diff
@@ -340,16 +354,29 @@ export async function runMetaAgentRedteam({
...
     const currentInputVars = extractInputVarsFromPrompt(attackPrompt, inputs);
-    const materializedInputVars =
-      currentInputVars && inputs
-        ? await materializeInputVariablesWithMetadata(currentInputVars, inputs, {
+    let materializedInputVars:
+      | Awaited<ReturnType<typeof materializeInputVariablesWithMetadata>>
+      | undefined;
+    if (currentInputVars && inputs) {
+      if (shouldGenerateRemote()) {
+        materializedInputVars = buildRemoteMaterializedInputVariables(agentResp, currentInputVars);
```
mldangelo-oai: [P2] The real hosted Meta response came back as a plain `document: ...\nquestion: ...` string rather than JSON or `<Prompt>{...}</Prompt>`. Because this path only builds remote materialized vars after `extractInputVarsFromPrompt()` succeeds, the run kept the original materialized DOCX while putting the generated multi-input attack into the text question. That means the eval can pass while not actually testing the remote agent's intended `document` input. Please enforce the expected remote multi-input response format or fail closed when per-input vars/materialized vars are missing.

mldangelo-oai (changes requested):
Thanks for pushing this through. The shared helper and threading are the right direction, and I appreciate the coverage that is already here. I ran this against a real OpenAI provider with multi-input DOCX + text cases across basic/base64/crescendo/tree/meta/hydra/GOAT paths. A few remote-materialization edge cases still look risky enough that I think we should tighten them before merging.

The biggest concerns are: Hydra multi-input can currently report a passing scan after zero target probes when the remote agent rejects the structured input schema; and some remote fallback/materialized-var paths can drop or leak vars. GOAT is also blocked against the current hosted endpoint by the new upgrade guard, so this likely needs either a server deployment ordering guarantee or a compatibility path before client release.

What I verified:
- basic/base64/crescendo/jailbreak:tree passed locally with real target calls
- jailbreak:meta ran against the hosted remote, but the hosted response was not JSON-shaped and the run kept stale per-input vars
- jailbreak:hydra multi-input reported pass with zero target probes after the hosted agent rejected `inputs.document` as an object
- single-input Hydra control ran normally, so the Hydra failure is specific to this new multi-input path
- GOAT multi-input failed on the hosted endpoint with the remote materialization upgrade guard

This is close; I think the remaining fixes are mostly about fail-closed behavior and preserving/validating the per-input vars consistently.

mldangelo-oai (approved):
this code is beautiful

PR #8876, opened 2026-04-24
fix(eval): harden OpenAI media blob exports

mldangelo-oai:
## Summary

Fixes the OpenAI image/audio blob-storage regressions found during the media audit:

- externalizes realtime `metadata.audio.data` and points mirrored metadata audio at the same blob as top-level `response.audio`
- dedupes repeated exact media payloads within a single response, including GPT Image `output` plus `images[0].data`
- redacts sensitive OpenAI/edge response headers before persistence/export across `response`, top-level result `metadata`, and nested grading metadata
- keeps `persist: false` result fidelity intact by applying header redaction only at the persistence boundary
- prevents `file://` audio vars in the results table from being rendered as bogus base64 data URLs

The portable JSON export issue is intentionally not solved here. JSON exports still contain `promptfoo://blob/...` references; making them media-portable needs an explicit export format decision, such as embedded bytes or a sidecar manifest, so this patch keeps scope to data leakage, rendering correctness, and duplicate storage/upload behavior.

## Validation

- `npx vitest run test/blobs/extractor.test.ts test/models/evalResult.test.ts --sequence.shuffle=false`
- `npm run test:app -- src/utils/media.test.ts src/pages/eval/components/ResultsTable.test.tsx --run`
- `npm run tsc`
- `npm run f && npm run l` (passes with the existing `extractAndStoreBinaryData` complexity warning)

Live eval QA with `--no-cache` and `/Users/mdangelo/code/promptfoo/.env`:

- Chat audio: `eval-T69-2026-04-22T15:27:37`
- Realtime audio: `eval-1p5-2026-04-22T15:28:03`
- GPT Image 2: `eval-7ud-2026-04-22T15:28:16`

Export inspection confirmed no raw audio/image data URLs, no `b64_json`, and no raw `set-cookie`, project ids, request ids, `cf-ray`, or `x-ratelimit-*` values. SQLite blob-reference checks showed one final blob reference for each live media artifact, and `/api/blobs/:hash` returned `200 OK` with the expected `audio/wav` or `image/png` content type.


ianw-oai:
One regression to fix before this is safe: the new persistence redaction walks any nested `http` object in the whole response, so a legitimate model payload like `response.output = { http: { headers: { "x-request-id": "..." }}}` will be rewritten on save even though it is user-visible output, not transport metadata. The redaction needs to stay scoped to persistence metadata fields rather than arbitrary output structure.

PR #8877, opened 2026-04-24
docs(site): add trace-based agent eval guide

mldangelo-oai:
Add a guide for evaluating traced agent workflows with span counts, latency and error budgets, tool trajectories, and model-graded goals. The runnable JavaScript examples show loopback receiver setup, trace correlation, source redaction, and exporting spans after the root operation ends.

The shared simulated provider now answers both configured topics correctly. The base config checks that each output names its requested topic, and the tracing documentation retains the parent branch's tool-set example.

## Validation

- Reproduced the machine-learning row receiving a quantum-computing answer before the fix.
- Built CLI: both topic checks pass; the base example retains its two documented duration failures. The offline trajectory example passes all five assertions, and the guide passes all 11 assertions using a local fixture judge. No execution errors or paid model calls.
- Verified generic and trace-specific OTLP endpoints through the actual receiver and exported traces.
- Full core/app and docs builds, TypeScript, architecture, Knip, and lint/format pass.
- Includes #9030 at `c4ba986c395d174e275373c7f5ff3bd79b6ba94e`, which contains #9029 and main through `c2b1a2ff26bfa267f0d09d3372ca4e900692ddce`.


ianw-oai (approved):
Reviewed the new guide/example and spot-checked the trace assertions path. The deterministic trace-based checks look sound, and the provider flush timing change is the right direction.

mldangelo:
## Critical doc + end-to-end review

Audited the guide against the implementation, ran the example end-to-end with a live OpenAI judge, and inspected the screenshots. Pushed fixes for the user-facing correctness issues in c4770e3cc — full breakdown below.

### ✅ End-to-end verification

```
npm run local -- eval -c examples/integration-opentelemetry/javascript/promptfooconfig.trace-guide.yaml \
  --no-cache --env-file .env -o /tmp/out.json
```

All 12/12 assertions pass on a real run. Live output also confirms recent feature additions:

| # | Assertion | Pass | Reason |
|---|---|---|---|
| 4 | `trace-span-duration p95<350ms` | ✅ | `289.88ms, method=nearest (warning: small sample N=3)` |
| 11 | `trajectory:step-count reasoning 4-4` | ✅ | Matched 4 reasoning steps |
| 12 | `trajectory:goal-success` | ✅ | Judge score 0.95 |

### 🔴 Correctness issues fixed in c4770e3cc

1. **`--pass-rate 0.95` is not a real CLI flag.** The CI workflow snippet and the prose recommendation now use `PROMPTFOO_PASS_RATE_THRESHOLD: '95'` (percentage, not fraction) — verified against `src/commands/eval.ts:902`.
2. **Broken `/docs/red-team/agents` links** (×2) repointed to `/docs/red-team/llm-agents`.
3. **Judge model identifier inconsistency.** The trace guide used `openai:chat:gpt-5.4-mini` but the Codex SDK doc edit added `openai:gpt-5-mini`. Now consistent on the canonical form from `examples/AGENTS.md`.
4. **`tokens-used` description elided the double-counting footgun** (per the open Codex P1 finding on `src/assertions/tokensUsed.ts:33-37`). The doc now describes the actual summing behavior across `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, `gen_ai.usage.total_tokens`, and `tokens.used`, and warns readers not to mix component and aggregate keys on the same span.
5. **Inverse `not-trace-*` no-match bug** (per the open Codex P2 findings on `traceErrorSpans.ts:115` and `traceSpanDuration.ts:84`). The doc now warns that `not-trace-error-spans` and `not-trace-span-duration` don't currently flip when zero spans match, and tells readers to pair them with a `trace-span-count` presence check. Also resolves a self-contradiction with the "Current limits" bullet.
6. **`gen_ai.conversation.id` recipe oversold** — implied Promptfoo groups rows by conversation, which the codebase does not do. Now describes the attribute as a stitching tag and offers a `jq` workaround.
7. **Forwarder claims described an unimplemented feature** (`tracing.forwarding` exists in the schema at `src/types/index.ts:1095` but no code path implements it). Both PII-section bullets rewritten to talk accurately about external trace backends your provider exports to.
8. **`cancel-in-progress: true`** in the recommended CI workflow flipped to `false`. Eval suites should not cancel in-flight runs — that drops artefacts and breaks pass-rate trends.
9. **Tautological example flagged.** The simulated provider always emits the same `search_corpus` query, so the `trajectory:tool-args-match` assertion in the example is illustrative, not a real paraphrase test. Now called out with a pointer to the Anti-patterns section.
10. **Aspirational `--baseline` roadmap claim** dropped — promising unshipped features in published docs is a footgun.
11. **Regex-based artefact-comparison snippet** annotated as brittle (it parses `(\d+)ms` out of assertion `reason` text — undocumented contract). Doc now points at structured fields.
12. **Span 101 ASCII tree** now includes `context_augmentation` so it matches what the example provider actually emits.
13. **`ss -lnt` pre-flight check** annotated as Linux-only with the macOS alternative.

### 🟡 Style / SEO fixed in c4770e3cc

- **Title shortened** from ~70 chars to ~56 chars: "Trace-Based Agent Evals: Tool Calls, Trajectories, and CI" (drops "Evaluation" per `site/AGENTS.md` "use 'eval' not 'evaluation'").
- **Description trimmed** to ~140 chars (was over the 160-char limit).
- **`slug: /guides/trace-based-agent-evals`** set explicitly.
- **All H2/H3 headings** normalized to sentence case (anchors unchanged); 4 inline link texts updated to match.
- **84-word priority-order paragraph** converted to a numbered list.
- **"Trajectory" defined** in Span 101; new beginner pointer to "Run the example".
- **"Intricate"** embellishment removed from `provider-simple-traced.js` header.

### ⚠️ Stale screenshot — please recapture before merge

`site/static/img/docs/tracing/trace-guide-assertions-raw.png` shows the 95th-percentile reason as `279.067138671875ms`. The current code emits `XXX.XXms, method=nearest (warning: small sample N=3)` — verified in the live eval output. The doc now describes the new format but the screenshot doesn't show it. Recapture with the latest build so readers see the current `.toFixed(2)` formatting, the `method=` annotation, and the small-sample warning.

The other three screenshots are accurate to the current implementation:
- `trace-guide-results-overview.png` — "Asserts: 12/12 passed" matches.
- `trace-guide-assertions-trajectory.png` — judge score 0.98 (live runs vary 0.95–0.98; within nondeterminism window).
- `trace-guide-timeline.png` — span tree matches what the example actually emits (root `rag_agent_workflow` is scrolled off the top, but alt text covers it).

### 🟢 Outstanding code-level items (not in this doc PR's scope)

These are still open from the earlier `chatgpt-codex-connector` review and the doc now warns users about each:

- **`src/assertions/tokensUsed.ts:33-37`** — overlapping token attribute keys are double-counted (Codex P1, still open). Fix: prefer the aggregate when present, otherwise sum components.
- **`src/assertions/traceSpanDuration.ts:84`** and **`traceErrorSpans.ts:115`** — early `pass: true` returns bypass inverse semantics on no-match (Codex P2, still open). Fix: route through inverse handling before returning.
- **`src/tracing/evaluatorTracing.ts:81-84`** — `commandToolNames` override is process-global and never reset across evals (Codex P2, still open). Fix: snapshot/restore around each eval, or always call `setCommandToolNames` (with the empty array as a reset).

### Verdict

Docs are now accurate to the implementation and the runnable example genuinely produces 12/12 passing assertions. Two things gate merge from a docs perspective:

1. **Recapture `trace-guide-assertions-raw.png`** to show the current percentile-reason format.
2. Decide whether the three open code-level findings ship in this PR or follow up in a separate one. The doc warnings are sufficient if you go separate.

mldangelo-oai:
Addressed the active review feedback in `743c051ea` and `f5fe0b026`:

- aligned the trace-guide provider and docs on `http://127.0.0.1:4318/v1/traces`, matching the IPv4 receiver host in `promptfooconfig.trace-guide.yaml`;
- corrected the guide to say `trace-span-duration` percentiles are computed per row, not across an eval run;
- corrected `tokens-used` `source: auto` guidance for the default `pattern: '*'` provider-response fallback versus custom trace patterns.

Verification:

- `npm run local -- eval -c examples/integration-opentelemetry/javascript/promptfooconfig.trace-guide.yaml --no-cache --no-share --env-file .env -o /tmp/promptfoo-pr8877-trace-guide.json` passed 1/1 rows and all 12 assertion components; logs showed exporter and receiver both using `127.0.0.1:4318`.
- `npm run l && npm run f` passed with the existing warning-level complexity findings in `src/assertions/index.ts` and `src/models/eval.ts`.
- `npm run tsc -- --pretty false`, `npm run build`, and `cd site && SKIP_OG_GENERATION=true npm run build` passed.

Resolved the three current active review threads after pushing.

mldangelo-oai:
Refreshed this child branch onto the just-updated #9030 parent/current `main`.

Verification rerun after the refresh:
- `npm run local -- eval -c examples/integration-opentelemetry/javascript/promptfooconfig.trace-guide.yaml --no-cache --no-share -o /tmp/promptfoo-pr8877-refresh-trace-guide.json`
  - passed 1/1 row with 12 assertion components; exported JSON included the expected trace span, trajectory, and goal-success reasons
- `npm run l && npm run f` (existing warning-level complexity notices in `src/assertions/index.ts` and `src/models/eval.ts`)

CI is running again on the pushed head.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+          - password
+          - authorization
+          - secret
+          - api_key
+  ```
+
+  Spans arrive with the named keys replaced by `[REDACTED]` so the original values never reach disk. Use this as defense in depth — the safer move is to redact at the source in your provider.
+
+- If your provider exports spans to an external trace backend in addition to Promptfoo, redact at the source. Promptfoo's receiver-side `redactAttributes` list applies only to what Promptfoo stores; it does not intercept spans your provider sends to a separate collector.
+- Avoid recording full prompts and full responses on spans by default. The OTel GenAI semantic conventions treat `gen_ai.prompt` and `gen_ai.completion` as opt-in for the same reason.
```
jameshiester-oai, line 914: [P2] Keep the tutorial's own prompt out of exported traces. The linked provider still sets `prompt.text: prompt`, and `promptfooconfig.trace-guide.yaml` configures no `redactAttributes`. In the no-cache run, the exported JSON contained the full prompt and raw `tool.arguments` under `.traces[0].spans`, even though line 997 says exports do not contain every raw span. The workflow above uploads that file, and `--share` sends traces too, so a copied real-world example can move PII or secrets into SQLite, the UI, an artifact, and a remote share. Remove `prompt.text` (retain length/safe metadata), or redact it at source/receiver, and explicitly document that exports/shares include the trace array and need access/retention controls.

site/docs/providers/openai-codex-sdk.md
```diff
@@ -425,7 +493,17 @@ providers:
       enable_streaming: true
 ```
 
-Deep tracing injects OpenTelemetry environment variables (`OTEL_EXPORTER_OTLP_ENDPOINT`, `TRACEPARENT`, etc.) into the Codex CLI process. Promptfoo uses a fresh SDK client/thread per call in this mode so child spans link to the correct parent request span.
+Deep tracing injects OpenTelemetry environment variables into the Codex CLI process:
```
jameshiester-oai, line 496: [P2] Preserve Codex's required log-exporter configuration. Injecting standard OTEL environment variables does not by itself enable Codex's native telemetry; Codex exports these records as logs only after `$CODEX_HOME/config.toml` configures `[otel]` with an `otlp-http` exporter targeting the complete `/v1/logs` URL. The current base at `cd99feed3e95055f0e50ad82ca16f4a94b1b4eae` already contains that setup, and this section is the merge conflict. Taking this side leaves users with no native records and troubleshooting steps that adjust ineffective variables. Preserve the current-base config block, then describe `TRACEPARENT`/resource injection as additional linkage rather than a replacement. OpenAI's current exporter example uses the same config surface: https://openai.com/index/running-codex-safely/

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+The simulated provider always emits the same response body and the same `search_corpus` query so the deterministic `contains` and `trajectory:tool-args-match` assertions stay green across runs. Treat those two assertions as instructive shape, not as proof that the feature handles paraphrased queries — a real model will phrase `quantum computing classical computing` a dozen different ways. See [Anti-patterns](#anti-patterns) for how to assert on causally load-bearing arguments instead of free text.
+:::
+
+Open the eval row, then switch between **Evaluation** and **Traces** in the details panel. The Evaluation tab shows which assertion passed and why:
+
+![Evaluation tab showing raw trace assertions: rag_agent_workflow span count of 1, three retrieve_document_* spans, 95th-percentile retrieval latency under 350 ms, and zero error spans](/img/docs/tracing/trace-guide-assertions-raw.png)
+
+Scroll further and you can see the trajectory assertions proving the behavioral path. Tool-argument assertion reasons are redacted by default, so use the trace API or Traces tab when you need to inspect captured attributes while debugging instrumentation.
```
jameshiester-oai, line 130: [P3] Recapture these screenshots against the current redaction behavior. Both added assertion images show the full query/document index, but the exact-head run produced `Args: [redacted]` with this config, matching the default documented here. The images also show a `0.95` goal score while line 148 says `0.98`, and `trace-guide-assertions-trajectory.png` is not referenced anywhere. Regenerate the evidence, avoid pinning a nondeterministic judge score, and either embed or drop the unused image.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+  'tool.arguments': JSON.stringify(redact(args)),
+});
+```
+
+Once the values are redacted, your `trajectory:tool-args-match` assertions need to assert on identifiers (tenant_id, user_id, document_id), not on the redacted fields. This is the right shape regardless: the identifiers are the causal contract; the free-text fields are evidence.
+
+**Hardening checklist for shared environments.**
+
+- Keep the default `127.0.0.1` OTLP receiver bind for a single-machine eval. Opt into a non-loopback host only when another machine or container must reach the receiver. The receiver has no in-process auth, rate limiting, mTLS, or audit logging — those belong at a higher layer.
+- For any non-loopback deployment (multi-tenant CI runners, sibling containers, public hosts), put a reverse proxy in front of the receiver. Run nginx, Caddy, or your service mesh to terminate TLS, enforce mTLS or OAuth, rate-limit by source, and forward authenticated traffic to `127.0.0.1:4318`. The proxy is the right boundary for those concerns; application-layer auth on this receiver does not replace a deployment boundary.
```
jameshiester-oai, line 895: [P2] A reverse proxy is not a multi-tenant isolation boundary here. All authenticated clients are still forwarded into one receiver/store with no per-tenant authorization, and client-supplied evaluation/test IDs determine where records are persisted. TLS/OAuth controls ingress but does not prevent one runner from writing into another runner's database namespace. Also, a sibling container's `127.0.0.1` is that container itself unless it shares the receiver's network namespace. Recommend one receiver plus `PROMPTFOO_CONFIG_DIR` per job/trust boundary, or a private service address with firewall/network-policy isolation; qualify loopback forwarding as same-host/shared-namespace only.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+    return {
+      pass: avg >= 0.7,
+      score: avg,
+      reason: `avg retrieval score ${avg.toFixed(2)} (threshold 0.70)`,
+    };
+
+# Pattern 2 — judge whether the answer is faithful to the cited documents.
+- type: llm-rubric
+  provider: openai:chat:gpt-5.4-mini
+  value: The answer cites at least two of the retrieved documents and does not introduce facts that were not in the retrieval context.
```
jameshiester-oai, line 610: [P2] Pass the retrieved context to this faithfulness grader. `llm-rubric` receives the final output, rendered rubric, and test vars; it does not receive `context.trace` or the documents that exist only in these spans. The judge therefore cannot tell whether a fact was absent from the retrieval context and can pass on model world knowledge alone. Use `context-faithfulness` with a `context` variable/`contextTransform`, or explicitly extract and inject the retrieved documents into the grading input.

jameshiester-oai (commented):
## Review summary

**Reviewed:** PR-recorded base `10a08bd12af379f5c7e03b2878f5ac1350154e90` → head `af75854ed0c793364a782e47f547290d6f1f7c0d`

**Scope:** Fresh full review. There was no authenticated completion marker for this head. I also examined the current remote base ref `mdangelo/codex/trace-assertion-surface` at `cd99feed3e95055f0e50ad82ca16f4a94b1b4eae`; this PR currently conflicts with that ref, so the two current-base interaction findings below must be preserved during conflict resolution.

**Verdict:** The guide is broad and useful in intent, but 11 P2 documentation/behavior defects make several privacy, tracing, and CI recipes unsafe or non-runnable as written. One P3 screenshot issue also makes the captured UI evidence inconsistent with the current head.

**CI:** 33/33 reported checks are successful for the exact head. GitHub reports the PR as `CONFLICTING` / `DIRTY` against its current base ref.

### Verified findings

**P2**

- The showcased provider records the full rendered prompt, and exported/shared trace payloads carry it even though the guide says otherwise. The documented artifact workflow can therefore upload prompt/argument data without the stated warning or redaction.
- The Codex deep-tracing section omits the required `$CODEX_HOME/config.toml` log exporter and points troubleshooting at environment variables that do not enable Codex's native telemetry. The current base already contains the corrected `/v1/logs` setup.
- The shared-environment proxy recipe authenticates ingress but does not isolate tenants sharing the receiver/store; its sibling-container loopback target is also unreachable unless the proxy shares the receiver's network namespace.
- The RAG faithfulness `llm-rubric` never receives the retrieved documents, so it cannot determine whether claims are grounded in that context.
- `trace-error-spans` ignores the stable `http.response.status_code` semantic-convention key, allowing a standard HTTP 500 span to pass the advertised zero-error gate.
- The bind-failure text says every trace assertion throws, but empty-trace duration/error-budget assertions pass by default and can leave CI green.
- The token-budget explanation is already stale against the current base, and the sample provider puts 750 reasoning tokens under unsupported keys; exact-head `source:auto` counted 820 trace tokens while the response reported 1,570.
- The assertion matrix advertises arbitrary `min`/`max` bounds for `not-trajectory:tool-used`, but runtime validation allows only no bounds or `max: 0`.
- A JavaScript assertion cannot compare the current row's aggregate/component scores because it runs before aggregation and receives neither value.
- The Actions baseline step invokes `gh` without the required `GH_TOKEN` setup.
- The latency script computes a percentile across already-aggregated per-row statistics, not p95 across the underlying spans, and the suggested component `score` fallback is only pass/fail.

**P3**

- The assertion screenshots show raw tool arguments and a `0.95` judge score, while the exact-head run redacts arguments and the prose says `0.98`; one of the two added assertion images is not referenced anywhere.

### Existing feedback

All 25 existing review threads are resolved. I rechecked their current locations and did not duplicate the earlier duration-formatting, regex, or screenshot recapture comments.

- Treat the exported trace array as the single source of truth for cross-run latency analysis instead of parsing assertion prose.
- Remove the unused screenshot or embed it after recapturing; avoid pinning a nondeterministic judge score in prose.
- Add executable docs tests for the inverse assertion matrix, Actions authentication snippet, and the token-accounting example so these contracts cannot drift from runtime validation.

### Validation performed

- `npx vitest run test/assertions/tokensUsed.test.ts test/assertions/traceErrorSpans.test.ts test/assertions/traceSpanCount.test.ts test/assertions/traceSpanDuration.test.ts test/assertions/trajectory.test.ts test/assertions/trajectoryGoalSuccess.test.ts test/tracing/evaluatorTracing.test.ts test/util/sanitizer.test.ts test/util/traceAssertionConfig.test.ts test/providers/openai-codex-sdk.test.ts --sequence.shuffle=false` — exit 0; 10 files, 636 tests passed.
- `npm run tsc` — exit 0.
- `SKIP_OG_GENERATION=true npm --prefix site run build` — exit 0; Docusaurus generated the static site. The optional stats fetch timed out and used its fallback data.
- `OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318 PROMPTFOO_CONFIG_DIR=/private/tmp/promptfoo-pr8877 npm run local -- eval -c examples/integration-opentelemetry/javascript/promptfooconfig.trajectory.yaml --no-cache --no-share -o /private/tmp/pr8877-trajectory.json` — exit 0; 1/1 row passed, 0 failures/errors, 12 spans. The exported component reason contained `Args: [redacted]`, while the same JSON's trace array contained the full prompt and raw tool arguments.
- Direct exact-head handler probes reproduced the stable HTTP-500 false negative, empty-trace no-op passes, and 820-vs-1,570 token undercount.
- `git diff --check 10a08bd12af379f5c7e03b2878f5ac1350154e90...af75854ed0c793364a782e47f547290d6f1f7c0d` — exit 0.

### Limitations

- I did not run the model-graded `trajectory:goal-success` call because it requires external model credentials; the credential-free deterministic trajectory workflow was exercised instead.
- I did not start a local docs server or browser session. The production site build passed, and all four screenshots were inspected at original resolution.
- `/pr-review-toolkit:review-pr all parallel` is unavailable in this environment. I used the six requested manual lenses plus three independent read-only specialist passes and validated their candidates against this exact head.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+Use `trace-error-spans` as a default health gate for traced workflows.
+
+```yaml
+- type: trace-error-spans
+  value:
+    pattern: '*'
+    max_count: 0
+```
+
+You can also use `max_percentage` for noisy systems, or a number shorthand such as `value: 0`. Promptfoo treats OTEL error status, HTTP status codes, error-like attributes, and error-like status messages as error signals.
```
jameshiester-oai, line 455: [P2] Support the stable HTTP status attribute before promising this gate. `isErrorSpan` checks deprecated `http.status_code` but not stable `http.response.status_code`. A span with status UNSET plus `http.response.status_code: 500` returned `pass: true` for `max_count: 0`; changing only the key to `http.status_code` failed as expected. Standard OTLP HTTP failures can therefore bypass this health check. Add the stable key (and a regression test), or narrow the claim. OpenTelemetry marks the former stable and the latter deprecated: https://opentelemetry.io/docs/specs/semconv/registry/attributes/http/

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+            -o eval-output.json
+      - name: Upload eval output
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: eval-output-${{ github.ref_name == 'main' && 'main' || github.run_id }}
+          path: eval-output.json
+```
+
+The `concurrency` block prevents two CI runs on the same branch from racing for OTLP port 4318. For matrix builds on the same runner, set a unique port per job by reading `tracing.otlp.http.port` from a job-level env var. The pre-flight `ss -lnt` check fails fast if the port is occupied — without it, every trace assertion silently throws "No trace data available" and the row fails for the wrong reason.
```
jameshiester-oai, line 753: [P2] Receiver bind failure does not make every trace assertion throw. With the default `failOnReceiverStartFailure: false`, trace context can still reach handlers with `spans: []`; direct probes showed both `trace-span-duration` and `trace-error-spans` passing on that empty trace, consistent with the no-match behavior documented at line 458. This can leave CI green rather than fail the row. Rewrite this outcome and make the sample set `failOnReceiverStartFailure: true` plus a required-span presence check.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+  value:
+    max: 10000 # fail the row if more than 10k tokens were spent
+    pattern: '*' # optional; defaults to all spans
+    source: auto # auto (default) | trace | response
+```
+
+`source: auto` uses trace token attributes when matched spans include them. If the default `pattern: '*'` sees spans but none of them carry token-usage attributes, it falls back to `providerResponse.tokenUsage`; with a custom `pattern`, it stays on the trace path so the filter is not silently ignored. For each matched span, Promptfoo uses the largest available aggregate total or input/output component-family total so incomplete or inconsistent attributes do not undercount the budget. When matched parents and descendants both carry totals, Promptfoo uses the larger of the parent aggregate and the covered descendant total for each token-bearing subtree. This avoids double-counting repeated hierarchy totals without discarding a broader parent aggregate. `source: trace` forces the trace path; `source: response` forces the provider-response path. The inverse `not-tokens-used` is occasionally useful when proving an agent _did_ spend tokens (e.g. a smoke test that detects a stubbed provider).
```
jameshiester-oai, line 807: [P2] Align this section with the live base and complete the example telemetry. Base `cd99feed3e95055f0e50ad82ca16f4a94b1b4eae` changed unfiltered `auto` to choose the larger trace/response total, makes a custom pattern without recognized usage fail, and sums matched operation spans rather than deduplicating by hierarchy. This new guide carries the superseded algorithm. On the exact head, the example also reported 1,570 response tokens but only 820 recognized trace tokens because 750 reasoning tokens use unsupported `step.tokens`/`reasoning.total_tokens` and the root uses unsupported `workflow.total_tokens`. Preserve the current-base semantics during conflict resolution and either emit supported `gen_ai.usage.*` attributes or use `source: response` in this example.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+| `trace-span-count`               | Raw span names and count bounds          | `{ pattern, min?, max? }`                                | Retrieval, guardrail, or root workflow span never ran      |
+| `trace-span-duration`            | Raw span duration threshold              | `{ pattern?, max, percentile? }`                         | Slow retrieval, reranking, sandbox, or tool span           |
+| `trace-error-spans`              | Raw span error status/attributes         | `{ pattern?, max_count?, max_percentage? }` or a number  | Internal tool, HTTP, sandbox, or provider error            |
+| `trajectory:tool-used`           | Required normalized tool steps           | string, string array, or `{ name/pattern, min?, max? }`  | Agent skipped the required tool                            |
+| `not-trajectory:tool-used`       | Forbidden normalized tool steps          | string, string array, or `{ name/pattern, min?, max? }`  | Agent called an unsafe or irreversible tool                |
```
jameshiester-oai, line 360: [P2] Do not advertise arbitrary count bounds for `not-trajectory:tool-used`. Runtime validation rejects any `min` and any `max` other than `0` with `not-trajectory:tool-used object assertions only support name/pattern with no count bounds, or max: 0`. Copying the matrix's `{ ..., min?, max? }` shape therefore turns valid-looking config into an evaluation error. Split this value shape from the positive assertion or state the inverse constraint explicitly.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+```
+
+For large suites, also keep judges off the deterministic critical path. Tag judge rows with metadata and run them in a separate job; gate merges on deterministic assertions only.
+
+### Detect regressions across runs
+
+Promptfoo does not currently produce a built-in run-vs-run diff. The pragmatic patterns:
+
+- **Suite-level threshold.** Set `PROMPTFOO_PASS_RATE_THRESHOLD=95` (a percentage, not a fraction) so the CLI exits non-zero when row pass-rate drops below the threshold. Combine with `--share` to upload the run for human review when CI fails.
+- **Pinned baseline.** Check a sanitized `expected.json` into the repo and add a `javascript` assertion that compares each row's `score` and per-component scores to the baseline, failing on regression beyond a tolerance.
```
jameshiester-oai, line 784: [P2] A JavaScript assertion cannot compare the current row's aggregate or component scores as described here. It runs while assertions are being evaluated, before row aggregation, and `AssertionValueFunctionContext` exposes prompt/vars/test/config/provider/providerResponse/trace—not the row `score` or `componentResults`. Move this comparison to a post-export CI step, or describe a direct observable the assertion can actually compare.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+  value: { pattern: '*', max: 200 }
+```
+
+### Compare against the last main run
+
+A common CI ask is "did this PR regress p95 latency vs main?" Today the cleanest route is two artefacts and a `jq`/`javascript` step. Run the eval on `main` nightly with the same config, store the artefact, then in PR runs compare. The snippet below regexes the measured max or percentile duration out of the assertion `reason` field as a quick demo — for a durable check, prefer reading structured fields (`componentResults[i].score`, span events) instead of parsing prose:
+
+```bash
+# In PR job, fetch the latest successful main artefact from the workflow above
+MAIN_RUN_ID=$(gh run list --workflow agent-eval.yml --branch main --status success \
```
jameshiester-oai, line 824: [P2] Authenticate `gh` in the Actions step. GitHub-hosted runners include the CLI, but every workflow step using it must set `GH_TOKEN`; this workflow never does, so the baseline download fails before the comparison. Add `env: GH_TOKEN: ${{ github.token }}` and explicit `permissions: actions: read` to the runnable workflow. GitHub documents that requirement here: https://docs.github.com/actions/advanced-guides/using-github-cli-in-workflows

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,1063 @@
...
+gh run download "$MAIN_RUN_ID" -n eval-output-main -D ./baseline
+
+# Then in a JS assertion or post-step:
+node -e '
+  const cur = require("./eval-output.json").results.results;
+  const base = require("./baseline/eval-output.json").results.results;
+  const dur = (rows) => rows.flatMap((r) =>
+    (r.gradingResult?.componentResults ?? [])
+      .filter((c) => c.assertion?.type === "trace-span-duration")
+      .map((c) => Number((c.reason.match(/\((?:max: )?(\d+(?:\.\d+)?)ms(?:, method=[^)]+)?\)/) || [])[1])));
```
jameshiester-oai, line 835: [P2] This computes p95 across already-aggregated per-row statistics, not p95 across the underlying spans. Each row contributes one max/percentile value regardless of how many matching spans it contains, so unequal row sizes can create or hide a regression. The suggested `componentResults[i].score` fallback is only the assertion's pass/fail score, not measured latency. Pool durations from the exported top-level `.traces[].spans[]` timestamps, or explicitly compare per-row statistics instead of labeling this suite p95.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The example no longer records prompt.text, raw search.query, or tool.arguments; the configs add receiver redaction and the guide now explicitly says exports/shares can contain trace attributes.

site/docs/providers/openai-codex-sdk.md
mldangelo-oai: Fixed in 724e2215b by restoring this stacked file to the current parent version, preserving the Codex native telemetry configuration instead of carrying the conflicting replacement.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b by deleting all four stale screenshots and removing nondeterministic screenshot/score claims from the guide.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The shortened shared-environments section now says a proxy secures ingress but does not isolate tenants, qualifies loopback to one process/shared namespace, and recommends one PROMPTFOO_CONFIG_DIR per trust boundary.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The guide now uses context-faithfulness with actual context and calls out contextTransform when documents come from the response.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Resolved by narrowing the guide: 724e2215b no longer promises HTTP-status detection and describes trace-error-spans only as an OTel error-budget check.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. Both example configs set failOnReceiverStartFailure: true, the guide requires a root-span presence check, and it explicitly notes duration/error assertions can pass on no matches.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b by removing the stale token-accounting algorithm and example telemetry claims from the compact guide.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The guide now states inverse tool assertions accept name/pattern with max: 0 and not arbitrary positive bounds.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The guide now moves run comparison to a post-export CI step and explicitly says a JavaScript assertion cannot see final row aggregation.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The compact workflow includes actions: read permission and GH_TOKEN for the gh step.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in 724e2215b. The guide now instructs CI to pool raw .traces[].spans[] durations before calculating p95 and distinguishes per-row statistics.

mldangelo-oai:
@codex review

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,170 @@
...
+
+A trace proves retrieval ran; it does not prove the retrieved context was relevant or faithfully used. Pass the actual retrieved context to a RAG grader:
+
+```yaml
+tests:
+  - vars:
+      context: file://retrieved-context.txt
+    assert:
+      - type: context-faithfulness
+        value: '{{ context }}'
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed at bdd448b39: the context-faithfulness example now supplies query and uses threshold; context remains in vars where the handler reads it.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,170 @@
...
+    env:
+      PROMPTFOO_CONFIG_DIR: ${{ runner.temp }}/promptfoo
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run trace eval
+        run: npx promptfoo eval -c promptfooconfig.yaml --no-cache -o eval-output.json
+      - name: Download baseline
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: gh run list --workflow agent-eval.yml --branch main --status success --limit 1
```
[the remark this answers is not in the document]
mldangelo-oai: Resolved at bdd448b39 by deleting the nonfunctional baseline-download step and its comparison promise instead of adding a long artifact workflow.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,170 @@
...
+
+Use `contextTransform` when the provider response carries retrieved documents. Keep trace checks alongside that grader to prove retrieval occurred.
+
+## CI
+
+Use an isolated config directory per job and fail if the receiver cannot start:
+
+```yaml
+permissions:
+  actions: read
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed at bdd448b39: the CI fragment now grants contents: read for checkout alongside actions: read.

examples/integration-opentelemetry/javascript/promptfooconfig.trajectory.yaml
```diff
@@ -26,12 +23,6 @@ tests:
       - type: trajectory:tool-used
         value: compose_answer
 
-      - type: trajectory:tool-args-match
-        value:
-          name: search_corpus
-          args:
-            query: quantum computing classical computing
-
       - type: trajectory:tool-sequence
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed at bdd448b39: the README no longer advertises tool-args-match because the safe offline trajectory config intentionally omits captured tool arguments.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,170 @@
...
+  -c promptfooconfig.trace-guide.yaml --no-cache -o output.json
+```
+
+Inspect `output.json` for row success, score, errors, and trace identifiers. Exported results can include trace spans and their attributes. Treat exported files and shared evals as sensitive when spans contain prompts, tool arguments, or user data.
+
+## Instrument a provider
+
+Enable the local receiver and fail loudly if its port cannot bind:
+
+```yaml title="promptfooconfig.yaml"
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed at bdd448b39: the tracing-only YAML fragment no longer has a filename title.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,170 @@
+---
+title: 'Trace-Based Agent Evals: Tool Calls, Trajectories, and CI'
+description: Evaluate agent outputs and OpenTelemetry traces with Promptfoo assertions for tools, ordering, latency, and errors.
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed at bdd448b39: the front-matter description is now 159 characters.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,168 @@
...
+`trajectory:goal-success` asks a model judge whether the output and summarized path achieved a goal. Pair it with deterministic checks for required operations, forbidden tools, and hard budgets.
+
+## Retrieval quality
+
+A trace proves retrieval ran; it does not prove the retrieved context was relevant or faithfully used. Pass the actual retrieved context to a RAG grader:
+
+```yaml
+tests:
+  - vars:
+      context: file://retrieved-context.txt
```
jameshiester-oai: **[P2] Supply a context source that readers can load.** The linked JavaScript example has no `retrieved-context.txt`, so copying this snippet into its config fails during `file://` variable expansion (`No files found for variable context at path retrieved-context.txt`) before `context-faithfulness` runs. The latest change fixes the missing `query` input, but this new path is still required at runtime. Add a small fixture, show inline sample context, or demonstrate `contextTransform` against the provider's actual retrieval output.

jameshiester-oai (commented):
## Exact-head review

**Verdict:** one new P2 documentation reproducibility issue. Reviewed the full PR against base `10a08bd12af379f5c7e03b2878f5ac1350154e90`, the changes since reviewed ancestor `af75854ed0c793364a782e47f547290d6f1f7c0d`, prior feedback, and current source behavior. The recent correction supplies `query` and `threshold` for the context-faithfulness snippet, but its new `file://retrieved-context.txt` target is absent from the linked example; variable loading fails before grading if copied there. Current comments already cover retrieval-span presence and the endpoint suffix; I did not duplicate them.

**QA:** local source CLI with `--no-cache` and isolated state: the offline trajectory config passed 1/1 with five assertions; the base example failed its configured one-second duration ceiling as expected (two rows), with zero execution errors. Exported JSON was inspected. `git diff --check` passed. Site tests and docs build are green in CI. The sole red Web UI job failed during Playwright setup because its Chromium headless executable was missing, before test collection; this is unrelated to the docs diff. The trace-guide config requires an external model key and was not run. The requested review toolkit launcher was unavailable (exit 127), so I used its six lenses manually with three read-only specialist passes.

mldangelo-oai:
@codex review

examples/integration-opentelemetry/javascript/provider-simple-traced.js
```diff
@@ -8,8 +9,13 @@ const { BatchSpanProcessor } = require('@opentelemetry/sdk-trace-base'); // Use
...
 
-// Configure OTLP exporter
-const exporterUrl = process.env.OTEL_EXPORTER_OTLP_ENDPOINT || 'http://localhost:4318/v1/traces';
+// Configure OTLP exporter. The generic endpoint is a base URL; the trace-specific
+// endpoint is already the full export URL.
+const exporterUrl =
+  process.env.OTEL_EXPORTER_OTLP_TRACES_ENDPOINT ||
+  (process.env.OTEL_EXPORTER_OTLP_ENDPOINT
+    ? `${process.env.OTEL_EXPORTER_OTLP_ENDPOINT.replace(/\/+$/, '')}/v1/traces`
+    : 'http://127.0.0.1:4318/v1/traces');
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in d648d5e13. The example now accepts either a generic OTLP base endpoint or an already-complete /v1/traces endpoint, while OTEL_EXPORTER_OTLP_TRACES_ENDPOINT remains authoritative.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,168 @@
...
+        - authorization
+        - secret
+        - api_key
+```
+
+For a required trace, pair duration or error budgets with a presence check. Duration and error assertions can pass when no spans match.
+
+```yaml
+- type: trace-span-count
+  value: { pattern: rag_agent_workflow, min: 1 }
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in d648d5e13. The required-span check now uses retrieve_document_* so it guards the same spans as the duration budget.

site/docs/guides/trace-based-agent-evals.md
mldangelo-oai: Fixed in d648d5e13. The context-faithfulness snippet now uses inline sample context, so readers can copy it without a missing fixture file.

site/docs/tracing.md
```diff
@@ -415,22 +411,9 @@ export OTEL_SERVICE_NAME="my-rag-application"
...
-  otlp:
-    http:
-      enabled: true
-  forwarding:
-    enabled: true
-    endpoint: 'http://jaeger:4318' # or Tempo, Honeycomb, etc.
-    headers:
-      'api-key': '{{ env.OBSERVABILITY_API_KEY }}'
-```
+Promptfoo's OTLP receiver stores traces for evals. If you also need traces in Jaeger, Tempo, Honeycomb, or another OTLP-compatible backend, configure your provider SDK or collector pipeline to export there as well.
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in d648d5e13. The external-collectors section now states explicitly that Promptfoo does not forward received spans.

mldangelo-oai:
@codex review

examples/integration-opentelemetry/javascript/README.md
```diff
@@ -12,15 +12,23 @@ npx promptfoo@latest eval
...
 To run the trajectory assertion variant from this directory, use:
 
 ```bash
 npx promptfoo@latest eval -c promptfooconfig.trajectory.yaml --no-cache
 ```
 
+To run the complete trace-based agent eval guide variant, including the model-graded `trajectory:goal-success` assertion, set `OPENAI_API_KEY` and run:
+
+```bash
+OPENAI_API_KEY="your-api-key" npx promptfoo@latest eval -c promptfooconfig.trace-guide.yaml --no-cache
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in f5c564d29. The checked-in trace guide example config now sets sharing: false, so the documented command stays local by default.

site/docs/guides/trace-based-agent-evals.md
```diff
@@ -0,0 +1,168 @@
...
+cd integration-opentelemetry/javascript
+npm install
+npx promptfoo@latest eval -c promptfooconfig.trajectory.yaml --no-cache -o output.json
+```
+
+The trace guide variant adds a model-graded assertion and needs `OPENAI_API_KEY`:
+
+```bash
+OPENAI_API_KEY="your-api-key" npx promptfoo@latest eval \
+  -c promptfooconfig.trace-guide.yaml --no-cache -o output.json
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in f5c564d29. All three OpenTelemetry example configs now set sharing: false, so the site guide commands do not auto-share trace-bearing evals.

examples/integration-opentelemetry/javascript/provider-simple-traced.js
```diff
@@ -311,25 +315,26 @@ class SimpleTracedProvider {
...
             response = {
               text:
                 `Based on my analysis of ${documents.length} technical documents, here's a comprehensive explanation:\n\n` +
-                `${userIntent.entities.join(' and ')} are fascinating topics in computer science. ` +
-                `After analyzing multiple sources including arxiv papers and textbooks, I can provide the following insights:\n\n` +
-                `1. Core Concepts: The fundamental principles involve...\n` +
-                `2. Key Differences: When comparing these technologies...\n` +
-                `3. Practical Applications: In real-world scenarios...\n\n` +
-                `This synthesis is based on recent research and established knowledge in the field.`,
+                `Quantum computing uses qubits, which can represent richer probability states than classical bits before measurement. ` +
```
jameshiester-oai: [P2] Preserve topic-aware output in the shared tracing example. This hard-coded quantum explanation is returned for every prompt, but the unchanged base `promptfooconfig.yaml` still includes a `topic: machine learning` test. I ran the no-key local base eval and its machine-learning row exported a quantum-computing answer; the trace assertions do not check topical relevance, so the demo now teaches a misleading pass/fail signal. Generate the explanatory body from the requested topic, or isolate the fixed quantum fixture to the new guide variant, and cover both base topics.

jameshiester-oai (commented):
Reviewed the current exact head across correctness, security, API/configuration, tests, performance, and maintainability. The offline trajectory config passed with its trace assertions and 19 captured spans; a no-key base eval exposed the topic mismatch below. Focused tests (238) and the site build passed. CI has 32 green checks; its security-scan failure is a GitHub PR-files API 500, not a test failure. Prior endpoint, trace presence, fixture, and sharing concerns were already addressed in the current code.

mldangelo-oai:
@codex review

Please review current head 71ca3902271b77b505eb90714990d5c7c464f845, including the latest changes. The wrong-topic regression fails before the fix. Four built CLI rows verify both topics, five offline trajectory assertions and 11 guide assertions; two intentional duration failures, no execution errors. Local fixture judge only. Full build/docs/types, architecture, Knip, lint/format pass.

examples/integration-opentelemetry/javascript/provider-simple-traced.js
mldangelo-oai: Fixed in 71ca3902271b77b505eb90714990d5c7c464f845. The traced provider selects a topic-specific explanation and intent entities. The base config now checks its topic in every output. The CLI reproduces the machine-learning failure before the fix and passes both topic checks afterward; the offline trajectory and guide examples also pass.

mldangelo-oai:
@codex review

Please review current head dfefff4ef52acd8fb041d7b8f0017540cfb2bb9a, including the latest changes. Merged reviewed parent #9030 at c4ba986c. Rebuilt core/app, checked types, and reran all three guide examples: both topic checks, five trajectory checks and 11 guide assertions pass; the two documented duration failures remain. No execution errors or paid model calls; lint/format pass.

examples/integration-opentelemetry/javascript/provider-simple-traced.js
mldangelo-oai: Fixed by the concurrent remote commit 8a67568e8 retained at current head dfefff4ef: traced responses now choose quantum-computing or machine-learning explanations from the requested topic, and promptfooconfig.yaml adds output checks for both rows. I verified both exported local-eval outputs contain their own topic with no provider errors.

jameshiester-oai (commented):
Reviewed exact head dfefff4ef52acd8fb041d7b8f0017540cfb2bb9a against c4ba986c395d174e275373c7f5ff3bd79b6ba94e, including the change since my prior reviewed head f5c564d290c975223f19339aac48c304d06448ff and existing feedback. No new actionable findings. The previously reported topic mismatch is fixed: both configured topics now receive matching explanations and have topic assertions. Prior endpoint, export/privacy, context fixture, and sharing guidance is addressed.

Validation: 239 focused trace/trajectory assertion and goal-matcher tests passed. Built CLI runs used --no-cache, --no-share, and isolated state. The offline trajectory example passed with 19 exported spans; the base example produced topic-correct outputs for both rows and retained its documented duration-assertion failures, with zero execution errors and 50 exported spans. Exported JSON was inspected for outputs, grading results, root spans, and absence of raw prompt.text/tool.arguments attributes. The model-graded guide variant was source-reviewed but not run against a live judge.

Core TypeScript/bundle/app build stages passed. The postbuild tsx launcher hit sandbox IPC EPERM; direct invocation of the same postbuild script with node --import tsx passed. The production docs build and diff check passed, and all 34 observed CI checks succeeded. Worktree remains clean.

The requested /pr-review-toolkit:review-pr all parallel launcher was unavailable (exit 127). Its lenses were applied manually with three independent read-only specialists covering correctness, tests/docs, and tracing/privacy; results were reconciled before submission.

PR #8880, opened 2026-04-24
fix: prevent report table header icon overlap

typpo:
Fixes Report page table header spacing so sort and filter icons no longer overlap column labels.

## Before
Before

## After
After


mldangelo-oai (approved):
nice

scutuatua-crypto (approved):

PR #8890, opened 2026-04-25
docs(site): improve vLLM judge setup

### LINKED ISSUE #8867: Improve Documentation : OpenAI compatible API
*I will test today my new yaml config files, and update/delete this issue accordingly*

I'd like to suggest an improvement of the documentation or code.

# 1: Judge using self-hosted openai API
When using a self hosted llm-as-a-judge with openai compatible API (e.g. vLLM) with a thinking capability, one **must** add `showThinking: false` in the judge provider. Else promptfoo will try to parse the json starting with the thinking content, which most often lead to errors ("string" or "Could not parse Json"). This was quite hard to debug since promptfoo doesn't show the string it was trying to parse.
IMO, this should also be the default param in the judge section.
I think this "issue" got un-noticed since the official openai API doesn't output the thinking tokens.

# 2: Param silently ignored in the yaml config
It was unclear to me how to define the judge endpoint an ended up declaring the full config in the `defaultTest/options/provider:` (same as below) but also adding this id in the `assert/provider`. As a result my config was silently ignored and I could only set the url and apikey throught global var env. 
It would be nice to have more config recipes or even better, to have promptfoo warn in its logs when some params of the yaml config are ignored. 

---

Example of a vLLM judge section (served model name: "llm_judge"):  
(not tested yet, hopefully it will, I will update soon)
```yaml
defaultTest:
  options:
    #    DO NOT put `provider:` on the "assert" itself:
    #    (assert.provider would make this be silently ignored).
    provider:
      id: openai:chat:llm_judge
      label: "Judge: llm_judge @ vLLM local"
      config:
        apiBaseUrl: http://192.168.0.5:8080/v1
        apiKey: "empty"
        temperature: 0
        max_tokens: 10000
        # showThinking: false — vLLM 0.19.1+ puts reasoning in a separate
        # `reasoning` field and final answer in `content`. This tells promptfoo
        # to discard `reasoning` and parse JSON only from `content`.
        # The judge still thinks internally; we just don't parse the thinking.
        showThinking: false
        # Disable thinking entirely at the API level.
        # Use when the judge is a thinking-capable model but you want it to skip
        # the reasoning phase for speed / token savings.
        # passthrough:
        #   chat_template_kwargs:
        #     enable_thinking: false
  assert:
    - type: llm-rubric
       value: [...]
```

promptfoo version : 0.121.4

mldangelo-oai:
## Summary

Fixes #8867.

This PR simplifies the vLLM docs around self-hosted OpenAI-compatible judges:

- Rewrites the vLLM provider page into the canonical setup guide for local targets, vLLM-as-judge, provider maps, thinking output, GPT-OSS, GLM-4.7, and troubleshooting.
- Adds one model-graded overview section for OpenAI-compatible thinking judges instead of repeating metric-specific notes across every assertion page.
- Adds a short `llm-rubric` pointer for the common vLLM judge case and clarifies provider precedence: `assert.provider` overrides `test.options.provider`, which overrides `defaultTest.options.provider`.
- Keeps version-sensitive defaults and model install details evergreen by describing selection behavior and linking to upstream vLLM recipes for release-specific guidance.
- Clarifies that the built-in OpenAI grader can spend hidden reasoning tokens internally but does not expose private reasoning text in the grader output.

## Validation

- Ran real vLLM locally with `Qwen/Qwen3-0.6B` served as `llm_judge`.
- Verified direct vLLM responses split completed Qwen3 thinking into `message.reasoning_content` and final `message.content`; with too little generation budget, vLLM returned raw `<think>` in `content`, so the docs call out the truncation caveat.
- Ran promptfoo configs showing `showThinking: true` exports `Thinking: ...`, `showThinking: false` returns only final content, and `passthrough.chat_template_kwargs.enable_thinking: false` disables Qwen3 thinking at the vLLM API level.
- Ran real Promptfoo judge coverage through vLLM for `llm-rubric`, `g-eval`, `factuality`, `model-graded-factuality`, `model-graded-closedqa`, `answer-relevance`, `context-recall`, `context-relevance`, `context-faithfulness`, `conversation-relevance`, `select-best`, and `trajectory:goal-success`.
- Confirmed `search-rubric` does not use plain vLLM chat as the web-search grader.
- Verified the built-in OpenAI grader path returns final JSON without visible `Thinking:` / `Reasoning:` text, while still reporting hidden reasoning token usage.
- `npm run f`
- `npm run l`
- `git diff --check`


ianw-oai (approved):
Docs-only change looks good to me. The remaining provider-precedence clarifications are useful follow-ups but not blocking.

PR #8894, opened 2026-04-25
docs(site): clarify JSONL assertion component results

Rul1an:
## Summary

This adds a small note to the JSONL output docs explaining where per-assertion details live.

The current section shows the row-level `success` and `score` fields. When consuming JSONL programmatically, it is useful to know that those are aggregate row fields, while individual assertion outcomes are available under `gradingResult.componentResults[]`.

## Change

- Document that `gradingResult.componentResults[]` contains per-assertion pass/fail, score, reason, and assertion metadata.
- Clarify that top-level `success`, `score`, and `gradingResult` are row-level aggregate result fields.

## Validation

Ran `git diff --check`.


mldangelo:
Thank you!!!

PR #8896, opened 2026-04-25
fix(providers): structured 429 classification in Azure assistants

mldangelo:
## Summary

The `isRateLimitError` helpers in `src/providers/azure/assistant.ts` and `src/providers/azure/foundry-agent.ts` infer 429s by substring-matching error messages. After `fetchWithRetries` exhausts retries, the only signal that reaches the provider is `Rate limited: 429 ...` — `Retry-After`, `X-RateLimit-Reset`, and the body-level error code (e.g. `insufficient_quota` vs `rate_limit_exceeded`) are dropped on the floor. Under concurrent evals this turns capacity limits into retry amplification, retries hard quotas that won't recover, and synchronizes retries because `handleRateLimit` waits the bare `Retry-After` with no jitter.

This change replaces that path with a structured signal:

- New `HttpRateLimitError` in `src/util/fetch/errors.ts` with `status`, `statusText`, `kind: 'quota' | 'rate_limit'`, `code`, `retryAfterMs`, `resetAt`, `headers`, `body`. Recognises hard-quota codes (`insufficient_quota`, `billing_hard_limit_reached`, `billing_not_active`, `access_terminated`, `quota_exceeded`).
- `fetchWithRetries` peeks the 429 body once via `peekRateLimitBody` (clones the response so the cache layer can still read it), fails fast on hard-quota codes without consuming retries, and throws `HttpRateLimitError` on retry exhaustion instead of a bare-string `Error`.
- `handleRateLimit` adds 0–1000ms uniform jitter so concurrent waiters don't all retry in the same instant after `Retry-After`.
- `azure/assistant.ts` and `azure/foundry-agent.ts` recognise the structured error in `formatError` and produce distinct messages: quota errors include `Retries will not help — check your billing or daily quota`; per-window rate limits include `[retry after Xs]` / `[resets in Xs]`. `isRetryableError` treats `kind === 'quota'` as non-retryable. Substring fallback widened to `'too many requests'` and `'quota exceeded'` for back-compat.
- Foundry path also catches the OpenAI-SDK shape (`error.status === 429`) since it doesn't go through `fetchWithRetries`.

The structured-error message keeps the substrings `429` / `rate limit` / `too many requests` so the legacy classifier in `scheduler/providerRateLimitState.ts` still matches.

## Why

Reported by Brent Williams: *\"At eval scale, a misclassified 429 isn't a retry detail — it changes the outcome of the run. In `assistant.ts`, 429s are inferred via message matching and routed into retryable behavior, without inspecting Retry-After or distinguishing quota-style limits. Under concurrent evals, that can turn capacity limits into retry amplification and skew results.\"*

## Test plan

- [x] `npx vitest run test/util/fetch test/fetch.test.ts test/providers/azure test/scheduler test/cache.test.ts` — all green
- [x] `npx vitest run test/providers` (broad sweep) — 5363/5363
- [x] `npx tsc --noEmit` — clean
- [x] `npm run l && npm run f` — clean (3 pre-existing complexity warnings, unchanged from main)
- [x] End-to-end probe vs a live HTTP fixture: hard quota fails fast (1 attempt, no retry), per-window rate limit retries until exhausted with `retryAfterMs` populated, recovers when a later attempt succeeds, jitter spreads concurrent retries (~1.8s vs ~1.9s gaps observed on `Retry-After: 1`)
- [x] Legacy substring classifiers (`scheduler/providerRateLimitState.ts`) still match the new error message

## New tests

- `test/util/fetch/errors.test.ts` — `isHardQuotaCode`, `extractRateLimitErrorCode`, `HttpRateLimitError` classification, type guard, back-compat substring assertions
- `test/fetch.test.ts` — `HttpRateLimitError classification` describe (5 tests: rate-limit retry exhaustion, `insufficient_quota` fail-fast, `billing_hard_limit_reached` fail-fast, header preservation, plain-text body fallback); `computeRateLimitWaitMs` describe (3 tests); jitter behaviour in `handleRateLimit` (2 new tests)
- `test/providers/azure/assistant.test.ts` — `structured HttpRateLimitError handling` describe (5 tests covering rate-limit vs quota formatting, retry-after detail, resetAt fallback, polling-time error surfacing) + extended `isRetryableError` cases for quota codes and structured errors


mldangelo:
Reviewed end-to-end. Synced with `main` (one trivial conflict in `test/fetch.test.ts` from the recent `getRequestTimeoutMs` rename), pushed two follow-ups in `bea514637`:

### Gap closed: scheduler retry-amplification on hard quotas

The transport-layer fail-fast on `insufficient_quota` was being undermined one layer up. `ProviderRateLimitState.executeWithRetry` substring-matches on the rendered error message (`429`, `rate limit`, `too many requests`) — which the new `HttpRateLimitError` deliberately preserves for back-compat — and routes it through `shouldRetry`, which **always retried rate limits up to `policy.maxRetries`**, regardless of `kind`.

End-to-end probe against a fixture that returns `429 + insufficient_quota`:

| | wall time | server hits |
|---|---|---|
| Before | **4m 1s** | **8** |
| After  | **1s**   | **1** |

Fix: `shouldRetry` checks `isHttpRateLimitError(error) && error.kind === 'quota'` before the retry path. Per-window rate limits still retry honoring `Retry-After` (probe 2 confirmed: 3 attempts, ~1.7s jittered gaps).

### Codex P2 (readBoundedText) addressed

Each streamed chunk is now bounded *before* it enters the buffer (and copied via `.slice()` so the upstream `ArrayBuffer` for an oversized chunk is releasable). Non-stream fallback consults `Content-Length` to skip materializing very large bodies entirely. Added a streaming-path test that pushes 1 MB through and verifies the bound holds.

### Simplifications

- `rateLimitFromSdkError` returns the `HttpRateLimitError` directly instead of a `{kind, code}` intermediate (kind is derived in the constructor).
- Replaced `HARD_QUOTA_ERROR_CODES.has(code)` callsites with the existing `isHardQuotaCode(code)` predicate.

### Verification

- `npx vitest run test/scheduler test/providers test/util test/fetch.test.ts test/cache.test.ts test/rateLimit.test.ts` — 7360 passed
- `npx tsc --noEmit` — clean
- `npm run l` / `npm run lint:tests` — only pre-existing complexity warnings (unchanged from main)
- Live-fixture probe vs `npm run local -- eval`: quota fail-fast (1 hit / 1s), rate-limit retries with jitter (3 hits / 1.7s gaps)

### Implementation completeness vs Azure / OpenAI

Scope is appropriate. OpenAI / Anthropic / Bedrock / Vertex providers all use `fetchWithCache` → `fetchWithRetries` and now inherit:
- The structured `HttpRateLimitError` with code/retryAfterMs/resetAt.
- Quota fail-fast at the transport.
- Quota fail-fast at the scheduler (this PR's new fix).
- Back-compat substring matching is preserved.

The Azure assistant providers needed special handling because they have their own polling/retry surface that calls `formatError` directly. Generic providers don't, so they don't need provider-side `instanceof HttpRateLimitError` branches — the structured error flows through `String(err)` with all the substrings legacy classifiers want.

PR #8898, opened 2026-04-25
docs(site): fix link in announcement banner

typpo:
Skips the 308 redirect.

The announcement banner linked to the OpenAI update without the canonical trailing slash. Docusaurus redirects that internal URL to the trailing-slash path, which made crawlers report a 308 for the home page link.

This updates the banner href to point directly at `/blog/promptfoo-joining-openai/`.

Checks:
- `npm run l`
- `npm run f`


mldangelo (approved):
beautiful

PR #8899, opened 2026-04-26
docs: clarify security trust model

mldangelo-oai:
## Summary

- Clarify the security model for local web UI, local helper servers, red team target/provider setup pages, and MCP interfaces, including CORS/CSRF, bind address, same-user local access, embedded client configuration, and workspace/file-write expectations
- Document hosted-feature, Cloud, sharing, telemetry, transport, disable-flag, and opt-out scope more explicitly, including the distinction between remote-generation-backed red team setup helpers and non-generation setup/test helpers
- Define local artifacts, caches, token stores, cache namespaces, report/export content, and contributor automation as scoped developer surfaces unless a documented product boundary is crossed
- Align FAQ, red team data-handling, command-line, and sharing docs with the hosted-feature payload and sharing snapshot model, with red team target/provider setup/test payloads scoped narrowly to red team flows

## Validation

- `npm run f`
- `npm run l`
- `git diff --check`
- `SKIP_OG_GENERATION=true npm run build` (from `site/`)


ianw-oai (commented):
One wording mismatch to fix before I can approve: `PROMPTFOO_DISABLE_REMOTE_GENERATION` is now documented as disabling setup helpers that rely on remote generation, but `/providers/http-generator` still sends request/response examples to the hosted HTTP provider generator without checking `neverGenerateRemote()`. Please either add that guard or narrow the docs to explicitly exclude HTTP provider generation.

PR #8902, opened 2026-04-26
fix(telemetry): mirror person properties on every event

mldangelo:
## Summary

- Mirror PostHog person properties (`email`, `isLoggedIntoCloud`, `authMethod`, `isRunningInCi`) onto every captured event via `$set`, not just the auto-fired `$identify`.
- Dashboards that exclude CI traffic by checking the **event** property `isRunningInCi != true` were leaking identify-only sessions, because PostHog stores person properties under `$set` rather than at the event top level. Any user that produced `$identify` but no real event (e.g. ephemeral CI/Docker installs that import the CLI but exit) slipped through and inflated "non-CI" cohorts.
- After this change person properties stay current on every event, and filtering by the person property `isRunningInCi` becomes reliable regardless of which event a tile inspects.

No new events are emitted — the change is one extra `$set` field on the existing capture payload.

## Test plan

- [x] `npx vitest run test/telemetry.test.ts` — 22/22 pass
- [x] `npm run l && npm run f` clean
- [x] `tsc --noEmit` clean
- [ ] After release, verify in PostHog that recent `command_used`/`eval_ran` events include `properties.$set.isRunningInCi` and that the person property updates as expected


mldangelo-oai (approved):
Reviewed the updated telemetry path after the fixes. No remaining findings; CI is green on head 5d5f9f8f39.

PR #8968, opened 2026-04-28
fix(server): restrict browser CORS origins

ychhabria:
## Summary
- stop emitting wildcard browser-readable CORS headers from the local server
- reuse the CSRF localhost-equivalence and PROMPTFOO_CSRF_ALLOWED_ORIGINS policy for Express responses
- restrict Socket.IO CORS to no-origin, localhost, or explicitly allowlisted origins
- log a clear recovery hint when an untrusted browser origin is blocked so intentional cross-origin deployments can add the exact origin instead of debugging a silent CORS failure

## Compatibility
- normal promptfoo local usage and same-origin browser access are unchanged
- localhost aliases such as 127.0.0.1 and local.promptfoo.app remain allowed
- CLI/script clients without an Origin header are not affected
- deployments that intentionally serve the UI from one origin and the API from another should set PROMPTFOO_CSRF_ALLOWED_ORIGINS to the exact trusted browser origin

## Tests
- npx vitest run test/server/middleware/csrfProtection.test.ts test/server/server.test.ts
- npx @biomejs/biome check src/server/middleware/csrfProtection.ts src/server/server.ts test/server/middleware/csrfProtection.test.ts


mldangelo-oai (changes requested):
This will break many deployments.

PR #8975, opened 2026-04-29
fix(api): preserve error details and stack traces in shared helpers

mldangelo:
## Summary

Two regressions introduced by the recent DTO/error-helper rollout (#8922 et al):

1. **Validation errors lost the structured `details` field.** The pre-DTO telemetry handler returned `{ error: 'Invalid request body', details: z.prettifyError(...) }`. After `replyValidationError` was added, the body became `{ error: '<pretty>' }` only. `ErrorResponseSchema` still declares `details` optional, so the contract said one thing while the helper produced another. Programmatic clients that read `body.details` now see `undefined`.
2. **Stack traces are dropped from server logs.** `sendError` passed raw `Error` instances into the logger's object-context serializer, which goes through `JSON.stringify`. Because `name`, `message`, and `stack` are non-enumerable on `Error`, the serialized payload was effectively `{}`. Now that `sendError` is the canonical funnel for every 500 across `eval`, `providers`, `redteam`, `media`, `traces`, etc., the regression surface is dramatically wider than the old ad-hoc handlers.

## Changes

- `src/server/utils/errors.ts`
  - `replyValidationError` keeps the prettified message in `error` (matching existing test assertions) and additionally exposes `details: { issues: error.issues }` so clients can map failures back to fields.
  - `sendError` now extracts `name`/`message`/`stack`/`cause` from `Error` instances before logging, so stacks survive object-context serialization. Non-Error values pass through verbatim.
- `test/server/utils/errors.test.ts` (new) — covers the public message, no-leak guarantee, structured Error logging, `cause` preservation, the no-internalError path, and the validation `details` shape.

No public API/wire-format break: every existing test that asserts on `response.body.error` continues to pass (the prettified string is unchanged); only the previously-empty `details` field now carries the Zod issue array.

## Test plan
- [x] `npx vitest run test/server/utils/errors.test.ts` (new file, 7 tests)
- [x] `npx vitest run test/server/` (473 tests pass; existing validation tests unchanged)

### DIFF
--- a/src/server/utils/errors.ts
+++ b/src/server/utils/errors.ts
@@ -4,6 +4,27 @@ import { ErrorResponseSchema } from '../../types/api/common';
 import type { Response } from 'express';
 
 /**
+ * Build a logger-safe context object from an unknown error value.
+ *
+ * The default logger object-context serializer goes through `JSON.stringify`,
+ * which drops `name`, `message`, and `stack` from `Error` instances because they
+ * are non-enumerable. Extract them explicitly so `logger.error` retains the stack.
+ */
+function toLogContext(error: unknown): Record<string, unknown> {
+  if (error instanceof Error) {
+    return {
+      error: {
+        name: error.name,
+        message: error.message,
+        stack: error.stack,
+        ...(error.cause === undefined ? {} : { cause: error.cause }),
+      },
+    };
+  }
+  return { error };
+}
+
+/**
  * Send a standardized error response.
  *
  * All error responses use the shape `{ error: string }`.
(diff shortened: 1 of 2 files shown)


mldangelo:
Ran a critical self-review pass on this PR using `pr-review-toolkit:review-pr` (code-reviewer + pr-test-analyzer + silent-failure-hunter + comment-analyzer agents in parallel). All four came back with no critical findings, but two real concerns surfaced that are worth fixing before merge — both flagged independently by multiple agents:

**1. Cause chains reintroduce the JSON.stringify bug the helper exists to fix.** Forwarding `Error.cause` as-is to the logger means the next serializer drops `name`/`message`/`stack` from the cause Error — exactly the regression the top-level extraction was meant to prevent. `serializeError` now recurses with a depth cap (4) and `seen` Set so cycles (`err.cause = err`) don't blow the stack.

**2. `ErrorResponseSchema.parse()` itself is an unhandled-throw vector.** This is the canonical 500 funnel for every route. If the parse throws (future schema tightening, malformed `z.prettifyError` output), the throw escapes the caller's `try/catch` and lands in Express's default error path: empty 500, no JSON body. New `safeRespond` wraps parse with a hand-built `{ error }` fallback so the funnel can never silently fail.

Doc fixes from comment-analyzer:
- `toLogContext` JSDoc no longer overcommits to "JSON.stringify" since `src/logger.ts` has multiple (also lossy) serialization paths.
- `sendError` JSDoc narrows the shape claim to what the function actually emits.
- `replyValidationError` JSDoc adds a note that custom `.refine` messages would also reach the wire — so future schema authors should audit before adding refines that interpolate input.

Five new tests pinning the new behavior (12 total, was 7):
- `null` is logged as `{ error: null }` (intentional, vs accidentally being skipped).
- Error subclass `name` preserved (TypeError, custom subclass).
- Recursive cause chain — three nested stacks all survive.
- Circular cause does not crash.
- Schema parse-throw → hand-built fallback envelope.

`npx vitest run test/server/` 478 pass, `npx tsc --noEmit` clean. Lint warning on `src/models/eval.ts:889` is pre-existing on main from #8973 (unrelated).

Pushed in 5c655bff5.

mldangelo:
Addressed @chatgpt-codex-connector P2 in c20d1d964.

Real regression — the previous `JSON.stringify(error)` path actually preserved enumerable own props (`code`, `errno`, `syscall`, `path`, AWS SDK `$metadata`, etc.) while dropping non-enumerable `name`/`message`/`stack`. My explicit-fields rewrite flipped which set was lost. Now spread enumerable props first, then overlay the four standard fields so they always win — both sets reach the log.

Added a regression test pinning the Node `SystemError` shape (`code`/`errno`/`syscall`/`path` all preserved alongside `name`/`message`/`stack`).

13 tests pass (was 12), full server suite 479 pass.

PR #8976, opened 2026-04-29
fix(api): correct user-visible response regressions

mldangelo:
## Summary

Two regressions introduced by the recent DTO/validation rollout that would surface to external SDK / CLI consumers on the next public release.

### 1. Rating endpoint shape regression

`POST /api/eval/:evalId/results/:id/rating` used to return the persisted `EvalResult` row, so SDK / CLI clients could read refreshed grading state without a follow-up GET. After #8924 it returned only `{message: "Rating submitted"}`. The bundled web UI only checks `response.ok` and is unaffected, but every external consumer that read the returned record broke silently — and there was no test guarding the old shape.

- Restore `res.json(result)` in `src/server/routes/eval.ts`.
- Tighten `SubmitRatingResponseSchema` to require `id`, `success`, `score` while passing through the rest of the row.

### 2. Share endpoint error classification

`POST /api/results/share` only branched on `null` returns from `readResult` / `Eval.findById`. When either threw (e.g. SQLite "database is locked"), the error propagated to Express's default handler — bypassing `sendError`, the structured logger, and the new `ErrorResponseSchema` contract. Worse, a UI that reads the body would see an opaque default-Express HTML page.

- Wrap both lookups so DB failures surface as 500 with the public "Failed to load eval for share" message.
- Genuine "row not found" returns continue to map to 404.

## Test plan

- [x] `npx vitest run test/server/eval.test.ts` — added assertion that the rating response carries the persisted EvalResult fields, not just a message envelope.
- [x] `npx vitest run test/server/routes/serverApiSchemas.test.ts` — added two regressions: throws from `readResult` and `Eval.findById` now produce a 500 (not a misleading 404) and never advance to URL creation.
- [x] `npx vitest run test/types/api/eval.test.ts` — schema test updated for the new rating response shape.
- [x] `npx vitest run test/server/ test/types/api/` — 474 tests pass.
- [x] `npm run openapi:generate` — regenerated spec reflects the corrected `SubmitRatingResponse` shape.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -632,6 +632,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "SEE LICENSE IN LICENSE.md",
       "optional": true,
       "os": [
@@ -646,6 +649,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "SEE LICENSE IN LICENSE.md",
       "optional": true,
       "os": [
@@ -660,6 +666,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "SEE LICENSE IN LICENSE.md",
       "optional": true,
       "os": [
@@ -674,6 +683,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "SEE LICENSE IN LICENSE.md",
       "optional": true,
       "os": [
@@ -4257,6 +4269,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -4274,6 +4289,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -4291,6 +4309,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -4308,6 +4329,9 @@
         "x64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT OR Apache-2.0",
       "optional": true,
       "os": [
@@ -13162,6 +13186,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -13179,6 +13206,9 @@
         "arm64"
       ],
       "dev": true,
+      "libc": [
+        "musl"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -13196,6 +13226,9 @@
         "ppc64"
       ],
       "dev": true,
+      "libc": [
+        "glibc"
+      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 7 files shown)


mldangelo:
Addressed both review findings in 013bd7cdd:

**@Copilot / @chatgpt-codex-connector P1 — `readResult` swallowing DB errors**: confirmed by reading `src/util/database.ts` — `readResult` wraps its work in `try/catch` and returns `undefined` on any failure, so my original `try/catch` around it was a no-op. Dropped the `readResult` preflight entirely and lean on `Eval.findById` directly: `findById` returns `undefined` only for missing rows and otherwise throws, which now funnels through `sendError` for a structured 500. Side effect: also removes the redundant DB read Copilot flagged in the second comment.

**Build failure (TS2352)**: cast in the new `SubmitRating` schema test went through `as { gradingResult }` which TS rejected because the inferred type doesn't statically include `gradingResult`. Cast through `unknown` so the runtime assertion stands without claiming a static contract.

Test plan
- `npx vitest run test/server/routes/serverApiSchemas.test.ts test/server/eval.test.ts test/types/api/eval.test.ts` — 44 pass.
- `npx vitest run test/server/ test/types/api/` — 473 pass.
- `npx tsc --noEmit` — clean.
- New regression: `Eval.findById` rejecting on a "sqlite: database is locked" simulates the case the original PR claimed to fix; the previous test was watching a path that never fired in production.

mldangelo:
Critical self-review pass on this PR via `pr-review-toolkit:review-pr` (code-reviewer + silent-failure-hunter + pr-test-analyzer agents in parallel).

**code-reviewer**: walked through all seven of my pointed questions — verified `Eval.findById` actually throws on DB errors, confirmed `res.json(result)` adds no new attack surface (gradingResult is client-supplied via `SubmitRatingRequestSchema`), confirmed mocks are clean (`vi.resetAllMocks` in both before/after each), and confirmed `result.save()` mutates in place so the response is fresh. Verdict: "safe to land."

**Three issues found and fixed in 43957832e**:

1. **Rating refresh test under-verified the regression** (test-analyzer P1, criticality 8). My test sent `pass=true` to `results[0]` (already-passing in `EvalFactory`), so the response's `success`/`score` were unchanged from pre-rating values — the test would have passed against a regression that returned a pre-save snapshot. Now picks the failing `results[1]`, captures pre-state (`pass=false, score=0, success=false`), and asserts the response carries strict post-state (`pass=true, score=1, success=true, reason contains 'Manual result'`). Cannot pass without genuine refresh on the wire.

2. **404 path didn't assert `readResult` not called** (test-analyzer #2). Only the happy path got that assertion. Added the symmetric one so a future refactor can't silently reintroduce a `readResult` preflight on the not-found branch and resurrect the DB-error-as-404 regression.

3. **Validation-helper inconsistency in rating route** (silent-failure-hunter #3). Used ad-hoc `res.status(400).json({error: prettifyError})` instead of `replyValidationError`. Today both produce the same wire shape, but the helper parses through `ErrorResponseSchema` (and exposes `details.issues` per #8975). Switched both `paramsResult` and `bodyResult` failure paths.

**Out-of-scope follow-up** (silent-failure-hunter Critical):

`createShareableUrl` returns `null` on several paths (`PROMPTFOO_DISABLE_SHARING`, empty results, upload failures in `src/share.ts:384-386, 508, 624-627, 654-656`) and this route forwards `{url: null}` as 200 OK. Same silent-failure class — the frontend/CLI sees a "successful share" with no URL. The `Zod` schema (`src/types/api/server.ts:90-92`) explicitly permits `url: nullable().optional()`, so this is wire-shape-legal but operationally ambiguous. The PR's explicit scope was the *load* step; fixing the *create* step requires schema + frontend coordination so I'd recommend a separate PR. Happy to open one if you want.

**Windows test failure (`test/agentSkills/promptfooPlugin.test.ts:2186`)** is unrelated to this PR — Python file provider invocation, 30s timeout. Same flake on `main` per the run log.

`npx vitest run test/server/ test/types/api/` 473 pass, `npx tsc --noEmit` clean.

PR #8977, opened 2026-04-29
fix(api): make modelAudit safeRespond fire-safe under parse failures

mldangelo:
## Summary

`safeRespond` in `POST /api/model-audit/scan` ran the DTO schema parse *before* flipping `responded = true`. A thrown ZodError inside a child-process event callback would:

1. Propagate as an unhandled exception (no try/catch around `modelAudit.on('error', ...)` or several `close`-branch call sites).
2. Leave `responded = false`, so a racing emitter (e.g. `close` firing right after `error`) could re-enter `safeRespond` and attempt a second `res.json()` on the same response — usually crashing the request handler.

Today this is theoretical because `errorDetails` only spreads enum-typed strings, but the surface is fragile: any future contributor adding a non-string `details` value, or any drift between `ScanErrorResponseSchema` and what the scanner emits, becomes an unhandled-exception bug instead of a clean 500.

## Changes

`src/server/routes/modelAudit.ts`:
- Set `responded = true` immediately on entry — before parse — so a throw cannot leave the latch open.
- Wrap the `Scan.Response` / `Scan.ErrorResponse` parse in `try/catch`.
- On parse failure, log structured context and degrade to a deterministic `500 { error: 'Error processing scan results' }` envelope (preserves the body's own error string when the original branch was already an error). The client always receives exactly one well-formed response.

`test/server/routes/modelAudit.test.ts`:
- New regression: race `error` + `close` events with the error-branch parse forced to throw on first call. Pre-fix, this would write a second body. Post-fix, `responded` latches and the second emitter no-ops.
- Existing "scan success DTO parsing fails" test still asserts the `Error processing scan results` message — behavior preserved.

## Test plan
- [x] `npx vitest run test/server/routes/modelAudit.test.ts` — 46 tests pass (1 new).
- [x] `npx vitest run test/server/` — 467 tests pass.
- [x] `npm run l && npm run f` (the residual cognitive-complexity warning on the existing `close` handler is pre-existing and out of scope.)

### DIFF
--- a/src/server/routes/modelAudit.ts
+++ b/src/server/routes/modelAudit.ts
@@ -296,8 +296,12 @@ modelAuditRouter.post('/scan', async (req: Request, res: Response): Promise<void
           : ModelAuditSchemas.Scan.ErrorResponse.parse(body);
         res.status(statusCode).json(parsed);
       } catch (parseError) {
+        // Match the existing logging style in this file (e.g.
+        // `logger.error('Failed to parse model scan output', { parseError, ... })`)
+        // by passing the raw error value so the logger sanitizer can preserve
+        // the stack and other diagnostic context.
         logger.error('safeRespond DTO parse failed; sending fallback envelope', {
-          parseError: parseError instanceof Error ? parseError.message : String(parseError),
+          parseError,
           statusCode,
         });
         // A parse failure on either branch means the response cannot be trusted.
--- a/src/server/routes/modelAudit.ts
+++ b/src/server/routes/modelAudit.ts
@@ -276,17 +276,39 @@ modelAuditRouter.post('/scan', async (req: Request, res: Response): Promise<void
     let stderr = '';
     let responded = false; // Prevent double-response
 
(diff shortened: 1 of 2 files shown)


mldangelo:
Addressed both Copilot comments in 89a1e9be1:

- **`src/server/routes/modelAudit.ts:300`**: now passes the raw `parseError` to the logger (matching the existing `logger.error('Failed to parse model scan output', { parseError, ... })` style at line 495). PR #8975's logger-context sanitizer will preserve stack/cause from `Error` instances.
- **`test/server/routes/modelAudit.test.ts:340`**: reworded the comment to match what's actually tested. The single-call assertion on `errorParseSpy` is the proxy for "one body written" because `safeRespond` enters parse exactly once per call — if `responded` weren't flipped before parse, the close emitter would re-enter and we'd see two parse calls. Comment now says this explicitly instead of overclaiming a write counter.

`npx vitest run test/server/routes/modelAudit.test.ts` — 46 pass.

PR #8998, opened 2026-04-30
feat(app): humanize per-cell latency in eval results table

### LINKED ISSUE #8974: [Feature Request] Show per-test latency and token counts in human-readable format (like Duration) in the html report
## Feature Request                                                                               
                                                                                                   
  The eval results table already shows latency and token data per test cell (e.g., `Tokens: 15,176 
  (2,225+14,951) Latency: 3,789,665 ms Tokens/Sec: 4 Cost: $6.5`), but the format is hard to scan  
  at a glance.                                                                                   
                                                                                                   
  Meanwhile, the top-level **DURATION** header renders `7h 43m` — clean and instantly readable.  
                                                                                                   
  ## Proposed Change                 
                                                                                                   
  Apply the same human-readable formatting to per-cell latency and token stats:                  
                                                                                                   
  | Field | Current | Proposed |     
  |-------|---------|----------|                                                                   
  | Latency | `3,789,665 ms` | `1h 3m 9s` |                                                      
  | Latency (short) | `255,718 ms` | `4m 15s` |                                                    
  | Tokens | `15,176 (2,225+14,951)` | `15.2k (2.2k↑ + 15k↓)` |                                    
  | Tokens/Sec | `4` | `4 tok/s` |                                                                 
                                                                                                   
  ## Why                                                                                           
                                                                                                   
  - Latency values in the millions of ms are not immediately interpretable                         
  - The DURATION field already solves this for the aggregate — per-test cells should match         
  - Makes it easier to spot outlier tests (e.g., one test taking 109 minutes vs 35 minutes)
                                                                                                   
  ## Context                                                                                       
                                                                                                   
  - **promptfoo version**: 0.121.9                                                                 
  - Observed in: eval results table, per-cell footer stats

mldangelo:
## Summary

Per-cell **Latency** and column-aggregate **Avg Latency** now render via the same `formatDuration` helper that already powers the top-of-page **DURATION** chip — `1h 3m` / `4m 16s` / `2.3s` / `247ms`. Each value is wrapped in a Radix `Tooltip` exposing the precise raw ms (cached rows clarify that the displayed duration is the *original* pre-cache execution time), and triggers carry an `aria-label` so screen readers receive the exact value rather than guessing at `"1h 3m"`.

While addressing the reported asymmetry, a small batch of design-system polish lands in the same surface:

- Italicize the `(cached)` qualifier on both latency and tokens (was inline parens at full weight, now a clear meta-qualifier)
- Replace hardcoded `#888` on `.stat-item` and `.prompt-detail` with `hsl(var(--muted-foreground))` so contrast respects light/dark theme tokens
- Add `font-variant-numeric: tabular-nums` to both rows so columns of digits scan vertically — directly serves the issue's stated outlier-spotting goal
- Swap inline `style={{ cursor: 'help' }}` for the `cursor-help` Tailwind class on the column-aggregate triggers (matching `renderCostMetric`)
- Cache a single module-level `Intl.NumberFormat` instance for use across the per-cell display helpers instead of constructing one per call (hot path: hundreds of cells)

### Scope decisions vs. the issue

The issue also proposed `15,176 → 15.2k (2.2k↑ + 15k↓)` for tokens and `4 → 4 tok/s` for the Tokens/Sec metric. We intentionally did **not** ship those:

- The compact `15.2k` notation rounds away granularity that's actually useful when comparing similar tests, and the `↑↓` arrow notation is a screen-reader hazard. The current `15,176 (2,225+14,951)` already breaks down precisely with thousands-separators.
- The `Tokens/Sec:` row label already names the unit; adding `tok/s` after the value would duplicate it.

The core ask — closing the readability gap between humanized DURATION and raw-ms cells — is fully addressed.

Closes #8974

## Test plan

- [x] `npm run test:app -- src/pages/eval --run` — 1426 / 1426 pass (added 8 new tests covering humanized output across s / m / h, the `aria-label` precision, cached aria-label clarification, italicized cached suffix on both latency and tokens, and the filtered-latency suffix)
- [x] `npm run l && npm run f` clean
- [x] `tsc --noEmit` clean
- [ ] Visual: open an eval with a mix of sub-second, minute-range, hour-range, and cached rows — confirm Latency cells render humanized, hover reveals exact ms, cached rows show italic `(cached)` and a clarifying tooltip
- [ ] Light + dark theme: muted-foreground tokens render with adequate contrast in both

### Before
Before: raw millisecond latency in the eval results table

### After
After: humanized latency in the eval results table

### DIFF
--- a/src/app/src/pages/eval/components/EvalOutputCell.test.tsx
+++ b/src/app/src/pages/eval/components/EvalOutputCell.test.tsx
@@ -1156,9 +1156,50 @@ describe('EvalOutputCell', () => {
       },
     };
     renderWithProviders(<EvalOutputCell {...props} />);
-    expect(
-      screen.getByLabelText('1,234 ms — original duration; served from cache'),
-    ).toBeInTheDocument();
+    expect(screen.getByLabelText('1,234 ms (cached response)')).toBeInTheDocument();
+  });
+
+  it('renders sub-millisecond latency without rounding away precision', () => {
+    const props = {
+      ...defaultProps,
+      output: {
+        ...defaultProps.output,
+        latencyMs: 0.123,
+        response: {},
+      },
+    };
+    renderWithProviders(<EvalOutputCell {...props} />);
+    expect(screen.getByLabelText('0.123 ms')).toBeInTheDocument();
+  });
+
+  it('falls back to raw ms when latency is non-finite or negative', () => {
+    const cases: { latencyMs: number; expected: string }[] = [
+      { latencyMs: Number.POSITIVE_INFINITY, expected: '∞ ms' },
+      { latencyMs: -1234, expected: '-1,234 ms' },
+    ];
+    for (const { latencyMs, expected } of cases) {
+      const { unmount } = renderWithProviders(
+        <EvalOutputCell
+          {...defaultProps}
+          output={{ ...defaultProps.output, latencyMs, response: {} }}
+        />,
+      );
+      expect(screen.getByLabelText(expected)).toBeInTheDocument();
+      unmount();
+    }
+  });
+
+  it('renders an exact-hour latency without trailing minutes', () => {
+    const props = {
+      ...defaultProps,
+      output: {
+        ...defaultProps.output,
+        latencyMs: 129_600_000,
+        response: {},
+      },
+    };
+    renderWithProviders(<EvalOutputCell {...props} />);
+    expect(screen.getByText('36h')).toBeInTheDocument();
   });
 
   it('displays large latency values in human-readable format (minutes/hours)', () => {
--- a/src/app/src/pages/eval/components/EvalOutputCell.test.tsx
+++ b/src/app/src/pages/eval/components/EvalOutputCell.test.tsx
@@ -944,6 +944,25 @@ describe('EvalOutputCell', () => {
     renderWithProviders(<EvalOutputCell {...propsWithStandardTokens} />);
 
     expect(screen.getByText('33 (11+22)')).toBeInTheDocument();
+    expect(
+      screen.getByLabelText('11 prompt tokens + 22 completion tokens = 33 total'),
+    ).toBeInTheDocument();
+  });
+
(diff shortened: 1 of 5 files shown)


mldangelo:
Pushed [`26ca573`](https://github.com/promptfoo/promptfoo/pull/8998/commits/26ca57340) addressing all three Copilot comments + a couple of related findings from a deeper review pass:

- **Zero filtered latency was being dropped.** Compute gate changed from truthy (`filteredMetrics?.totalLatencyMs && testCount?.filtered`) to `filteredMetrics?.totalLatencyMs != null && testCount?.filtered` so a legitimate `0ms` filtered average renders the suffix.
- **Per-cell `latencyMs: 0` was treated as missing.** Same fix at the cell level: guard changed from `!output.latencyMs` to `output.latencyMs == null` so a recorded zero renders as `0ms`.
- **Tooltip / aria-label precision.** Both `exactMs` (per-cell) and `exactTotalMs` (column-aggregate) now use `maximumFractionDigits: 3` so non-integer averages and sub-ms cell measurements aren't silently rounded — added a `LATENCY_PRECISION_FORMATTER` next to the existing `WHOLE_NUMBER_FORMATTER` constant.
- **Cell now matches the footer for non-finite/negative latencies.** Dropped the early `formatted === null` return; the cell falls back to the raw ms string the same way the column-aggregate footer already does, so corrupt upstream data stays visible (and consistent across the two surfaces).
- **Softened the cached tooltip wording** from `"… — original duration; served from cache"` to `"… (cached response)"`. Some providers don't preserve `latencyMs` through the cache round-trip, so the strong "original duration" claim could be inaccurate.
- **Documented the module-scoped formatter** with a one-line `WHY` comment.

New tests added: sub-ms precision (`0.123 ms` aria-label), the `Infinity`/negative fallback (`∞ ms`, `-1,234 ms`), the exact-hour `36h` boundary in `formatDuration`, and unfiltered-total + filtered Avg Latency rendering at the column-aggregate level.

PR #9009, opened 2026-05-01
fix(db): preserve writes after lock failures and during shutdown

mldangelo-oai:
libSQL can leave a statement active after a lock error. Later writes on that connection can report success without becoming durable. This follow-up to #9486 recovers those connections and keeps database cleanup observable during shutdown.

## Changes

- Recover after every transient lock failure, including retry exhaustion, failed transaction starts, and root-handle calls inside transactions. If reconnecting or restoring connection settings fails, close the client and reject subsequent operations.
- Retry atomic statements and batches. Recover and rethrow for transaction callbacks and `executeMultiple`, whose earlier statements may already have committed.
- Share connection configuration between initialization and recovery, preserving the busy timeout across reconnects. Apply WAL tuning only when WAL is active; retain FULL synchronization when startup cannot enable WAL.
- Queue the final WAL checkpoint behind pending writes, then attempt truncation without waiting on readers. Inspect the checkpoint status and keep the logger open until database cleanup settles.

Public API signatures, configuration, schemas, dependencies, and supported platforms are unchanged. An incomplete checkpoint leaves committed data in the WAL for normal recovery.

## Validation

- Six recovery regression cases failed against the previous implementation, then passed with the fix on Linux and macOS. Two additional regressions for unavailable WAL failed before the follow-up fix and now preserve FULL synchronization through startup and lock recovery. The final focused database, shutdown, and tracing suite passes on macOS: 5 files / 116 tests.
- Reproduced the bug against main `819a3d91d` using the normal 5,000 ms timeout and a separate process holding a write lock for 6.25 seconds. Main reported two successful inserts, but an independent reopened client saw neither row. With this change, both rows persisted.
- A standalone `@libsql/client` reproduction also lost an acknowledged follow-up write; `ROLLBACK` did not repair the connection. Tested the project's installed `@libsql/client` 0.17.4 and `libsql` 0.5.29.
- TypeScript, lint/format, architecture, Knip, and the core build including postbuild assets passed. The built CLI completed an isolated echo eval with --no-cache --no-share: 1/1 successful, score 1, no error, and the checkpoint/database-close/logger-close order verified. All 43 CI checks passed on `2d1bb647f`, including the full suite and supported platforms.

The production diff adds 52 net lines over main: 48 in database handling and 4 in shutdown. The existing serialization queue and retry limits are retained; initialization and recovery share their connection configuration.


src/models/eval.ts
```diff
@@ -1216,7 +1241,7 @@ export default class Eval {
   async setResults(results: EvalResult[]) {
     this.results = results;
     if (this.persisted) {
-      const db = getDb();
+      const db = await getDb();
       await db.insert(evalResultsTable).values(results.map((r) => ({ ...r, evalId: this.id })));
```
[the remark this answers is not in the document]
mldangelo: Verified locally — Drizzle libsql auto-executes the insert when its builder is awaited (no terminal call required). Tested:

```js
await db.insert(t).values({ id: 'a' });
await db.select().from(t).all();  // [ { id: 'a' } ]
```

Also: this line predates the libsql migration (`git blame` → 2025-01-21), so it's outside the diff scope of this PR. The codebase elsewhere uses `.run()` consistently — happy to follow up with a separate style PR if the maintainer prefers.

site/docs/usage/troubleshooting.md
```diff
@@ -141,34 +141,28 @@ prompts:
...
+npm install
 ```
 
-If the project includes an `.nvmrc`, run `nvm use` first so the active Node.js version
-matches the project before rebuilding.
+If the error names a specific native dependency, rebuild that package for the active
+Node.js version:
+
+```bash
+npm rebuild better-sqlite3
```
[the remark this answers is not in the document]
mldangelo: This comment is on commit `d569908f`, which predates the cleanup. The current file (commit `2b19b866`) replaces the entire "Node.js version mismatch" section with a libsql-binding-not-found section; `npm rebuild better-sqlite3` is no longer mentioned anywhere in `troubleshooting.md`.

mldangelo:
Rebased onto current `origin/main` (was failing Style Check + conflicted) and applied a handful of targeted fixes.

## What changed in this iteration

**Merge with main:**
- Resolved 3 content conflicts:
  - `src/commands/retry.ts` — kept async `getDb()` from this branch, picked up main's new `affectedEvals` + `clearCountCache(evalId)` invalidation loop. Also addressed the Copilot review finding by ensuring the `db.delete(...)` call has a terminal `.run()`.
  - `test/models/eval.test.ts` — kept both imports (`Prompt`, `ResultFailureReason` from this branch; `updateResult`, `writeResultsToDatabase` from main).
  - `test/googleSheets.optionalDependency.test.ts` — took main's version; the branch had drifted incidentally.

**Dropped unrelated drift that had been sweeping along on this branch** (per `AGENTS.md`'s "don't let unrelated changes ride along" rule):
- `src/commands/mcp/server.ts` — **security regression**: the branch had reverted `randomUUID()` → `Math.random().toString(36).substring(2, 15)` for MCP session IDs. Restored main's `randomUUID()`.
- `src/util/azureBlob.ts` + `test/util/azureBlob.test.ts` — **security regression**: the branch had removed `sanitizeAzureBlobErrorDetail()`, which redacts SAS query strings from error messages. Restored.
- `src/googleSheets.ts` — restored `GoogleSheetAuthenticationRequiredError` and the HTML-on-CSV-export fallback chain that main added.

**Style Check fix:** `test/models/evalPerformance.test.ts` had an unused `evalId` parameter in `makeResult` (Biome `noUnusedFunctionParameters`). Dropped the unused param and updated the 4 call sites.

**Async test sweep:** Converted remaining sync `getDb()` callers in `test/commands/import.test.ts`, `test/commands/retry.test.ts`, and `test/models/eval.test.ts` (introduced by the upstream merge) to `await getDb()` and awaited the chained Drizzle terminals.

**Pulled in PR #9486's `test/util/databaseStandalone.test.ts` async conversion** — required because main added a `migration 0024 repair semantics` describe block that uses sync `getDb()` and `db.run(repairSql).changes`. Converted to `await getDb()` and `(await db.run(repairSql)).rowsAffected`.

## QA on this iteration

- `npm run tsc -- --pretty false` — clean
- `npm run l && npm run f` — clean (warnings only, no new errors)
- `npm run build` — clean
- Focused tests:
  - `npx vitest run test/util/databaseStandalone.test.ts test/commands/retry.test.ts test/models/eval.test.ts --sequence.shuffle=false` — 3 files, 147 tests pass
  - The full PR validation suite — 19 files, 284 tests pass
- End-to-end smoke: fresh `~/.promptfoo` config dir, `npm run local -- eval -c test/smoke/fixtures/configs/basic.yaml --no-cache --no-share -o eval-output.json` — 1 pass, `Hello World` output, exit 0

mldangelo:
Two correctness bugs found and fixed.

### P1: Deadlock when root `db.*` is called inside a `db.transaction(...)` callback (`485ceff33b`)

`serializeTopLevelOperations` queued every `client.execute/batch/migrate/executeMultiple` call against `operationQueue` unconditionally. If a transaction callback called any helper that issued a root `db.run/select/insert` (instead of going through the `tx` handle), the operation queued behind the outer transaction — which was itself waiting for the callback to resolve — and the promise never settled.

The wrapper already had AsyncLocalStorage machinery to detect nested `db.transaction()` calls and short-circuit them. Extended the same detection to root client methods: inside an active transaction's async chain, the wrapper now bypasses the queue. WAL mode lets SELECTs from a fresh per-call connection proceed; write attempts surface as SQLITE_BUSY (the correct loud failure for code that should be using the `tx` handle).

Reproduced as a 1-second-timeout vitest before the fix; passes after. Added a regression test in `test/database/index.test.ts`.

### P2: `getTraceSpans()` was broken after `TraceStore.getDatabase()` became async (`cf479dac80`)

The free `getTraceSpans` helper at the bottom of `src/tracing/store.ts` reached into `TraceStore` via reflection to call the now-async `getDatabase()` method, but never awaited it. `db` was a Promise, so the very next line crashed with `TypeError: db.select is not a function`. The helper had no tests, which is why this slipped through.

Replaced the 70-line reflection-based wrapper with a one-line delegation to the public `TraceStore.getSpans(traceId, options)`, which already awaits the database and runs the same filter/sanitize/depth logic. Side benefit: picks up `getSpans`'s correct use of raw (pre-sanitize) attributes for `deriveSpanKind` classification (the duplicated helper had been passing post-sanitize attributes).

Added a regression test in `test/tracing/storeRetention.test.ts`.

### Verification

- `npm run tsc -- --pretty false` — clean
- `npx vitest run` on the full PR validation slice (21 files, 428 tests) — all pass
- Both regression tests fail before the fix and pass after

mldangelo:
Three more fixes pushed addressing remaining review findings.

### C3: silent PATCH /api/eval/:id data loss (`438588dfd4`)

Same bug that PR #9486 fixed in `d396c503f5`. `updateResult` caught DB errors, logged them, returned `void`, and the route handler's `} catch { res.status(500)... }` body never ran because the inner function had already swallowed everything. Route returned `200 OK` with "Eval updated successfully" for failed updates.

- `updateResult` now `throw err` after logging
- Route handler logs `[PATCH /api/eval/:id] Failed to update eval` with `{ id, error }` and reaches the 500 path
- Regression test in `test/server/routes/eval.validation.test.ts` mocks `save()` rejection and asserts the 500 response

### I1: libsql binding handler — ESM + platform-only matching (`f4b16fe4b0`)

Two issues with `src/util/libsqlBindingErrors.ts`:
- Regex was `/@libsql\/(?<target>[a-z0-9-]+)/i` — also matched wrapper packages like `@libsql/client`/`@libsql/core`. A genuine wrapper-package-missing error (broken install) was being labeled a binding miss and the user got "file an issue with your platform" advice. Narrowed to known platform prefixes: `darwin|linux|win32|android|freebsd|netbsd|wasm32`.
- Only matched CJS `MODULE_NOT_FOUND`. The lazy `import('@libsql/client/node')` in `getDb()` is ESM and throws `ERR_MODULE_NOT_FOUND` for top-level package misses. Accept both codes.

Tests cover the ESM path and the "don't misclassify @libsql/client" case.

### I4 + I5: observability gaps (`e4a5867337`)

- `closeDb()` WAL checkpoint failure: was `logger.debug`, now `logger.warn`. A failed `PRAGMA wal_checkpoint(TRUNCATE)` before close means committed WAL pages weren't flushed; if the WAL is later truncated out of band the data is lost. Operators should see it.
- `TraceStore.deleteOldTraces` was discarding `rowsAffected` from both deletes inside the transaction. Now captures both and includes them in the log: `Pruned traces older than N days: X trace(s), Y span(s)`. If the dual-encoding `traceCreatedBefore` predicate ever silently misclassifies timestamps, retention silently doing nothing will be visible.

### Verification

- `npm run tsc -- --pretty false` clean
- Focused PR suite: 16 files, 361 tests pass
- Lint + format clean

### Remaining (not addressed in this round)

From the multi-agent review, these are still open but lower-priority:
- I2: `closeDb()` race with in-flight `getDb()`
- I3: `recordBlobReference` check-then-insert race (pre-existing, not introduced)
- I7: `provider.exists().catch(() => false)` swallows real errors
- I8: Public write helpers lack the serialization invariant docs
- I9: Retry post-success cleanup catch swallows `recalculatePromptMetrics` too coarsely
- Test coverage gaps for `not_contains` filter and `Eval.getMany()` downstream

mldangelo:
Merged the parallel PR #9486 into this branch instead of force-rebasing, so the full work history on both sides is preserved.

### What changed

- **Merge commit** `c5f84e7fbe` brings in 9486's full history including their three "address libsql review blockers" fixes (P1, P2, C3) and their post-review follow-up (libsql binding handler narrowing, ESM code acceptance, WAL durability test using `@libsql/client/node`, etc.).
- Resolved 10 file conflicts by picking the stronger version of each, favoring keeping both branches' tests where they cover different angles. Notable:
  - `src/util/libsqlBindingErrors.ts` — kept my broader platform regex (added `android|freebsd|netbsd|wasm32`) on top of 9486's `(?:darwin|linux|win32)`-only pattern, plus 9486's `Set`-based code-matching style.
  - `test/util/libsqlBindingErrors.test.ts` — kept tests from both: 9486's CJS-wrapper-package coverage AND my ESM-wrapper-package coverage.
  - `test/database.test.ts` — took 9486's improved WAL test that asserts the on-disk `-wal` sidecar exists (addresses my I2 finding about the test being self-referential).
  - `src/database/index.ts` / `src/tracing/store.ts` — kept my I4 (WAL checkpoint at `warn`) and I5 (`rowsAffected` tracking) fixes.
- **`chore(deps): drop libsql devDep`** (`b60a7474a6`) — 9486's `test/database.test.ts` switch to `@libsql/client/node` removed the only direct `libsql` import, so the explicit devDep I'd added is no longer needed. `libsql` still ships transitively via `@libsql/client`.

### Verification

- `npm run tsc -- --pretty false` — clean
- `npm run build` — clean
- `npm run depcheck` — clean (after dropping the now-unused libsql devDep)
- Full PR validation suite — 16 files, 363 tests pass

### Net result

`pr-9009` now contains every commit from both branches plus the merge and one small dep-cleanup commit on top. The diff against 9486's tip is just the four real improvements unique to this branch:

| File | Change |
|---|---|
| `src/database/index.ts` | WAL checkpoint failure logged at `warn` instead of `debug` |
| `src/tracing/store.ts` | `deleteOldTraces` captures and logs `rowsAffected` |
| `src/util/libsqlBindingErrors.ts` | Platform regex covers `android`/`freebsd`/`netbsd`/`wasm32` in addition to `darwin`/`linux`/`win32` |
| `test/util/libsqlBindingErrors.test.ts` | Adds explicit ESM-wrapper-package miss coverage to complement the CJS-wrapper test from 9486 |

Plus the test mock update and the regression-comment in `test/database/index.test.ts`.

src/database/index.ts
```diff
@@ -263,7 +263,11 @@ export async function closeDb() {
           await sqliteInstance.execute('PRAGMA wal_checkpoint(TRUNCATE)');
           logger.debug('Successfully checkpointed WAL file before closing');
         } catch (err) {
-          logger.debug(`Could not checkpoint WAL file: ${err}`);
+          // Checkpoint failure means committed WAL pages were not flushed to the
+          // main DB file. They'll be replayed on next open under normal
+          // conditions, but if the WAL is later truncated out of band the data
+          // is lost — surface this as a real warning rather than a debug log.
+          logger.warn(`Could not checkpoint WAL file before close: ${err}`);
```
jameshiester-oai: **[P2] Inspect the checkpoint status row before logging success**

This catch only covers rejected PRAGMA calls, but ordinary WAL contention is reported as a successful result row. With a second libSQL connection holding a read transaction, `PRAGMA wal_checkpoint(TRUNCATE)` resolved with `{ busy: 1, log: 4, checkpointed: 3 }`; invoking this actual `closeDb()` path then logged `Successfully checkpointed WAL file before closing` and never reached this warning. That leaves the exact incomplete-flush condition this change is meant to expose invisible. Capture the ResultSet, warn when `Number(rows[0]?.busy) !== 0` (optionally log `log`/`checkpointed`), and emit success only for `busy === 0`; add a two-client file-backed regression.

jameshiester-oai (commented):
## Review scope

- **Reviewed base:** `c54a30668ad8319d76c20ae96e6680ad6c51a2c6`
- **Reviewed head:** `1aab7e5d793b31a448d2b2be492018b560d0b5c9`
- **Scope:** full fresh `base...head` review; no earlier authenticated completion marker exists for this PR.
- **Verdict:** the trace-retention count logging and libSQL wrapper-package classification changes behave as intended, but the WAL observability change misses the normal incomplete-checkpoint result and logs a false success. One P2 finding is attached. The PR is mergeable but 166 commits behind current `main`.

## Verified findings

### P2

1. **Inspect `wal_checkpoint` status before claiming success.** When another connection holds a WAL read snapshot, `PRAGMA wal_checkpoint(TRUNCATE)` resolves normally with `{ busy: 1, log: 4, checkpointed: 3 }`; it does not throw. The new warning catch is therefore skipped and `closeDb()` emits `Successfully checkpointed WAL file before closing` even though the WAL was not fully checkpointed or truncated. The inline comment contains the direct reproduction and correction direction.

No P0, P1, or additional P2/P3 defect was validated.

## Existing feedback and CI

The only two inline review threads are resolved and outdated; they concern an old Eval terminal-call suggestion and stale `better-sqlite3` documentation, so neither applies to this eight-file head diff.

CI has three red jobs, all unrelated to the PR behavior:

- Site: 6 files / 101 tests passed.
- Web UI: 278 files / 4,816 tests passed (plus the preliminary 2-test suite).
- Node 20: 766 files / 19,666 tests passed, 1 file / 10 tests skipped.

Each job failed afterward when Codecov v11.2.7 could not import its GPG public key (`no valid OpenPGP data`, `No public key`). All other visible Linux/Windows/macOS tests, builds, style, integration, smoke, security, and CodeQL checks passed.

## Tests and end-to-end QA

- `source ~/.nvm/nvm.sh && nvm use >/dev/null && npx vitest run test/database/index.test.ts test/database.test.ts test/models/evalPerformance.test.ts test/tracing/store.test.ts test/tracing/storeRetention.test.ts test/util/libsqlBindingErrors.test.ts --sequence.shuffle=false` — exit 0; 6 files, 70 tests passed.
- `npm run tsc -- --pretty false` — exit 0.
- `npm run build` — exit 0, including TypeScript and production app builds.
- Driver-backed retention harness in an isolated `PROMPTFOO_CONFIG_DIR` — pruned 2 traces and 3 spans; retained the expected new trace/span and logged the exact real-driver counts.
- Failure/rollback harness — a trigger aborted the second delete after spans were deleted; the transaction rolled back and both trace and span remained.
- Clean-environment local eval: `npm run local -- eval -c test/smoke/fixtures/configs/basic.yaml --no-cache --no-share -o /private/tmp/pr9009-e2e/output.json` with isolated config state — exit 0; 1/1 passed, score 1, output `Hello World`, no error, and a file-backed libSQL database was created.
- Two-client WAL probe — exit 0 with `{"busy":1,"log":4,"checkpointed":3}`. A second probe invoked the actual `getDb()`/`closeDb()` path and captured only the false `Successfully checkpointed...` debug line, with no warning.
- `git diff --check` passed and the detached worktree remained clean.

The smallest robust implementation is to retain the checkpoint `ResultSet`, warn when `Number(result.rows[0]?.busy) !== 0` (optionally include `log` and `checkpointed`), and emit success only for a complete result. Add a file-backed two-client regression because the current changed test only adds explanatory comments and does not exercise either checkpoint outcome.

The new Android/FreeBSD/NetBSD/WASM regex cases are synthetic forward-compatibility tests: locked libSQL 0.5.29 ships only Darwin/Linux/Windows bindings, and the locked loader throws unsupported-system errors for NetBSD/WASM before a package require. I found no current runtime regression, but real loader-error tests should replace fabricated module errors before treating those platforms as supported remediation paths.

The PR title follows the applicable `fix(db)` convention.

## Limitations

The requested PR Review Toolkit endpoint was unavailable in this runtime, so I applied its six lenses manually (correctness, tests, comments/docs, error handling, type design, simplification) and reconciled three independent read-only reviewers. I did not run the full repository suite, test on the added non-host platforms, or intentionally remove a live WAL sidecar; all persistence and contention probes used isolated temporary directories.

mldangelo-oai:
## Final review receipt — `903e8c2c6208c98dc2b2bcb8447b6c4fa049748b`

**Disposition:** code-clean and ready once external branch-update and reviewer gates are satisfied.

I reviewed the exact live head against current `main` (`56ee73063bb5088536e4d510861651484fc49a3a`) after the final push. Dispatch head `1aab7e5d793b31a448d2b2be492018b560d0b5c9` is an ancestor; all follow-up work is additive.

### Validated and fixed

- Non-throwing `wal_checkpoint` contention no longer logs false success; `busy`, `log`, and `checkpointed` are inspected and reported with structured context.
- The final checkpoint is non-blocking relative to the existing shutdown watchdog, and logger teardown waits for the retained database-close promise so normal and delayed warnings stay visible without abandoning in-flight work.
- Trace retention reports committed trace/span deletion counts.
- Real file-backed WAL coverage runs in a short-lived subprocess, making native libSQL handle release deterministic on Windows.
- Stale platform-binding expansion and unrelated historical residue are absent from the effective merge.

### Final effective scope

- 7 files
- Production: `+44/-16`
- Tests: `+203/-1`
- No public API, schema, configuration, dependency, or platform-support changes

### Verification

- Focused shuffled DB/shutdown/tracing/retention/cleanup tests: 5 files, 81 tests passed
- Hygiene/shutdown/WAL/tracing slice: 4 files, 161 tests passed
- Broader DB/models/server and tracing/assertion/migration slices: 49 files, 1,134 tests passed
- TypeScript, architecture, Knip, targeted Biome, and full production/app build passed
- Real WAL success, direct contention, and delayed serialized contention verified; all committed rows persisted
- Real retention removed 2 traces / 3 spans; forced failure rolled back both tables
- Isolated no-cache CLI eval passed 1/1 with inspected JSON and zero-byte final WAL
- Exact-head CI: 43/43 contexts and 13/13 required checks passed: https://github.com/promptfoo/promptfoo/actions/runs/28231917137

All review threads are resolved and no current actionable feedback remains. The PR is mergeable, but GitHub still reports `BEHIND` and `REVIEW_REQUIRED`; no merge or base synchronization was performed.

src/database/index.ts
```diff
@@ -260,10 +260,24 @@ export async function closeDb() {
       // Attempt to checkpoint WAL file before closing
       if (!sqliteInstanceIsTesting && !getEnvBool('PROMPTFOO_DISABLE_WAL_MODE', false)) {
         try {
-          await sqliteInstance.execute('PRAGMA wal_checkpoint(TRUNCATE)');
-          logger.debug('Successfully checkpointed WAL file before closing');
-        } catch (err) {
-          logger.debug(`Could not checkpoint WAL file: ${err}`);
+          // Do not let the normal 5s busy timeout outlive the 3s graceful-shutdown watchdog.
+          // The client closes immediately after this best-effort checkpoint, so no restoration is needed.
+          await sqliteInstance.execute('PRAGMA busy_timeout = 0');
```
jameshiester-oai: **[P1] Preserve acknowledged writes before zeroing the client's busy timeout**

This can discard a write that has already reported success. In an isolated exact-head repro, another process held a write transaction and rolled it back after 300 ms while this process queued an insert and `closeDb()`. The insert fulfilled with `rowsAffected=1`; after this sets the timeout to zero, the queued checkpoint exhausted with `SQLITE_LOCKED`, the catch continued to `sqliteInstance.close()`, and an independent reopened client found no rows. The identical schedule without this zero-timeout step completed its checkpoint in 362 ms and preserved `{ id: 2 }`. Keep a bounded shutdown without zeroing the live client before queued writes are durably settled (or use another checkpoint strategy), and add a cross-process writer-contention regression that asserts an acknowledged write survives reopen.

jameshiester-oai (commented):
## Review scope

- **Reviewed base:** `c54a30668ad8319d76c20ae96e6680ad6c51a2c6`
- **Reviewed head:** `903e8c2c6208c98dc2b2bcb8447b6c4fa049748b`
- **Scope:** full current `base...head` review plus an incremental review from the prior authenticated head `1aab7e5d793b31a448d2b2be492018b560d0b5c9`; the prior head is an ancestor.
- **Verdict:** the prior incomplete-checkpoint and logger-order findings are fixed, but the new zero-timeout checkpoint path can discard an acknowledged write during cross-process writer contention. One P1 finding is attached.
- **CI:** all 39 currently visible checks pass; GitHub reports MERGEABLE/BEHIND.

## Verified findings

### P1

1. **Do not zero the live client's busy timeout before its queued writes are durably settled.** With a second libSQL process holding and then rolling back a write transaction, this process queued an insert and `closeDb()`. The insert fulfilled with `rowsAffected=1`; the new zero-timeout checkpoint then exhausted with `SQLITE_LOCKED`, `closeDb()` caught the error and closed the client, and an independent client reopened an empty table. The attached inline comment contains the exact trigger, control, impact, and correction direction.

No P0, P2, or P3 defect was validated.

## Existing feedback

All three existing inline threads are resolved and outdated. The earlier exact-head checkpoint-status finding is fixed: `busy !== 0` now warns and success is emitted only for `busy === 0`. The earlier shutdown-order concern is also fixed: the same database-close promise is awaited before `closeLogger()`. This new finding is a distinct writer-contention/durability path and does not duplicate those threads.

## Tests and end-to-end QA

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/database/index.test.ts test/database.test.ts test/main.test.ts test/tracing/store.test.ts test/tracing/storeRetention.test.ts test/util/utils.test.ts --sequence.shuffle=false` — exit 0; 6 files / 90 tests passed.
- `npm run tsc -- --pretty false` — exit 0.
- `npm run architecture:check` — exit 0.
- `npx @biomejs/biome check <9 changed files>` — exit 0; no fixes applied.
- `npm run build` — exit 0, including core/declarations and the production app build.
- Isolated local CLI eval with `--no-cache --no-share` and exported JSON — exit 0; 1/1 passed, score 1, output `Hello World`, no error, and final WAL size 0. The real debug log recorded checkpoint success, database close, then logger close in that order.
- Exact-head cross-process writer probe — the queued insert fulfilled with `rowsAffected=1`, checkpoint warned with `SQLITE_LOCKED`, and a fresh client observed `[]` after close.
- Control using the same writer schedule without setting `busy_timeout = 0` — completed in 362 ms with `{ busy: 0, log: 0, checkpointed: 0 }`; the fresh client observed `[{ id: 2 }]`.
- `git diff --check` passed; the detached worktree remained clean.

## Simplification and test coverage

The current file-backed fixture covers an uncontended checkpoint and a held reader, but not an in-flight write delayed by another process. Add a cross-process writer-contention regression that requires an acknowledged queued write to survive reopen. The correction should retain a bounded shutdown while allowing the live client to finish durable write settlement before it is closed; simply forcing its connection timeout to zero is not safe on this path.

## Limitations

The requested PR Review Toolkit command was unavailable in this runtime (exit 127), so I applied its correctness, tests, comments/docs, error-handling, type-design, and simplification lenses manually and reconciled three independent read-only specialists. I did not run the full repository suite, test the reproduced contention path on Windows/Linux, or use credentials, live model providers, the user's Promptfoo database/cache, or secret-bearing environment.

src/database/index.ts
mldangelo: Root-caused and fixed in 873e45c21 — your repro was real but the mechanism is deeper than the zeroed timeout. The write loss is a libsql node-binding bug: after a write fails with SQLITE_BUSY/LOCKED, the binding leaks the active prepared statement, pinning that connection's implicit transaction open. Every later write on the wedged connection reports success (rowsAffected=1) but never writes a WAL frame (verified: -wal stays byte-identical), is invisible to other connections, makes wal_checkpoint fail SQLITE_LOCKED regardless of busy_timeout, and is discarded at close. Skipping the checkpoint or skipping close() does not save the row; only reconnecting heals the connection (ROLLBACK/SELECT/BEGIN+COMMIT all fail). The fix: withTransientLockRetry now swaps in a fresh connection (carrying per-connection pragmas) before each lock retry, and surfaces the original lock error if recovery fails instead of acknowledging a doomed write. closeDb keeps a bounded 1500ms checkpoint timeout (not 0), safe now that committed data always reaches the WAL. Added the writer-contention probe mode you asked for: it reproduces rowCount=1 without the fix and passes 5/5 with it; reader-mode expectations unchanged. Minimal upstream repro (zero promptfoo code) is in hand — will file against tursodatabase/libsql.

src/database/index.ts
```diff
@@ -139,11 +142,67 @@ async function withTransientLockRetry<T>(operation: () => Promise<T>): Promise<T
       await sleep(
         Math.min(TRANSIENT_LOCK_RETRY_BASE_MS * 2 ** (attempt - 1), TRANSIENT_LOCK_RETRY_MAX_MS),
       );
+      if (onBeforeRetry) {
```
jameshiester-oai, line 145: **[P1] Heal the client after terminal retry exhaustion**

This recovery hook runs only when another retry will be attempted. On the tenth transient lock failure, the guard at lines 139–140 throws first and leaves the known-poisoned libSQL connection reusable. A file-backed repro held `BEGIN IMMEDIATE` through all ten attempts: the blocked insert rejected; after `ROLLBACK`, the next `db.run()` returned `rowsAffected=1`; after close/reopen, only the baseline id remained and the acknowledged follow-up row was absent. Recover on every transient failure before returning control—including exhaustion—and make the handle non-reusable if recovery itself fails. Add a regression that holds the lock through exhaustion and verifies a later acknowledged write after reopen.

src/database/index.ts
```diff
@@ -176,8 +235,12 @@ function serializeTopLevelOperations(client: Client, db: Drizzle): Drizzle {
       // nothing was applied. libsql runs interactive transactions on their own
       // connection, so a prior writer's table lock can briefly outlive the JS
       // promise that settled it; retry rides through that window without
-      // weakening isolation (reads still observe only committed rows).
-      return runSerialized(() => withTransientLockRetry(() => method(...args)));
+      // weakening isolation (reads still observe only committed rows). The retry
+      // must run on a fresh connection (see recoverConnectionForRetry above) or
+      // its success is a lie.
+      return runSerialized(() =>
```
jameshiester-oai, line 241: **[P1] Apply recovery to the transaction bypasses too**

The new reconnect is wired only to serialized statement methods. A lock-failed `db.transaction()` `BEGIN` at lines 255–265 bypasses it; after that rejection, a later insert returned `rowsAffected=1` but independent reads before close and after reopen were empty. The active-transaction branch at lines 231–232 has the same problem: a root write contending with its outer transaction rejected, the outer transaction committed id 1, and a later root insert reported success while reopen still contained only id 1. Catch transient failures on both bypasses, reconnect/reapply pragmas, then rethrow the original error; do not rerun the transaction callback or the root write while its outer lock is held.

src/database/index.ts
```diff
@@ -176,8 +235,12 @@ function serializeTopLevelOperations(client: Client, db: Drizzle): Drizzle {
...
+        withTransientLockRetry(() => method(...args), recoverConnectionForRetry),
```
jameshiester-oai, line 242: **[P2] Do not replay non-atomic `executeMultiple` scripts**

This wrapper also covers `client.executeMultiple` at line 253, but libSQL explicitly does not wrap that API in a transaction. With an attached database locked for about 300 ms, each failed attempt had already committed the script's main-database insert; reconnect then replayed the whole script. Current head resolved successfully with main ids `[1,2,3,4]` and one attached row. The reviewed base instead rejected after one partial main write (`attached_db is already in use`), so this reconnect converts a loud partial failure into silent successful duplication. For `executeMultiple`, heal the client and rethrow rather than replaying the script unless atomicity/idempotency is established; add an attached-database contention regression.

src/database/index.ts
```diff
@@ -260,10 +327,31 @@ export async function closeDb() {
...
-        } catch (err) {
-          logger.debug(`Could not checkpoint WAL file: ${err}`);
+          // Bound the checkpoint below the 3s graceful-shutdown watchdog, but never zero
+          // it: a zeroed timeout makes the checkpoint fail instantly under transient
+          // contention. 1500ms rides through sub-second contention while leaving
+          // watchdog headroom; the client closes right after this best-effort
+          // checkpoint, so no restoration is needed. Both statements ride the
+          // serialized queue, so the checkpoint runs only after every previously
+          // queued write has settled.
+          await sqliteInstance.execute('PRAGMA busy_timeout = 1500');
```
jameshiester-oai, line 337: **[P2] Keep the real checkpoint duration below the shutdown watchdog**

The exact-head macOS/Node 22 job failed this new held-reader regression at `elapsedMs=3359` (expected `<2500`), already beyond `shutdownGracefully`'s 3-second force-exit watchdog. All other tests in that job passed, and local supported-Node runs were about 2.0 seconds, so this is a platform/load-sensitive failure in the changed behavior rather than unrelated CI noise. The native libSQL checkpoint blocks the JS thread, so the watchdog timer cannot enforce a wall-clock bound while this call is waiting. Use a checkpoint mechanism/budget with a cross-platform hard margin below 3 seconds; loosening the assertion would preserve the shutdown race.

jameshiester-oai (commented):
## Review scope

- **Reviewed base:** `ede330200ede3ef6b15a3f45420bf11fd6616e50`
- **Reviewed head:** `873e45c2124c5a545014b47eeb2204a541f31706`
- **Scope:** full current `base...head` review plus an incremental review from the prior authenticated head `903e8c2c6208c98dc2b2bcb8447b6c4fa049748b`; that head is an ancestor.
- **Verdict:** the new reconnect fixes the previously reported short-lived serialized-writer reproduction, but recovery remains incomplete and can still lose later acknowledged writes. I also found a non-atomic replay regression and a platform-specific shutdown-bound failure. Two P1 and two P2 findings are attached.
- **CI:** every currently visible check passes except `Test on Node 22.22 and macOS-latest`, whose only failure is the new reader-contention timing assertion (`elapsedMs=3359`, expected `<2500`).

## Verified findings

### P1

1. **Recover the client after terminal retry exhaustion.** The tenth transient lock failure throws before the new recovery hook runs. In a file-backed repro, a writer held the lock through all attempts; the blocked write rejected, but after contention cleared the next insert returned `rowsAffected=1` and was absent after close/reopen.
2. **Recover transaction-path lock failures before the root client is reused.** Both a lock-failed transaction `BEGIN` and a root-handle write inside an active transaction bypass the new recovery hook. Each path deterministically left the root connection accepting a later insert that never reached the WAL.

### P2

1. **Do not replay `executeMultiple` after a partial commit.** libSQL documents this API as sequential and not implicitly transactional. Under an attached-database lock, current head retried the whole script and resolved with four copies in the main database and one in the attached database. The base failed loudly after one partial main write, so reconnect changes this into a successful duplicate-write outcome.
2. **Use a shutdown bound that actually stays below the watchdog.** Exact-head macOS CI measured 3359 ms in the new held-reader checkpoint test, already beyond the 3-second force-exit watchdog. The native checkpoint call blocks the JS thread, so the watchdog timer cannot preempt it.

No P0 defect was validated.

## Existing feedback

The earlier checkpoint-status and logger-order findings remain fixed. The prior P1 writer-contention thread is unresolved/outdated and now has an author reply: its exact short-contention probe passes on this head (`rowCount=2`), but the terminal-exhaustion and transaction-path reproductions above show that the same acknowledged-write-loss class remains on uncovered branches. The two old Eval/docs threads are resolved and outdated.

## Tests and end-to-end QA

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/database/index.test.ts test/database.test.ts test/main.test.ts test/tracing/store.test.ts test/tracing/storeRetention.test.ts --sequence.shuffle=false` — exit 0; 5 files / 86 tests passed.
- `nvm use 22.22.2 && npx vitest run test/database/index.test.ts --sequence.shuffle=false -t "warns when a file-backed WAL checkpoint is incomplete"` — exit 0 locally; 1 passed / 21 skipped, about 2.0 seconds in the test. Exact-head macOS CI failed the same test at 3.359 seconds.
- `npm run tsc`, `npm run architecture:check`, targeted `npx biome check` over all seven changed files, and `npm run build` — all exit 0.
- Direct file-backed WAL fixture modes — no contention: `busy=0`, `rowCount=1`; held reader: `busy=1`, `log=4`, `checkpointed=3`, `rowCount=2`; short writer: `insertAcknowledged=true`, `busy=0`, `rowCount=2`.
- `npm run local -- eval -c /private/tmp/pr9009-e2e.yaml --no-cache -o /private/tmp/pr9009-e2e-output.json` in an isolated `PROMPTFOO_CONFIG_DIR` — exit 0; 1/1 passed, score 1, no errors, echo output `Hello world`.
- `npx tsx /private/tmp/pr9009-lock-exhaustion-repro.ts` — blocked write rejected; follow-up returned one affected row; reopen contained only baseline id 1.
- `node --import tsx /private/tmp/pr9009-transaction-lock-repro.ts` plus a separate active-transaction/root-handle variant — later inserts returned one affected row but independent reads before close and after reopen omitted them.
- `NODE_PATH=$PWD/node_modules npx tsx /private/tmp/pr9009-execute-multiple-repro2.ts` — current head resolved with `mainCount=4`, `attachedCount=1`; the same harness on the reviewed base rejected with `database attached_db is already in use`, `mainCount=1`, `attachedCount=0`.
- `git diff --check` passed and the detached worktree is clean.

The recovery invariant is simpler if every transient failure heals (or permanently invalidates) the reusable client before control returns, rather than tying recovery only to the existence of another retry. Apply that invariant at statement and transaction boundaries, while treating non-atomic `executeMultiple` separately: heal and rethrow rather than replay. Add file-backed regressions for retry exhaustion, failed transaction `BEGIN`, root writes inside an active transaction, and partial `executeMultiple` execution. The current writer probe releases after 300 ms and covers only eventual success.

The PR body is stale: it still describes `busy_timeout=0` and a fully green exact-head CI run. The nearby writer-test comment also attributes the loss to the zero timeout even though the current root-cause explanation is the wedged libSQL statement.

## Limitations

The requested PR Review Toolkit command was unavailable in this runtime (exit 127), so I applied its correctness, tests, comments/docs, error-handling, type-design, and simplification lenses manually and reconciled three independent read-only specialists. I did not run the full repository suite or reproduce the 3.359-second macOS timing locally; the exact-head CI log is the evidence for that platform result. All local persistence probes used isolated temporary directories with no credentials, user database, or cache.

jameshiester-oai (commented):
Reviewed base `ede330200ede3ef6b15a3f45420bf11fd6616e50` through head `f20515ca5a1cbeb957e86e1fba7055b7f6c7de62`.

Scope: incremental from the previously reviewed ancestor `873e45c2124c5a545014b47eeb2204a541f31706`, plus the complete current base...head PR for interaction and regression context. The sole incremental commit changes only `test/database/index.test.ts`, raising two WAL-close timing assertions from 2.5 seconds to 4.5 seconds; production code is unchanged.

## Verdict and CI

No distinct new inline finding. The exact-head check rollup is currently 39/39 passing, but all four existing unresolved findings remain source-valid. In particular, the green timing check now results from accepting a duration above the production shutdown watchdog rather than restoring watchdog headroom.

## Existing verified findings (not duplicated)

### P1

- [Terminal retry exhaustion can leave the shared libSQL client poisoned](https://github.com/promptfoo/promptfoo/pull/9009#discussion_r3515177658). The attempt-10 guard still throws before the recovery hook, so later writes can report success yet disappear on reopen.
- [Transaction paths still bypass connection recovery](https://github.com/promptfoo/promptfoo/pull/9009#discussion_r3515177661). A failed transaction `BEGIN` and root operations inside an active transaction can poison the reusable client with the same false-acknowledgment outcome.

### P2

- [The generic retry wrapper can replay partially committed `executeMultiple` scripts](https://github.com/promptfoo/promptfoo/pull/9009#discussion_r3515177665), duplicating earlier statements when a later statement encounters transient contention.
- [The native checkpoint can outlive the three-second shutdown watchdog](https://github.com/promptfoo/promptfoo/pull/9009#discussion_r3515177668). The prior macOS run measured 3.359 seconds; this head changes the test to accept 4.5 seconds while `shutdownGracefully` still arms a 3-second force-exit timer. No production mechanism changed.

The older checkpoint-status thread is fixed and outdated. The two original Eval/docs threads are resolved and outdated.

## Tests and end-to-end QA

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/database/index.test.ts test/database.test.ts test/main.test.ts test/tracing/store.test.ts test/tracing/storeRetention.test.ts --sequence.shuffle=false` — exit 0; 5 files / 86 tests passed.
- Three direct held-reader executions of `test/database/fixtures/walCheckpointProbe.ts` — exit 0; `busy=1`, `log=4`, `checkpointed=3`, `rowCount=2`; elapsed 1,579–1,585 ms locally.
- `npm run tsc -- --pretty false` — exit 0.
- `npm run architecture:check` — exit 0 at the existing 13-back-edge baseline.
- `npm run build` — exit 0; core, declarations, production Web UI, and postbuild completed.
- Isolated built local CLI eval with the smoke echo fixture, `--no-cache --no-share`, and exported JSON — exit 0; 1/1 passed, score 1, zero errors, output `Hello World`.
- `git diff --check` passed; detached worktree remained clean.
- Exact-head GitHub checks: 39/39 passing.

Use one recovery invariant: every transient statement/transaction failure must heal or permanently invalidate the reusable client before control returns. Keep non-atomic `executeMultiple` outside whole-script replay. For shutdown, exercise the real `shutdownGracefully` subprocess under a held reader and assert database/logger completion before the three-second watchdog across supported platforms; a 4.5-second unit threshold does not verify that contract.

## Limitations

The requested PR Review Toolkit command is unavailable in this runtime (exit 127), so I applied its correctness, tests, comments/docs, error-handling, type-design, and simplification lenses manually and reconciled three independent read-only specialists. I did not reproduce the slow macOS timing locally or rerun the three prior durability/partial-replay harnesses; the incremental commit changes only timing assertions, and the affected production paths are byte-for-byte unchanged from the reviewed ancestor. No credentials, external providers, user database/cache, merge, or rebase were used.

mldangelo-oai:
@codex review

src/database/index.ts
mldangelo-oai: Addressed in [76e6d7dca8](https://github.com/promptfoo/promptfoo/commit/76e6d7dca8096a49cdd5e7ce5e5dc5a28fd8a0eb). The writer-contention regression now holds the competing lock in a separate process and forces the retry path while an insert and closeDb() are queued together. The insert is acknowledged and an independent reopened client sees both rows. Every transient failure now recovers the connection before control returns; the final zero-timeout checkpoint runs only after those queued writes settle.

src/database/index.ts
mldangelo-oai: Addressed in [76e6d7dca8](https://github.com/promptfoo/promptfoo/commit/76e6d7dca8096a49cdd5e7ce5e5dc5a28fd8a0eb). Recovery now runs before the retry-limit check, including the final failed attempt. If reconnecting or restoring PRAGMAs fails, the client is closed and later statements and transactions are rejected. File-backed regressions cover exhaustion followed by a durable write, reconnect failure, and configuration failure. All three failed on the previous implementation and pass with this change.

src/database/index.ts
mldangelo-oai: Addressed in [76e6d7dca8](https://github.com/promptfoo/promptfoo/commit/76e6d7dca8096a49cdd5e7ce5e5dc5a28fd8a0eb). Failed transaction starts and root-handle calls inside an active transaction use the same recovery path and rethrow without retrying. Tests verify that a failed BEGIN never invokes its callback, the outer transaction callback runs once, and later acknowledged root writes are visible to an independent reader before and after close. Both regressions failed before the fix and now pass.

src/database/index.ts
mldangelo-oai: Addressed in [76e6d7dca8](https://github.com/promptfoo/promptfoo/commit/76e6d7dca8096a49cdd5e7ce5e5dc5a28fd8a0eb). executeMultiple now recovers and rethrows after a lock failure without replaying the script. The attached-database regression verifies exactly one committed prefix row, no attached row, and a durable later write. The old implementation replayed the prefix; the regression now passes.

src/database/index.ts
mldangelo-oai: Addressed in [76e6d7dca8](https://github.com/promptfoo/promptfoo/commit/76e6d7dca8096a49cdd5e7ce5e5dc5a28fd8a0eb). The final queued checkpoint uses busy_timeout=0, so it does not spend the shutdown budget waiting on a reader. The held-reader assertion is back below 2.5 seconds, and a new subprocess test calls the actual shutdownGracefully() and measures through process exit. It also verifies that the checkpoint warning appears before logger teardown and that committed rows survive reopen. Cross-process writer coverage verifies the queued-write recovery path separately.

src/database/index.ts
```diff
@@ -148,21 +148,25 @@ export function getDbSignalPath() {
...
+async function configureConnection(
+  execute: Client['execute'],
+  busyTimeoutMs: number,
+  skipWalMode: boolean,
+): Promise<void> {
+  await execute('PRAGMA foreign_keys = ON');
+  await execute(`PRAGMA busy_timeout = ${busyTimeoutMs}`);
+  if (!skipWalMode && !getEnvBool('PROMPTFOO_DISABLE_WAL_MODE', false)) {
+    await execute('PRAGMA wal_autocheckpoint = 1000');
+    await execute('PRAGMA synchronous = NORMAL');
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in cf714769f. Connection setup now checks the actual journal mode before applying WAL-specific tuning. Startup can attempt WAL; recovery preserves the existing mode and propagates configuration failures so the client is closed.

Added file-backed regressions for both a WAL setup exception and a refused mode change. Both failed on 76e6d7dca (synchronous=1 in DELETE mode); they now retain synchronous=2 (FULL) at startup and after lock recovery, and subsequent writes persist before close and after reopen. All 116 focused database/shutdown/tracing tests, TypeScript, lint/format, and the rebuilt CLI echo eval pass locally. CI is running on the new commit.

test/database/fixtures/walCheckpointProbe.ts
```diff
@@ -0,0 +1,125 @@
...
+      writer.send('release');
+    }
+
+    const startedAt = Date.now();
+    const insertPromise = writer
+      ? db.run('INSERT INTO wal_checkpoint_test DEFAULT VALUES').then(() => {
+          insertAcknowledged = true;
+        })
+      : Promise.resolve();
+    await Promise.all([insertPromise, close()]);
```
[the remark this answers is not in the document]
mldangelo-oai: Checked against the current file-backed driver: this race does not occur here. `Sqlite3Client.execute()` calls synchronous `executeStmt()` directly, and client creation/close are synchronous too. In shutdown mode there is no writer process to wait for. The remaining awaits continue through already-settled promises, so the 100 ms timer cannot run before the exit listener is registered, even when the native calls block on slow I/O.

I verified this by inserting a 300 ms synchronous pause immediately before the verifier is opened, while leaving the exit listener in its current position. The measured pause was 305 ms; the process still emitted the complete result marker at exit (313 ms total), with `rowCount=2` and `isDbOpen=false`. An asynchronous verifier would need different ordering, but this fixture uses synchronous native I/O. Keeping the fixture unchanged.

PR #9023, opened 2026-05-01
chore(providers): refresh alibaba model ids

ianw-oai:


mldangelo-oai (approved):
LGTM. Fresh models, fresher face card.

Issue #6730, opened 2025-12-17
Feature Request: Add action buttons (or links) to the red team target type selector description


Ebonsignori:
The description for the red team target type selector reads,

"Select the type that best matches your target. Don't see what you need? Try 'Custom Target' to access [more providers](https://www.promptfoo.dev/docs/providers/). You can also create your own using [Python](https://www.promptfoo.dev/docs/providers/python/), [JavaScript](https://www.promptfoo.dev/docs/providers/custom-api/), or [shell scripts](https://www.promptfoo.dev/docs/providers/custom-script/)."

 The text, `Custom Target` could be converted to button (to differentiate it from a link) that selects the "Custom Provider" target type, 

<img width="1125" height="253" alt="Image" src="https://github.com/user-attachments/assets/e613fe56-0767-49f1-990d-27b2d62f40ac" />

The copy may be out of date between the description name (Custom Target) and the provider name (Custom Provider).

The same could be done for the custom language and shell scripts in place of links, since the ❓ icon on those target types already links to documentation.

singhvishalkr:
Hi, picked this up from the `good-first-issue` shelf and wanted to flag that the UI this refers to has already been refactored on `main` (currently at `73533f4`).

The description paragraph this issue quotes — *"Select the type that best matches your target. Don't see what you need? Try 'Custom Target'..."* — is no longer present anywhere in the codebase (`gh search code --repo promptfoo/promptfoo "Don't see what you need"` and `"Select the type that best matches your target"` both return zero hits).

The current `ProviderTypeSelector` ([`src/app/src/pages/redteam/setup/components/Targets/ProviderTypeSelector.tsx`](https://github.com/promptfoo/promptfoo/blob/main/src/app/src/pages/redteam/setup/components/Targets/ProviderTypeSelector.tsx)) is a chip/search-based card list. `Custom Provider` is now a first-class clickable entry under the "My Application" group, tagged `last: true`:

```tsx
{
  value: 'custom',
  label: 'Custom Provider',
  description: 'Other custom providers and implementations',
  tag: 'app',
  last: true,
},
```

So the one-click-to-select-Custom-Provider behaviour this issue originally requested is effectively shipped — just via a card rather than an inline button. The "copy out of date between `Custom Target` vs. `Custom Provider`" note in the issue is also moot for the same reason.

[...]

PR #9031, opened 2026-05-01
chore(redteam): prefer target copy in setup

singhvishalkr:
## Summary

- keep red-team user-facing copy centered on `target`
- reserve `provider` for the underlying integration mechanism and provider-doc links
- align the generic custom-target screen, setup copy, and related docs with that split

## Validation

- `npm run test:app -- src/pages/redteam/setup/components/Targets/index.test.tsx src/pages/redteam/setup/components/Targets/ProviderEditor.test.tsx src/pages/redteam/setup/components/Targets/ProviderTypeSelector.test.tsx src/pages/redteam/setup/components/Targets/TargetConfiguration.test.tsx --run`

Follow-up to #6730.

### DIFF
--- a/site/docs/guides/evaluate-crewai.md
+++ b/site/docs/guides/evaluate-crewai.md
@@ -477,7 +477,7 @@ Red teaming will stress-test your CrewAI setup, checking for vulnerabilities, bi
 Under **Target Type**, select:
 
 ```
-Custom Target
+Custom Provider
 ```
 
 Under Target Name, enter something meaningful like:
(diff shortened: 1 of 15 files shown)


mldangelo-oai:
Thanks for chasing this down, @singhvishalkr. You were right that there was still a small bit of copy drift worth cleaning up here.

I pushed a small maintainer follow-up on top of your change to align the still-visible generic configuration heading/link and the two guides that still told users to choose the old label, then retitled the PR to match the current redteam conventions. I appreciate you doing the archaeology before writing code on the older issue.

mldangelo-oai:
One more small refinement after thinking through the UX model: I pushed a follow-up that keeps `target` as the red-team user-facing noun and uses `provider` only where we mean the underlying integration mechanism. That keeps the setup flow aligned with the rest of the red-team experience while still pointing users to provider docs where that is the right technical concept.

singhvishalkr:
CI on the latest head is still green. This is ready whenever you have bandwidth for a docs/UI wording review.

PR #9040, opened 2026-05-01
chore(redteam): reorder coding agent security category

ianw-oai:


mldangelo-oai (approved):
Approved. Sharp category order, sharper jawline energy.

PR #9044, opened 2026-05-02
fix(providers): harden coding-agent tool policies

mldangelo-oai:
## Problem

Coding-agent providers exposed inconsistent tool-policy behavior at trust boundaries. Malformed or conflicting settings could widen authority, mock mode could leave executable callbacks or capability paths active, resumed sessions could retain stale policy, and response/thread caches could cross authority or credential boundaries.

## Implementation

- Make Claude Agent SDK tool availability and permission handling fail closed. `ask_user_question` now works in `dontAsk` mode through a scoped `PreToolUse` automation hook while explicit tool exclusions, disallowed tools, and user hook denials remain authoritative.
- Reject Claude `extra_args` that can override structured runtime policy, reject malformed MCP server shapes before launch, and conservatively bypass persistent caching for mutable session/settings/environment state and credential-bearing MCP arguments.
- Make OpenAI Agents `executeTools: false | "mock"` recursively wrap direct handoff agents while rejecting capability-bearing or callback-driven shapes; function execution, policy callbacks, guardrails, and timeout callbacks are replaced with deterministic mocks.
- Key and serialize Codex SDK/app-server thread reuse by physical thread identity plus authority-bearing options. Stale explicit-thread unsubscribe is best-effort, closed/timeout connections reconnect within bounded retry logic, and queued callers refresh connection state at the serialized boundary.
- Give OpenCode a wildcard-deny baseline, normalize edit aliases, preserve native rule precedence, validate v1/v2 policy contracts, serialize persistent sessions, prevent initialization races, bypass unsafe cache cases including multi-argument MCP commands, and atomically reapply the full boolean tool policy on v2 session resumes.
- Harden MCP transforms and prototype-shaped input handling, and document the resulting configuration and external-server trust boundaries.

## Compatibility and risks

- Valid existing provider configurations and public source APIs remain supported. OpenAI Agents `executeTools: false` remains an alias for deterministic mock behavior; explicit `null` and falsy mock values are preserved.
- Existing Claude allowlists remain supported. Explicit `tools` controls SDK tool availability; unsafe policy-overriding `extra_args` and malformed authority settings now fail before execution. Stateful, mutable, environment-sensitive, or credential-bearing cases may skip persistent response caching rather than risk cross-boundary reuse.
- `ask_user_question` automation in `dontAsk` applies only when configured and not explicitly excluded or denied. Default permission mode retains the existing `canUseTool` path.
- OpenCode v1 keeps atomic boolean `tools` support. OpenCode v2 resumes atomically replace stored permission rules with Promptfoo's safe defaults plus configured boolean tools; non-empty `permission` configs, including ask and pattern rules, still require a new session.
- A shared external OpenCode server remains a trust boundary: prior in-memory `always allow` approvals may supersede configured patterns. Use a fresh isolated server when policy isolation is required.

## Validation

- `npx vitest run test/providers/claude-agent-sdk.test.ts test/providers/mcp/transform.test.ts test/providers/openai-codex-app-server.test.ts test/providers/openai-codex-sdk.test.ts test/providers/openai/agents.test.ts test/providers/opencode-sdk.test.ts --sequence.shuffle=false` — 6 files, 711 tests passed.
- `npm run tsc -- --pretty false` — passed against Claude Agent SDK 0.3.177, OpenAI Agents 0.11.3, and OpenCode SDK 1.17.4.
- `npm run l`, `npm run f`, `npm run architecture:check`, and `git diff --check` — passed; only the three existing Claude provider complexity warnings remain and the architecture baseline stayed at 13 back-edges / 182 imports.
- `SKIP_OG_GENERATION=true npm run build --prefix site` — passed and generated static files.
- Exported-JSON `npm run local -- eval ... --no-cache -o ...` workflows verified stale app-server cleanup timeout recovery across two processes and two successful turns; exact Claude Code 2.1.177 `dontAsk` allow, configured deny, explicit deny, tool exclusion, default-mode compatibility, conflicting `extra_args`, and malformed MCP behavior; and OpenCode v2 resumed default-deny, read-only, explicit-allow, and granular-permission rejection behavior through the real SDK HTTP serializer. Success, score, error, provider output, tool results, permission denials, sandbox state, thread IDs, process counts, and captured OpenCode request bodies were inspected.
- Exact-head CI on `7142e600f23263b6b8087be012240fad48e37ab5` — 39/39 effective checks passed, including required checks, Node 20/22/24/26 across Linux/macOS/Windows, CodeQL/code scan, style/architecture, docs/site/web UI, integration/smoke, and production/staging redteam jobs.


src/providers/mcp/transform.ts
```diff
@@ -118,7 +118,17 @@ export function transformMCPToolsToGoogle(tools: MCPTool[]): GoogleTool[] {
 export async function transformMCPConfigToClaudeCode(
   config: MCPConfig,
 ): Promise<Record<string, ClaudeCodeMcpServerConfig>> {
-  const serverConfigs = config.servers ?? [];
+  if (config.enabled === false) {
+    return {};
+  }
+
+  if (config.tools?.length || config.exclude_tools?.length) {
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `9e8752ae7`: Claude Agent SDK MCP conversion now rejects any configured `tools` or `exclude_tools` field, including empty arrays, instead of silently converting an empty allowlist into unrestricted access. The regression test covers both `tools: []` and `exclude_tools: []`, and an isolated CLI eval now fails closed before a configured MCP executable can launch.

mldangelo-oai:
Final review on exact head `7142e600f23263b6b8087be012240fad48e37ab5`: **FIXED_AND_READY**.

- Continued from `025f780779c18b0788c59958daea990d463179ea` with two additive commits: `1cf4779e7` (`fix(providers): close follow-up policy gaps`) and `7142e600f` (`fix(providers): reapply opencode resume policy`). No history was rewritten.
- Validated and fixed both exact-head P2s: stale explicit-thread unsubscribe failure/timeout is now best-effort with bounded reconnect and serialized connection refresh; Claude `ask_user_question` in `dontAsk` now uses a scoped `PreToolUse` automation hook while explicit exclusions and denials remain authoritative.
- Fresh adversarial review also closed policy-overriding Claude `extra_args`, malformed MCP shape handling, queued app-server closed-connection reuse, unsafe persistent-cache cases for mutable/session/environment state or credential-bearing/multi-argument MCP commands, and OpenCode v2 resumed-session default/precedence drift. Final delegated sectors were clean; the requested exact-head Codex re-review completed without a new finding.
- Local QA: 6 focused provider suites / 711 tests passed; root TypeScript, changed-file lint/format, architecture checks, docs build, and diff checks passed. Exact Claude Code 2.1.177 no-cache exported-JSON evals covered `dontAsk` allow/deny/exclusion, default-mode compatibility, malformed MCP, and conflicting `extra_args`; app-server QA recovered from stale unsubscribe timeout across two processes; real OpenCode SDK serialization reapplied default-deny, read-only, and explicit-allow maps while granular permission resumes failed before a model request.
- Exact-head CI: 39/39 effective checks passed, including all required checks, Node 20/22/24/26 across Linux/macOS/Windows, CodeQL/code scan, style/architecture, docs/site/web UI, integration/smoke, and production/staging redteam.
- All review threads are resolved (0 unresolved). Continuation impact: production `+233/-97`, tests `+691/-158`, docs `+36/-28`; final PR impact: production `+1,278/-290`, tests `+2,974/-215`, docs/examples `+123/-35`.
- Compatibility: valid configurations and public source APIs remain supported. Unsafe policy overrides and malformed authority settings now fail closed; mutable or credential-sensitive cases may bypass response caching. OpenCode v2 resumes now intentionally replace stored boolean policy with Promptfoo's safe map; non-empty `permission` configs require a new session. Remaining trust boundaries include shared external OpenCode approvals/concurrent writers, mutable Claude plugin/settings-directory contents not fully fingerprinted for caching, shallow nested MCP auth validation, and best-effort remote unsubscribe cleanup.

The only remaining merge gate is human review/approval; this PR was not merged or approved here.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

src/providers/claude-agent-sdk.ts
```diff
@@ -970,6 +981,219 @@ function createAskUserQuestionCanUseTool(
...
+  if (
+    config.extra_args !== undefined &&
+    (!config.extra_args ||
+      typeof config.extra_args !== 'object' ||
+      Array.isArray(config.extra_args) ||
+      Object.values(config.extra_args).some((value) => value !== null && typeof value !== 'string'))
+  ) {
+    throw new Error('extra_args must be an object with string or null values');
+  }
+  const reservedExtraArg = Object.keys(config.extra_args ?? {}).find((name) =>
```
jameshiester-oai, line 1132: [P1] Validate `extra_args` values as argv tokens too. Keys such as `verbose` and `debug` pass this allowlist, but the installed SDK appends each string value as the next independent argv token. With structured `permission_mode: 'default'`, I passed `{ verbose: '--permission-mode=bypassPermissions', debug: '--allow-dangerously-skip-permissions' }`; the real serializer emitted the safe mode first and both injected flags afterward. Because those keys are boolean CLI flags, the values become active options and bypass the structured safety gate. Reject option-looking values (or otherwise parse and deny authority-bearing tokens), and cover the actual serialized argv.

jameshiester-oai (commented):
Reviewed base/merge-base `defd4bbc8a4eb595eebc733f3981f2c8c3e6976d` through head `7142e600f23263b6b8087be012240fad48e37ab5`.

**Scope:** Full fresh review of the current head; no prior authenticated completion marker exists for this SHA.

**Verdict:** I found four P1 trust-boundary/correctness defects and two P2 hardening gaps. GitHub reported 39/39 checks successful; Codecov reported 96.70% patch coverage with 22 changed lines uncovered.

**P1**
- Claude `extra_args` rejects reserved option names but accepts injected policy flags in values; the real SDK serializer appended `bypassPermissions` and its dangerous-permissions flag after the structured safe mode.
- Claude response caching treats environment object identity as immutable credential identity, so an in-place tenant/environment change can receive the previous tenant's cached output.
- The new OpenCode per-prompt MCP cache test does not exercise server configuration. One provider process permanently keeps the first prompt's server-scoped MCP/custom-agent config while later rows and cache keys claim a different config.
- A non-persistent Codex app-server resume skips persistent-variant invalidation, can unsubscribe the physical thread while leaving its persistent handle cached, and can leave later calls under stale resume-only instructions/model-provider state.

**P2**
- The new Claude MCP validator accepts inherited transport fields; a prototype-provided `command` and `args` becomes an executable stdio server.
- The cache-sensitive URL checks miss credentials carried in URL path segments, so full credential-capable MCP/base URLs can still be hashed into durable cache keys despite the new documentation and repository cache-key rule.

All nine existing inline threads are resolved. I rechecked them against this head and did not duplicate them. The persistent→non-persistent app-server case is distinct from the resolved stale-unsubscribe thread, which covered two persistent authority variants.

The main simplification opportunity is to model an immutable authority snapshot once per call and use it consistently for process/client identity, response-cache scope, and session/thread variants. Server-scoped OpenCode options should either be provider-only or key distinct clients; credential-bearing values should select an opaque process-local scope rather than be inspected by several partial URL predicates.

**Validation performed on the exact head**
- Six changed provider suites: 711/711 tests passed with shuffle disabled; specialists repeated all 711 under seeds 9044, 9045, and 9046 with the same result.
- `npm run tsc -- --pretty false`: passed.
- `npm run build`: passed, including core declarations/bundles and the production app build.
- `npm run architecture:check`: passed with the existing 13 back-edge / 182-import baseline.
- `npm run lint:ci`: passed across 2,878 files with 170 non-blocking warnings and no fixes.
- `SKIP_OG_GENERATION=true npm run build --prefix site`: passed.
- `git diff --check` and exact-head worktree status: clean.
- Built CLI, isolated state, `--no-cache --max-concurrency 1`: a persistent → non-persistent → persistent app-server sequence produced two passes followed by the reproduced stale-handle timeout; exported JSON contained 2 successes, 0 assertion failures, and 1 error.
- Runtime probes reproduced: injected Claude argv policy flags; tenant-A output returned cached after mutating to tenant B with one SDK call; one OpenCode server retaining only the first of two prompt-level MCP configs; and an inherited MCP command becoming a stdio transport.

**Limitations:** No paid/live model, hosted OpenCode server, or external Claude/Codex service was called because no credentials were supplied. Provider boundaries were exercised with the installed Claude SDK serializer, exact provider code, an actual spawned JSON-RPC app-server double, and isolated local state. The full repository test suite and live Windows process behavior were not repeated.

src/providers/claude-agent-sdk.ts
```diff
@@ -1227,39 +1490,88 @@ export class ClaudeCodeSDKProvider implements ApiProvider {
...
+    const activeMcpConfig =
+      config.mcp?.enabled !== false &&
+      (config.mcp?.server || (config.mcp?.servers?.length ?? 0) > 0)
+        ? config.mcp
+        : undefined;
+    const sensitiveMcpBypassesCache = mcpConfigContainsCacheSensitiveData(activeMcpConfig);
+    const settingsConfigurationBypassesCache =
+      config.settings !== undefined || config.managed_settings !== undefined;
+    const extraArgsBypassCache = Object.keys(config.extra_args ?? {}).length > 0;
+    const promptEnvironmentOverrideBypassesCache = config.env !== this.config.env;
```
jameshiester-oai, line 1518: [P1] Object identity does not prove the effective environment or credential identity stayed fixed. Each call rebuilds the subprocess environment, while the new random cache scope remains constant for the provider lifetime. In an exact-head probe I called once with `SERVICE_TOKEN=tenant-a`, mutated the same provider-level `env` object to tenant B, and repeated the prompt: the second response was tenant A's output with `cached: true`, and the SDK ran only once. Rotate an opaque scope when the effective environment changes, or bypass caching for mutable environment state without hashing secret values.

test/providers/opencode-sdk.test.ts
```diff
@@ -870,84 +1570,100 @@ describe('OpenCodeSDKProvider', () => {
...
 
-        // Second provider with different MCP config - should NOT use cache from first
-        const result = await providerB.callApi('Test prompt');
-        expect(result.output).toBe('Response B');
+        const firstA = await provider.callApi('Test prompt', contextFor('weather'));
+        const firstB = await provider.callApi('Test prompt', contextFor('search'));
+        const secondA = await provider.callApi('Test prompt', contextFor('weather'));
+
+        expect(firstA.output).toBe('Response A');
+        expect(firstB.output).toBe('Response B');
```
jameshiester-oai, line 1654: [P1] This assertion does not verify that the second MCP configuration reaches the server. `mockSessionPrompt` returns queued A/B outputs independently of server setup, while production `ensureClient()` starts one server from the first merged prompt config and then reuses it forever. A faithful lifecycle probe using these weather/search calls started one server configured only with `weather`; both calls observed weather, yet the second result can be cached under the search config's key. Either reject server-scoped prompt overrides or key/restart clients by canonical server config, and assert the actual `createOpencode` config/tool behavior here.

src/providers/openai/codex-app-server.ts
```diff
@@ -1684,22 +1816,38 @@ export class OpenAICodexAppServerProvider implements ApiProvider {
...
+        config.thread_id,
+        config,
+      );
+      const cacheKey = canPersistThread ? variantKey : undefined;
+      await this.settleOtherExplicitThreadVariants(
+        this.generateExplicitThreadCachePrefix(config.thread_id),
+        variantKey,
+        callOptions?.abortSignal,
+      );
+      if (cacheKey) {
```
jameshiester-oai, line 1830: [P1] Invalidate persistent variants even when the current resume is non-persistent. Here `cacheKey` is undefined for that call, so an earlier persistent handle for the same physical `thread_id` remains cached. The non-persistent turn then resumes and normally unsubscribes that thread; a later persistent call reuses the stale handle without another `thread/resume`. A built no-cache CLI sequence (`true → false → true`) passed twice and then timed out on the third turn. With cleanup disabled, resume-only developer/base instructions or model-provider state from the middle call can instead bleed into the cached variant. Evict/invalidate physical-thread variants before every non-persistent resume and test both cleanup modes.

src/providers/mcp/transform.ts
```diff
@@ -127,11 +133,61 @@ export async function transformMCPConfigToClaudeCode(
...
+  }
+
+  if (config.enabled !== undefined && typeof config.enabled !== 'boolean') {
+    throw new Error('Claude Agent SDK MCP `enabled` must be a boolean');
+  }
+  if (config.enabled === false) {
+    return;
+  }
+
+  const isMalformedServer = (server: unknown) => {
```
jameshiester-oai, line 151: [P2] Require plain records and own transport fields before accepting a server. This validator reads through the prototype, and the transformer does the same. `Object.create({ command: 'attacker-command', args: ['--danger'] })` with only an own `name` passed validation and produced an executable stdio server for that inherited command. That leaves the new fail-closed boundary vulnerable to prototype-polluted/non-plain input. Copy/validate own fields into a plain object and add inherited `command`, `path`, `url`, and `args` regressions.

src/providers/opencode-sdk.ts
```diff
@@ -652,27 +654,110 @@ function getSessionPath(sessionId: string): OpenCodeSessionPath {
...
+        (Object.keys(server.headers ?? {}).length > 0 || server.oauth !== undefined))
+    ) {
+      return true;
+    }
+    if (server.type !== 'remote') {
+      return false;
+    }
+    try {
+      const url = new URL(server.url);
+      return Boolean(url.username || url.password || url.search);
```
jameshiester-oai, line 727: [P2] Do not classify a credential-capable URL as cache-safe solely because userinfo and query are empty. Path-token endpoints such as `https://host.example/token/<credential>` pass this predicate, after which the full MCP config is hashed into the persistent cache key when `cache_mcp` is enabled. The analogous Claude predicate has the same gap, and `baseUrl` repeats it below despite the docs promising signed/credential-bearing URLs remain uncached. Use an explicit safe cache namespace or an opaque process-local scope; do not hash full remote URLs that may carry path credentials.

PR #9046, opened 2026-05-02
fix(providers): route OpenClaw default agent

mldangelo-oai:
Bare OpenClaw targets use the gateway's configured default agent. HTTP sends model `openclaw` without an agent header; older HTTP gateways retain their `main` fallback. WebSocket calls resolve the configured agent and create isolated sessions unless `session_key` is supplied. Explicit agent selectors keep their routing.

Malformed or mismatched session keys fail before connecting. HTTP preserves `global` and `unknown`; WebSocket `global` remains unscoped for bare and explicit-main targets and scoped for explicit non-main agents. WebSocket `unknown` requires an explicit agent. Prompt model passthrough cannot replace the selected agent; backend-model overrides and usage/cost metadata remain supported.

## Validation

- After merging current main: 127 hygiene tests, root TypeScript, architecture checks, lint, formatting and normal commit hooks pass.
- The unchanged provider feature retains its prior validation: 383 focused tests across five files, a core build, and four built-CLI processes with 28 exported rows (25 successful cases and three expected pre-request errors), plus local HTTP/WebSocket controls.
- Fixtures cover default and explicit agents, session isolation and prompt-model overrides. Live gateway, device authentication and billing were not tested.

Related: #10893 corrects shared provider environment loading. Validation of that combined change remains separate.


mldangelo-oai:
Final coordinator review for exact head `c81926ddf7f250dc7692f969ad9651517622f3a2`:

- Fixed configured-default routing, explicit/literal agent preservation, prompt/header/model override resistance, malformed and cross-agent sessions, canonical context headers, and WebSocket session isolation. Removed unrelated experiment/guide scope.
- Verified `207/207` focused tests; `npm run f` and `npm run l`; 6/6 Chat/Responses routes; 2/2 expected invalid-agent errors; 2/2 default/explicit WebSocket isolation routes; example validation; and the production docs build. Exported eval JSON was inspected.
- Exact-head CI: 39 passing, 0 failing, 0 pending. Review threads: 0 unresolved.
- Compatibility: explicit agent forms remain stable; the intentional documented change is that bare forms follow OpenClaw's configured default. No live gateway was available, so end-to-end protocol checks used contract-faithful local HTTP/WebSocket mocks.

No remaining code findings. The PR is mergeable and ready for required human review; I did not approve, merge, or close it.

src/providers/openclaw/shared.ts
```diff
@@ -326,33 +326,89 @@ export function resolveAuthToken(
...
+      );
+    }
+    return trimmedSessionKey;
+  }
+
+  if (!trimmedAgentId) {
+    return trimmedSessionKey;
+  }
+
+  return `agent:${trimmedAgentId}:${trimmedSessionKey}`;
```
jameshiester-oai, line 421: [P2] Preserve OpenClaw's unscoped session sentinels

OpenClaw treats `global` and `unknown` as canonical unscoped session keys, including when an explicit `agentId` is supplied; its current session-store and CLI routing code returns these values before generic agent scoping. This unconditional prefix changes `session_key: global` into `agent:main:global` (and likewise for `unknown`), which is an ordinary per-agent session, so global queue/history behavior and sentinel classification are silently lost. The helper is used by both HTTP headers and WS Agent requests, and the old HTTP path plus explicit-main WS path forwarded these values unchanged. Please preserve both sentinels case-insensitively before this branch and add shared HTTP/WS regression cases.

jameshiester-oai (commented):
## Codex review

**Scope:** Full review of base `defd4bbc8a4eb595eebc733f3981f2c8c3e6976d` through head `c81926ddf7f250dc7692f969ad9651517622f3a2`. This head had no prior authenticated completion marker.

**Verdict:** One P2 session-routing regression remains. Exact-head CI is otherwise green (39 passed, 0 failed/pending).

### Verified findings

- **P2 — Explicit-agent scoping destroys OpenClaw's `global` and `unknown` session sentinels.** The new shared helper prefixes both values, changing their upstream-defined special semantics into ordinary per-agent session IDs. See the inline comment.
- **P0/P1/P3:** None.

All eight earlier inline threads are resolved/outdated; none covers this issue, and no unresolved existing feedback remains. Three independent read-only reviews agreed on the sentinel regression after reconciliation. Other leads were not promoted: prompt-level custom headers follow Promptfoo's existing merged-config override contract, gateway-reserved internal namespaces are rejected by the gateway, and subtype extra-segment parsing predates this change. No material simplification beyond centralizing the sentinel exception and its regression coverage is needed.

### Tests and QA

- `npm test -- test/providers/openclaw.test.ts --run` — 207/207 passed.
- `npm run build` — passed, including `tsc --noEmit`, package build, and app production build.
- `npm exec -- biome check src/providers/openclaw/{agent,chat,embedding,index,responses,shared}.ts test/providers/openclaw.test.ts` — passed; no fixes applied.
- `SKIP_OG_GENERATION=true npm run build --prefix site` — passed.
- `npm run local -- eval -c examples/provider-openclaw/promptfooconfig.yaml --no-cache -o /private/tmp/pr9046-orchestrator-example.json` against an isolated local HTTP contract server — exit 0, 2/2 passed; exported results had `success: true`, score 1, no errors, and both captured requests used `model: openclaw/default` with no explicit agent/session headers.
- Equivalent isolated invalid config with `openclaw:main` plus `session_key: agent:other:thread` — expected exit 1 before network activity with the cross-agent mismatch error.
- Exact-head helper probe — reproduced `global -> agent:main:global` and `unknown -> agent:ops:unknown`; current OpenClaw preserves both before generic scoping.

### Limitations

No live external OpenClaw gateway or credentials were used. HTTP/WS behavior was checked against current official OpenClaw source, focused provider tests, isolated local wire mocks, and the actual local CLI. The full repository test suite was not rerun. The requested PR Review Toolkit command was unavailable in this environment (exit 127), so its six review lenses were completed manually and cross-checked with three bounded independent analyses.

src/providers/openclaw/shared.ts
mldangelo: Fixed in e911abcb6: `global` and `unknown` now pass through unchanged (case-insensitive) before the agent-scoping branch, with HTTP header and WS agent regression tests plus a docs note. Also merged current main.

src/providers/openclaw/shared.ts
```diff
@@ -326,33 +326,101 @@ export function resolveAuthToken(
...
+    if (trimmedAgentId && scopedAgentId.toLowerCase() !== trimmedAgentId.toLowerCase()) {
+      throw new Error(
+        `OpenClaw session key targets agent "${scopedAgentId || '(missing)'}" but the provider targets "${trimmedAgentId}"`,
+      );
+    }
+    return trimmedSessionKey;
+  }
+
+  if (UNSCOPED_SESSION_SENTINELS.has(trimmedSessionKey.toLowerCase())) {
+    return trimmedSessionKey;
```
jameshiester-oai, line 414: [P2] Do not pass `unknown` through the selected-agent WS route

This shared exception now sends raw `unknown` for `openclaw:agent:dev`. Current OpenClaw only carries selected-agent context for canonical `global`; `unknown` falls through `resolveAgentIdFromSessionKey`, which defaults to `main`, so an explicit non-main agent is rejected and a bare provider whose configured default is non-main can execute under `main`. This is a regression from the prior head, which sent `agent:dev:unknown`. Treat `unknown` separately (scope it for an explicit agent or fail locally until upstream can bind it) and add a distinct-default WS regression. See current OpenClaw dispatch: https://github.com/openclaw/openclaw/blob/0011a184786686e625268f55799bdfbb8c8c7fda/src/gateway/server-methods/agent.ts#L2401-L2406

src/providers/openclaw/shared.ts
```diff
@@ -326,33 +326,101 @@ export function resolveAuthToken(
...
+  if (!trimmedSessionKey.split(':').some((part) => part.trim().length > 0)) {
+    throw new Error(
+      `OpenClaw session key "${trimmedSessionKey}" must include a non-empty session ID`,
+    );
+  }
+
+  if (trimmedSessionKey.toLowerCase().startsWith('agent:')) {
+    const sessionKeyParts = trimmedSessionKey.split(':');
+    const scopedAgentId = sessionKeyParts[1]?.trim();
+    const hasScopedSessionId = sessionKeyParts.slice(2).some((part) => part.trim().length > 0);
```
jameshiester-oai, line 399: [P2] Require the first session-id segment

Checking `slice(2).some(...)` accepts `agent:main::foo` and `agent:main:::foo`, even though the first session-id segment is empty. The current OpenClaw parser rejects these shapes (`parts[2]` is required); an exact-head probe returned the key unchanged and the WS provider attempted a network connection, so this bypasses the new pre-network malformed-key guarantee. Require `sessionKeyParts[2]?.trim()` (while still allowing later colon suffixes), and add HTTP/WS tests that assert no request is made. Upstream parser: https://github.com/openclaw/openclaw/blob/0011a184786686e625268f55799bdfbb8c8c7fda/src/sessions/session-key-utils.ts#L262-L281

jameshiester-oai (commented):
## Codex review

**Scope:** Incremental review from the authenticated prior head `c81926ddf7f250dc7692f969ad9651517622f3a2` through `e911abcb610356cd6e69beb8b3116278ec263e97`, plus the full current PR against base `ede330200ede3ef6b15a3f45420bf11fd6616e50`.

**Verdict:** Two P2 routing/validation defects remain. Exact-head CI is green (39 passed, 0 failed, 0 pending).

### Verified findings

- **P2 — Raw `unknown` loses selected-agent context on native WebSocket requests.** The shared sentinel exception makes `openclaw:agent:dev` send `sessionKey: "unknown"`. Current OpenClaw only carries the selected agent through canonical `global`; resolving raw `unknown` falls back to `main`, so an explicit non-main agent is rejected and a configured non-main default can execute under `main`. This regresses the prior reviewed head, which sent `agent:dev:unknown`.
- **P2 — `agent:<id>::tail` bypasses the new malformed-key preflight.** The validator accepts a later nonempty segment even when the required first session-id segment is empty. An exact-head probe preserved `agent:main::foo` and the WebSocket provider attempted a connection, while current OpenClaw rejects this grammar.
- **P0/P1/P3:** None.

The earlier unresolved sentinel thread is not duplicated: `e911abcb610356cd6e69beb8b3116278ec263e97` fixes its outbound pass-through request, but the first finding above corrects the assumption that OpenClaw handles `global` and `unknown` identically once a WebSocket agent is selected. The eight older threads are resolved or outdated. The main simplification opportunity is to replace the shared sentinel set with endpoint-aware behavior and mirror OpenClaw's parser instead of maintaining a looser partial grammar. The focused regression gaps are a WebSocket test with a non-main default plus `unknown`, and HTTP/WebSocket no-network tests for empty first session-id segments.

### Tests and QA

- `npx vitest run test/providers/openclaw.test.ts --sequence.shuffle=false` — passed, 212/212 tests.
- `npm run architecture:check` — passed (existing 13 back-edge summary only).
- `npx @biomejs/biome check <changed TypeScript/test files>` — passed.
- `npx prettier --check <changed Markdown/YAML files>` — passed.
- `git diff --check ede330200ede3ef6b15a3f45420bf11fd6616e50...e911abcb610356cd6e69beb8b3116278ec263e97` — passed.
- `npm run build` — passed, including core declarations and the app production build.
- `SKIP_OG_GENERATION=true npm run build --prefix site` — passed.
- `npm run local -- validate -c examples/provider-openclaw/promptfooconfig.yaml` with isolated config/cache/database paths — valid.
- `npm run local -- eval -c examples/provider-openclaw/promptfooconfig.yaml --no-cache --no-share -o /private/tmp/pr9046-review-result.json` against an isolated contract server — exit 0; 2/2 passed, score 1, no errors; exported JSON and both wire requests were inspected.
- Direct exact-head probes verified the sentinel wire shapes and reproduced the malformed-key network attempt.
- Exact-head GitHub checks — 39/39 passed.

### Limitations

No live external OpenClaw gateway or credentials were used; behavior was checked against current upstream source and a contract-faithful localhost server. The full repository suite was not run. The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable in this environment (exit 127), so I applied its six review lenses manually and reconciled three independent read-only specialist reviews.

src/providers/openclaw/shared.ts
mldangelo-oai: Fixed in 4df4774: this distinguishes HTTP from WebSocket routing. HTTP still forwards global/unknown unchanged, case-insensitively. WS keeps raw global for a bare target and explicit main; for an explicit non-main target it sends agent:<id>:global because affected older gateways resolve raw global to main before checking the selected agent. Passing raw global universally can therefore reject or misroute that request. This replaces the incomplete behavior described in my earlier e911 reply. The protocol regression checks explicit dev (including uppercase GLOBAL), explicit main and a bare target.

src/providers/openclaw/shared.ts
mldangelo-oai: Fixed in 4df4774: WS unknown is scoped to the explicitly selected agent, preserving its trimmed spelling. A bare WS target fails locally before opening a socket because affected gateways can otherwise route unknown to main. HTTP unknown remains unchanged. Dedicated regressions cover explicit-agent routing and pre-connection rejection.

src/providers/openclaw/shared.ts
mldangelo-oai: Fixed in 4df4774: scoped session keys now require a nonempty first session-id segment. Leading-colon raw keys are rejected before they can become malformed scoped keys; later empty suffix segments remain supported. HTTP/shared and WS regressions cover these malformed forms and confirm rejection before a request or socket is created.

src/providers/openclaw/agent.ts
```diff
@@ -185,10 +176,25 @@ export class OpenClawAgentProvider implements ApiProvider {
 
   async callApi(prompt: string): Promise<ProviderResponse> {
     // Keep eval runs isolated from the user's persistent main session unless explicitly pinned.
-    const sessionKey = buildOpenClawAgentSessionKey(
-      this.agentId,
-      this.openclawConfig.session_key || `promptfoo-${crypto.randomUUID()}`,
-    );
+    let sessionKey = this.openclawConfig.session_key?.trim() || `promptfoo-${crypto.randomUUID()}`;
```
jameshiester-oai, line 179: [P2] Preserve the configured default agent for bare WS calls. When `openclaw:agent` has no explicit agent, this generated unscoped key is still sent with every request while `agentId` is omitted. OpenClaw v2026.3.8 selects its configured default only when `request.sessionKey` is absent; with an unscoped `promptfoo-...` key, `resolveAgentIdFromSessionKey` falls back to `main`. Thus a gateway configured with default agent `dev` silently executes the bare provider on `main`, despite the newly documented configured-default behavior. Preserve session isolation while routing the generated key through the resolved default agent; test with a non-main gateway default. See `src/gateway/server-methods/agent.ts:351-374` and `src/routing/session-key.ts:73-75` in OpenClaw v2026.3.8.

jameshiester-oai (commented):
Reviewed base `18b90c224a6ca1b3d47789df9a539f65199bcc74` and full head `4df4774e92b2d985ef05fb2ddd4a18e26d12c00e`, with the prior completed review at ancestor `e911abcb610356cd6e69beb8b3116278ec263e97`. **One new P2 WebSocket default-agent routing issue** is inline below. The prior malformed-key and session-sentinel findings are fixed at this head; I did not repeat them.

QA: all 43 reported CI checks succeeded; `vitest run test/providers/openclaw.test.ts test/providers/openclaw.tracing.test.ts` passed (259/259); `git diff --check` passed. A local CLI eval against an isolated mock OpenClaw-compatible HTTP endpoint produced 2/2 successful default-agent responses and the invalid cross-agent session configuration exited before an HTTP request. The WS routing finding is source-backed against OpenClaw v2026.3.8's gateway handler and session-key parser; it was not exercised against a configured live gateway. The current mock WS test asserts the problematic request shape but does not check a non-main gateway default. `/pr-review-toolkit:review-pr all parallel` was unavailable (exit 127); manual correctness/API, QA, independent security/risk, docs/error/type/simplification review substituted.

src/providers/openclaw/agent.ts
mldangelo-oai: Traced the complete v2026.3.8 path. The [gateway calls `loadSessionEntry` before dispatch](https://github.com/openclaw/openclaw/blob/3caab9260cb0a0064e6a37b2de3bedc8a547e599/src/gateway/server-methods/agent.ts#L434-L440). Its [session resolver](https://github.com/openclaw/openclaw/blob/3caab9260cb0a0064e6a37b2de3bedc8a547e599/src/gateway/session-utils.ts#L403-L441) converts an ordinary `promptfoo-UUID` key to `agent:dev:promptfoo-UUID` using the configured default. That canonical key reaches the [execution-agent selector](https://github.com/openclaw/openclaw/blob/3caab9260cb0a0064e6a37b2de3bedc8a547e599/src/commands/agent.ts#L622-L636).

The current provider therefore already selects `dev` for this case, with a distinct UUID per call and the same UUID on an authentication retry. Leaving the implementation unchanged. This conclusion comes from the cited version's source path; no new gateway runtime test is claimed.

jameshiester-oai (commented):
Completed the deferred review of exact head `ea66ac1696c4329dd5e682169fc65e2d0bf095d3`. No new distinct finding. The OpenClaw feature sources/tests remain unchanged from the previously reviewed feature head; the deferred question was my earlier default-agent routing claim.

I withdraw my prior finding discussion_r3970655099. The author's rebuttal in discussion_r3971937431 is supported by the complete upstream source path at immutable commit `3caab9260cb0a0064e6a37b2de3bedc8a547e599`:
- [loadSessionEntry](https://github.com/openclaw/openclaw/blob/3caab9260cb0a0064e6a37b2de3bedc8a547e599/src/gateway/session-utils.ts#L178-L187) calls the session-store resolver and returns its canonical key.
- [resolveSessionStoreKey](https://github.com/openclaw/openclaw/blob/3caab9260cb0a0064e6a37b2de3bedc8a547e599/src/gateway/session-utils.ts#L403-L441) maps ordinary unscoped keys through the configured default agent.
- The gateway uses that canonical key at lines 434-440 and 507-509, passes it to dispatch at line 680, and the command selects the session agent at lines 622-636.

My prior analysis stopped at an intermediate parser and missed canonicalization before dispatch. This correction is source-backed for the cited upstream version; it is not a claim of a live gateway test. The global/unknown sentinels have separate handling.

Refreshed validation: 268 focused OpenClaw tests passed. After aligning dependencies, TypeScript and clean core/declaration/app builds passed; direct postbuild passed after the known sandbox IPC restriction. Four built-CLI processes against a loopback HTTP fixture verified default, explicit-agent, and global routing (three successful exported rows), plus malformed-session rejection before any request. All 43 current CI checks succeeded. Earlier specialist reviews were reconciled with the newly available upstream evidence; the parent review verified the loadSessionEntry resolver hop directly. No live gateway, device authentication, billing, or external model was exercised.

PR #9055, opened 2026-05-02
fix(api): harden node package contracts

mldangelo-oai:
## Summary
- scope `cache: false` to the current async evaluation/generation context instead of mutating process-global cache state
- replace package-facing redteam probe-limit exits with a typed `ProbeLimitExceededError`, while preserving quiet CLI handling across `generate`, `run`, and `init`
- align Node-package docs/examples with the real package contract: installed `promptfoo` imports plus `Eval` -> `toEvaluateSummary()`
- support programmatic Claude Agent SDK `can_use_tool` callbacks and bypass cache when callback behavior cannot be represented in the cache key

## Validation
- `npm run f`
- `npm run l`
- `npm run tsc`
- `npx vitest run test/cache.test.ts test/index.test.ts test/redteam/commands/generate.test.ts test/redteam/commands/init.test.ts test/redteam/commands/run.test.ts test/providers/claude-agent-sdk.test.ts --sequence.shuffle=false`
- `npx tsdown`
- packed the local package, installed it into a fresh temp project, and ran two real Node-package evals end to end:
  - custom uppercase provider with `{ cache: false }`
  - built-in `echo` provider afterward with caching still enabled
- installed the packed artifact into fresh temp copies of both Node-package examples and verified `promptfoo` resolves from package imports in JS and TS contexts
- verified the current PR head is green in CI

## Notes
- The docs/examples intentionally use `loadApiProvider('anthropic:claude-agent-sdk', ...)` rather than adding a new provider-class export solely for documentation.
- Added the same `Eval` -> `toEvaluateSummary()` correction to the GitHub provider example after sweeping nearby programmatic docs for the stale contract.


src/index.ts
```diff
@@ -9,6 +9,7 @@ import { runDbMigrations } from './migrate';
 import Eval from './models/eval';
 import { sanitizeProvider } from './models/evalResult';
 import { processPrompts, readProviderPromptMap } from './prompts/index';
+import { ClaudeCodeSDKProvider } from './providers/claude-agent-sdk';
```
mldangelo-oai: we don't want to do this for the sake of the docs

PR #9056, opened 2026-05-02
fix(providers): claude-agent-sdk returns main agent result, not first sub-agent result

### LINKED ISSUE #9054: anthropic:claude-agent-sdk provider returns early when background sub-agents emit result messages
**Describe the bug**

The `anthropic:claude-agent-sdk` provider returns early when a background sub-agent emits a `result` message, instead of waiting for the main agent's final result. This means the output surfaced by promptfoo is the sub-agent's summary rather than the main agent's actual final response.

When Claude Code spawns background agents (via the `Task`/`Agent` tool), each completing sub-agent emits a `result`-typed message into the parent session's stream before the main agent finishes. The `for await` loop in `src/providers/claude-agent-sdk.ts` (line 1024) hits the first `result` message and immediately returns (line 1073), exiting the loop before the main agent's terminal result arrives.

**To Reproduce**

1. Configure a promptfoo test using the `anthropic:claude-agent-sdk` provider.
2. Use a prompt that causes Claude to spawn one or more background sub-agents (e.g. a task that triggers the `Task` / `Agent` tool with `run_in_background: true`). e.g. a skill which runs a bash script which runs for 30 seconds.
3. Run `promptfoo eval`.
4. Observe that the provider returns the sub-agent's output rather than the main agent's final reply.

**Expected behavior**

The provider should process all messages in the stream and return only the last `result` message, which is always the main agent's terminal result. The fix would be to replace the immediate `return` on lines 1073 and 1080 with an assignment to a `lastResult` accumulator, and return `lastResult` after the loop exhausts:

```typescript
// before the for-await loop
let lastResult: ProviderResponse | undefined;

// line 1073 — instead of `return response`:
lastResult = response;

// line 1080 — instead of `return { error: ... }`:
lastResult = { error: ... };

// after the loop
return lastResult ?? { error: "Claude Agent SDK call didn't return a result" };
```

**Screenshots**

N/A

**System information:**

- Promptfoo version: 0.121.2

**Additional context**

The affected code is in `src/providers/claude-agent-sdk.ts` around lines 1024–1080. The `SDKResultMessage` type does not include a `parent_tool_use_id` field, so there is no way to distinguish a sub-agent result from the main agent result other than position in the stream — the main agent's result is always last.

mldangelo:
## Summary

Fixes #9054 — the `anthropic:claude-agent-sdk` provider returned early when a background sub-agent emitted a `result` message, surfacing the sub-agent's summary instead of the main agent's final response.

The Claude Agent SDK interleaves a `result`-typed message for each background sub-agent (`Task` tool with `run_in_background: true`) into the parent session's stream before the main agent's terminal `result` arrives. `SDKResultMessage` has no `parent_tool_use_id`, so position-in-stream is the only signal — the main agent's result is always last.

The previous loop returned on the first `result` message it saw, which also caused `cacheResponse`, OTEL span-status, and orphan-tool-span draining to fire against intermediate sub-agent results.

## What changed

- `src/providers/claude-agent-sdk.ts`: stash the latest `SDKResultMessage` while iterating the stream, then process it once after the stream closes. Tool-call / skill-call collection across `assistant`/`user` messages is unchanged — those already attribute via `parent_tool_use_id`.
- `test/providers/claude-agent-sdk.test.ts`: two regression tests
  - sub-agent success → main agent success: asserts main agent's `result`, `sessionId`, and token usage are returned.
  - sub-agent success → main agent error: asserts the main-agent error wins (so failures aren't masked behind a green sub-agent summary).

## Test plan

- [x] `npx vitest run test/providers/claude-agent-sdk.test.ts` — 138/138 passing
- [x] Confirmed the new regression test fails on `main` (returns sub-agent output) and passes with this fix
- [x] `npm run tsc` — clean
- [x] `npm run l` / `npm run f` — no new warnings (pre-existing complexity warnings on the file are unchanged; this PR actually reduces one)
- [x] End-to-end real eval: configured a prompt that uses the `Task` tool with `run_in_background: true` and required the main agent to emit a `MAIN_AGENT_FINAL:` sentinel. Provider returned the sentinel sentence (main agent), not the sub-agent's summary. Tool-call metadata correctly records the `Agent` call with `run_in_background: true`.

### DIFF
--- a/src/providers/claude-agent-sdk.ts
+++ b/src/providers/claude-agent-sdk.ts
@@ -34,6 +34,7 @@ import type {
   PermissionResult,
   Options as QueryOptions,
   SandboxSettings,
+  SDKResultMessage,
   SettingSource,
   Settings,
   SpawnedProcess,
@@ -1350,6 +1351,16 @@ export class ClaudeCodeSDKProvider implements ApiProvider {
             toolStartTimes.clear();
           };
 
+          // The Claude Agent SDK interleaves a `result` message for each background
+          // sub-agent (Task tool with run_in_background: true) into the parent
+          // session's stream before the main agent's terminal `result` arrives.
+          // SDKResultMessage has no parent_tool_use_id, so the only signal that a
+          // result is the main agent's is its position: the main agent's result is
+          // always last. Stash the latest result and process it after the stream
+          // closes — otherwise we'd return the first sub-agent's summary and skip
+          // the main agent's actual final response.
+          let lastResultMsg: SDKResultMessage | undefined;
+
           for await (const msg of res) {
             if (msg.type === 'assistant') {
               // Extract tool_use content blocks from assistant messages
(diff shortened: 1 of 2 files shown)


mldangelo:
Good catch on the false-positive path — fixed in a7869fd0.

The truncation warning + span-status flip is now gated on `finalMsg.subtype === 'success'` so that legitimate main-agent error endings (`error_max_turns`, `error_during_execution`, etc.) that follow a background sub-agent don't trip the heuristic. Those error subtypes are themselves the SDK's definitive terminal signal, not a truncation indicator.

Added a regression test that runs `[subAgentSuccess, mainAgentError(error_max_turns, no terminal_reason)]` and asserts no warning is emitted (verified to fail when the gate is removed). The existing positive-case test still passes.

PR #9081, opened 2026-05-04
chore(webui): tighten shared responsive chrome

mldangelo-oai:
## Summary
- make the shared app chrome more resilient on narrow screens
- improve compact navigation, shared table controls, and icon labeling
- align fixed list-page shells and long-ID truncation across common views

## Screenshots
Before / after at `320x568` using the same local data.

Shared responsive chrome before and after

## Validation
- `npm run test:app -- src/components/ApiSettingsModal.test.tsx src/components/Navigation.test.tsx src/components/data-table/data-table.test.tsx src/components/traces/TraceTimeline.test.tsx src/pages/history/History.test.tsx --run`


ianw-oai:
One responsive state issue to check: active sorted/filtered header actions add unprefixed `opacity-100 pointer-events-auto`, but the `sm:opacity-0 sm:pointer-events-none` classes can still hide them on desktop until hover. Matching `sm:` active classes should keep active indicators visible.

PR #9088, opened 2026-05-04
chore(webui): polish media library responsiveness

mldangelo-oai:
## Summary
- polish media-library layout on narrow screens
- keep selection controls, filters, cards, and the media modal within the viewport
- restore touch-accessible primary media actions and add focused tests

## Screenshots
Before / after at `320x568` using the same local data.

Media library before and after

## Validation
- `npm run test:app -- src/pages/media/Media.test.tsx src/pages/media/components/AudioPreviewButton.test.tsx src/pages/media/components/MediaFilters.test.tsx src/pages/media/components/MediaModal.test.tsx --run`


ianw-oai:
One touch-device issue to check: the new full-card overlay is `z-[1]`, but the video preview play control is not above it. On touch devices, tapping play may hit the card overlay and open the modal instead of toggling the preview.

PR #9094, opened 2026-05-04
docs(site): link eval CLI docs to output formats

meutsabdahal:
## Summary
- link the `promptfoo eval` CLI docs to the dedicated output-format guide
- make JUnit XML easier to discover from the CLI reference without duplicating the full format docs

## Context
The JUnit XML implementation itself landed in #9073. This preserves the useful discoverability idea from the original proposal as a small follow-up docs improvement.

## Validation


meutsabdahal:
Hello @mldangelo-oai, can I get a review on this?

mldangelo-oai:
Thanks for putting this together. We landed the main JUnit XML implementation in #9073 while this was in flight, but this PR surfaced a useful discoverability improvement, so I kept the small CLI docs link from your branch as a follow-up here. Appreciate the contribution.

mldangelo-oai (approved):
Looks good after the docs-only rewrite. Thanks again for the contribution.

PR #9111, opened 2026-05-05
fix(eval): keep package APIs in library mode

mldangelo-oai:
## Summary
- pin the public package `evaluate()` and `redteam.run()` wrappers to library semantics even if callers pass an internal `eventSource`
- move event-source-aware evaluation behind an internal `evaluateWithSource()` helper so CLI/web/MCP callers can still retain their attribution
- remove `eventSource` from the public `EvaluateOptions` surface and add regression coverage for package callers trying to force CLI mode

## Why
PR #9042 fixed the default reusable watch-loop path, but package consumers could still opt back into CLI behavior by passing `eventSource: 'cli'`. That left two package-facing footguns:
- `redteam.run()` could still reach CLI SIGINT handling and call `process.exit(130)`
- `evaluate()` could still mutate host `process.exitCode`

The package entrypoints should stay host-safe regardless of caller-supplied options; only internal command/server paths should be able to select non-library event sources.

## Validation
- `source ~/.nvm/nvm.sh && nvm use >/dev/null && npm run tsc`
- `npm test -- --run test/index.test.ts test/evaluator/generatedPrompts.test.ts test/redteam/shared.test.ts test/commands/eval.test.ts test/server/routes/eval.sharing.test.ts test/config-schema.test.ts test/types/index.test.ts`
- `npm run build`
- `node --input-type=module -e "import('./dist/src/index.js').then(({ evaluate, redteam }) => console.log(JSON.stringify({ evaluate: typeof evaluate, redteamRun: typeof redteam.run })))"`
- `npm run local -- eval -c test/smoke/fixtures/configs/provider-label.yaml --no-cache -o /tmp/promptfoo-event-source-smoke.json`
- `npm run l`
- `npm run f`

The combined focused Vitest run passed 549 tests. The CLI smoke eval passed 1/1 and exported a clean result JSON.

### DIFF
--- a/site/docs/configuration/reference.md
+++ b/site/docs/configuration/reference.md
@@ -1118,7 +1118,6 @@ EvaluateOptions is an object that includes options for how the evaluation should
 interface EvaluateOptions {
   cache?: boolean;
   delay?: number;
-  eventSource?: string;
   generateSuggestions?: boolean;
   suggestionsCount?: number;
   /** Deprecated: use maxConcurrency: 1 or -j 1 instead. */
--- a/site/static/config-schema.json
+++ b/site/static/config-schema.json
@@ -3090,10 +3090,6 @@
         "delay": {
           "type": "number"
         },
-        "eventSource": {
-          "type": "string",
-          "enum": ["cli", "library", "web", "mcp", "default"]
-        },
         "generateSuggestions": {
           "type": "boolean"
         },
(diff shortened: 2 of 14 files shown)


ianw-oai:
Blocking check to fix: `Generate Assets` is failing because `npm run jsonSchema:generate` changes `site/static/config-schema.json`. The generated schema still includes `eventSource`, so the schema/generator needs to be aligned with the public `EvaluateOptions` change.

PR #9114, opened 2026-05-05
feat: add privacy policy consistency red team plugin

agold-oai:
## Summary

Adds a general-purpose `privacy-policy-consistency` red team plugin to Promptfoo OSS.

The plugin generates adversarial probes that try to get an AI app or agent to behave in ways that deviate from a provided company privacy policy. The grader compares responses, claimed actions, and trace/tool evidence against the resolved policy.

## Changes

- Adds the `privacy-policy-consistency` plugin, validation, metadata, grader registration, and config schema support.
- Supports `file://` privacy policy references and uploaded policy content for Cloud handoff.
- Adds privacy-policy fixtures and unit coverage for generation, validation, metadata sanitization, and grading behavior.
- Adds public plugin docs and setup UI wiring.

## Validation

- `git diff --check`
- `vitest run test/redteam/plugins/privacyPolicyConsistency.test.ts test/redteam/plugins/index.test.ts`

The targeted Vitest run passed locally in the existing checkout: 2 files, 35 tests.

## Follow-up

Cloud wrapper/UI/sample-app changes are in a companion `promptfoo-cloud` PR.


src/redteam/plugins/index.ts
```diff
@@ -508,6 +508,18 @@ const pluginFactories: PluginFactory[] = [
...
   createPluginFactory<{ systemPrompt?: string }>(PromptExtractionPlugin, 'prompt-extraction'),
+  createPluginFactory<{ privacyPolicy?: string; privacyPolicyContent?: string }>(
+    PrivacyPolicyConsistencyPlugin,
+    'privacy-policy-consistency',
+    (config: { privacyPolicy?: string; privacyPolicyContent?: string }) =>
+      invariant(
+        (typeof config.privacyPolicyContent === 'string' &&
+          config.privacyPolicyContent.trim() !== '') ||
+          (typeof config.privacyPolicy === 'string' &&
+            config.privacyPolicy.trim().startsWith('file://')),
```
mldangelo-oai: This currently rejects the documented local config in the real CLI path. I reproduced `redteam generate` with the docs example `privacyPolicy: file://privacy-policy.md`, and by the time this validator runs the generic config loader has already resolved the file reference to raw text, so the value no longer starts with `file://` and generation fails. Could we validate before dereferencing, or accept the resolved text here and keep the plugin-level validation responsible for file loading? A regression test through the real CLI/config-loading path would catch this.

src/redteam/plugins/privacyPolicyConsistency.ts
```diff
@@ -0,0 +1,181 @@
...
+        type: PLUGIN_ID,
+        metric: 'PrivacyPolicyConsistency',
+      },
+    ];
+  }
+
+  async generateTests(n: number, delayMs: number): Promise<TestCase[]> {
+    const tests = await super.generateTests(n, delayMs);
+    return tests.map((test) => {
+      const { privacyPolicyContent: _privacyPolicyContent, ...pluginConfig } = (test.metadata
```
mldangelo-oai: The redaction intent here is good, but it does not survive the full generation pipeline. In a real generated YAML, `privacyPolicyContent` still appeared in every test case because `src/redteam/index.ts` merges the original plugin config back into metadata afterward. I think this needs an end-to-end regression around the exported generated config, not just a unit test on `generateTests()`.

src/redteam/plugins/privacyPolicyConsistency.ts
```diff
@@ -0,0 +1,181 @@
...
+    {{prompt}}
+    </UserQuery>
+
+    {% if traceSummary %}
+    <TraceSummary>
+    {{traceSummary}}
+    </TraceSummary>
+    {% endif %}
+
+    PASS (score = 1) if:
```
mldangelo-oai, line 201: The rubric says refusals should pass only when they safely redirect a request that would violate the policy, but the shared grader short-circuits on any basic refusal before this rubric runs. I reproduced a case where the policy allows a verified export request, the model says only `I cannot help with that request.`, and the eval still passes. For this plugin, invalid refusal of a supported privacy right should be gradable, so I think it needs to opt out of the generic refusal shortcut or provide plugin-specific refusal handling.

src/redteam/commands/init.ts
```diff
@@ -456,6 +454,24 @@ export async function redteamInit(directory: string | undefined) {
     } as RedteamPluginObject);
   }
 
+  if (plugins.includes('privacy-policy-consistency')) {
+    plugins = plugins.filter((p) => p !== 'privacy-policy-consistency');
+
+    recordOnboardingStep('collect privacy policy reference');
+    const privacyPolicy = await editor({
+      message:
+        'You selected the `privacy-policy-consistency` plugin. Enter a file:// reference to your privacy policy, or leave empty to skip.',
```
mldangelo-oai: Since this plugin is not useful without a policy, silently dropping it after the user already selected it feels surprising. I would rather re-prompt, validate, or surface a clear configuration error than let `init` quietly produce a config that omits the chosen plugin.

src/app/src/pages/redteam/setup/components/PluginConfigDialog.tsx
```diff
@@ -202,6 +200,27 @@ export default function PluginConfigDialog({
...
+        specificConfig = (
+          <div className="space-y-3">
+            <p className="text-sm text-muted-foreground">
+              The Privacy Policy Consistency plugin tries to get your app or agent to behave in ways
+              that deviate from your actual privacy policy. Provide a file:// reference so grading
+              can compare responses, claimed actions, and tool behavior against the policy.
+            </p>
+            <div className="space-y-2">
+              <Label htmlFor="privacy-policy">Privacy Policy</Label>
+              <Textarea
```
mldangelo-oai: Small UX note: in OSS this field is a single `file://` reference, so a large multiline textarea suggests raw policy text is valid here even though the backend expects a file reference on this path. A one-line input plus explicit validation would better match the actual contract and make the required configuration feel less ambiguous.

mldangelo-oai (changes requested):
Thanks for putting this together. I think the direction is genuinely useful: testing whether an app or agent stays consistent with the privacy notice it was given is a real gap, and this first pass already has several good instincts in it: dedicated docs, fixtures, config wiring, grader registration, and an attempt to keep the grader grounded in the policy text rather than treating this as generic privacy QA.

I did a fairly deep pass because this feels like a feature where end-to-end behavior matters more than the isolated unit shape.

### What I tested

I checked out the PR branch and ran:

- `git diff --check`
- `npx vitest run test/redteam/plugins/privacyPolicyConsistency.test.ts test/redteam/plugins/index.test.ts --sequence.shuffle=false`
- `npm run test:app -- src/pages/redteam/setup/components/PluginConfigDialog.test.tsx src/pages/redteam/setup/constants.test.ts --run`

I also exercised the feature end to end:

- started the local app/server and verified the redteam setup UI loads
- ran `redteam generate` with the documented local `file://` config path
- ran three real generations with different policy fixtures (support, adtech, SaaS) to inspect prompt quality/diversity
- ran real evals against hand-authored target outputs to check grader behavior, including cases that should fail and a valid privacy-rights request that should not be treated as safe just because the model refused

### Must fix before merge

I found three issues that I think should block landing this version:

1. The documented local `file://` configuration path does not work in the real CLI flow. The docs tell users to use `privacyPolicy: file://privacy-policy.md`, but by the time plugin validation runs in `redteam generate`, the generic config loader has already resolved the file reference to raw text, so validation rejects the documented config.
2. Uploaded policy content leaks back into generated YAML even though this plugin tries to strip it from metadata. In real generated output, `privacyPolicyContent` still appeared in every test case because the original plugin config is merged back in later.
3. The grader currently false-passes any basic refusal before the privacy-policy rubric runs. That means a legitimate supported request, such as a verified user asking for an export that the policy allows, can pass even when the target simply says “I cannot help with that request.” For this plugin, unsupported stonewalling is itself a policy-consistency problem and needs to be gradable.

I left inline comments on each of those code paths.

### Product / quality notes

I think the product framing is sound, but the current generation strategy is weaker than the grading strategy. Right now the grader is policy-aware, but generation is intentionally policy-blind. In practice, across three different policy fixtures with the same system purpose, the generated probes clustered around the same small set of generic privacy themes (sharing, deletion, opt-outs, personalization, location), and some outputs drifted outside the supplied policy or became malformed/incomplete.

That leaves a lot of value on the table. A stronger next version would be policy-grounded and diversity-aware, for example:

- extract concrete policy clauses / commitments first, then generate probes against those clauses
- deliberately cover different dimensions instead of near-duplicates: data category, actor, action, workflow, disclosure mode, explanation vs action, supported right vs unsupported request
- include both behavior probes and explanation probes, because privacy notices can be violated by doing the wrong thing or by saying the wrong thing
- avoid inventing unsupported domains that are absent from the policy unless the intended probe is explicitly about unsupported behavior
- add grader examples for valid refusal, invalid refusal of a supported right, unsupported claim of compliance, omission/misstatement, and trace evidence contradicting text

On the policy side, the docs are directionally good. Typical privacy notices usually need categories of data, sources, purposes, sharing/recipients, retention, and rights; depending on jurisdiction and product, they may also need controller/contact details, transfers, automated decision-making, and child-specific handling. I would keep this plugin clearly framed as a **consistency check against the supplied notice**, not as a legal-compliance oracle.

### UX / DX suggestions

A few smaller things I would tighten while this is still fresh:

- require config consistently across CLI, server, and UI surfaces
- do not silently drop the plugin in `redteam init` after the user selected it
- use a single-line file-reference control locally rather than a large textarea for a `file://` path
- make validation explicit and consistent: missing, wrong scheme, unreadable file, empty content
- document the remote-generation choice clearly. Since policy text is excluded from generation, `canGenerateRemote = false` is a noticeable UX tradeoff without an obvious privacy win unless we plan to make generation policy-aware later

### Overall

I like the direction and I think this is worth continuing. Before merge, though, I would want the three correctness issues above fixed, plus at least one end-to-end regression test that exercises the real config-loading/export path instead of only the plugin in isolation. Once those are addressed, I think the most valuable follow-up is improving policy grounding and prompt diversity so this becomes meaningfully more relevant than a generic privacy prompt pack with a custom scorer afterward.

mldangelo-oai:
Thanks again for putting this together — this is great work and I think we can get it merged tomorrow once the review items are addressed. Could you also please resolve the current conflicts with `main` (or just merge `main` into this branch) when you do the next pass?

PR #9115, opened 2026-05-05
chore(redteam): default setup skill to meta

ianw-oai:


plugins/promptfoo/skills/promptfoo-redteam-setup/SKILL.md
```diff
@@ -94,11 +94,11 @@ a resolved Promptfoo Cloud policy object.
 
 ### 4. Choose strategies conservatively
 
-- Use `basic` for the first setup/generation pass.
```
mldangelo-oai, line 97: I spent some time vibe checking this when I wrote it, if you are sure you don't want the multi pass approach I am happy to defer to you

mldangelo-oai (approved):

PR #9124, opened 2026-05-06
fix(cache): stabilize Anthropic cache hashes across runs

wholley-oai:
## Summary
- keep Anthropic cache-key hashing stable across fresh processes
- fingerprint Anthropic credentials without embedding raw API keys in cache entries
- add a regression test that compares hashes produced by two separate Node processes

## Why
`d1f1e5b48` introduced per-process random HMAC material for Anthropic cache keys. That kept secrets out of the key string, but it also meant the same request generated a different disk-cache key after each fresh process start, so later CLI runs could not reuse prior Anthropic cache entries.

## Validation
- `npm_config_cache=/tmp/promptfoo-npm-cache npx vitest run test/providers/anthropic/completion.test.ts test/providers/anthropic/messages.test.ts --sequence.shuffle=false`
- `npm_config_cache=/tmp/promptfoo-npm-cache npm run l`
- `npm_config_cache=/tmp/promptfoo-npm-cache npm run f`

### DIFF
--- a/src/providers/anthropic/generic.ts
+++ b/src/providers/anthropic/generic.ts
@@ -40,17 +40,28 @@ interface AnthropicBaseOptions {
 
 const ANTHROPIC_CACHE_HASH_CONTEXT = 'promptfoo:anthropic:cache-key:v1';
 
-// Canonicalize before hashing so semantically identical objects with different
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thank you!

PR #9149, opened 2026-05-07
fix(redteam): handle MCP target prompt materialization

jameshiester-oai:
## Summary

Fixes redteam scans against MCP targets when generated probes or runtime attack strategies produce plain text instead of MCP tool-call JSON.

- Treat MCP `config.tools` string and string-array values as tool-name filters instead of OpenAI-style tool definitions during grading.
- Wrap MCP target providers during redteam execution so Basic, Meta, Hydra, and other strategies materialize target prompts at the provider boundary.
- Require inference-based materialization for non-JSON or schema-invalid MCP target prompts instead of guessing tool names or arguments deterministically.
- Allow already-valid MCP tool-call JSON to pass through without inference.
- Allow the MCP provider to parse the direct rendered prompt when `context.vars.prompt` is not populated.

## Root Cause

MCP provider configs use `tools` as a target-side filter, but redteam grading treated `provider.config.tools` as a list of LLM tool definitions and attempted to load it as such. Separately, MCP targets require JSON tool-call payloads, but redteam strategies like Hydra and Meta can hand natural-language prompts to `context.originalProvider` at runtime.

## Behavior

When a redteam MCP target receives a valid MCP tool-call JSON payload, promptfoo passes it through. When the prompt is plain text or schema-invalid MCP JSON, promptfoo now asks the redteam inference provider to produce a schema-valid MCP tool call. If no inference provider is available, the target call fails clearly with a materialization error instead of silently guessing a tool or argument mapping.

## Validation

- `npx vitest run test/redteam/mcpMaterialization.test.ts test/redteam/mcpTargetProvider.test.ts test/redteam/plugins/base.test.ts test/redteam/commands/generate.test.ts test/providers/mcp/util.test.ts` passed: 5 files, 231 tests.
- `npx prettier --check ...` passed for changed files.
- `npm run tsc -- --noEmit` still fails on existing Claude SDK type issues around `skills` and `origin`; no MCP-related type errors were reported.

## Notes

An earlier local end-to-end MCP scan exposed the missing-inference-provider failure path. After tightening this PR to require inference materialization rather than deterministic fallback, that local no-key scan is expected to fail unless a redteam inference provider is configured.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -26,6 +26,7 @@ import { CIProgressReporter } from './progress/ciProgressReporter';
 import { maybeEmitAzureOpenAiWarning } from './providers/azure/warnings';
 import { providerRegistry } from './providers/providerRegistry';
 import { isPromptfooSampleTarget } from './providers/shared';
+import { maybeWrapMcpProviderForRedteam } from './redteam/mcpTargetProvider';
 import { redteamProviderManager } from './redteam/providers/shared';
 import { throwIfTargetPromptExceedsMaxChars } from './redteam/shared/promptLength';
 import { getSessionId } from './redteam/util';
@@ -834,13 +835,17 @@ async function callActiveProvider({
   traceContext: Awaited<ReturnType<typeof generateTraceContextIfNeeded>>;
   vars: Vars;
 }): Promise<ProviderResponse> {
-  const activeProvider = isApiProvider(test.provider) ? test.provider : provider;
+  const originalProvider = maybeWrapMcpProviderForRedteam(provider, test);
+  const activeProvider = maybeWrapMcpProviderForRedteam(
+    isApiProvider(test.provider) ? test.provider : originalProvider,
+    test,
+  );
   logger.debug(`Provider type: ${activeProvider.id()}`);
 
   const callApiContext = buildCallApiContext({
     evalId,
     filters,
-    originalProvider: provider,
+    originalProvider,
     promptForRender,
     repeatIndex,
     test,
--- a/src/providers/mcp/index.ts
+++ b/src/providers/mcp/index.ts
@@ -56,7 +56,7 @@ export class MCPProvider implements ApiProvider {
   }
 
   async callApi(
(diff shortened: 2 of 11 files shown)


src/redteam/mcpMaterialization.ts
```diff
@@ -0,0 +1,206 @@
...
+
+import type { MCPTool } from '../providers/mcp/types';
+import type { ApiProvider } from '../types/index';
+
+type McpToolCall = {
+  tool: string;
+  args: Record<string, unknown>;
+};
+
+const TOOL_NAME_FIELDS = ['tool', 'toolName', 'function', 'functionName', 'name'] as const;
```
jameshiester-oai, line 15: @codex are these and the tool args fields from the MCP spec?

src/providers/mcp/index.ts
```diff
@@ -67,7 +67,8 @@ export class MCPProvider implements ApiProvider {
       // Parse the prompt as JSON to extract tool call information
       let toolCallData: any;
       try {
-        toolCallData = JSON.parse(context?.vars.prompt as string);
+        const rawToolCall = context?.vars?.prompt ?? prompt;
```
zcrab-oai, line 70: fine...

src/evaluator.ts
```diff
@@ -834,13 +835,17 @@ async function callActiveProvider({
   traceContext: Awaited<ReturnType<typeof generateTraceContextIfNeeded>>;
   vars: Vars;
 }): Promise<ProviderResponse> {
-  const activeProvider = isApiProvider(test.provider) ? test.provider : provider;
+  const originalProvider = maybeWrapMcpProviderForRedteam(provider, test);
+  const activeProvider = maybeWrapMcpProviderForRedteam(
```
zcrab-oai, line 839: this looks very codex...

zcrab-oai: i don't like this syntax but maybe it's better for cleaner testing

zcrab-oai (approved):

PR #9207, opened 2026-05-12
feat(providers): support Azure Responses reasoning config

typpo:
## Summary

- add first-class Azure Responses `reasoning` configuration for effort and provider-generated summaries, including templating and custom deployment detection
- preserve legacy and `passthrough.reasoning` precedence, reject malformed non-object configuration, and keep sampling controls only when the effective or documented model-default effort is `none`
- consolidate the Azure documentation around the native configuration and use a summary-capable `o4-mini` example

## Why

Azure Responses already normalized returned reasoning items, but requesting summaries required raw `passthrough` configuration. This adds a native provider surface while retaining the raw escape hatch and leaving Azure chat/completion and non-Azure providers unchanged.

## Validation

- two shuffled runs of six focused provider/processor suites (169 tests per seed)
- deterministic no-credential loopback HTTP and built-CLI QA covering summary and tool output, usage/cost, empty and partial summaries, content filtering, retry, malformed JSON, and serialized eval output
- `npm run build`
- `SKIP_OG_GENERATION=true npm run build` in `site`


mldangelo-oai:
## Comprehensive review result: FIXED_AND_READY

Final head `c7f7c72accbb7027b46995478942e9aebee0040b` was reviewed against current `main` `81adadadd2fd1bb19b80136ebe66c8c64102bc55`.

### What changed

- reconciled current `main` additively and removed the branch conflict without rewriting contributor history
- kept the new first-class Azure Responses `reasoning` configuration while preserving legacy `reasoning_effort` and `passthrough.reasoning` precedence
- preserved `temperature` and `top_p` when the final or documented model-default effort is `none`, while omitting them for active reasoning
- rejected null, string, and array `reasoning` configuration before request serialization
- corrected the Azure docs to use a summary-capable `o4-mini` example, scope sampling behavior to Responses, and consolidate the native configuration with the existing summary documentation

### Verification

- Azure Responses focused suite: 52/52 passed
- two shuffled compatibility runs across six Azure/OpenAI/shared processor suites: 169/169 passed for each seed
- TypeScript, formatting/lint, architecture boundaries, production build, and Docusaurus production build passed; Biome reports only the existing cognitive-complexity warning
- deterministic credential-free loopback exercised 10 real local HTTP requests plus the built CLI, covering summary + tool output, usage/cost, empty and partial summaries, content filtering, 429 retry, malformed JSON, serialization, and no-cache eval output
- fresh independent correctness/API, security/adversarial, runtime/docs, and simplification reviews are clean on the final SHA
- exact-head GitHub CI: all 44 rollup entries successful, 0 pending, 0 failed, 0 cancelled; Codecov checks are green and project coverage increased
- review threads: 2 resolved/outdated, 0 unresolved

The PR is mergeable. Its title and body now describe the final native configuration and validation. The remaining GitHub gate is the required human review; no approval or merge was performed.

One pre-existing boundary remains outside this diff: Azure Responses `stream: true` still reaches the provider's existing JSON response parser rather than an SSE accumulator. The loopback reproduced that base behavior; this PR neither introduces nor worsens it.

src/providers/azure/responses.ts
```diff
@@ -122,7 +158,29 @@ export class AzureResponsesProvider extends AzureGenericProvider {
       input = prompt;
     }
 
-    const isReasoningModel = this.isReasoningModel();
+    const renderedReasoning = renderVarsInObject(config.reasoning, context?.vars);
+    if (
+      renderedReasoning !== undefined &&
+      (renderedReasoning === null ||
```
mldangelo-oai, line 164: **[P2] Preserve the valid `null` reasoning override**

The [Azure Responses request schema](https://learn.microsoft.com/en-us/rest/api/microsoft-foundry/azureopenai/responses) defines `reasoning` as `OpenAI.Reasoning or null`, but this new guard rejects the `null` half of that contract (and `AzureResponsesOptions.reasoning` excludes it too). This matters beyond raw parity: the provider shallow-merges `context.prompt.config` over provider config, so `reasoning: null` is the natural per-prompt way to clear a provider-level reasoning object. Today that valid override throws before the request is built. Please accept and type top-level null, treat it as inactive when deciding custom-deployment promotion/sampling suppression, and preserve the explicit null in the final body so it also clears any lower-precedence reasoning value. A regression should start with provider-level active reasoning, override it with prompt-level null, and assert `reasoning: null` is emitted without stripping `temperature`/`top_p` for a custom deployment.

mldangelo-oai (commented):
Reviewed exact head `c7f7c72accbb7027b46995478942e9aebee0040b` (comment only; I did not modify or push to this branch).

The first-class `azure:responses` reasoning feature has not been superseded on current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`). The two older inline threads remain fixed/outdated. I found one additional P2 contract mismatch: the new modeled field rejects top-level `reasoning: null`, although Azure's Responses request contract explicitly permits it and prompt-level config needs it to clear a provider-level reasoning object.

Local exact-head validation passed: 7 focused Azure/OpenAI/shared Responses test files (235 tests), TypeScript, core package build, app build, and the Docusaurus production build. `git diff --check` and a merge-tree check against current main also passed. GitHub still reports `MERGEABLE`/`BEHIND`: the branch is 61 commits behind, its required checks are green but stale from June 25 under strict freshness, and the required code-owner approval is still outstanding. The two changed-in-both files (`azure.md` and `azure/types.ts`) merge cleanly and no semantic conflict was found, so this looks refreshable after the null case is addressed and current CI reruns.

PR #9234, opened 2026-05-15
fix(fetch): decompress pooled undici responses

### LINKED ISSUE #9214: fetch() with custom undici Agent does not decompress Content-Encoding responses (Brotli/gzip → garbage)
## Summary

After upgrading to undici 7 (PR #7137, shipped in 0.121.x), promptfoo's pooled `fetchWithProxy` produces raw compressed bytes instead of decoded JSON when the remote endpoint returns `Content-Encoding: gzip` or `br`. The user sees `Error parsing response from <url>: Unexpected token '<x>', "<binary>"... is not valid JSON. Received text: <binary>` and every test against that endpoint errors out.

The bug fires whenever OpenAI's CDN compresses the response (Brotli on responses >~200 B), which on web-search-enabled evals is essentially every call.

## Root cause

In `src/util/fetch/fetch.ts` (`fetch-B1Rb0VdL.js` in the bundled build), `getOrCreateAgent` constructs a custom `undici.Agent` for connection pooling and assigns it as `finalOptions.dispatcher`. In undici v7 the `decompress` step moved from a default Agent behavior to an **opt-in interceptor** (`undici/lib/interceptor/decompress.js`). The custom Agent does not install that interceptor, so the dispatcher returns the raw compressed body. The wrapping `fetch()` reads it via `await resp.text()`, runs `JSON.parse(...)` on the raw bytes, and throws.

Plain `fetch()` without a custom dispatcher works because Node's global default dispatcher ships with the decompress interceptor.

## Minimal repro (Node 24+ / undici 7)

```js
const { Agent } = require('undici');

// Mirrors getOrCreateAgent — custom Agent, no interceptors set.
const agent = new Agent({ headersTimeout: 60000, connections: 10 });

const r = await fetch('https://api.openai.com/v1/responses', {
  method: 'POST',
  headers: {
    'Authorization': 'Bearer ' + process.env.OPENAI_API_KEY,
    'Content-Type': 'application/json',
  },
  body: JSON.stringify({
    model: 'gpt-5.5',
    input: [{ role: 'user', content: [{ type: 'input_text', text: 'Reply: hi' }] }],
    max_output_tokens: 50,
  }),
  dispatcher: agent,
});

console.log('content-encoding:', r.headers.get('content-encoding'));   // null (header stripped)
console.log(Buffer.from((await r.text()).slice(0, 16)).toString('hex')); // raw Brotli/gzip bytes
```

vs the same fetch without `dispatcher` — returns clean JSON.

## Impact

- Any provider that hits an OpenAI endpoint via the responses API (`/v1/responses`) or chat completions (`/v1/chat/completions`).
- Particularly nasty on web-research evals (web_search-enabled), where responses are large and always compressed.
- Symptom is the JSON parse error with binary garbage in `Received text:` — the binary is the actual Brotli stream (first byte commonly `0x1B`, `0x0B`, or `0x0F`).
- Triggered by undici v7 upgrade (#7137); does not affect promptfoo versions still on undici v6.

## Workaround

Add `Accept-Encoding: identity` to the provider config. Both the responses and chat-completions providers spread `config.headers` into the outgoing fetch, so this works per-provider:

```yaml
providers:
  - id: openai:responses:gpt-5.5
    config:
      headers:
        Accept-Encoding: identity
```

Server skips compression entirely, sidestepping the bug.

## Suggested fix

Install undici's `decompress` interceptor on the custom Agent in `getOrCreateAgent` and `getOrCreateProxyAgent`:

```js
const { Agent, interceptors } = require('undici');

const agent = new Agent({
  headersTimeout: getRequestTimeoutMs(),
  keepAliveTimeout: 30_000,
  keepAliveMaxTimeout: 60_000,
  connections: concurrency,
  connect: tlsOptions,
}).compose(interceptors.decompress());
```

(or pass `interceptors: [interceptors.decompress()]` to the Agent constructor, depending on undici 7 API).

This restores the parity with plain `fetch()` and removes the need for per-provider `Accept-Encoding: identity` workarounds.

## Environment

- promptfoo: 0.121.11 (also reproduced on 0.121.9 after `npm rebuild` for Node 26)
- undici (bundled): 7.25.0
- Node.js: v26.0.0 (also reproduces on v24.x once you're on undici 7)
- OS: macOS 14.x / Linux

Happy to send a PR with the interceptor wiring if it'd help — the change is small and the repro above can become a regression test.

Gujiassh:
## Summary

- compose promptfoo's pooled Undici Agent with `interceptors.decompress()`
- apply the same decompression interceptor to ProxyAgent-backed dispatchers
- keep caller-provided dispatchers untouched
- add fetch tests that lock the decompression interceptor onto both cached dispatcher types

Fixes #9214.

## Tests

- `npx vitest run test/fetch.test.ts --testNamePattern "decompression|dedicated Agent dispatcher|dedicated ProxyAgent dispatcher|preserve a caller-provided dispatcher"`
- `npm run tsc`
- `npx @biomejs/biome check src/util/fetch/index.ts test/fetch.test.ts`
- `git diff --check`

### DIFF
--- a/src/util/fetch/index.ts
+++ b/src/util/fetch/index.ts
@@ -3,7 +3,7 @@ import path from 'path';
 import type { ConnectionOptions } from 'tls';
 
 import { getProxyForUrl } from 'proxy-from-env';
-import { Agent, ProxyAgent } from 'undici';
+import { Agent, type Dispatcher, interceptors, ProxyAgent } from 'undici';
 import cliState from '../../cliState';
 import { DEFAULT_MAX_CONCURRENCY, VERSION } from '../../constants';
 import { getEnvBool, getEnvInt, getEnvString } from '../../envars';
@@ -33,8 +33,8 @@ import type { FetchOptions } from './types';
 // Note: TLS options (rejectUnauthorized, CA cert) are captured at agent
 // creation time. This is acceptable because these env vars don't change
 // mid-process. If that assumption changes, add cache-invalidation logic.
-const cachedAgents: Map<number, Agent> = new Map();
-const cachedProxyAgents: Map<string, ProxyAgent> = new Map();
+const cachedAgents: Map<number, Dispatcher> = new Map();
+const cachedProxyAgents: Map<string, Dispatcher> = new Map();
 
 /**
  * Get the connection pool size for HTTP agents.
@@ -72,7 +72,7 @@ export function clearAgentCache(): void {
   cachedProxyAgents.clear();
 }
 
-function getOrCreateAgent(tlsOptions: ConnectionOptions): Agent {
+function getOrCreateAgent(tlsOptions: ConnectionOptions): Dispatcher {
   const concurrency = getConnectionPoolSize();
   const existing = cachedAgents.get(concurrency);
   if (existing) {
@@ -84,7 +84,7 @@ function getOrCreateAgent(tlsOptions: ConnectionOptions): Agent {
     keepAliveMaxTimeout: 60_000,
     connections: concurrency,
     connect: tlsOptions,
-  });
+  }).compose(interceptors.decompress());
   cachedAgents.set(concurrency, agent);
   return agent;
 }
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thanks for jumping on this so quickly. I reproduced the underlying Node 26 + Undici 7 failure locally and confirmed that this interceptor wiring fixes the pooled-dispatcher path for compressed successful responses, which is the core regression reported in #9214. Really appreciate the focused patch and the targeted coverage here.

I’m going to land this smaller dispatcher-level fix first, then follow up separately on the remaining compressed non-2xx response behavior so we fully close out the edge case.

PR #9237, opened 2026-05-15
fix(providers): refresh xai grok support

mldangelo-oai:
## Summary
- preserve Grok 4.3 reasoning effort support in chat requests and keep Responses streaming output limited to final answer deltas
- add Grok Imagine quality-model support with redirect-aware fallback pricing coverage
- refresh xAI docs and examples with evergreen model-selection and legacy-alias guidance

## Verification
- `./node_modules/.bin/vitest run test/providers/xai/chat.test.ts test/providers/xai/responses.test.ts test/providers/xai/image.test.ts --sequence.shuffle=false`
- `npm run tsc`
- `npm --prefix site run build`
- `npm run l`
- `npm run f`
- `git diff --check`

## Notes
- The docs build completed successfully; external stats fetches fell back to the repo's built-in fallback values before Docusaurus build work proceeded.
- `npm run l` / `npm run f` completed with the repository's existing cognitive-complexity warning class on touched xAI provider methods and no blocking lint or formatting errors.

### DIFF
--- a/examples/xai/chat/README.md
+++ b/examples/xai/chat/README.md
@@ -75,7 +75,7 @@ The recommended starting point for general text workflows:
 
 ### Legacy Model Note
 
-xAI periodically retires older model slugs and may keep them working through redirects to newer replacements. This example sticks to current Grok 4.3 and Grok 4.20 families so it remains a useful starting point for new evaluations.
+xAI periodically retires older model slugs and may keep them working through redirects to newer replacements. This example uses Grok 4.3 plus alias-style Grok 4.20 family IDs, matching xAI's guidance for configs that should track the current release within a family.
 
 ### Agent Tools (Responses API)
 
--- a/examples/xai/chat/README.md
+++ b/examples/xai/chat/README.md
@@ -32,7 +32,7 @@ promptfoo view
 
 This example includes configurations to test different Grok capabilities:
 
-- **Text Generation** (`promptfooconfig.yaml`) - Mathematical reasoning with Grok 4.3, Grok 4.20, Grok 4.1 Fast, Grok 4 Fast, Grok 4, and Grok 3 models
+- **Text Generation** (`promptfooconfig.yaml`) - Mathematical reasoning with the current Grok 4.3 and Grok 4.20 families
 - **Image Generation** (`promptfooconfig.images.yaml`) - Artistic image creation using Grok's image models
 - **Search Tools** (`promptfooconfig.search.yaml`) - Real-time web and X search using the Responses API
 - **Agent Tools (Responses API)** (`promptfooconfig.responses.yaml`) - Autonomous web and X search using Agent Tools
@@ -64,6 +64,7 @@ promptfoo eval -c promptfooconfig.promptfoo-search.yaml
 The recommended starting point for general text workflows:
 
 - `xai:grok-4.3` - General-purpose reasoning model
+- `reasoning_effort` - Supports `none`, `low`, `medium`, and `high` in chat configs
 - `xai:responses:grok-4.3` - Recommended form for server-side tools
 
 ### Grok 4.20
@@ -72,25 +73,9 @@ The recommended starting point for general text workflows:
 - `xai:grok-4.20-non-reasoning` - Non-reasoning model
 - `xai:grok-4.20-multi-agent` - Multi-agent variant
 
-### Grok 4.1 Fast
+### Legacy Model Note
 
-A frontier model optimized for agentic tool calling with a 2M context window:
-
-- `xai:grok-4-1-fast-reasoning` - Maximum intelligence with reasoning
-- `xai:grok-4-1-fast-non-reasoning` - Fast responses without reasoning
-
-### Grok 4 Fast
-
-Fast reasoning models with 2M context:
-
-- `xai:grok-4-fast-reasoning` - Reasoning variant
-- `xai:grok-4-fast-non-reasoning` - Non-reasoning variant
-
-### Grok 4
-
-Flagship reasoning model:
-
-- `xai:grok-4` - Full reasoning capabilities
+xAI periodically retires older model slugs and may keep them working through redirects to newer replacements. This example sticks to current Grok 4.3 and Grok 4.20 families so it remains a useful starting point for new evaluations.
 
 ### Agent Tools (Responses API)
 
--- a/examples/xai/chat/promptfooconfig.yaml
+++ b/examples/xai/chat/promptfooconfig.yaml
@@ -14,19 +14,19 @@ providers:
       max_completion_tokens: 4096
 
   # Grok 4.20 reasoning family
-  - id: xai:grok-4.20-0309-reasoning
+  - id: xai:grok-4.20-reasoning
     config:
       temperature: 0.7
       max_completion_tokens: 4096
 
   # Grok 4.20 non-reasoning family
-  - id: xai:grok-4.20-0309-non-reasoning
+  - id: xai:grok-4.20-non-reasoning
     config:
       temperature: 0.7
       max_completion_tokens: 4096
 
   # Grok 4.20 multi-agent family
-  - id: xai:grok-4.20-multi-agent-0309
+  - id: xai:grok-4.20-multi-agent
     config:
       temperature: 0.7
       max_completion_tokens: 4096
--- a/examples/xai/chat/promptfooconfig.yaml
+++ b/examples/xai/chat/promptfooconfig.yaml
@@ -10,36 +10,26 @@ providers:
   - id: xai:grok-4.3
     config:
       temperature: 0.7
-      max_completion_tokens: 4096
-
-  # Grok 4.1 Fast Models (optimized for agentic tool calling)
-  - id: xai:grok-4-1-fast-reasoning
-    config:
-      temperature: 0.7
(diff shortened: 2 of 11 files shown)


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/236145ba-ba3c-4040-a5e1-f36df710a704?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on 9b1fca89b1a52ece7de8cd1d7dca52f750ee52ac. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 84dbd6da9177b421582650eefbf65ece39aa1806 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/8eb3cf90-ae27-4b82-a3f8-59e20895542a?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 15, 2026 2:28PM |
| b7875171ffeb29c4b22d6711d123badea4c7c95d | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/8e25be07-7f7a-4299-8260-9eff3a9b5151?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 15, 2026 2:38PM |
| 9b1fca89b1a52ece7de8cd1d7dca52f750ee52ac | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/236145ba-ba3c-4040-a5e1-f36df710a704?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 15, 2026 2:51PM |

</p>
</details>

mldangelo:
## Critical review

Tested the branch end-to-end against the real xAI API (`grok-4.3` chat + Responses, with `reasoning_effort` ∈ {`none`, `low`, `high`} and streaming). Behavior is correct:

- `reasoning_effort: none` → 0 reasoning tokens
- `reasoning_effort: low` → ~364 reasoning tokens
- `reasoning_effort: high` → ~644 reasoning tokens
- Streaming Responses output is clean (no reasoning-summary leakage)

Cross-checked the PR's redirect/retirement assumptions against xAI's official May 14 retirement email and `/v1/language-models` + `/v1/image-generation-models`. The email confirms the core premise: retired slugs auto-redirect to `grok-4.3` (reasoning variants → `low` effort; non-reasoning → `none`) and are billed at standard Grok 4.3 pricing.

### Gaps I found and pushed a follow-up commit for (84dbd6da9)

1. **Image-quality aliases silently routed to `grok-2-image`.** The live `/v1/image-generation-models` lists `grok-imagine-image-quality-latest` and `grok-imagine-image-quality-20260403` as canonical aliases. They were missing from `XAIImageProvider.getApiModelName()`'s modelMap, so requests fell through to the `grok-2-image` default — wrong model, wrong fallback price. Added both, plus a pass-through guard so any future `grok-imagine-*` slug stays as-is instead of being silently downgraded.
2. **Redirect set missed several aliases xAI's catalog collapses into the retired ids.** Per `/v1/language-models`, `grok-3` includes `grok-3-beta`, `grok-3-fast`, `grok-3-fast-latest`, `grok-3-fast-beta` aliases, and `grok-code-fast-1` includes a dated `grok-code-fast-1-0825` alias. Without those, configs spelling the legacy variants would have been priced against the old `$3 / $15` per-1M Grok-3 table instead of the post-retirement `$1.25 / $2.50` Grok 4.3 rate.
3. **Docs caution was vague.** Replaced "may keep them working through redirects" with the authoritative behavior from xAI's retirement email — explicit list, the `low` vs `none` reasoning-effort split, and the standard Grok 4.3 billing.

Test coverage added for each.

### Items I flagged but did not change

- The fallback pricing for `grok-imagine-image-quality` uses `$0.05 / 1k` and `$0.07 / 2k` (and `$0.01` per source-image input on edits) from xAI's docs page. The live `/v1/image-generation-models` API reports a single `image_price = $0.04` and `edit_input_image_price: null`. In practice `usage.cost_in_usd_ticks` takes priority for image responses, so the fallback only matters when usage is missing — bounded impact, worth a watch.
- The PR's pro→quality fallback for `grok-imagine-image-pro` is a reasonable bet, but xAI's email confirms retirement of that slug without naming a redirect target. If xAI publishes a definitive image-pro redirect (or starts setting `deprecated`/`retired` on `/v1/image-generation-models`), it would be cleaner to drive this from the API metadata rather than a hard-coded mapping.
- Pre-existing: chat completions never see `cost_in_usd_ticks` because the OpenAI base provider doesn't surface the raw body, so the local table is the source of truth for chat billing. Worth a follow-up to plumb xAI's per-request reported cost through for chat, the way Responses and image already do.

Otherwise LGTM — the reasoning_effort change, the SSE filtering to `response.output_text.delta`, and the annotations preservation on streamed responses are all correct.

mldangelo:
## Follow-up pass: PR review + simplifier findings

Ran both \`pr-review-toolkit:code-reviewer\` and \`pr-review-toolkit:code-simplifier\` over the branch. Pushed [b7875171f](https://github.com/promptfoo/promptfoo/commit/b7875171f) with the load-bearing fix and the simplifications.

### Important: streamed tool calls were being dropped

The code-reviewer caught a real behaviour asymmetry between streaming and non-streaming Responses calls. In \`readStreamingResponse\`, whenever any \`response.output_text.delta\` accumulated, the function returned a synthesized object that **replaced** \`latestResponse.output\` with a single rebuilt assistant \`message\` item. Every \`function_call\`, \`web_search_call\`, \`code_interpreter_call\`, \`mcp_call\`, etc. that the model emitted alongside the text was silently discarded. The non-streamed path forwards those items to \`ResponsesProcessor\`, so \`stream: true\` and \`stream: false\` produced materially different outputs for any tool-using config — and \`functionToolCallbacks\` never fired under streaming.

Fix: keep every non-\`message\`, non-\`reasoning\` output item from \`latestResponse\` and append the rebuilt message. \`reasoning\` is still filtered out (that was the whole point of the original change in this PR), but tool-call items now survive.

Verified against the real xAI API with \`xai:responses:grok-4.3\` + \`stream: true\` + \`tools: [{type: web_search}]\`: the output now contains the \`Web Search: "..."\` and \`Opening page: ...\` narratives alongside the final answer (matching non-streamed runs). Added a regression test that feeds a streamed \`response.completed\` payload containing both a \`web_search_call\` and a message output and asserts the tool-call content survives.

### Simplifications (behaviour-preserving)

- \`src/providers/xai/chat.ts\`: \`callApi\` had two parallel 14-line blocks for reasoning-token extraction — one under \`typeof response.raw === 'string'\`, one under \`typeof response.raw === 'object'\`. Collapsed into a single parse step that produces \`rawData\` either way, followed by one shared extraction block. Net -37 lines.
- \`src/providers/xai/image.ts\`: \`calculateImageCost\` used a nested ternary to compute \`mediaInputCost\` that was only consumed inside two branches that already knew their \`pricingModel\`. Inlined the constants into each branch and named the per-image rate.
- \`src/providers/xai/responses.ts\`: extracted the streaming annotation \`flatMap\` chain into a small \`collectOutputTextAnnotations\` helper with a doc comment.

### Final state

- 233 / 233 xAI provider tests passing
- \`npm run l\`, \`npm run f\`, \`npm run tsc\` clean (only the pre-existing complexity warnings the PR description already calls out)
- Verified end-to-end against real xAI: \`grok-4.3\` chat reasoning_effort scales correctly, Responses streaming with reasoning is clean, Responses streaming with \`web_search\` tools preserves the tool-call narrative.

mldangelo:
### Re: Codex P2 finding — keeping the strip, with a code comment

Looked at the Codex suggestion to forward `reasoning_effort` on the legacy aliases that redirect to `grok-4.3`. Tested live before changing anything:

\`\`\`
xai:grok-4-fast-reasoning + reasoning_effort: high
→ 400 Client specified an invalid argument
→ "Model grok-4-fast-reasoning does not support parameter reasoningEffort."
\`\`\`

So even though xAI's retirement email says the legacy slugs server-side redirect to `grok-4.3` post-cutoff, the chat-completions endpoint currently rejects the parameter on the legacy names. Forwarding it would break runtime for anyone still on those slugs before the 12:00 PM PT cutover.

Kept the original strip and pushed [9b1fca89b](https://github.com/promptfoo/promptfoo/commit/9b1fca89b) with a code comment recording why — Codex's reasoning was plausible but didn't match live API behavior. Users who want to tune `reasoning_effort` should target `xai:grok-4.3` directly. If xAI's redirect later starts accepting the parameter on the legacy names, we can revisit then.

Re-ran both legacy slugs with `reasoning_effort: none` and `high`: both 200 OK (reasoning tokens reported from xAI's internal default since we stripped the param), and all 233 xAI tests still pass.

PR #9243, opened 2026-05-16
feat(cli): add searchable team selector

ianw-oai:


src/commands/auth.ts
```diff
@@ -161,13 +161,24 @@ async function setupTeamContext(
...
-              value: team.id,
-              description: team.slug,
-            })),
+            source: async (input) => {
+              const searchTerm = input?.trim().toLowerCase();
+              const matchingTeams = searchTerm
+                ? organizationTeams.filter(
+                    (team) =>
+                      team.name.toLowerCase().includes(searchTerm) ||
+                      team.slug.toLowerCase().includes(searchTerm),
```
mldangelo-oai, line 172: The new searchable picker only matches team name and slug, but this same auth flow already treats team IDs as valid selectors (`--team` resolves `team.id`, and the help/docs describe name, slug, or ID). A user who pastes a valid team ID here gets no match even though `promptfoo auth login --team <id>` works. Please include `team.id` in this predicate and cover that case in the new source test.

mldangelo-oai (commented):
This is a thoughtful usability improvement for larger team lists. I left one selection-contract note inline: the new interactive search matches names and slugs, but the same auth flow already accepts team IDs elsewhere. Supporting IDs in the search predicate too would keep the experience consistent with the documented CLI behavior.

PR #9244, opened 2026-05-16
fix(deps): avoid incompatible npm release-age config

ianw-oai:
Remove the project-level npm release-age setting because npm rejects it when another supported install cutoff is already configured, which can prevent routine npm commands from starting.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,1 @@
 engine-strict=true
-# Global floor for manual npm installs; Renovate rules should not undercut it.
-min-release-age=7


.npmrc
```diff
@@ -1,3 +1 @@
 engine-strict=true
```
mldangelo-oai, line 1: Removing `min-release-age=7` drops a repo-wide supply-chain delay that was added intentionally and strengthened just three days ago. The branch still tells contributors that npm applies this release-age policy (`site/docs/contributing.md`, `AGENTS.md`), and related policy text still assumes the `.npmrc` floor exists. Please preserve an equivalent control or update the related policy/docs in the same change.

mldangelo-oai (approved):

mldangelo-oai (changes requested):
I agree the npm incompatibility is real and worth fixing. The current patch goes broader than that, though: it removes the repo-level release-age guard outright while related contributor/setup guidance still describes that policy as active. I left an inline comment with the specific concern. I would rather preserve an equivalent control or update the surrounding policy/docs in the same change before merging this.

mldangelo-oai (approved):
Over-zealous agent requested changes. This is good.

PR #9245, opened 2026-05-16
fix(eval): allow header horizontal scrolling

ianw-oai:
Make the eval results header horizontally scrollable while keeping it synchronized with the table body.

### DIFF
--- a/src/app/src/pages/eval/components/ResultsTable.tsx
+++ b/src/app/src/pages/eval/components/ResultsTable.tsx
@@ -1371,6 +1371,7 @@ function ResultsTableHeader({
   tableWidth,
   maxTextLength,
   wordBreak,
+  headerScrollRef,
   theadRef,
   stickyHeader,
   setStickyHeader,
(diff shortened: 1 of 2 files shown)


src/app/src/pages/eval/components/ResultsTable.tsx
```diff
@@ -2147,38 +2156,39 @@ function ResultsTable({
...
-    let rafId: number | null = null;
+    const tableContainer = tableRef.current;
+    const headerContainer = headerScrollRef.current;
 
-    const handleScroll = () => {
-      if (rafId !== null) {
-        cancelAnimationFrame(rafId);
+    const syncScrollLeft = (source: HTMLDivElement, target: HTMLDivElement) => {
+      if (target.scrollLeft !== source.scrollLeft) {
+        target.scrollLeft = source.scrollLeft;
```
mldangelo-oai, line 2168: This direct bidirectional `scrollLeft` copy can snap the body back from its true right edge. The body scroller can own a vertical scrollbar (`overflow: auto`) while the new header scroller is horizontal-only, so their `maxScrollLeft` values can differ. If the body reaches a larger rightmost offset, assigning that value into the header clamps it; the header `scroll` event then writes the smaller clamped offset back into the body. Please normalize against a shared reachable range or otherwise avoid feeding the clamped header value back into the table.

mldangelo-oai (changes requested):
The new header scroller addresses a real layout issue, but I found a scroll-sync edge case that can snap the table body back from the far-right position when the body and header containers have different reachable scroll ranges. I left the concrete path inline. I would fix that interaction before merging this.

mldangelo-oai (approved):

PR #9246, opened 2026-05-16
fix(app): sort custom metrics before truncation

ianw-oai:


mldangelo-oai (changes requested):
Sorting before truncation fixes the ordering bug, but the current implementation also makes collapsed views sort the full metric list on every render. That is already called out in the unresolved inline thread on this PR, and I agree it is material because this component appears in hot table paths. I would tighten that before merging.

mldangelo-oai (approved):

PR #9247, opened 2026-05-16
chore(app): move zoom into table settings

ianw-oai:
Move the eval table zoom control into Table Settings so the toolbar stays focused on primary table actions.

Table Settings zoom control

### DIFF
--- a/src/app/src/pages/eval/components/TableSettings/components/SettingsPanel.tsx
+++ b/src/app/src/pages/eval/components/TableSettings/components/SettingsPanel.tsx
@@ -1,8 +1,24 @@
 import React, { useCallback, useState } from 'react';
 
+import { Select, SelectContent, SelectItem, SelectTrigger } from '@app/components/ui/select';
 import { useResultsViewSettingsStore } from '../../store';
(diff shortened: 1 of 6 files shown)


src/app/src/pages/eval/components/TableSettings/components/SettingsPanel.tsx
```diff
@@ -140,6 +163,22 @@ const SettingsPanel = () => {
         />
 
         <div className="mt-3">
```
mldangelo-oai, line 165: Since this PR moves Zoom out of the toolbar and into Table Settings, can we update `site/docs/usage/web-ui.md` in the same change? It still lists Zoom under `Toolbar` and the `Table Settings` section does not mention the new location.

mldangelo-oai (commented):
Nice cleanup of the toolbar layout. I left one docs-alignment note inline: the web UI guide still tells users Zoom lives in the toolbar, but this PR moves it into Table Settings. Updating that page alongside the UI move would keep the shipped behavior and docs in step.

mldangelo-oai (approved):

PR #9248, opened 2026-05-16
fix(app): improve custom metrics dialog readability

ianw-oai:
Group prompt metrics under readable headers, preserve single-provider ergonomics, and fix grouped table header spans.


mldangelo-oai (changes requested):
The grouped metrics presentation is a clear readability improvement. I rechecked the live branch tip after it moved and one interaction problem remains: the Columns menu is still exposing grouped header ids rather than user-facing labels. That matches the unresolved inline thread already on the PR, so I am not duplicating it, but I do think it should be fixed before merge.

mldangelo-oai (approved):

PR #9249, opened 2026-05-16
fix(app): align table settings info buttons

ianw-oai:
Keeps the Table Settings info buttons beside their labels, and prevents help-button clicks from toggling the parent setting.

Updated Table Settings dialog

Checked with:
- `../../node_modules/.bin/vitest src/pages/eval/components/TableSettings/components/CompactToggle.test.tsx src/pages/eval/components/TableSettings/components/SettingsPanel.test.tsx --run`


mldangelo-oai (approved):
I checked the change carefully and did not find a PR-introduced bug. The click-propagation fix matches the component intent, and the alignment tweak is nicely contained. The new test for info-button behavior is useful too.

mldangelo-oai (approved):
Approved. Small and well-contained UI fix.

PR #9250, opened 2026-05-16
feat(providers): normalize OpenCode skill usage

ianw-oai:
Normalize OpenCode SDK native skill tool parts into `metadata.skillCalls` so `skill-used` works for OpenCode, and document the same skill-eval workflow alongside Claude and Codex.


src/providers/opencode-sdk.ts
```diff
@@ -1370,16 +1377,45 @@ export class OpenCodeSDKProvider implements ApiProvider {
...
+      const skillDir =
+        typeof part.state?.metadata?.dir === 'string' ? part.state.metadata.dir.trim() : '';
+
+      return [
+        {
+          name: skillName,
+          input: part.state?.input,
+          ...(skillDir ? { path: path.join(skillDir, 'SKILL.md') } : {}),
+          source: 'tool',
+          ...(part.state?.status === 'error' ? { is_error: true } : {}),
```
mldangelo-oai, line 1413: Because `skill-used` matches every `metadata.skillCalls` entry by name and never filters `is_error`, this new error-path entry can satisfy `skill-used` even when OpenCode failed to load the skill. Please either keep failed attempts out of `skillCalls` or make the assertion ignore errored entries.

mldangelo-oai (changes requested):
Thanks for adding first-class OpenCode skill normalization. I found one correctness issue before this is safe to merge: errored skill tool attempts are now surfaced in `metadata.skillCalls`, which means the existing `skill-used` assertion can pass even when the skill failed to load. I left an inline comment on the exact spot. Once that success/error boundary is tightened, this looks much sturdier.

mldangelo-oai (approved):

PR #9251, opened 2026-05-16
docs(redteam): tighten plugin docs

ianw-oai:
Tighten red-team plugin docs by replacing repetitive framing with plugin-specific guidance.


site/docs/red-team/plugins/rag-document-exfiltration.md
```diff
@@ -59,12 +59,6 @@ The plugin assesses the AI's response based on:
...
 ## Importance in Gen AI Red Teaming
 
-Testing for document exfiltration vulnerabilities is a critical aspect of generative AI security and robustness evaluation. It helps ensure that AI systems:
-
-- Protect sensitive information
-- Maintain strict access controls
-- Prioritize data security over convenience
-
-By incorporating the RAG Document Exfiltration plugin in your LLM red teaming strategy, you can identify and address potential vulnerabilities in your AI system's ability to handle sensitive document content securely.
+Document exfiltration testing matters when retrieval quality and disclosure limits pull in opposite directions. This plugin checks whether the system summarizes or answers within scope instead of dumping the underlying source material.
```
mldangelo-oai: This weakens the boundary the grader actually enforces. The exfiltration rubric fails any unauthorized reveal of sensitive document content; it is not limited to dumping whole source material, and an unauthorized summary can still be an exfiltration failure. Could we rephrase this around unauthorized disclosure rather than summarization as the dividing line?

site/docs/red-team/plugins/hallucination.md
```diff
@@ -5,17 +5,15 @@ description: Red team LLM hallucination vulnerabilities by testing fabrication t
 
 # Hallucination Plugin
 
-The Hallucination red teaming plugin is designed to test whether a large language model can be led to fabricate answers or generate false information.
-
-This plugin helps identify potential vulnerabilities where the AI might produce incorrect or misleading content.
+Use the Hallucination plugin to test whether a model fabricates specifics when a prompt sounds ordinary but asks for information that is unlikely, unverifiable, or false. It targets overconfident answers, not every factual mistake.
```
mldangelo-oai: This narrows the plugin more than the current grader does. The rubric still checks factual accuracy, unsupported claims, inconsistency, misinterpretation, fabricated details, and any significant hallucination, not only overconfident fabrication. Could we soften this sentence so readers do not infer that ordinary false but materially misleading answers fall outside the plugin scope?

site/docs/red-team/plugins/shell-injection.md
```diff
@@ -5,17 +5,15 @@ description: Red team shell injection vulnerabilities by testing AI systems' res
...
 
-The Shell Injection plugin helps to:
+It focuses on whether the system:
 
-1. Assess the AI's ability to recognize and prevent attempts to execute shell commands.
-2. Identify potential vulnerabilities where the system might be tricked into running unauthorized commands.
-3. Test the AI's response to malicious input disguised as part of legitimate queries or commands.
+1. Recognizes command payloads embedded in otherwise legitimate requests.
+2. Avoids constructing or appearing to execute unauthorized shell operations.
+3. Responds without exposing system details that make follow-on exploitation easier.
```
mldangelo-oai: This overstates the grader contract. The shell-injection rubric explicitly treats refusals that mention system details, explain system architecture, or describe tools/capabilities as PASS, and says the deciding question is only whether the output looks like actual command execution/results. Could we narrow this sentence to the behavior the grader actually flags?

mldangelo-oai (changes requested):
Thanks for tightening these pages up. The rewrite is noticeably clearer overall, but I found three spots where the new prose drifts from the graders that actually define plugin behavior, so I left inline comments before merge: Hallucination now reads narrower than its rubric, RAG Document Exfiltration frames summarization as the safe boundary even though unauthorized summaries can still fail, and Shell Injection promises system-detail handling that the grader explicitly treats as PASS in a refusal. Once those are aligned with the current contracts, this should be in strong shape.

mldangelo-oai (approved):

PR #9254, opened 2026-05-16
fix(openclaw): support protocol negotiation for versions 3 and 4

Yenwod:
## Summary

The OpenClaw provider previously hard-coded protocol version 3 by sending:

```javascript
const OPENCLAW_PROTOCOL_VERSION = 3;
...
minProtocol = OPENCLAW_PROTOCOL_VERSION
maxProtocol = OPENCLAW_PROTOCOL_VERSION
```

Recent OpenClaw versions require protocol 4, which caused the provider to fail
with a protocol mismatch during the WebSocket handshake.

This change introduces separate min and max protocol constants and advertises
support for protocol versions 3 through 4, allowing the gateway to negotiate
the highest mutually supported version.

## Testing

- Confirmed that protocol 3 only reproduces the protocol mismatch.
- Confirmed that advertising support for versions 3–4 successfully connects,
  pairs the device, and passes Promptfoo evaluations against an OpenClaw
  gateway requiring protocol 4.

### DIFF
--- a/src/providers/openclaw/agent.ts
+++ b/src/providers/openclaw/agent.ts
@@ -17,7 +17,8 @@ import type { ApiProvider, ProviderOptions, ProviderResponse } from '../../types
 import type { OpenClawDeviceIdentity } from './device-auth';
 import type { OpenClawConfig } from './types';
 
-const OPENCLAW_PROTOCOL_VERSION = 3;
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you @Yenwod

PR #9255, opened 2026-05-17
fix: isolate loadFunction cache entries

pragnyanramtha:
## Summary
- cache loaded functions by resolved file path instead of the raw relative file path
- include the selected export mode/default function name in the cache key
- add regression coverage for basePath and defaultFunctionName collisions

## Root cause
loadFunction previously keyed cached functions by the unresolved filePath and optional explicit functionName. Reusing the same relative filename from different basePath values, or loading the same module with different defaultFunctionName values, could return a stale function from the first call.

## Validation
- npx vitest run test/util/functions/loadFunction.test.ts
- npx @biomejs/biome check src/util/functions/loadFunction.ts test/util/functions/loadFunction.test.ts
- npm run tsc


mldangelo-oai (approved):
Thank you!

PR #9286, opened 2026-05-19
fix(providers): execute Anthropic MCP tool results

jameshiester-oai:
## Summary
- Execute Anthropic MCP-backed `tool_use` blocks with `MCPClient.callTool`
- Continue the Anthropic Messages conversation with `tool_result` blocks until a final response or `max_tool_calls` cap
- Preserve existing non-MCP tool behavior and add regression coverage for success, pass-through, and loop-cap cases

## Testing
- `npx vitest run test/providers/anthropic/messages.test.ts --sequence.shuffle=false` passed before refreshing dependencies
- `npm run l` passed with existing `callApiInternal` complexity warning
- `npm run f` passed with existing `callApiInternal` complexity warning
- `git diff --check` passed

## Notes
- After fast-forwarding from latest `main`, `npm run tsc` initially failed because `@inquirer/search` was missing from stale local `node_modules`.
- Running `npm ci` to refresh dependencies was blocked by the OpenAI Socket policy for newly published `ws@8.20.1`, which then left local `node_modules` incomplete. Subsequent Vitest/TypeScript runs could not start in this workspace.

### DIFF
--- a/examples/anthropic/mcp/README.md
+++ b/examples/anthropic/mcp/README.md
@@ -1,0 +1,48 @@
+# anthropic/mcp (Anthropic Messages + MCP tools)
+
+This example wires Claude up to a [Model Context Protocol](https://modelcontextprotocol.io) server through the Anthropic Messages provider. Promptfoo exposes the MCP server's tools to Claude, executes any `tool_use` blocks the model emits, and feeds the `tool_result` back into the conversation until Claude produces a final reply.
+
+```bash
+npx promptfoo@latest init --example anthropic/mcp
+cd anthropic/mcp
+```
+
+## Setup
+
+```bash
+export ANTHROPIC_API_KEY=sk-ant-...
+npx promptfoo@latest eval
+```
+
+The bundled config uses the public [deepwiki MCP server](https://mcp.deepwiki.com/) so the example works out of the box without installing anything. To swap in your own server, replace the `mcp` block:
+
+```yaml
+mcp:
+  enabled: true
+  # Local stdio server
+  server:
+    command: npx
+    args: ['-y', '@modelcontextprotocol/server-filesystem', '/tmp/workspace']
+  # …or remote SSE / streamable HTTP
+  # servers:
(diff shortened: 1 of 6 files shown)


mldangelo:
Pushed three small follow-ups while QA'ing this PR end-to-end:

**Refactor (60ccbac0b)** — Collapsed the stream/non-stream branches in `callApiInternal`. They were doing the same ~60 lines of work after `resolveMcpToolUse` returned (cache, output build, refusal handling, cost). Extracted a `getAnthropicCostFromMessage` helper that's now reused by the cached/success/error paths.

**Fix bundled in the same refactor** — The `max_tool_calls` error path returned `tokenUsage` but no `cost`, which silently dropped cost data when a long MCP loop aborted. Added `cost: getAnthropicCostFromMessage(...)` to the error return and asserted `result.cost > 0` in all three `max_tool_calls`-exceeded tests.

**Docs + example (027512b6a, 709cf6468)** — Added a "Model Context Protocol (MCP)" subsection under Tool Calling in `site/docs/providers/anthropic.md` (with `mcp` and `max_tool_calls` in the Options table), removed the stale "use the Agent SDK for MCP" tip, and added `examples/anthropic/mcp/` running against the public deepwiki MCP server so it works out of the box.

**Live eval** — Ran the new example against the real Anthropic API (Sonnet 4.6) + deepwiki MCP:

| Test | Tokens | Cost | Result |
|------|-------:|-----:|--------|
| Transport protocols (multi-round MCP) | 295,331 | \$1.7842 | ✅ |
| Promptfoo architecture (multi-round MCP) | 281,683 | \$1.7047 | ✅ |
| "What does MCP stand for?" (zero MCP rounds) | 2,279 | \$0.0092 | ✅ |

The Codex/Copilot feedback flagged on earlier commits (caching MCP results, mixed-tool follow-ups, forced `tool_choice` clearing) is all addressed by the existing commits on the branch — nothing additional needed there.

PR #9305, opened 2026-05-20
feat(cli): import OpenAI Evals dashboard exports

mldangelo-oai:
## Summary

- import OpenAI Evals dashboard `eval_items_*.jsonl` exports through `promptfoo import`
- preserve dashboard output rows, OpenAI item metadata, samples, grader scores, pass states, grader samples, errors, and token usage in imported Promptfoo results
- represent multi-run exports as prompt columns while keeping rows with colliding run-local indices separate unless both index and source item content match
- document the historical-results import path and its mapping limits

## Why

OpenAI Evals dashboard exports output-item JSONL rows rather than Promptfoo eval JSON. This gives users historical run artifacts that are useful in Promptfoo, but they need an import path that preserves raw OpenAI data and does not infer unsupported assertion definitions from grader result keys.

## Scope

This imports dashboard output-item JSONL exports as historical Promptfoo eval records. It does not reconstruct a runnable Promptfoo config or migrate OpenAI eval definition data such as `testing_criteria`, datasource config/schema, or run pipeline config because those fields are not present in the dashboard JSONL.

## Validation

- `npx vitest run test/commands/import.test.ts --sequence.shuffle=false`
- `npm run l`
- `npm run local -- import --help`
- live Platform dashboard export QA:
  - imported a complex three-run export with nested datasource items, shared-item runs, mismatched same-index rows, `string_check`, `text_similarity`, `label_model`, `score_model`, and `python` grader results
  - verified two shared item pairs align into comparable Promptfoo rows while unrelated same-index rows stay separate
  - verified raw dashboard rows round-trip at `metadata.openai.outputItem` with samples, grader scores, passes, grader samples, projected outputs, finish reasons, and token usage preserved
  - imported additional real dashboard exports covering no-sample results, sampled multi-run rows, and a live `sample.error` row


mldangelo:
# Critical Review

**Verdict:** Solid work overall. CI is green, TypeScript compiles, all 18 tests pass, smoke-tested live against `npm run local -- import` with single-run, multi-run, empty, mixed, and invalid fixtures. The shape of the import (alignment by `data_source_idx` + canonical item hash, error-vs-assert failure distinction, score-only row support) is well thought through and matches what the docs promise.

One **medium-priority correctness concern** worth addressing before merge, plus a few smaller items.

---

## P1 — `promptId` collision across runs silently drops eval→prompt links

`src/commands/import.ts:127-137`
```ts
function createOpenAIEvalsPrompt(runId: string, rows: OpenAIEvalsJsonlRow[]): CompletedPrompt {
  const hasStringInput = rows.every((row) => typeof row.item['input'] === 'string');
  const raw = hasStringInput ? '{{item.input}}' : 'Imported OpenAI eval item';
  const label = hasStringInput ? 'OpenAI eval item input' : 'Imported OpenAI eval item';

  return { raw, label, provider: `openai-evals:${runId}` };
}
```

`generateIdFromPrompt` (`src/models/prompt.ts:24`) hashes `label` first, so every OpenAI-imported prompt with the same input shape produces an **identical `promptId`** regardless of `runId`. Verified by importing the multi-run fixture and inspecting the DB:

```
Prompts:
  { label: 'Imported OpenAI eval item', provider: 'openai-evals:evalrun_error_fixture' }
  { label: 'Imported OpenAI eval item', provider: 'openai-evals:evalrun_pass_fixture' }
Result promptIds (unique): 1  // ['d616e64b...']
```

This trips two things:

1. **`evalsToPromptsTable` silently loses rows.** `src/models/eval.ts:505-511` inserts each prompt with `onConflictDoNothing()` on the composite PK `(evalId, promptId)`. For a multi-run import, only one row lands in `evals_to_prompts`, so cross-eval analytics or queries that fan out via the join table will see N runs but only one prompt link.
2. **`EvaluateResult.promptId` is identical for every result row across runs.** UI paths that key by `promptIdx` (`ResultsCharts.tsx`, `hooks.ts`) are fine, but anything filtering or grouping by `promptId` will conflate runs.

You could argue ''same template across providers = same `promptId`'' is intentional in Promptfoo's model. Fair — but in that scenario the prompts table genuinely holds one row replayed against multiple providers. Here each run is added as a separate `CompletedPrompt` entry, and the data layer silently disagrees.

**Suggested fix** (cheapest): include the runId in the label so each run is a distinct prompt identity, e.g.
```ts
label: hasStringInput
  ? `OpenAI eval item input (${runId})`
  : `Imported OpenAI eval item (${runId})`,
```
Setting an explicit `id` won't work alone — `generateIdFromPrompt` prefers `label` over `id`. Add a regression test asserting `prompts[0].id !== prompts[1].id` for the multi-run fixture and that `evals_to_prompts` has N rows.

---

## P2 — minor cleanups

- **`src/commands/import.ts:289`**: `(JSON.stringify(row.item, null, 2) ?? '{}')` — `JSON.stringify` on a `Record<string, unknown>` produced by `JSON.parse` cannot return `undefined`. The `?? '{}'` is dead. Either drop it or wrap in `try/catch` to defend against BigInt/circular refs that a non-JSON caller could (in theory) hand in.
- **`src/commands/import.ts:447-450`**: `openai-evals-${runIds[0]}` doesn't sanitize the OpenAI run ID. OpenAI's `evalrun_…` IDs are alphanumeric+underscore today, but defensive `runId.replace(/[^A-Za-z0-9_-]/g, '_')` would future-proof against URL/route weirdness if the format ever changes.
- **`metadata.openai.outputItem: row`** (`import.ts:336`) stores the entire row redundantly with `item`, `sample`, `grades`, `passes`, `graderSamples` already split out alongside it. This doubles storage per result for what is effectively the same data. If the goal is forward-compatibility with future OpenAI fields, a one-line comment would help reviewers.
- **`fs.readFile(file, 'utf-8')`** (`import.ts:534`) loads the entire file into memory. Dashboard exports can be tens of MB to GB. Not a blocker (pre-existing behavior for JSON files), but worth a future ticket to stream JSONL.

## P3 — test coverage gaps

Two additional cases worth adding given the failure mode above:
- Assertion that distinct runs in a multi-run import produce distinct `prompts[i].id` and distinct `EvaluateResult.promptId` per `promptIdx`.
- A row with `data_source_idx` of `-0` / fractional / `NaN` to confirm the validator rejects them (currently relying on `Number.isInteger`, which does the right thing — easy belt-and-suspenders test).

---

## What's good

- Strict structural validation (`isOpenAIEvalsJsonlRow`, `isNumberRecord`, `isBooleanRecord`) before treating a file as OpenAI Evals format — no accidental misclassification of a Promptfoo JSON export.
- Alignment key `${data_source_idx}:sha256(canonicalize(item))` correctly keeps same-index different-content rows separate (covered by the dataset-mismatch test).
- Error vs. assertion failure distinction (`ResultFailureReason.ERROR` when `sample.error` is present) preserves semantically correct failure attribution.
- Score-only rows (no `passes`) are explicitly documented and treated as pass with grader scores preserved — matches the doc copy and is tested.
- Telemetry adds `source` so OpenAI-imports are distinguishable from Promptfoo-imports in product analytics.
- Docs in `command-line.md` are honest about what's not migrated (testing criteria, datasource config, run pipeline config).

---

**TL;DR:** Blocker is the `promptId` collision in P1 — one-line fix in `createOpenAIEvalsPrompt`. Everything else is polish. Otherwise a clean, well-tested feature.

ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/10b2c538-2678-4380-8ee8-55a52fa75bd4?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on 692edb11351b4c1ecb4c464a091662b007569eb8. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 692edb11351b4c1ecb4c464a091662b007569eb8 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/10b2c538-2678-4380-8ee8-55a52fa75bd4?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 20, 2026 11:59PM |

</p>
</details>

PR #9308, opened 2026-05-20
fix(code-scan): scope OIDC token to scan subprocess

### LINKED ISSUE #9307: Harden code scan action OIDC token handling
## Summary

The code scan action currently requests a GitHub OIDC token before running `npm install -g promptfoo`, then stores that token on `process.env`. The install subprocess inherits a copied environment, so the OIDC token is available during package installation even though only the later `promptfoo code-scans run` subprocess needs it.

## Why this matters

This is a defense-in-depth hardening opportunity for the action. Package installation and scanner execution have different credential needs, so the action should keep the install environment tokenless and pass the minted OIDC token only to the scan subprocess.

## Proposed change

- Return the OIDC token from the auth helper instead of mutating global process state.
- Build a sanitized base subprocess environment that strips stale `GITHUB_OIDC_TOKEN`, GitHub OIDC request env vars, and action token input env vars.
- Use the sanitized tokenless env for `npm install -g promptfoo`.
- Add the freshly minted `GITHUB_OIDC_TOKEN` only to the scan subprocess env.
- Add tests covering successful OIDC minting, stale inherited token values, and OIDC minting failure.

## Validation

Focused validation should include:

- `npx vitest run test/code-scan-action`
- `npm --prefix code-scan-action run tsc`
- `npm --prefix code-scan-action run build`

fallintoplace:
## Summary

- keep the code scan action's package install subprocess tokenless
- return the GitHub OIDC token from authentication instead of writing it to global process state
- pass the freshly minted `GITHUB_OIDC_TOKEN` only to the `promptfoo code-scans run` subprocess
- add tests covering successful OIDC minting, stale inherited credentials, and OIDC minting failure

Fixes #9307

## Validation

- `npx vitest run test/code-scan-action`
- `npm --prefix code-scan-action run tsc`
- `npm --prefix code-scan-action run build`
- `npm run tsc`


mldangelo (approved):
Thanks for the clean fix here, @fallintoplace 🙏

This correctly scopes the OIDC minting credentials (`ACTIONS_ID_TOKEN_REQUEST_TOKEN` / `_URL`) and any stale `GITHUB_OIDC_TOKEN` out of the `npm install` subprocess, and threading the token through as a return value instead of mutating `process.env` globally is a nice cleanup. The stale-inherited-token test also covers a latent bug the issue didn't even name.

Approving and merging now — we'll layer a bit of extra defense-in-depth on top in a follow-up. Appreciate the contribution!

PR #9310, opened 2026-05-20
fix(eval): preserve eval export import parity

mldangelo-oai:
## Summary

Make `promptfoo export eval` / `promptfoo import` a faithful round trip. The export envelope (`OutputFile`) gains four optional fields — `vars`, `runtimeOptions`, `traces`, and opt-in `blobAssets` — and `import` now restores all of them instead of dropping or re-deriving them.

- **Export** — `promptfoo eval -o` and `promptfoo export eval` now share one envelope builder (`createOutputData`); a new `--include-media` flag embeds referenced blob bytes for portable, self-contained exports.
- **Import** — preserves vars, runtime options, durations, and traces; decodes/validates embedded media; remaps imported trace IDs against target-store collisions.
- **Hardening** — imported blob MIME types are sanitized against a passive-media allowlist (active types like `text/html` / `image/svg+xml` fall back to `application/octet-stream`) so an untrusted export cannot plant same-origin active content; runtime options are sanitized before export; `--force` replacement preflights the artifact before deleting the existing eval, and discloses the loss if a replacement still fails after deletion.
- Trace payloads, embedded media, and runtime options honor the same `PROMPTFOO_STRIP_*` redaction settings as result data.
- Shared blob-reference parsing (`src/blobs/blobRefs.ts`) is now reused by export, import, and share inlining.

## Why

Eval exports are used as portable artifacts for CI, sharing, and migration. The previous round trip could lose fields that import already persisted (vars, runtime options, durations), omitted traces and media needed for parity, and allowed an invalid forced-replacement artifact to delete an existing eval before the import failed. This change narrows those gaps while keeping backwards-compatible import handling for older artifact shapes.

## Format changes

`OutputFile` adds optional `vars: string[]`, `runtimeOptions: Partial<EvaluateOptions>`, `traces: TraceData[]`, and `blobAssets: ExportedBlobAsset[]` (`{ hash, mimeType, sizeBytes, data }`, base64). All four are optional — older exports without them import unchanged, and a payload that is neither a recognizable v3 nor v2 shape is now rejected explicitly. `promptfoo eval -o` and `promptfoo export eval` emit the same envelope. Docs (`command-line.md`, `ci-cd.md`) updated; the `ci-cd.md` `OutputFile` schema was corrected to match the actual v3 shape.

## Data and security notes

- Trace metadata follows the same strip/projection settings as exported result data, so vars/metadata and request/response span bodies do not leak through a new trace side channel.
- Portable media stays opt-in (`--include-media`); embedded blobs are hash-verified, size-capped, and MIME-sanitized on import.
- Imported trace IDs are remapped when a target-store collision would otherwise append spans to another eval's trace.
- Forced replacement validates result, trace/span, and embedded-media shape before deleting the prior eval. Malformed traces/spans/blobs are skipped with a `warn` rather than silently, and a `--force` failure after deletion explicitly tells the user the original eval is gone.

## Validation

- `npm run tsc`, `npm run build`, `npm run l` (clean; 3 pre-existing complexity warnings, none new)
- Backend suites for import/export/output/blobs/traces/sharing/models/importers — all green (700+ tests), including new `test/blobs/blobRefs.test.ts` and new import regression tests (oversized embedded blob, malformed trace, `--force` post-deletion disclosure)
- Smoke test `test/smoke/export-import.test.ts` round-trips a portable eval through real command actions
- Live built-CLI QA with a real provider: export (JSON/YAML) then import into an isolated `PROMPTFOO_CONFIG_DIR` then re-export, with byte-for-byte parity on evalId/author/vars/runtimeOptions/durations/results; collision modes (`--force`, `--new-id`); trace round trips; embedded-media round trips; legacy v2 import; OpenAI Evals JSONL import; and hostile corrupt-artifact cases (bad hashes, oversized blobs, active MIME types)
- Merged latest `origin/main`; full CI green


mldangelo:
## Comprehensive Review — PR #9310

Reviewed at `c29a8d913`. CI green, mergeable, but I'm requesting changes — see P1 below.

### Verification performed
- Read every changed file plus relevant unchanged code (`Eval.create`/`Eval.delete`, `getBlobByHash`, `src/server/routes/blobs.ts`, `genaiTracer.PromptfooAttributes`, `sanitizeRuntimeOptions`, `inlineBlobsForShare`, `extractor`).
- `npm run tsc` clean.
- `npx vitest run` on the PR's 5 test files (133 tests) plus 15 adjacent files (389 tests) — all pass.
- Live built-CLI round trip in isolated `PROMPTFOO_CONFIG_DIR`: baseline eval → export to JSON+YAML → import to fresh dir → re-export. Byte-for-byte parity for `evalId`, `metadata.author`, `metadata.evaluationCreatedAt`, `vars`, `runtimeOptions`, result counts, `durationMs`, `evaluationDurationMs`.
- Collision flows: bare collision fails with correct message; `--force` replaces; `--new-id` mints a new ID.
- Trace round-trip: `traceId` preserved on first import, remapped on `--new-id` reimport, remapped on cross-eval collision (matches the new tests).
- JSONL export still works after the `createOutputData` refactor.
- Adversarial fixtures (corrupt-hash, oversized base64, unsupported-results, orphan-blob, prototype-pollution): all correctly rejected or filtered.
- **XSS reproduction**: confirmed live (details below).

### High-level take

Design is sound and well-tested. The shared `blobRefs.ts` module is a real de-duplication win. The preflight-before-delete ordering for `--force` (`importBlobAssets` → `replaceExistingEval` → create new) is correct and proven by the `embedded media storage failed` test. Most P1/P2s from earlier Codex/Copilot rounds are already resolved in the later commits (`c29a8d913 honor portable export redaction`, `1fdc5597a refactor(import): share eval replacement path`, `ea6344082 guard trace collisions`, `fcaacee35 preflight forced eval replacements`).

That said, one P1 security regression is real and currently exploitable, plus two smaller follow-ups. Not mergeable as-is.

---

## 🔴 P1 — Stored same-origin XSS via `blobAssets` MIME-type injection on import

**Files:** `src/commands/import.ts:126-140`, `src/server/routes/blobs.ts:22,490`

`isImportableBlobAsset` accepts **any** non-empty `mimeType` string. `prepareBlobAssets` → `importBlobAssets` then call `storeBlob(data, asset.mimeType)`, and `recordImportedBlobReferences` writes a `blob_references` row so the blob route will serve it. When the UI server replies, `SAFE_MIME_TYPE_REGEX = /^[a-z]+\/[a-z0-9_+-]+$/i` admits `text/html`, `image/svg+xml`, `application/xhtml+xml`, `text/javascript`, etc., so the response's `Content-Type` is whatever the attacker put in the export.

Reproduced live in a fresh `PROMPTFOO_CONFIG_DIR` against the current branch tip:

```
$ promptfoo import malicious-html-mime.json
Eval with ID eval-html-injection-test has been successfully imported.

$ curl -sI http://localhost:15511/api/blobs/<hash>
HTTP/1.1 200 OK
Content-Type: text/html
Content-Length: 39
Access-Control-Allow-Origin: *
Cache-Control: public, max-age=31536000, immutable

$ curl -s http://localhost:15511/api/blobs/<hash>
<script>alert(document.domain)</script>
```

Same result with `image/svg+xml` + `<svg onload="alert(document.domain)"/>` — served with `Content-Type: image/svg+xml`.

**Impact.** Same-origin script execution in the local Promptfoo UI after a user runs `promptfoo import <file>` on an attacker-supplied export. From that origin, script can read every other eval, the user's config, share tokens, and any other localhost-served data. The blob URL is `Cache-Control: immutable` and stable on the content hash, so it persists for the lifetime of the eval reference.

**Why this is a regression introduced by this PR.** Before this change, blob ingestion only happened through provider code paths where MIME types were normalized (`normalizeAudioMimeType`) or derived from `data:` URLs in provider responses, both of which are governed by trusted code. This PR introduces a brand-new path where untrusted JSON controls both the bytes and the MIME stored alongside them. `promptfoo-scanner` was looking at LLM API call surfaces — it would not catch a parsing/storage security issue like this.

**Fix.** In `isImportableBlobAsset` (or a wrapper before `storeBlob`), restrict imported MIME types to a renderer-safe allowlist (`image/png|jpeg|gif|webp|avif`, `audio/*`, `video/mp4|webm`) and force anything else to `application/octet-stream`. Prefer the explicit allowlist over tightening `SAFE_MIME_TYPE_REGEX` — that regex is currently used for legitimate trusted blobs and shouldn't have to do double duty as an import-time trust boundary. Please add a regression test that imports a file with `text/html` and `image/svg+xml` and asserts the stored MIME is `application/octet-stream`.

---

## 🟡 P2 — `--no-write` evals can leak unserializable runtime options into YAML/TXT exports

**Files:** `src/util/output.ts:200`, `src/models/eval.ts:621`

`createOutputData` writes `...(evalRecord.runtimeOptions && { runtimeOptions: evalRecord.runtimeOptions })` without sanitizing. For evals loaded via `Eval.findById` this is fine because runtime options were sanitized at write time (`Eval.create` line 500, `Eval.save` line 654). For in-memory evals constructed via `new Eval(config, { runtimeOptions })`, the constructor at line 621 just stores `opts?.runtimeOptions` raw — the `--no-write` path in `src/commands/eval.ts:653`, plus any library caller.

Reproduced with a one-shot test:

```ts
const evalRecord = new Eval({}, {
  runtimeOptions: {
    progressCallback: () => {},
    abortSignal: new AbortController().signal,
  } as any,
});
const output = await createOutputData(evalRecord, null);
expect(typeof (output.runtimeOptions as any).progressCallback).toBe('function'); // passes — leak
expect((output.runtimeOptions as any).abortSignal).toBeInstanceOf(AbortSignal);  // passes
```

`JSON.stringify` silently drops the function (leaving an empty `runtimeOptions.abortSignal: {}` artifact); `js-yaml` throws `unacceptable kind of an object to dump [object Function]` and fails the export entirely.

For the standard CLI today the leak is non-exploitable — the `options` literal at `eval.ts:576` doesn't include `progressCallback`, and the `evaluateOptions.abortSignal` mutation at line 657 happens on `evaluateOptions`, not the cloned `options`. But the path is fragile to future additions, and any library caller passing callbacks will trip it.

**Fix.** `runtimeOptions: sanitizeRuntimeOptions(evalRecord.runtimeOptions)` in `createOutputData`. One line, mirrors `Eval.save`/`Eval.create`.

---

## 🟡 P2 — Legacy v2 imports silently drop embedded `blobAssets`

**File:** `src/commands/import.ts:346-352`

`createOutputData` emits `blobAssets` for any export, v2 or v3 (the scan walks the summary regardless of version). On import the gate `importV3 && Array.isArray(evalData.blobAssets) && evalData.blobAssets.length > 0` means v2 imports always skip blob ingestion. Verified live: legacy v2 export with `blobAssets[]` imports cleanly, eval row is created, but `~/.promptfoo/blobs/` is empty and no `blob_references` row exists — the blob URIs in the v2 results 403 from `/api/blobs/<hash>`.

**Fix.** Move `prepareBlobAssets` / `importBlobAssets` / `recordImportedBlobReferences` out of the `if (importV3)` block so v2 imports also restore embedded media, or suppress `blobAssets` emission for v2 exports in `createOutputData`. Option A is true parity; option B is asymmetric but easier. The P1 allowlist fix protects both paths.

---

## 🟢 Minor — `recordImportedBlobReferences` ordering

**File:** `src/commands/import.ts:359-373`

If `EvalResult.createManyFromEvaluateResult` or `importTraces` throws, the prepared blob bytes are already stored but no `blob_references` row exists — dangling blobs. Not a regression and self-healing on the next eval that references the same hash. Lowest priority.

---

## 🟢 Already-addressed earlier review comments (verified)

To save reviewers cycle time, the following Codex/Copilot findings from earlier commits are obsolete against the current tip:

- **Trace span body stripping for `STRIP_PROMPT_TEXT`/`STRIP_RESPONSE_OUTPUT`** — handled in `projectTracesForOutput` at `output.ts:104-124`, covered by `omits stripped prompt and response bodies from trace span attributes`.
- **Trace metadata leak when only `vars` is present** — handled via the `else if` branch with `traceWithoutMetadata` at `output.ts:89-95`, covered by `omits trace metadata when stripped vars are the only metadata`.
- **`STRIP_RESPONSE_OUTPUT` + `--include-media` embedding response bytes** — handled by `resultsForMediaExportScan` at `output.ts:129-153`, covered by `should not embed response blob bytes when response output stripping is enabled`.
- **Orphan blobs in imports** — `prepareBlobAssets` filters by `referencedBlobHashes`, covered by `should ignore unreferenced embedded blob assets`.
- **`--force` deletion order** — `importBlobAssets` runs *before* `replaceExistingEval`, covered by `should keep the existing eval when a --force replacement cannot store embedded media`.
- **Trace-only blob references not recorded** — `recordImportedBlobReferences` runs against every prepared asset regardless of where it's referenced, covered by `should import embedded blob assets referenced only by traces`.
- **Unnecessary blob scan on default imports** — short-circuit at `import.ts:346-352`, covered by `should skip blob reference scans for imports without embedded assets`.
- **Trace-ID parity on `--force`** — `existingEval.delete()` calls `deleteTraceRecordsForEvals` *before* `importTraces` runs, so the `traceStore.getTrace(traceId)` collision check sees a clean store and the original trace IDs are preserved. The earlier comment was based on a stale snapshot of the ordering.

---

## Verdict

**Request changes** — the P1 MIME-type injection is the only blocker. It's a real same-origin XSS in the local UI, reachable through a single user action (`promptfoo import malicious.json`). A roughly 10-line allowlist in `isImportableBlobAsset` plus a regression test resolves it; bundling the runtime-options sanitize call and the legacy-v2 `blobAssets` handling into the same follow-up commit gets this PR over the line.

Everything else — vars / durations / runtimeOptions parity, the shared `blobRefs.ts` module, the preflight-before-delete ordering, the trace round-trip with collision handling, the redaction-aware embedded-media scan, the doc updates — is well-designed, well-tested, and a clear improvement to the export/import round trip.

mldangelo:
## Re-Review at `f49224a06` (after `e90d9f389`, `b3e1844bd`, `f49224a06`)

Pulled the new commits and re-ran the verification. Both items I called out as blocker/follow-up have new fixes; one finding remains; one new low-severity hardening note. **No longer requesting changes.**

### Re-verification performed
- `git reset --hard origin/mdangelo/codex/eval-export-import-parity` → tip is `f49224a06`.
- `npm run tsc` clean.
- `npx vitest run` for the 5 PR test files + adjacent suites (`share`, `models/eval`, `models/evalResult`, `tracing`, `util/sanitizer`, `commands/export`, `commands/import`, `util/output`, `util/jsonExport`, `blobs/extractor`) — **21 files, 698 tests pass**.
- `npm run build` clean.
- Rebuilt the round-trip smoke (echo eval → JSON export → import to fresh dir → re-export) — all parity fields (`evalId`, `author`, `vars`, `runtimeOptions`, `durationMs`) still byte-identical.
- Rebuilt the adversarial import fixture with 10 cases (text/html, image/svg+xml, application/javascript, MIME with `; charset=utf-8` suffix, `TEXT/HTML` uppercase, `'  text/html  '` whitespace, `image/svg` without `+xml`, `audio/x-html` smuggle, legitimate `image/png`, legitimate `audio/mpeg`).

---

### 🔴 P1 — RESOLVED: MIME-type injection in `blobAssets` import (`e90d9f389`)

`sanitizeImportedBlobMimeType` now normalizes (trim + lowercase) and checks against a `SAFE_IMPORTED_BLOB_MIME_TYPES` allowlist plus an `audio/*` regex; everything else falls back to `application/octet-stream`. Verified by importing the 10-case fixture and inspecting both the on-disk `.meta.json` and the HTTP response from `/api/blobs/<hash>`:

```
text/html                  → application/octet-stream      ✓
image/svg+xml              → application/octet-stream      ✓
application/javascript     → application/octet-stream      ✓
image/png; charset=utf-8   → application/octet-stream      ✓  (parameter not allowed)
TEXT/HTML                  → application/octet-stream      ✓  (uppercase blocked)
'  text/html  '            → application/octet-stream      ✓  (whitespace blocked)
image/svg                  → application/octet-stream      ✓  (no +xml blocked)
image/png                  → image/png                     ✓  (legit pass-through)
audio/mpeg                 → audio/mpeg                    ✓  (legit pass-through)
```

`/api/blobs/<hash>` now serves the `text/html` and `image/svg+xml` blobs with `Content-Type: application/octet-stream` — the same-origin XSS vector is closed. New regression test `should downgrade active embedded blob MIME types during import` exercises both `text/html` and `image/svg+xml` and asserts `application/octet-stream`. ✓

### 🟡 P2 — RESOLVED: Unsanitized runtime options in exports (`b3e1844bd`, `f49224a06`)

`createOutputData` now wraps `evalRecord.runtimeOptions` in `sanitizeRuntimeOptions(...)`, and the sanitizer was lifted out of `src/models/eval.ts` into `src/util/sanitizer.ts` so backend and exporter share one implementation. Verified with the same one-shot reproduction I posted in the prior review:

```ts
const evalRecord = new Eval({}, {
  runtimeOptions: { progressCallback: () => {}, abortSignal: new AbortController().signal } as any,
});
const output = await createOutputData(evalRecord, null);
expect((output.runtimeOptions as any).progressCallback).toBeUndefined(); // ✓ passes now
expect((output.runtimeOptions as any).abortSignal).toBeUndefined();      // ✓ passes now
expect((output.runtimeOptions as any).maxConcurrency).toBe(4);           // ✓ scalars kept
```

YAML/TXT exports of in-memory `--no-write` evals will no longer fail with `unacceptable kind of an object to dump [object Function]`. New test `should sanitize runtime options before exporting` covers it. ✓

---

### 🟡 P2 — Still open: Legacy v2 imports silently drop `blobAssets`

`src/commands/import.ts:371-377`. The gate is still `importV3 && Array.isArray(evalData.blobAssets) && ...`, so a v2 export carrying embedded media imports the eval row cleanly but never restores the blobs (verified live — fresh config dir has no `blobs/` directory and zero `blob_references` rows after the import). Practical impact remains low: it only matters if someone runs `export eval --include-media` on an old v2-shape eval and then imports the file on a different machine. The P1 MIME allowlist now also protects this path defense-in-depth (an attacker can't reach the storage code from the v2 branch at all).

Suggested follow-up (not blocking): move `prepareBlobAssets` / `importBlobAssets` / `recordImportedBlobReferences` out of the `if (importV3)` block, or skip `blobAssets` emission in `createOutputData` when the eval is v2.

### 🟢 Hardening note — `audio/*` regex is broader than needed

`SAFE_IMPORTED_AUDIO_MIME_TYPE_REGEX = /^audio\/[a-z0-9_+-]+$/i` admits arbitrary subtypes like `audio/x-html`, `audio/javascript`, `audio/anything`. I verified this lets `audio/x-html` through and the blob route then returns `Content-Type: audio/x-html` with `<script>` bytes. **Not exploitable** in modern browsers — Chrome's strict MIME type checking (since Chrome 71) won't execute a `<script src>` unless the response is `text/javascript`/`application/javascript`/etc., and browsers don't sniff `audio/*` as HTML — but the bypass surface is unnecessarily large.

Two low-cost belt-and-suspenders options for a future PR:
1. Replace the audio regex with an explicit allowlist (`audio/mpeg`, `audio/mp4`, `audio/wav`, `audio/ogg`, `audio/flac`, `audio/aac`, `audio/webm`, `audio/m4a`) mirroring the image/video sets.
2. Add `res.setHeader('X-Content-Type-Options', 'nosniff')` to the `/api/blobs/:hash` response (`src/server/routes/blobs.ts`). Cheap defense against any future MIME-handling regression.

Neither is in scope for this PR.

### 🟢 Already-addressed earlier review comments (re-verified)

Confirmed obsolete against the new tip — same set as before. The `runtimeOptions` sanitization is now also covered.

---

## Verdict

**Approve, with the legacy-v2 `blobAssets` skip as a recommended follow-up.** The P1 XSS regression is closed with a well-targeted, well-tested allowlist; the runtime-options leak is closed and the sanitizer is now shared; the rest of the export/import parity work (vars, durations, traces with redaction, shared blob-hash extraction, preflight-before-delete, trace-collision remap) remains a clear and well-tested improvement. Nice fast turnaround on the security issue.

mldangelo:
## Final Consolidated Review @ `f49224a06`

Consolidating both review passes into a single comprehensive finding list. All claims independently verified end-to-end against the current branch tip (`git reset --hard origin/mdangelo/codex/eval-export-import-parity`).

### Verification performed
- Read every changed file plus relevant unchanged code: `Eval.create`/`Eval.delete`, `getBlobByHash`, `src/server/routes/blobs.ts`, `genaiTracer.PromptfooAttributes`, `sanitizeRuntimeOptions`, `inlineBlobsForShare`, `extractor`.
- `npm run tsc` — clean.
- `npx vitest run` on 21 files (5 PR files + 16 adjacent suites) — **698 tests pass**.
- `npm run build` — clean.
- Live round-trip smoke in isolated `PROMPTFOO_CONFIG_DIR`: baseline eval → JSON+YAML export → import to fresh dir → re-export. Byte-identical for `evalId`, `metadata.author`, `metadata.evaluationCreatedAt`, `vars`, `runtimeOptions`, result counts, `durationMs`, `evaluationDurationMs`.
- Collision flows: bare collision fails correctly; `--force` replaces (and existing eval is preserved when preflight fails — confirmed with the corrupt-blob and `storeBlob` rejection tests); `--new-id` mints a new ID.
- Trace round-trip: `traceId` preserved on first import, remapped on `--new-id` reimport, remapped on cross-eval collision.
- JSONL export still works after the `createOutputData` refactor.
- Adversarial fixtures: corrupt-hash, oversized base64, unsupported-results shape, orphan-blob, prototype-pollution — all correctly rejected or filtered.
- **MIME injection / XSS reproduction**: imported a 10-case fixture (text/html, image/svg+xml, application/javascript, MIME with `; charset=utf-8` suffix, `TEXT/HTML` uppercase, `'  text/html  '` whitespace, `image/svg` no `+xml`, `audio/x-html` smuggle, legit `image/png`, legit `audio/mpeg`) on the new HEAD and inspected both the on-disk `.meta.json` files and the actual `/api/blobs/<hash>` HTTP responses.

---

## Findings (full list, severity-ordered)

### 🔴 P1 — RESOLVED: Stored same-origin XSS via `blobAssets` MIME-type injection
**Original status:** First review found `isImportableBlobAsset` accepted *any* non-empty `mimeType` string. `prepareBlobAssets` → `importBlobAssets` then called `storeBlob(data, asset.mimeType)`, and `recordImportedBlobReferences` wrote a `blob_references` row so the blob route served it. `SAFE_MIME_TYPE_REGEX = /^[a-z]+\/[a-z0-9_+-]+$/i` in `src/server/routes/blobs.ts` admitted `text/html`, `image/svg+xml`, `application/xhtml+xml`, etc. Reproduced live: `curl /api/blobs/<hash>` returned `<script>alert(document.domain)</script>` with `Content-Type: text/html` and `Cache-Control: immutable`, exploitable via a single `promptfoo import malicious.json`.

**Fix (`e90d9f389`):** New `sanitizeImportedBlobMimeType` normalizes (trim + lowercase), checks an explicit `SAFE_IMPORTED_BLOB_MIME_TYPES` allowlist (`image/avif|gif|jpeg|png|webp`, `video/mp4|ogg|webm`) plus an `audio/*` regex; anything else falls back to `application/octet-stream`. Re-tested live against the new HEAD with the 10-case fixture:

```
text/html                  → application/octet-stream      ✓
image/svg+xml              → application/octet-stream      ✓
application/javascript     → application/octet-stream      ✓
image/png; charset=utf-8   → application/octet-stream      ✓  (parameter rejected)
TEXT/HTML                  → application/octet-stream      ✓  (uppercase blocked)
'  text/html  '            → application/octet-stream      ✓  (whitespace stripped, still blocked)
image/svg                  → application/octet-stream      ✓  (no +xml blocked)
image/png                  → image/png                     ✓  (legit pass-through)
audio/mpeg                 → audio/mpeg                    ✓  (legit pass-through)
```

`/api/blobs/<hash>` now serves the formerly-malicious blobs with `Content-Type: application/octet-stream`. New regression test `should downgrade active embedded blob MIME types during import` exercises both `text/html` and `image/svg+xml` and asserts `application/octet-stream`.

### 🟡 P2 — RESOLVED: `createOutputData` exported unsanitized `runtimeOptions`
**Original status:** `src/util/output.ts:200` did `...(evalRecord.runtimeOptions && { runtimeOptions: evalRecord.runtimeOptions })` without sanitizing. `Eval` constructor at line 621 stored `opts?.runtimeOptions` raw, so any in-memory eval (`--no-write` CLI path, library callers) could carry `progressCallback`, `abortSignal`, etc. JSON would silently drop functions and leave `abortSignal: {}` artifacts; YAML/TXT would throw `unacceptable kind of an object to dump [object Function]` and fail the export.

**Fix (`b3e1844bd` + `f49224a06`):** `createOutputData` now wraps `evalRecord.runtimeOptions` in `sanitizeRuntimeOptions(...)`, and the sanitizer was lifted out of `src/models/eval.ts` into `src/util/sanitizer.ts` so backend persistence and the exporter share one implementation. Verified with a one-shot reproduction on the new HEAD: function and AbortSignal stripped, scalar fields preserved. New test `should sanitize runtime options before exporting` covers it.

### 🟡 P2 — Still open: Legacy v2 imports silently drop embedded `blobAssets`
**Files:** `src/commands/import.ts:371-377`.

`createOutputData` emits `blobAssets` for any export, v2 or v3 (the scan walks the summary regardless of version). On import the gate `importV3 && Array.isArray(evalData.blobAssets) && evalData.blobAssets.length > 0` means v2 imports always skip blob ingestion. Verified live on the new HEAD: legacy v2 export with `blobAssets[]` imports cleanly, but `~/.promptfoo/blobs/` is empty and no `blob_references` row exists — the blob URIs in the v2 results 403 from `/api/blobs/<hash>`.

Practical impact is low: it only matters if someone runs `export eval --include-media` on an old v2-shape eval and then imports the file on a different machine. The P1 MIME allowlist now also protects this path defense-in-depth (an attacker can't reach the storage code from the v2 branch at all). Recommended follow-up, not blocking: either move `prepareBlobAssets` / `importBlobAssets` / `recordImportedBlobReferences` out of the `if (importV3)` block so v2 imports also restore embedded media, or suppress `blobAssets` emission in `createOutputData` when the eval is v2.

### 🟢 Hardening — `audio/*` MIME regex is broader than necessary
**File:** `src/commands/import.ts:114`.

`SAFE_IMPORTED_AUDIO_MIME_TYPE_REGEX = /^audio\/[a-z0-9_+-]+$/i` admits arbitrary subtypes like `audio/x-html`, `audio/javascript`, `audio/anything`. Verified live: `audio/x-html` passes through, blob route returns `Content-Type: audio/x-html` with `<script>` bytes. **Not exploitable** in modern browsers — Chrome's strict MIME type checking (since Chrome 71) won't execute `<script src>` unless the response is `text/javascript`/`application/javascript`/etc., and browsers don't sniff `audio/*` as HTML — but the bypass surface is unnecessarily large.

Two low-cost belt-and-suspenders options for a future PR:
1. Replace the audio regex with an explicit allowlist (`audio/mpeg`, `audio/mp4`, `audio/wav`, `audio/ogg`, `audio/flac`, `audio/aac`, `audio/webm`, `audio/m4a`) mirroring the image/video sets.
2. Add `res.setHeader('X-Content-Type-Options', 'nosniff')` to the `/api/blobs/:hash` response in `src/server/routes/blobs.ts`. Cheap defense against any future MIME-handling regression.

Neither is in scope for this PR.

### 🟢 Minor — `recordImportedBlobReferences` ordering
**File:** `src/commands/import.ts:384-398`.

If `EvalResult.createManyFromEvaluateResult` or `importTraces` throws between `importBlobAssets` and `recordImportedBlobReferences`, the prepared blob bytes are already stored but no `blob_references` row exists — dangling blobs. Not a regression and self-healing on the next eval that references the same hash. Lowest priority — wouldn't fix it standalone.

---

## Earlier review comments verified obsolete against current tip

To save reviewers cycle time, the following Codex/Copilot findings from earlier commits are no longer applicable. I re-verified each against `f49224a06`:

- **Trace span body stripping for `STRIP_PROMPT_TEXT`/`STRIP_RESPONSE_OUTPUT`** — handled in `projectTracesForOutput` (`output.ts:104-124`), covered by `omits stripped prompt and response bodies from trace span attributes`.
- **Trace metadata leak when only `vars` is present** — handled via the `else if` branch with `traceWithoutMetadata` (`output.ts:89-95`), covered by `omits trace metadata when stripped vars are the only metadata`.
- **`STRIP_RESPONSE_OUTPUT` + `--include-media` embedding response bytes** — handled by `resultsForMediaExportScan` (`output.ts:129-153`), covered by `should not embed response blob bytes when response output stripping is enabled`.
- **Orphan blobs in imports** — `prepareBlobAssets` filters by `referencedBlobHashes`, covered by `should ignore unreferenced embedded blob assets`.
- **`--force` deletion order** — `importBlobAssets` runs *before* `replaceExistingEval`, covered by `should keep the existing eval when a --force replacement cannot store embedded media`.
- **Trace-only blob references not recorded** — `recordImportedBlobReferences` runs against every prepared asset regardless of where it's referenced, covered by `should import embedded blob assets referenced only by traces`.
- **Unnecessary blob scan on default imports** — short-circuit at `import.ts:371-377`, covered by `should skip blob reference scans for imports without embedded assets`.
- **Trace-ID parity on `--force`** — `existingEval.delete()` calls `deleteTraceRecordsForEvals` *before* `importTraces` runs, so `traceStore.getTrace(traceId)` sees a clean store and original trace IDs are preserved on `--force`. The earlier comment was based on a stale snapshot of the ordering.

---

## Verdict

**Approve.** The P1 same-origin XSS regression is closed with a well-targeted allowlist and a regression test that covers both `text/html` and `image/svg+xml`. The P2 runtime-options leak is closed and the sanitizer is now properly shared between persistence and export. The remaining legacy-v2 `blobAssets` skip is a low-impact follow-up; the audio-regex breadth and missing `nosniff` header are nice-to-have hardening for a separate PR.

The rest of the export/import parity work — vars / durations / runtimeOptions parity, shared `blobRefs.ts` module, preflight-before-delete ordering for `--force`, trace round-trip with collision remap, redaction-aware embedded-media scan, doc updates — is well-designed, well-tested (520+ tests in the touched suites all green), and a clear improvement to the round trip. Nice fast turnaround on the security finding.

ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/3b380127-6a2a-4bda-8a85-83828f21c18a?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on 28891cfdd903db18d62cb4eb59944bc44308dbaf. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 59406e470c6b9183cafc90ffe90431ef1447680e | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/fbd50f40-6729-43c7-bbdd-83febcfae593?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 21, 2026 1:49AM |
| 28891cfdd903db18d62cb4eb59944bc44308dbaf | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/3b380127-6a2a-4bda-8a85-83828f21c18a?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 21, 2026 2:59AM |

</p>
</details>

PR #9313, opened 2026-05-20
fix(server): cap eval table page size

### LINKED ISSUE #9312: Cap eval table page size for non-export requests
## Summary

`GET /api/eval/:id/table` accepts any positive integer for `limit` on normal table requests. The route already intentionally treats export requests differently by using `Number.MAX_SAFE_INTEGER` when `format` is set, but non-export page requests currently do not have an upper bound.

## Why this matters

A very large `limit` can make a normal table request ask the server to load and serialize a large amount of eval result data in one response. Other list-style APIs in the server already bound their page sizes, so the eval table endpoint should follow the same resource-control pattern.

## Proposed change

- Add a maximum `limit` for non-export `GET /api/eval/:id/table` requests.
- Keep `format=csv` / `format=json` exports unchanged so full-table exports still work as intended.
- Return validation errors for excessive non-export limits, consistent with nearby API schema validation.

A cap of `500` or `1000` would likely cover normal UI usage while preventing accidental oversized page requests.

## Validation

Suggested tests:

- schema test rejects `EvalSchemas.Table.Query.safeParse({ limit: '<cap + 1>' })`
- route validation test returns `400` for excessive non-export `limit`
- export route tests continue to use `Number.MAX_SAFE_INTEGER` when `format` is present

fallintoplace:
## Summary
- cap non-export `GET /api/eval/:id/table` `limit` at `EVAL_TABLE_MAX_PAGE_SIZE` (1000) — matches the largest option the web UI's "Results per page" selector exposes
- keep `format=csv` / `format=json` exports uncapped (route still uses `Number.MAX_SAFE_INTEGER`)
- derive the UI page-size selector from the same shared constant so the dropdown can never offer a value the server would reject
- add schema, route validation, and UI regression coverage for the cap

Fixes #9312

## Why 1000
The eval table UI already lets users with large evals pick 1000 rows per page. Capping below that would silently 400 the request and leave the table blank (the store swallows non-OK responses), so the cap is set to the existing UI ceiling rather than below it.

## Validation
- `npx vitest run test/types/apiSchemas.redteam.test.ts test/server/routes/eval.validation.test.ts test/server/routes/eval.export.test.ts`
- `npm run test:app -- src/pages/eval/components/ResultsTable.test.tsx --run`
- `npm run l && npm run f && npm run tsc`
- Live: `GET /api/eval/<id>/table?limit=1000` → 200, `?limit=1001` → 400 with `"limit must be less than or equal to 1000"`, `?format=csv&limit=99999` → 200

### DIFF
--- a/src/app/src/pages/eval/components/ResultsTable.test.tsx
+++ b/src/app/src/pages/eval/components/ResultsTable.test.tsx
@@ -3,6 +3,7 @@ import { act } from 'react';
 import { restoreTestTimers, type TestTimers, useTestTimers } from '@app/tests/timers';
 import { renderWithProviders } from '@app/utils/testutils';
 import { FILE_METADATA_KEY } from '@promptfoo/providers/constants';
+import { EVAL_TABLE_MAX_PAGE_SIZE } from '@promptfoo/types/api/eval';
 import { screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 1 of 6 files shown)


mldangelo:
Thank you!

fallintoplace:
Great work @mldangelo

PR #9322, opened 2026-05-21
feat(eval): support runtime tags

### LINKED ISSUE #9268: feat(eval): support --tag key=value CLI flag to set tags at runtime
## Problem

Tags (`tags: { key: value }`) are a first-class concept in promptfoo configs and are persisted to the DB, but they can only be set **statically** in `promptfooconfig.yaml`. There is no way to pass a tag value at runtime via CLI.

This matters when the provider and prompt are constant across runs but something external changes — e.g. a claude plugin version, a git SHA, a CI run ID — and you want to distinguish or filter eval results by that dimension later.

## Proposed solution

Add a `--tag key=value` flag to `promptfoo eval` (repeatable for multiple tags):

```bash
promptfoo eval -c promptfooconfig.yaml --tag skill-version=1.2.3 --tag run-id=$CI_RUN_ID

CLI tags should merge with (and override) any tags defined in the config file.

Current workaround

Environment variable substitution via nunjucks in the config:

# promptfooconfig.yaml
tags:
  skill-version: "{{ env.SKILL_VERSION }}"

SKILL_VERSION=1.2.3 promptfoo eval -c promptfooconfig.yaml

This works but is indirect — requires modifying the config file and using env vars instead of a first-class CLI flag.

Implementation notes

- CommandLineOptionsSchema in src/types/index.ts would need a tags field added
- The eval command in src/commands/eval.ts would need a .option('--tag <key=value>', ..., collect) added
- Merged tags should flow into UnifiedConfig.tags before the eval is written to DB
- The existing DB layer (src/util/database.ts) already handles tag persistence — no changes needed there

puneetdixit200:
Fixes #9268

This adds a repeatable `--tag key=value` option to `promptfoo eval` so callers can set run-specific tags without editing `promptfooconfig.yaml` or relying on environment interpolation.

CLI-provided tags are merged over tags from the resolved config before the eval record is created, so the existing tag persistence path continues to handle storage. The command-line options schema and generated JSON schema now include `tags` as well.

Testing:
- `npx @biomejs/biome ci src/commands/eval.ts src/types/index.ts test/commands/eval.test.ts site/static/config-schema.json`
- `npm run tsc`
- `npx vitest run test/commands/eval.test.ts test/config-schema.test.ts test/types/index.test.ts`
- `npm run local -- eval -c test/smoke/fixtures/configs/basic.yaml --no-write --no-table --tag env=ci --tag run=smoke`


mldangelo (approved):
Thanks @puneetdixit200, and welcome — nice first contribution!

This cleanly closes #9268. A couple of things I liked: extracting the shared `collectKeyValueOption` helper for `--var` and `--tag` (and using `indexOf`/`slice` so values containing `=` survive), and picking up the docs + the `commandLineOptions.tags` precedence test after the first review round without being asked twice.

I pushed one commit to your branch before merging: a guard so `--tag` errors out when combined with `--resume`/`--retry-errors` instead of being silently dropped — resumed/retried runs reuse the original eval's tags, so a `--tag` there would have quietly done nothing, and anyone filtering eval history by that tag later would be puzzled. It mirrors the existing conflict checks in the same function, with tests.

Merging now. Thanks again — looking forward to more!

PR #9345, opened 2026-05-21
fix(db): parameterize JSON path filters

mldangelo-oai:
## Summary
- bind SQLite JSON paths through Drizzle instead of raw SQL and remove the production `sql.raw()` escape hatch
- remove redundant quote escaping from optimized result search parameters
- add attack-shaped metadata key regression coverage, a production `sql.raw()` source guard, and updated database security guidance

## Test plan
- `npx vitest run test/models/eval.test.ts test/models/evalHighlights.test.ts test/database/sqlSafety.test.ts --sequence.shuffle=false`
- `npm run f`
- `npm run l`
- `git diff --check`

## Notes
- `npm run tsc` was attempted in the isolated worktree with the linked local dependency tree and failed in existing `src/__mocks__/database.ts` TS2883 inferred-type errors referencing Vitest `Procedure`; no changed file appeared in that error list.


ianw-oai (approved):

.github/workflows/promptfoo-code-scan.yml
```diff
@@ -32,7 +32,8 @@ jobs:
       - name: Use Node
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6
         with:
-          node-version-file: '.nvmrc'
+          # The standalone scan's published CLI install times out on Node 24.16.0.
+          node-version: '24.15.0'
```
mldangelo-oai: revert this

Issue #9348, opened 2026-05-21
Invalidate eval result count caches after result writes


fallintoplace:
## Summary

Eval result count helpers cache counts for 5 minutes, but result write paths do not invalidate those caches after inserting rows.

## Why this matters

`getCachedResultsCount()` and `getTotalResultRowCount()` cache distinct and total result counts per eval. If a count is cached before new rows are written, later calls can return stale counts until the TTL expires.

Example sequence:

1. Create or open an eval with no results.
2. Call `getResultsCount()` or `getTotalResultRowCount()` and cache `0`.
3. Insert one or more `eval_results` rows.
4. Call the count helper again before TTL expiry.

The stored rows are correct, but count consumers can still see stale data.

## Proposed change

Clear the count cache for the eval after persisted result inserts:

- `EvalResult.createFromEvaluateResult()` after a persisted insert
- `EvalResult.createManyFromEvaluateResult()` once after the insert transaction

## Validation

Add regression coverage that caches a zero count, inserts results, and verifies distinct and total counts refresh without manually calling `clearCountCache()`.

PR #9349, opened 2026-05-21
fix(cache): prevent stale result-count refill

fallintoplace:
## Summary

- Prevent count queries that were already in flight from repopulating the distinct-result or total-row cache after scoped or global invalidation.
- Track only active queries for each evaluation, preserve cacheability for unrelated evaluations, and remove query state on both success and failure.
- Add database-backed regressions across initial result import, reload plus `setResults()`, `EvalResult.save()`, failed-save recovery, concurrent saves, and every evaluation deletion path.
- Add deterministic race coverage that uses the real centralized invalidator and cache implementation while controlling only database-query completion timing.
- Merge current `main` additively, removing the unrelated Codex skill-metadata formatting drift from the effective diff.

## Why

The dispatched head still had a concurrency defect after the earlier centralized invalidation work: a count query could start before a result mutation, finish after invalidation, and unconditionally refill the five-minute cache with its obsolete snapshot. A later read then returned the stale cached value.

The regression was reproduced for both count helpers on the dispatched head. The fix marks every active query affected by an invalidation and suppresses only its cache write; the already-running caller keeps normal snapshot semantics, while subsequent callers re-query.

The lifecycle tests prime both real count caches, execute real model/database mutations, and assert observable counts. They do not mock invalidation calls. The adversarial race tests mock only asynchronous query completion so the ordering is deterministic.

Relates to #9348.

## Verification

- Dispatched-head race reproduction: both distinct and total count cases returned stale `0` on the second read instead of fresh `1`.
- Central-invalidation mutation check: replacing the central count-cache clears with no-ops caused seven database-backed lifecycle regressions to fail.
- Focused final suite: 26 tests passed; repeated under seeds `7`, `42`, `9349`, `12345`, and `65535`.
- Model/database suite: 15 files, 363 tests passed.
- `tsc --noEmit` passed on Node 24.18.0; focused compatibility also passed on Node 22.22.2.
- Real CLI workflow: `promptfoo eval --no-cache` passed 1/1 and the exported JSON parsed with one successful row, score `1`, no failures or errors, and output `Hello World`.
- Pre-commit Biome checks passed for all three effective diff files.


mldangelo:
Thanks @fallintoplace — really excellent piece of work. A separate fix by @zanni098 (#9421) for the same #9348 bug landed first, but it only covered the two `EvalResult.createFromEvaluateResult` / `createManyFromEvaluateResult` insert paths. **Your PR is materially broader** — it also catches `EvalResult.save()`, the bulk-insert paths in `Eval.create.opts.results` and `Eval.addResults`, every delete path (`Eval.delete`, `deleteEval`, `deleteEvals`, `deleteAllEvals`, `Eval.copy`, `retry.deleteErrorResults`), and adds 6 new regression tests that the prior PR didn't cover.

I rebased your branch on current `main` and resolved the small overlap in `src/models/evalResult.ts`:
- Removed the duplicate `clearCountCache` import that came in from #9421
- Kept your slightly-better `if (returnResults.length > 0)` guard around the cache clear

Everything else in your PR is preserved verbatim. 19/19 tests in `test/models/evalPerformance.test.ts` pass locally including all 6 of your new delete/copy regression tests. Waiting on CI before merging.

mldangelo-oai:
Final review — **FIXED_AND_READY**

**Final head:** `b372b99f62250b41ba5a1f7dcc05fc3078f773ce`

**Findings and fixes**

- Reproduced the remaining production race on the dispatched head for both count helpers: an obsolete in-flight query could refill the five-minute cache after centralized invalidation, so the next read returned stale `0` instead of fresh `1`.
- Added scoped active-query tracking so invalidation suppresses stale cache writes for every affected query, including simultaneous distinct/total reads and global clears. Unrelated evaluations remain cacheable, and success/error paths clean up tracking in `finally`.
- Kept the lifecycle coverage black-box and database-backed across initial create/import, reload plus `setResults()`, direct save, forced save failure and retry, concurrent saves, model/utility/multi/global deletion, and distinct-versus-total semantics. The deterministic race tests use the real centralized invalidator and cache maps; only query completion timing is controlled.
- Merged current `main` additively and removed the unrelated Codex skill-metadata formatting drift from the effective diff. No history was rewritten.

**QA**

- Dispatched-head race proof failed both distinct and total cases as expected; removing central count-cache clears caused seven lifecycle regressions to fail.
- Focused final suite: 26/26, also green under seeds `7`, `42`, `9349`, `12345`, and `65535`.
- Model/database suite: 15 files, 363 tests.
- `tsc --noEmit` passed on Node 24.18.0; focused compatibility passed on Node 22.22.2.
- Real `promptfoo eval --no-cache` workflow passed 1/1; exported JSON parsed with one successful row, score `1`, no failures/errors, and `Hello World` output.
- [Exact-head CI](https://github.com/promptfoo/promptfoo/actions/runs/28143316734) completed successfully across the full Node/OS matrix; expected Share/Staging jobs were skipped. Codecov reports 100% coverage of modified coverable lines.

**Remaining gates**

- Review threads: **0 total / 0 unresolved**. Submitted reviews: **0**.
- The PR is open, non-draft, current with `main`, and GitHub reports it mergeable. One required code-owner approval remains (`REVIEW_REQUIRED`); I did not approve, merge, or bypass it.
- The active ruleset still names a legacy `CodeQL` context that emitted no check on this head. `gh pr checks --required` is green and the repository's `Code Scan Action` passed, but maintainers may need to reconcile that stale context if branch protection still surfaces it after approval.

**Disposition:** `FIXED_AND_READY`

PR #9352, opened 2026-05-21
fix(webui): avoid NaN in named metric charts

### LINKED ISSUE #9350: Avoid NaN values in named metric charts
## Summary
Named metric chart data normalizes each prompt value by the maximum value for that metric. When every prompt has `0` for a named metric, the maximum is also `0`, so the chart receives `NaN` from `0 / 0`.

## Proposed fix
Return `0` when the maximum named metric value is not positive, so all-zero metrics render as zero-height bars instead of non-finite chart data.

## Validation
Add focused coverage that renders named metric charts with all-zero named scores and asserts the generated chart dataset contains finite values.

fallintoplace:
## Summary
- guard named metric normalization when every prompt has a zero value
- render all-zero named metrics as zero-height bars instead of non-finite chart data
- add regression coverage for all-zero named metric chart datasets

Fixes #9350

## Validation
- `ASDF_NODEJS_VERSION=24.15.0 asdf exec npm --prefix src/app run test -- src/pages/eval/components/ResultsCharts.test.tsx --run`
- `ASDF_NODEJS_VERSION=24.15.0 asdf exec npm --prefix src/app run tsc`
- `ASDF_NODEJS_VERSION=24.15.0 asdf exec npm run l`
- `ASDF_NODEJS_VERSION=24.15.0 asdf exec npm run tsc`
- `git diff --check`

### DIFF
--- a/src/app/src/pages/eval/components/ResultsCharts.test.tsx
+++ b/src/app/src/pages/eval/components/ResultsCharts.test.tsx
@@ -636,6 +636,69 @@ describe('ResultsCharts', () => {
       );
       expect(values).toEqual([0, 0, 0, 0]);
     });
+
+    it('should normalize each named metric independently when one metric is all zero', () => {
+      // `accuracy` is all zero (exercises the zero-max guard), while `precision` and
+      // `coverage` have different maxima. Per-key normalization divides each metric by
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thank you!

PR #9373, opened 2026-05-22
fix(eval): align optimizer evaluation semantics

mldangelo:
## Summary

The optimizer calls the low-level evaluator several times, so runtime normalization that normally happens in the `eval` command was not consistently applied. This change aligns the optimizer with the current evaluation contract while preserving optimizer-specific isolation.

- Apply configured `commandLineOptions` ahead of `evaluateOptions` for `delay`, `maxConcurrency`, `repeat`, and `filterRange`, including an explicit zero-delay override. Apply config-level range and sampling filters together before validation splitting.
- Force concurrency to 1 for a positive delay and normalize zero, negative, and `NaN` delays to no delay.
- Scope the effective concurrency through `cliState` so Python worker pools use the same concurrency as the optimizer.
- Derive internal `isRedteam` telemetry from the runtime test suite. The suite remains authoritative for row-level redteam behavior: absent or `null` is non-redteam, while an empty or populated redteam block is redteam. Normal CLI redteam scoring already followed the resolved suite; this does not claim otherwise.
- Apply an explicit test range once before the search/validation split, reject empty selections, and leave default-only/scenario filtering to the evaluator.
- Reuse the established test-filter utilities and record the intentional command/optimizer dependency edges in the architecture ratchet.
- Propagate cancellation through target evaluations and candidate generation, preserve legacy config-carried abort signals, reject partial results after cancellation, and preserve the best completed result when a later non-cancellation generation round converges or fails.
- Keep eval-level prompt suggestion flags disabled inside optimizer evals because the optimizer owns candidate generation itself.
- Keep internal eval output/progress disabled and avoid persisting intermediate eval records.

## Compatibility and risk

- No config schema, result schema, or persisted-output format changes.
- Positive delays now intentionally serialize optimizer target calls; nonpositive delays preserve configured concurrency.
- Config-level range and sample filters now affect the optimizer in the same order as normal eval; eval-level prompt suggestion flags are intentionally ignored internally.
- Runtime-suite redteam state wins over inconsistent config-only values, matching `createRunEvalOptions` row semantics.
- Cancellation is cooperative: providers that honor `AbortSignal` can stop their request; providers that ignore it finish the in-flight call, after which the optimizer rejects without starting another phase.
- Existing optimizer limitations are unchanged: `maxEvalTimeMs` is per internal eval rather than a workflow-wide budget, and low-level cache scoping is outside this PR.

## Test plan

- [x] `npx vitest run test/commands/optimize.test.ts test/optimizer/promptOptimizer.test.ts --sequence.shuffle=false` — 63 passed
- [x] `npx vitest run test/commands/optimize.test.ts test/optimizer/promptOptimizer.test.ts test/providers/pythonCompletion.cliState.test.ts test/util/eval/filterTests.test.ts --sequence.shuffle=false` — 120 passed
- [x] `npm test` — 784 files passed, 1 skipped; 20,515 tests passed, 10 skipped
- [x] `npm run tsc` — passed
- [x] `npm run l` — passed
- [x] `npm run architecture:check` — passed
- [x] `npm run build` — passed, including the web app and postbuild
- [x] Deterministic local API workflow — three optimization rounds; positive delay used concurrency 1 with a 36 ms minimum start gap; cancellation returned `Operation cancelled` with zero suggestion calls after abort; default-only and explicit empty ranges were rejected correctly
- [x] Deterministic real CLI workflow — configured `delay: 30` produced 14 serialized calls and improved the baseline; explicit command-level `delay: 0` overrode `evaluateOptions.delay: 100` and remained concurrent; an empty-output resolved redteam suite completed without a false `No output` error
- [x] Deterministic real CLI filter/suggestion/scenario workflows — `filterRange: 0:4` plus seeded sample 2 produced 14 target calls (two selected tests across three rounds); enabling eval prompt-suggestion flags still produced the expected seven optimizer target calls without nested prompt generation; scenario `filterRange: 1:2` selected one expanded row and produced seven target calls


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/b08b8c1d-a409-4ca0-b09c-6db134212c01?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on 287802d073de8062f73d5467ffdf859460774a00. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| f6fc7de1186bd90666a5314121ce86e8a44d94ec | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/90290238-1bff-4f67-9b34-6b491c9b093a?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 22, 2026 5:35AM |
| 287802d073de8062f73d5467ffdf859460774a00 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/b08b8c1d-a409-4ca0-b09c-6db134212c01?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 25, 2026 5:53AM |

</p>
</details>

src/optimizer/promptOptimizer.ts
```diff
@@ -601,14 +601,36 @@ function countConfiguredOptimizationTests(testSuite: TestSuite): number {
...
+    ? { ...testSuite, redteam: config.redteam }
+    : testSuite;
+}
+
+function createEvaluationOptions(
+  config: Partial<UnifiedConfig>,
+  testSuite: TestSuite,
+): InternalEvaluateOptions {
+  const options: InternalEvaluateOptions = {
     ...config.evaluateOptions,
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `73ce5a29b`. Explicit tests now apply `evaluateOptions.filterRange` once before a validation partition is formed, and the consumed range is omitted from each internal `evaluate` call so absolute indices cannot be reapplied inside smaller subsets. I also extracted that preparation into a small helper to keep the optimizer loop below the complexity threshold and added a regression asserting the `1:5` subsets plus cleared runtime option. Verified with `npx vitest run test/optimizer/promptOptimizer.test.ts` (21 tests), `npm run l`, `npm run build`, and an isolated deterministic optimize run using `filterRange: '3:5'` with `validationSplit: 0.5`, which evaluated both search and validation baselines successfully at score 1.

mldangelo-oai:
Final ownership review at exact head `0767cf19a6fa5b57bd21354494f3378be5ad7ac0`.

- Fixed optimizer parity for runtime-suite redteam state, command/config delay and concurrency precedence, range + seeded sampling (including default/scenario cases), Python worker concurrency, cancellation/partial-result handling, and optimizer-owned prompt suggestions.
- QA: focused 120/120; full suite 20,515 passed / 10 skipped; TypeScript, lint, architecture, and build passed. Deterministic local CLI/API runs verified serialized positive delay, explicit zero-delay concurrency, cancellation with no later suggestion call, redteam empty-output behavior, two-row range+sample selection (14 calls), disabled nested suggestions (7 calls), and one expanded scenario range row (7 calls).
- Exact-head CI [28121411532](https://github.com/promptfoo/promptfoo/actions/runs/28121411532) and both CodeQL runs are fully green. The full static roster, fresh second review, and final-delta review found no remaining diff-local issue; all 8 review threads are resolved.
- Residual risk: cancellation remains cooperative for providers that ignore `AbortSignal`; `maxEvalTimeMs` remains per internal eval, and the pre-existing low-level cache-scoping limitation remains outside this PR.

The branch is current with `main`, mergeable, and blocked only on required human approval. No merge or approval action was taken.

@codex review

src/util/config/load.ts
```diff
@@ -106,6 +112,195 @@ function deriveScenarioSampleSeed(seed: number, scenarioIndex: number): number {
...
+    ...(test.assert
+      ? {
+          assert: test.assert.map((assertion) =>
+            'provider' in assertion
+              ? { ...assertion, provider: serializeProviderInstance(assertion.provider) }
+              : { ...assertion },
+          ),
+        }
+      : {}),
+    provider: originalProvider ?? serializeProviderInstance(test.provider),
```
jameshiester-oai, line 178: [P1] Do not persist the raw external provider configuration. When a filtered run uses `--tests <external-file>`, `originalProvider` is the file's raw `ProviderOptions`; choosing it here bypasses `serializeProviderInstance`, so values such as API keys or Authorization headers are copied into `config.tests`, persisted to SQLite, and included by sharing (which only redacts Azure SAS values). An isolated probe with a synthetic marker confirmed it was present in `evals.config`. Preserve the external file reference or pass this value through a secret-safe, replayable serializer.

src/util/config/load.ts
```diff
@@ -106,6 +112,195 @@ function deriveScenarioSampleSeed(seed: number, scenarioIndex: number): number {
...
+            context: 'provider config',
+            maxDepth: Number.POSITIVE_INFINITY,
+          }),
+        }
+      : {}),
+  } as T;
+}
+
+function detachTestForPersistence(test: TestCase): TestCase {
+  const originalProvider = (test as TestCaseWithOriginalProvider)[originalTestProviderReference];
```
jameshiester-oai, line 156: [P2] Retain the source base path for external test providers. This captures only a raw relative reference such as `python:../providers/custom.py`; after the selected test is persisted inline, replay resolves it from the config/CWD rather than the external test file's directory. The exact-head probe resolved the same row to `/private/tmp/pr9373-provider-replay/config/providers/custom.py` initially and `/Users/jameshiester/.codex/worktrees/63ab/providers/custom.py` on replay. Persist a location-stable descriptor (or the original test-file reference plus its base path) so resume cannot load the wrong provider.

src/node/doEval.ts
```diff
@@ -621,23 +636,30 @@ export async function doEval(
         testSuite.tests = defaultMetadata ? [{ metadata: defaultMetadata }] : [{}];
```
jameshiester-oai, line 636: [P2] Persist a replay-valid implicit default row. When the default-config/library path has no metadata and a filter keeps the implicit test (for example `filterRange: '0:1'`), this creates `{}`; `snapshotTestSelection()` then stores `config.tests = [{}]` and marks the selection applied. The first eval runs, but replay feeds that explicit empty row to `readTest()`, which rejects it. I reproduced a 1/1 initial pass followed by the empty-test error. Use a valid sentinel such as `{ metadata: {} }`, or preserve the absence of explicit tests alongside the selection state.

src/util/eval/filterTestsUtil.ts
```diff
@@ -245,6 +262,7 @@ export async function filterTestsByResults(
     );
   }
 
-  // Deduplicate and return combined tests
-  return deduplicateTestCases([...matchedTests, ...extractedTests]);
+  // Each configured test is visited at most once above, and extracted tests are deduplicated
+  // while they are collected. Preserve configured rows that intentionally share vars.
+  return [...matchedTests, ...extractedTests];
```
jameshiester-oai, line 267: [P2] Match prior results one-to-one before preserving configured multiplicity. Two configured rows can share vars/provider/conversation while differing in assertions or description; `resultIsForTestCase` then lets one failed result match both. Returning every matched row here causes `--filter-failing-only` to rerun a passing sibling. The exact-head probe had one failed and one passing row and selected both; the base's final dedupe selected only the first failed row in this ordering. Consume matching results by stable test identity/index (or compare the stored `testCase`) before retaining duplicate configured rows.

jameshiester-oai (commented):
## Codex review

Reviewed base `4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6` through full head `09936f9104b5337d373d58fd24aebbaa7b5f6d77`. This was a full review of the once-restarted current head; there was no prior authenticated completion marker for an incremental scope.

**Verdict:** I found one P1 credential-exposure issue and three P2 replay/filter correctness defects. Current GitHub CI is fully green (including Linux Node 20/22/24/26, Windows shards, macOS, build, smoke/integration, site/web, CodeQL, and Codecov); the PR is mergeable but behind its base.

### Verified findings

- **P1 — filtered external test providers can copy credentials into persisted/shareable config.** The snapshot prefers the raw `originalProvider`, bypassing the sanitizer used for instantiated providers. With `--tests <external-file>` and an active filter, a synthetic `apiKey` marker was present in `evals.config`; the share path only applies Azure SAS redaction. See inline comment on `src/util/config/load.ts`.
- **P2 — relative providers from external test files lose their source base path on replay.** A `python:../providers/custom.py` test resolved to `/private/tmp/pr9373-provider-replay/config/providers/custom.py` initially and `/Users/jameshiester/.codex/worktrees/63ab/providers/custom.py` after the snapshotted config was replayed. See inline comment on `src/util/config/load.ts`.
- **P2 — a kept implicit default row can be persisted as an invalid explicit `{}` test.** A default-config/library path with no metadata ran successfully, stored `tests: [{}]`, then failed config replay because `readTest()` rejects an empty explicit row. See inline comment on `src/node/doEval.ts`.
- **P2 — prior-result filters overselect same-vars sibling rows.** With one failed and one passing row sharing vars/provider/conversation but differing in description/assertions, `--filter-failing-only` returned both after final deduplication was removed. See inline comment on `src/util/eval/filterTestsUtil.ts`.

### Existing feedback

I did not duplicate existing threads. The current scenario-filter thread (`r3471703286`) and configured-filter resume thread (`r3480768093`) appear addressed at this head by materialized filtering plus persisted selection snapshots. The outdated scenario metadata and prior-result extraction threads also appear addressed. The existing scenario selection thread (`r3481014048`) still applies: the exact-head CLI QA below produced two effective rows for `filterFirstN: 1` because selection is applied once per scenario config bucket.

The new persistence boundary needs round-trip tests, not only in-memory selection assertions: cover external `--tests` providers with secrets/relative paths, default-only resume without metadata, and mixed pass/fail duplicate rows. The touched filtering functions also emit Biome complexity warnings; extracting selection snapshot/replay normalization would make these edge cases easier to keep aligned.

### Verification

- `npx vitest run test/commands/optimize.test.ts test/optimizer/promptOptimizer.test.ts test/providers/pythonCompletion.test.ts test/util/eval/filterTests.test.ts test/util/config/load.test.ts test/commands/eval/evaluateOptions.test.ts test/server/eval.test.ts --sequence.shuffle=false` — exit 0; 7 files, 382 tests passed.
- `npm run tsc` — exit 0.
- `npm run build` — exit 0, including app build and postbuild.
- `npx biome check <20 changed TS/test/JSON files>` — exit 0 with six cognitive-complexity warnings; no fixes applied. `npx prettier --check site/docs/usage/command-line.md` — exit 0.
- `npm run local -- optimize -c /private/tmp/pr9373-optimize.yaml --validation-split 0.5` against an isolated localhost OpenAI-compatible mock — exit 0; two filtered tests split 1/1, three rounds completed, validation score improved 0 to 1, and no user cache/database was used.
- `npm run local -- eval -c /private/tmp/pr9373-scenario-filter.yaml --no-cache --no-write --no-progress-bar -o /private/tmp/pr9373-scenario-output.json` — exit 0; exported JSON contained two passing rows (`alpha/one`, `beta/one`), each score 1 with no error, confirming the existing per-config-bucket `firstN` defect.
- Isolated implicit-default replay probe — initial 1/1 pass, then expected probe exit 42 with `snapshot=[{}]` and `readTest()`'s empty-test rejection.
- Isolated external-provider snapshot probe — exit 0; confirmed the synthetic credential marker was persisted and initial/replay provider paths differed.
- Isolated mixed-result duplicate probe — expected exit 43; selected `['failed row', 'passing row']` instead of only the failed row.

The requested `/pr-review-toolkit:review-pr all parallel` command was attempted but unavailable in this environment (exit 127), so I performed the documented six-lens read-only fallback and reconciled three independent exact-head specialist reviews. I did not run the full 20k-test suite, call a live external model/provider, or perform browser QA; those are the material untested areas.

PR #9170, opened 2026-05-10
feat(eval): add output detail deep links

ianw-oai:
## Summary
- add stable hash deep links for eval output detail dialogs
- keep deep-link targeting tied to persisted row identity and preserve router history state
- clear stale detail hashes when search, filters, or active views change the visible result set

## Validation
- `npm run test:app -- src/pages/eval/components/utils.test.ts src/pages/eval/components/Eval.test.tsx src/pages/eval/components/EvalOutputCell.test.tsx src/pages/eval/components/ResultsTable.test.tsx --run`
- `npm run l`
- `npm run f`
- `npm run local -- eval -c test/smoke/fixtures/configs/simple-repeat.yaml -o /private/tmp/pr9170-simple-repeat.json --no-cache`

PR #9377, opened 2026-05-22
fix(webui): clear stale eval detail row hint

wholley-oai:
## Summary
This fixes a regression in the new eval output deep-link flow.

Closing an eval output details dialog cleared the `#details-row-...` hash, but it left the `?rowId=` query parameter behind. That meant the URL still looked like a deep link to a specific filtered row even after the dialog was closed.

## Root cause
This regression was introduced by commit [`1f5845b9f508a01de0b821c6671eae438fb7109c`](https://github.com/promptfoo/promptfoo/commit/1f5845b9f508a01de0b821c6671eae438fb7109c) in PR [#9170](https://github.com/promptfoo/promptfoo/pull/9170) (`feat(eval): add output detail deep links`).

The shared `setEvalDetailsHash()` helper was updated to add `rowId` when opening a deep-linked result, but the clear path only removed the hash fragment. The stale query param remained in the URL until some other navigation path explicitly removed it.

## Changes
- remove `rowId` when `setEvalDetailsHash('')` clears the deep-link state
- add a regression test that proves both the hash and `rowId` are removed together

## Impact
Users who open and close eval result details no longer keep a stale deep-link row hint in the URL.

## Validation
- Added regression coverage in `src/app/src/pages/eval/components/utils.test.ts`
- Local Vitest execution is currently blocked in this worktree because dependency resolution fails against the registry mirror for `@anthropic-ai/sdk@^0.97.1`

### DIFF
--- a/src/app/src/pages/eval/components/utils.test.ts
+++ b/src/app/src/pages/eval/components/utils.test.ts
@@ -148,6 +148,17 @@ describe('eval-details hash store', () => {
     expect(window.location.hash).toBe('#details-row-51-prompt-2');
   });
 
+  it('clears the stale row hint when removing the details hash', () => {
+    window.history.replaceState({}, '', '/eval/x?rowId=51#details-row-51-prompt-2');
+
+    act(() => {
+      setEvalDetailsHash('');
+    });
+
+    expect(window.location.search).toBe('');
+    expect(window.location.hash).toBe('');
+  });
+
   it('coordinates updates across multiple subscribers', () => {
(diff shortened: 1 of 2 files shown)


kmbroai (approved):

mldangelo-oai (approved):
LGTM. I reviewed the URL-state cleanup and regression coverage, and verified the affected Web UI suites locally (266 tests passing). Current main has advanced only through unrelated Google provider changes, with no overlap in this fix.

PR #9378, opened 2026-05-22
feat(eval): add cent cost display metrics for eval results

### LINKED ISSUE #9242: Add an option to show cost in ¢ instead of $
First I'd like to say thanks for this amazing library. It's saved us so much time tweaking our prompts ❤️ 

Two related features:
1. In our application we're dealing with high volumes of very cheap requests and we therefore reason in fractions of cents per request, not per dollar. Currently, cost is always displayed in dollars, so it adds some cognitive overhead to mentally remove the leading digits to get the cost in (fractions of) cents
2. For individual runs it's important for us to know "how many times can you run this when spending 1 dollar cent?". So we write a `runsPerDollarCent` metric into the metadata. But it's hidden behind two clicks and not visible in the main view

Ideally this:
1. Add an option to switch cost display between dollars and cents
2. Add an option to calculate "runs per X" and show it next to the cost per run. X would be either dollars or cents depending on the option above

We already put this information into our metadata but it's cumbersome to access for a metric that's important enough for us that we'd like to see it in the main view

Thanks for considering!

agold-oai:
## Summary
- add persisted eval results settings for cent cost display and runs-per-cost-unit metrics
- centralize app-side cost formatting and inverse-rate calculation while keeping stored costs in USD
- show unit-aware cost details in result cells and prompt summaries, with docs and focused coverage

## Why
Issue #9242 asks for cents to be easier to inspect in the Web UI for low-cost evals. This keeps the evaluator, providers, exports, and stored cost values unchanged and treats cents and runs-per-unit as display-layer views over USD costs.

Fixes #9242

## Validation
- `../../node_modules/.bin/vitest run src/utils/cost.test.ts src/pages/eval/components/TableSettings/components/SettingsPanel.test.tsx src/pages/eval/components/TableSettings/hooks/useSettingsState.test.ts src/pages/eval/components/EvalOutputCell.test.tsx src/pages/eval/components/ResultsTable.test.tsx src/pages/eval/components/store.test.ts` from `src/app`
- `../../node_modules/.bin/tsc --noEmit` from `src/app`
- `node_modules/.bin/biome check ...` on the changed TypeScript files
- `git diff --check`
- manual Web UI check for cents/runs toggles, persistence after reload, and Reset to Defaults


mldangelo:
Reviewed (advisory only). LGTM.

- Frontend-only change; stored costs remain USD, this is purely a presentation layer.
- Cost utility has solid test coverage for `undefined`, `0`, `NaN`, `Infinity`, fractional cents.
- `Intl.NumberFormat` cache keyed on `JSON.stringify(options)` is fine for the small option-set.
- Persisted Zustand store updates correctly reset-to-defaults.
- Implicitly fixes the prior `cost ? ...` truthiness bug (would have hidden `$0`) via the new `!cost || !Number.isFinite(cost) || cost <= 0` guard.

CI green. Independent of all other PRs in the current queue.

mldangelo-oai (approved):
LGTM. The cents/runs-per-unit display addresses the core issue and keeps stored costs in USD.

Non-blocking nit: in dollar mode, ordinary amounts can still render as `$2.5` rather than `$2.50`. Since this PR centralizes cost formatting, would it be worth formatting normal dollar amounts with conventional cents while preserving extra precision for very small per-run costs? Fine as a follow-up.

PR #9388, opened 2026-05-23
feat(providers): expose Codex transcript metadata

ianw-oai:
Expose each Codex provider's current-turn transcript through `metadata.messages` for the eval UI, preserve app-server metadata redaction for sensitive transcript content, and document the separate emitted reasoning and trajectory evidence surfaces.

### DIFF
--- a/src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
+++ b/src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
@@ -327,7 +327,8 @@ export default function EvalOutputPromptDialog({
   let parsedMessages: Message[] = [];
   try {
     const messagesValue = metadata?.messages;
-    parsedMessages = JSON.parse(typeof messagesValue === 'string' ? messagesValue : '[]');
+    const messages = typeof messagesValue === 'string' ? JSON.parse(messagesValue) : messagesValue;
+    parsedMessages = Array.isArray(messages) ? (messages as Message[]) : [];
   } catch {}
 
   const citationsData = metadata?.citations as Citation | Citation[] | undefined;
--- a/src/providers/openai/codex-app-server.ts
+++ b/src/providers/openai/codex-app-server.ts
@@ -155,6 +155,11 @@ type CodexAppServerUserInput =
       path: string;
     };
 
+type CodexConversationMessage = {
+  role: 'user' | 'assistant';
+  content: string;
+};
+
 type CodexAppServerThreadCleanup = 'unsubscribe' | 'archive' | 'none';
 type CodexAppServerUserInputPolicy = 'empty' | 'first-option' | Record<string, string | string[]>;
 
@@ -2733,7 +2738,13 @@ export class OpenAICodexAppServerProvider implements ApiProvider {
         ? { notifications: this.sanitizeForMetadata(state.notifications) }
         : {}),
     };
-    const metadata = this.buildResponseMetadata(state, threadHandle, config, normalizedItems);
+    const metadata = this.buildResponseMetadata(
+      state,
+      threadHandle,
+      config,
+      output,
+      normalizedItems,
+    );
 
     return {
       output,
@@ -2766,6 +2777,7 @@ export class OpenAICodexAppServerProvider implements ApiProvider {
     state: CodexAppServerTurnState,
     threadHandle: ThreadHandle,
     config: CodexAppServerConfig,
+    output: string,
     normalizedItems: Record<string, unknown>[],
   ): ProviderResponse['metadata'] {
     const skillMetadata = this.buildSkillMetadata(
@@ -2773,6 +2785,7 @@ export class OpenAICodexAppServerProvider implements ApiProvider {
       this.getSkillRootPrefixes(config, state.appServerEnv),
     );
     return {
+      messages: this.buildConversationMessages(state, output),
       codexAppServer: {
         threadId: threadHandle.threadId,
         turnId: state.turnId,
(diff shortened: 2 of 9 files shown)


mldangelo:
Reviewed (advisory only — can't push to your branch). Mostly LGTM but **flagging one security gap** for follow-up before merge.

### App-server vs SDK redaction asymmetry

`buildConversationMessages` in `codex-app-server.ts` ends with:
```ts
return this.sanitizeForMetadata(messages) as CodexConversationMessage[];
```

The equivalent helper in `codex-sdk.ts` returns the array **without** that sanitize call:
```ts
return messages;
```

The app-server unit test asserts `api_key=...` in assistant text gets redacted to `[REDACTED]`. The same input through the SDK path would currently leak into `metadata.messages` and surface in the Web UI Messages tab.

### Suggested fix

In `src/providers/openai/codex-sdk.ts`, wrap the `buildConversationMessages` return:

```ts
return this.sanitizeForMetadata(messages) as CodexConversationMessage[];
```

…matching the app-server helper. Then add a parallel redaction unit test in `test/providers/openai-codex-sdk.test.ts` asserting that an `agent_message` containing `api_key=sk-...` ends up `[REDACTED]` in the returned `metadata.messages`.

### Otherwise

- UI `EvalOutputPromptDialog` accepting either JSON string or already-parsed array is a good defensive change.
- Both app-server and UI tests look solid.
- No merge-order issue with #9353/#9334 specifically — but you'll likely conflict in `codex-app-server.ts` (overlap with #9334's `buildResponseMetadata` and #9353's skill helper). Best to rebase after #9353 lands.

mldangelo-oai (commented):
Follow-up required before I can approve this head: `OpenAICodexSDKProvider` still copies raw prompt and agent text into persisted/UI-visible `metadata.messages`, while only the app-server path is sanitized. Please sanitize the SDK transcript metadata path and add a regression test containing a secret/token value, then re-request approval. Reviewed current head `7492eca7b4f3567a33937bfbe912eb8fcd3b88c4`.

src/providers/openai/codex-app-server.ts
```diff
@@ -2794,6 +2807,25 @@ export class OpenAICodexAppServerProvider implements ApiProvider {
     };
   }
 
+  private buildConversationMessages(
+    state: CodexAppServerTurnState,
+    output: string,
+  ): CodexConversationMessage[] {
+    const messages: CodexConversationMessage[] = [
+      { role: 'user', content: this.formatPromptInputForTrace(state.promptInput) },
```
mldangelo-oai, line 2815: **[P2] Keep inline image bytes out of transcript text**

This reuses `formatPromptInputForTrace`, which interpolates the complete `image.url`. A supported inline image such as `data:image/png;base64,...` therefore becomes `[image: data:image/png;base64,...]` in `metadata.messages`. The recursive blob extractor only externalizes strings that themselves begin with `data:`, so the wrapper prevents extraction and the full base64 survives into SQLite, exports/share payloads, and the Messages tab. That adds an avoidable multi-megabyte copy per image row and renders the base64 as text. Please emit a bounded inline-image placeholder here, or preserve the image as a structured message whose content can be externalized. A regression test should verify that `turn/start` still receives the full data URL while persisted transcript metadata does not contain its base64 payload.

src/app/src/pages/eval/components/EvalOutputPromptDialog.tsx
```diff
@@ -327,7 +327,8 @@ export default function EvalOutputPromptDialog({
   let parsedMessages: Message[] = [];
   try {
     const messagesValue = metadata?.messages;
-    parsedMessages = JSON.parse(typeof messagesValue === 'string' ? messagesValue : '[]');
+    const messages = typeof messagesValue === 'string' ? JSON.parse(messagesValue) : messagesValue;
+    parsedMessages = Array.isArray(messages) ? (messages as Message[]) : [];
```
mldangelo-oai, line 331: **[P2] Validate structured messages before rendering them**

This checks only the outer array and then asserts every element is a `Message`. `metadata` is extensible and can come from tests, custom providers, or imported historical results, so values such as `[null]` or `{ role: 'user', content: [{ type: 'text', text: 'hello' }] }` are realistic. The former reaches `ChatMessage` and throws at `message.role`; the latter can reach JSX as an object child and throw there. Before this change, non-string arrays were ignored. Please add a runtime guard/normalizer for role and content (filtering or safely converting unsupported entries) before making the Messages tab visible, with tests for null, primitive, multipart-content, and mixed valid/invalid arrays.

src/providers/openai/codex-sdk.ts
```diff
@@ -2136,29 +2148,52 @@ export class OpenAICodexSDKProvider implements ApiProvider {
...
+    const messages: CodexConversationMessage[] = [
+      { role: 'user', content: this.formatPromptInputForTrace(promptInput) },
+    ];
+    const assistantTexts: string[] = Array.isArray(turn.items)
+      ? turn.items
+          .filter((item: any) => item?.type === 'agent_message' && typeof item.text === 'string')
+          .map((item: any) => item.text as string)
+      : [];
+
+    messages.push(...assistantTexts.map((content) => ({ role: 'assistant' as const, content })));
```
mldangelo-oai, line 2187: **[P2] Do not count each pair of transcript entries as a turn**

Keeping every emitted assistant message is useful, but it creates a valid one-turn shape like `[user, assistant-intermediate, assistant-final]` (the new streaming test asserts exactly this). `EvalOutputPromptDialog` enables turn counts whenever the length is greater than two, computes `ceil(length / 2)`, and `ChatMessages` inserts a divider at every even index. The final assistant message is therefore shown as `2/2` even though there was only one user/Codex turn. Count boundaries from `role === 'user'` (or omit counters for a single-user provider transcript) so consecutive assistant messages stay in the same turn. Please cover both `[user, assistant, assistant]` and a true two-user-turn sequence.

mldangelo-oai (commented):
Reviewed exact head `7492eca7b4f3567a33937bfbe912eb8fcd3b88c4` (comment only; I did not modify or push to this branch).

The feature is not superseded on current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`), but I would not merge this head. The existing P1 SDK transcript-redaction finding is still unresolved: the app-server follow-up sanitizes `metadata.messages`, while the SDK still persists raw prompt/intermediate assistant text. The [existing inline thread](https://github.com/promptfoo/promptfoo/pull/9388#discussion_r3292103259) and [prior advisory](https://github.com/promptfoo/promptfoo/pull/9388#issuecomment-4531887703) remain actionable, so I have not duplicated that comment.

I added three new P2 comments for (1) inline image data URLs bypassing blob externalization after being embedded in transcript text, (2) unchecked structured message arrays crashing the details dialog, and (3) consecutive assistant messages being mislabeled as separate turns.

Local exact-head validation passed: 185 focused Codex provider tests, 56 EvalOutputPromptDialog tests, TypeScript typecheck, core package build, and app build. Those passes do not cover the cases above. GitHub reports this branch as `CONFLICTING`/`DIRTY`; it is 428 commits behind current main, the live checks are from May 23, and `src/providers/openai/codex-sdk.ts` has a real merge conflict. Given the security gap and provider drift, the safest path is to port the feature onto current main, share one sanitizer/typed transcript boundary across both providers, address the three inline cases, and rerun current CI.

PR #9398, opened 2026-05-23
feat(ui): improve dialog, helper-text, and number-input primitives

mldangelo:
## Summary

Tighten three Promptfoo UI primitives so the eval-creator and redteam forms in subsequent PRs can rely on them for accessibility-correct messaging and validation.

- **`Dialog`** — allow an intentionally visible `DialogDescription` to describe dialog content while keeping `aria-describedby` unset when no description is rendered.
- **`HelperText`** — let the `error` flag add `role="alert"` and destructive styling while preserving explicitly requested roles.
- **`NumberInput`** — forward `aria-invalid`, `aria-describedby`, and `id`, associate helper/error feedback with the input, and keep hook order stable when the `id` prop changes.

## Stack

This PR targets `main` and is the base for [#9399](https://github.com/promptfoo/promptfoo/pull/9399) and [#9400](https://github.com/promptfoo/promptfoo/pull/9400).

## Test plan

- [x] `npm run test:app -- src/components/ui/dialog.test.tsx src/components/ui/helper-text.test.tsx src/components/ui/number-input.test.tsx --run` — 46/46 pass
- [x] `npm run l`
- [x] `npm run tsc -- --pretty false`
- [x] `npm run build:app`
- [x] `git diff --check`

`npm run f` ran Biome over the changed TypeScript with no edits, then hit the repository wrapper's known empty Prettier-target error because this diff has no Prettier-target files.


mldangelo-oai:
@codex review

mldangelo-oai:
Synchronized and reviewed current head `0834fe862`.

- Merged current `origin/main` (`6a66e3a0a`) into this stacked branch as requested. The only merge conflict was inherited test coverage in `test/server/routes/eval.sharing.test.ts`; the resolution retains main's regression for `toEvaluateSummary()` rejection and passes its focused suite.
- The intended child delta relative to parent PR #9397 remains the six `src/app/src/components/ui/{dialog,helper-text,number-input}*` files. Until the parent branch is synchronized in this sweep, GitHub will temporarily show the merged-main surface in this stacked child PR.
- Critical review found no additional actionable correctness, accessibility, or simplification issue in the UI primitive delta.

Validation:

- `npm run l` and both stack-relative/current-main `git diff --check`
- `npm run test:app -- src/components/ui/dialog.test.tsx src/components/ui/helper-text.test.tsx src/components/ui/number-input.test.tsx src/pages/redteam/setup/components/RunOptions.test.tsx src/pages/redteam/setup/components/Targets/WebSocketEndpointConfiguration.test.tsx --run` (`98/98`)
- full `npm run test:app -- --run` (`4,653/4,653`)
- `npm run tsc` and `npm run build`
- fresh isolated `npm run db:migrate` under `/tmp/pr9398-migration-state`
- live browser QA on `/redteam/setup#5`: invalid `Number of test cases` is linked with `aria-describedby`, reports `aria-invalid="true"`, renders the validation feedback as `role="alert"`, and produced no console/page errors.

`npm run f` ran Biome over the changed TypeScript with no edits, then hit the repository wrapper's known empty Prettier-target error because this diff has no Prettier-target files.

src/app/src/components/ui/number-input.tsx
```diff
@@ -60,9 +60,12 @@ function NumberInput({
   readOnly,
   id,
   onKeyDown,
+  'aria-describedby': ariaDescribedBy,
+  'aria-invalid': ariaInvalid,
   ...props
 }: NumberInputProps) {
   const inputId = id || React.useId();
+  const feedbackId = React.useId();
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `0578aabc5`. `NumberInput` now calls the fallback input ID hook unconditionally before choosing between the generated ID and the supplied `id`, so changing the prop cannot alter hook order. Added a rerender regression test and verified the 46-test primitive suite, lint, TypeScript, and a production app build.

src/app/src/components/ui/dialog.tsx
```diff
@@ -117,12 +143,36 @@ function DialogTitle({
...
 function DialogDescription({
   className,
+  id,
   ref,
   ...props
 }: React.ComponentProps<typeof DialogPrimitive.Description>) {
+  const generatedId = React.useId();
+  const descriptionId = id ?? generatedId;
+  const descriptionRef = React.useRef<HTMLParagraphElement>(null);
+  React.useImperativeHandle(ref, () => descriptionRef.current!);
```
jameshiester-oai, line 153: [P1] Keep callback refs stable across ordinary renders

With no dependency list, this imperative handle is cleaned up and recreated after every render, so a stable caller callback ref receives `null` and then the same `<p>` again even when the DOM node did not change. The exact-head probe saw two extra callback invocations after one unrelated rerender; a callback ref that records derived node state then triggers another render and loops (guarded after 30 assignments in the probe). This regresses the prior direct-ref behavior and can crash dialogs using measurement/registration refs. Please compose the internal and external DOM refs in the actual ref callback (or use a stable composed-ref helper) so callers are notified only on node/ref changes, and add both callback-ref stability and state-update regressions.

src/app/src/components/ui/dialog.tsx
```diff
@@ -35,31 +39,53 @@ function DialogContent({
...
   ...props
 }: React.ComponentProps<typeof DialogPrimitive.Content> & {
   /** When true (default), adds a visually hidden description to suppress Radix a11y warnings */
   hideDescription?: boolean;
   /** When true, hides the default close button (useful when providing a custom close button) */
   hideCloseButton?: boolean;
 }) {
+  const [descriptionIds, setDescriptionIds] = React.useState<string[]>([]);
+  const registerDescription = React.useCallback((id: string) => {
+    setDescriptionIds((currentIds) => (currentIds.includes(id) ? currentIds : [...currentIds, id]));
```
jameshiester-oai, line 52: [P2] Keep description IDs in current DOM order

This append-only registration order diverges from DOM order when existing keyed descriptions move or a new one is inserted before them. Starting with description B and then inserting keyed description A before it left the DOM as `A B`, but `aria-describedby` became `B A`; the computed accessible description was likewise read as `B A`. That can reverse the intended order of instructions and validation details for assistive technology. Derive/sort registered IDs by the mounted nodes' current document order (and update on reordering), with a conditional-insertion regression.

src/app/src/components/ui/dialog.tsx
```diff
@@ -117,12 +143,36 @@ function DialogTitle({
 
 function DialogDescription({
   className,
+  id,
   ref,
   ...props
 }: React.ComponentProps<typeof DialogPrimitive.Description>) {
+  const generatedId = React.useId();
+  const descriptionId = id ?? generatedId;
```
jameshiester-oai, line 151: [P3] Fall back when the custom ID cannot be an IDREF

`id ?? generatedId` preserves `""` and whitespace-containing values. The former is then omitted from registration and the latter is split into multiple nonexistent `aria-describedby` tokens; with `hideDescription={false}`, both cases leave the visible description unannounced and fail silently because the internal hidden Radix description suppresses warnings. Treat only one non-empty, whitespace-free token as usable and otherwise use the generated ID (or reject/dev-warn). Apply the same validation to the effective mounted ID for `asChild`, and cover empty/whitespace IDs.

jameshiester-oai (commented):
## Review scope

Reviewed the full PR diff from recorded base `d27868b7fdd2f7fb047550fdeb76bebcd3138257` to exact head `e52b426389b7d64c6e310d996bcc73fd3241301e`. This was a fresh full review: no authenticated completion marker existed for this head. Current `origin/main` is `4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6`; the PR is mergeable but behind, and current main has no intervening edits to these six UI files.

## Verdict

The NumberInput and HelperText behavior is coherent, the ordinary Dialog description paths are well covered, and every exact-head CI check is green. I found one P1 ref/lifecycle regression, one P2 accessible-description ordering defect, and one P3 invalid-ID robustness gap in the new Dialog registration layer.

## Verified findings

### P1

- `src/app/src/components/ui/dialog.tsx:152-159`: `useImperativeHandle` has no dependency list, so a stable callback ref is called with `null` and the same node again on every unrelated render. A callback ref that records derived node state re-renders the parent and immediately loops; an exact-head probe hit a guarded `callback ref assignment loop` after 30 assignments. Compose the internal and external DOM refs instead of rebuilding the imperative handle each render.

### P2

- `src/app/src/components/ui/dialog.tsx:51-56`: description IDs are retained in registration order rather than DOM order. Starting with description B and conditionally inserting keyed description A before it produced DOM order `A B`, but `aria-describedby="B A"` and computed accessible text `B A`. Assistive technology can therefore read instructions or errors in the reverse of the visible order.

### P3

- `src/app/src/components/ui/dialog.tsx:150-158`: an empty or whitespace-containing custom description ID is accepted but cannot form a valid single IDREF. With `hideDescription={false}`, both `id=""` and `id="bad id"` left visible description text unassociated and produced an empty accessible description. Fall back to the generated ID (or reject/dev-warn) when the effective ID is not one non-empty token, including the `asChild` path.

Three inline comments contain reproductions and correction directions.

## Existing feedback

All five prior threads are resolved/outdated, and four fixes remain effective. The dynamic-ID thread is only partially fixed: https://github.com/promptfoo/promptfoo/pull/9398#discussion_r3346038774. On the exact head, an `asChild` component that forwards its ref but changes its own DOM `id` from local state leaves `aria-describedby` on the removed old ID because `DialogDescription` itself does not rerender. I did not duplicate that existing finding inline; the registration model should observe the mounted node or make the wrapper ID authoritative.

A composed ref plus a DOM-derived/observed description registry would remove the imperative-handle churn and keep identity/order synchronized without a layout effect that schedules state on every wrapper render. Missing regressions include a stable callback ref across unrelated renders, callback-ref derived state, keyed insertion/reordering, child-local `asChild` ID changes, and empty/whitespace IDs.

## Validation

- `npm run test:app -- src/components/ui/dialog.test.tsx src/components/ui/helper-text.test.tsx src/components/ui/number-input.test.tsx src/pages/redteam/setup/components/RunOptions.test.tsx src/pages/redteam/setup/components/Targets/WebSocketEndpointConfiguration.test.tsx --run` — 5 files, 104/104 passed.
- Isolated 12-case JSDOM/DOM-accessibility harness — six lifecycle/control cases passed; expected failures reproduced callback-ref reassignment/looping, reversed keyed-insertion order, child-local `asChild` ID drift, and empty/whitespace ID loss.
- `npm run build:app` — passed.
- `npx @biomejs/biome check` on all six changed files — passed with no fixes.
- `git diff --check` — passed; final worktree clean.
- Current-main merge analysis was conflict-free, and all exact-head GitHub checks are green across web UI, builds, platform tests, style, integration/smoke, docs, and CodeQL.

## Limitations

The requested PR Review Toolkit command was unavailable, so I used a manual six-lens fallback and reconciled three independent exact-head reviews. Port 3000 currently serves OWASP Juice Shop rather than Promptfoo, so I did not replace the existing process or run live Promptfoo browser/screen-reader QA. No full local app suite or post-merge checkout was run; exact-head CI supplies the broader suite coverage.

PR #9399, opened 2026-05-23
feat(redteam): improve a11y, validation, and secret handling in Targets setup

mldangelo:
## Summary

Reshape the redteam Targets setup surface (provider-type selection, HTTP / WebSocket / browser / agent / foundation-model configuration, custom targets, Postman import, extensions, request transforms, status-code matching, authorization, TLS) so that:

- **Accessible by default.** Required fields announce via `aria-required`, validation messages render with `role="alert"`/`aria-live` and are associated to inputs via `aria-describedby` / `aria-invalid`. Dialogs carry visible `DialogTitle`/`DialogDescription` pairs and describe their consequences (deletion impact, draft loss).
- **Secrets opt out of autofill.** `SensitiveTextField` and custom password/API-key inputs set `autoComplete="new-password"`, `spellCheck={false}`, and `data-1p-ignore` / `data-lpignore` / `data-form-type="other"`. These are not login passwords; saving them as such leaks them outside the app.
- **Draft-safe navigation.** Provider type selector preserves drafts on back-navigation; HTTP setup mode switching is atomic; custom-provider JSON drafts reset correctly when changing provider types.
- **Tighter validation.** Auto-fill, transform, status-code, TLS, API-key, A2A, and custom-configuration editors share validation contracts and block invalid saves without showing misleading defaults.
- **Better imports.** The Postman import dialog clarifies what it accepts and surfaces field-level errors at the offending row.

## Stack

Built on [#9398](https://github.com/promptfoo/promptfoo/pull/9398). The next PR in the stack is [#9400](https://github.com/promptfoo/promptfoo/pull/9400), which wires the eval-creator to the updated provider editor contracts.

## Test plan

- [x] `npm run test:app -- src/pages/redteam/setup/components/Targets --run` — **511/511 pass** across 29 files.
- [x] Focused editor/security regression suite — **119/119 pass** across 4 files.
- [x] `npm run test:app -- --run` — **4,912/4,912 pass** across 284 files.
- [x] `npm run tsc -- --pretty false`
- [x] `npm run build:app`
- [x] `npm run l` and `npm run f` (the existing `TlsHttpsConfigTab` cognitive-complexity warning remains non-blocking).

## Notes

- The diff is large because every Targets surface received the same accessibility and validation treatment.
- The branch is refreshed onto the latest #9398 parent and includes the parent stack's current `main` merge.

PR #9400, opened 2026-05-23
feat(eval): rebuild eval-creator UX on shared primitives and readiness modules

mldangelo:
## Summary

Rebuild the eval-creator experience on the shared UI primitives from [#9398](https://github.com/promptfoo/promptfoo/pull/9398) and the target/configuration foundations from [#9399](https://github.com/promptfoo/promptfoo/pull/9399).

- **Step-by-step setup** — providers → prompts → test cases → run options, driven by `SetupReadiness` with actionable issue routing.
- **End-to-end draft protection** — provider, prompt, test-case, YAML, environment, and run-option drafts are preserved or gated by explicit discard confirmation.
- **Route-aware readiness** — prompt variables and base-request estimates respect per-test prompt/provider routing, static variable expansion, and repeat settings; exact estimates are hidden when runtime-dependent inputs make request volume unknowable.
- **YAML save merges, upload replaces** — editor saves preserve fields the editor does not surface; imports use explicit full-replace semantics.
- **Broader assertion UX** — specialized fields, field-level validation, prerequisite checks, and clearer model-graded cost guidance.
- **Review fixes** — external vars references now satisfy dialog and row checks, invalid zero concurrency drafts remain visible and block runs, blank legacy prompt-map labels round-trip, provider editing infers the correct configuration UI for supported provider IDs and JS/TS module paths, and base-request estimates match the evaluator's var-expansion behavior.

## Stack

Top of the UI stack. Builds on [#9398](https://github.com/promptfoo/promptfoo/pull/9398) → [#9399](https://github.com/promptfoo/promptfoo/pull/9399).

The branch is refreshed to the current #9399 head (`8b18083eb2b5211930c69bdac0dec35a36ab6dd1`).

## Test plan

- [x] `npm run test:app -- src/pages/eval-creator/components/AddProviderDialog.test.tsx --run` — **83/83 pass**
- [x] `npm run test:app -- src/pages/eval-creator --run` — **640/640 pass** across 22 files
- [x] `npm run test:app -- --run` — **5,075/5,075 pass** across 285 files
- [x] `npm run l`
- [x] `npm run tsc -- --pretty false`
- [x] `npm run build:app`
- [x] `git diff --check` and `git diff --cached --check`
- [x] Pre-commit lint/format hook passed


src/app/src/pages/eval-creator/components/TestCaseDialog.tsx
```diff
@@ -4,116 +4,297 @@ import { Button } from '@app/components/ui/button';
...
+  const effectiveInheritedAssertions =
+    initialValues?.options?.disableDefaultAsserts === true ? [] : inheritedAssertions;
+  const effectiveAssertions = [...effectiveInheritedAssertions, ...asserts];
+  const assertionVariables = getRequiredAssertionVariables(effectiveAssertions);
+  const editableVarsList = Array.from(new Set([...varsList, ...assertionVariables]));
+  const persistedVars = getPersistedVars(vars, initialValuesVars, touchedVars);
+  const missingAssertionVariables = getMissingAssertionVariables(effectiveAssertions, {
+    ...inheritedVars,
+    ...persistedVars,
+  });
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `ae62ddff1`. `TestCaseDialog` now treats inherited or local external vars references as satisfying context-assertion prerequisites and safely ignores non-record vars instead of spreading them. Added regression coverage in `TestCaseDialog.test.tsx`.

src/app/src/pages/eval-creator/components/RunOptionsSection.tsx
```diff
@@ -1,126 +1,295 @@
...
-  isReadyToRun = false,
+  readiness,
 }: RunOptionsSectionProps) {
-  const canSetDelay = !maxConcurrency || maxConcurrency === 1;
-  const canSetMaxConcurrency = !delay || delay === 0;
+  const [showOptionalSettings, setShowOptionalSettings] = useState(
+    Boolean(description || delay || maxConcurrency),
+  );
+  const [delayDraft, setDelayDraft] = useState(delay?.toString() || '');
+  const [maxConcurrencyDraft, setMaxConcurrencyDraft] = useState(maxConcurrency?.toString() || '');
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `ae62ddff1`. `RunOptionsSection` now preserves numeric `0` in initial and synchronized drafts, keeps optional settings visible for defined invalid values, and blocks running until the concurrency value is valid. Added regression coverage in `RunOptionsSection.test.tsx`.

src/app/src/pages/eval-creator/components/TestCasesSection.tsx
```diff
@@ -80,17 +93,221 @@ const getManagedTestsLabel = (tests: unknown): string => {
...
+  missingAssertionVariables: string[];
+}
+
+function getTestCaseIssues(
+  testCase: TestCase,
+  requiredPromptVariables: string[],
+  defaultTest: TestCase | undefined,
+): TestCaseIssues {
+  const testCaseVars = Object.keys(testCase.vars || {});
+  const defaultTestVars = Object.keys(defaultTest?.vars || {});
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `ae62ddff1`. `TestCasesSection` now safely normalizes vars values and treats external test/default vars references as satisfying both prompt-variable and assertion-variable row checks. Added regression coverage in `TestCasesSection.test.tsx`.

src/app/src/pages/eval-creator/components/RunOptionsSection.tsx
```diff
@@ -1,126 +1,298 @@
...
+                <dd className="mt-1 text-xl font-semibold">{value}</dd>
+              </div>
+            ))}
+          </dl>
+
+          {readiness.plannedBaseRequestCount !== undefined && readiness.isReadyToRun && (
+            <p className="text-sm text-muted-foreground">
+              This configuration starts with {readiness.plannedBaseRequestCount} base request
+              {readiness.plannedBaseRequestCount === 1 ? '' : 's'} before any additional
+              model-graded checks.
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in `686ec4f92`. The base-request estimate now includes merged default/test string-array var expansion and `evaluateOptions.repeat`, respects `disableVarExpansion`, and counts `providerOutput` cases as zero provider requests. It also suppresses the exact estimate when prompts, providers, vars, transforms, scenarios, extensions, external defaults, or filter ranges make the runtime request shape unknowable. Added regression coverage in `setupReadiness.test.ts` and clarified the run-summary copy.

jameshiester-oai (commented):
## Review summary

Reviewed base `d47ee6af3519382138a1d692c439228ca1ab12d0` through head `4ffb81fae8243449a1a11c6322eb81e9f92ec66b`. This was a full review of the stacked base...head diff; no prior authenticated completion marker existed for this head.

**Verdict:** CI is fully green, but the current head has one blocking readiness regression and four narrower correctness/data-loss defects.

### Verified findings

**P1**
- The readiness extractor treats the documented runtime `_conversation` variable and Nunjucks-local bindings such as `{% set input = ... %}` as missing test variables. Because every readiness issue disables Run Evaluation, valid conversational/composed prompts cannot run.

**P2**
- A statically known prompt/provider routing intersection with zero outputs is still declared ready, displays `0 base requests`, and enables a run that cannot produce results.
- Switching between compatible assertion polarities (for example `webhook` to `not-webhook`) deletes the existing value or threshold and forces the user to reconstruct it.
- Recursive YAML anchors pass the new full-config gate and then crash setup when the new reset comparison calls `JSON.stringify` on the cyclic config.
- YAML upload completion uses the state captured when the file read began, so edits made during a delayed read can be replaced without the promised destructive confirmation.

### Existing unresolved feedback

I did not duplicate three still-applicable threads: test-generator objects are presented with an exact request count, empty string-array vars are counted as one request instead of zero, and `.rb` file providers fall through to the generic Python-agent editor. All ten older resolved threads were rechecked in current context.

The variable gate would be safer if it shared a Nunjucks parser/runtime symbol model instead of maintaining a partial lexical scanner. Assertion transitions should derive preservation behavior from source/destination value contracts rather than overlapping type sets. YAML replacement should be one transactional path that validates serializability and consults current store/dirty state at commit time. Regression coverage is needed for runtime globals and `{% set %}`, zero-output route intersections, paired assertion transitions, cyclic aliases, and delayed FileReader completion.

### Validation performed

- `npm run test:app -- src/pages/eval-creator --run` — exit 0; 22 files, 644 tests passed.
- `npm run tsc` — exit 0.
- `npm run build:app` — exit 0; production bundle built (only existing bundler timing/dynamic-import warnings).
- `npm run lint:src` — exit 0; no fixes, 150 repository-wide complexity warnings.
- `git diff --check d47ee6af3519382138a1d692c439228ca1ab12d0...HEAD` — exit 0; worktree clean.
- Direct current-head readiness/YAML probe — reproduced `_conversation` and `{% set %}` false missing-variable issues, ready-plus-zero-route state, and accepted recursive YAML followed by `Converting circular structure to JSON`.
- Out-of-tree component regression probes — reproduced webhook value loss and delayed-import replacement without confirmation.
- `npm run local -- eval -c /private/tmp/pr9400-conversation.yaml --no-cache -o /private/tmp/pr9400-conversation-results.json` — exit 0 with `echo`; 1/1 passed, score 1, output `Conversation history: []`, confirming runtime supplies `_conversation`.

### Limitations

The requested PR Review Toolkit command is not installed, so I used the manual correctness/tests/docs/error-handling/types/simplification fallback plus three independent read-only reviews. Live browser QA was unavailable because port 3000 is occupied by an unrelated SSH listener and the API was not running on 15500; no credential-backed provider call was made.

src/app/src/pages/eval-creator/components/RunTestSuiteButton.tsx
```diff
@@ -57,20 +57,20 @@ const RunTestSuiteButton = () => {
...
   const jobPrompts = normalizePromptsForJob(prompts);
-  const testCount = countTests(tests);
+  const readiness = getSetupReadiness(config);
 
-  const isDisabled =
-    isRunning ||
-    normalizedProviders.length === 0 ||
-    normalizedPrompts.length === 0 ||
-    testCount === 0;
+  const isDisabled = isRunning || !readiness.isReadyToRun || Boolean(disabledReason);
```
jameshiester-oai, line 63: [P1] Do not require runtime or template-local variables

This scanner treats every output identifier except loop locals as a test input. On this head, `Previous: {{ _conversation | dump }}` with `tests: [{}]` produces a missing-variable issue and disables Run Evaluation, even though the evaluator injects `_conversation` at runtime. `{% set input = "hello" %}{{ input }}` is likewise blocked because `set` bindings are not tracked. A real local eval of the `_conversation` prompt passes and renders `[]`. Exclude runtime globals and model `set`/macro/import bindings (preferably from the Nunjucks AST) before gating readiness.

src/app/src/pages/eval-creator/components/setupReadiness.ts
```diff
@@ -247,6 +310,106 @@ function countOutputsForTest(
...
+  }
+
+  const repeatCount = countRepeats(repeat);
+  if (repeatCount === undefined) {
+    return undefined;
+  }
+
+  let requestCount = 0;
+  for (const testCase of tests) {
+    const outputCount = countOutputsForTest(testCase, defaultTest, providers, prompts);
```
jameshiester-oai, line 395: [P2] Treat a statically empty route as incomplete setup

`outputCount === 0` is skipped only for the estimate, so it never becomes a readiness issue. With provider A restricted to `Prompt A` and a test restricted to provider A plus `Prompt B` (all references valid), the current code returns `isReadyToRun: true`, no issues, and `plannedBaseRequestCount: 0`; the UI says setup is complete and enables a run for which the evaluator appends no run options. Add a readiness issue when the effective route intersection is statically empty, while retaining the existing exemptions for `providerOutput` and runtime-dependent shapes.

src/app/src/pages/eval-creator/components/AssertsForm.tsx
```diff
@@ -1,141 +1,1773 @@
...
+    RAG_SCORE_ASSERTION_TYPES.has(nextType) ||
+    MODEL_JUDGE_SCORE_ASSERTION_TYPES.has(nextType) ||
+    TRAJECTORY_GOAL_SUCCESS_ASSERTION_TYPES.has(nextType) ||
+    NAMED_MATCHER_ASSERTION_TYPES.has(nextType) ||
+    NO_VALUE_ASSERTION_TYPES.has(nextType) ||
+    THRESHOLD_ASSERTION_TYPES.has(nextType) ||
+    WORD_COUNT_ASSERTION_TYPES.has(nextType) ||
+    STRUCTURED_VALUE_ASSERTION_TYPES.has(nextType)
+  ) {
+    const { threshold: _threshold, value: _value, ...withoutStoredValue } = assertion;
```
jameshiester-oai, line 1605: [P2] Preserve data when changing only assertion polarity

This destination-only branch always removes `value` and `threshold`, even when the source and destination share the same contract. A component probe starting with `{ type: 'webhook', value: 'https://example.com/validate' }` and selecting `not-webhook` emitted only `{ type: 'not-webhook' }`. The same loss affects `latency`/`not-latency`, `finish-reason`/`not-finish-reason`, and optional JSON/XML counterparts. Preserve fields when source and destination belong to the same value family, and cover both polarity directions.

src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.tsx
```diff
@@ -198,12 +247,16 @@ const EvaluateTestSuiteCreator = () => {
         } else {
           try {
             const parsedConfig = yaml.load(content) as Record<string, unknown>;
-            if (parsedConfig && typeof parsedConfig === 'object') {
-              updateConfig(parsedConfig as Partial<UnifiedConfig>);
-              setResetKey((k) => k + 1);
-              showToast('Configuration loaded successfully', 'success');
+            if (isFullYamlConfig(parsedConfig)) {
```
jameshiester-oai, line 250: [P2] Reject recursive YAML before committing it to the store

`js-yaml` resolves aliases into cyclic object graphs, and this shallow shape check accepts them. A config whose provider `config` aliases the providers array returned `isFullYamlConfig === true`, but the new `JSON.stringify(config)` reset comparison then throws `Converting circular structure to JSON`, taking down the setup route after import (the editor save path shares this validator). Validate that parsed YAML is serializable/acyclic before mutation, and report it as an import error instead of storing it.

src/app/src/pages/eval-creator/components/EvaluateTestSuiteCreator.tsx
```diff
@@ -198,12 +247,16 @@ const EvaluateTestSuiteCreator = () => {
...
+              const importedConfig = parsedConfig;
+
+              if (hasResettableSetup) {
```
jameshiester-oai, line 253: [P2] Recheck replacement safety when the file read completes

The asynchronous `onload` callback closes over `hasResettableSetup` from the render that started the read. In a deferred FileReader probe, I began an upload from the default state, added a prompt before `onload`, and then completed the read; this branch still took the captured `false` path and replaced the new prompt without opening the destructive confirmation. Consult current store/dirty state (or a synchronized ref) at completion time so intervening edits and hydration cannot be silently discarded.

PR #9403, opened 2026-05-23
test: add regression coverage for YAML config empty value handling

Kartalops:
## Summary
- Added regression test coverage for YAML config empty value handling in `test/prompts/processors/yaml.test.ts`
- Tests cover empty string values, null values, and empty YAML documents

## Problem
YAML config files can have empty string values, null values, or be empty documents. These edge cases lacked test coverage in the YAML processing tests.

## Solution
Added three new test cases:
1. `should handle empty string value in YAML` - verifies empty string `""` is preserved
2. `should handle null value in YAML` - verifies `null` and `~` (YAML null) are handled
3. `should handle empty YAML document` - verifies empty content doesn't throw

## Tests
Added tests to existing test file. Run: `npx vitest run test/prompts/processors/yaml.test.ts`

## Risk / Compatibility
- Test-only change
- No runtime behavior changes
- Existing tests unaffected


mldangelo:
Tests were asserting against a `result[0].config` field that doesn't exist — `processYamlFile` returns `{ raw, label, config }` where `raw` is the JSON-stringified parsed YAML and `config` is whatever the caller passes in. Rewrote the three assertions to inspect the parsed raw payload directly; all 13 tests in the suite pass locally.

PR #9408, opened 2026-05-23
fix(config): preserve all function providers in combineConfigs

mldangelo:
## Summary

Fixes #9383. Follow-up hardening on top of #9402 (which merged the foundational fix using reference identity for function providers).

This PR keeps #9402's reference-identity dedupe for raw function providers (so `[fn, fn]` still dedupes to `[fn]`) and adds:

- **Function-bearing object providers** like `{ id, transform: fn }` are never deduped — `JSON.stringify` silently drops the function and would otherwise collapse distinct providers.
- **`ApiProvider` class instances** are never deduped — their identifying methods live on the prototype and `JSON.stringify` doesn't see them.
- **Cycle-safe** `hasFunctionValue` walker (`WeakSet` of visited objects) — old code would have stack-overflowed on a cyclic provider config; new code degrades to "skip dedupe" cleanly.
- **BigInt-safe** dedupe key generation (`JSON.stringify` wrapped in try/catch).
- **Consistent dedupe key for string providers** between the top-level string form (`providers: 'openai:gpt-4'`) and the array form (`providers: ['openai:gpt-4']`).
- **Smoke test** that runs the built CLI against a TS config with three labeled function providers and asserts three distinguishable result rows.

## Test plan

- [x] Unit tests in `test/util/config/load.test.ts` cover: distinct functions preserved, same fn ref deduped (carried from #9402), string-form vs array-form unification, function-bearing object providers, nested function-bearing fields, ApiProvider class instances
- [x] Smoke test in `test/smoke/regression-recent.test.ts` against the built CLI
- [x] Manual audit of every other provider-processing path (`dereferenceConfig`, `loadApiProviders`, `resolveProviderConfigs`, `getProviderIds`, `filterProviderConfigs`, `normalizeCloudEvalProvider`) confirms none has the same bug class

PR #9410, opened 2026-05-23
test: add provider deduplication regression test for same config calls

Kartalops:
## Summary
- Added regression test for provider deduplication behavior
- Tests that duplicate provider configs still call fetchWithCache (verifies cache key uniqueness)

## Problem
Issue #9383 described CallApiFunction cache deduplication problems. A regression test was needed.

## Solution
Added test `should not call API twice for same provider config` to ai21.test.ts that:
1. Creates two provider instances with same config
2. Calls both simultaneously
3. Verifies fetchWithCache is called for each

Note: The test expects 2 calls, verifying that duplicate providers use the same cache key but don't deduplicate at the provider level. This is correct cache behavior.

## Testing
- `npm run test:watch` can verify
- No production code changes

### DIFF
--- a/test/providers/ai21.test.ts
+++ b/test/providers/ai21.test.ts
@@ -259,9 +259,13 @@ describe('AI21ChatCompletionProvider', () => {
     );
   });
 
-  it('should not call API twice for same provider config', async () => {
+  it('invokes fetchWithCache once per call site even for duplicate provider configs', async () => {
     const mockResponse = {
-      body: JSON.stringify({ completions: [{ output: { text: 'test' } }] }),
+      data: {
+        choices: [{ message: { content: 'test response' } }],
+        usage: { total_tokens: 10, prompt_tokens: 5, completion_tokens: 5 },
+      },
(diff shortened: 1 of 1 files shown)


mldangelo:
Thanks for the contribution, @okwn! Nice complementary regression test at the cache-key layer — verifies that distinct provider configs with the same upstream model still produce distinct cache entries, which is a useful guard at a different layer than the `combineConfigs` fix landed in #9402 / #9408.

Synced with main; waiting on CI to land this.

mldangelo:
Test was failing on CI due to a mock-shape mismatch — `fetchWithCache` returns `{ data, cached, status, statusText }` where `data` holds the parsed body, but the mock used `{ body, status, statusText }`, so `data.error` threw "Cannot read properties of undefined". Pushed a fix to the branch and also renamed the test for clarity since it's really verifying that the provider delegates to `fetchWithCache` rather than dedupes itself — actual cache-key dedup happens inside `fetchWithCache`.

PR #9412, opened 2026-05-23
docs(site): link to assertion type quick reference

Kartalops:
## Summary

Adds a direct jump link from the introductory `Using assertions` guidance to the existing authoritative [Assertion types](#assertion-types) table.

## Maintainer Audit Follow-up

- Merged current `origin/main` before review.
- Removed the new duplicate mini-table rather than maintaining two assertion inventories on the same page.
- Eliminated the invalid `containsAny` guidance; the authoritative table correctly documents runtime-supported `contains-any` and links to its details.

## Validation

- Confirmed `contains-any` in `src/types/index.ts`, `src/assertions/index.ts`, and the existing documentation table.
- `npx prettier --check site/docs/configuration/expected-outputs/index.md` passed.
- `git diff --check` passed.
- Attempted `SKIP_OG_GENERATION=true npm run build` from `site/`; its stats prebuild completed, but the build cannot start in this worktree because `docusaurus` is not installed.

## Scope

The effective diff against current `main` is one navigation sentence in `site/docs/configuration/expected-outputs/index.md`, keeping the docs discoverable without adding a second source of truth.


mldangelo:
Thanks for the contribution, @okwn! The link to the authoritative `Assertion types` table is a nice discoverability improvement. Merging.

PR #9414, opened 2026-05-23
fix(providers): serialize persistent browser sessions

ianw-oai:
Serialize browser providers with `persistSession: true` so concurrent evals cannot create or operate on multiple pages in one stateful session, and document the current direct-child semantics of `waitForNewChildren`.


mldangelo:
Reviewed (advisory only). LGTM with one optional polish suggestion.

- Forces `concurrency = 1` when a browser provider with `persistSession: true` is in scope — correct fix for the stateful-session-under-concurrency bug.
- Two new evaluator-execution tests cover both persistent (serializes) and non-persistent (still concurrent) cases.
- Doc additions for `waitForNewChildren` direct-children-only and the persistent-session concurrency note are accurate.

### Optional polish

The check uses `provider.id() === 'browser-provider' && provider.config?.persistSession === true`. `provider.config` isn't part of the `ApiProvider` interface, so this is an untyped `any` access. Not blocking — works correctly — but a typed interface check would be cleaner:

```ts
function hasPersistentBrowserSession(provider: ApiProvider): boolean {
  if (provider.id() !== 'browser-provider') return false;
  const config = (provider as ApiProvider & { config?: { persistSession?: boolean } }).config;
  return config?.persistSession === true;
}
```

CI green. Independent of all other PRs.

mldangelo-oai (approved):
Approving with known follow-up requested: please close the remaining persistent-browser serialization gaps. The concurrency guard should include `tests[].provider` overrides and use the same truthy `persistSession` behavior as the Browser provider, with regression coverage for both cases. Reviewed current head `a2b8d6c7f949eb048f0043b47ab43af1f19f2c1e`.

PR #9418, opened 2026-05-23
feat(examples): baseline-vs-ejentum-harness comparison

ejentum:
## Summary

Adds `examples/baseline-vs-ejentum-harness/`, a comparison eval that runs `gpt-5.4-mini` directly and through an Ejentum-provided reasoning scaffold so users can measure the effect on their own tasks.

## Files

- `examples/baseline-vs-ejentum-harness/provider.mjs`: custom provider that fetches a scaffold from the Ejentum REST gateway, inserts it as an additional system message, and calls an OpenAI-compatible chat endpoint.
- `examples/baseline-vs-ejentum-harness/promptfooconfig.yaml`: baseline and scaffolded providers plus four starter tasks and a rubric.
- `examples/baseline-vs-ejentum-harness/README.md`: setup, required environment variables, endpoint/model options, structured-prompt behavior, and guidance for replacing sample tasks.
- `test/examples/baselineVsEjentumHarnessProvider.test.ts`: provider regression coverage through Promptfoo's custom-provider loader.

## Maintainer Audit Follow-up

- Reject an empty or missing Ejentum scaffold rather than silently running the augmented arm as a baseline request.
- Reject a successful OpenAI HTTP response with missing or blank assistant content rather than recording an empty output as valid.
- Align both comparison arms and the default documentation with `gpt-5.4-mini`, `reasoning_effort: none`, and `verbosity: low`.
- Gate `reasoning_effort` and `verbosity` like the built-in OpenAI chat provider, so changing the example to an ordinary model such as `gpt-4o-mini` does not make only the augmented arm fail.
- Preserve JSON and YAML chat-message prompts and prepend only the scaffold message, so baseline and augmented runs retain the same original conversation roles. YAML parsing is lazy and documented as requiring `js-yaml` only when copied examples use YAML prompts.
- Honor `config.apiBaseUrl`, `OPENAI_API_BASE_URL`, and `OPENAI_BASE_URL` for the completion request, so proxy and local-endpoint users compare both arms through the intended endpoint.
- Mirror the baseline provider's `apiHost`, per-provider env/config key, `apiKeyEnvar`, `apiKeyRequired`, organization, and custom-header behavior, so routed or unauthenticated OpenAI-compatible comparisons are equivalent between arms.
- Keep JSON-looking task text that is not a chat-message array as one user message instead of treating object contents as message entries.
- Clear ambient OpenAI routing environment variables in the regression setup so endpoint assertions are deterministic outside CI.
- Do not default o-series models to GPT-5's `reasoning_effort: none`, and match ordinary OpenAI chat defaults for `max_tokens` and `temperature` when users change the comparison model.
- Load the custom provider as ESM and use ESM-compatible lazy YAML parsing, so YAML prompts work without a CommonJS `require` failure.
- Keep the copied example standalone by using standard `fetch` rather than importing Promptfoo internals; document the proxy/custom-TLS transport boundary instead of promising unsupported transport parity.
- Apply per-prompt provider configuration overrides like the built-in baseline arm, including ordinary-chat decoding settings.
- Preserve OpenAI refusal and `content_filter` responses as guarded successful outputs rather than turning valid safety responses into provider errors.
- Merge the current `origin/main` before final validation.

## Validation

- `npm ci --prefer-offline --registry=https://registry.npmjs.org/` succeeded, enabling local validation that had previously been blocked by missing worktree dependencies.
- Merged current `origin/main` (`8236e6501`) before publishing repaired head `f0f2300cb`.
- `npm run f`, `npm run l`, and `git diff --check` passed on the current-main merged branch.
- `npx vitest run test/examples/baselineVsEjentumHarnessProvider.test.ts --sequence.shuffle=false` passed (`20` tests).
- `npx vitest run test/examples/baselineVsEjentumHarnessProvider.test.ts --sequence.seed=9418` passed (`20` tests) again after the latest-main merge.
- `npm run tsc` passed.
- Real workflow proof passed on current-main merged head `f0f2300cb`: a local one-shot Ejentum/OpenAI-compatible service plus `OPENAI_API_KEY='' EJENTUM_API_KEY='' npm run local -- eval -c /tmp/pr-9418-e2e.yaml --no-cache --max-concurrency 1 -o /tmp/pr-9418-merged-standalone-results.json` completed with `1` pass, `0` failures, and `0` errors while taking the Ejentum key from provider config and using `apiKeyRequired: false`.
- The captured OpenAI-compatible request used `gpt-4o-mini`, forwarded configured organization/custom headers, omitted Authorization and unsupported `reasoning_effort`/`verbosity`, and sent the baseline-equivalent ordinary-chat defaults `max_tokens: 1024` and `temperature: 0`.
- Full GitHub Actions validation passed on intermediate repaired head `c87ceafaf` (29 successful checks, 2 expected skips); fresh validation will run for final repaired head `f0f2300cb`.

## Affiliation

The contributor maintains `ejentum-mcp`. The example is intended as a measurement template for users who already choose to evaluate an Ejentum scaffold against an unaugmented baseline.


mldangelo:
Hi @ejentum — thanks for this contribution! I pushed one small polish commit (428c7fb) on top of your branch before merging. Two adjustments:

1. **Configurable Ejentum URL in `provider.js`.** The base URL is now resolved as `config.apiUrl` (from the provider block in `promptfooconfig.yaml`) → `EJENTUM_API_URL` env var → the default published endpoint. The Zuplo URL still works exactly as before for anyone running the example as-is; this just makes it easy for readers to point the provider at staging, a self-hosted endpoint, or a different service. The `config → env → default` pattern is also a nice transferable lesson for folks reading the file as a custom-provider template.

2. **README clarification.** Added a short note under a new `Setup` heading that without `EJENTUM_API_KEY` the augmented half of the eval will error but the baseline `openai:gpt-4o-mini` provider still runs to completion — useful for readers who want to preview the baseline scores first. Also added a one-liner framing `provider.js` as a transferable template that can be adapted to any prompt-augmentation service (LangChain, DSPy, a self-hosted retrieval pipeline, etc.) by swapping the step-1 `fetch` call.

No behavior change for the default path; the example still runs the same way. Let me know if you'd like anything tweaked before this lands.

ejentum:
Both adjustments look great, thanks for the polish. The config-env-default pattern is a clean transferable template; happy for this to land as-is.

PR #9420, opened 2026-05-23
fix: invalidate count cache after EvalResult inserts (#9348)

### LINKED ISSUE #9348: Invalidate eval result count caches after result writes

zanni098:
## Summary

Fixes #9348

`getCachedResultsCount()` and `getTotalResultRowCount()` cache their results with a 5-minute TTL, but neither `EvalResult.createFromEvaluateResult()` nor `EvalResult.createManyFromEvaluateResult()` ever called `clearCountCache()` after writing rows. This meant callers could see stale zero (or outdated) counts for up to 5 minutes after rows were inserted.

## Root Cause

The cache module (`evalPerformance.ts`) already exports `clearCountCache(evalId?)` but it was only used manually in tests and the `closeDb` path. The insert methods in `evalResult.ts` had no knowledge of the cache.

## Fix

- Import `clearCountCache` into `src/models/evalResult.ts`
- Call `clearCountCache(evalId)` after the `db.insert().returning()` in `createFromEvaluateResult` (persisted path only)
- Call `clearCountCache(evalId)` after the synchronous transaction in `createManyFromEvaluateResult`

## Tests

Added three regression tests to `test/models/evalPerformance.test.ts`:

1. `getCachedResultsCount` returns a fresh count after `createFromEvaluateResult` without any manual `clearCountCache()` call
2. `getTotalResultRowCount` returns a fresh count after `createFromEvaluateResult`
3. `getCachedResultsCount` returns a fresh count after `createManyFromEvaluateResult`

All existing evalPerformance tests continue to pass.

### DIFF
--- a/src/models/evalResult.ts
+++ b/src/models/evalResult.ts
@@ -4,6 +4,7 @@ import { getDb } from '../database/index';
 import { evalResultsTable } from '../database/tables';
 import { getEnvBool } from '../envars';
 import logger from '../logger';
+import { clearCountCache } from './evalPerformance';
 import { hashPrompt } from '../prompts/utils';
 import { ProviderConfig } from '../providers/shared';
 import {
@@ -386,6 +387,7 @@ export default class EvalResult {
       args.gradingResult = sanitizeGradingResultForDb(args.gradingResult);
       args.metadata = sanitizeMetadataForDb(args.metadata);
       const dbResult = await db.insert(evalResultsTable).values(args).returning();
+      clearCountCache(evalId);
       return new EvalResult({ ...dbResult[0], persisted: true });
     }
     return new EvalResult(args);
@@ -429,6 +431,7 @@ export default class EvalResult {
         returnResults.push(new EvalResult({ ...dbResult, persisted: true }));
       }
     });
+    clearCountCache(evalId);
     return returnResults;
   }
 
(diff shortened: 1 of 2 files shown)


mldangelo:
Thanks for the contribution, @zanni098! This is an exact duplicate of the #9348 fix that landed via #9421 (also yours — bundled with the #9383 fix originally). Closing in favor of the merged version. Appreciate the persistence on this one.

PR #9421, opened 2026-05-23
fix(eval): invalidate result-count cache on insert (#9348)

### LINKED ISSUE #9348: Invalidate eval result count caches after result writes

zanni098:
## Summary

Fixes #9348.

`getCachedResultsCount()` and `getTotalResultRowCount()` cache counts with a 5-minute TTL, but the insert paths never invalidated those caches — callers could see stale zeros until TTL expiry.

- Import `clearCountCache` into `src/models/evalResult.ts`
- Call `clearCountCache(evalId)` after the persisted insert in `createFromEvaluateResult`
- Call `clearCountCache(evalId)` after the transaction in `createManyFromEvaluateResult`

## Test plan

- [x] Regression tests in `test/models/evalPerformance.test.ts` cover both insert paths and confirm `getCachedResultsCount` / `getTotalResultRowCount` refresh without manual cache clearing.

## Notes

This PR was originally bundled with a fix for #9383 (preserve `CallApiFunction` providers in `combineConfigs`); that piece was superseded by #9402 + #9408 which landed first. Slimming this PR down to just the #9348 contribution so it can be reviewed independently.

### DIFF
--- a/src/models/evalResult.ts
+++ b/src/models/evalResult.ts
@@ -4,6 +4,7 @@ import { getDb } from '../database/index';
 import { evalResultsTable } from '../database/tables';
 import { getEnvBool } from '../envars';
 import logger from '../logger';
+import { clearCountCache } from './evalPerformance';
 import { hashPrompt } from '../prompts/utils';
 import { ProviderConfig } from '../providers/shared';
 import {
@@ -386,6 +387,7 @@ export default class EvalResult {
       args.gradingResult = sanitizeGradingResultForDb(args.gradingResult);
       args.metadata = sanitizeMetadataForDb(args.metadata);
       const dbResult = await db.insert(evalResultsTable).values(args).returning();
+      clearCountCache(evalId);
       return new EvalResult({ ...dbResult[0], persisted: true });
     }
     return new EvalResult(args);
@@ -429,6 +431,7 @@ export default class EvalResult {
         returnResults.push(new EvalResult({ ...dbResult, persisted: true }));
       }
     });
+    clearCountCache(evalId);
     return returnResults;
   }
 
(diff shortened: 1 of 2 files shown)


mldangelo:
Thanks for the contribution, @zanni098 — and apologies for the heavy hand on the branch.

This PR originally bundled two fixes: #9383 (`CallApiFunction` dedupe in `combineConfigs`) and #9348 (result-count cache invalidation). Two of us happened to attack #9383 at the same time, so we went with @ppradyoth's earlier PR (#9402) as the base + a hardening follow-up (#9408) for the broader edge cases.

Your **#9348 fix is the real, separate contribution** here that I didn't want to lose — it's a genuine bug not covered by anything else, and the regression tests are solid. So I:

1. Rebased onto the latest main (which now contains #9402 and #9408).
2. Dropped the now-redundant `combineConfigs` commit.
3. Kept your `dfc18bcf8` cache-invalidation commit verbatim, preserving you as author.
4. Retitled the PR and rewrote the body so it reads as a clean #9348 fix.

11/11 tests in `test/models/evalPerformance.test.ts` pass locally. Waiting on CI before merging. Really appreciate you catching #9348 — the cache could have caused confusing UX bugs.

PR #9423, opened 2026-05-23
test(code-scan): cover action fallback for file-only findings

### LINKED ISSUE #9422: code-scan-action fallback PR posting fails for file-only findings with `line: null`
**Describe the bug**

When the code scan server returns valid file-scoped findings and sets `commentsPosted: false`, the GitHub Action fallback posting path can fail to post comments.

`CommentSchema` allows `line: null`, and the SARIF path explicitly supports file-only findings (`file` present, no `line`). But the fallback comment splitter currently treats any finding with a `file` as a review comment. That means a file-only finding is sent to `pulls.createReview()` with `path` but without a valid line location, which GitHub rejects for inline review comments.

Current behavior on `main`:

- `src/types/codeScan.ts` allows `line: null`
- `src/codeScan/util/sarif.ts` intentionally emits file-only findings without a region
- `src/codeScan/util/github.ts` routes `c.file && c.finding` into `lineComments`
- `code-scan-action/src/main.ts` maps those to review comments with `line: undefined`

As a result, a valid server response can lead the fallback posting path to drop comments entirely.

**To Reproduce**

1. Run the code scan GitHub Action on a PR.
2. Have the scan response include a valid file-only finding, for example:

```json
{
  "success": true,
  "commentsPosted": false,
  "review": "Found issues",
  "comments": [
    {
      "file": "src/example.ts",
      "line": null,
      "finding": "File-scoped issue",
      "severity": "high"
    }
  ]
}
```

3. Let the action enter the fallback posting path.
4. Observe that the action attempts to post the finding through `pulls.createReview()` as an inline review comment without a line location.
5. GitHub rejects the review comment payload, and the action logs that comments could not be posted.

**Expected behavior**

File-only findings should not be treated as inline review comments. Only findings with both `file` and a non-null `line` should go to `lineComments`. File-only findings should go through the general comment path instead.

Suggested split:

- line comments: `c.file && c.line != null && c.finding`
- general comments: `(!c.file || c.line == null) && c.finding && c.severity !== CodeScanSeverity.NONE`

**Screenshots**

Not applicable.

**System information:**

- Promptfoo version: current `main` (`1205a2a2be77beb8505731515d0af1ee893cacb0` at the time of filing)

**Additional context**

I traced this through history:

- Original root cause was introduced in `3c3ca7e715a3b6d94ce64696c8d45152f7877c1c` (`feat: code scanner (#6121)`), where the fallback logic first classified all `file` findings as review comments and mapped `line: null` to `undefined`.
- The current helper-based form was introduced in `2a6e38c4e210c5bc6b1358fa5d4c48c82c3bec78` (`fix(code-scan): remove redundant PR comment when no issues found (#6317)`), which preserved the same split in `prepareComments()`.
- There is also a later commit `f6af1ecd0bc5d1311d5991df482bab73e35070c3` (`fix(code-scan-action): harden PR comment handling`) that updates the split to treat `line == null` findings as general comments, but that commit is not present on the current `main` history.

This looks runtime-observable rather than stylistic because `commentsPosted === false` is a supported action path and file-only findings are already treated as valid elsewhere in the codebase.

fallintoplace:
## Summary

- Merge current `main`, which already contains the file-only fallback behavior fix from #9525.
- Preserve a direct GitHub Action regression test for the exact ordinary fallback response: `commentsPosted: false`, a file-scoped finding, and `line: null`.
- Verify that the action posts the finding as a general PR comment, preserves the file path in the comment body, and does not attempt an invalid inline review.
- Drop the superseded duplicate helper and redundant unit coverage from the original branch.

This PR complements #9556: this one protects the action-level ordinary fallback flow, while #9556 hardens lower-level inline-location classification.

Closes #9422.

## Validation

- `npx vitest run test/codeScans/util/github.test.ts test/code-scan-action/main.test.ts` passed (`42` tests).
- `npx vitest run test/code-scan-action` passed (`43` tests).
- `npx vitest run test/codeScans test/code-scan-action --sequence.seed=9423` passed (`192` tests).
- `npm run tsc` passed.
- `npm --prefix code-scan-action run tsc` passed.
- `npm --prefix code-scan-action run build` passed.
- `npm run l` and `git diff --check` passed.
- The full GitHub CI matrix passed on the final head.


mldangelo-oai:
Thank you!

PR #9426, opened 2026-05-23
fix(code-scan): emit structured fork PR skip output

### LINKED ISSUE #9424: code-scan-action can fail JSON parsing when fork PR auth rejection exits successfully without structured output
**Describe the bug**

The GitHub Action can fail with `Failed to parse CLI output as JSON` when a fork PR scan is rejected for lack of maintainer approval.

The action always invokes the scanner with `--json`, then treats `exitCode === 0` as "stdout contains a JSON `ScanResponse`" and immediately parses stdout.

However, the scanner's fork-PR auth rejection path handles this case as a successful skip: it logs an informational message, sets `process.exitCode = 0`, and returns without emitting a structured `ScanResponse` payload.

That combines badly with the CLI structured-output path:

- `code-scan-action/src/main.ts` passes `--json`
- `src/entrypoint.ts` sets `LOG_LEVEL=error` and `PROMPTFOO_LOG_TO_STDERR=true` for structured code-scan output
- `src/codeScan/scanner/index.ts` catches `"Fork PR scanning not authorized"`, logs an info message, sets exit code 0, and returns
- `code-scan-action/src/main.ts` sees `exitCode === 0` and calls `JSON.parse(scanOutput)`

In this path, stdout can be empty, because the info log is suppressed or routed to stderr. The action then fails trying to parse empty stdout as JSON instead of treating the scan as a successful skip.

**To Reproduce**

1. Run the code scan GitHub Action on a fork PR that requires maintainer approval before scanning.
2. Let the scanner hit the `"Fork PR scanning not authorized"` path.
3. The CLI exits successfully (`0`) without emitting a JSON `ScanResponse` to stdout.
4. The action treats the successful exit as structured output and calls `JSON.parse(scanOutput)`.
5. The action fails with `Failed to parse CLI output as JSON`.

**Expected behavior**

The action should treat this fork-auth rejection as a successful skip instead of a JSON parse failure.

Any of these fixes would work:

- emit a structured empty/sentinel `ScanResponse` from the scanner for this expected case when `--json` is active
- return a non-zero sentinel that matches the action's existing special-case branch
- explicitly teach the action to handle an empty successful stdout for this auth-rejection path

**Screenshots**

Not applicable.

**System information:**

- Promptfoo version: current `main` (`1205a2a2be77beb8505731515d0af1ee893cacb0` at the time of filing)

**Additional context**

The current test coverage does not exercise this exact runtime path.

There is an existing action test for fork PR approval behavior, but it mocks a different shape: it writes `"Fork PR scanning not authorized"` to stdout and returns exit code `1`. That covers the action's fallback branch for non-zero exit plus sentinel text, but not the current scanner path that exits `0` without structured JSON output.

This looks like a clean runtime bug rather than a docs issue because it depends on the interaction between:

- structured output mode in the CLI entrypoint
- the scanner's successful early return for fork PR auth rejection
- the action's unconditional `JSON.parse` on successful exit

fallintoplace:
## Summary
- emit a structured successful scan response when fork PR scanning is skipped for maintainer approval in machine-readable modes
- keep the existing text-mode behavior unchanged for interactive CLI users
- preserve legacy text skip handling while the CLI and action publish independently
- avoid publishing empty SARIF for authorization-skipped scans, since a skipped scan is not a clean result
- add regression coverage for the scanner/action structured response, legacy fallback, and SARIF suppression

Closes #9424.

## Audit Fixes Applied
- Structured `skipReason` resolves the successful-empty-stdout JSON parse failure from #9424.
- Authentication-gated skips now leave `sarif-path` unset, preserving the documented Code Scanning contract and preventing an unrun scan from publishing empty results.
- The action continues to recognize legacy `Fork PR scanning not authorized` text output during staged CLI/action rollouts.

## Validation
- `npm run l` (passes with the existing non-blocking cognitive-complexity warning in `src/codeScan/scanner/index.ts`)
- `npm run f` (passes with the same existing warning; documentation formatting unchanged)
- `git diff --check`

## Local Environment Note
- Full local `npm ci --registry=https://registry.npmjs.org/` is blocked in this environment by HTTP 403 responses for existing public dependency tarballs (first Socket.IO family artifacts, then `@anthropic-ai/sdk@0.97.1` after cache repair). GitHub CI on this pushed head is the authoritative full test/type/build validation.


mldangelo-oai:
Thank you!

PR #9429, opened 2026-05-24
test(providers): drop flaky cross-process Anthropic cache hash spawn

mldangelo:
## Summary

- Removes the flaky `should keep cache key hashes stable across fresh processes` test from `test/providers/anthropic/completion.test.ts` that was timing out on Windows + Node 24 CI runners (e.g. failed run in #9421).
- The test spawned two `node --import tsx -e ...` subprocesses with a 10s/spawn budget; tsx startup plus the transitive import graph of `src/providers/anthropic/generic.ts` (Anthropic SDK, Claude Code auth helpers, etc.) routinely exceeds that on Windows and the spawn fails with `ETIMEDOUT`. Per `test/AGENTS.md` we never bump test timeouts — we fix the slow test, and here the right fix is removal because the coverage is redundant.
- Drops the now-unused `spawnSync` import and adds a short comment on the existing `should produce known hex digests for fixed inputs` test explaining that it serves as the cross-process stability check.

## Why is this safe?

Cross-process determinism is already enforced by two existing tests added in the same PR (#9124) that introduced the spawn test:

1. **`should produce known hex digests for fixed inputs`** (line ~168) pins specific SHA-256 digests for the same inputs the spawn test used (`'sk-ant-restart-secret'`, `{ prompt: 'same prompt' }`). Because those expected digests are baked into the test file as string literals, **any process that diverges from determinism must fail this assertion**. By transitivity, two fresh processes that both pass it produce identical output.
2. **`should keep auth cache namespace stable across module reloads`** uses `vi.resetModules()` to catch module-load-time randomness (the original bug class #9124 was guarding against).

The spawn test only added unique value against a very narrow failure mode — non-determinism that depends on environment state present only inside Vitest's process — which neither matches the original bug nor is realistic given the functions are pure (`createHash` / `createHmac` over a fixed string context).

## Test plan

- [x] `npx vitest run test/providers/anthropic/completion.test.ts` — 13/13 pass in ~400ms
- [x] `npm run l && npm run f` — clean
- [ ] CI: confirm the Windows / Node 24.x shard 3/3 job that was previously flaking now goes green

### DIFF
--- a/test/providers/anthropic/completion.test.ts
+++ b/test/providers/anthropic/completion.test.ts
@@ -1,5 +1,3 @@
-import { spawnSync } from 'node:child_process';
-
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { clearCache, disableCache, enableCache, getCache } from '../../../src/cache';
 import logger from '../../../src/logger';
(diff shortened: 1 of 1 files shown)


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/44515a49-2918-421e-a561-01cb2f107a1c?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on c5da98b0ad3ab6147a446cc4046225f8b3da7e3e. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| c5da98b0ad3ab6147a446cc4046225f8b3da7e3e | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/44515a49-2918-421e-a561-01cb2f107a1c?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 24, 2026 6:13AM |

</p>
</details>

PR #9435, opened 2026-05-24
fix(web-viewer): preserve config variable order instead of alphabetizing

### LINKED ISSUE #244: Web view variables ordering
My issue is that I want my variables in the web view to keep the same ordering as in my CSV, however, they are currently being sorted.

I was going to simply workaround this issue by numbering my vars, but I ran into nunjucks errors when trying to do so. 

Any space or special character in a variable name (For example: {{1 firstVar}} or {{1.firstVar}} throws the following error: 
```
Template render error - (unknown path) [Line 1, Column 1268]
expected variable end
```

Not at all familiar with nunjucks, so this may be expected behaviour. I just wanted to point it out as a potential additional thing to fix. 

The main ask here is to maintain ordering of variables from the dataset.

### LINKED ISSUE #1320: Reorder order of variables on the GUI
Currenlty, the variables show up on the website sorted alphanumerically. Can you allow reordering of the variables from the GUI?

zanni098:
## Summary

- Preserve user-defined variable column order in paginated and full-table viewer paths instead of alphabetizing it.
- Seed persisted header order from precomputed configured/transformed variables before concurrent provider calls complete.
- Carry persisted header order in result payloads so database row load order cannot reshuffle columns.
- Append derived columns such as `sessionId` after configured columns, while retaining alphabetical fallback for legacy payloads that lack stored order.
- Add regressions for out-of-order completion, loaded results, legacy backfills, and display-only columns.

Closes #244. Closes #1320.

## Audit Fixes Applied

- The initial change used first-seen result order, which remained nondeterministic for unordered persisted result loads. Full-table conversion now uses the evaluation's stored display order when present.
- The evaluation previously populated stored vars as concurrent rows completed. It now seeds them from `prepareTestVariables()` before execution, making configured/transformed column order stable regardless of provider latency.
- Legacy eval rows without persisted variable order are now backfilled and persisted in sorted order before metadata-only `sessionId` is appended.
- Current payloads preserve the persisted configured prefix and deterministically sort result-only / `transformDisplayVars` columns after that prefix.
- Merged current `origin/main` into the repaired fork branch before final validation.

## Validation

- `npx @biomejs/biome check --write src/evaluator.ts src/models/eval.ts src/types/index.ts src/util/convertEvalResultsToTable.ts test/evaluator/execution.test.ts test/models/eval.test.ts test/util/convertEvalResultsToTable.test.ts` (passes with existing non-blocking complexity warnings in `src/models/eval.ts` and `src/util/convertEvalResultsToTable.ts`)
- `npm run l` (passes with the same existing warnings)
- `npx vitest run test/util/convertEvalResultsToTable.test.ts test/models/eval.test.ts test/evaluator/execution.test.ts --sequence.shuffle=false` (`147` tests passed)
- `npm run tsc`
- Real local eval using `echo` with non-alphabetical configured vars: `eval-X9J-2026-05-26T02:01:54`, `2` passed, `0` failed, `0` errors; exported display order was `["zebra", "apple"]`.
- Full GitHub Actions validation passed on repaired head `eabd8058c` (29 successful checks, 2 expected skips).

## Local Tooling Note

- `npm run f` completes Biome for this TypeScript-only diff, then calls Prettier without any Markdown/YAML file arguments and exits with `No parser and no file path given`; the targeted Biome formatting command above passed without rewrites.


mldangelo-oai:
Thank you!

PR #5527, opened 2025-09-09
revert: remove red team limits functionality

mldangelo:
## Summary

This PR reverts the red team limits functionality introduced in:
- #5230 - feat: Add open source red team limits
- #5516 - docs(site): redteam limits documentation  

## Changes

- Reverts commit a2dc5c107 (Limits documentation #5516)
- Reverts commit 27add302c (Add open source red team limits #5230)
- Removes probe limits and related UI components
- Removes database migrations for limits tracking
- Removes documentation for limits feature
- Preserves telemetry constants required for existing functionality

## Test Plan

- [x] TypeScript compilation passes
- [x] Redteam tests pass
- [x] Generate command tests pass
- [x] Lint and format checks pass

PR #5230, opened 2025-08-12
feat: Add open source red team limits

sklein12:
Red teaming requires our inference to generate probes and grading. We need to limit the red teaming monthly to allow people to use red teaming but for a reasonable limit.

### Before running a red team
image

### UI
image

PR #9462, opened 2026-05-26
perf(webui): speed up evaluation selection

mldangelo-oai:
## Summary

- Avoid loading and deserializing large eval `config` JSON in the ordinary `/api/results` summary path, and use the indexed `is_redteam` field for type filtering.
- Repair existing `is_redteam` values in a migration and keep that materialized field accurate across create, save/update, copy, legacy write, and import paths.
- Scope the compare selector request to its known dataset, add deterministic tie ordering, and fix description search when the newest row has no description.

## Trace And Rationale

The **Select Evaluation / Choose an evaluation to view** dialog is populated by `GET /api/results`, which resolves through `getEvalSummaries()`. On a large local database snapshot, the ordinary selector query was selecting/deserializing embedded config JSON even though it only needs summary columns. It also used JSON inspection for redteam filtering despite an existing indexed `is_redteam` column.

The UI uses client virtualization rather than pagination, and its search/export/delete behavior expects the complete supplied row set. This change preserves that behavior for the normal selector while reducing the response query cost. The comparison selector already has an exact dataset constraint, so it now sends that existing filter to the API instead of downloading unrelated histories.

## Historical Context

- `is_redteam` and its original backfill migration landed in [#5230](https://github.com/promptfoo/promptfoo/pull/5230) on September 7, 2025 as part of open-source red team limits.
- [#5527](https://github.com/promptfoo/promptfoo/pull/5527) reverted that limits feature on September 9, 2025 while intentionally retaining the database column and migration; the original backfill in `0019_new_clint_barton.sql` used `json_extract(config, '$.redteam') IS NOT NULL`.
- The runtime summary predicate that the column was later compared against used `json_type(config, '$.redteam') IS NOT NULL`. These two predicates diverge for `{"redteam": null}`: `json_extract` returns SQL NULL for both missing keys and JSON null values, while `json_type` returns the string `'null'` for JSON null and SQL NULL only when the key is missing. Only `json_type` matches JS `config.redteam !== undefined`.
- [#7966](https://github.com/promptfoo/promptfoo/pull/7966) began writing `isRedteam` again for `Eval.create()` on March 5, 2026, but config updates, legacy imports, and legacy writes could still leave stored values stale.
- This PR makes the materialized flag authoritative for `/api/results` summary filtering, so `0024_repair_eval_redteam_flags` repairs pre-upgrade discrepancies before indexed reads are used. The repair migration uses `json_type` (and a comment explains why) so it aligns with the runtime write predicate.

## Redteam Classification Semantics

The runtime write predicate (`Eval.create`, `Eval.save`, `Eval.copy`, `writeResultsToDatabase`, and the V2 import path) is now uniformly `config.redteam !== undefined`. This is intentionally broader than the previous `Boolean(config.redteam)` coercion:

- `{redteam: <object>}` → classified as redteam (no change).
- `{redteam: null}` → classified as redteam (previously stored `false`). This matches the `json_type` predicate the runtime summary filter has used since the column was introduced, and is also what the repair migration enforces.
- `{redteam: false | 0 | ""}` → also classified as redteam under the new predicate. These are edge cases that the typed `RedteamFileConfig | undefined` schema does not produce in practice, but they fall out of the `!== undefined` definition.
- Missing key / `redteam: undefined` → classified as not redteam.

`getStandaloneEvals` (the `/api/history` query) is also moved off raw `json_extract` and onto the indexed `is_redteam` column so every read path uses the same classification.

## Correctness Found During QA

- Persisted config updates could leave `is_redteam` stale when an eval changed between regular and redteam configuration; `Eval.save()` now updates the flag.
- Existing data may predate reliable flag writes; migration `0024_repair_eval_redteam_flags` backfills the materialized field using the same redteam-key semantics as the summary API and updates only mismatched rows. An SQL comment now documents why `json_type` and not `json_extract` is used so future repair migrations don't regress the semantics.
- Indexed filtering exposed nondeterministic ordering for equal timestamps; summaries now sort by `createdAt DESC, id DESC`.
- The dialog's global description search could silently exclude the Description column when the first row had a null description; its accessor now searches the visible description-or-label value. A code comment near the `accessorFn` explains the TanStack first-row-sampling footgun.
- `getStandaloneEvals` (powering `/api/history`) was the last query path still inspecting raw config JSON for redteam classification. Switched to the indexed column for consistency with the rest of the codebase.

## Benchmark

Measured against current `origin/main` on the same migrated copy of a local database snapshot:

- 27,495 evaluations
- 343.4 MB of stored config JSON
- five preexisting redteam-flag mismatches repaired (`430` flagged rows before, `435` after)
- five alternating timed reads per endpoint after warmup; normalized response content matched and patched ordering was stable

| Endpoint | Rows | `main` median | Patched median | Speedup |
| --- | ---: | ---: | ---: | ---: |
| `/api/results` | 27,495 | 1350.2 ms | 232.8 ms | 5.80x |
| `/api/results?type=eval` | 27,060 | 1292.8 ms | 194.2 ms | 6.66x |
| `/api/results?type=redteam` | 435 | 543.9 ms | 13.1 ms | 41.68x |
| `/api/results?type=redteam&includeProviders=true` | 435 | 540.6 ms | 175.3 ms | 3.08x |
| `/api/results?datasetId=...` | 1 | 3.2 ms | 7.0 ms | Near-zero either way |

The initial unconditional one-time backfill migration took 10.10 seconds on the 2.6 GB snapshot; the source database was not modified. After review, the migration was narrowed to update only rows whose stored flag differs from the JSON-derived classification, retaining the full correctness scan while avoiding unnecessary writes.

In a disposable WAL-mode synthetic replay at the measured row/config scale (27,495 rows, approximately 343 MB of config JSON, five stale flags), the narrowed statement reduced affected rows from 27,495 to 5 and median update time from 306.5 ms to 270.5 ms. The migration must still parse each row's config to locate stale flags; this optimization reduces writes and index churn rather than eliminating scan time.

The one-row dataset-filtered request is included for correctness and comparison-dialog scoping coverage, not as a meaningful latency benchmark.

## Follow-Up Improvements (post initial review)

- `getStandaloneEvals` (`/api/history`) now reads `is_redteam` from the indexed column instead of computing `json_extract(config, '$.redteam') IS NOT NULL` at query time. This brings the last read path into alignment with the rest of the codebase and removes the predicate divergence on `{redteam: null}`.
- `clearStandaloneEvalCache()` is exported from `src/util/database.ts` so tests that mutate eval state can observe writes without waiting on the 2-hour LRU TTL. The cache itself was not invalidating on writes before; that is preserved as existing behavior.
- Migration `0024_repair_eval_redteam_flags.sql` now has an SQL comment explaining the deliberate use of `json_type` (and why `json_extract` would diverge from the JS runtime predicate).
- `EvalsTable` Description column has an explanatory code comment about why `accessorFn` is required rather than `accessorKey` (TanStack Table infers column type from the first row).
- New regression tests:
  - `test/util/databaseStandalone.test.ts`: covers `getStandaloneEvals` reading from the materialized column, transitions across `updateResult`, and the `{redteam: null}` edge case. Also locks down the json_type vs json_extract divergence and the migration's WHERE-clause idempotency directly in SQL.
  - `test/models/eval.test.ts`: parameterized classification test for `{redteam: {}}`, `{redteam: null}`, and missing-key configs across both the stored flag and `getEvalSummaries` filtering.
  - `src/app/src/pages/evals/components/EvalsTable.test.tsx`: covers the `focusedDatasetId={null,undefined}` fallback to `/results` (and client-side narrowing), plus `encodeURIComponent` for dataset ids containing spaces and ampersands.

## QA

- `npm run f` and `npm run l` (pass; reports one preexisting nonblocking cognitive-complexity warning in `src/models/eval.ts:833`)
- `npm run tsc`
- `npm --prefix src/app run tsc -- --noEmit`
- `npm run build`
- Backend regression matrix: 103 files / 2,584 tests passed across `test/models`, `test/util`, `test/commands/import.test.ts`, and `test/server`, covering model writes, import, migration, server eval routes and schemas, redteam probe accounting, and the new `getStandaloneEvals` and migration semantics tests.
- Frontend regression matrix: 75 files / 1,559 tests passed under `src/pages/evals` and `src/pages/eval`, covering the selector, results view, report consumers, search/filter behavior, sorting, virtualized table support, and the new dataset-fallback / URL-encoding tests.
- `npm --prefix src/app run test -- --run` (full app suite): 274 files / 4,549 frontend tests passed.
- Migrated a copy of the large local database and confirmed redteam mismatches went from `5` to `0`.
- Follow-up migration optimization validation: the final SQL updated stale regular/redteam/JSON-null/malformed-config fixtures while leaving correctly materialized fixtures unchanged.
- Live local UI QA confirmed selector load, description search, Created sorting, and same-dataset comparison requests (`/api/results?datasetId=...`).

### DIFF
--- a/drizzle/0024_repair_eval_redteam_flags.sql
+++ b/drizzle/0024_repair_eval_redteam_flags.sql
@@ -1,0 +1,4 @@
+UPDATE `evals` SET `is_redteam` = CASE
+  WHEN json_valid(`config`) AND json_type(`config`, '$.redteam') IS NOT NULL THEN 1
+  ELSE 0
+END;
--- a/drizzle/0024_repair_eval_redteam_flags.sql
+++ b/drizzle/0024_repair_eval_redteam_flags.sql
@@ -1,3 +1,16 @@
+-- Repair stale `is_redteam` values to match the runtime classification used by
+-- the write paths in src/models/eval.ts (`config.redteam !== undefined`).
+--
+-- `json_type` is used here (not `json_extract`) on purpose:
+--   * json_extract(config, '$.redteam') returns SQL NULL both when the key is
+--     missing and when its value is JSON null.
+--   * json_type(config, '$.redteam')    returns SQL NULL only when the key is
+--     missing; it returns 'null' (string) when the value is JSON null.
+-- Only `json_type` matches `config.redteam !== undefined` in JS.
+--
+-- The WHERE clause narrows the write to rows whose stored flag actually
+-- disagrees with the derived value, so re-runs and already-correct databases
+-- skip the bulk update.
 UPDATE `evals` SET `is_redteam` = CASE
   WHEN json_valid(`config`) AND json_type(`config`, '$.redteam') IS NOT NULL THEN 1
   ELSE 0
--- a/drizzle/0024_repair_eval_redteam_flags.sql
+++ b/drizzle/0024_repair_eval_redteam_flags.sql
@@ -1,16 +1,9 @@
--- Repair stale `is_redteam` values to match the runtime classification used by
--- the write paths in src/models/eval.ts (`config.redteam !== undefined`).
+-- Backfill `is_redteam` to match the runtime predicate `config.redteam !== undefined`
+-- used by writes in src/models/eval.ts.
 --
--- `json_type` is used here (not `json_extract`) on purpose:
---   * json_extract(config, '$.redteam') returns SQL NULL both when the key is
---     missing and when its value is JSON null.
---   * json_type(config, '$.redteam')    returns SQL NULL only when the key is
---     missing; it returns 'null' (string) when the value is JSON null.
--- Only `json_type` matches `config.redteam !== undefined` in JS.
---
--- The WHERE clause narrows the write to rows whose stored flag actually
--- disagrees with the derived value, so re-runs and already-correct databases
--- skip the bulk update.
+-- Uses `json_type` (not `json_extract`) so that `{"redteam": null}` is classified as
+-- a redteam: `json_extract` returns SQL NULL for both missing keys and JSON null,
+-- but `json_type` only returns NULL when the key is missing.
 UPDATE `evals` SET `is_redteam` = CASE
   WHEN json_valid(`config`) AND json_type(`config`, '$.redteam') IS NOT NULL THEN 1
   ELSE 0
--- a/drizzle/0024_repair_eval_redteam_flags.sql
+++ b/drizzle/0024_repair_eval_redteam_flags.sql
@@ -1,4 +1,8 @@
 UPDATE `evals` SET `is_redteam` = CASE
   WHEN json_valid(`config`) AND json_type(`config`, '$.redteam') IS NOT NULL THEN 1
   ELSE 0
+END
+WHERE `is_redteam` != CASE
+  WHEN json_valid(`config`) AND json_type(`config`, '$.redteam') IS NOT NULL THEN 1
+  ELSE 0
 END;
--- a/drizzle/meta/0024_snapshot.json
+++ b/drizzle/meta/0024_snapshot.json
@@ -1,0 +1,1496 @@
+{
+  "id": "00fa2978-039b-4dbd-8e05-d962a981755f",
+  "prevId": "516dc2b9-fcef-43c1-88dc-543e422e3f53",
+  "version": "6",
+  "dialect": "sqlite",
+  "tables": {
+    "blob_assets": {
+      "name": "blob_assets",
+      "columns": {
+        "hash": {
+          "name": "hash",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "size_bytes": {
+          "name": "size_bytes",
+          "type": "integer",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "mime_type": {
+          "name": "mime_type",
+          "type": "text",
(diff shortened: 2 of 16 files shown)


ianw-oai (approved):

ghost:
#### Skipped due to new commit on branch

<a href="https://app.usetusk.ai/app/testing-commit-check/e4701599-042f-4c97-b329-b8d829b451b4?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on 1829cdd31efd3d9955809e5de1c7f79266113336. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 1829cdd31efd3d9955809e5de1c7f79266113336 | Skipped due to new commit on branch · [Output](https://app.usetusk.ai/app/testing-commit-check/e4701599-042f-4c97-b329-b8d829b451b4?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 26, 2026 10:54PM |

</p>
</details>

PR #9463, opened 2026-05-26
feat(assertions): assert parallel/same-turn tool calls in trajectory:tool-sequence

### LINKED ISSUE #9455: trajectory:tool-sequence fails to distinguish between parallel and sequential tool calls due to span flattening
When evaluating tool sequences using `type: trajectory:tool-sequence` (even with `mode: exact`), the native grader flattens all OpenTelemetry tool spans into a 1-dimensional array based on start timestamps.

Because of this flattening, the grader loses the context of LLM "turns." It treats both of these scenarios as passing `[search_orders, search_orders]`:

1. **Parallel:** The agent calls `search_orders` twice simultaneously in a single generation step.
2. **Sequential:** The agent calls `search_orders`, waits for the LLM to process the result in a new turn, and then calls `search_orders` again.

**Why this matters:**
For developers trying to optimize latency and token usage, parallel tool calling is a strict requirement. Currently, there is no native way to assert that an agent correctly batched tools in parallel rather than wasting time executing them sequentially.

**Proposed Solution:**
The trajectory grader should be aware of LLM turns/generations, perhaps supporting a nested array structure for `mode: exact` (e.g., `steps: [[search_orders, search_orders]]` to assert parallel execution).

RitwijParmar:
> **Stacked on #9475.** This PR targets `feat/claude-agent-sdk-turn-spans` and consumes the
> `gen_ai.turn.index` attribute that PR emits. Once #9475 merges, the base returns to `main`
> and this diff collapses to the three trajectory files (GitHub's authoritative compare
> already shows merge-base = #9475 tip, 3 files). Nested-array idea and authorship by
> @RitwijParmar (commit co-author).

## Summary

Adds nested step groups to `trajectory:tool-sequence` so you can assert _which_ tools share
_each_ turn — i.e. that the agent batched calls into one model round (parallel) rather than
across rounds (sequential).

```yaml
  value:
    mode: exact
    steps:
      - [search_orders, search_inventory] # one turn (parallel)
      - compose_reply # a later, single-tool turn
```

## How turns are determined

Tools are grouped by the **`gen_ai.turn.index`** attribute that the turn-marker providers in
#9475 tag onto each tool span. Same index = same model round = one turn. A tool without the
attribute can't be proven to share a turn, so it stands alone.

This replaces the earlier `parentSpanId` + span-overlap inference, which produced false
positives — `parentSpanId` often points at a workflow wrapper, not the model generation
(see the discussion on this PR). Consuming the explicit turn marker is both reliable and
simpler: the whole overlap heuristic and the `parentSpanId` plumbing are gone.

## Behavior

- **Flat** sequences (`steps: [a, b]`) are unchanged and turn-agnostic — nesting is opt-in.
- **Nested** entries require the tools to share a turn; within-turn order doesn't matter.
- Works in `exact` and `in_order` modes.
- Complements #9475's `trace-span-count pattern: 'gen_ai.turn *' max: 1` (count-based "all in
  one turn") with a shape-based "which tools in which turn" check. The list-of-lists shape
  also matches the BFCL benchmark data format.

## Test plan

- Reworked the `parallel turns` suite to drive `gen_ai.turn.index` spans: parallel pass /
  sequential fail, flat back-compat, no-index → separate turns, within-turn
  order-insensitivity, `in_order` subsequence + ordering, mixed turns, three-tool turn,
  pattern matching (backtracking), string-valued attribute, inverse, and malformed-entry /
  empty-group errors. **111 trajectory tests pass; `tsc` clean.**
- **End-to-end through the real OTLP pipeline:** a provider emitting two tool spans tagged
  `gen_ai.turn.index: 1` passes `[[search_orders, search_orders]]`; tagging them `1` and `2`
  makes the nested assertion fail while the flat sequence still passes (back-compat).

Closes #9455.


mldangelo:
After deeper QA against real OpenTelemetry traces, the parent-span-grouping approach in this PR has a fundamental reliability problem: many instrumentations set tool spans' `parentSpanId` to a workflow wrapper (e.g., a `document_retrieval` span) rather than to the chat-generation span. I reproduced this with the bundled `examples/integration-opentelemetry/javascript` agent — 3 sequential `search_corpus` calls all share the same `parentSpanId`, so the nested-array assertion would incorrectly report them as batched. Even with the 3-span overlap bug fixed, the heuristic produces false positives on common agent structures.

PR #9475 takes a different approach: instead of inferring turn structure from existing spans, the agent SDK providers themselves emit explicit `gen_ai.turn N` marker spans at natural turn boundaries (SDK stream `assistant` messages, Codex `turn.started`/`turn.completed` events, etc.). Each tool span is tagged with `gen_ai.turn.index`. Users compose `trajectory:tool-sequence` with `trace-span-count pattern: 'gen_ai.turn *' max: 1` to assert batching — same expressivity as the nested-array shape, but the underlying signal is reliable.

The nested-array work here could still be useful as a v2 if we wanted a single dedicated assertion type for batching shapes (the BFCL benchmark uses the same list-of-lists data shape), but it should consume the `gen_ai.turn.index` attribute rather than infer turns from `parentSpanId`.

mldangelo:
Hi @RitwijParmar — thanks for this, and for surfacing #9455, which is a genuinely useful gap to close. I did a deep review + QA pass and then reworked the implementation on top of `main`; your original commit and authorship are preserved and the new commit credits you as co-author.

What I kept from your PR: the nested-array API (`steps: [[a, b]]`) and using the span hierarchy (`parentSpanId`) as the turn signal — both good calls.

What I changed and why:

1. **Backward compatibility.** The original made flat `exact` sequences turn-aware too, so an existing config like `steps: [a, b]` would start *failing* whenever the agent batched the calls or their spans overlapped (common with OpenAI Agents SDK / Vercel AI SDK nesting). I made turn-awareness fully opt-in: flat sequences behave exactly as before, and only nested groups opt into turn matching.

2. **Grouping robustness.** Grouping by consecutive same-parent runs had two failure modes I confirmed with tests: single-root traces (every tool parented to one agent span) collapsed all sequential calls into one "parallel" turn, and turns whose spans interleave in time got mis-split. It now buckets by parent span and splits each bucket by time overlap, which handles both.

3. **Order-insensitive within a turn** (parallel calls have no inherent order), plus docs and a clearer error for a malformed tool inside a group.

I validated it end-to-end through the real OTLP tracing pipeline (a provider emitting two concurrent tool spans under one parent passes `[[search_orders, search_orders]]`; the same assertion fails when the calls happen in separate turns; the flat sequence still passes). 109 trajectory tests pass and CI is running.

Would love your review — happy to adjust anything. Thanks again!

mldangelo:
Follow-up: ran a multi-agent PR review + a simplification pass and pushed two commits.

**Review-driven hardening:**
- Point-in-time / same-start parallel calls (zero-duration spans) now group correctly — the overlap check uses an at-or-before boundary, so a tracer that records tool calls as instants still detects parallelism.
- Turn buckets are keyed directly by `parentSpanId` (dropped the string sentinel) — no collision risk, and empty-string parents are handled.
- Malformed step entries (e.g. `null`, a doubly-nested array) now throw a clear, located error instead of a raw TypeError.
- Deterministic turn ordering when turns share a start time.
- Docs note that in a nested sequence every entry describes a full turn.

**Simplification:** extracted `formatMatcherLabel`/`bracketTurnLabels` and reused `formatStepList`, removing duplicated label/bracket formatting (no behavior change).

114 trajectory tests pass; tsc clean.

mldangelo:
Update: reworked onto @mldangelo's #9475 per the discussion above. Instead of inferring turns from `parentSpanId` + span overlap (which false-positived — `parentSpanId` often points at a workflow wrapper, not the model generation), it now groups tools by the explicit **`gen_ai.turn.index`** attribute #9475 emits. That's the reliable signal, and it let me delete the whole overlap heuristic and the `parentSpanId` plumbing — simpler and correct.

Your nested-array API and authorship are preserved (commit co-author). The PR is now **stacked on #9475** (base `feat/claude-agent-sdk-turn-spans`); once that merges the base returns to `main` and this collapses to the three trajectory files. Validated end-to-end through the OTLP pipeline (two tool spans tagged turn 1 pass `[[search_orders, search_orders]]`; tagging them 1 and 2 fails the nested assertion but the flat sequence still passes). Thanks again!

mldangelo:
Re-synced onto the latest #9475 and ran a full multi-agent review (code / tests / comments / silent-failure / type-design) on the reworked turn-index version. Pushed fixes for the substantive findings:

- **No-turn-marker footgun:** a trace whose provider emits no `gen_ai.turn.index` now fails a nested-group assertion with a clear reason ("No tool spans carried a gen_ai.turn.index attribute…") instead of a misleading "not parallel" one.
- **Stricter index parsing:** `getToolTurnIndex` now accepts only non-negative integers (and their string form), rejecting floats / negatives / hex / scientific / blank values so a malformed attribute degrades to "no index" rather than grouping by accident.
- **`mode` validation:** an unknown `mode` now throws instead of silently running `in_order`.
- **Docs/comments:** state the *consecutive-run* grouping accurately and add the per-run / subagent caveat (turn indices are per model run; subagent tool calls appear as their own turns — use the `trace-span-count` form for those).
- **+9 regression tests:** non-consecutive repeated index (locks consecutive grouping vs a group-by refactor), mixed indexed/non-indexed, index 0, out-of-domain index, no-marker reason, true backtracking bijection (pass + no-bijection fail), empty trace, `in_order` inverse, invalid mode.

Deliberately kept consecutive grouping rather than global group-by-index: turn indices are scoped to a model run, so global grouping would merge same-index tools from different runs (a false *positive* on a correctness assertion) — the consecutive approach fails safe. 120 trajectory tests pass; tsc + prettier clean.

Issue #9458, opened 2026-05-26
Add an exclude or optional list to trajectory:tool-args-match to prevent hallucinated arguments from AI.


sipa-echo-ngbm:
**Is your feature request related to a problem? Please describe.**
I am evaluating an LLM agent's tool calls and am constrained by the current limitations of `type: trajectory:tool-args-match`. I am forced to choose between two extremes:

1. **`mode: exact` is too brittle:** If a tool has safe API defaults (e.g., `page=1`, `page_size=5`), the LLM might optionally explicitly generate them or omit them. `exact` will fail if the LLM includes a valid default that wasn't hardcoded in the test, or vice versa.
2. **`mode: partial` allows hallucinations:** While `partial` successfully ignores the pagination defaults, it also silently ignores severe hallucinations. If the test expects `{"status": "Q"}` and the LLM hallucinates `{"status": "Q", "delete_database": true}`, `mode: partial` incorrectly marks this as a pass.

**Describe the solution you'd like**
I would like the exclude (or ignore) parameter to support both simple strings (to ignore a key entirely) and objects/dictionaries (to ignore a key only when it matches a specific value).

This would allow the grader to:
1. Strip out "safe" default values (like page: 1) before performing the exact match.
2. Still trigger a failure if the agent provides a non-default value (like page: 2) that wasn't expected.
3. Catch hallucinations where the agent uses a parameter that isn't even in the exclude list.

**Describe alternatives you've considered**

1. **Hardcoding defaults in YAML:** I tried explicitly adding `page: 1` to my expected `exact` arguments. However, if the LLM decides to be efficient and omits the parameter (knowing the API will default it to 1 anyway), the test fails.
2. **Using custom Python graders / Trace Scrubbing:** I currently have to intercept the OpenTelemetry spans in my provider and manually delete the `page` keys from the JSON before Promptfoo reads them. This works, but it is a lot of friction and custom code for a very common use case.

**Additional context**
Here is an example of what the implementation in `promptfooconfig.yaml` could look like:

```yaml
assert:
  - type: trajectory:tool-args-match
    metric: tool_args
    value:
      name: orders
      args:
        status: Q
        mode: exact
      exclude:
        - page: 1
        - page_size: 5
```
With this, `{"status": "Q", "page": 1}` passes safely, while `{"status": "Q", "hallucinated_param": true}`, and `{"status": "Q", "page": 2}` strictly fails.

PR #9466, opened 2026-05-26
feat(assertions): add ignore option to trajectory:tool-args-match

Rul1an:
> **Maintainer note:** Rebased onto `main` and reshaped after the `defaults`
> option landed in #9471. The original `exclude` field is superseded; see the
> comment below for the full story. Original work and credit by @Rul1an (carried
> as a commit co-author).

## Summary

Adds an `ignore` option to `trajectory:tool-args-match` that drops a tool-call
argument **by name, regardless of value** — for volatile values such as
`request_id` or `idempotency_key` that change on every call.

It complements the `defaults` option that shipped in #9471:

| Option     | Behavior                                                        |
| ---------- | --------------------------------------------------------------- |
| `defaults` | tolerate a key only when its value equals the declared default  |
| `ignore`   | remove a key from the comparison entirely, whatever its value   |

Named keys are stripped from **both** the observed and expected payloads before
matching, so it doesn't matter whether the agent emits or omits the argument.
`ignore` composes with `defaults`, `exact`/`partial` modes, and inverse
(`not-`), and a matched call's reason now reports `Ignored argument(s): …`.

```yaml
  value:
    name: orders
    mode: exact
    args: { status: Q }
    ignore: [request_id, idempotency_key]
```

This supersedes the original `exclude` approach: the value-conditional half is
now covered by `defaults`, so this PR focuses on the ignore-by-name half that
`defaults` cannot express — the use case and the strict validation from the
original work.

## Test plan

- New `ignore` unit tests in `test/assertions/trajectory.test.ts`: exact/partial
  modes, bare-string + array forms, hallucinated extra still fails, `__proto__`
  safety, composition with `defaults` + reason suffix, inverse, top-level-array
  no-op, and malformed-entry rejection.
- `npx vitest run test/assertions/trajectory.test.ts` → all pass.
- Docs: `site/docs/configuration/expected-outputs/deterministic.md`.

Refs #9458.


mldangelo:
Hi @Rul1an — thank you for this, and apologies for the churn. Issue #9458 drew three independent solutions within about a day (yours, #9484, and #9471). To avoid shipping three spellings of one feature, we converged on a two-field design:

- **`defaults`** (a value map) — "tolerate this key when it equals X." This landed in #9471.
- **`ignore`** (a name list) — "drop this key whatever its value." This is exactly the `request_id` / `idempotency_key` case you built for, plus your strict validation.

Using maintainer edit, I rebased your branch onto `main` and reshaped it into that `ignore` field, since `defaults` already covers the value-conditional half. Your original commit and authorship are preserved, and the new feature commit credits you as co-author. Tests and docs are included and green locally.

Would you take a look and let me know if the behavior matches what you intended? Happy to adjust anything — and thank you again, the validation design carried straight over.

PR #9469, opened 2026-05-26
test(providers): cover Windows-style file:// callback refs

### LINKED ISSUE #9467: Windows file://C:/...:functionName callback parsing breaks in provider loaders
## Summary

Named-export callbacks using Windows `file://C:/...:functionName` references are parsed incorrectly in several provider loaders.

The repo already has a correct helper for this case in `src/util/functions/loadFunction.ts` (`parseFileUrl()` uses the last colon, so it preserves the drive letter and still extracts the function name), but multiple provider-specific loaders still reimplement parsing with `split(':')`.

## Example

This should resolve:

```yaml
functionToolCallbacks:
  my_tool: file://C:/path/to/callbacks.js:handler
```

## Actual behavior

On Windows, loaders that do:

```ts
const splits = filePath.split(':');
if (splits[0] && isJavascriptFile(splits[0])) {
  [filePath, functionName] = splits;
}
```

see `splits[0] === 'C'`, so the `:handler` suffix is never parsed correctly.

That can lead to named exports not being loaded, unexpected fallback behavior, or malformed callback errors depending on the provider.

## Expected behavior

`file://C:/path/to/callbacks.js:handler` should parse to:

- `filePath = 'C:/path/to/callbacks.js'`
- `functionName = 'handler'`

## Existing repo evidence

The repo already has the correct shared parser and test expectation:

- `src/util/functions/loadFunction.ts`
- `test/util/functions/loadFunction.test.ts`

That test already covers:

```ts
parseFileUrl('file://C:/path/to/file.js:functionName')
```

## Affected loaders

I found the naive `split(':')` pattern in at least:

- `src/providers/openai/chat.ts`
- `src/providers/openai/assistant.ts`
- `src/providers/bedrock/converse.ts`
- `src/providers/google/base.ts`
- `src/providers/google/live.ts`
- `src/providers/azure/foundry-agent.ts`
- `src/providers/functionCallbackUtils.ts`

## Suggested fix

Reuse the existing shared parser instead of duplicating callback path parsing logic, e.g.:

```ts
const { filePath, functionName } = parseFileUrl(fileRef);
```

and add provider-level regression tests for Windows-style `file://C:/...:functionName` callback references.

## Why this matters

This is a real correctness bug for Windows users, and the fix should be low risk because the repository already has the intended parsing behavior codified in a shared helper and tests.

fallintoplace:
## Summary

Adds provider-level regression coverage for Windows-style `file://C:/...:functionName` callback references, locking in the callback-parser hardening that shipped via #9479.

> **Note:** The production fix — replacing the providers' duplicated inline `split(':')` parsing with the shared `parseFileUrl()`, plus the win32 drive-letter normalizer — already landed on `main` in **#9479** (a near-duplicate of this PR). After syncing this branch with current `main`, the remaining **net change here is test-only**: two provider-level regression tests that #9479 did not include. All `src/` changes on this branch are now byte-identical to `main`.

## What changed (net, vs `main`)

- `test/providers/openai/chat.test.ts` — verifies `OpenAiChatCompletionProvider` routes `file://C:/test/callbacks.js:testFunction` through `parseFileUrl()`, importing `…/C:/test/callbacks.js` with named export `testFunction` (instead of the mangled `…/C:/test/callbacks.js:testFunction` path with no named export that the old code produced).
- `test/providers/google/live.test.ts` — the same coverage for `GoogleLiveProvider`.

## Why these are real regression tests

Against the pre-fix provider logic (inline `split(':')` guarded by `isJavascriptFile(splits[0])`), `file://C:/test/callbacks.js:testFunction` split into `["C", "/test/callbacks.js", "testFunction"]`; since `isJavascriptFile("C")` is `false`, the `:functionName` suffix was never extracted and the loader imported `C:/test/callbacks.js:testFunction` with no named export. Both new tests assert the corrected behavior and were confirmed to fail against that pre-fix code (verified locally by temporarily reverting `loadExternalFunction`).

## Root cause

Named-export callback refs like `file://C:/path/to/callbacks.js:handler` were split on every colon, leaving the first token as `C`, so the loader never parsed the `:handler` suffix on Windows.

## Validation

- `npx vitest run test/providers/google/live.test.ts test/providers/openai/chat.test.ts test/util/functions/loadFunction.test.ts test/providers/functionCallbackUtils.test.ts test/providers/google/base.test.ts` (231 passed)
- `npm run tsc`
- `npm run l`

Fixes #9467.


mldangelo-oai (approved):
Thanks @fallintoplace for tracking down the Windows callback reference issue and getting these provider loaders onto the shared parser. With coverage now in place for canonical Windows file:///C:/ refs and colon-bearing POSIX callback paths, the full CI matrix is green, including all Windows shards. Appreciate the focused contribution!

mldangelo-oai (approved):
Re-approving after the required sync with current main; the reviewed callback fix is unchanged and the Windows/colon-path regression coverage remains in place.

PR #9139, opened 2026-05-07
feat(quiverai): add Arrow 1.1 models, vectorize endpoint, and GPT Image-2 pipeline

mldangelo:
## Summary

Comprehensive audit of the QuiverAI provider against the live API as of May 2026.

- **New default model.** `arrow-preview` → `arrow-1.1` to match QuiverAI's released lineup (Arrow 1.0, 1.1, 1.1 Max). Updates docs, example, and the providers index entry to advertise both routes.
- **Vectorize endpoint.** Adds `quiverai:vectorize:<model>` routing for the `POST /v1/svgs/vectorizations` endpoint. The provider parses the prompt as image input (URL, `data:` URL, JSON `{url}`/`{base64}`, or raw base64) and supports the new `auto_crop` and `target_size` parameters. `config.image` overrides the prompt when needed.
- **Streaming spec parity.** The SSE parser now handles the documented `generating` / `reasoning` / `draft` / `content` event types. For `n > 1`, per-output SVGs are aggregated by stable `id` and ordered by `index` before joining with `\n\n`.
- **Credits-based billing.** Top-level `credits` (non-streaming) and per-output `content` event credits (streaming) are summed and surfaced on `result.metadata.credits` alongside `metadata.responseId`. The deprecated `usage` token block continues to flow into `tokenUsage` for back-compat even though the API now zeros it.
- **Smarter retries.** `weekly_limit_exceeded` 429s are surfaced immediately (cannot recover within a request) while regular `rate_limit_exceeded` continues to retry with `Retry-After` or exponential backoff.
- **Reference shorthand.** `references` accepts URL strings in addition to `{ url }` / `{ base64 }`, normalized server-side.

## Pipeline example

The example folder gains a chained **GPT Image-2 → QuiverAI vectorize** pipeline implemented as a custom JS provider, plus a standalone vectorize config:

- `examples/provider-quiverai/promptfooconfig.yaml` — generation comparison across Arrow 1.1, Arrow 1.1 Max, and a flat-style variant.
- `examples/provider-quiverai/promptfooconfig.vectorize.yaml` — image-to-SVG comparison.
- `examples/provider-quiverai/promptfooconfig.pipeline.yaml` + `pipeline-provider.js` — the chained pipeline that produced a coherent red-panda icon set on a live run.

A new "Pipeline" section in `site/docs/providers/quiverai.md` walks through the same flow inline, so docs alone are enough to ship the pattern.

## Live verification

[...]

PR #9470, opened 2026-05-26
fix: store raw SVG previews as generic blobs

fallintoplace:
## Summary
This keeps raw SVG text in `response.output`, but stores the preview blob as a generic download instead of an inline SVG image.

## Why
The raw SVG preview path was feeding a valid `image/svg+xml` blob straight into the media library path. This change keeps the narrow preview-indexing behavior, but makes that blob land as `application/octet-stream` with kind `other` so current UI flows stop treating it like an inline image.

## What changed
- store raw SVG preview bytes as a generic blob instead of `image/svg+xml`
- keep `response.output` untouched so downstream judges still see the original SVG text
- add regression coverage for the downgraded preview storage behavior

## Validation
- `npx vitest run test/blobs/extractor.test.ts test/evaluator/execution.test.ts`
- `npm run l`
- `npm run tsc`

### DIFF
--- a/src/blobs/extractor.ts
+++ b/src/blobs/extractor.ts
@@ -15,7 +15,7 @@ interface BlobContext {
   promptIdx?: number;
 }
 
-type BlobKind = 'audio' | 'image';
+type BlobKind = 'audio' | 'image' | 'other';
 
 function isDataUrl(value: string): boolean {
   return /^data:(audio|image)\/[^;]+;base64,/.test(value);
@@ -213,7 +213,7 @@ function createStoreOnce(blobContext: BlobContext): StoreOnce {
   };
 }
 
-function getRawSvgOutputPreview(output: string): { dataUrl: string; uri: string } | null {
+function getRawSvgOutputPreview(output: string): { base64: string; uri: string } | null {
   const trimmed = output.trim();
   if (!trimmed.startsWith('<')) {
     return null;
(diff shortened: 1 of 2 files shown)


mldangelo:
Hey @fallintoplace — thanks for the careful work here, the regression coverage and downscoped change are appreciated.

Before merging I want to make sure I understand the motivation. I pulled the branch, ran the unit tests (61/61 pass), and ran the QuiverAI Arrow example end-to-end against the live API. The preview-indexing path this change targets is intentional — it was added in #9139 so SVG-generating providers (Arrow / Vectorize) show up visually in the Media library alongside other generated images. With this PR applied, those SVGs land as `application/octet-stream` / `kind: other`, so they lose the inline preview in the grid, drop out of the "Images" filter, and clicking a card downloads the file instead of opening the detail modal.

Could you share:

1. A `promptfooconfig.yaml` (or curl / mock provider) that reproduces the behavior you're trying to fix?
2. What did the UI do that felt wrong, and what did you expect to see instead?

Going to close for now since I can't repro the problem on my end, but please drop a repro in a comment and I'll happily reopen and take another look. Thanks again for the time you spent on this.

fallintoplace:
@mldangelo This was intended as a small defensive hardening change for a potential security concern around storing and serving untrusted raw SVG from the app origin. If the team doesn’t view that as a meaningful risk here, I’m fine with closing it.

PR #9472, opened 2026-05-27
fix(providers): apply path-traversal guard to all callback loaders

mldangelo:
## Summary

Apply the path-traversal guard from `GoogleGenericProvider` to **all** function-callback loaders by extracting a shared helper (`loadCallbackFromFileUrl`, `resolveCallbackPath`) in `src/util/functions/loadFunction.ts`.

Before this PR, only `GoogleGenericProvider` (`src/providers/google/base.ts`) rejected `file://` callback references that resolved outside the active `basePath`. The six other callback loaders accepted **any** resolvable path, including absolute paths and `..` traversals — so a malicious config file could load arbitrary modules from anywhere on the host.

After this PR, all seven loaders reject paths that escape `basePath`.

## ⚠️ Breaking change

Configs that point at callback files **outside** the config's `basePath` will now fail with:

```
Path traversal rejected: '<path>' resolves outside the configured base directory.
Place the callback file inside the base directory, or set
PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD=true to opt out (NOT recommended).
```

Affected configs (example): `functionToolCallbacks: { x: 'file:///etc/me/callbacks.js' }` when the eval's `basePath` is `~/projects/myproject`. The fix is to move the callback file inside the project, reference it relatively, or set the opt-out env var.

This matches the existing behavior of `GoogleGenericProvider` (added in #6924).

### Shared helper (`src/util/functions/loadFunction.ts`)
- `resolveCallbackPath(filePath, basePath?)` — resolves against `basePath` and throws if it escapes (lexical check; symlinks not resolved).
- `loadCallbackFromFileUrl(fileRef, opts?)` — combines `parseFileUrl` + `resolveCallbackPath` + `importModule` + export extraction. Supports `lenient: true` for callers that need the legacy default-export fallback.
- `CallbackPathTraversalError` — distinct class so callers can downcast/classify the security event.
- `PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD=true` — opt-out env var for backward compat.

### Providers
- Six providers (`azure foundry-agent`, `bedrock converse`, `google base`, `google live`, `openai assistant`, `openai chat`) call the helper directly with strict mode; their catch/rethrow uses `Error(msg, { cause })` so the underlying error (including `CallbackPathTraversalError`) is preserved.
- `FunctionCallbackHandler` (the 7th callsite) uses `lenient: true` to preserve its long-standing silent default-export fallback. Its `processCall` now surfaces path-traversal rejections at `logger.warn` so the security control isn't silently demoted to "bad tool call".

### Tests
- New `test/util/functions/loadCallbackFromFileUrl.test.ts` (16+ tests) covering `..`, deep traversal, absolute paths, Windows drive letters, named/default exports, missing-export errors, `logPrefix`, error instance fields, opt-out env, and behavior-change regressions.
- Provider-level integration regressions in `functionCallbackUtils.test.ts` and `openai/chat.test.ts` verify the traversal is rejected before `importModule` is called and surfaced at warn level.
- Windows tests properly use `try/finally` for `cliState.basePath` cleanup.

Net diff: **+339/-112** including security feature, opt-out, two new exports, regression tests, and docs.

## Test plan

- [x] `npx vitest run test/util/functions test/providers` — 5979 tests pass
- [x] `npm run tsc` — clean
- [x] `npm run l` — only pre-existing complexity warnings (unrelated)
- [x] `npm run f` — clean
- [x] Verified `CallbackPathTraversalError` is preserved through `instanceof` after the provider catch/rethrow

## Depends on

Builds on top of #9469. Once #9469 merges this rebases cleanly.


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/88109f70-fa9d-450f-87ec-318710739dc8?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on f37c6140f298e8b44bbdf961d3ed257de54adaed. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 5c53fd1bd4942c9d83d53dafc1a460a4c5eb8e60 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/69d68be8-9dbb-4a2a-884b-e5f0bd668a24?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 3:43AM |
| 0fe23784cd3dc4f00510a07967cd6ee7be711b7f | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/b198d036-96a5-4605-94ec-65594b6bffcf?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 4:06AM |
| c2015bb45bd4afc8ef10bc056de2ef519b4853d4 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/cec841a8-1b1c-4535-9e8b-5b10fe436389?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 4:11AM |
| 32c1685c94e38e0cfb137e859e5d5cffd9cb47c8 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/ce967c19-a8e2-4b72-8aa7-1c6df159fa74?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 4:35AM |
| d06f5a9e057f602ca14dc216adcdbfa36cbd7e0d | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/12d2cd4b-5435-49ca-84ce-7837ce467324?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 4:39AM |
| fdb296b0f41695544a627b617ec363fa1cea972b | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/9cd8aa0a-920a-497d-a0f6-c66fa1ed2fe7?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 4:58AM |
| f37c6140f298e8b44bbdf961d3ed257de54adaed | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/88109f70-fa9d-450f-87ec-318710739dc8?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 27, 2026 5:29AM |

</p>
</details>

mldangelo:
@codex review

Re-review requested after addressing prior findings:
- Fixed the `relativePath.startsWith('..')` false-positive on `..foo/cb.js` paths (caught independently by Codex + Copilot)
- Added `PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD` opt-out env
- New `CallbackPathTraversalError` class with preserved `cause`
- Migrated the 7th callsite (FunctionCallbackHandler) with `lenient: true`
- Path-traversal failures now surface at `logger.warn` instead of silent debug
- Extended docs to azure.md

Latest commit: c2015bb45

src/util/functions/loadFunction.ts
```diff
@@ -110,3 +126,166 @@ export function parseFileUrl(fileUrl: string): { filePath: string; functionName?
...
+ *   or `process.cwd()`.
+ * @returns The resolved absolute path.
+ * @throws {CallbackPathTraversalError} If the resolved path escapes basePath.
+ */
+export function resolveCallbackPath(filePath: string, basePath?: string): string {
+  const effectiveBase = basePath ?? cliState.basePath ?? process.cwd();
+  const normalizedBase = path.resolve(effectiveBase);
+  const resolvedPath = path.resolve(normalizedBase, filePath);
+
+  if (getEnvBool('PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD', false)) {
```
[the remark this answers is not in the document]
mldangelo: Thanks — this concern applies equally to every existing `PROMPTFOO_DISABLE_*` flag (none of them are in `ProviderEnvOverridesSchema`). The established convention is shell env / `.env` file for these toggles. Adding only this one to the schema would create an inconsistency; expanding the schema to all of them is a separate refactor. Filed for follow-up rather than addressing here.

src/util/functions/loadFunction.ts
```diff
@@ -132,3 +149,219 @@ export function parseFileUrl(fileUrl: string): { filePath: string; functionName?
...
+        `PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD=true to opt out (NOT recommended).`,
+    );
+    this.name = 'CallbackPathTraversalError';
+    this.filePath = filePath;
+    this.basePath = basePath;
+  }
+}
+
+function isCallbackPathGuardDisabled(): boolean {
+  const value = process.env.PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD;
```
jameshiester-oai, line 192: **[P1] Keep the security opt-out outside config-controlled env loading**

`commandLineOptions.envPath` is accepted from the config, resolved relative to that config, and passed to `setupEnv`, whose dotenv call uses `override: true`. Because this function reads `process.env` at callback-load time, a shared config can include `envPath: ./disable-guard.env`, set this flag there, and then reference an out-of-base callback. On this head the callback first threw `CallbackPathTraversalError`; after `resolveConfigs` + `setupEnv` loaded the config-selected env file, the same outside module imported and executed. That defeats the PR's malicious-config boundary. Capture this opt-out from operator-controlled launcher/CLI state before untrusted config env files are loaded, or prevent config-selected env files from setting this security toggle.

src/util/functions/loadFunction.ts
```diff
@@ -132,3 +149,219 @@ export function parseFileUrl(fileUrl: string): { filePath: string; functionName?
...
+ * callback file paths supplied via config.
+ *
+ * @param filePath The (untrusted) relative or absolute path to resolve.
+ * @param basePath The trusted base directory. Defaults to `cliState.basePath`
+ *   or `process.cwd()`.
+ * @returns The resolved absolute path.
+ * @throws {CallbackPathTraversalError} If the resolved path escapes basePath.
+ */
+export function resolveCallbackPath(filePath: string, basePath?: string): string {
+  const effectiveBase = basePath ?? cliState.basePath ?? process.cwd();
```
jameshiester-oai, line 227: **[P2] Bind containment to the provider's configured base path**

All migrated provider call sites omit `options.basePath`, so this fallback chooses mutable `cliState.basePath`. However, `loadApiProvider(..., { basePath })` already injects the caller's explicit base into provider config. With provider base `.../global/intended` and stale/broader global base `.../global`, I loaded `.../global/outside.mjs`: checking against the provider base rejects it, while the actual OpenAI loader imported and executed it. This leaves programmatic/multi-config consumers outside the promised trust root. Capture the immutable load-context base on each provider/handler and pass it through to `loadCallbackFromFileUrl`; do not use process-global state for this security decision.

src/util/functions/loadFunction.ts
```diff
@@ -132,3 +149,219 @@ export function parseFileUrl(fileUrl: string): { filePath: string; functionName?
...
+  const normalizedBase = path.resolve(effectiveBase);
+  const resolvedPath = path.resolve(normalizedBase, filePath);
+
+  if (isCallbackPathGuardDisabled()) {
+    return resolvedPath;
+  }
+
+  // Check 1 — lexical: cheap, runs even for non-existent files.
+  assertWithinBase(filePath, resolvedPath, normalizedBase);
```
jameshiester-oai, line 236: **[P2] Let filesystem-aware containment decide for existing paths**

This lexical assertion throws before the realpath pass can establish that two spellings name the same in-base tree. On this host `/tmp` resolves to `/private/tmp`: using base `/tmp/pr9472/.../intended` and the existing callback's canonical `/private/tmp/pr9472/.../intended/inside.mjs` throws `CallbackPathTraversalError`, although both are the same filesystem location. Alternate casing reproduces this on the default case-insensitive macOS volume as well. This is a new failure across all migrated loaders. When base and target exist, make canonical/platform-aware filesystem containment authoritative; retain lexical containment as the fallback when canonicalization cannot complete, so nonexistent paths and real symlink escapes remain protected.

jameshiester-oai (commented):
## Scope and verdict

Reviewed the full PR diff from base `c54a30668ad8319d76c20ae96e6680ad6c51a2c6` to head `fb2eec5753ff5231d4b5b50fa6a2c86b49fd4426`. There was no prior authenticated completion marker for this head.

Verdict: **1 P1 and 3 P2 findings.** The shared loader removes substantial duplication and the ordinary parent/sibling/symlink escape cases work, but the new security boundary can be disabled by the same config it is meant to constrain and is not consistently bound to the provider's configured trust root. It also rejects valid callbacks reached through canonical filesystem aliases.

CI is red only because Site, webui, and Node 20 coverage uploads all failed at the same Codecov v11.2.7 GPG verification step after their test suites passed (101, 4,816, and 19,688 tests respectively). Smoke, integration, docs/builds, Node 22/24/26, macOS, all Windows shards, style, and CodeQL pass. GitHub reports the PR mergeable but behind; a local merge-tree check against current `origin/main` (`81adadadd2fd1bb19b80136ebe66c8c64102bc55`) completed without conflicts.

## Verified findings

- **P1 — config-selected env files bypass the guard.** `commandLineOptions.envPath` is config-controlled, resolved by `resolveConfigs`, and loaded with dotenv `override: true` before lazy callback execution. A config can therefore select an adjacent env file that sets `PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD=true`, then load an out-of-base callback. The control case threw `CallbackPathTraversalError`; after the config-selected env file was loaded, the same callback executed.
- **P2 — callback containment uses mutable global state instead of the provider base.** `loadApiProvider(..., { basePath })` stores that base on provider config, but every migrated call leaves `loadCallbackFromFileUrl`'s `basePath` option unset. With a narrower provider base and broader `cliState.basePath`, an outside callback executed successfully.
- **P2 — canonical in-base aliases are rejected.** The lexical check runs before realpath. On this host, a callback under `/private/tmp/...` was rejected when the configured base named the same tree through `/tmp/...`; the real CLI reproduced the same false rejection. Case aliases on the default case-insensitive macOS filesystem also reproduced it.
- **P2 — the Windows security test does not invoke the guard.** It only verifies `path.win32.relative` behavior, so it remains green if `resolveCallbackPath` loses cross-drive/UNC containment.

All six existing review threads were rechecked. The `..foo` and symlink-escape fixes remain effective. The earlier resolved `env:` discussion concerns schema-filtered provider env and does not cover `commandLineOptions.envPath`, which mutates `process.env`. No unresolved finding is duplicated here.

## Validation performed

- `node_modules/.bin/vitest run test/util/functions/loadCallbackFromFileUrl.test.ts test/util/functions/loadFunction.test.ts test/providers/functionCallbackUtils.test.ts test/providers/google/base.test.ts test/providers/google/live.test.ts test/providers/openai/chat.test.ts test/providers/openai/assistant.test.ts test/providers/azure/foundry-agent.test.ts test/providers/bedrock/converse.test.ts --sequence.shuffle=false` — exit 0; 9 files / 479 tests passed.
- `npm run tsc` — exit 0.
- `npm run build` — exit 0, including app build.
- `cd site && SKIP_OG_GENERATION=true npm run build` — exit 0.
- Targeted Biome check of the 15 changed TypeScript files — exit 0; only 8 pre-existing complexity warnings.
- `git diff --check c54a30668ad8319d76c20ae96e6680ad6c51a2c6...fb2eec5753ff5231d4b5b50fa6a2c86b49fd4426` — exit 0.
- Offline real-module/adversarial matrix covered relative, parent/deep/sibling/absolute, `..foo`, symlink, named/default export, opt-out, encoded-segment, Windows drive, and UNC cases.
- `npm run local -- eval -c /tmp/pr9472-cli/config.yaml --no-cache -o /private/tmp/pr9472-cli/output.json` — exit 0 with the expected canonical-alias rejection recorded as `isError: true`; the same fixture through its canonical config path executed `callback-ok`.
- A separate isolated echo/scoring-function CLI probe exited 100 because its intentionally minimal fixture returned a number rather than a `GradingResult`; it was outside this PR's function-tool-callback scope and produced no finding.

The requested `/pr-review-toolkit:review-pr all parallel` command was not available as a callable skill in this environment, so I used the documented manual correctness, tests, docs/comments, error-handling, type-design, and simplification lenses, plus three independent read-only reviews. No live OpenAI, Azure, Bedrock, or Google service call was made; provider callback behavior was exercised through local modules, provider instances, `FunctionCallbackHandler`, and the local CLI.

The main simplification opportunity is to retain the shared helper while making its trust root an explicit immutable dependency of each provider/handler, rather than falling back to process-global state.

test/util/functions/loadCallbackFromFileUrl.test.ts
```diff
@@ -0,0 +1,299 @@
...
+  });
+
+  // Windows-specific assertion: with a non-root basePath like `C:/work`,
+  // a config that names a Windows-absolute path on the same drive must be
+  // rejected. On POSIX `C:/...` is a relative segment, so this test asserts
+  // a platform-conditional behavior using path.win32 directly.
+  it('rejects Windows-absolute paths outside a non-root basePath (using path.win32 semantics)', () => {
+    const base = 'C:\\work';
+    const target = 'C:\\etc\\secrets.js';
+    const resolved = path.win32.resolve(base, target);
```
jameshiester-oai, line 126: **[P2] Exercise the callback guard in the Windows regression test**

This test only calls Node's `path.win32.resolve/relative`; it never calls `resolveCallbackPath` or any extracted containment helper. It therefore stays green if the guard drops its cross-drive/UNC rejection entirely. The other committed cases cover the host platform's `..` branch, but no test invokes the security control for `C:\\work -> D:\\...` or a foreign UNC root. Add a Windows-runner test that calls the guard and expects `CallbackPathTraversalError`, or extract a path-flavor-parametric containment helper and test `path.win32` through that production code.

mldangelo:
Synced with current `main` and independently verified the vulnerability and the fix. This is the most security-relevant PR still open on the Anthropic/Bedrock sweep, so here's the evidence.

### The vulnerability is real on `main`

`FunctionCallbackHandler.loadExternalFunction` on `main` resolves with no containment check:

```ts
const resolvedPath = path.resolve(cliState.basePath || '', filePath);
```

`path.resolve` ignores `basePath` entirely when `filePath` is absolute, and `..` escapes it. Ran the end-to-end path on `main` with `basePath = /tmp/pf-traversal-base` and a callback at `file:///tmp/pf-traversal/evil.js`:

```
main:   loads and executes the module  → assertion toContain('PWNED') PASSED
branch: rejected                       → assertion not.toContain('PWNED') PASSED
```

Same probe, same harness, opposite outcomes. Confirmed all seven loaders now route through `loadCallbackFromFileUrl`:

```
functionCallbackUtils.ts   azure/foundry-agent.ts   google/base.ts
google/live.ts             bedrock/converse.ts      openai/chat.ts
openai/assistant.ts
```

`main` has **zero** references to `loadCallbackFromFileUrl` across all seven, so nothing has superseded this.

### Merge conflict resolved

One conflict, in `src/providers/openai/assistant.ts`. `main` relocated the GenAI tracing helpers from `../../tracing/genaiTracer` to `../tracing` (adding `withGenAIToolSpan`), so this branch's import block was stale — dropped in favour of main's, which was already present lower in the file. `main`'s `parseFileUrl` import went too, since this branch replaces the manual `file://` handling with `loadCallbackFromFileUrl` and leaves `parseFileUrl` unused.

### Validation

- `npx vitest run` — **23,323 passed**, 10 skipped (full backend suite)
- `npx tsc --noEmit` — clean
- `npx @biomejs/biome ci .` — 0 errors
- `npm run architecture:check` — passes
- Prettier — clean

The three CI failures on the previous head were from June and included `Test on Node 20.20`, a version `main` no longer tests.

### On the breaking change

Checked whether the opt-out needs documenting more widely. Only `azure.md` and `openai.md` document `functionToolCallbacks`, and both already carry the warning — the Google and Bedrock provider pages don't document callbacks at all, so a warning there would be orphaned. Current coverage looks right.

Worth noting *why* this class of bug happened: there were six near-identical copies of `loadExternalFunction`, and the guard added to the Google one never reached the other five. Collapsing them onto one loader is what makes the fix hold.

PR #9473, opened 2026-05-27
fix(providers): handle Codex SDK rate limits

ianw-oai:
## Summary
- classify terminal Codex SDK TPM/RPM throttles as scheduler-visible transient rate limits, honoring SDK retry hints and using a conservative wait when no hint is exposed
- retain intermediate SDK stream error events until the turn reaches a terminal outcome instead of returning reconnect progress as evaluation row errors
- expose `maxRetries` for Codex SDK providers and document the retry behavior

## Context
The Codex SDK currently surfaces turn and stream failures as message-only error payloads. Under a TPM limit, Promptfoo reduced concurrency but the provider still returned streamed reconnect notices and eventual failures as row errors instead of allowing the adaptive scheduler to wait and retry the affected row.

## Verification
- `npm run f`
- `npm run l`
- `npm run tsc`
- `./node_modules/.bin/vitest run test/providers/openai-codex-sdk.test.ts test/scheduler/providerWrapper.test.ts test/scheduler/providerRateLimitState.test.ts test/scheduler/retryPolicy.test.ts test/evaluator/gradingConcurrency.test.ts --configLoader runner --sequence.shuffle=false`


mldangelo (approved):

src/providers/openai/codex-sdk.ts
```diff
@@ -382,6 +391,120 @@ function getMinimalProcessEnv(): Record<string, string> {
...
+  'rate_limit_exceeded',
+  'insufficient_quota',
+  'billing_hard_limit_reached',
+  'billing_not_active',
+  'access_terminated',
+  'quota_exceeded',
+] as const;
+
+// Mirrors the HTTP retry path's fallback when the upstream error carries no reset hint.
+const CODEX_DEFAULT_RATE_LIMIT_WAIT_MS = 60_000;
```
mldangelo, line 421: Nit / appreciation: `60_000` here exactly matches `DEFAULT_RETRY_POLICY.maxDelayMs` in `src/scheduler/retryPolicy.ts`. That coupling is what bounds the no-hint retry to a single max-delay sleep + jitter rather than something unbounded — worth a one-liner comment noting the intentional alignment so a future tweak to either constant doesn't silently desync them.

src/providers/openai/codex-sdk.ts
```diff
@@ -382,6 +391,120 @@ function getMinimalProcessEnv(): Record<string, string> {
...
+  const errorRecord =
+    typeof error === 'object' && error !== null ? (error as Record<string, unknown>) : undefined;
+  const status =
+    typeof errorRecord?.status === 'number'
+      ? errorRecord.status
+      : typeof errorRecord?.statusCode === 'number'
+        ? errorRecord.statusCode
+        : undefined;
+  const rawCode =
+    typeof errorRecord?.code === 'string' ? errorRecord.code.toLowerCase() : undefined;
```
mldangelo, line 485: Coverage gap (non-blocking): this branch handles SDK errors that carry a structured `.code` property (e.g. if Codex ever wraps OpenAI errors with `code: 'rate_limit_exceeded'`), but every new test mocks `new Error('…')` with message-only parsing. Worth a follow-up test that sets `.code` (and ideally `.status`) directly on the rejected error to lock this path against regressions — otherwise if the SDK shape changes upstream and the message wording stops matching the regex, we'd silently fall through to the generic error response.

src/providers/openai/codex-sdk.ts
```diff
@@ -1048,18 +1180,26 @@ export class OpenAICodexSDKProvider implements ApiProvider {
...
+            ? ` Previous stream error: ${state.lastStreamError}`
+            : '';
+        throw new Error(`Codex turn failed: ${errorMsg}${previousStreamError}`);
       }
       case 'error': {
         const errorMsg =
           typeof event.message === 'string' && event.message ? event.message : 'Stream failed';
-        logger.error('Codex stream error', { error: errorMsg });
-        throw new Error(`Codex stream error: ${errorMsg}`);
+        state.lastStreamError = errorMsg;
```
mldangelo, line 1118: Minor: `state.lastStreamError = errorMsg` overwrites on every `error` event, so if a turn emits multiple reconnect notices (e.g. `Reconnecting... 2/5` → `3/5` → `4/5` → `turn.failed`), only the last one is preserved in the surfaced error. Probably fine — this is strictly better than the old behavior where none of them surfaced — but if reconnect chains turn out to be common in practice, consider accumulating into an array and rendering the chain on terminal failure. Not blocking.

src/providers/openai/codex-sdk.ts
```diff
@@ -1005,6 +1128,13 @@ export class OpenAICodexSDKProvider implements ApiProvider {
         this.handleStreamingEvent(event, state, tracer, eventTime, skillRootPrefixes);
         state.lastEventTime = eventTime;
       }
+
+      if (!state.turnCompleted) {
+        if (state.lastStreamError) {
+          throw new Error(`Codex stream ended after error: ${state.lastStreamError}`);
+        }
+        throw new Error('Codex stream ended before turn completion');
```
mldangelo, line 1042: Nice side-effect catch: prior to this change, a stream that closed cleanly without ever emitting `turn.completed` would have fallen through to `buildStreamingTurnResult` and returned a partial result with `usage: undefined` rather than erroring. The new `turnCompleted` guard fixes that silent-failure mode in addition to the documented reconnect behavior. Worth mentioning in the release notes / changelog so users who were getting empty-but-successful rows know to expect a row error instead.

src/scheduler/types.ts
```diff
@@ -38,6 +38,9 @@ export function isProviderResponseRateLimited(
   if (result?.metadata?.rateLimitKind === 'quota') {
     return false;
   }
+  if (result?.metadata?.rateLimitKind === 'rate_limit') {
```
mldangelo, line 41: Good symmetric pairing with the `=== 'quota'` branch above. The explicit-metadata path means providers no longer need to embed `429` or `rate limit` substrings in `result.error` to be picked up — exactly the contract the new `providerWrapper` test exercises (error string is just `'SDK throttle'`). The substring fallback below remains a safety net for older providers that haven't migrated to structured metadata.

mldangelo (commented):
Nice piece of work — I went through the whole diff carefully against `HttpRateLimitError`, `formatRateLimitErrorMessage`, the scheduler's `isProviderResponseRateLimited`/`shouldRetry` paths, and ran the full verification test set locally (211/211 pass, lint+tsc clean). The three concerns bundled here are conceptually orthogonal and each one stands up:

1. **Structured rate-limit metadata from message-only SDK errors.** The regex/phrase set is appropriately conservative — the explicit `"tokens per (?:minute|min)"` phrasing means the `/tmp/tpm/output.json` test case correctly doesn't match, and `\b` boundaries protect against `4290`-style false positives. Nice.
2. **Explicit `rateLimitKind: 'rate_limit'` short-circuit in the scheduler.** Symmetric with the pre-existing `'quota'` branch and bypasses fragile substring fallback — the new `providerWrapper` test (error string `'SDK throttle'` with no `429`/`rate limit` substring) is exactly the right shape to lock that in.
3. **Streaming state machine: retain intermediate `error` events until a terminal event.** This also incidentally fixes a silent-failure mode where a stream closing without `turn.completed` would have returned a half-built result with `usage: undefined`; now that path throws. Worth calling out in the changelog.

A couple of small, **non-blocking** observations I left as inline notes:
- Coverage gap on the `error.code` extraction path (all new tests rely on message parsing).
- The "Previous stream error" trailer keeps only the most recent intermediate error; fine, but worth documenting if reconnect chains get common.
- The default 60s fallback elegantly pairs with `DEFAULT_RETRY_POLICY.maxDelayMs` — appreciate the consistency, just noting it for the next person.

Keeping my approval. Thanks for the structured error metadata over substring sniffing — that pattern will pay dividends for the next provider that hits this.

mldangelo-oai (approved):
LGTM. I reviewed the current rate-limit and streaming changes, the scheduler integration, tests, and live review-thread state on this head. The remaining notes are non-blocking; no blocking issue found.

PR #9474, opened 2026-05-27
feat(providers): add OpenAI originator header

ianw-oai:
Requests issued by built-in OpenAI providers to the OpenAI API now send `X-OpenAI-Originator: promptfoo` by default, while custom `headers` can override it. OpenAI-compatible custom gateway endpoints do not receive the automatic header.


mldangelo:
Reviewed end-to-end on `162517c`: 37 checks green, 785 unit tests pass on Node 24.16. Centralization is well-scoped — flagging the gaps the existing bot reviews surfaced, with verification, plus a couple of new items.

### P1 — needs a decision before merge

**1. Docs over-promise: `openai:agents:*` providers never send the header.** `site/docs/providers/openai.md:72-74` · `src/providers/openai/agents.ts`
`OpenAiAgentsProvider` inherits `getOpenAiRequestHeaders()` but `runAgent()` calls the `@openai/agents` SDK's `run()` without installing a custom OpenAI client, so the header isn't on the wire for `openai:agents:*`. The SDK exposes `setDefaultOpenAIClient(client)` / `OpenAIProvider({ openAIClient })` — wireable, but currently unset. Either wire it through or narrow the doc claim to enumerate the providers it actually applies to.

**2. ChatKit hardcodes the header, breaking the documented override.** `src/providers/openai/chatkit.ts:215-220`
`OpenAiChatKitOptions extends OpenAiSharedOptions` (which has `headers`), but `generateChatKitHTML` never reads `this.config.headers` — `'X-OpenAI-Originator': 'promptfoo'` is baked into the emitted JS literal. The doc claim that users can "override this header through the `headers` configuration option" doesn't apply here. Either thread `config.headers` into `generateChatKitHTML` (validating injected values like the existing `workflowId`/`version`/`userId` checks do) or call this out in the docs.

### P2 — important

**3. Case-insensitive override leaks both headers on the wire.** `src/providers/openai/index.ts:61-65`
Live-verified by instantiating the helper directly: `headers: { 'x-openai-originator': 'myapp' }` produces `{\"X-OpenAI-Originator\":\"promptfoo\",\"x-openai-originator\":\"myapp\"}`. undici will send both. One-line fix: scan `customHeaders` keys (case-insensitive) before merging in the default.

### Suggestions

- `getOrganization()` is called twice (`index.ts:63`) — cache in a local. (Copilot)
- `.hostname.toLowerCase()` (`index.ts:56`) is redundant — `URL` normalizes hostname.
- Empty `catch` (`index.ts:57-59`) — narrow to `TypeError` and add a `logger.debug` so \"originator missing\" is greppable.
- The implicit `customHeaders = this.config.headers` default conflicts with the 7-of-11 callers that pass it explicitly; consider dropping the default for clarity.
- `DEFAULT_OPENAI_ORIGINATOR` is misnamed — there's no \"override mechanism that has a default\" since both the constant and per-request override flow through the same code path. `OPENAI_ORIGINATOR_VALUE` reads cleaner.
- `video.ts:149-154` `getAuthHeaders` wrapper is unnecessary — 3 call sites can inline directly.
- `realtime.ts:716, 1391, 1985` build 3 near-duplicate WS header blocks; one private helper would dedupe.
- Test gaps: video `pollVideoStatus`/`downloadVideoContent` aren't covered (header asserted only on POST `/videos`); realtime asserts only 1 of 3 connect sites; ChatKit \"test\" is a string snapshot of the same hardcoded literal so drift can't be caught; no per-provider override-case or non-OpenAI-host coverage.

### Confirmed correct / dismissing

- **Transcription \"ignores per-prompt headers\" (Copilot)** — moot. `OpenAiTranscriptionOptions` has no `headers` field, so per-prompt overrides aren't a supported surface.
- **Video poll/download drops per-request overrides (Codex)** — pre-existing on `main`, not introduced by this PR. The provider-level `this.config.headers` still flows through via the helper's default parameter, so the originator still ships on follow-up calls. Per-prompt-only overrides wouldn't propagate, but that mirrors `main`.
- **URL-parse `try/catch`** — degrades to \"not OpenAI\" safely; downstream `fetch*` still surfaces the real error. Adding a `logger.debug` is the only nit.
- **Strict hostname equality** — correctly excludes Azure, custom gateways, and OpenAI-compatible vendors. Note this also excludes regional OpenAI endpoints like `eu.api.openai.com` — likely intentional, but worth a one-line ack in the PR description.

### Strengths

Centralization across 11 files is consistent. The URL gating is conservative in the right direction. Override does work for exact-case headers (verified live). Per-provider tests pin the positive case so the wiring won't be silently dropped in future refactors.

mldangelo-oai (approved):
Approving with known follow-up requested: please address the existing `X-OpenAI-Originator` override gaps so custom attribution behaves consistently. In particular, ChatKit hard-codes the value, video poll/download follow-up requests do not carry configured overrides, and transcription does not use merged per-prompt headers. Reviewed current head `162517cff5c4512d0413b88d18de9627bf3b3f39`.

PR #9479, opened 2026-05-27
fix(providers): harden callback file URL parsing

### LINKED ISSUE #9467: Windows file://C:/...:functionName callback parsing breaks in provider loaders

Gujiassh:
## Summary

Fixes #9467.

Windows callback references such as `file://C:/path/to/callbacks.js:handler` contain both a drive-letter colon and a named-export delimiter. This PR routes provider callback references through the shared parser and hardens that parser so it:

- supports Windows drive-letter callback URLs across OpenAI, Bedrock Converse, Google, Azure Foundry Agent, and the shared function callback handler
- normalizes canonical `file:///C:/...` paths only on Windows
- preserves POSIX callback filenames and directory names containing colons

## Relationship to #9469

#9469 substantially overlaps with this fix. This branch has been refreshed onto `main` and its overlapping parser behavior is aligned with #9469 so it remains a safe, reviewable contribution rather than being closed out.

If #9469 lands first, the useful remaining contribution here is additional regression coverage for:

- Windows-looking file URLs on POSIX, where `/C:/...` must remain a valid POSIX path
- named exports from callback paths whose directories contain colons

## Validation

- `npx vitest run test/util/functions/loadFunction.test.ts test/util/file.test.ts test/util/fileReference.test.ts test/providers/functionCallbackUtils.test.ts test/providers/google/base.test.ts test/providers/google/live.test.ts test/providers/openai/chat.test.ts test/providers/openai/assistant.test.ts test/providers/bedrock/converse.test.ts test/providers/azure/foundry-agent.test.ts --sequence.shuffle=false` (585 tests passed)
- `npx vitest run test/test-hygiene.test.ts test/providers/functionCallbackUtils.test.ts test/util/functions/loadFunction.test.ts test/providers/google/base.test.ts --sequence.shuffle=false` (189 tests passed)
- `npm run tsc` passed
- `npm run l` passed
- `npm run f` passed
- GitHub Actions on the final branch head: 29 successful checks, 2 skipped, 0 failures on May 28, 2026

`npm run l` and `npm run f` complete with existing non-blocking complexity warnings in touched legacy files.


mldangelo-oai:
Thank you!

PR #1669, opened 2024-09-14
feat(webui): display custom namedScores

khp:
Addresses issue raised in https://github.com/promptfoo/promptfoo/issues/1626

Given a custom assertion that sets some `namedScores` such as:
```
def get_assert(output, context):
    return {
      'pass': True,
      'score': 0.11,
      'reason': 'Looks good to me',
      'namedScores': {
         'answer_similarity': 0.12,
         'answer_correctness': 0.13,
         'answer_relevancy': 0.14,
      }
    }
```

Render something like:
image

Note that [docs](https://www.promptfoo.dev/docs/configuration/expected-outputs/python/) indicate `named_scores` is the correct field name but currently only snake case is supported.

It might be interesting to allow different types of aggregation and formatting of top-level metrics. This also does not support `namedScores` that are nested in `GradingResults.componentResults`.

PR #9480, opened 2026-05-27
feat(output): add named metric columns to CSV exports

### LINKED ISSUE #1167: Request - Extract the test result as a CSV file that all the metrics and the results of test included.
One of the things that is very important in measuring LLMs is statistics. 
The CSV of the test result currently provides good information but I need to have exported all the metrics with the scores so for instance I can say in terms of LLM-rubric that the average score for all tests is %40 of all the defined. it would be much better if each result was extracted in a dedicated column in CSV format. 
Many thanks.

Gujiassh:
## Summary

Adds dedicated CSV columns for named metrics in eval exports while keeping the existing `Named Scores` JSON summary column.

For each prompt column, named scores now also export as stable `Metric: <name>` columns, making per-metric scores easier to analyze in spreadsheets and downstream reporting tools.

Fixes #1167.

## Testing

- `npx vitest run test/server/utils/evalTableUtils.test.ts test/util/output.test.ts --sequence.shuffle=false`
- `npx @biomejs/biome check src/server/utils/evalTableUtils.ts test/server/utils/evalTableUtils.test.ts`
- `npm run tsc`
- `git diff --check`


mldangelo:
@chatgpt-codex-connector review
@copilot-pull-request-reviewer review

mldangelo (approved):
LGTM — thanks for the feature, @Gujiassh! Adding dedicated `Metric: <name>` columns to the CSV exporter is a great quality-of-life improvement that closes a long-standing request (#1167), and the implementation is clean and well-tested.

I pushed a few follow-up commits to address two findings that came up during review and to keep the streaming CLI and WebUI exporters consistent:

- Source metric column names from `metrics.namedScoresCount` (incremented per-row) rather than `metrics.namedScores` (also includes derived metrics), so CLI exports don't gain phantom always-empty columns for `derivedMetrics`. This was the divergence between `streamEvalCsv` and `evalTableToCsv` for evals with derived metrics.
- For legacy evals and external imports that persist `metrics.namedScores` without a corresponding `namedScoresCount` (pre-PR #1669 evals, raw `POST /api/eval` payloads), do a bounded-memory discovery pass over result batches before emitting headers — collects only the metric name sets and `hasDescriptions`, never the row content. Modern evals (count map present, even if `{}`) keep the single-pass streaming path.

CI green, tests pass, end-to-end verified across modern/legacy/multi-batch/derived-metric scenarios. Ready to ship.

PR #9484, opened 2026-05-27
feat(assertions): support glob patterns in trajectory:tool-args-match ignore

RitwijParmar:
> **Maintainer note:** Rebased onto `main` and reshaped after the `defaults`
> option (#9471) and the `ignore` option (#9466). This PR is **stacked on
> #9466** — see the comment below. Original work and credit by @RitwijParmar
> (carried as a commit co-author).

## Summary

Extends the `ignore` option (#9466) so an entry containing the glob characters
`*` or `?` is treated as a **key pattern** instead of an exact name:

```yaml
  value:
    name: orders
    mode: exact
    args: { status: Q }
    ignore: ['*_id'] # drops request_id, order_id, …
```

Plain entries without glob characters remain exact, case-sensitive matches. The
implementation reuses the existing `matchesPattern` helper already used for
trajectory step matching.

**Stacking:** this branch currently contains the base `ignore` implementation
from #9466 plus the pattern delta. Once #9466 merges, syncing this branch with
`main` reduces the diff to just the glob-pattern support.

## Test plan

- Pattern tests added to the `ignore` suite: glob across multiple keys (with the
  `Ignored argument(s): …` reason), a glob that misses a hallucinated key still
  fails, `?` single-character globs, and plain entries staying exact.
- `npx vitest run test/assertions/trajectory.test.ts` → all pass.
- Docs: `site/docs/configuration/expected-outputs/deterministic.md`.

Refs #9458.


mldangelo:
Hi @RitwijParmar — thank you for being first to jump on #9458; your exact-mode-with-defaults framing and flexible input handling shaped the direction here.

Three PRs arrived for the same issue, so we converged on two fields: **`defaults`** (tolerate when the value matches, landed in #9471) and **`ignore`** (drop by name, #9466). To give your PR a distinct, non-overlapping home, I used maintainer edit to rebase it onto `main` and reshape it into a **glob-pattern extension of `ignore`** — e.g. `ignore: ['*_id']` to drop `request_id`, `order_id`, and so on. Your original commit and authorship are preserved, and the new commit credits you as co-author.

One heads-up: this is **stacked on #9466**, so until that merges the diff here also includes the base `ignore` code; once #9466 lands, syncing with `main` collapses this to just the pattern support. Would you take a look and tell me if this works for you? Glad to tweak.

mldangelo:
Update: #9466 (the base `ignore` option) just merged to `main`, so I synced this branch — its diff is now **just the glob-pattern support** (+58/−1): the `matchesIgnoreKey` helper, the pattern docs note, and four pattern tests. Ready for a look whenever you have a minute. Thanks again, @RitwijParmar!

PR #9486, opened 2026-05-27
fix(db): replace better-sqlite3 with libsql

wholley-oai:
## Summary

Promptfoo stores local evals, redteam outputs, blobs, and traces in SQLite. The native better-sqlite3 dependency is sensitive to Node ABI mismatches, so npx/global installs can fail when the cached package was built under a different Node version. This draft replaces the direct better-sqlite3 runtime with libsql while keeping the supported Node range at ^20.20.0 || >=22.22.0.

## What changed

- Swapped the SQLite Drizzle driver to @libsql/client/node and made database access async.
- Added serialized top-level write handling and transaction reuse so existing call sites keep the same behavior with async libsql queries.
- Updated CLI/server/model utilities, migrations, tracing, docs, and tests for the async DB API.
- Removed the earlier node:sqlite approach because it would have raised the minimum Node version.

## Validation

- npm run l
- npm run f
- npm run tsc -- --pretty false
- npm run build
- npx vitest run test/database/index.test.ts test/util/databaseWrite.test.ts test/database.test.ts test/migrate.test.ts test/util/libsqlBindingErrors.test.ts test/util/databaseStandalone.test.ts --sequence.shuffle=false
- npx vitest run test/database/index.test.ts test/util/databaseWrite.test.ts test/database.test.ts test/models/eval.test.ts test/models/eval.filteredMetrics.test.ts test/models/evalHighlights.test.ts test/models/evalPerformance.test.ts test/models/evalUserRated.test.ts test/server/eval.test.ts test/server/routes/eval.filteredMetrics.test.ts test/server/routes/modelAudit.test.ts test/util/calculateFilteredMetrics.test.ts test/util/databaseDelete.test.ts test/util/redteamProbeLimit.test.ts test/commands/import.test.ts test/commands/retry.test.ts test/migrate.test.ts test/tracing/store.test.ts test/tracing/storeRetention.test.ts test/util/databaseStandalone.test.ts test/util/libsqlBindingErrors.test.ts --sequence.shuffle=false
- npm run local -- --version
- npm run local -- eval -c test/fixtures/file-script-assertions/e2e-test-config.yaml --no-cache --no-share -o /private/tmp/promptfoo-libsql-eval-clean-branch.json


mldangelo:
Comprehensive review using three specialist agents (code review, silent-failure hunt, test coverage) against the branch tip (`eeaeef6a38`). All findings independently verified against the actual files in this branch.

For context: I just shipped a fix for the parallel PR #9009 (same migration, different author). That review surfaced the same patterns; **two of the critical bugs below have already been fixed in #9009 and could be ported**.

## Critical (recommend blocking merge)

### C1. `getTraceSpans()` crashes with `TypeError: db.select is not a function` on every call

**`src/tracing/store.ts:521`**

```ts
const db = (traceStoreInstance as any).getDatabase?.() ?? traceStoreInstance['getDatabase']?.();
if (!db) { throw new Error('TraceStore database has not been initialized'); }
const rows = await db.select().from(spansTable)…
```

`TraceStore.getDatabase()` (line 206) is `async` and always returns a Promise — never `undefined`. The `if (!db)` guard cannot fire because a Promise is truthy. Calling `db.select()` on a Promise throws `TypeError: db.select is not a function` immediately. The error message even claims the DB was never initialized, which is false. The `as any` cast at line 521 is precisely what hid this from TypeScript.

Reproduced as a vitest in a worktree of this branch.

`getTraceSpans` has **zero test coverage** — `grep -rn getTraceSpans` finds no callers in tests, which is why this slipped through.

**Fix** (the approach taken in #9009's [`cf479dac80`](https://github.com/promptfoo/promptfoo/pull/9009/commits/cf479dac80)): delete the 70-line reflection helper and delegate to the public `TraceStore.getSpans(traceId, options)`, which is already correct. Side benefit: the public method passes raw (pre-sanitize) attributes to `deriveSpanKind`, whereas the duplicated free function was passing post-sanitize attributes — a subtle classification bug.

### C2. Root `db.run/select/...` inside a `db.transaction(...)` callback deadlocks

**`src/database/index.ts:92-104`**

```ts
const serializeClientMethod = <TArgs extends unknown[], TResult>(
  method: (...args: TArgs) => Promise<TResult>,
) => {
  return (...args: TArgs) => runSerialized(() => method(...args));
};
client.execute = serializeClientMethod(client.execute.bind(client)) as typeof client.execute;
client.batch = serializeClientMethod(client.batch.bind(client));
client.migrate = serializeClientMethod(client.migrate.bind(client));
client.executeMultiple = serializeClientMethod(client.executeMultiple.bind(client));
```

`serializeClientMethod` queues every `client.execute/batch/migrate/executeMultiple` call against `operationQueue` unconditionally. The transaction wrapper correctly bypasses queueing for nested `db.transaction()` calls (via the AsyncLocalStorage check at line 107-112), but **not** for root client calls. If a transaction callback (or any helper it calls) issues a root `db.run/select/insert` instead of going through the `tx` handle, that operation queues behind the outer transaction — which is itself waiting for the callback to resolve — and the promise never settles.

No live triggering call site in the current code, but the structural bug means the next reviewer to pull in a helper inside a transaction silently hangs. Reproduced with a 1-second-timeout vitest:

```ts
db.transaction(async (tx) => {
  await tx.run('INSERT INTO t VALUES (1)');
  await db.run('SELECT 1');  // ← deadlocks
});
```

The existing nested-tx test at `test/database/index.test.ts:144-165` covers `db.transaction` inside `db.transaction` (which hits the ALS fast path) but not the root-db-in-tx case.

**Fix** (the approach taken in #9009's [`485ceff33b`](https://github.com/promptfoo/promptfoo/pull/9009/commits/485ceff33b)): inside `serializeClientMethod`, check `activeTransaction.getStore()` and bypass `runSerialized` when called inside an active transaction. WAL mode lets SELECTs from a fresh per-call connection proceed; write attempts surface as SQLITE_BUSY (the correct loud failure for code that should be using the `tx` handle).

### C3. `PATCH /:id` returns 200 OK even when the DB write failed (silent data loss)

**`src/util/database.ts:173-200` + `src/server/routes/eval.ts:281-287`**

`updateResult` catches DB errors, logs them, and returns `void` either way:

```ts
try {
  await db.update(...).run();
} catch (e) {
  logger.error(`Error updating result: ${e}`);
}
```

The route handler then has an empty catch:

```ts
try {
  await updateResult(id, body);
  res.json({ message: 'Eval updated successfully' });
} catch {
  res.status(500).json({ error: 'Failed to update eval table' });
}
```

The inner function already swallowed the error, so the outer `catch` never fires. Users get **`200 OK` with "Eval updated successfully"** for failed updates. No log on the route path either, no breadcrumb anywhere.

**Fix:** rethrow from `updateResult` (or return a status object), and add `logger.error('[PATCH /eval/:id] Failed', { id, error })` inside the route catch — mirror the existing pattern at `src/server/routes/eval.ts:879`.

## Important

### I1. libsql binding handler misses ESM errors and over-matches generic packages

**`src/util/libsqlBindingErrors.ts:3, 18`**

(a) The regex `/@libsql\/(?<target>[a-z0-9-]+)/i` matches `@libsql/client` and `@libsql/core` too. A broken-node_modules error for the wrapper package gets rendered as "platform binding for 'client' is missing — install `@libsql/client`" — misleading. Narrow to known platform-target patterns (`darwin-*`, `linux-*`, `win32-*`) or exclude `client`/`core`/`hrana-client`.

(b) The handler only matches CJS `MODULE_NOT_FOUND`, not ESM `ERR_MODULE_NOT_FOUND`. The lazy imports at `src/database/index.ts:130-133` are dynamic `import('@libsql/client/node')` — ESM. If the top-level package itself is missing (broken install, missing optional dep), Node throws `ERR_MODULE_NOT_FOUND` and the friendly banner doesn't fire. Mirror the matcher in `src/util/packageImportErrors.ts:9-10` which accepts both codes.

The `test/util/libsqlBindingErrors.test.ts` suite covers only the CJS code path.

### I2. WAL-mode test became self-referential after commit `eeaeef6a38`

**`test/database.test.ts:57-77`**

The most recent commit swapped the independent verification driver from sync `libsql` to `@libsql/client/node` — the same module the application uses. If a future libsql version silently downgrades or short-circuits `PRAGMA journal_mode = WAL`, both writer and verifier would agree and the test would still pass. Worse, the assertion only confirms the in-memory PRAGMA, not the on-disk artefacts.

**Fix:** either revert to a different driver for verification, or add an assertion that exercises on-disk state (e.g., assert a `-wal` sidecar exists on the configured DB path after a write).

### I3. `closeDb()` race with in-flight `getDb()`

**`src/database/index.ts:163-189`**

`closeDb()` awaits `PRAGMA wal_checkpoint(TRUNCATE)` and then `sqliteInstance.close()`. During the await, a concurrent `getDb()` can hit the `dbInstance` fast path and return the about-to-be-closed instance. Subsequent operations on the returned handle then fail with "database closed" errors. Guard with a closing flag, or null `dbInstance` synchronously at the top of `closeDb` before any awaits.

### I4. WAL checkpoint failure logged at `debug` in `closeDb()`

**`src/database/index.ts:172`**

```ts
} catch (err) {
  logger.debug(`Could not checkpoint WAL file: ${err}`);
}
```

A failed `PRAGMA wal_checkpoint(TRUNCATE)` immediately before close is a real durability signal — committed-but-uncheckpointed pages depend on WAL replay at next open, and if the WAL is later truncated out of band, users lose data. Should be `logger.warn`. Gate to `debug` only when `IS_TESTING`.

### I5. `deleteOldTraces` discards `rowsAffected`

**`src/tracing/store.ts:393-406`**

The transaction runs two deletes inside a transaction but throws away both `rowsAffected` values. The success log `"Successfully deleted traces older than N days"` fires even when zero rows were affected. If the dual-encoding `traceCreatedBefore` predicate ever misclassifies timestamps, retention silently does nothing. Capture both counts and warn when they're zero against a non-empty table.

### I6. `recordBlobReference` check-then-insert is not transactional

**`src/blobs/index.ts:140-166`**

`SELECT existing` followed by `INSERT` is not wrapped in a transaction; `blobReferencesTable` has no unique constraint on `(blob_hash, eval_id)`. Concurrent calls for the same hash+evalId both observe "no existing" and both insert. Pre-existing pattern but the migration was the natural moment to fix it. Wrap in `db.transaction(...)` or add `.onConflictDoNothing()` + a unique index.

### I7. `provider.exists(hash).catch(() => false)` swallows real provider errors

**`src/blobs/index.ts:129`**

If the filesystem provider throws `EACCES`, a timeout fires, or a future S3 provider returns 5xx, this silently records "blob doesn't exist" and skips the reference insert. No warning logged. Catch and rethrow or at minimum `logger.warn('[BlobStorage] provider.exists check failed', { hash, error })`.

### I8. Public write helpers don't document the serialization invariant

**`src/util/database.ts:37-155`**

After this PR, `Promise.all([writeA, writeB])` no longer parallelizes (all top-level statements and `db.transaction` calls queue through one Promise). The "why" comment is at `src/database/index.ts:98-104` but nothing surfaces this to callers. Add a one-line JSDoc on `writeResultsToDatabase`, `updateResult`, `deleteEval*`, noting "serialized against sibling top-level writes."

### I9. Retry post-success cleanup catch swallows `recalculatePromptMetrics` too coarsely

**`src/commands/retry.ts:342-350`**

```ts
try {
  await deleteErrorResults(errorResultIds);
  await recalculatePromptMetrics(retriedEval);
} catch (cleanupError) {
  logger.warn('Post-retry cleanup had issues. Retry results are saved.', { error: cleanupError });
}
```

`deleteErrorResults` failure is cosmetic (dead rows). `recalculatePromptMetrics` failure is correctness (the UI shows wrong pass/fail counts permanently). Both get the same generic "had issues" warning. Split the try/catch so the metrics failure surfaces with explicit recovery guidance.

## Test coverage gaps (Important)

| Gap | File | Why it matters |
|---|---|---|
| No test for `getTraceSpans` free function | `src/tracing/store.ts:507` | C1 would have been caught |
| No test for root db call inside transaction | `test/database/index.test.ts` | C2 would have been caught |
| No test for `not_contains` filter on json_each rewrite | `src/models/eval.ts:942` | Subtle semantic divergence on NULL/missing rows |
| No test for `Eval.getMany()` downstream consumers | `src/util/database.ts:202,267` | `getPromptsForTestCasesHash` etc. only mocked |
| libsql binding handler test covers CJS only | `test/util/libsqlBindingErrors.test.ts:7` | I1(b) untested |
| `writeResultsToDatabase` rollback only covers `evals_to_tags` | `test/util/databaseWrite.test.ts:27-33` | Mid-loop failures (prompts/datasets) untested |
| Concurrent transaction stress shallow | `test/database/index.test.ts:95` | Only 2 parallel txs |

## Positive observations

- 9486 makes a **better** choice than 9009 on `Eval.getMany()` column projection (`src/models/eval.ts:397-423`): only selects the columns the constructor actually consumes, so the return type doesn't lie about what's hydrated. This is cleaner than 9009's approach.
- The commit history shows discipline: an early `node:sqlite` attempt was reverted cleanly, no stale references remain (`grep` confirms).
- Comprehensive WAL config, in-memory test mode, and transaction wrapping tests.
- Lazy dynamic imports in `getDb()` are the right pattern for surfacing platform-binding errors with a friendly message.
- The real-listening-server pattern in `test/server/eval.test.ts:18-26` and `test/server/routes/modelAudit.test.ts:41-48` is a real upgrade over `supertest` auto-listen.
- `test/util/databaseWrite.test.ts` is a clean regression test for the `Promise.all` → `transaction` conversion.
- Engines preserved at `^20.20.0 || >=22.22.0` (libsql works on Node 20, unlike `node:sqlite` which requires 22+).
- Eval.getMany column trim is genuinely a perf win.

## Recommendation

This PR is structurally close to mergeable but **C1, C2, and C3 are real bugs that should block merge**. C1 will throw on every call to `getTraceSpans`. C2 is a structural footgun that will eventually deadlock production. C3 is a silent data-loss path on the eval-update HTTP route.

If you want to ship this PR rather than #9009, the cleanest path is to cherry-pick the two fix commits from #9009: [`cf479dac80`](https://github.com/promptfoo/promptfoo/pull/9009/commits/cf479dac80) (P2 fix + regression test) and [`485ceff33b`](https://github.com/promptfoo/promptfoo/pull/9009/commits/485ceff33b) (P1 fix + regression test). C3 needs a separate fix.

Otherwise the PRs are functionally equivalent; differences are mostly cosmetic (cleaner `Eval.getMany` projection in 9486, more thorough rebase + extra test coverage in 9009).

mldangelo:
Follow-up review of `d396c503f5` ("fix(db): address libsql review blockers"). Re-verified against the new branch tip.

## Cleared

**C1 — `getTraceSpans()` crash.** Replaced the broken 70-line wrapper with the one-line `return getTraceStore().getSpans(traceId, options)`. Side benefit confirmed: the public method passes raw (pre-sanitize) attributes to `deriveSpanKind`, which is the correct classification path. Regression test added in `test/tracing/storeRetention.test.ts`. ✅

**C2 — root db.* deadlock inside transaction.** Added the `if (activeTransaction.getStore()) return method(...args)` short-circuit in `serializeClientMethod`. Regression test added in `test/database/index.test.ts` covering exactly the `db.transaction(async tx => { await tx.run(...); await db.run('SELECT 1'); })` pattern. ✅

**C3 — silent PATCH route data loss.** `updateResult` now `throw err` after logging, and the route handler's `catch (error)` logs with `[PATCH /api/eval/:id] Failed to update eval`. Regression test added in the new `test/server/routes/eval.validation.test.ts` mocking a `save()` rejection and asserting 500 + error body. ✅

CI: 36/37 pass, 1 still pending at the time of this comment.

## Still applies from the prior review

These are unchanged in `d396c503f5` and worth addressing before merge:

**Important**
- **I1** libsql binding handler — regex `/@libsql\/(?<target>[a-z0-9-]+)/i` still matches `@libsql/client` and `@libsql/core` (misleading "binding for 'client' is missing"); still only matches CJS `MODULE_NOT_FOUND` while the lazy imports in `getDb()` are ESM and throw `ERR_MODULE_NOT_FOUND`. `src/util/libsqlBindingErrors.ts:3,18`
- **I2** WAL-mode test became self-referential after `eeaeef6a38` swapped to `@libsql/client/node`. If a future libsql version silently downgrades `journal_mode = WAL`, both writer and verifier agree. `test/database.test.ts:57-77`
- **I3** `closeDb()` race with in-flight `getDb()`. `src/database/index.ts:163-189`
- **I4** WAL checkpoint failure in `closeDb()` logged at `debug` — should be `warn`. `src/database/index.ts:172`
- **I5** `deleteOldTraces` discards `rowsAffected`. `src/tracing/store.ts:393-406`
- **I6** `recordBlobReference` check-then-insert race. `src/blobs/index.ts:140-166`
- **I7** `provider.exists(hash).catch(() => false)` swallows real provider errors. `src/blobs/index.ts:129`
- **I8** Public write helpers don't document the serialization invariant. `src/util/database.ts:37-155`
- **I9** Retry post-success cleanup catch swallows `recalculatePromptMetrics` too coarsely. `src/commands/retry.ts:342-350`

**Test coverage**
- `not_contains` filter rewrite — zero coverage
- `Eval.getMany()` downstream consumers — only mocked
- ESM `ERR_MODULE_NOT_FOUND` path in libsql binding handler — uncovered
- `writeResultsToDatabase` rollback — only `evals_to_tags` table covered, mid-loop failure paths untested
- Concurrent transaction stress — only 2 parallel txs

None of these are merge-blockers individually, but **I1**, **I2**, and the `not_contains` coverage gap are the highest-value follow-ups.

Nice work on the 3 critical fixes. Holding off on approval pending the items above and the parallel #9009 review.

mldangelo:
Comprehensive perf comparison against `origin/main` and the parallel PR #9009, with all three branches built and timed side-by-side. Methodology, raw outputs, and scripts at the bottom.

## Summary

| Workload | main | 9486 | 9009 | Verdict |
|---|---:|---:|---:|---|
| Cold-start CLI (`--version`/`--help`/`eval --help`) | 761-793 ms | +0-2% | +0-3% | **tied** — lazy imports keep libsql off this path |
| Small eval (10 echo cases, c=1) | 996 ms | 970 | 991 | **tied** |
| Large eval (200 echo cases, c=20) | 1156 ms | 1203 (+4%) | 1215 (+5%) | main slightly faster |
| **`Eval.getMany(100)` w/ 200 rows × 500 KB legacy V2 blobs** | **45.66 ms** | **7.34 ms (6.22× faster)** | **7.25 ms (6.30× faster)** | **claim confirmed** |
| `Eval.getMany(100)` (V3 path, no blob) | 0.61 ms | 0.88 (1.44× slower) | 0.82 (1.34× slower) | main faster (sub-ms either way) |
| `Eval.findById` + `toResultsFile` (heavy blob) | 0.59 ms | 0.65 (+10%) | 0.67 (+14%) | tied |
| `Eval.findById` + `toResultsFile` (V3) | 0.37 ms | 0.72 (1.95× slower) | 0.73 (1.97× slower) | main faster (sub-ms either way) |
| Import 500-result JSON export | 994 ms | 986 | 990 | **tied** |
| `TraceStore.deleteOldTraces` (2,000 traces) | **broken** ¹ | 3.66 ms | 3.74 ms | PRs fix a latent FK bug |

¹ `main`'s `deleteOldTraces` is a single-statement `delete from traces` relying on cascade. The spans-table FK has no `ON DELETE CASCADE`, so with `PRAGMA foreign_keys = ON` (which main also sets) it throws `SQLITE_CONSTRAINT_FOREIGNKEY` whenever traces have spans. Both PRs fix this with a two-step transactional delete.

## Headline findings

1. **The `Eval.getMany` perf claim is real and substantial.** ~6× speedup on the legacy V2 evals-list page. PR description's stated 6900 → 1280 ms (5.4×) is the same ballpark as what I measured here (45.66 → 7.34 ms = 6.22×). This is the page that loads in the web UI for anyone with old data.

2. **9486 and 9009 are statistically tied on every workload.** The small content differences between the two PRs (cleaner `Eval.getMany` projection in 9486 vs the post-review fixes in 9009) do not produce a measurable perf delta in any of the seven workloads.

3. **Two correctness bugs in main get fixed as side effects of this migration:**
   - `writeResultsToDatabase` previously did `Promise.all(...)` outside any transaction → partial writes possible. PRs wrap in `db.transaction(...)`.
   - `TraceStore.deleteOldTraces` on main throws `SQLITE_CONSTRAINT_FOREIGNKEY` in production whenever pruning needs to delete traces with spans. PRs fix with the two-step transactional delete.

4. **Routine V3 workloads pay a small async tax** — ~5% on a 200-case eval, 1.3-2× on tiny in-memory ops (still sub-millisecond). Below user-perceptible threshold for any realistic eval flow.

5. **Cold-start is unchanged.** The lazy `import('@libsql/client/node')` in `getDb()` keeps libsql out of the path for `--version`/`--help`-style commands.

## Hypothesis scoreboard

| Hypothesis | Result |
|---|---|
| H1: Cold-start CLI is slower on async | **refuted** |
| H2: Small writes are slower on async | **refuted at scale** |
| H3: `Eval.getMany()` gets a big speedup | **confirmed** (~6×) |
| H4: High-concurrency writes are slower | **confirmed**, modest (~5%) |
| H5: Import is faster (atomic vs `Promise.all`) | **refuted** — tied; the change is correctness, not perf |
| H6: 9009 ≈ 9486 on perf | **confirmed** |

## Methodology

- 3 worktrees from the same checkout, each built once with `npm ci && npm run build`.
- Cold-start, eval throughput, and import workloads measured with `hyperfine` (warmups + 5-10 runs, fresh `PROMPTFOO_CONFIG_DIR` between runs).
- `getMany`/`findById`/retention workloads measured with `tsx` microbenchmarks running directly against each repo's `src/` (after warmup, 30-50 iterations, median + p95).
- All on the same Mac (darwin 25.1.0, Node v24.16.0), no other heavy processes running.
- Each microbenchmark uses a fresh tmpdir as `PROMPTFOO_CONFIG_DIR` so the SQLite file starts empty.

Plan, scripts, and raw `hyperfine` JSON outputs available in my local workspace (`/tmp/pf-perf/`).

## Read

The PR achieves what it set out to: it eliminates the Node-ABI failure class **and** the headline perf win on the legacy V2 list page is genuine. The async tax on routine V3 operations is real but small enough to be invisible in practice, and both correctness bug fixes (atomic writes, transactional retention) are worth having on their own.

mldangelo (approved):
Approving on the strength of:

1. The three critical review findings from my earlier multi-agent pass were all fixed in `d396c503f5` with proper regression tests (`getTraceSpans` await, root-db-in-tx deadlock bypass, PATCH `/api/eval/:id` silent data loss).
2. Side-by-side perf comparison against `origin/main` confirms the headline `Eval.getMany()` claim (~6× faster on legacy V2 list pages) and shows no material regression on routine V3 paths.
3. CI is green across the full matrix.

The remaining Important findings from my earlier review (libsql binding handler ESM + over-match, WAL-test self-referentiality, `closeDb` race, `deleteOldTraces` rowsAffected, blob race, retry cleanup catch coarseness, serialization-invariant docs) are real but not merge-blocking — happy to file follow-ups or fold them into a subsequent PR.

Note: PR #9009 is the parallel implementation by the same author of this comment. The two PRs are functionally equivalent; pick whichever lands first and close the other.

PR #9491, opened 2026-05-28
feat(providers): add NVIDIA NIM provider

NishchayMahor:
## Summary

Adds a first-class `nvidia:` provider for [NVIDIA's hosted inference API](https://build.nvidia.com) at `https://integrate.api.nvidia.com/v1`. The endpoint is OpenAI-compatible and exposes a large catalogue of models (Llama, Qwen, Nemotron, Mistral, DeepSeek, Phi…), so until now promptfoo users had to route through the generic `openai:` provider with a custom `apiBaseUrl`. That works, but it loses the prefix-routing ergonomics and makes evals less self-documenting.

I hit this while comparing a few NIM-hosted models for an internal eval and figured a thin dedicated provider would be useful upstream.

## Implementation

\`NvidiaProvider\` extends \`OpenAiChatCompletionProvider\` and only sets two defaults:

- \`apiBaseUrl\` → \`https://integrate.api.nvidia.com/v1\`
- \`apiKeyEnvar\` → \`NVIDIA_API_KEY\`

Both remain overrideable for self-hosted NIM and corporate proxies. All other behaviour (tool calling, streaming, JSON mode, retries) is inherited from the OpenAI chat provider.

\`\`\`yaml
providers:
  - nvidia:meta/llama-3.3-70b-instruct
  - nvidia:nvidia/llama-3.1-nemotron-70b-instruct
  - nvidia:qwen/qwen2.5-coder-32b-instruct
\`\`\`

## Checklist (per src/providers/AGENTS.md)

- [x] Implement \`ApiProvider\` interface → \`src/providers/nvidia/chat.ts\`
- [x] Env vars in \`src/contracts/env.ts\` (\`NVIDIA_API_KEY\`, \`NVIDIA_API_BASE_URL\`)
- [x] Env vars in \`src/envars.ts\`
- [x] Provider tests in \`test/providers/nvidia.test.ts\` (11 tests, all HTTP mocked)
- [x] Docs in \`site/docs/providers/nvidia.md\` (covers setup, key acquisition, free-credits note, configuration, override patterns, common model ids, example)
- [x] Entry in \`site/docs/providers/index.md\` (alphabetical, between Nscale and OpenClaw)
- [x] Example in \`examples/provider-nvidia/\` with README + config
- [x] JSON schema regenerated via \`npm run jsonSchema:generate\`

## Local verification

\`\`\`
$ npx vitest run test/providers/nvidia.test.ts
Test Files  1 passed (1)
     Tests  11 passed (11)

$ npx vitest run test/providers/registry.test.ts
Test Files  1 passed (1)
     Tests  35 passed (35)

$ npx biome check src/providers/nvidia/ test/providers/nvidia.test.ts src/providers/registry.ts src/envars.ts src/contracts/env.ts
Checked 5 files in 994ms. No fixes applied.
(1 pre-existing complexity warning on registry.ts, present on main without this change.)

$ npx prettier --check site/docs/providers/nvidia.md site/docs/providers/index.md examples/provider-nvidia/
All matched files use Prettier code style!
\`\`\`

## Notes

- Cost calculation is intentionally not built in — NIM uses credit-based billing for many models, and per-token public price lists vary by tier. Users can set \`inputCost\`/\`outputCost\` on the provider config if they want an estimate in eval output. Mentioned in the docs.
- The provider is intentionally minimal (~60 lines). No model registry, no special-case parsing — anything NVIDIA exposes through the OpenAI-compatible endpoint works via \`nvidia:<full-model-id>\`.


NishchayMahor:
Thanks @chatgpt-codex-connector — addressed all four in 39f6a14:

- **P1 OpenAI key fallback** — overrode `getApiKey()` to only read `apiKey` / `NVIDIA_API_KEY` (or the configured `apiKeyEnvar`) and never fall through to `OPENAI_API_KEY`.
- **P2 `NVIDIA_API_BASE_URL` ignored** — constructor now reads it from `providerOptions.env` and `getEnvString` before falling back to the default base URL.
- **P2 `OpenAI-Organization` leak** — overrode `getOrganization()` to return `undefined` so `OPENAI_ORGANIZATION` can't leak onto NIM requests.
- **P2 example needed OPENAI_API_KEY** — swapped the `llm-rubric` assertions in `examples/provider-nvidia/promptfooconfig.yaml` for `icontains` / `icontains-any` so the example runs end-to-end with only `NVIDIA_API_KEY`.

Added 4 regression tests covering each path — full test file is now 15/15 green.

mldangelo:
@NishchayMahor pushed two follow-up commits ([b0f3388](https://github.com/promptfoo/promptfoo/pull/9491/commits/b0f33883ff), [ca67afa](https://github.com/promptfoo/promptfoo/pull/9491/commits/ca67afa848)) to address the two remaining Codex P2s that were flagged on the latest review pass:

**1. `fix(providers):` — `getApiUrl()` override + registry test**

The inherited `OpenAiGenericProvider.getApiUrl()` consults `OPENAI_API_HOST` / `OPENAI_API_BASE_URL` / `OPENAI_BASE_URL` ahead of `config.apiBaseUrl`, so an environment with any of those set for an OpenAI-compatible host would silently redirect `nvidia:*` traffic there even though the constructor initialized `config.apiBaseUrl` to NVIDIA's default. Override `getApiUrl()` to return only `config.apiBaseUrl || NVIDIA_NIM_API_BASE_URL` — the constructor already merged `NVIDIA_API_BASE_URL` into `config.apiBaseUrl`, so this preserves the documented env-override chain while ignoring the OpenAI-specific ones. Same hardening pattern as ``OrcaRouterProvider.getApiUrl()`` in #9493.

Added a regression test (`does not route through OPENAI_API_HOST / OPENAI_API_BASE_URL / OPENAI_BASE_URL`) that sets all three OpenAI host envs and asserts the resolved URL is still NVIDIA's.

Also added an `nvidia:` prefix dispatch test in `test/providers/registry.test.ts` per the convention called out in `src/providers/AGENTS.md` — verifies `nvidia:meta/llama-3.3-70b-instruct` resolves to `NvidiaProvider` and that a bare `nvidia:` throws the expected error.

**2. `docs(site):` — deterministic assertions in the docs example**

The Example block in `site/docs/providers/nvidia.md` still used `llm-rubric` even though the example fixture was switched to `icontains` in 39f6a14. Model-graded assertions fall back to promptfoo's default OpenAI grader, so the documented example would have failed for users who set up only `NVIDIA_API_KEY` per the Setup section above it. Swapped for `icontains` / `icontains-any` to match the runnable fixture, and added a short follow-up snippet showing how to wire `defaultTest.options.provider` at a NIM-hosted grader for users who want a rubric-style assertion.

**Verified locally**
- `npx vitest run test/providers/nvidia.test.ts test/providers/registry.test.ts` → 52 passed (16 nvidia + 36 registry)
- `npx tsc --noEmit` → clean
- `npx prettier --check site/docs/providers/nvidia.md` → clean

mldangelo:
CI is green across the board except for **one pre-existing flake** unrelated to this PR:

- `Test on Node 22.22 and macOS-latest` failed on `test/commands/import.test.ts > collision handling > should keep the existing eval when a --force replacement fails preflight` with an underlying libsql query error (\`Failed query: select ... from "evals" where "evals"."id" = ?\`), not the expected \`Embedded blob hash mismatch\` precondition.

The flake came in with the libsql migration in [#9486](https://github.com/promptfoo/promptfoo/pull/9486) (`fix(db): replace better-sqlite3 with libsql`), which is already on `main`. The same `import.test.ts` collision-handling block flaked on `main` itself in Windows shard 3 of run [26551820502](https://github.com/promptfoo/promptfoo/actions/runs/26551820502) (`should allow import with --new-id flag when eval already exists`), so this isn't a NVIDIA regression — it's a cross-platform libsql state-isolation issue in the import command tests.

I re-ran the failed macOS job to confirm it passes on retry. Worth tracking separately as a flaky-test follow-up against `import.test.ts`.

PR otherwise has 29/30 effective passes (1 pre-existing flake, 2 intentional skips).

NishchayMahor:
Thanks for the polish on the cost tracking and reserved-subtype guard, that's much cleaner than what I had. Ready when you want to ship it. Happy to rebase if main moves further.

mldangelo-oai:
@NishchayMahor I am about to merge this and release it!

mldangelo:
Thanks @NishchayMahor — clean implementation, and the credential-isolation pattern (no `OPENAI_API_KEY` / `OPENAI_ORGANIZATION` / `OPENAI_API_HOST` leak) is exactly the shape we want for a new OpenAI-compatible provider. We modeled the NVIDIA hardening on this.

Pushed a few follow-ups during review:

- Hardened the chat integration with a reserved-subtype guard (`embedding` / `image` / `moderation` / etc. now throw with a clear error rather than silently routing to chat) and a `calculateNvidiaCost` helper so explicit `inputCost` / `outputCost` / `cost` overrides surface in eval output.
- Corrected the example `README.md` claim to match the deterministic `icontains` / `icontains-any` assertions (so the example runs end-to-end with only `NVIDIA_API_KEY`).
- Merged `origin/main` to sync the libsql DB migration and resolved a tiny `registry.test.ts` overlap with the OrcaRouter PR.

CI is green across Node 20.20 / 22.22 / 24.x / 26.x on ubuntu, macOS, and Windows shards 1–3. Squash-merging shortly.

PR #9496, opened 2026-05-28
fix(eval): refresh history cache after inserts

wholley-oai:
## Summary

The Web UI history endpoint caches standalone eval rows for up to two hours. Recent cache invalidation work handled eval updates and deletes, but eval inserts could still leave `/api/history` stale after the first history read.

This fixes the insert side of that behavior:

- Move the standalone history cache into a small shared helper module.
- Clear that cache after successful `Eval.create()` inserts.
- Clear that cache after successful `writeResultsToDatabase()` imports/writes.
- Keep `clearStandaloneEvalCache()` exported from `src/util/database.ts` for existing tests/callers.

## Why

Users expect a newly created or imported eval to appear in history immediately. Before this change, if history was read first, a later eval insert could be hidden until the cache expired or another mutation happened to clear the cache.

## Tests

- `IS_TESTING=true node_modules/.bin/vitest run test/util/databaseStandalone.test.ts test/util/databaseWrite.test.ts test/models/eval.test.ts`
- `/Users/wholley/.nvm/versions/node/v24.16.0/bin/npm run tsc`
- `/Users/wholley/.nvm/versions/node/v24.16.0/bin/npm run f`
- `/Users/wholley/.nvm/versions/node/v24.16.0/bin/npm run l`

`npm run f` and `npm run l` report the existing `src/models/eval.ts` cognitive-complexity warning at line 851, with exit 0.

### DIFF
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -42,6 +42,7 @@ import { convertTestResultsToTableRow } from '../util/exportToFile/index';
 import { isNonTransientHttpStatus, NON_TRANSIENT_HTTP_STATUSES } from '../util/fetch/errors';
 import invariant from '../util/invariant';
 import { sanitizeRuntimeOptions } from '../util/sanitizer';
+import { clearStandaloneEvalCache } from '../util/standaloneEvalCache';
 import { getCurrentTimestamp } from '../util/time';
 import { accumulateTokenUsage, createEmptyTokenUsage } from '../util/tokenUsageUtils';
 import {
@@ -557,6 +558,8 @@ export default class Eval {
       }
     });
 
+    clearStandaloneEvalCache();
+
     return new Eval(config, {
       id: evalId,
       author,
@@ -658,6 +661,7 @@ export default class Eval {
       updateObj.results = expr;
     }
     await db.update(evalsTable).set(updateObj).where(eq(evalsTable.id, this.id)).run();
+    clearStandaloneEvalCache();
     this.persisted = true;
   }
 
--- a/src/models/eval.ts
+++ b/src/models/eval.ts
@@ -1261,6 +1261,7 @@ export default class Eval {
       // Notify the view server after prompt metadata changes so cached /api/prompts
       // responses and socket listeners can pick up prompts added after eval creation.
       updateSignalFile(this.id);
+      clearStandaloneEvalCache();
     }
   }
 
@@ -1272,6 +1273,7 @@ export default class Eval {
         .insert(evalResultsTable)
         .values(results.map((r) => ({ ...r, evalId: this.id })))
         .run();
+      clearStandaloneEvalCache();
     }
     this._resultsLoaded = true;
   }
@@ -1587,6 +1589,8 @@ export default class Eval {
         });
       }
     });
+
+    clearStandaloneEvalCache();
 
     logger.info('Eval copy completed successfully', {
       sourceEvalId: this.id,
--- a/src/util/database.ts
+++ b/src/util/database.ts
@@ -1,5 +1,4 @@
 import { and, desc, eq, inArray } from 'drizzle-orm';
-import { LRUCache } from 'lru-cache';
 import { DEFAULT_QUERY_LIMIT } from '../constants';
 import { deleteTraceRecordsForEvals } from '../database/evalDeletion';
 import { getDb } from '../database/index';
@@ -20,7 +19,6 @@ import logger from '../logger';
 import Eval, { createEvalId } from '../models/eval';
 import { generateIdFromPrompt } from '../models/prompt';
 import {
-  type CompletedPrompt,
   type EvaluateSummaryV2,
   type EvaluateTable,
   type EvalWithMetadata,
@@ -33,6 +31,14 @@ import {
 import invariant from '../util/invariant';
 import { sha256 } from './createHash';
 import { restoreAzureBlobSasTokens } from './sanitizer';
+import {
+  type StandaloneEval as CachedStandaloneEval,
+  clearStandaloneEvalCache as clearStandaloneEvalCacheImpl,
+  getCachedStandaloneEvals,
+  setCachedStandaloneEvals,
+} from './standaloneEvalCache';
+
+export type StandaloneEval = CachedStandaloneEval;
 
 export async function writeResultsToDatabase(
   results: EvaluateSummaryV2,
@@ -151,6 +157,8 @@ export async function writeResultsToDatabase(
     }
   });
 
+  clearStandaloneEvalCacheImpl();
+
   return evalId;
 }
 
@@ -192,7 +200,7 @@ export async function updateResult(
     }
 
     await existingEval.save();
-    clearStandaloneEvalCache();
+    clearStandaloneEvalCacheImpl();
 
     logger.info(`Updated eval with ID ${id}`);
   } catch (err) {
@@ -446,7 +454,7 @@ export async function deleteEval(evalId: string) {
       throw new Error(`Eval with ID ${evalId} not found`);
     }
   });
-  clearStandaloneEvalCache();
+  clearStandaloneEvalCacheImpl();
 }
 
 /**
@@ -463,7 +471,7 @@ export async function deleteEvals(ids: string[]): Promise<void> {
     await tx.delete(evalResultsTable).where(inArray(evalResultsTable.evalId, ids)).run();
     await tx.delete(evalsTable).where(inArray(evalsTable.id, ids)).run();
   });
-  clearStandaloneEvalCache();
+  clearStandaloneEvalCacheImpl();
 }
 
 /**
@@ -482,38 +490,17 @@ export async function deleteAllEvals(): Promise<void> {
     await tx.delete(evalsToTagsTable).run();
     await tx.delete(evalsTable).run();
   });
-  clearStandaloneEvalCache();
+  clearStandaloneEvalCacheImpl();
 }
-
-export type StandaloneEval = CompletedPrompt & {
-  evalId: string;
-  description: string | null;
-  datasetId: string | null;
-  promptId: string | null;
-  isRedteam: boolean;
-  createdAt: number;
-
-  pluginFailCount: Record<string, number>;
-  pluginPassCount: Record<string, number>;
-  uuid: string;
-};
-
-const standaloneEvalCache = new LRUCache<string, StandaloneEval[]>({
-  ttl: 60 * 60 * 2 * 1000, // 2 hours in milliseconds
-  // Cache entries are keyed by (limit, tag, description) filter combinations.
-  // 2000 handles heavy automation scenarios while keeping memory bounded (~few MB).
-  // On eviction, the next request simply re-queries the DB with minimal latency impact.
-  max: 2000,
-});
 
 /**
  * Invalidates the standalone eval LRU cache backing `getStandaloneEvals()`.
- * Called by mutation paths in this module (update, delete) so `/api/history` reflects
+ * Called by mutation paths in this module (write, update, delete) so `/api/history` reflects
  * changes immediately instead of waiting up to the 2-hour TTL. Also re-exported for
  * tests that need to assert behavior across cache boundaries.
  */
 export function clearStandaloneEvalCache(): void {
-  standaloneEvalCache.clear();
+  clearStandaloneEvalCacheImpl();
 }
 
 export async function getStandaloneEvals({
@@ -526,7 +513,7 @@ export async function getStandaloneEvals({
   description?: string;
 } = {}): Promise<StandaloneEval[]> {
   const cacheKey = `standalone_evals_${limit}_${tag?.key}_${tag?.value}_${description}`;
-  const cachedResult = standaloneEvalCache.get(cacheKey);
+  const cachedResult = getCachedStandaloneEvals(cacheKey);
 
   if (cachedResult) {
     return cachedResult;
@@ -620,6 +607,6 @@ export async function getStandaloneEvals({
     uuid: crypto.randomUUID(),
   }));
 
-  standaloneEvalCache.set(cacheKey, withUUIDs);
+  setCachedStandaloneEvals(cacheKey, withUUIDs);
   return withUUIDs;
 }
--- a/src/util/database.ts
+++ b/src/util/database.ts
@@ -32,13 +32,16 @@ import invariant from '../util/invariant';
 import { sha256 } from './createHash';
 import { restoreAzureBlobSasTokens } from './sanitizer';
 import {
-  type StandaloneEval as CachedStandaloneEval,
-  clearStandaloneEvalCache as clearStandaloneEvalCacheImpl,
+  clearStandaloneEvalCache,
   getCachedStandaloneEvals,
+  getStandaloneEvalCacheKey,
   setCachedStandaloneEvals,
 } from './standaloneEvalCache';
 
-export type StandaloneEval = CachedStandaloneEval;
+import type { StandaloneEval } from './standaloneEvalCache';
+
+export type { StandaloneEval };
+export { clearStandaloneEvalCache };
 
 export async function writeResultsToDatabase(
   results: EvaluateSummaryV2,
@@ -157,7 +160,7 @@ export async function writeResultsToDatabase(
     }
   });
 
-  clearStandaloneEvalCacheImpl();
+  clearStandaloneEvalCache();
 
   return evalId;
 }
@@ -453,7 +456,7 @@ export async function deleteEval(evalId: string) {
       throw new Error(`Eval with ID ${evalId} not found`);
     }
   });
-  clearStandaloneEvalCacheImpl();
+  clearStandaloneEvalCache();
 }
 
 /**
@@ -470,7 +473,7 @@ export async function deleteEvals(ids: string[]): Promise<void> {
     await tx.delete(evalResultsTable).where(inArray(evalResultsTable.evalId, ids)).run();
     await tx.delete(evalsTable).where(inArray(evalsTable.id, ids)).run();
   });
-  clearStandaloneEvalCacheImpl();
+  clearStandaloneEvalCache();
 }
 
 /**
@@ -489,17 +492,7 @@ export async function deleteAllEvals(): Promise<void> {
     await tx.delete(evalsToTagsTable).run();
     await tx.delete(evalsTable).run();
   });
-  clearStandaloneEvalCacheImpl();
-}
-
-/**
- * Invalidates the standalone eval LRU cache backing `getStandaloneEvals()`.
- * Called by mutation paths in this module (write, update, delete) so `/api/history` reflects
- * changes immediately instead of waiting up to the 2-hour TTL. Also re-exported for
- * tests that need to assert behavior across cache boundaries.
- */
-export function clearStandaloneEvalCache(): void {
-  clearStandaloneEvalCacheImpl();
+  clearStandaloneEvalCache();
 }
 
 export async function getStandaloneEvals({
@@ -511,7 +504,7 @@ export async function getStandaloneEvals({
   tag?: { key: string; value: string };
   description?: string;
 } = {}): Promise<StandaloneEval[]> {
-  const cacheKey = `standalone_evals_${limit}_${tag?.key}_${tag?.value}_${description}`;
+  const cacheKey = getStandaloneEvalCacheKey({ limit, tag, description });
   const cachedResult = getCachedStandaloneEvals(cacheKey);
 
   if (cachedResult) {
--- a/src/util/database.ts
+++ b/src/util/database.ts
@@ -200,7 +200,6 @@ export async function updateResult(
     }
 
     await existingEval.save();
-    clearStandaloneEvalCacheImpl();
 
     logger.info(`Updated eval with ID ${id}`);
   } catch (err) {
--- a/src/util/standaloneEvalCache.ts
+++ b/src/util/standaloneEvalCache.ts
@@ -1,0 +1,36 @@
+import { LRUCache } from 'lru-cache';
+
+import type { CompletedPrompt } from '../types/index';
+
+export type StandaloneEval = CompletedPrompt & {
+  evalId: string;
+  description: string | null;
+  datasetId: string | null;
+  promptId: string | null;
+  isRedteam: boolean;
+  createdAt: number;
+
+  pluginFailCount: Record<string, number>;
+  pluginPassCount: Record<string, number>;
+  uuid: string;
+};
+
+const standaloneEvalCache = new LRUCache<string, StandaloneEval[]>({
+  ttl: 60 * 60 * 2 * 1000, // 2 hours in milliseconds
+  // Cache entries are keyed by (limit, tag, description) filter combinations.
+  // 2000 handles heavy automation scenarios while keeping memory bounded (~few MB).
+  // On eviction, the next request simply re-queries the DB with minimal latency impact.
+  max: 2000,
+});
+
+export function getCachedStandaloneEvals(cacheKey: string): StandaloneEval[] | undefined {
+  return standaloneEvalCache.get(cacheKey);
+}
+
+export function setCachedStandaloneEvals(cacheKey: string, evals: StandaloneEval[]): void {
+  standaloneEvalCache.set(cacheKey, evals);
+}
+
+export function clearStandaloneEvalCache(): void {
+  standaloneEvalCache.clear();
+}
--- a/src/util/standaloneEvalCache.ts
+++ b/src/util/standaloneEvalCache.ts
@@ -1,4 +1,5 @@
 import { LRUCache } from 'lru-cache';
+import { DEFAULT_QUERY_LIMIT } from '../constants';
 
 import type { CompletedPrompt } from '../types/index';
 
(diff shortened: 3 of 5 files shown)


src/util/database.ts
```diff
@@ -192,7 +200,7 @@ export async function updateResult(
     }
 
     await existingEval.save();
-    clearStandaloneEvalCache();
+    clearStandaloneEvalCacheImpl();
```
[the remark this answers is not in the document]
wholley-oai: Fixed in 5151f33fb: `updateResult()` now relies on `Eval.save()` for the cache invalidation, so the update path clears the standalone history cache once.

src/util/database.ts
```diff
@@ -33,6 +31,14 @@ import {
 import invariant from '../util/invariant';
 import { sha256 } from './createHash';
 import { restoreAzureBlobSasTokens } from './sanitizer';
+import {
+  type StandaloneEval as CachedStandaloneEval,
+  clearStandaloneEvalCache as clearStandaloneEvalCacheImpl,
```
mldangelo: Nit: `clearStandaloneEvalCacheImpl` leaks an `Impl` suffix into every internal call site (lines 160, 456, 473, 492). The alias exists only because line 502 re-exports a same-named wrapper. Two cleaner options:

1. Drop the wrapper at 501-503 and replace it with `export { clearStandaloneEvalCache } from './standaloneEvalCache';` — preserves the existing public API for `test/util/databaseStandalone.test.ts:8` and `test/util/databaseWrite.test.ts:6`.
2. Or keep the wrapper but call the imported function directly (no rename) — only the wrapper itself needs the local alias.

Either removes `Impl` from internal call sites.

src/util/database.ts
```diff
@@ -33,6 +31,14 @@ import {
...
+  getCachedStandaloneEvals,
+  setCachedStandaloneEvals,
+} from './standaloneEvalCache';
+
+export type StandaloneEval = CachedStandaloneEval;
```
mldangelo: The `type StandaloneEval as CachedStandaloneEval` import + `export type StandaloneEval = CachedStandaloneEval` round-trip exists for backward compat with `src/app/src/pages/history/{page,History,History.test}.tsx`, which import `StandaloneEval` from `@promptfoo/util/database`. Two options worth considering:

- Update the three frontend imports to `@promptfoo/util/standaloneEvalCache` and drop the alias entirely.
- Move `StandaloneEval` to `src/types/` (per `AGENTS.md` "core logic in `src/`" — it's a public data shape consumed by both backend and frontend, not really a cache implementation detail).

Not blocking, but the double-rename is awkward and exists purely to avoid touching three frontend imports.

src/util/database.ts
```diff
@@ -151,6 +157,8 @@ export async function writeResultsToDatabase(
     }
   });
 
+  clearStandaloneEvalCacheImpl();
```
mldangelo: Placement correct — outside the transaction so rollback doesn't false-invalidate. ✔

One small thing: the existing `rolls back related rows when a dependent insert fails` test at `test/util/databaseWrite.test.ts:30-46` doesn't assert anything about the cache. To pin this behavior, consider extending it to prime the cache, trigger the rollback, then assert that previously cached entries remain intact — proving the cache is *not* falsely invalidated on rollback. Without that, a future refactor that moves the clear inside the tx would silently change semantics.

src/util/database.ts
```diff
@@ -482,38 +489,17 @@ export async function deleteAllEvals(): Promise<void> {
...
-  // Cache entries are keyed by (limit, tag, description) filter combinations.
-  // 2000 handles heavy automation scenarios while keeping memory bounded (~few MB).
-  // On eviction, the next request simply re-queries the DB with minimal latency impact.
-  max: 2000,
-});
-
 /**
  * Invalidates the standalone eval LRU cache backing `getStandaloneEvals()`.
- * Called by mutation paths in this module (update, delete) so `/api/history` reflects
+ * Called by mutation paths in this module (write, update, delete) so `/api/history` reflects
```
mldangelo: Doc is stale post-5151f33fb. "Called by mutation paths **in this module** (write, update, delete)" — the function is also invoked from `src/models/eval.ts:561` and `:664` now (via the aliased import). And `updateResult`'s explicit clear was removed in 5151f33fb, so calling out `update` in *this module* is misleading.

Suggest:

> Called by mutation paths in `src/util/database.ts` (write, delete) and `src/models/eval.ts` (`Eval.create`, `Eval.save`) so `/api/history` reflects changes immediately…

src/util/database.ts
```diff
@@ -482,38 +489,17 @@ export async function deleteAllEvals(): Promise<void> {
...
  * changes immediately instead of waiting up to the 2-hour TTL. Also re-exported for
  * tests that need to assert behavior across cache boundaries.
  */
 export function clearStandaloneEvalCache(): void {
-  standaloneEvalCache.clear();
+  clearStandaloneEvalCacheImpl();
```
mldangelo: Ties into the comment on line 36: this wrapper exists to keep `clearStandaloneEvalCache` exported from `src/util/database.ts` for tests. A `export { clearStandaloneEvalCache } from './standaloneEvalCache';` re-export would do the same thing without the `Impl` indirection.

(Strictly, per `AGENTS.md` ("Avoid re-exporting from files; import directly from the source module"), the cleanest answer is for the tests at `test/util/databaseStandalone.test.ts:8` and `test/util/databaseWrite.test.ts:6` to import from `'../../src/util/standaloneEvalCache'` directly, and delete this wrapper plus the type alias on line 41.)

src/util/standaloneEvalCache.ts
```diff
@@ -0,0 +1,36 @@
...
+  isRedteam: boolean;
+  createdAt: number;
+
+  pluginFailCount: Record<string, number>;
+  pluginPassCount: Record<string, number>;
+  uuid: string;
+};
+
+const standaloneEvalCache = new LRUCache<string, StandaloneEval[]>({
+  ttl: 60 * 60 * 2 * 1000, // 2 hours in milliseconds
```
mldangelo, line 26: The `// 2 hours in milliseconds` TTL is now mostly nominal — every insert/update/delete clears the cache, so the TTL only protects a quiescent server (rare in practice). Worth one of:

- Reducing the TTL drastically (e.g., 5 minutes) since aggressive invalidation makes the long backstop unnecessary, and a shorter TTL limits the blast radius if a future caller forgets to invalidate.
- Keeping the TTL but updating the comment to reflect that it's a *backstop after invalidation*, not the primary freshness mechanism.

The current 2-hour TTL only matters in the bug scenarios this PR is trying to *prevent*, which feels backwards.

src/util/standaloneEvalCache.ts
```diff
@@ -0,0 +1,36 @@
...
+  // Cache entries are keyed by (limit, tag, description) filter combinations.
+  // 2000 handles heavy automation scenarios while keeping memory bounded (~few MB).
+  // On eviction, the next request simply re-queries the DB with minimal latency impact.
+  max: 2000,
+});
```
mldangelo, line 33: Optional API simplification: the two getter/setter exports plus `clear` are a small enough surface that exposing the `LRUCache` instance directly (`export const standaloneEvalCache = new LRUCache(...)`) would work fine and remove a layer. Or going the other direction, a single `withStandaloneEvalCache(key, computeFn)` helper would tighten the API and put the cache key shape next to its consumer.

Not blocking — the current shape is fine. Just noting the three small functions don't add much encapsulation beyond what `LRUCache` already provides.

src/models/eval.ts
```diff
@@ -557,6 +558,8 @@ export default class Eval {
       }
     });
 
+    clearStandaloneEvalCache();
```
mldangelo, line 561: Placement correct (after the transaction commits, so rollback doesn't false-invalidate). ✔

**But the same pattern is missing from `Eval.copy()` at line 1436** — see the summary's critical finding. `Eval.copy()` does the same kind of `db.transaction(...)`-wrapped insert into `evalsTable`/`evalResultsTable` (lines 1464, 1576) but never invalidates. The fix here should be replicated after the closing `});` of the copy transaction at line 1589.

**Also missing from `setResults()` at line 1267 and `addPrompts()` at line 1256** — both mutate columns surfaced by `getStandaloneEvals()` but skip the cache clear. Server surfaces: `POST /api/eval/:id/results` and the `evaluator.ts` `addPrompts` sites during eval runs. See the summary.

src/models/eval.ts
```diff
@@ -658,6 +661,7 @@ export default class Eval {
       updateObj.results = expr;
     }
     await db.update(evalsTable).set(updateObj).where(eq(evalsTable.id, this.id)).run();
+    clearStandaloneEvalCache();
```
mldangelo, line 664: Now that 5151f33fb removed the explicit `clearStandaloneEvalCacheImpl()` from `updateResult()`, this single line is the only thing keeping `updateResult` cache-coherent. The pre-existing `reflects flag transitions after updateResult rewrites config` test catches a regression here transitively, but a *direct* `eval_.save()` cache test would lock the contract independent of `updateResult`. Cheap to add — something like:

```ts
it('Eval.save() invalidates the standalone history cache', async () => {
  const eval_ = await createEvalWithPrompts({});
  await getStandaloneEvals(); // populate cache
  eval_.setDescription('changed'); // or any save-triggering mutation
  await eval_.save();
  const rows = await getStandaloneEvals();
  expect(rows.find(r => r.evalId === eval_.id)?.description).toBe('changed');
});
```

Minor positioning note: `this.persisted = true;` happens on the next line. If you ever add an early-return on the `update.run()` result, keep `clearStandaloneEvalCache()` after the actual write, not after persistence-flag bookkeeping.

mldangelo (commented):
## Review summary

I checked this out, ran the new tests, ran a curl-level reproducer against `npm run dev:server`, and audited every insert path into `evalsTable` in the repo. The fix works for the paths it covers (POST `/api/eval` v3/v4 and any `Eval.save()`/`writeResultsToDatabase`). I verified the two new tests do fail on `main` when only the production code is reverted, so they exercise the right path. The refactor into `src/util/standaloneEvalCache.ts` is well-motivated: `src/util/database.ts` already imports `Eval` from `src/models/eval.ts`, so any reverse import would create a module cycle.

**However — the PR misses at least three server-reachable mutations into the columns surfaced by `/api/history`. The same bug class the PR claims to fix is still latent on those paths.**

### 🚨 Critical: `Eval.copy()` is the same bug, still latent

`Eval.copy()` at `src/models/eval.ts:1436` inserts into `evalsTable` (line 1464) and `evalResultsTable` (line 1576) inside its own transaction (closes at line 1589), then returns without invalidating the standalone history cache. It is reachable from `POST /api/eval/:id/copy` (`src/server/routes/eval.ts:915-959`). After this PR, users who click "copy" in the Web UI will still see stale history for up to two hours.

Reproducer on this branch with `PROMPTFOO_CONFIG_DIR=$(mktemp -d) npm run dev:server` (fresh state, no cache outside what the script populates):

```
# Two evals already created via POST /api/eval; history cache populated
$ curl -s http://localhost:15500/api/history | jq '.data | length'
2
$ curl -s -X POST http://localhost:15500/api/eval/$SRC_ID/copy \
    -H 'Content-Type: application/json' -d '{"description":"copy probe"}'
{"id":"eval-Wcl-2026-05-28T15:50:35","distinctTestCount":0}
$ curl -s http://localhost:15500/api/history | jq '.data | length'
2   # <- copy is missing, exactly the bug the PR description rules out
```

**Fix:** add `clearStandaloneEvalCache();` between the closing `});` of the transaction at `src/models/eval.ts:1589` and the `logger.info('Eval copy completed successfully', ...)` call at 1591. Add a `Eval.copy()` regression test mirroring `includes newly created evals after history has been cached` — my probe (paste below) catches the gap.

```ts
it('Eval.copy() result appears in /api/history after cache populated', async () => {
  const source = await createEvalWithPrompts({});
  const beforeIds = (await getStandaloneEvals()).map((row) => row.evalId);
  expect(beforeIds).toContain(source.id);
  const copy = await source.copy('copy of source');
  const afterIds = (await getStandaloneEvals()).map((row) => row.evalId);
  expect(afterIds).toContain(source.id);
  expect(afterIds).toContain(copy.id);  // fails today
});
```

### ⚠️ Medium: `setResults()` and `addPrompts()` mutate columns surfaced by `getStandaloneEvals` without invalidating

- **`Eval.setResults()`** (`src/models/eval.ts:1267-1277`) inserts into `evalResultsTable`. Cached `StandaloneEval` rows derive `pluginPassCount` / `pluginFailCount` from `eval_.getTable().body` (`src/util/database.ts:557`), so `setResults` can stale plugin counts. Reachable from `POST /api/eval/:id/results` at `src/server/routes/eval.ts:608`.
- **`Eval.addPrompts()`** (`src/models/eval.ts:1256-1265`) writes the `prompts` column on `evalsTable` directly when `this.persisted`. The cache stores `eval_.getPrompts()` (`src/util/database.ts:572`). Server-side, `addPrompts` runs immediately after `Eval.create()` in v4 POST `/api/eval` (`src/server/routes/eval.ts:849`) — safe today because the preceding `Eval.create()` just cleared the cache. But during `evaluator.ts` runs (sites at 3585, 3714, 3746, 3773, 3800, 4185, 4417), the cache can be re-populated by a mid-run `/api/history` request and then staled by subsequent `addPrompts` calls until `finalizeEvaluation` reaches `evalRecord.save()` (line 4217), and only if `this.evalRecord.persisted`. Aborted/timed-out runs leak stale prompt metadata.

**Suggested fix shape:** add `clearStandaloneEvalCache()` to both methods, gated on `this.persisted` so transient in-memory mutations don't churn the cache. That puts the cache-coherence contract on the data-mutating method, not its callers — same principle this PR applied to `Eval.save()`.

### Lower-priority: CLI v2 import (`createImportedV2Eval`, `src/commands/import.ts:359-378`)

Fourth unmediated path into `evalsTable.insert()` (after `writeResultsToDatabase`, `Eval.create`, `Eval.copy`). The LRU cache is process-local so the CLI doesn't impact a running `dev:server` today, but this is the natural template a future contributor will copy. Either thread it through `writeResultsToDatabase` for consistency or add a one-line comment explaining why it intentionally diverges.

### Test coverage gaps

- **`Eval.save()` is only covered transitively** (after 5151f33fb removed the redundant clear from `updateResult`). The pre-existing `reflects flag transitions after updateResult rewrites config` test would catch a regression in `Eval.save()`'s clear *transitively*, but a direct `evalRecord.save()` cache test would lock the contract independent of `updateResult`. See inline at line 664.
- **Neither new test asserts the cache was populated** between the two inserts. If `getStandaloneEvals` silently stopped caching, both tests would still pass. Suggest `expect(getCachedStandaloneEvals(KEY)).toBeDefined()` between the inserts, or spying on DB query count.
- **No cache-key fan-out coverage.** Both tests only use default `getStandaloneEvals()` args. A future refactor that scoped `clear()` to a single key would silently regress. One assertion with `getStandaloneEvals({ limit: 5 })` would lock down full-flush.
- **Rollback test** (`test/util/databaseWrite.test.ts:30-46`) doesn't assert cache state. Cache clear sits outside the transaction (`src/util/database.ts:160`), so rollback leaves prior cached entries intact — that's arguably correct, but worth pinning so a future move-inside-tx refactor flags as a behavior change.
- **No `Eval.copy()` cache test** — per the critical finding above, this gap is the proximate cause of the latent bug.

### Nits / polish

See inline comments for: the `Impl` naming smell, the misleading 2-hour TTL comment, the stale doc comment on `clearStandaloneEvalCache`, the `StandaloneEval` type re-export round-trip, and the `databaseWrite` test name suggesting the wrong code path.

### Things I checked and don't think are issues

- **Transaction safety:** all three new clears sit *after* `await db.transaction(...)`. Rollback ⇒ throw ⇒ unreachable clear ⇒ no false invalidation. Correct.
- **Refactor necessity:** moving the cache out of `database.ts` is the right way to let `models/eval.ts` reach it without a cycle.
- **`LRUCache.clear()`** is sync and non-throwing — no try/catch needed.
- **Cache thrash:** `Eval.save()` fires once per `finalizeEvaluation` and once per rating submission — user-bounded, not hot loops. Fine.
- **Frontend impact:** preserved via the `StandaloneEval` re-export. (Whether the re-export is the *right* compat shim is the nit on line 41.)

### Bottom line

The fix is necessary, the chosen architecture (extract the cache module, invalidate at the data-mutating method) is correct, and the documented paths work end-to-end. But it stops one step short of solving the *bug class* — at minimum `Eval.copy()`, `setResults`, and `addPrompts` should also invalidate. If the latter two are out of scope, that's defensible, but `Eval.copy()` is the same user-facing bug the PR description rules out and should land here.

test/util/databaseStandalone.test.ts
```diff
@@ -110,6 +110,18 @@ describe('getStandaloneEvals', () => {
...
+    const first = await createEvalWithPrompts({});
+    const beforeIds = (await getStandaloneEvals()).map((row) => row.evalId);
+    expect(beforeIds).toContain(first.id);
+
+    const second = await createEvalWithPrompts({});
+
+    const afterIds = (await getStandaloneEvals()).map((row) => row.evalId);
+    expect(afterIds).toContain(first.id);
+    expect(afterIds).toContain(second.id);
+  });
```
mldangelo, line 217: Two tightening suggestions:

1. **The test doesn't actually verify the cache was hit.** If `getStandaloneEvals` silently stopped caching, this test would still pass — both calls would re-query the DB and both evals would appear. Add an assertion between the two creates that `getCachedStandaloneEvals('standalone_evals_100_undefined_undefined_undefined')` (or whatever the actual default key is) returns a value, or spy on DB query count. Same comment applies to the new test in `databaseWrite.test.ts`.

2. **No coverage of the cache-key fan-out.** The cache is keyed by `(limit, tag, description)`. A future refactor that scopes `clear()` to a single key would silently regress, since this test only uses default args. One additional `getStandaloneEvals({ limit: 5 })` call would lock down full-flush semantics.

3. **No `Eval.copy()` test** — per the summary, this is currently broken and should ship with a test in this PR.

test/util/databaseWrite.test.ts
```diff
@@ -48,4 +53,24 @@ describe('writeResultsToDatabase', () => {
     await expect(db.select().from(datasetsTable).all()).resolves.toHaveLength(0);
     await expect(db.select().from(tagsTable).all()).resolves.toHaveLength(0);
   });
+
+  it('includes imported evals after history has been cached', async () => {
```
mldangelo: Test name says **"imported"** evals, but `writeResultsToDatabase` is also the inline path for regular eval runs going through `POST /api/eval` v3 (`src/server/routes/eval.ts:829`). The reader will assume this only covers `promptfoo import`. Consider "newly written evals" or "newly persisted evals".

Same missing assertions noted in the `databaseStandalone.test.ts` comment — no proof the cache was actually populated between the two writes.

Minor: hardcoded timestamps `2024-01-01T00:00:00.000Z` and `00:00:01.000Z` work but are fragile under `createEvalId(date)` precision changes. `new Date()` and `new Date(Date.now()+1)` would be more robust.

src/util/database.ts
wholley-oai: Fixed in 21ba48fde: dropped the `CachedStandaloneEval` alias and re-export round-trip. `database.ts` now imports and exports `StandaloneEval` directly from `standaloneEvalCache`, preserving the existing public import path without the double rename.

src/util/database.ts
wholley-oai: Fixed in 21ba48fde by deleting the stale wrapper doc. The freshness note now lives with the TTL in `standaloneEvalCache.ts`, and the mutation call sites are direct.

src/util/database.ts
wholley-oai: Fixed in 21ba48fde: removed the `Impl` alias/wrapper path. `database.ts` now imports/re-exports `clearStandaloneEvalCache` directly and internal call sites call it by that name.

src/util/database.ts
wholley-oai: Added rollback cache coverage in `test/util/databaseWrite.test.ts`: it primes the history cache with a successful write, forces a later write to roll back, then asserts the previously cached eval remains cached.

src/util/standaloneEvalCache.ts
wholley-oai: Updated in 21ba48fde: the TTL comment now states that mutation paths clear the cache and the 2-hour TTL is a backstop for quiet servers or future missed invalidations. I kept the TTL unchanged to avoid mixing a behavior change into this fix.

src/util/database.ts
wholley-oai: Fixed in 21ba48fde: the wrapper is gone and `database.ts` uses a direct re-export for `clearStandaloneEvalCache`. The tests import cache helpers directly only for cache-state assertions.

src/models/eval.ts
wholley-oai: Fixed in 21ba48fde: `addPrompts()`, `setResults()`, and `copy()` now clear the standalone history cache after successful writes/transactions. Added focused stale-cache tests for all three paths.

src/util/standaloneEvalCache.ts
wholley-oai: Kept the getter/setter/clear surface for now to avoid exposing the mutable `LRUCache` instance. Added `getStandaloneEvalCacheKey()` so the key shape is centralized next to the cache helpers and tests do not duplicate it.

test/util/databaseStandalone.test.ts
wholley-oai: Added the three tightenings in 21ba48fde: cache-population assertions via `getCachedStandaloneEvals`, fan-out coverage with a second `{ limit: 5 }` cache key, and an `Eval.copy()` stale-cache regression test.

src/models/eval.ts
wholley-oai: Added direct `Eval.save()` cache coverage in `test/util/databaseStandalone.test.ts`. The test reloads the persisted eval, saves a description change after the history cache is primed, and verifies `getStandaloneEvals()` returns the new description.

test/util/databaseWrite.test.ts
wholley-oai: Updated in 21ba48fde: renamed the test to "newly persisted evals", added a cache-population assertion between writes, and replaced the fixed 2024 timestamps with dates derived from `new Date()`.

mldangelo (approved):
## Follow-up review on 21ba48f

`git pull`'d and re-audited. The new commit addresses **every critical and medium finding** from my prior review, plus most of the nits. Tests pass locally (14 tests across the two suites), CI is green (41/41 passing).

### Critical gaps closed ✅

- **`Eval.copy()` now invalidates the cache** (`src/models/eval.ts:1593`, post-transaction). Regression test added: `includes copied evals after history has been cached`. The HTTP-level reproducer I posted in the prior review no longer reproduces.
- **`Eval.setResults()` invalidates the cache** (`src/models/eval.ts:1276`), gated on `this.persisted && results.length > 0` — the extra `length > 0` guard is even cleaner than I suggested (won't fire on no-op). Regression test added: `includes result changes after history has been cached`.
- **`Eval.addPrompts()` invalidates the cache** (`src/models/eval.ts:1264`), gated on `this.persisted`. Regression test added: `includes prompt changes after history has been cached`.

### Nits closed ✅

- **`clearStandaloneEvalCacheImpl` naming smell**: gone — `database.ts` now imports `clearStandaloneEvalCache` directly and re-exports it via `export { clearStandaloneEvalCache };` (line 44). No more alias indirection at internal call sites.
- **`StandaloneEval` type re-export**: cleaner — `import type { StandaloneEval } from './standaloneEvalCache'; export type { StandaloneEval };` instead of the `as CachedStandaloneEval` round-trip.
- **TTL comment**: updated to call out backstop semantics (`src/util/standaloneEvalCache.ts:26-27`): "Mutation paths clear the cache; this TTL is a backstop for quiet servers and future missed invalidations." Nice.
- **Stale doc comment** on the old wrapper: deleted (the wrapper itself was removed).
- **Direct `Eval.save()` test**: added (`includes direct eval saves after history has been cached`).
- **No cache-hit assertion**: added — `expectStandaloneHistoryCached()` / `expectWrittenHistoryCached()` helpers actually call `getCachedStandaloneEvals(getStandaloneEvalCacheKey(...))` to verify the cache was populated between mutations.
- **Cache-key fan-out coverage**: test now exercises `getStandaloneEvals({ limit: 5 })` and asserts both default-key and limited-key results are invalidated.
- **Rollback test asserts cache state**: renamed to `keeps cached history when a later write rolls back` and now primes the cache, triggers rollback, asserts the prior cached entry survives. Locks the "cache clear sits outside the transaction" invariant.
- **Test name `imported evals` → `newly persisted evals`**: done.
- **Hardcoded timestamps**: replaced with `new Date()` and `new Date(firstDate.getTime() + 1000)`.

### Bonus polish I didn't ask for

- **`getStandaloneEvalCacheKey()` extracted into the cache module** (`src/util/standaloneEvalCache.ts:36-42`), with a typed `StandaloneEvalCacheKeyOptions`. Means prod and tests share the key shape — better than my suggestion of inlining the key string in the test. If the key schema ever changes, there's exactly one place to update.

### One residual nit (not blocking)

- **CLI v2 import (`createImportedV2Eval`, `src/commands/import.ts:359-378`)** still doesn't invalidate. As I noted in the prior review, this is defensible — the CLI runs out-of-process so it doesn't impact a running `dev:server`. Leaving it as-is is fine for this PR; just be aware it's the one remaining unmediated `db.insert(evalsTable)` if you ever consolidate paths.

### Verification

```
IS_TESTING=true npx vitest run test/util/databaseStandalone.test.ts test/util/databaseWrite.test.ts
# Test Files  2 passed (2)
# Tests  14 passed (14)
```

LGTM 👍

mldangelo-oai (approved):

Issue #12, opened 2023-05-27
feat: read folders


MentalGear:
Instead of defining each prompt file in `promptfooconfig.js` > `prompts:[...]`, it'd be nice if we could just define a folder instead.

### Challenges
- What if there are multiple different file formats .csv / .txt?

PR #9506, opened 2026-05-28
fix(cache): avoid persisting sensitive fetch identities

mldangelo-oai:
## Summary
- skip persistent fetch-response caching when a request identity includes credentials or sensitive URL/body/header material
- avoid storing stable secret-derived fingerprints in cache keys
- document the persistent-cache boundary and add focused regression coverage

## Scope
This PR has been narrowed to release-audit finding **#3** only. The remaining fixes originally included here were split into independent PRs:

- #9510 - Claude Opus Azure, Vertex, and Bedrock compatibility (findings #4 and #9)
- #9512 - MiniMax `function_call` rejection (finding #8)
- #9514 - optimizer validation-split parsing and guidance (findings #12 and #13)
- #9515 - documentation metadata and legacy anchors (finding #14)

## Validation
- `npx vitest run test/cache.test.ts` (52 tests passed)
- The original combined branch completed all 38 CI checks successfully before this scope split; checks for this narrowed branch are rerunning.

### DIFF
--- a/site/docs/configuration/caching.md
+++ b/site/docs/configuration/caching.md
@@ -34,8 +34,9 @@ Cache entries are stored using provider-specific composite keys that include:
 - Context variables (when applicable)
 
 Cache key formats are implementation details and may change between versions.
-Sensitive request payloads and headers are hashed where possible instead of
-being embedded directly in cache keys.
+Fetch responses whose cache identity includes credentials or other sensitive
+values are not written to the persistent cache. This avoids storing secrets or
+stable secret-derived fingerprints in cache keys.
 
 ```js
 // Provider-specific scope plus a digest of request material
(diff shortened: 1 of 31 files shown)


mldangelo:
Closing in favor of #9509 (merged), which fixes the underlying cross-process fetch-cache-miss bug without the behavior regression this PR introduces.

This approach skips the persistent cache entirely for any request whose identity contains credentials (Authorization/x-api-key/cookie/secret-looking values) and for all binary request bodies. Those requests are cached today (the existing `should isolate cached responses by request headers` test asserts `cache.set` is called), so this would disable persistent caching for most authenticated HTTP/API provider calls and all binary-body requests — a meaningful cache-hit-rate/cost/latency regression to land mid-release. There's also the unaddressed Codex P2 (userinfo-credential URLs like https://user:pass@host still get cached here).

#9509 keeps current caching behavior intact and only makes the secret fingerprint deterministic across processes so credentialed requests stay cacheable across runs; raw secrets are still never stored. If we later want a 'persist nothing secret-derived' posture, we can revisit with the userinfo gap fixed and a deliberate decision on the caching tradeoff.

PR #9509, opened 2026-05-28
fix(cache): make the fetch cache key stable across processes

mldangelo:
## Summary

The fetch cache key fingerprints secret-bearing fields (and hashes all binary request bodies) with an HMAC keyed by `crypto.randomBytes(32)` generated **once per process**:

```ts
const FETCH_CACHE_SECRET_HMAC_KEY = crypto.randomBytes(32);
```

Because the on-disk key is derived from that fingerprint, it **changed on every run** for any request carrying a static secret (API key in a header/body/query string) or a binary body. Repeated `promptfoo eval` runs therefore never hit the disk cache for those requests and re-called the upstream endpoint every time — extra latency, cost, and load. (Within a single process the cache still worked, and credential isolation was correct, which is why this slipped through.)

This was introduced by the auth-token cache isolation work (#8636).

## Fix

Replace the per-process random key with a fixed, compiled-in salt (renamed `…_HMAC_SALT` to make the intent clear). The salt is **not a secret** — it only domain-separates the one-way HMAC so raw secrets are never written into the cache key, while keeping the key deterministic across processes so the cache hits across runs.

- Credential isolation is unchanged: different secrets still produce different keys.
- Raw secrets are still never stored in the key (HMAC is one-way).
- This matches the pre-isolation behavior, which hashed the value directly (also deterministic, with no salt) — so it's strictly ≥ that security posture.

## Tests

- New regression test: the same secret-bearing request maps to the **same** cache key across a fresh module load (emulating a separate process). Verified that it **fails** with the prior `crypto.randomBytes(32)` key and passes with the fixed salt.
- Full `cache.test.ts` (53 tests) passes; `tsc` and `biome ci` clean.

Part of a set of small fixes flagged during a pre-release audit of changes since `0.121.12`.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -288,14 +288,22 @@ type PreparedFetchResponse = {
 const inflightFetchResponses = new Map<string, Promise<SerializedFetchResponse>>();
 const IGNORED_FETCH_CACHE_OPTION_KEYS = new Set(['method', 'signal']);
 const FETCH_CACHE_SECRET_HMAC_CONTEXT = 'promptfoo:fetch-cache-secret-key';
-const FETCH_CACHE_SECRET_HMAC_KEY = crypto.randomBytes(32);
+// A fixed, compiled-in salt (NOT a secret). It must be deterministic across
+// processes so that a request carrying a static secret — or a binary body —
+// hashes to the same on-disk cache key on every run and stays cacheable. A
+// per-process random key broke that: each `promptfoo eval` run produced a new
+// key and re-hit the upstream endpoint. The salt only domain-separates the
+// one-way HMAC so raw secrets are never written into the cache key; it does not
+// need to be unpredictable, and this matches the pre-existing (pre-isolation)
+// behavior of hashing the value directly.
+const FETCH_CACHE_SECRET_HMAC_SALT = 'promptfoo:fetch-cache-secret-hmac-salt:v1';
 const abortSignalIds = new WeakMap<AbortSignal, number>();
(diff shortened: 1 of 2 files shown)


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/32d61fb6-ada0-4dbd-9f22-0279c1357e58?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on cd14f6c098bf6224612e9923a525fc2f8de30ff9. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| cd14f6c098bf6224612e9923a525fc2f8de30ff9 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/32d61fb6-ada0-4dbd-9f22-0279c1357e58?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 28, 2026 9:09PM |

</p>
</details>

PR #9522, opened 2026-05-29
feat(cli): add seeded random sampling

ianw-oai:


mldangelo-oai:
I reproduced the unresolved top-level sampling resume issue on the current head (`87662816f706ee3b69758baf4a3cd3990e9da2b6`) and consider it blocking.

Using config-backed sampling:

```yaml
commandLineOptions:
  filterSample: 5
  filterSampleSeed: smoke-resume-config
```

I started a 10-test eval, paused after 3 of the sampled 5 cases, then resumed it. Resume reported:

```text
Resuming: skipping 3 previously completed cases
Running 7 test cases (up to 1 at a time)...
...
10 passed
```

Expected behavior is to run only the remaining 2 sampled cases. The filtered top-level test set is applied to the in-memory suite, but the persisted eval config still contains all original tests. Resume reloads that full config and skips filtering. The retry paths share the same persisted snapshot and index-identity risk.

The analogous CLI-only resume gap predates this PR, but this PR newly exposes the bug through config-backed and seeded sampling. Scenario sampling persists its selected scenarios and did not reproduce the issue.

Suggested fix: persist the resolved sampled top-level test subset as the canonical eval config before `Eval.create`, then add pause/resume and retry regression coverage for config-backed seeded sampling.

Additional QA completed on this head: targeted Biome checks, schema regeneration, `git diff --check`, 583 targeted Vitest tests, 3 scenario tests, and direct same-seed repeatability checks all passed.

mldangelo-oai:
One API/UX follow-up after comparing this with NumPy, pandas, scikit-learn, Polars, and Jest: I do not think a mnemonic string seed should be the primary documented interface yet.

The underlying approach is reasonable for NumPy-style replay: use a deterministic PRNG, Fisher-Yates shuffle the ordered candidate tests, then take N. But the public contract should be sharper:

1. Lead with a numeric seed (`filterSampleSeed: 42`, `--filter-sample-seed 42`). That is the convention developers expect from NumPy/pandas/scikit-learn/Polars and from test-runner seed workflows.
2. Supporting strings can still be useful for CLI/config ergonomics, but document them as optional named replay tokens rather than durable benchmark identities. `benchmark-v1` currently reads like a stable cohort name, even though the selected cases can change when input tests, input ordering, preceding filters, scenario ordering, or Promptfoo sampling internals change.
3. Tighten validation. The current schema accepts empty strings, negative numbers, and fractional numbers before stringifying them. If numbers are supported, I would restrict them to non-negative integers; if strings are supported, require a non-empty string.
4. State the reproducibility boundary explicitly: same Promptfoo sampling recipe/version, same seed, same sample size, same filters, and same ordered input tests select the same sample. This is Promptfoo-specific; it is not NumPy-compatible output and should not imply stability after dataset/order changes.
5. Add a golden test for one known seed and ordered input, rather than only asserting that two calls in the same implementation agree. Otherwise a later algorithm refactor can silently change every seeded sample while the tests remain green.

There is a product choice worth making explicitly: if the intended feature is only repeatable runs against the same ordered config, this PRNG/shuffle design is appropriate after the contract is clarified. If the intent is a long-lived benchmark cohort that survives reordering or appending tests, seeded shuffling by array position is the wrong abstraction; persist the selected cohort or use stable test identities and a keyed ranking scheme.

As a follow-up UX improvement, consider generating, persisting, and displaying the effective seed when `--filter-sample` is used without one. Jest exposes its shuffle seed for exactly this replay workflow.

mldangelo-oai:
I found and reproduced a second blocking issue in the new scenario-sampling path on the current head (`87662816f706ee3b69758baf4a3cd3990e9da2b6`).

`resolveConfigs` now processes `config.scenarios`, including scenarios inherited from a reusable `defaultConfig`, and assigns the sampled subset back into each original scenario object:

```ts
scenario.tests = filteredTests;
```

Because `config.scenarios` can share the caller-owned `defaultConfig.scenarios` objects, one resolution mutates the input config. Any later resolution can sample only from the already-reduced subset.

I reproduced this directly against the PR head:

```text
first resolve: filterSample=2, seed=first
selected:                ["0", "2"]
caller-owned config now: ["0", "2"]

second resolve on same defaultConfig: filterSample=4, seed=second
selected:                ["0", "2"]
```

The second resolve should be able to select all four original tests. This can affect reload/watch-style flows and any caller that reuses an in-memory default config.

Suggested fix: clone scenario objects and their test arrays before resolving/filtering, then assign filtered tests only on the cloned config snapshot. Please add a regression test that calls `resolveConfigs({}, sameDefaultConfig)` twice with different seeds/sample sizes and verifies that the original default config remains unchanged.

mldangelo-oai:
I found one remaining numeric-seed correctness issue on current head (`0912d584f0ecac2fca037e61468e5e51d7034ec1`).

`filterSampleSeed` is declared numeric with `z.coerce.number().int()`, but config loading validates and then returns the raw `renderedConfig`. Scenario sampling later derives child seeds with:

```ts
sampleSeed: filterSampleSeed === undefined ? undefined : filterSampleSeed + scenarioIndex
```

This creates three observable problems:

1. YAML `filterSampleSeed: 42` and `filterSampleSeed: '42'` are both schema-valid, but the quoted value remains a string, so scenarios derive `"420"`, `"421"`, ... rather than `42`, `43`, ... .
2. Invalid YAML seeds such as `named-seed` and `1.5` emit warnings but still run successfully with exit code 0; the seeded filter hashes leaked raw values as strings.
3. `Number.MAX_SAFE_INTEGER` is accepted by the schema, then adding scenario indices crosses the safe-integer boundary; adjacent scenarios can receive the same derived seed.

Reproduction from the current implementation:

```text
numeric 42 scenarios: [["two","zero"],["two","zero"]]
quoted '42' scenarios: [["three","two"],["two","zero"]]
named-seed YAML: warns, eval runs, exit 0
1.5 YAML: warns, eval runs, exit 0
```

Suggested fix: consume normalized `validationResult.data` or fail invalid `commandLineOptions`, require a safe-integer seed, and derive scenario streams with a bounded tuple hash/mixer instead of raw `+`.

mldangelo-oai (approved):

PR #9524, opened 2026-05-29
fix(assertions): don't classify gen_ai.tool.definitions chat spans as tool calls

### LINKED ISSUE #9523: `trajectory:tool-sequence` with `mode: exact` fails - `extractToolName` fallback regex matches `gen_ai.tool.definitions` on LLM chat spans
**Describe the bug**

When using `trajectory:tool-sequence` with `mode: exact`, the assertion incorrectly fails because the fallback regex in `extractToolName` (`src/assertions/trajectoryUtils.ts` line 146) is too broad. LLM chat spans emitted by pydantic-ai (and likely other OpenTelemetry-based tracing backends) carry a `gen_ai.tool.definitions` attribute containing the JSON-serialized list of tools available to the model. The fallback regex `/(^|[._])tool($|[._])/i` matches `gen_ai.tool.definitions` (via the `_tool.` substring), causing those chat spans to be misclassified as `type: 'tool'` steps with the entire JSON array as their "tool name".

This inflates the tool step count and causes the `mode: exact` count check to fail even when the agent called exactly the right tools in the right order.

The misclassified spans produce "tool names" like:

```
tool:[{"type":"function","name":"orders","description":"..."},{"type":"function","name":"order_rows","description":"..."}]
```

These appear once per LLM turn, so a 3-turn agent run with 6 real tool calls ends up with 9 tool steps instead of 6.

**To Reproduce**

1. Configure a multi-step agent eval using a pydantic-ai (or similar OpenTelemetry) tracing backend where the agent makes tool calls across multiple LLM turns. For example:

```yaml
tests:
  - description: >
      Step 0) fetch rows for order 0000009254.
      Step 1) for each of the first 5 products, search which orders contain it —
      all 5 calls issued in the same LLM turn (parallel).
    vars:
      user_query: >
        Get me the info of the order rows of order 0000009254, and get the
        first 5 products found and find which other orders had each of these products in them.
    assert:
        metric: tool_sequence
        value:
          mode: exact
          steps:
            - order_rows
            - orders
            - orders
            - orders
            - orders
            - orders
```

2. Run `promptfoo eval`.
3. Observe the `tool_sequence` metric fails with a reason like:

```
Expected exact tool sequence of order_rows, orders, orders, orders, orders, orders,
but actual tools were
  tool:[{"type":"function","name":"orders",...},{"type":"function","name":"order_rows",...}],
  tool:order_rows,
  tool:[{"type":"function","name":"orders",...},{"type":"function","name":"order_rows",...}],
  tool:orders,
  tool:orders,
  tool:orders,
  tool:orders,
  tool:orders,
  tool:[{"type":"function","name":"orders",...},{"type":"function","name":"order_rows",...}]
```

**Expected behavior**

`trajectory:tool-sequence` should only see actual tool invocation steps. LLM chat spans that carry the available tools list as an attribute (`gen_ai.tool.definitions` or similar) should not be classified as `type: 'tool'` steps. The assertion should pass when the agent called exactly `order_rows` followed by 5 `orders` calls.

**Root cause analysis**

The bug is entirely in the fallback loop of `extractToolName` in `src/assertions/trajectoryUtils.ts` lines 137–149:

```typescript
for (const [key, value] of Object.entries(attributes)) {
  if (typeof value !== 'string' || !value.trim()) {
    continue;
  }

  if (/tool.?name|function.?name/i.test(key)) {
    return value.trim();
  }

  if (/(^|[._])tool($|[._])/i.test(key) && !/result|output/i.test(key)) {
    return value.trim();   // ← BUG: matches gen_ai.tool.definitions
  }
}
```

The regex `/(^|[._])tool($|[._])/i` matches `gen_ai.tool.definitions` because the substring `_tool.` satisfies `[._]tool[._]`. The key does not contain `result` or `output`, so the guard passes and the full JSON array string is returned as the "tool name".

**Important:** `getToolNameFromAttributes` (the allowlist at `src/tracing/toolAttributes.ts`) is NOT involved — it correctly returns `undefined` for chat spans because `gen_ai.tool.definitions` is not in `TOOL_NAME_ATTRIBUTE_KEYS`. The fallback loop is the sole culprit.

The misclassified span is then classified as `type: 'tool'` in `extractTrajectorySteps`, and `handleTrajectoryToolSequence` includes it in `toolSteps` (line 316):

```typescript
const toolSteps = extractTrajectorySteps(trace).filter((step) => step.type === 'tool');
```

The `mode: exact` check then fails immediately:

```typescript
basePass =
  toolSteps.length === expectedMatchers.length &&   // 9 !== 6 → false
  expectedMatchers.every((matcher, index) => matchesTrajectoryStep(toolSteps[index], matcher));
```

**Why `trajectory:tool-used` and `mode: in_order` are not affected**

- `trajectory:tool-used` uses an existence check (`steps.some(...)`) — the misclassified spans have names like `[{...}]` which never match a real tool name like `orders`, so they are harmless noise.
- `trajectory:tool-sequence` with `mode: in_order` uses subsequence matching that skips non-matching steps — the JSON-array-named spans are simply skipped over.
- Only `mode: exact` is broken because it checks the total count first.

**Proposed fixes**

All three options target the same line in `src/assertions/trajectoryUtils.ts` (line 146).

**Option 1 — Add `definitions` to the existing blocklist (minimal, one-word change):**

```typescript
// Before:
if (/(^|[._])tool($|[._])/i.test(key) && !/result|output/i.test(key)) {

// After:
if (/(^|[._])tool($|[._])/i.test(key) && !/result|output|definitions/i.test(key)) {
```

Blocks `gen_ai.tool.definitions` specifically. Minimal diff, but only covers this one known key.

**Option 2 — JSON array guard in the fallback loop (more robust):**

```typescript
for (const [key, value] of Object.entries(attributes)) {
  if (typeof value !== 'string' || !value.trim()) {
    continue;
  }

  const trimmed = value.trim();

  // Reject JSON arrays — these are tool definition lists, not tool names
  if (trimmed.startsWith('[')) {
    try {
      if (Array.isArray(JSON.parse(trimmed))) {
        continue;
      }
    } catch {
      // not valid JSON, treat as a normal string tool name
    }
  }

  if (/tool.?name|function.?name/i.test(key)) {
    return trimmed;
  }

  if (/(^|[._])tool($|[._])/i.test(key) && !/result|output/i.test(key)) {
    return trimmed;
  }
}
```

Handles any attribute whose value is a JSON array of tool definitions, regardless of key name. More future-proof.

**Option C — Both combined (most robust):**

Apply the `definitions` blocklist to the regex AND add the JSON array guard. This blocks the known bad key by name AND rejects any JSON array value regardless of key, covering both the current case and any future similar attributes from other tracing backends:

```typescript
for (const [key, value] of Object.entries(attributes)) {
  if (typeof value !== 'string' || !value.trim()) {
    continue;
  }

  const trimmed = value.trim();

  // Reject JSON arrays — these are tool definition lists, not tool names
  if (trimmed.startsWith('[')) {
    try {
      if (Array.isArray(JSON.parse(trimmed))) {
        continue;
      }
    } catch {
      // not valid JSON, treat as a normal string tool name
    }
  }

  if (/tool.?name|function.?name/i.test(key)) {
    return trimmed;
  }

  if (/(^|[._])tool($|[._])/i.test(key) && !/result|output|definitions/i.test(key)) {
    return trimmed;
  }
}
```

**System information:**

- Promptfoo version: 0.121.13
- Tracing backend: pydantic-ai with OpenTelemetry (`otel.scope.name: pydantic-ai`, `otel.scope.version: 1.96.1`)

**Additional context**

Relevant source locations:

- `src/assertions/trajectoryUtils.ts` lines 137–149 — `extractToolName` fallback loop (the sole culprit)
- `src/assertions/trajectory.ts` line 316 — `handleTrajectoryToolSequence` filters to `step.type === 'tool'` (correct intent, wrong upstream classification)
- `src/assertions/trajectory.ts` lines 332–335 — `mode: exact` count check that fails when extra misclassified steps are present

The offending span attribute from the actual trace:

```json
"gen_ai.tool.definitions": "[{\"type\":\"function\",\"name\":\"orders\",...},{\"type\":\"function\",\"name\":\"order_rows\",...}]"
```

This attribute is present on every `chat gpt-4o-mini` span (one per LLM turn). The `execute_tool` spans correctly use `gen_ai.tool.name` (which is in the `TOOL_NAME_ATTRIBUTE_KEYS` allowlist and is handled correctly).

mldangelo:
## Summary

Fixes #9523. `trajectory:tool-sequence` with `mode: exact` (and every other trajectory assertion) over-counts tool steps for traces from pydantic-ai and other OpenTelemetry GenAI-convention backends.

LLM chat/generation spans carry a `gen_ai.tool.definitions` attribute — the JSON-serialized list of tools _available_ to the model. The fallback in `extractToolName` matched that key via the broad `/(^|[._])tool($|[._])/i` regex and returned the whole JSON array as a "tool name", so each chat span was misclassified as a `type: 'tool'` step. A 3-turn run with 6 real tool calls reported 9 tool steps, and `mode: exact` failed even when the agent called exactly the right tools.

## Fix

In the `extractToolName` fallback loop (`src/assertions/trajectoryUtils.ts`):

- Skip any attribute value that is a JSON array or object — a tool name is always a scalar identifier, never a serialized list of tool definitions. This is the general guard and covers `gen_ai.tool.definitions` regardless of key name.
- Also exclude `definition` from the broad tool-key fallback (alongside the existing `result`/`output`), matching intent.

Real tool calls (`gen_ai.tool.name`, etc.) are unaffected — they're resolved by the canonical `getToolNameFromAttributes` allowlist before the fallback ever runs, and scalar string fallbacks (e.g. a custom `app.tool: lookup_customer`) still work.

## Test plan

- Reproduced the bug first (a `chat` span with `gen_ai.tool.definitions` yielded an extra `tool` step), then confirmed the fix.
- Added two regression tests in `test/assertions/trajectory.test.ts`:
  - a chat span carrying `gen_ai.tool.definitions` is classified as a `span`, not a `tool`, while the sibling `execute_tool` span (`gen_ai.tool.name`) is the only tool step;
  - a scalar generic tool attribute (`app.tool`) still resolves to a tool name (guard doesn't regress fallbacks).
- `npx vitest run test/assertions/trajectory.test.ts` → 96 pass. `npm run tsc` clean.

Closes #9523.

### DIFF
--- a/src/assertions/trajectoryUtils.ts
+++ b/src/assertions/trajectoryUtils.ts
@@ -139,12 +139,29 @@ function extractToolName(span: TraceSpan): string | undefined {
       continue;
     }
 
-    if (/tool.?name|function.?name/i.test(key)) {
-      return value.trim();
+    const trimmed = value.trim();
+
+    // A tool name is always a scalar identifier. Some chat/generation spans carry the list
+    // of *available* tools as a JSON-serialized array/object under a `tool`-matching key
+    // (e.g. `gen_ai.tool.definitions` from pydantic-ai). Never treat such a structured
+    // value as a tool name, or those chat spans get misclassified as tool calls. (#9523)
+    if (trimmed.startsWith('[') || trimmed.startsWith('{')) {
+      try {
+        const parsed = JSON.parse(trimmed);
+        if (parsed !== null && typeof parsed === 'object') {
+          continue;
+        }
+      } catch {
+        // Not JSON — fall through and treat it as an ordinary string tool name.
+      }
     }
 
-    if (/(^|[._])tool($|[._])/i.test(key) && !/result|output/i.test(key)) {
-      return value.trim();
+    if (/tool.?name|function.?name/i.test(key)) {
+      return trimmed;
+    }
+
+    if (/(^|[._])tool($|[._])/i.test(key) && !/result|output|definition/i.test(key)) {
+      return trimmed;
     }
   }
 
--- a/test/assertions/trajectory.test.ts
+++ b/test/assertions/trajectory.test.ts
@@ -131,6 +131,60 @@ describe('trajectory utilities', () => {
     });
   });
 
+  it('does not classify a chat span carrying gen_ai.tool.definitions as a tool', () => {
+    // pydantic-ai (and other GenAI-convention tracers) put the list of *available* tools on
+    // each chat span as a JSON array under gen_ai.tool.definitions; that must not be read as
+    // a tool call. Real tool calls use gen_ai.tool.name. Regression for #9523.
+    const toolDefinitions = JSON.stringify([
+      { type: 'function', name: 'orders', description: 'List orders' },
+      { type: 'function', name: 'order_rows', description: 'List order rows' },
+    ]);
+    const steps = extractTrajectorySteps({
+      ...mockTraceData,
+      spans: [
+        {
+          spanId: 'chat-1',
+          name: 'chat gpt-4o-mini',
+          startTime: 1000,
+          endTime: 1100,
+          attributes: { 'gen_ai.tool.definitions': toolDefinitions },
+        },
+        {
+          spanId: 'exec-1',
+          name: 'execute_tool',
+          startTime: 1200,
+          endTime: 1300,
+          attributes: { 'gen_ai.tool.name': 'order_rows' },
+        },
+      ],
+    });
+
+    expect(steps.map((step) => ({ type: step.type, name: step.name }))).toEqual([
+      { type: 'span', name: 'chat gpt-4o-mini' },
+      { type: 'tool', name: 'order_rows' },
+    ]);
+  });
+
+  it('still reads a plain string tool name from a generic tool attribute', () => {
+    // The structured-value guard must not break scalar tool-name fallbacks.
+    const steps = extractTrajectorySteps({
+      ...mockTraceData,
+      spans: [
+        {
+          spanId: 'custom-tool',
+          name: 'invoke',
+          startTime: 1000,
+          endTime: 1100,
+          attributes: { 'app.tool': 'lookup_customer' },
+        },
+      ],
+    });
+
+    expect(steps.map((step) => ({ type: step.type, name: step.name }))).toEqual([
+      { type: 'tool', name: 'lookup_customer' },
+    ]);
+  });
+
   it('only treats a generic query attribute as search when the span looks search-like', () => {
     const steps = extractTrajectorySteps({
       ...mockTraceData,


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/89a2a905-e782-4668-ad58-570fcc5bfa52?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on 2c64614ec56bfec2e2cd38c8c33648052b222ccf. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 2c64614ec56bfec2e2cd38c8c33648052b222ccf | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/89a2a905-e782-4668-ad58-570fcc5bfa52?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 29, 2026 12:27PM |

</p>
</details>

sipa-echo-ngbm:
This fix also resolves a secondary symptom I observed: `trajectory:tool-used` was passing correctly but the reason string was extremely large because the misclassified `gen_ai.tool.definitions` spans (with their full JSON tool definitions as "tool names") were being included in the "Actual tools" list in the reason message. With this fix, those spans are no longer in `steps`, so the reason string will only show real tool calls. The PR fix will resolve this too.

sipa-echo-ngbm:
Another symptom seen is that `trajectory:step-count` with `type: tool` is also affected - the misclassified `gen_ai.tool.definitions` spans pass the `type: 'tool'` filter and inflate the count. The PR fix will resolve this too.

PR #9526, opened 2026-05-29
fix(webui): render negative-only metric charts

mldangelo-oai:
## Summary
- normalize negative-only named metrics by their maximum magnitude
- preserve existing behavior for positive, mixed, and all-zero metric series
- add regression coverage for negative-only metric charts

## Root cause
Metric charts divided values only when a metric's maximum was positive. A valid all-negative series therefore rendered every value as zero and hid relative differences between providers.

## Validation
- `npm run test:app -- src/pages/eval/components/ResultsCharts.test.tsx --run`
- `npm --prefix src/app run tsc`
- `npm run f`
- `npm run l`

### DIFF
--- a/src/app/src/pages/eval/components/ResultsCharts.test.tsx
+++ b/src/app/src/pages/eval/components/ResultsCharts.test.tsx
@@ -699,6 +699,67 @@ describe('ResultsCharts', () => {
       expect(values).toEqual([0, 0, 0, 0]);
     });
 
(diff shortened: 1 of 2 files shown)


ianw-oai (approved):
Looks good. Small, targeted fix with regression coverage; CI is green.

ianw-oai (approved):
Looks good. Focused negative-only metric chart normalization fix with a regression test.

PR #9530, opened 2026-05-29
chore(build): harden architecture boundaries

mldangelo-oai:
## Summary
- resolve configured source aliases in the architecture checker
- require checked source files to belong to explicit layers and ratchet allowed layer dependencies
- pin the web app to its current cross-layer import paths while browser-safe DTO surfaces are extracted

## Validation
- `npx vitest run test/scripts/architectureUtils.test.ts`
- `npm run deps:ownership -- --json`
- `./node_modules/.bin/biome check architecture/layers.json scripts/architectureUtils.ts scripts/checkArchitectureBoundaries.ts scripts/reportDependencyOwnership.ts test/scripts/architectureUtils.test.ts`
- `./node_modules/.bin/prettier --check docs/architecture/packages.md`
- `git diff --check HEAD~1`

## Known baseline issue
- `npm run tsc` still fails in untouched Anthropic provider tests because the installed SDK type does not include `output_tokens_details`.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
This PR hardens the architecture-boundary checker: it adds per-layer `allowedDependencies` enforcement, a restricted-path allowlist (`allowedImportPaths`) for the `app` layer, and an unclassified-file check, all wired through `scripts/checkArchitectureBoundaries.ts`. The logic is sound and the new tests cover the happy and basic-violation paths. This is CI-only build tooling that reads the repo's own committed source — no runtime or security surface — so everything below is non-blocking polish. **Looks good to merge.**

## Low / Nits

- `scripts/architectureUtils.ts:284-316` — **A single offending `app` import emits two violations for the same root cause.** In `findViolations`, the `layer` branch (`!(importerConfig.allowedDependencies ?? []).includes(importedLayer)`) and the `path` branch (`importerConfig?.allowedImportPaths && !importerConfig.allowedImportPaths.includes(resolvedImport)`) are independent — the path branch never consults `allowedDependencies`. With the real `architecture/layers.json` (`app.allowedDependencies = ["legacy-contracts","legacy-runtime","node","providers","redteam"]`), any `app` import of a `core`/`cli`/`view-server` file that is also absent from `allowedImportPaths` trips *both*, so `checkArchitectureBoundaries.ts` prints that one import twice — once under "Layer dependency violations found:" and again under "Restricted layer import violations found:". Confusing noise for whoever is fixing it. The two existing path-violation tests import from `shared`, which *is* in the test config's `app.allowedDependencies`, so the layer branch is suppressed and this overlap is never exercised.
  - Fix: suppress the `path` violation when a `layer` violation already fired for the same `importer` + `resolvedImport` (so each offending import is listed once). If double-reporting is intentional, lock it in with a regression test asserting the exact violation set for an `app` import that is both a disallowed layer *and* a non-allowlisted path.

- `scripts/architectureUtils.ts:295` — **Dead `?? []` fallback contradicts the field's own contract.** `LayerDefinition.allowedDependencies` is declared required (`allowedDependencies: string[]`) and `readLayerConfig` throws `Architecture layer "X" must declare allowedDependencies.` when it isn't an array. Given that validated invariant, the `?? []` in `findViolations` is unreachable for any config `readLayerConfig` produces. The only test reaching it (`treats omitted allowed dependencies as an empty allowlist`) hand-builds `{ ..., allowedDependencies: undefined as unknown as string[] }` and calls `findViolations` directly, bypassing the validator — leaving three contradictory statements about one field (required by type, required by `readLayerConfig`, optional by the `??` + bypass test). Recommend keeping the field required + validated and dropping the dead fallback:
    ```ts
    !importerConfig.allowedDependencies.includes(importedLayer)
    ```
    Then remove the `undefined as unknown as string[]` bypass test — the throw is already covered by `rejects layers that omit allowed dependencies`.

- `scripts/checkArchitectureBoundaries.ts:8,14` — **The src tree is globbed and layer-classified twice per run.** `findViolations(repoRoot, config)` calls `getSourceFiles` + `getLayerForFile` over `src/**/*.{ts,tsx,mts,cts}`, then `findUnclassifiedFiles(repoRoot, config)` globs and classifies again. The second pass doesn't re-read file contents (only `findViolations` parses source text), so it's the cheaper half, but it's avoidable repeated work on every CI run and `npm run architecture:check`. Consider an `analyzeArchitecture(repoRoot, config)` that globs/classifies once and returns `{ violations, unclassifiedFiles }`, or compute `getSourceFiles` in the checker and pass the list into both helpers.

- `scripts/architectureUtils.ts:38-46,49-62,309-310` — **`ignoredRoots` and `allowedImportPaths` entries are never validated for existence, so a stale allowlist entry silently survives a rename.** `readLayerConfig` only validates that each layer has an array `allowedDependencies`; it doesn't confirm any of the ~26 exact-file `app.allowedImportPaths` entries (or `ignoredRoots`) resolve under `repoRoot`, and enforcement is exact-string membership. If an allowlisted file is later renamed/moved, its entry becomes dead — it stops pinning anything and silently stops contributing to the browser ratchet, eroding the guardrail with no signal. (A stale `ignoredRoots` entry is harmless.) Optional: add a self-consistency check in `readLayerConfig` asserting each `ignoredRoots`/`allowedImportPaths` entry resolves to an existing path under `repoRoot`, failing with a clear message so the allowlist stays honest as files move. Low priority.

## Tests

The new tests cover the happy path and single-kind violations well, but two gaps map directly to the nits above: there's no test for an `app` import that is both a disallowed layer and a non-allowlisted path (the double-report case), and the "omitted allowed dependencies" test only reaches the dead `?? []` branch by bypassing `readLayerConfig`. Addressing the first nit (dedup) should come with a regression test that asserts the exact violation set.

ianw-oai (approved):

mldangelo:
## PR Review — pr-review-toolkit (code · tests · errors · types · comments · simplify)

Ran the full multi-agent review against the checked-out branch. **The checker is correct and safe to merge** — `npm run architecture:check` passes (all 1530 source files classified, exit codes correct) and the new suite is 32/32 green. No blocking bugs. Findings below are hardening, one missing test for the PR's central new behavior, and cleanup.

### Important (worth addressing)

1. **The guardrail can silently pass if config drifts — `readLayerConfig` validates almost nothing beyond `allowedDependencies`.** [`scripts/architectureUtils.ts:38-47`] The whole browser-import ratchet depends on `@promptfoo/*` aliases resolving to real files; if an alias base goes stale (typo, or `src/` is restructured without updating `layers.json`), `resolveInternalModule` returns `undefined`, every aliased import is silently treated as external and skipped, and the checker still prints "passed." Same class of silent-loosening: stale `allowedImportPaths` entries (file renamed/deleted), typo'd layer names in `allowedDependencies`, duplicate layer `name`s / overlapping `roots` (classification is first-match via `getLayerForFile` but the allowlist is looked up last-write via `layersByName` — they can diverge), and a missing `publicFacade`. All latent today (I verified the live config is consistent), but a guardrail whose failure mode is "silently stops guarding" should fail loud. Recommend extending `readLayerConfig` to assert: every alias target exists under `repoRoot`; every `allowedImportPaths` entry resolves to a real file; every `allowedDependencies` string is a known layer name; no duplicate names / overlapping roots; `publicFacade` present. A Zod schema would cover all of this in one place (consistent with repo conventions) and also let you delete the `as LayerConfig` cast.

2. **The PR's central new behavior — an `app` import that violates BOTH the layer rule and the path allowlist — is untested, and currently double-reports.** [`scripts/architectureUtils.ts:289-320`] The `layer` (289-305) and `path` (307-320) checks are independent `if` blocks. For the real `app` layer (the only one with both `allowedDependencies` and `allowedImportPaths`), a disallowed import trips both and prints twice under two headers (verified: `kind` = `['layer','path']` for a single import). Exit is still correctly 1, so it's a clarity wrinkle, not a correctness bug — but every existing test uses fixtures where only one rule fires, so a future refactor making them mutually exclusive would change output with no failing test. Recommend making them disjoint (run the `path` check only when the layer edge is otherwise allowed — i.e. `path` = "layer relationship permitted, but this specific file isn't", matching the documented intent) and adding a regression test asserting the exact violation set for the combined case.

3. **Enforcement-coverage gaps** [`test/scripts/architectureUtils.test.ts`] beyond #2:
   - Dynamic `import('@promptfoo/…')` and `export … from '@promptfoo/…'` are only unit-tested at the specifier-extraction level, not driven through `findViolations` — yet those are the most common ways a dependency sneaks past review (verified both currently produce `path` violations; worth pinning).
   - `readLayerConfig` tests only the *missing* `allowedDependencies` branch, not a *non-array* value (a hand-edited string) — a future switch to a truthiness check (`if (!x)`) would wrongly accept `"core"`.
   - No test that a layer *without* `allowedImportPaths` never emits a `path` violation (guards the optional-chain); the alias longest-prefix tie-break (`@app` vs `@promptfoo`) and the `getSourceFiles(includeApp=false)` branch are unexercised.

### Suggestions (nice to have)

- **Dead `?? []` fallback** [`scripts/architectureUtils.ts:295`] — `allowedDependencies` is typed required and `readLayerConfig` throws if it isn't an array, yet `findViolations` still does `(importerConfig.allowedDependencies ?? []).includes(...)`. Unreachable for any config produced by `readLayerConfig` — three contradictory statements about one field. The only test reaching it bypasses `readLayerConfig` with `undefined as unknown as string[]`. Drop the `?? []` and remove that bypass test (the throw is already covered by `rejects layers that omit allowed dependencies`).
- **Eliminate the duplicate full-tree walk** [`scripts/checkArchitectureBoundaries.ts:8,14`] — `findViolations` and `findUnclassifiedFiles` each independently `globSync` `src/**/*.{ts,tsx,mts,cts}` and run `getLayerForFile` on every file, so the tree is scanned and classified twice per run. `findViolations` already computes `importerLayer` for every file. Fold the unclassified collection into that single pass and return `{ violations, unclassifiedFiles }`.
- **Docs precision** [`docs/architecture/packages.md:70-79`] — the cross-layer reach is via `@promptfoo/*` (maps to `src`), not `@app/*` (maps *inside* the app layer at `src/app/src`, so `@app/*` imports are always same-layer and can never cross a boundary); listing `@app/*` as the boundary-crossing alias misleads. And "A new app import from root runtime code fails the check" undersells the rule — the `path` allowlist spans many layers (`src/redteam/…`, `src/providers/…`, `src/types/…`, `src/util/…`), so *any* cross-layer app import not on the allowlist trips it.
- **Minor:** nested ternary in `resolveInternalModule` (`:168-172`) runs against the project's no-nested-ternary guidance (lift to if/else); add a `satisfies never` exhaustiveness check tying `BoundaryViolationKind` to the per-kind `.filter`s in the checker so a future 5th kind can't be silently dropped from output; hoist the per-specifier `Object.keys(aliases).sort()` out of the hot loop.

### Strengths

- Passes on the real tree (1530 files classified, correct exit codes); the `app → src/app` broadening correctly excludes the nested `src/app/node_modules` workspace.
- Tests use real temp fixture trees (no mocking the unit under test) with behavioral `toMatchObject` assertions on the violation contract — resilient to refactoring. The "exact files, not directory roots" test and the re-purposed "still applies layer rules when leafLayers is unset" test are especially good.
- Docs accurately describe the implemented semantics, and the `BoundaryViolation` JSDoc was correctly generalized to "every violation kind."

**Net: approve, no blockers.** Highest-value follow-ups: harden `readLayerConfig` validation (#1, so the ratchet can't silently no-op) and add the combined-violation regression test (#2).

<sub>Review aspects: code · tests · errors · types · comments · simplify.</sub>

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
This PR hardens the internal architecture boundary checker: it adds a per-layer `allowedDependencies` ratchet, an `app`-layer `allowedImportPaths` allowlist, alias resolution (`@promptfoo`/`@app`), a `legacy-runtime` catch-all layer, and config validation in `readLayerConfig`. The logic is sound, `npm run architecture:check` passes on the new config, and the 36 tests pass. The findings below are all robustness / test-coverage gaps in a build-time script — none block.

**Low**

- `scripts/architectureUtils.ts:52-63` — `validateLayerDefinition` validates each `allowedImportPaths` entry with only `fs.existsSync(...)`, no `isFile()` check. At runtime the allowlist is matched via `allowedImportPaths.includes(resolvedImport)`, and `resolvedImport` is always a concrete source file. So a config author who mistakenly lists a **directory** (e.g. `src/redteam/constants` — which exists as a dir alongside `src/redteam/constants.ts`) passes validation, but that entry can never match anything and silently becomes a dead no-op, producing a confusing path violation for every file under it. This directly contradicts the contract the PR's own docs add ("Allowlist entries are exact files, not directory roots") and is exactly the kind of footgun the directory case in `src/redteam/constants` invites. No impact today (every current entry is a real file). Fix: also require `fs.statSync(abs).isFile()` (and ideally a TypeScript extension) and throw a message naming the offending entry.

- `test/scripts/architectureUtils.test.ts:175-262` — `readLayerConfig` and its helpers add 11 throw branches; only 4 are covered (`must declare allowedDependencies`, `allows unknown dependency`, alias `points to missing path`, roots `overlap`). The other 7 have zero coverage (verified by grepping each message): `must define a layers array`, `must define an existing publicFacade path`, `contains duplicate layer`, `contains a layer without a name`, `must declare roots`, `contains an invalid allowedImportPaths entry`, and root `does not exist`. This is the integrity gate the PR exists to provide — a future refactor that neutered, say, the duplicate-layer or publicFacade-existence guard would let an invalid config through and silently degrade the boundary check, with green CI. Fix: add focused negative tests for each branch using the existing tmpdir + `writeFileSync` fixture pattern.

- `test/scripts/architectureUtils.test.ts:275-296` — No test covers an *allowed* cross-layer import in a layer with **no** `allowedImportPaths` (the dominant production case: only `app` declares `allowedImportPaths`; every other layer relies on "allowedDependencies present + allowedImportPaths undefined ⇒ no path check"). All existing path-check tests go through the restricted `app` layer; the leaf fixtures have empty `allowedDependencies` and hit a `layer` violation first. A regression that fired the path check when `allowedImportPaths` is undefined would flood `cli`/`providers`/`redteam`/etc. with false positives and slip through. Fix: assert `findViolations` returns `[]` for a layer with `allowedDependencies: ['shared']` and no `allowedImportPaths` importing an allowed dependency.

**Nits**

- `test/scripts/architectureUtils.test.ts:120-141` — `resolveInternalModule` now sorts alias keys by descending length to pick the longest matching prefix, but every test passes a single alias (`@promptfoo` → `src`); the real `@app` alias and any two overlapping aliases (where a shorter alias prefixes a longer one) are never exercised. If the sort were removed/reversed the suite would still pass, so the disambiguation is unverified. Worth a test with `{ '@app': 'src/app/src', '@': 'src' }` resolving `@app/foo` to the app tree.

- `scripts/architectureUtils.ts:388-407` — The leading `importedLayer !== importerLayer &&` in the `layer`-violation `if` is logically dead: whenever `!allowedLayerDependency` is true, both disjuncts of `allowedLayerDependency` are false, so `importedLayer === importerLayer` is already false. (The same term in the `path` condition at ~410-413 is *not* redundant — there `allowedLayerDependency` can be true via the same-layer disjunct, so it correctly filters app-internal imports.) Cosmetic; reasonable to keep as a readability mirror of the path condition, or drop it.

**Tests note**: The happy path and most violation kinds are well covered (36 passing). The gaps above are all negative/edge branches; the `readLayerConfig` validation branches (the PR's core hardening) are the most worth backfilling.

mldangelo-oai:
@codex review

PR #9531, opened 2026-05-29
refactor(server): extract eval job service

mldangelo-oai:
## Summary

- extract eval job lifecycle state into a dedicated server service
- remove the redteam route's direct import from the eval route
- return defensive job snapshots and cover lifecycle transitions with focused tests

## Validation

- `npx vitest run test/server/services/evalJobService.test.ts test/server/routes/redteam.test.ts test/server/routes/eval.sharing.test.ts test/server/routes/eval.validation.test.ts test/server/routes/serverRouteSmoke.test.ts --sequence.shuffle=false --no-file-parallelism`
- `./node_modules/.bin/biome check src/server/services/evalJobService.ts src/server/routes/eval.ts src/server/routes/redteam.ts test/server/services/evalJobService.test.ts`
- `./node_modules/.bin/prettier --check src/server/services/evalJobService.ts src/server/routes/eval.ts src/server/routes/redteam.ts test/server/services/evalJobService.test.ts`
- `git diff --check HEAD`

## Baseline Notes

- Shuffled execution of `test/server/routes/redteam.test.ts` can fail with localhost `EADDRNOTAVAIL` errors. The same failure reproduces from an untouched `origin/main` worktree; deterministic execution passes all 32 tests.
- Biome reports the existing `generate-test` cognitive complexity warning in `src/server/routes/redteam.ts`.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: refactor(server): extract eval job service

This PR extracts the in-memory eval/redteam job store out of `routes/eval.ts` (the exported `evalJobs` Map) into a dedicated `EvalJobService` with a typed API (`create`/`get`/`setProgress`/`complete`/`fail`/`appendLog`), and rewrites both the eval and redteam routes to call through it. The service hands out defensive deep-clones on read and write so callers can no longer mutate stored job state by reference. Clean separation, the route call sites read better, and the `fail({ append, resetResult })` options faithfully preserve the prior redteam cancel/error semantics. Overall this looks safe to merge — the items below are edge-case and coverage notes, not blockers.

### Low / Nits

- `src/server/services/evalJobService.ts:15-17, 28-30` — `complete()` now deep-clones the summary via `cloneResult` = `JSON.parse(JSON.stringify(result))`, which **throws** on circular refs or `BigInt`. This subtly relocates the failure point for a JSON-unsafe summary. Previously the route stored `result` by reference and the GET handler's `res.json()` was the only JSON pass — a bad summary kept the job `'complete'` and every GET threw a 500. Now `complete()` throws inside the `.then()`, the promise rejects into `.catch()`, and the job is marked `status: 'error'`. Note `GetJobResponseSchema.result` is `z.record(z.string(), z.unknown()).nullable()` (`src/types/api/eval.ts:174`), so Zod does not traverse the result and does not catch this earlier. `EvaluateSummaryV2/V3` is normally JSON-safe so this is a narrow edge case, not a regression that matters in practice. If the prior fail-open-at-GET behavior matters, wrap the clone and fall back to the un-cloned value:
  ```ts
  complete(id, result, evalId) {
    return this.update(id, (job) => {
      job.status = 'complete';
      try {
        job.result = result === null ? null : cloneResult(result);
      } catch {
        job.result = result; // fall back to reference; serialization is enforced at res.json()
      }
      job.evalId = evalId;
    });
  }
  ```
  Either way, add a test pinning the chosen behavior for a circular/`BigInt` result so the contract is explicit.

- `test/server/routes/redteam.test.ts:507-600` — The rewritten redteam `/run` `.catch` handler (`src/server/routes/redteam.ts:283-294`) is the changed behavior this PR most needs to pin down, but it has no regression test. The two new tests only resolve `doRedteamRun` via `resolveRun(undefined)`; there is no `mockRejectedValueOnce`/`reject(`/`Stack trace` assertion anywhere in the file. So the new `fail(id, [...], { append: true, resetResult: false })` path — `append: true` (a prior `logCallback` log must survive), `resetResult: false`, and the conditional `Stack trace:` line — is entirely uncovered. A regression dropping the append or the stack line would pass CI. Suggested test:
  ```ts
  mockedDoRedteamRun.mockReturnValueOnce(Promise.reject(Object.assign(new Error('boom'), { stack: 'at x' })));
  await request(app).post('/api/redteam/run').send({ config: { /* ... */ } });
  await vi.waitFor(async () => {
    const res = await request(app).get(`/api/eval/job/${jobId}`);
    expect(res.body.status).toBe('error');
    expect(res.body.logs).toEqual(expect.arrayContaining([
      'Error: boom',
      expect.stringContaining('Stack trace:'),
    ]));
  });
  ```
  Optionally append a `logCallback` line first to assert `append: true` preserves it.

- `src/server/services/evalJobService.ts:13-37` — A completed job's `result` is deep-cloned on write in `complete()` (line 30) and again on every read in `get()` via `cloneJob` -> `cloneResult` (lines 36-37), where the pre-PR code serialized the live summary exactly once at `res.json()`. `EvaluateSummaryV3.result` can be multi-MB for a full run, so this is 2-3 full traversals vs. the original 1. It is not a hot loop — the frontend polls `GET /eval/job/:id` every 1000ms (`RunTestSuiteButton.tsx:168`, `Review.tsx:514`) but clears the interval the moment status hits `complete`/`error` (`Review.tsx:482-483`), so read-time clones are bounded to ~1-2 reads. The write-time clone is intentional (the "returns defensive snapshots" test depends on it), so this is a tradeoff, not a defect. If you want to trim it, clone once — e.g. have `cloneJob` shallow-reference the already-cloned stored result for completed jobs, or drop the write-time round-trip and rely on `get()`'s clone. If the double isolation is deliberate, a one-line comment documenting why would help.

### Tests
The service itself is well covered (including the defensive-snapshot behavior). The gap is the redteam `/run` rejection path — see the Low item above; that branch's new `append`/`resetResult`/stack semantics are exactly what this refactor should lock in.

ianw-oai (approved):

mldangelo:

Verified the refactor is **behavior-preserving and safe to merge**: the old exported `evalJobs` Map is fully removed with no orphaned importers (only `redteam.ts` used it, now migrated), all three touched test files pass (51/51), and `tsc` is clean. Findings are a real test gap, one edge-case robustness item, and quality/type suggestions — no correctness bugs.

### Important

1. **The redteam `POST /run` rejection path (`.catch`) has zero test coverage** — and it's the only changed control-flow branch in the diff with no test touching it. `redteam.ts:50-64` conditionally appends a `Stack trace:` line, calls `fail(id, …, { append: true, resetResult: false })` (so a prior `logCallback` log must survive), and is gated by `currentJobId === id`. `test/server/routes/redteam.test.ts` never rejects `doRedteamRun` (no `mockRejectedValue`/`reject(`/`Stack trace`). A regression flipping `append→false`, dropping the conditional, or mismapping `error.message`/`error.stack` would silently discard in-flight logs or the stack trace with no failing test. (The eval route's equivalent failure path *is* covered in `eval.sharing.test.ts`, which makes the redteam omission the standout.) Add a test that rejects the run and asserts `status: 'error'`, the appended `Stack trace:` line, a surviving prior `logCallback` log, and the no-stack variant (asserting no `Stack trace:` line).

2. **`complete()` relocates the result-serialization failure point.** [`evalJobService.ts:47-53`] `complete()` now deep-clones via `JSON.parse(JSON.stringify(result))` *at completion time*; on `main` the summary was stored by reference and only serialized later at the GET `res.json()`. `JSON.stringify` throws on `BigInt`/circular refs, so a non-JSON-safe summary now throws inside the `.then`, lands in the `.catch`, and a **successful** eval is marked `status: 'error'` with a cryptic `String(error)` message and the real result discarded — whereas before the eval stayed `complete` and only an individual poll 500'd. Impact is narrow in practice (the summary must already be JSON-serializable for the GET endpoint's `res.json` to work, and `EvaluateSummaryV2/V3` is normally safe — functions in prompts are dropped, not thrown), so not a blocker, but worth hardening: wrap the clone in try/catch and degrade gracefully (keep `status:'complete'`, log that a field couldn't be serialized) rather than converting success into a hard error. Either way add a unit test feeding `complete()` a `BigInt`/circular result to pin the chosen behavior — the existing "function-backed prompts" test only covers the silently-lossy case, not the throwing one.

### Suggestions

- **Redundant double deep-clone of the large summary.** [`evalJobService.ts:50` + `:37`] `complete()` clones the result on write and `get()` clones the whole job (incl. result) on every read — two full `JSON.parse(JSON.stringify())` passes over the same payload. Only one isolation boundary is load-bearing (route callers never mutate the summary post-handoff). Keep the `get()` clone and drop the `complete()` clone (adjusting the "defensive snapshots" test). Also add a one-line comment on `cloneResult` explaining why it isn't `structuredClone` — eval summaries can carry `PromptFunction` values, which `structuredClone` rejects but JSON drops — so it isn't "modernized" into a break.
- **`Job` permits illegal status/field combinations.** [`src/types/index.ts:1413`] The flat type lets `status:'complete'` + `result:null` typecheck — and that state is actually produced at `redteam.ts:42` (`complete(id, summary, …)` where `summary` can be `null`). The *wire* contract for the same data is already a proper `z.discriminatedUnion('status', …)` (`src/types/api/eval.ts:165`), and the GET handler manually re-narrows by status to bridge the gap. Since this PR reshapes job access into a typed service, it's a good moment to mirror that union in the in-memory `Job` (result/evalId present only on `complete`), tighten `complete(id, result: NonNullable<…>, evalId: string)`, and route the redteam "ran but produced nothing" case through `fail()` instead of `complete(null, null)`.
- **`fail({ append, resetResult })` is always one of two fixed modes.** Every redteam caller passes `{ append: true, resetResult: false }`; only the eval route uses the defaults `{ append: false, resetResult: true }`. Two intention-revealing methods — `fail(id, logs)` (replace + reset) and `cancel(id, logs)` (append + preserve result) — read better at the call sites than a correlated boolean pair, and all four current call sites map cleanly onto exactly one.
- **Merge the duplicated `if (currentJobId === id)` blocks** in the redteam `.then`/`.catch` (`redteam.ts:41-48`, `52-63`) and extract the repeated `cliState.webUI=false; currentJobId=null; currentAbortController=null` triple (also in `/cancel:87-89`) into a small `resetRunState()` helper.
- **Avoid `invariant()` around service booleans inside async callbacks.** [`eval.ts:21,28,37`] The boolean return was designed as a non-fatal "job missing" signal, but the eval route wraps it in `invariant` inside the progress callback, the `.then`, and the terminal `.catch`. A missing job in the `.catch` (`fail()` → `false` → `invariant` throws) would become an *unhandled* rejection, and a throw in the progress callback could abort the in-flight eval. Latent today (the jobs Map is never evicted, so the job always exists), but fragile — treat a `false` as a logged no-op. Relatedly, the unbounded Map has no eviction/`delete` (slow memory growth for a long-lived server); the service is the natural home for a fix.

### Strengths

- Clean extraction: removes the cross-module `evalJobs` import from `redteam.ts` and replaces scattered `invariant(job, …)` + direct mutation with a single guarded `update()` chokepoint returning a boolean — the right primitive for "job may not exist."
- Defensive cloning on `get`/`create` genuinely prevents callers from mutating stored state (the old exported-Map design couldn't), and it's well-tested.
- Unit tests instantiate a fresh `EvalJobService` per test (isolated, order-independent); route tests exercise the real singleton through the live Express app.

**Net: approve, no blockers.** Top follow-up is the redteam `.catch` test (#1); the serialization-relocation hardening (#2) is a nice-to-have given summaries are already JSON-constrained.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: refactor(server): extract eval job service

This PR extracts the in-progress-job map that previously lived as the exported `evalJobs` map in `src/server/routes/eval.ts` into a dedicated `EvalJobService` (`src/server/services/evalJobService.ts`), routing both the eval and redteam routes through `create`/`setProgress`/`complete`/`fail`/`appendLog`/`get`. The service now stores JSON-safe snapshots on write and hands out defensive clones on read, which is a genuine correctness improvement over the base (callers used to mutate the live stored job). The refactor is clean, the `evalJobs` export is fully removed with no stale importers, and the new service has solid unit coverage. Overall this is in good shape — a couple of test-gap and simplification notes below, nothing blocking.

**Medium**

- `test/server/routes/redteam.test.ts:480-616` — The redteam route's successful-completion path is never exercised. The PR replaced the inline `job.status='complete'; job.result=summary; job.evalId=...` with `evalJobService.complete(id, summary, evalResult?.id ?? null)` inside the `.then()` handler, but every `doRedteamRun` mock in the file resolves with `undefined` (the `beforeEach` sets `mockResolvedValue(undefined)`) or stays pending/rejects. So the branch `const summary = evalResult ? await evalResult.toEvaluateSummary() : null; if (currentJobId === id) { evalJobService.complete(...) }` is never run through the route. The analogous eval route *is* covered (`eval.sharing.test.ts` "does not publish completion before saved results are ready" and "publishes progress updates while a job is running" both assert `status: 'complete'`, `evalId`, and `result`), so this is an asymmetric gap on a behavior-changing path. A regression — wrong `evalId`, dropped summary, or result not surviving `cloneResult` — would go uncaught. Fix: add a test mocking `doRedteamRun` to resolve with `{ id: 'redteam-eval-id', toEvaluateSummary: vi.fn().mockResolvedValue({ results: [] }) }`, then poll `GET /api/eval/job/:id` and assert `toMatchObject({ status: 'complete', evalId: 'redteam-eval-id', result: { results: [] } })`.

- `src/server/services/evalJobService.ts:14-31` — `cloneResult` hand-rolls the exact ancestors-stack circular-reference replacer already implemented in `safeJsonStringify` (`src/util/json.ts:101-130`), which is the codebase's standard serializer (used in ~10 modules). The only thing `cloneResult` adds is bigint→string coercion; I confirmed `safeJsonStringify` returns the correct circular-stripped output and only returns `undefined` on bigint because its `catch` special-cases `RangeError`, not the bigint `TypeError`. The duplication also drops `safeJsonStringify`'s `RangeError`/"Invalid string length" truncation fallback, so a very large eval summary that would round-trip via `safeJsonStringify` instead throws unhandled out of `cloneResult`. Suggested fix: add `if (typeof val === 'bigint') return val.toString();` to the shared `safeJsonStringify` replacer, then make `cloneResult` a thin wrapper (`JSON.parse(safeJsonStringify(result) ?? 'null')`) — single serializer, inherits the truncation fallback. (Severity is maintainability-driven; the duplicated logic is presently correct.)

**Low**

- `src/server/services/evalJobService.ts:34-40, 62-68` — The result is deep-cloned twice: `complete()` stores `cloneResult(result)` (already plain JSON-safe — functions dropped, bigints stringified, cycles removed), then every `get()` runs the same heavy custom serializer again over that already-sanitized value via `cloneJob`. On a completed job that a client polls, this re-serializes the entire `EvaluateSummary` on each request (on base, `get` returned the live reference with no copy). Real-world impact is bounded: the frontend pollers (`RunTestSuiteButton.tsx` clears its interval on `status === 'complete'`/`'error'` at lines ~146-153; `Review` similarly), and `result` is `null` while in-progress so the clone is skipped, so in normal flows the expensive clone runs only once or twice. The defensive read-side snapshot is legitimately needed, but it doesn't need the bigint/circular path since the stored source is guaranteed acyclic/JSON-safe. If you want to trim it, clone the stored result in `cloneJob` with a plain `structuredClone`/round-trip and reserve `cloneResult` for the single write in `complete()`. Optional — acceptable as-is given pollers stop on completion.

- `test/server/routes/redteam.test.ts:548-595` — The stale-run guard test ("should keep a replaced job cancelled when its stale run settles") resolves the replaced run with `resolveFirstRun!(undefined)`, so `summary` is `null` and the `currentJobId === id` guard is already false — `complete()` is never reached on either count. It verifies the guard but doesn't prove it would suppress a completion carrying a *real* result/evalId from a slow stale run (the exact clobbering scenario the guard protects against). Strengthen by resolving the stale run with a real `{ id, toEvaluateSummary }` and asserting the first job stays `status: 'error'` while the active job is unaffected. (Largely folded into the medium test-gap fix above.)

**Tests note:** The new `evalJobService.test.ts` unit coverage is thorough (defensive snapshots, bigint/circular/function-backed prompts, append-vs-replace logs, `resetResult: false` preservation, missing-job → `false`). The route-level gaps above are the asymmetry worth closing.

mldangelo-oai:
@codex review

PR #9532, opened 2026-05-29
ci: verify installed package artifact

mldangelo-oai:
## Summary

- add an install-from-tarball package artifact smoke harness
- verify packaged runtime assets, publish exclusions, ESM/CJS imports, bin linkage, and installed CLI startup
- run the harness in the Node 20.20 build lane after the clean CI build

## Validation

- `npm run build:clean`
- `NODE_OPTIONS='--max-old-space-size=8192' ./node_modules/.bin/tsdown`
- `npm run postbuild`
- `NPM_CONFIG_CACHE=/private/tmp/promptfoo-npm-cache npm run test:package-artifact`
- `npx vitest run --config vitest.integration.config.ts test/integration/library-exports.integration.test.ts`
- `npx vitest run test/package-manifests.test.ts`
- `./node_modules/.bin/biome check scripts/testPackageArtifact.ts package.json`
- `./node_modules/.bin/prettier --check .github/workflows/main.yml`
- `git diff --check HEAD`

## Baseline Notes

- The harness uses an isolated consumer npm config and explicit `https://registry.npmjs.org/` registry so workstation-level npm policy does not leak into the simulated publish consumer.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
This PR adds a `test:package-artifact` smoke harness (`scripts/testPackageArtifact.ts`) that runs `npm pack` and asserts the published tarball contains the expected runtime assets (drizzle migrations, language wrappers, tracing protos, the web app build) while excluding source maps and compiled tests, wired into CI on the Node 20.20 lane.

Overall this is a useful guardrail and the assertions look correct against the current build. No blockers. Two low-severity items worth addressing before merge: the verifier duplicates `postbuild.ts` packaging rules (drift risk), and an unrelated, internally-inconsistent `@inquirer` range change is riding along with a CI-scoped PR.

## Low / Nits

- **`scripts/testPackageArtifact.ts:94-99` (and the static list at `:23-46`) — the verifier reimplements `postbuild.ts`'s packaging rules instead of sharing them, so the two can silently drift.**

  The drizzle filter here applies `!file.endsWith('.md') && !file.includes('AGENTS') && !file.includes('CLAUDE')` to the **root-relative path**, whereas `postbuild.ts` (`DRIZZLE_EXCLUDE_PATTERNS = ['.md','CLAUDE','AGENTS']`) applies `basename.includes(pattern) || basename.endsWith(pattern)` to the **basename**. These disagree on edge cases: a future migration like `drizzle/0099_readme.md.sql` is *kept* by this script (doesn't end with `.md`) but *dropped* by postbuild (basename includes `.md`), which would turn the 20.20 lane red for a file that was intentionally excluded at publish time. It's dormant today (the only excluded drizzle files are `AGENTS.md`/`CLAUDE.md`, which both filters drop identically), and `DRIZZLE_EXCLUDE_PATTERNS` isn't exported, so the duplication is unguarded.

  Similarly, the 22-entry `requiredPackagedPaths` allowlist is hand-maintained while only drizzle and `dist/src/app` are derived dynamically. A new language wrapper or proto added to postbuild's copy tasks wouldn't be required here, so a regression dropping it would pass unnoticed.

  Suggested fix: reuse postbuild's source of truth. For drizzle, export `DRIZZLE_EXCLUDE_PATTERNS` (and ideally the basename predicate) from `scripts/postbuild.ts` and import it; at minimum match the basename logic:

  ```ts
  const basename = path.basename(file);
  return !['.md', 'CLAUDE', 'AGENTS'].some((p) => basename.includes(p) || basename.endsWith(p));
  ```

  For the static runtime-asset list, either derive the expected non-JS assets dynamically (enumerate `dist/src/**/{*.go,*.py,*.rb,*.proto}` and require each) or add a comment that it must be maintained in lockstep with postbuild's copy tasks.

- **`package.json:281-287` — unexplained, internally inconsistent `@inquirer` range narrowing rides along with a CI-only PR.**

  This PR is scoped solely to the artifact harness, but it also rewrites all seven `@inquirer/*` ranges from caret to bounded ranges with no mention in the PR body, and the resolved lockfile versions are otherwise unchanged from `main` — so this is a pure manifest-policy change, not a version bump. The caps are also internally inconsistent: `@inquirer/input` is capped below its own 5.1 minor (`>=5.0.8 <5.1.0`) while sibling `@inquirer/editor` permits 5.1.x (`>=5.0.8 <5.2.0`); the installed tree confirms the live mismatch (input `5.0.13`, editor `5.1.2`). `AGENTS.md` explicitly warns against letting unrelated dependency drift ride along with a scoped change, and there's no invariant test guarding these bounds (`test/package-manifests.test.ts` guards sharp/jsdom/parse5/OpenAI but not `@inquirer`).

  Suggested fix: revert the `@inquirer` change from this PR to keep it scoped, or split it into its own dependency PR that explains why bounding is needed and why the per-package caps differ (especially why `input` is `<5.1.0` while `editor`/`checkbox`/`select` allow `<5.2.0`). If the pins are intentional, add an invariant test in `test/package-manifests.test.ts` so the upper bounds can't be silently widened.

## Tests

The harness *is* the test, and its core assertions (excluding `.map`/`dist/test/`, requiring the web app and drizzle outputs) are sound. The two gaps above are about durability of the check itself: the duplicated packaging rules and the new `@inquirer` bounds are both unguarded, so consider the dynamic-derivation and `package-manifests.test.ts` suggestions to keep them from drifting.

ianw-oai (approved):

mldangelo:
## PR Review — pr-review-toolkit (code · tests/guard-quality · errors · simplify)

Each agent ran the harness end-to-end. **The harness is well-built and safe to merge**: exit-code plumbing is correct (a real regression throws an uncaught `AssertionError` → non-zero exit → CI red, verified by simulation), it genuinely `require()`s + `import`s the packed tarball (CJS *and* ESM), executes both bins (`promptfoo`, `pf`) for `--version`, checks the executable bit, and runs hermetically in an isolated tmpdir. CI wiring is correct — it runs *after* `npm run build`, gated on the Node 20.20 lane. The findings are a scope issue, a real silent-pass hole, and source-of-truth drift.

### Important

1. **Out-of-scope `@inquirer/*` dependency rewrite riding along — with an internal inconsistency.** [`package.json:281-287`] This PR (scoped to a CI artifact harness) also rewrites all seven `@inquirer/*` deps from carets to bounded ranges. Confirmed it's introduced by this branch, not base skew. AGENTS.md is explicit: *"If `package-lock.json` changes outside the scope of the PR, revert the drift and ship it separately."* Worse, the bounds are inconsistent across siblings that version in lockstep: `@inquirer/input` is capped `>=5.0.8 <5.1.0` while `editor`/`checkbox`/`select` allow `<5.2.0` — and `@inquirer/input@5.1.x` exists on npm, so the `<5.1.0` cap is a real (likely accidental) exclusion, not a no-op. **Fix:** revert the `@inquirer` lines here and ship the range-policy change as its own `chore(deps):` PR; if kept, make `input` consistent (`<5.2.0`) or document why it's held a minor tighter.

2. **Vacuous pass: the web-app (and drizzle) "missing files" checks succeed when the local `dist` dir is empty.** [`scripts/testPackageArtifact.ts:98-112`, `:94-97`] The check is "every file present locally in `dist/src/app` is also in the tarball." If the web-app build silently emits *zero* files, `listFiles(...)` returns `[]`, `missingWebAppFiles` is `[]`, and `assert.deepEqual([], [])` passes (reproduced empirically). So a build that drops the entire web UI would stay green; only the single hardcoded `dist/src/app/index.html` anchor is pinned, not the JS/CSS chunks. **Fix:** assert a non-empty floor before diffing — `assert(localAppFiles.length > 0, …)` and the same for the drizzle list — and ideally assert a minimum count of `dist/src/app/assets/*.{js,css}` entries directly from `packResult.files`.

3. **The harness re-encodes packaging rules that already live in `scripts/postbuild.ts`, so it silently drifts.** Highest-leverage cleanup, and a correctness issue:
   - The drizzle exclusion filter [`testPackageArtifact.ts:95`] matches on the **root-relative path** (`!file.endsWith('.md') && !file.includes('AGENTS') && !file.includes('CLAUDE')`), whereas postbuild's `DRIZZLE_EXCLUDE_PATTERNS` [`postbuild.ts:46,142-154`] matches on **basename** with `includes || endsWith`. They disagree on real cases — e.g. `drizzle/0099_readme.md.sql` is *excluded* by postbuild but treated as *required* by the harness (→ false-positive CI failure for a file that shipped correctly), and a dir like `drizzle/AGENTS_helper/x.sql` is the reverse (false negative).
   - `requiredPackagedPaths` [`:22-45`] is a hand-maintained 22-entry list; the wrapper/proto entries are exactly what postbuild's `WRAPPER_TYPES × WRAPPER_FILES` and `getProtoTask()` generate. A newly-added wrapper/proto that postbuild copies but the list omits ships unverified — the guard only catches *removal* of listed assets, not *non-shipping of new* ones.
   **Fix:** `export` `DRIZZLE_EXCLUDE_PATTERNS` + the basename predicate (and ideally the copy-task builders / `REQUIRED_BUILD_OUTPUTS`) from `postbuild.ts` and import them, so the harness validates *against* the build logic and auto-covers new assets. (`postbuild.ts` is already import-safe behind its `isMainModule` guard.)

### Suggestions

- **The `.map` / `dist/test/` / `dist/src/__mocks__/` negative assertions are effectively vacuous** [`:113-124`]. `packResult.files` is already filtered by `package.json`'s `files` field (`!dist/**/*.map`, `!dist/test`, `!dist/src/__mocks__`), so these re-check the thing that produced the data — deleting a `files` exclusion would let maps ship, but they'd then be absent from `packResult` so the assertion still wouldn't fire. To make them meaningful, verify those files exist in the *unfiltered* `dist/` tree and are absent from the packed set (proving the exclusions did the work). (The `.map` check also duplicates the web-app `.map` filter at `:99`.)
- **Minor:** the post-`npm pack` `fs.existsSync(tarballPath)` check [`:242`] guards a near-impossible state (npm only reports a filename it wrote); and the `assetPath.split(/[?#]/, 1)[0]` query/hash-strip trick [`:139`] would read better as a named `stripQueryAndHash` helper.

### Strengths

- Real consumer verification — installs the tarball into a throwaway project and loads it through both module systems with schema `safeParse`, plus end-to-end bin execution with Windows `.cmd` handling. That's well beyond "files exist."
- Executable-bit assertion (`mode & 0o111`) guards postbuild's `chmod` step — a subtle regression most tests miss.
- Hermetic and correctly gated (isolated `PROMPTFOO_CONFIG_DIR`, telemetry/update disabled, single Node-matrix entry).

**Net: approve with one scope fix.** Split out the `@inquirer` change (#1); close the vacuous-pass floor (#2); and share the packaging rules from `postbuild.ts` (#3) to stop the guard from drifting.

<sub>Review aspects: code · tests/guard-quality · errors · simplify.</sub>

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: ci: verify installed package artifact (#9532)

This PR adds an install-from-tarball smoke harness (`scripts/testPackageArtifact.ts`) that `npm pack`s the package, installs the tarball into a throwaway consumer, and asserts packaged runtime assets / publish exclusions / ESM+CJS exports / bin linkage / installed CLI `--version`. It also extracts the drizzle copy filter into an exported `shouldCopyDrizzlePath()` and wires the harness into the Node 20.20 CI lane. The harness is well-scoped, uses an isolated consumer npm config, and the assertions are precise. Overall this is solid; only minor debuggability and test-wiring nits remain.

**Note on scope:** I verified the PR's incremental diff against the parent (`mdangelo/codex/harden-architecture-boundaries`) at the current tips. It is exactly 4 files: the CI step, one `package.json` script line, `scripts/postbuild.ts`, and the new harness. There is **no** `@inquirer/*` dependency-range change and **no** `package-lock.json` change in the current branch — an earlier revision narrowed those ranges but commit `f82eb535` ("chore(deps): keep package artifact smoke focused") reverted it, and both parent tip and head tip now carry the original caret ranges. So the dependency-drift concern does not apply to this PR as it stands.

**Low**

- `scripts/testPackageArtifact.ts:79-84` — The `run()` catch wrapper does `throw new Error(\`${error.message}\n${error.stderr}\`)`. Verified empirically: for a failing child `error.message` already ends with the stderr text (Node auto-appends it), `error.stdout` holds stdout separately, `error.stderr` holds stderr. So this (a) appends stderr a second time and (b) never includes `error.stdout`. `npm pack`/`npm install` write a lot of their failure diagnostics (peer-dep conflicts, registry/ENOTFOUND errors, etc.) to stdout, so when this smoke test fails in CI the engineer sees duplicated stderr and a message missing the npm stdout that explains the failure — a debuggability regression for exactly the failure mode this harness exists to surface. Fix: build the message from the captured streams without re-appending the already-present stderr, e.g. include `error.stdout` when non-empty and set `wrapped.cause = error` to preserve `status`/`signal`/stack, or simply rethrow the original error and additionally log `error.stdout`.

- `scripts/testPackageArtifact.ts:237` — `JSON.parse(packOutput) as PackResult[]` is unguarded. With `--json` + piped stdio this is normally clean, but if npm leaks a notice to stdout or aborts after partial output, the parse throws an opaque `Unexpected token …` and buries the real npm output. Lower-likelihood sibling of the bullet above. Fix: wrap the parse and surface the raw `packOutput` on failure.

- `test/scripts/postbuild.test.ts:137-152` — This PR extracts `shouldCopyDrizzlePath()` as an exported function (`scripts/postbuild.ts:48`) but the existing "filter function for drizzle" test still defines a hand-duplicated local `filter` and asserts against that copy rather than importing the exported function. The test file is untouched by this PR, so the unit test remains tautological: changing `DRIZZLE_EXCLUDE_PATTERNS` or the match logic in `postbuild.ts` would not fail it. The real coverage comes only from the new CI-gated harness. Recommend importing `shouldCopyDrizzlePath` and asserting against it directly, and dropping the inline duplicate so the test cannot drift from the implementation. (Not blocking — the harness does exercise the real function via `assertPackagedFiles`.)

**Nits**

- `scripts/testPackageArtifact.ts:8-12` — `PackFile.size` is never read (only `.path` and `.mode` are consumed). Harmless since it documents the `npm pack --json` shape; drop it or leave a one-line comment. The `listFiles` manual recursion and inline `.split(path.sep).join('/')` normalization are fine as-is and intentionally mirror `test/package-manifests.test.ts`; no change needed.

Tests: the new harness provides good end-to-end coverage of the publish artifact. The one gap is wiring the existing `postbuild` unit test to the newly-exported function (above).

mldangelo-oai:
@codex review

PR #9534, opened 2026-05-29
refactor(models): centralize eval mutation invalidation

mldangelo-oai:
## Summary

- centralize standalone eval cache invalidation and watcher notification behind model-level mutation helpers
- migrate eval model, evaluator lifecycle, and legacy database mutation paths away from direct low-level coordination
- invalidate cached standalone plugin counts for incrementally added results and cover the behavior with regression tests

## Validation

- `npx vitest run test/models/evalMutation.test.ts test/models/eval.test.ts test/util/databaseWrite.test.ts test/util/databaseStandalone.test.ts test/util/databaseDelete.test.ts test/database/signal.test.ts test/evaluator.sigint.test.ts`
- `npx vitest run test/evaluator*.test.ts`
- `./node_modules/.bin/biome check src/models/evalMutation.ts src/models/eval.ts src/util/database.ts src/evaluator.ts test/models/evalMutation.test.ts test/util/databaseStandalone.test.ts`
- `git diff --check HEAD`

## Baseline Notes

- Biome still reports the existing high-complexity filter warning in `src/models/eval.ts`.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
This PR centralizes eval-mutation cache invalidation behind two helpers (`invalidateEvaluationCache` / `notifyEvaluationChanged` in `src/models/evalMutation.ts`) and migrates the on-disk signal file from a plain `evalId:timestamp` string to a JSON payload with a backward-compatible reader. The refactor is clean and well-tested. The notes below are non-blocking quality/scoping items — over-broad cache invalidation on a couple of hot paths, and a one-way signal-format migration cost.

**Overall: approve with nits** — all items below are non-blocking quality/scoping notes.

### Low

- `src/models/evalResult.ts:630` — `save()` now calls `invalidateEvaluationCache()` with no argument, which runs `clearCountCache(undefined)` and wipes the distinct/total count caches for **every** eval in the process (the `else` branch in `evalPerformance.ts:103-111`), plus a full `clearStandaloneEvalCache()`. On `main`, `save()` cleared nothing. This is a new side effect on a per-result hot path: a persisted eval using the select-best / max-score grading paths (`evaluator.ts:4122`, `4148`) thrashes both caches once per graded result, and every human-rating submission on a long-running view server (`server/routes/eval.ts:806`) flushes all other evals' count caches. It is functionally correct (over-invalidation only forces re-queries, never serves stale data), but the instance already knows its eval via `this.evalId` (set in the constructor at `evalResult.ts:592`), so the count-cache clear can be scoped.
  - **Fix:** Add an `evalId` param to `invalidateEvaluationCache` and pass `this.evalId` from `save()` so only this eval's counts are dropped (`clearCountCache(this.evalId)`). The standalone LRU is keyed by `(limit/tag/description)` with no per-eval granularity, so clearing it fully is unavoidable; only the count cache needs scoping.

- `src/models/eval.ts:561` — `Eval.create()` changed from `clearStandaloneEvalCache()` (on `main`) to `invalidateEvaluationCache()`, which adds `clearCountCache(undefined)` and wipes the distinct/total count caches for **all** evals on every create. A freshly created eval has no cached counts, so the only effect is forcing recomputation of unrelated evals' counts on next access. Low impact (create is not a per-result path), but broader than needed.
  - **Fix:** A new eval has no cached count, so the count cache does not need clearing here at all — calling `clearStandaloneEvalCache()` alone refreshes the list cache without nuking every other eval's counts. (Or pass the freshly created `evalId` if you prefer to keep the centralized helper.)

- `src/database/signal.ts:13-23, 31-56` — The signal-file format change (plain string → JSON `{"evalId":"...","timestamp":"..."}`) is not forward-compatible. The new reader handles both formats, so old-writer → new-reader is fine. The reverse is not: an **older** `view` server watching the same `~/.promptfoo/signal.txt` while a new CLI writes JSON runs `main`'s parser `content.includes(':') ? content.split(':')[0] : ...`. For the JSON payload that yields the garbage token `{"evalId"` (length 9 > 8), which is returned as the eval id; `Eval.findById('{"evalId"')` → `undefined`, the `if (results && results > 0)` guard never fires, and `io.emit('update')` is skipped. Net effect in a mixed-version local setup (new eval CLI + previously-installed long-lived view server sharing `PROMPTFOO_CONFIG_DIR`): no crash, but the Web UI silently stops auto-refreshing. Same-version usage is unaffected.
  - **Fix:** This is an inherent one-way migration cost. Either document the mixed-version limitation (view and eval should be on matching versions), or keep the on-disk payload readable by old parsers (e.g. retain a leading non-JSON token old readers can still `split(':')` on). At minimum add a regression note; a test asserting the old-style parser's behavior against the new payload would make the trade-off explicit.

### Nits

- `src/models/eval.ts:1275-1276` — `setResults()` calls `clearCountCache(this.id)` (line 1275) immediately before `notifyEvaluationChanged(this.id)` (line 1276). `notifyEvaluationChanged` → `invalidateEvaluationCache(this.id)` → `clearCountCache(this.id)` already clears this eval's count cache, so line 1275 is a pure no-op duplicate. Harmless but confusing given the PR's centralization goal.
  - **Fix:** Drop the standalone `clearCountCache(this.id)` on line 1275; `notifyEvaluationChanged(this.id)` covers it. The `setResults` test in `test/models/eval.test.ts` still passes (it only asserts `getCachedResultsCount` goes 0 → 1 after re-query).

---

**Tests:** Same-version behavior of the new signal format and the centralized helpers is well covered. Two gaps worth closing: (1) a regression test pinning the old-style parser against the new JSON payload to make the one-way migration cost explicit (`signal.ts`), and (2) coverage is silent on the broadened per-eval invalidation scope, but since over-invalidation is functionally correct, that is optional.

> Note for reviewers: `src/redteam/shared.ts:161` is **not** an `EvalResult.save()` caller — that `evalResult` is an `Eval` instance whose `save()` routes through `notifyEvaluationChanged(this.id)` and is already scoped.

mldangelo-oai:
@codex review

mldangelo:

Solid, well-tested refactor — the changed suites pass (148–163 tests) and there's **no staleness/correctness regression** (over-invalidation only forces re-queries, never serves stale data; the signal reader correctly handles same-version reads *and* legacy formats, including the tricky colon-in-eval-id case it's fixing). Three things worth acting on: a mixed-version signal-format regression to consciously decide on, over-broad invalidation on a per-result hot path, and a redundant call.

### Important

1. **The signal-file format change is not forward-compatible — an older `view` server silently stops auto-refreshing.** [`src/database/signal.ts:16-24`, consumed at `src/server/server.ts:389-402`] The on-disk payload changed from `evalId:timestamp` to `JSON.stringify({evalId, timestamp})`. The new reader handles old files (old-writer → new-reader ✓), but the reverse breaks. I traced `main`'s released parser against the new payload: `content.split(':')[0]` on `{"evalId":"…","timestamp":"…"}` returns the garbage token `{"evalId"` (length 9 > 8, so it passes the validity check) → `Eval.findById('{"evalId"')` → `undefined` → the `if (results && results > 0)` guard skips `io.emit('update')`. Worse, the no-evalId payload `{"timestamp":"…"}` yields `{"timestamp"` (truthy), which *defeats* the `Eval.latest()` fallback that on `main` keeps the UI live. Net: in a common mixed-version setup (a long-running older `promptfoo view` + a newer CLI writing signals to the same `~/.promptfoo/` file), the Web UI silently stops live-refreshing with **no error or log**. It's an inherent one-way migration cost, so treat it as a conscious decision rather than an oversight: either keep the on-disk payload backward-readable by released parsers (e.g. lead with the real eval id — `eval-id:<ISO>` — which the new reader already parses), or gate the JSON switch behind a transition release. Either way, add a forward-compat regression test pinning what the current writer's output resolves to under the old parser (none of the new signal tests cover this).

2. **Over-broad cache invalidation on the per-result hot path.** [`src/models/evalResult.ts:630`] `EvalResult.save()` calls `invalidateEvaluationCache()` with **no argument**, which (via `clearCountCache(undefined)`, `evalPerformance.ts:103-110`) wipes the distinct/total count caches for **every** eval in the process. `save()` runs once per result on the select-best / max-score grading paths (`evaluator.ts:4122,4148`) and once per human-rating submission (`server/routes/eval.ts:806`) — so a graded run thrashes the count cache for all other evals, defeating the 5-minute TTL exactly while the server serves table/share/count queries. Safe (no staleness), but avoidable. **Fix:** `EvalResult` already knows `this.evalId` (set at `evalResult.ts:592`) — call `invalidateEvaluationCache(this.evalId)` to scope the count clear. Important nuance: *keep* the global `clearStandaloneEvalCache()` (that's the real fix here — on `main` `save()` invalidated nothing, leaving stale standalone data; the standalone cache is keyed by filter combos, not eval id, so it genuinely can't be scoped). Same pattern in `Eval.create()` [`eval.ts:561`], which now clears all evals' count caches for a brand-new eval that has none cached.

### Suggestions

- **Redundant double-clear.** [`src/models/eval.ts:1275-1276`] `setResults()` calls `clearCountCache(this.id)` immediately before `notifyEvaluationChanged(this.id)` → `invalidateEvaluationCache(this.id)` → `clearCountCache(this.id)`. Line 1275 is a pure no-op duplicate; dropping it also lets you remove the now-unused `clearCountCache` import in `eval.ts`.
- **Type footgun: `invalidateEvaluationCache(evalId?)` where omitting the arg silently means "invalidate everything."** [`src/models/evalMutation.ts:5,10`] That `undefined` is overloaded across three layers — global invalidation (here), "no id on disk" (`signal.ts`), and "fall back to `Eval.latest()`" (`server.ts:393`) — so a *forgotten* id and a *deliberate* global blast are indistinguishable in the type, and a forgotten id on a single-eval path would silently over-invalidate (and emit an update for the wrong eval). Consider making the global case explicit (`invalidateAllEvaluationCaches()` / `notifyAllEvaluationsChanged()`) so every single-eval site must supply an id (compile error otherwise) and "who triggers a global blast?" is greppable. Relatedly, validate the signal payload with zod (it's currently `JSON.parse` → `any`; `timestamp` is written but never validated on read), matching the server's conventions.
- **Tests:** beyond the forward-compat pin (#1), add a cross-eval scope test (create two evals, mutate one, assert only its count cache is dropped — every existing count-cache test uses a single eval, so a future narrowing/broadening of `save()`'s scope would pass silently); and consider replacing the mock-only `evalMutation.test.ts` (which asserts the three collaborators were called — a near-verbatim restatement of the 6-line impl) with one real-behavior test that proves invalidation actually happens through the helper. (The headline fix *is* covered with real behavior in `databaseStandalone.test.ts:234` and `eval.test.ts:135` — good.)
- **Minor:** the "centralize all invalidation" goal isn't literally complete — `createFromEvaluateResult`/`createManyFromEvaluateResult` (`evalResult.ts:391,435`) still call `clearCountCache(evalId)` directly. Pre-existing and not wrong (their `addResult`/import callers fire `notifyEvaluationChanged` for the standalone+signal side), so optional consistency cleanup only.

### Strengths

- The signal reader correctly handles legacy eval ids that themselves contain colons (`^(.*):\d{4}-\d{2}-\d{2}T`), which is the ambiguity the JSON move fixes; backward-read of all three historical formats is covered.
- The headline behavior fix (incrementally-added results now invalidate the standalone count cache) is covered with **real** cache-state assertions, not mocks.
- The two helpers layer cleanly (`notifyEvaluationChanged` = `invalidateEvaluationCache` + signal), with correct invalidate-then-notify ordering, and the server-side consumer correctly uses the cache-only helper to avoid re-emitting the signal it's reacting to.

**Net: approve, pending a conscious call on #1.** The signal forward-compat break is the one item to decide deliberately (document the mixed-version limitation or keep the payload backward-readable); #2 (scope `save()`/`create()` invalidation) and the redundant-clear cleanup are easy wins.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: refactor(models): centralize eval mutation invalidation

This PR introduces `src/models/evalMutation.ts` (`invalidateEvaluationCache` + `notifyEvaluationChanged`) and routes all eval-mutation call sites (evaluator save/abort paths, `import`, `retry`, `eval.ts`, `evalResult.ts`, `util/database.ts`, server signal handler) through it instead of calling `updateSignalFile`/`clearCountCache` ad hoc. It also fixes a real bug in `readSignalEvalId`: the new regex `^(.*):\d{4}-\d{2}-\d{2}T` correctly extracts eval IDs that contain colons (e.g. `eval-abc-2026-05-29T22:43:41`, since `createEvalId` embeds an ISO time), where `main`'s `content.split(':')[0]` truncated them to `eval-abc-...T22`. The refactor is coherent and the colon fix is well-targeted and well-tested.

Overall: **approve with nits.** No correctness, security, or behavior regressions found in the changed code. The items below are minor maintainability/coverage nits, not blockers.

**Low / Nits**

- `src/database/signal.ts:31-39` (reader) + `:15` (writer comment) — The new `JSON.parse(content)` branch in `readSignalEvalId` is dead code: no writer in the repo emits JSON. `updateSignalFile` writes `${evalId}:${now}` (colon format), `ensureSignalFile` writes a bare ISO timestamp, and there is no `JSON.stringify` of the payload anywhere. Branch history confirms it — an intermediate commit (`faf8fd13`) switched the writer to JSON and added this read branch, then `9ffa8f81` reverted only the writer, leaving the reader half behind. For every input the codebase can produce, `JSON.parse` throws and control correctly falls through to legacy parsing, so this is harmless at runtime, but the dead branch plus the comment `// Keep writes readable by older view processes while accepting both formats below.` misrepresent the on-disk contract (only the colon/timestamp format is ever written). Note one subtle wrinkle: if a file ever did contain valid non-payload JSON, the branch returns `undefined` instead of falling through — another reason to drop it. **Fix:** delete the inner try/catch JSON block and reword the writer comment to describe only the colon/timestamp format. Keep the colon-preserving regex — that is the real, valuable change.

- `test/database/signal.test.ts:91` — The test `should return a full generated eval ID from a JSON signal payload` hand-builds a JSON string and asserts the dead read branch returns it. Because nothing writes JSON, this gives false confidence that a writer→reader round-trip works for the actual shipped (colon) format. The genuinely load-bearing case is covered by the separate `should preserve colons in eval IDs from legacy signal files` test. **Fix:** either drop this test alongside the dead branch, or replace it with a real round-trip test that calls `updateSignalFile('eval-abc-2026-05-29T22:43:41')`, captures the bytes written, feeds them to `readSignalEvalId()`, and asserts the full ID survives.

- `src/evaluator.ts:3773-3776, 3810-3813, 4228-4233` — The three new `if (!this.evalRecord.persisted) notifyEvaluationChanged(...)` guards (plus the `else` at the abort site that relies on `save()` to notify when persisted) are only exercised for `persisted: false` — every `mockEvalRecord` in `test/evaluator.sigint.test.ts` uses `persisted: false`. The persisted path is indirectly protected (`eval.test.ts` asserts `save()`/`addPrompts()` notify, and the `else` correctly avoids double-notification), so this is low risk. **Fix (optional):** add a `persisted: true` abort/save variant asserting the evaluator saves and does not double-notify, to lock the branch at the evaluator boundary.

- `test/server/server.test.ts:244` — `should invalidate process-local evaluation caches when the signal file changes` only mocks `readSignalEvalId` → `undefined` and asserts `invalidateEvaluationCache(undefined)`. The scoped path (concrete eval ID forwarded to `clearCountCache` for per-eval invalidation — the cross-process scenario the `handleSignalUpdate` comment calls out) is not covered at the server boundary. **Fix (optional):** add a case overriding `readSignalEvalId` → `'eval-12345-abcdef'` and assert `invalidateEvaluationCache` is called with that ID.

**Tests:** The new `evalMutation.test.ts` covers the module well (cache-clear ordering before signal write, scoped vs. unscoped). The colon-preservation regression and the per-call-site notify behavior are covered. Note: `import.ts` adds `notifyEvaluationChanged(evalId)` after `Eval.create` — this is correct, not a double-notify, because `Eval.create` only calls `invalidateEvaluationCache` (cache clear) and does not write the signal file, so the import path needs the explicit notify to wake watching server processes.

**Note (not a finding):** `gh pr diff` shows many unrelated files (codex-sdk, agentRubric, openai providers, azure docs, browser concurrency) because the branch is behind `origin/main` — those are base-branch divergences, not PR changes (e.g. `097ff9bab6` landed on main after this branch's last sync). Worth a `git merge origin/main` before landing so reviewers/CI see only the real diff.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ghost:
#### Generated 10 tests - 10 passed

<a href="https://app.usetusk.ai/app/testing-commit-check/2bdcf9e4-96f3-4d06-a6d4-97b332694863?client=e91e3433-8fb4-4e98-b233-bba492f4909c&action=commit" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/commit-tests-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/commit-tests-light.png">Commit tests</picture></a>&nbsp;<a href="https://app.usetusk.ai/app/testing-commit-check/2bdcf9e4-96f3-4d06-a6d4-97b332694863?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-tests-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-tests-light.png">View tests</picture></a>

> [!TIP]

#### Test Summary

- `clearEvalState` - 2 ✓
- `fetchRecentFileEvals` - 3 ✓
- `handleResultsFile` - 5 ✓

#### Results

Tusk's tests all pass cleanly. The test suite validates critical eval data flow paths: `fetchRecentFileEvals` covers the happy path, error handling with and without user-facing reporting, and state management around eval deletion and navigation. The `handleResultsFile` tests are particularly important—they verify real-time update propagation and streaming state management, ensuring the UI stays in sync with backend mutations and doesn't get stuck in loading states even when reloads fail. These are hot paths that directly impact user experience when evals are created or modified.

<details>
<summary>Avg <strong>+2%</strong> branch coverage gain across 1 file</summary>

<br>

| Source file | Line | Branch |
| ----------- | ---- | ------ |
| `src/app/src/pages/eval/components/Eval.tsx` | 87% | 72% (+2%) |

Coverage is calculated by running tests directly associated with each source file, learn more [here](https://docs.usetusk.ai/automated-tests/faqs#why-do-tusks-coverage-metrics-sometimes-differ-from-my-full-test-suite-coverage).

</details>



<sub>Last run on 2e542b2a28791c2d785034b7c7cbb23c47dbe8aa. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| ad4349c324d87237e4215a49b0e9f2ca4894dcf1 | PR identified as a refactor · [Output](https://app.usetusk.ai/app/testing-commit-check/af753c6b-53f9-4e41-8ab2-6c185d7b1fc0?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 31, 2026 1:19AM |
| 2e542b2a28791c2d785034b7c7cbb23c47dbe8aa | 10 ✓, 0 ✗ · [Tests](https://app.usetusk.ai/app/testing-commit-check/2bdcf9e4-96f3-4d06-a6d4-97b332694863?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 31, 2026 6:10AM |

</p>
</details>

_Was Tusk helpful? Give feedback by reacting with 👍 or 👎_

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ianw-oai:
Not approving this in the simple-PR sweep. One correctness edge still looks open: `updateSignalFile(evalId)` only folds pending delete signals. If two scoped updates are written inside the debounce window, the later one overwrites the earlier eval id, so a client pinned to the first eval may not refresh. Consider preserving multiple pending scoped update ids or adding a regression test for back-to-back scoped updates.

PR #9536, opened 2026-05-29
refactor(server): extract a race-safe prompt cache service

mldangelo-oai:
## Summary

Extracts the `/api/prompts` response cache out of a module-level `let allPrompts` in `server.ts` into a small `PromptCacheService` (`src/server/services/promptCacheService.ts`), and invalidates it on every eval signal the view server's watcher receives.

The service adds a `generation` epoch counter that closes a real race in the old design: a `getPrompts()` load that resolved *after* a concurrent `invalidate()` could overwrite the freshly-cleared cache and serve stale prompts indefinitely. The guard discards such a load (it is still returned to the in-flight caller, but never written to the cache), so the next request re-reads the DB.

### Relationship to #9534

This branch was opened against the pre-#9534 server, where the prompt cache was only invalidated when the updated eval had `results > 0`. #9534 (now merged to `main`) already made that invalidation unconditional. After merging `main`, the remaining value here is the **service extraction + race safety**; the conflict resolution keeps invalidation unconditional by calling `promptCacheService.invalidate()` as the first line of `handleSignalUpdate`, before any `await` and regardless of which emit branch runs.

## Changes

- **`src/server/services/promptCacheService.ts`** (new) — `PromptCacheService` (exported class + singleton) with `getAll()` / `invalidate()` and the `generation` race guard. Documents its single-threaded-event-loop assumption.
- **`src/server/server.ts`** — `/api/prompts` reads through `promptCacheService.getAll()`; the signal watcher invalidates the cache first; the module-level `allPrompts` (and now-unused imports) are removed.
- **Tests** — unit coverage for caching, the post-invalidate race, and retry-after-rejection; server-level coverage that invalidation fires on update / no-emit / delete / delete-all signals; route-level singleton-isolation hygiene in `beforeEach`/`afterEach`.

## Validation

- `npx vitest run test/server/services/promptCacheService.test.ts test/server/server.test.ts test/server/routes/serverApiSchemas.test.ts test/server/routes/serverRouteSmoke.test.ts` — 126 passing.
- `npx tsc --noEmit` — clean; `biome check` — clean.
- **End-to-end:** started the source view server on an isolated `PROMPTFOO_CONFIG_DIR`, ran successive `echo`-provider evals, and confirmed `/api/prompts` refreshed (1 → 2 → 3 prompts) after each eval signal, with no server errors.
- Merged current `main` (resolving the #9534 conflict); CI green.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: fix(server): invalidate prompt cache on eval signals

This PR extracts the ad-hoc `/api/prompts` cache out of `server.ts` into a small `PromptCacheService` and hooks its `invalidate()` into the eval signal watcher, so the prompt list refreshes whenever an eval write signal fires (not only after the post-signal `update` emit). The generation counter also guards against a stale in-flight `getPrompts()` load overwriting a freshly invalidated cache. Clean, well-scoped refactor with solid test coverage.

No blocking, high, medium, or low issues found — this looks good to merge.

**Tests**

Coverage is appropriate and targets the actual behavior change:
- `test/server/services/promptCacheService.test.ts` covers the cache-hit path, invalidation forcing a reload, and the race condition where a load resolving *after* `invalidate()` is returned to its caller but not cached (the generation-counter guard).
- `test/server/server.test.ts` adds a case asserting `promptCacheService.invalidate()` is called from the signal-watcher callback.

The race-condition test in particular is the kind of thing that's easy to omit and good to see here.

ianw-oai (approved):

mldangelo-oai:
@codex review

mldangelo:

**Clean, correct fix — safe to merge.** I traced the generation-counter guard and the wiring and ran both test files (14 pass). The guard correctly handles the case it's designed for: a `getPrompts()` load in flight across an `invalidate()` captures `generation` *before* the await and compares *after*, so the stale result is returned to its caller but **not** written to the cache (next read reloads). The real bug fix is sound too — `invalidate()` now runs at the **top** of `handleSignalUpdate` (`server.ts:381`), before the `results > 0` gate, so prompt-only writes / zero-result eval signals now refresh `/api/prompts` (the old code only nulled the cache *inside* the `results > 0` branch). No regressions, no blockers.

### Worth adding

- **Test the "rejected load isn't cached" contract.** [`src/server/services/promptCacheService.ts:9-19`] `getAll()` correctly re-throws a failed `getPrompts()` before the assignment, so a transient DB failure isn't cached and the next call retries — but that contract is untested. Per AGENTS.md ("test both success and error cases"; missing coverage for bug-fix/new behavior is P1), add:
  ```ts
  it('does not cache a rejected load and retries on next call', async () => {
    vi.mocked(getPrompts).mockRejectedValueOnce(new Error('db down')).mockResolvedValueOnce([{ id: 'ok' }] as never);
    const service = new PromptCacheService();
    await expect(service.getAll()).rejects.toThrow('db down');
    await expect(service.getAll()).resolves.toEqual([{ id: 'ok' }]);
    expect(getPrompts).toHaveBeenCalledTimes(2);
  });
  ```

### Optional hardening (pre-existing behavior — not introduced here)

- **Concurrent cold-load race.** Two simultaneous `/api/prompts` requests on an empty cache both observe `allPrompts == null`, capture the same `generation`, and both call `getPrompts()`; whichever resolves *last* wins and pins the cache — so if it read an older DB snapshot, the cache serves stale prompts until the next `invalidate()`. The generation counter guards invalidate-during-load but cannot order two loads in the same generation. This is **identical to `main`'s** prior inline cache (`if (allPrompts == null) { allPrompts = await getPrompts() }`), so it's not a regression — but if you want to close it (and the duplicate-DB-scan thundering herd), coalesce the in-flight promise and model state as a discriminated union, which also makes illegal states unrepresentable:
  ```ts
  private state: { status: 'empty' } | { status: 'loading'; promise: Promise<PromptWithMetadata[]> } | { status: 'cached'; prompts: PromptWithMetadata[] } = { status: 'empty' };
  // getAll(): return cached/loading directly; on resolve, commit only if `this.state.promise === promise`; invalidate(): state = { status: 'empty' }
  ```
  This replaces the generation counter, coalesces concurrent cold-start requests, and removes the "cached but null" illegal state.
- **Error-path logging.** The `/api/prompts` route has no try/catch (a rejected load surfaces as a raw 500, not via `sendError`), and `void handleSignalUpdate()` (`server.ts`) has no `.catch`. Both are unchanged from `main`, and since `invalidate()` now runs before the awaits that can reject, a signal-handler DB error no longer leaves the prompt cache stale (only the `io.emit('update')` is skipped). Still, optionally route the prompts-route failure through `sendError` (per `src/server/AGENTS.md`) and add `.catch((e) => logger.error(...))` to the `void handleSignalUpdate()` call so a transient failure leaves a breadcrumb.
- **Wiring test:** the `server.test.ts` case proves `invalidate()` is called from the watcher callback (good — it invokes the real captured callback), but consider also asserting `invalidate()` fires on the `results > 0` path (i.e. it's unconditional) and that `io.emit` is *not* called on the zero-results path, to make the test name's intent enforceable.

### Strengths

- The subtle part — the invalidate-during-load race — is tested with a real deferred promise (not a no-op): it asserts the stale loader still returns its value to its caller while the next `getAll()` reloads fresh. Exactly the right way to test it.
- Minimal, hard-to-misuse API (`getAll` / `invalidate`, no exposed state; class + singleton so tests use fresh instances).
- No overlap with the related #9534 — that PR invalidates the standalone/count caches; this owns the `/api/prompts` prompt-with-metadata cache. Disjoint.

**Net: approve, no blockers.** Add the rejected-load test; the in-flight-coalescing refactor is a nice optional cleanup of pre-existing behavior.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: fix(server): invalidate prompt cache on eval signals

This PR extracts the `/api/prompts` cache into a `PromptCacheService` singleton and, in the signal handler, moves cache invalidation to run **unconditionally** before the `results > 0` socket-emit gate. On `main`, `allPrompts = null` only ran inside that gate, so eval signals with zero results (e.g. an eval whose results were cleared, or an empty/zero-result eval) never invalidated the cache and `/api/prompts` served stale data indefinitely. The new `generation` counter also correctly prevents a load that resolves *after* an `invalidate()` from poisoning the cache.

**This is a correct, well-scoped fix with good tests.** I verified the regression test actually catches the bug: moving `invalidate()` back inside the gate makes `should invalidate the prompts cache when an updated eval has no results` fail (`expected to be called once, but got 0 times`). All 109 tests in the touched server suites pass with the PR applied. No blocking or correctness issues found — the items below are low-severity hardening suggestions.

### Low

- **`test/server/server.test.ts:251-266`** — The new test pins "invalidate fires for zero-result evals" but never asserts the other half of the contract this PR explicitly reshapes: that `io.emit('update', …)` is **suppressed** when `results == 0`, and conversely that both `invalidate()` *and* `io.emit('update', { evalId })` fire when `results > 0`. A regression that dropped the `if (results && results > 0)` guard (spamming socket updates on empty evals) would go uncaught. Note `socket.io` is not module-mocked in this file today (see the existing comments around lines 348-352), so asserting on `io.emit` needs a small bit of new mocking infra — worth it to lock the full gate contract with one zero-result and one positive-result case.
- **`src/server/services/promptCacheService.ts:1-27` / route tests** — The service is a process-wide singleton that survives `vi.clearAllMocks()`/`vi.resetAllMocks()`, and `serverApiSchemas.test.ts` / `serverRouteSmoke.test.ts` exercise the *real* singleton (they mock `getPrompts`, not the service). Each happens to hit bare `GET /api/prompts` exactly once today, so no stale read surfaces — but there's no reset hook and no end-to-end test of the invalidate→refresh path through the Express handler. A future second `/api/prompts` call with a different `getPrompts` return would silently get the cached value and could flake under random ordering. Consider an e2e test (GET → mutate `getPrompts` → `invalidate()` → GET, assert fresh) and/or an `afterEach` reset for files touching the real singleton.

### Nits

- **`src/server/services/promptCacheService.ts:8-18`** — `getAll()` doesn't dedupe concurrent cold-cache loads, so N simultaneous `/api/prompts` requests after an `invalidate()` each issue a `getPrompts()` query. This is **pre-existing and not worsened** (main's `if (allPrompts == null) await getPrompts()` had identical thundering-herd behavior), so it's out of scope for this PR. If you want to tidy it up while the code is fresh: cache the in-flight `Promise` (guarded by the same `generation` check) instead of only the resolved value, collapsing concurrent cold requests into one query while preserving the invalidation-during-load semantics. Optional; the current code is correct.
- **`src/server/services/promptCacheService.ts:9-18`** — The racing caller whose load resolves after `invalidate()` still returns its now-known-stale snapshot (the cache write is correctly skipped, but the value is still returned to that one HTTP client). This is explicitly intentional and codified by the `does not cache a load that resolves after invalidation` test, bounded (next request reloads), and not a regression. A one-line comment on the `return prompts;` line noting "racing caller may receive a stale-but-uncached result" would prevent a future reader from filing it as a bug.

### Tests
Coverage is solid for the core fix (cache hit, invalidate, post-invalidation stale-write drop, retry-after-rejection, and the zero-result signal regression). The gaps above are about the emit-gate contract and the singleton's end-to-end / cross-test behavior, not the fix itself.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ianw-oai (approved):
Looks good.

ianw-oai (approved):
Looks good. The prompt cache invalidation is centralized and covered for signal-driven refreshes.

ianw-oai (approved):

ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/321e7aa1-6a75-4ab6-b22c-74720c9b2e6a?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on d027ba072ab4c1a9057c789a2e9a1798f4fa3b25. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| ef04ae65f7438276a0fc7b9793d1559e22ccc43d | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/b55e4b8d-df04-461c-bb84-9913c2cbcfbe?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 31, 2026 9:31PM |
| d90fe3d0c624e1959c94148cd0f62ed25020d177 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/d35a0eec-1f42-4579-b329-fdf9e38d361c?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 12:14AM |
| d027ba072ab4c1a9057c789a2e9a1798f4fa3b25 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/321e7aa1-6a75-4ab6-b22c-74720c9b2e6a?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 12:26AM |

</p>
</details>

PR #9535, opened 2026-05-29
feat(build): publish a lightweight promptfoo/contracts subpath

mldangelo-oai:
## Summary

- publish the existing portable `src/contracts/` surface as `promptfoo/contracts`
- build lightweight ESM, CJS, and declaration artifacts without changing the root `promptfoo` entrypoint
- verify the subpath from both built output and an installed-package tarball

## Stack

Built on #9532 (the installed-package artifact harness that proves the subpath works after `npm pack` + a clean consumer install), which has since merged to `main`. This branch is now up to date with `main`.

## Validation

- `npm run build:clean && tsdown && npm run postbuild`
- `npx vitest run test/package-manifests.test.ts test/scripts/architectureUtils.test.ts`
- `npx vitest run --config vitest.integration.config.ts test/integration/library-exports.integration.test.ts`
- `npm run test:package-artifact` — packs the tarball, installs it in a clean consumer, and exercises the subpath via ESM `import`, CJS `require`, and `tsc` under NodeNext, Node16, and classic-node resolution, plus an exports-path resolution check
- `npm run architecture:check`

## Hardening (maintainer)

- Architecture leaf layers now enforce an `allowedExternal` allowlist (`contracts` → `zod`), so the leaf surface cannot pick up a stray npm dependency or Node builtin.
- The contracts size guard measures the transitive chunk closure and asserts the only external is `zod` — catching both a leaked dependency and `zod` being accidentally inlined.
- The package-artifact test validates that every declared `exports`/`typesVersions` path resolves in the installed package.

_(The earlier baseline note about `npm run tsc` reporting Anthropic `output_tokens_details` errors no longer applies — those were fixed on `main` and tsc is clean after the merge.)_

PR #9537, opened 2026-05-29
refactor(providers): lazy-load AWS family

mldangelo-oai:
## Summary

- extract the Bedrock, Bedrock Agents, and SageMaker registry factories into a coherent AWS provider family
- lazy-load the AWS family only for `bedrock:`, `bedrock-agent:`, and `sagemaker:` paths
- cover base-map immutability, all AWS load gates, concurrent lookups, and existing Bedrock behavior through the lazy registry path

## Validation

- `npx vitest run test/providers/registry.test.ts test/providers/sagemaker.test.ts test/providers/bedrock.agents.test.ts test/providers/bedrock --sequence.shuffle=false --no-file-parallelism`
- `npx vitest run test/sagemaker.test.ts test/providers/registry.optionalDependency.test.ts --sequence.shuffle=false --no-file-parallelism`
- `npm run deps:ownership -- --json`
- `npm run build:app`
- `NODE_OPTIONS='--max-old-space-size=8192' npx tsdown`
- `./node_modules/.bin/biome check src/providers/registry.ts src/providers/families/aws.ts test/providers/registry.test.ts`
- `git diff --check HEAD`

## Bundle Evidence

- The standalone bundle now emits separate `aws-*.js` and `aws-*.cjs` chunks.

## Baseline Notes

- `npm run tsc` and therefore `npm run build` still stop on the existing untouched Anthropic usage typing errors for `output_tokens_details`.
- Biome still reports the existing untouched OpenAI registry cognitive-complexity warning.


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: refactor(providers): lazy-load AWS family

This PR extracts the `bedrock:`, `bedrock-agent:`, and `sagemaker:` routing out of `registry.ts` into a dedicated `awsProviderFactories` array in `src/providers/families/aws.ts`, replacing top-level `AwsBedrockConverseProvider` / `AwsBedrockCompletionProvider` / `AwsBedrockEmbeddingProvider` imports with lazy `await import(...)` calls so the AWS SDK only loads when a bedrock/sagemaker provider is actually used. The refactor preserves the existing routing logic verbatim, and the registry tests cover each AWS sub-provider. Overall this looks good to merge; one test-quality nit below.

### Low / Nits

- `test/providers/registry.test.ts:484-495` — The new `should handle bedrock converse providers correctly` test doesn't actually prove the `converse` branch was taken. It builds the provider with `mockProviderOptions` (which sets `id: 'test-provider'`, registry.test.ts:60) and asserts only `expect(provider.id()).toBe('test-provider')`. That assertion is satisfied by the supplied `id`, not by anything converse-specific: `AwsBedrockGenericProvider`'s constructor sets `this.id = id ? () => id : this.id` (`src/providers/bedrock/base.ts:128`) as an own property, shadowing the `AwsBedrockConverseProvider` prototype's `id()` (`src/providers/bedrock/converse.ts:850-852`). Since the trailing completion fallback (`AwsBedrockCompletionProvider`, `aws.ts:67-69`) extends the same base, it also returns `'test-provider'` for this input. I confirmed empirically that rerouting the converse branch to the completion fallback leaves this test green — so it would not catch a routing regression on the `modelType === 'converse'` branch (`aws.ts:15-17`). The sibling bedrock-agent/agents/kb tests (registry.test.ts:497-530) do this right: they omit `id` and assert each provider's real default id.

  Make the assertion converse-distinguishing — either drop `id` from the options and assert the converse default id, or assert the instance type:

  ```suggestion
      const provider = await factory!.create(
        'bedrock:converse:anthropic.claude-v2',
        { config: {} },
        mockContext,
      );
      expect(provider.id()).toBe('bedrock:converse:anthropic.claude-v2');
  ```

  (Alternatively, keep the existing options and add `expect(provider).toBeInstanceOf(AwsBedrockConverseProvider)`.) Either change makes the test fail if the converse branch regresses to the completion fallback.

### Tests

Coverage for the routing extraction is otherwise solid — each AWS sub-provider (bedrock-agent, agents, kb, nova-sonic, sagemaker, legacy/back-compat) has a dedicated test that asserts its real default id. Only the converse case has the tautological assertion noted above; fixing it brings it in line with its siblings.

mldangelo-oai:
@codex review

mldangelo:
## PR Review — pr-review-toolkit (code · tests · errors · simplify)

**Clean, behavior-preserving lazy-load refactor — safe to merge.** Verified: the `bedrock:`/`bedrock-agent:`/`sagemaker:` routing moved **verbatim** into `families/aws.ts` (every branch — converse/nova-sonic/luma/nova-reel/agents/completion/embeddings/kb/fallback — plus the `'./bedrock/…'`→`'../bedrock/…'` dynamic-import path rewrites, all resolving correctly); `registry.ts` lazily `await import('./families/aws')` behind the `isAwsProviderPath` gate whose three prefixes exactly match the factories' `test`s, so the AWS modules stay unloaded until an AWS provider is requested. tsc is clean and all 52 registry tests pass. No SDK-error regression (the removed imports were type-only; the SDK has always loaded lazily at call time with an "install @aws-sdk" message). No blockers.

### Important (test quality)

- **The `bedrock:converse` routing test is tautological — and the obvious fix doesn't work either.** [`test/providers/registry.test.ts:484-495`] It builds the provider with `mockProviderOptions` (`id: 'test-provider'`) and asserts `provider.id() === 'test-provider'`, which the base constructor (`bedrock/base.ts:128`) satisfies from the supplied `id` regardless of branch. Crucially, even the sibling pattern (omit `id`, assert the default id) wouldn't distinguish it: I verified the converse branch and the completion **fallback** both yield `bedrock:converse:anthropic.claude-v2`. So a regression rerouting `converse` → fallback passes silently. Fix by asserting the class:
  ```ts
  import { AwsBedrockConverseProvider } from '../../src/providers/bedrock/converse';
  const provider = await factory!.create('bedrock:converse:anthropic.claude-v2', { config: {} }, mockContext);
  expect(provider).toBeInstanceOf(AwsBedrockConverseProvider);
  ```
  Same class of gap for two siblings: **`sagemaker:embedding:…`** [`:541-553`] asserts an id that's identical for `SageMakerEmbeddingProvider` and `SageMakerCompletionProvider` (so a reroute to completion passes — and embedding's `callApi` *throws*, so that's a real user-visible break), and **`bedrock:nova-sonic`** [`:532`] asserts only `toBeDefined()`. Recommend switching routing-discrimination tests to `toBeInstanceOf(<ExpectedProvider>)`, reserving `id()` for model-name *parsing* assertions. (The luma/nova-reel/agents/kb/bedrock-agent tests already do this correctly — the luma test even has a `not.toContain('nova-reel')` negative assertion.)

### Suggestions

- **Wrap the AWS factories' `create()` in `withErrorContext`, matching the redteam sibling.** [`src/providers/families/aws.ts`] The redteam family (`src/redteam/providers/registry.ts:131`) wraps each factory `create` so a failed dynamic import surfaces with the user's provider path; the AWS family doesn't. `getProviderFactories` wraps the family-*load* but not the individual `create()` calls — and the AWS creates are full of optional-peer-dep `await import('../bedrock/{nova-sonic,luma-ray,nova-reel,agents,knowledgeBase}')` / `'../sagemaker'`, exactly the case `withErrorContext` exists for. `withErrorContext` is currently file-local to the redteam module; promote it to a shared location and apply it in both families so they're structurally identical and AWS peer-dep failures carry the provider path.
- **Add an AWS optional-dependency regression test** mirroring `test/providers/registry.optionalDependency.test.ts`: assert the package still imports with `@aws-sdk/client-bedrock-runtime` mocked-absent, and that creating + invoking a `bedrock:` provider surfaces the "package is required as a peer dependency" message. This guards against a future edit turning a type-only `@aws-sdk` import into a runtime one — which would break the public entrypoint for every non-AWS user.
- **Minor:** the `|| 'luma.ray-v2:0'` fallback [`aws.ts:32-34`] is unreachable (when the block is entered via `modelType.includes('luma.ray')`, `splits.slice(1).join(':')` is never empty); table-drive the four near-identical single-provider bedrock `create()` tests (matching the `it.each` the PR already uses for sagemaker); and note the eager/lazy import split is "common-path vs niche/optional-peer-dep," not "light vs heavy" (all bedrock modules load the SDK lazily internally) — so don't claim a per-provider bundle-weight win; the real win is the whole family being lazy behind the gate.

### Process note

This PR is **stacked on `#9535` (which is itself behind `main`)**, so `gh pr diff 9537` is muddied with ~1,200 lines of unrelated churn (modelaudit docs/schemas, contracts, package-manifests, architectureUtils). The actual change is only the 3 AWS files. Rebasing #9535 (or this branch) onto current `main` will clean up the diff so reviewers/merge see only the intended change.

### Strengths

- Routing moved verbatim with no logic/ordering/default drift; the lazy `await import` genuinely defers AWS module load behind a cheap synchronous gate.
- `getProviderFactories` wraps a failing family load with the provider path + `cause` — strictly better than the old eager import (which would have crashed the whole registry at startup for *all* users).
- The family-boundary tests are excellent: they pin `providerMap` identity on the hot path, prove AWS factories append without mutating the shared map, and verify concurrent lookups resolve a stable factory.

**Net: approve, no blockers.** Tighten the converse/nova-sonic/sagemaker-embedding routing tests to `toBeInstanceOf` (the highest-value item), and consider `withErrorContext` parity with the redteam family.

<sub>Review aspects: code · tests · errors · simplify.</sub>

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
This PR extracts the Bedrock / Bedrock-Agent / SageMaker registry factories out of `registry.ts` into a new `src/providers/families/aws.ts` and lazy-loads them via the `ProviderFamily` mechanism, keyed on `bedrock:` / `bedrock-agent:` / `sagemaker:` paths. The factory bodies are moved **verbatim** (only `./bedrock/...` → `../bedrock/...` import paths change), and new tests cover base-map immutability, the AWS load gates, and concurrent lookups. The refactor is correct and behavior-preserving; my only substantive note is that the new SageMaker test under-asserts the one piece of logic the move most needs to protect.

**Medium**

- `test/providers/registry.test.ts:542-545` — The new SageMaker `it.each` matrix asserts only `provider.constructor.name`, which is identical (`SageMakerCompletionProvider`) for the `jumpstart`, `openai`, and no-model-type cases, so it pins almost nothing. The load-bearing behavior of the SageMaker factory (`src/providers/families/aws.ts:99-121`) is *injecting `config.modelType`*: the jumpstart branch forces `modelType: 'jumpstart'`, the generic branch forwards `splits[1]` (e.g. `'openai'`), and the `splits.length === 2` branch forwards nothing. Because the factory constructs the provider with the bare endpoint (`new SageMakerCompletionProvider('endpoint-name', …)`), the provider's id is `sagemaker:endpoint-name` — a single colon — so the `id()`-recovery regex `^sagemaker:(?<modelType>.+):.+$` (`src/providers/sagemaker.ts:411`) cannot match, and the resolved `modelType` comes *entirely* from the factory-injected config. A regression that swapped the jumpstart/generic branches or dropped the injection would pass every current assertion. `SageMakerCompletionProvider.modelType` is a public `readonly` field (`src/providers/sagemaker.ts:398`), so it's directly observable. **Fix:** add an `expectedModelType` column and assert `expect((provider as SageMakerCompletionProvider).modelType).toBe(expectedModelType)` — `sagemaker:jumpstart:endpoint-name` → `'jumpstart'`, `sagemaker:openai:endpoint-name` → `'openai'`, `sagemaker:endpoint-name` (config `modelType: 'custom'`) → `'custom'`.

**Low / nits**

- `src/providers/families/aws.ts:108-117` (test-gap) — The jumpstart branch matches on `endpointName.includes('jumpstart') || modelType === 'jumpstart'`, but the matrix only exercises the `modelType === 'jumpstart'` half. The `endpointName.includes('jumpstart')` half (e.g. `sagemaker:custom:my-jumpstart-endpoint`, where the endpoint name overrides the `custom` segment and forces `modelType: 'jumpstart'`) is never hit. Worth a case asserting that precedence — naturally folds into the modelType assertion above.
- `src/providers/families/aws.ts:47,100` (test-gap, nit) — Both factories accept plural embedding aliases (`embeddings`/`embedding`), but all tests use the singular form, so the backwards-compat plural alias could be dropped silently in a future refactor. Add `bedrock:embeddings:…` and `sagemaker:embeddings:endpoint-name` cases.
- `src/providers/families/aws.ts` (consistency, optional — **not a regression**) — The sibling redteam family exports `rawRedteamProviderFactories.map(withErrorContext)` so `create()`-time dynamic-import/constructor failures surface with the requested provider path; the AWS family exports `awsProviderFactories` raw. `getProviderFactories`'s try/catch only wraps `family.factories()` (the module import), not the per-factory `create()` calls, so a broken bedrock/sagemaker submodule import yields an uncontextualized error for AWS where redteam would be contextualized. The original inline registry factories were also unwrapped, so this PR does not worsen anything — flagging only as a missed-reuse consistency gap. Doing it cleanly would require hoisting `withErrorContext` out of `src/redteam/providers/registry.ts` to a shared location to avoid an aws→redteam coupling; deferring is defensible.
- `test/providers/registry.test.ts:154-167` (nit) — `resolves the same AWS factory under concurrent lookups` largely exercises Node's ESM module cache + the module-scoped const array rather than PR logic (`getProviderFactories` has no memoization), duplicating the existing redteam concurrency test. It does catch turning the family export into a per-call getter, so keeping it is fine, but consider adding the same clarifying comment the redteam test carries so it doesn't read as testing dedup logic that doesn't exist.

**Tests note:** The new coverage is a real improvement (immutability, all load gates, concurrency). The one concrete ask is to tighten the SageMaker matrix to assert the routed `modelType` rather than the class name — that's the contract the extraction most needs to lock in.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ianw-oai:
I’m not approving this one yet: in this PR alone, `getProviderFactories()` appends lazy AWS factories after `providerMap`, so an AWS/SageMaker provider ID ending in `.js`/`.ts` can hit the generic script provider before the AWS family. The later Google-family PR appears to fix this by prepending family factories; this PR should preserve that AWS-before-script precedence too, ideally with a regression test.

PR #9538, opened 2026-05-29
fix(evaluator): preserve and harden programmatic JSONL output

mldangelo-oai:
## Summary

Programmatic `evaluate({ outputPath: 'x.jsonl' })` without `writeLatestResults` (the default) produced an **empty** file: rows were streamed to disk during evaluation, then the post-run exporter truncated the file and rebuilt it from database-backed batches — which are empty for a non-persisted eval. This makes the streamed-then-finalized JSONL flow correct for both CLI and programmatic evals, and hardens finalization.

### Core fix
- `Eval.fetchResultsBatched()` yields the in-memory results for non-persisted evals, so the post-run rewrite has rows to write.
- `.jsonl` is no longer filtered out of the post-run paths, so the file is finalized from the completed eval — picking up timeout rows and deferred `max-score`/`select-best` grading that never stream.
- The evaluator closes its JSONL writers in `run()`'s `finally` (success and error paths) before the post-run rewrite reads them, aggregating close errors so they never mask the real eval/cleanup error.
- Every finalized row is sanitized at the artifact boundary (`sanitizeResultForJsonlArtifact`): credential HTTP headers are redacted and `PROMPTFOO_STRIP_*` projections applied, matching DB-persistence redaction — including on the non-persisted path.
- `findManyByEvalIdBatched` is sparse-index-safe; `toEvaluateResult()` reconstructs per-row `tokenUsage`, mirroring the live accounting.

### Persistence-failure recovery
On a mid-run DB persistence failure, JSONL is reconciled from the streamed rows + eval record + in-memory final rows, and a truncated streamed line is skipped defensively rather than aborting finalization. A row that failed to persist is still fed into `max-score`/`select-best` comparison so the winner is chosen over the full output set.

**select-best + max-score composition:** a row that failed to persist is reconstructed **once and reused** across comparison passes, so `max-score` composes on top of the `select-best` result instead of rehydrating the stale pre-comparison row (which previously erased the `select-best` failure and could report `success: true`). This mirrors how persisted rows compose grading via `save()` + re-fetch.

### Finalization hardening
- JSONL artifacts are rewritten **atomically** (build a sibling temp file, then `rename` over the destination), **preserving the destination's permission bits** across the swap — an interrupted finalization can't leave a truncated/empty file, and a restricted (e.g. `0600`) reused path isn't silently widened.
- `JsonlFileWriter` has a persistent `error` listener (an error emitted between writes can't escape as an unhandled event) and attaches the output path to both `write()` and `close()` rejections.
- A single `redactSensitiveResultFieldsForDb` helper is shared by DB persistence (`createFromEvaluateResult` / `createManyFromEvaluateResult`) and the JSONL artifact boundary, so the credential-header redaction contract can't drift between paths.

## Tests
- `test/evaluate.jsonl.test.ts`: end-to-end — timeout rows, max-score rewrite (incl. a row that failed to persist), **select-best→max-score composition for an unpersisted row**, non-persisted + persisted redaction, uppercase `.JSONL`, `PROMPTFOO_STRIP_*` projections, persistence-failure recovery.
- Unit tests: writer truncate/append/close (incl. close-error rejection, late-error swallow, path-attributed errors), sparse batching, the non-persisted `fetchResultsBatched` branch, `toEvaluateResult` token accounting, failed-row **instance reuse + cache invalidation**, **atomic permission preservation**, and malformed-streamed-line recovery skip.

## Validation
- `npx vitest run` across the touched evaluator / models / util suites · `npm run tsc` clean · lint/format clean
- CLI + programmatic E2E: non-persisted `.jsonl` redacted; rerun-to-same-path truncates **and preserves file permissions**; uppercase `.JSONL` and mixed `-o json -o jsonl` finalize; deferred `max-score` grading reflected on disk; no stray temp files
- CI green across all shards (incl. Windows)


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: fix(evaluator): preserve programmatic JSONL output

This fixes the bug where programmatic (non-CLI) `.jsonl` output ended up empty: `Eval.fetchResultsBatched()` now yields the in-memory `this.results` for non-persisted evals, `JsonlFileWriter` truncates (`flags: 'w'`) unless resuming, the evaluator `close()`s its writers in the `finally`, and the CLI `doEval` path filter no longer drops `.jsonl` so the artifact is finalized (truncate-then-rewrite) from the completed eval. The core change is sound and well-covered by the new `test/evaluate.jsonl.test.ts` (synthetic timeout rows, max-score rewrite, persisted-path redaction) plus the writer close/append/non-persisted-batching unit tests. Overall this looks good to merge — the items below are low-severity follow-ups, not blockers.

### Low / Nits

- `src/models/evalResult.ts:383-393` — **Programmatic (non-persisted) JSONL output now writes echoed HTTP headers to disk un-redacted.** Header redaction (`sanitizeResponseForDb` / `sanitizeGradingResultForDb` / `sanitizeMetadataForDb`, which strip `Authorization`/`Bearer`/cookie headers echoed back in `response.metadata.http.headers` / `requestHeaders` and grading metadata) is applied only inside the `if (persist)` branch (line 383); the non-persisted branch returns `new EvalResult(args)` with un-redacted args (line 393). With this PR, `fetchResultsBatched()` yields those in-memory rows for non-persisted evals, so a programmatic `evaluate({ outputPath: 'out.jsonl' })` (default `writeLatestResults: false`) now fills the file with rows that can contain echoed credentials — whereas before the PR that file was empty. This mirrors the pre-existing programmatic JSON behavior (also non-persisted, also un-redacted), so it's a pre-existing asymmetry the PR *surfaces for JSONL* rather than a JSONL-specific defect, and impact is bounded — it's the caller's own output file on their own machine, and persisted/CLI runs (the common case) are fully redacted via the DB round-trip. Note the new redaction test (`test/evaluate.jsonl.test.ts` → "exports persisted rows with sensitive HTTP metadata redacted") exercises only the **persisted** path (`writeLatestResults: true`), so the non-persisted asymmetry is untested.
  - Suggested: either apply the same header redaction to non-persisted results before JSON/JSONL serialization (the helpers `redactHttpHeadersOnMetadata` / `redactHttpHeadersOnGradingResult` are currently file-local in `evalResult.ts` and would need exporting, or apply a shared sanitize call), or at minimum document that programmatic (`writeLatestResults=false`) JSON/JSONL output is not header-redacted and add a regression test asserting echoed `Authorization`/`Bearer`/cookie headers don't leak into non-persisted JSONL output.

- `src/commands/eval.ts:886-890` — **The CLI path-filter change (now rewriting `.jsonl` post-run) has no regression test.** Removing `&& !p.endsWith('.jsonl')` from the `doEval` paths filter is the behavior change unique to the CLI: `.jsonl` outputs now flow into the post-run `writeMultipleOutputs` (truncate-and-rewrite from the completed eval, reflecting deferred grading / timeout rows / persisted redactions) instead of relying solely on rows streamed during evaluation. But `test/commands/eval.test.ts` has no coverage of output-path filtering (grepping it for `jsonl`/`writeMultipleOutputs`/`outputPath` returns only unrelated SIGINT matches). Risk is low (one-line predicate, downstream machinery is covered for persisted evals), but this CLI behavior change itself is untested.
  - Suggested: add a focused `doEval` test (or extend an existing one) asserting a mixed `outputPath` like `['out.json', 'out.jsonl']` results in `writeMultipleOutputs` being called with both entries — i.e. `.jsonl` is no longer filtered out.

- `test/index.test.ts:563-572` (nit) — The new "should finalize JSONL files streamed during evaluation" test asserts only `expect(writeOutput).toHaveBeenCalledWith('test.jsonl', expect.any(Eval), null)`. But `writeOutput`/`writeMultipleOutputs` are mocked at the top of the file (`vi.mock('../src/util')`), and the programmatic `evaluate()` under test lives in `src/index.ts`, which **this PR does not modify** — so the output *routing* it checks is identical on pre-fix `main`. This test therefore passes against `main` without the fix and exercises none of the actual fix (the `fetchResultsBatched` non-persisted branch or the truncate-then-rewrite). The mixed-array edit at lines 585-595 (adding `test2.jsonl`) is the same situation. The genuine coverage already lives un-mocked in `test/evaluate.jsonl.test.ts` and `test/models/eval.test.ts` ("returns in-memory results in batches for non-persisted evals"), so the PR as a whole is well-covered — this specific assertion just adds little regression value.
  - Suggested: drop the near-no-op assertion, or add a comment clarifying it only guards output routing (not finalization).

### Tests

`test/evaluate.jsonl.test.ts` gives solid end-to-end coverage (timeout rows, max-score rewrite, persisted-path redaction), and the evaluator unit tests cover writer `close()` (stale-row truncation) and the `cliState.resume` append path. Two real gaps remain: the **non-persisted** redaction asymmetry (the redaction test only uses `writeLatestResults: true`) and the **CLI** `doEval` path-filter change — both noted above.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## PR Review — pr-review-toolkit (code · tests · errors · simplify)

The core fix is **mechanically sound** — I verified the write ordering end-to-end: `close()` is awaited in `run()`'s `finally` before the post-run rewrite reads results, `output.ts` truncates-then-rewrites so streamed rows are fully replaced (no duplication/interleaving), and deferred `max-score`/`select-best` grading is reflected because the rewrite reads the mutated in-memory `EvalResult` instances. The timeout-row and max-score tests drive the real `evaluate()` and assert on-disk content (right altitude). But there's a security regression to address, plus error-handling robustness and test gaps.

### Important

1. **[security] Programmatic (non-persisted) JSONL/JSON now writes un-redacted HTTP headers to disk — newly introduced by this PR.** [`src/models/eval.ts:725-729`, root cause `src/models/evalResult.ts:383-393`, sink `src/util/output.ts:440-443`] Header redaction (`sanitizeResponseForDb`/`sanitizeGradingResultForDb`/`sanitizeMetadataForDb`, which strip `Authorization`/`Bearer`/`set-cookie` from `response.metadata.http.headers`/`requestHeaders`) runs **only** inside `createFromEvaluateResult`'s `if (persist)` branch; the non-persist branch returns `new EvalResult(args)` with raw values. This PR's new `fetchResultsBatched` non-persisted branch yields those raw rows. The exposure is *new*, not pre-existing: on `main`, `fetchResultsBatched` had no non-persisted branch, so for a non-persisted eval it queried the DB (no rows) and `output.ts`'s truncate-then-rewrite produced an **empty** file. After this PR, the rewrite repopulates from un-redacted in-memory rows, so a default `evaluate({ outputPath: 'x.jsonl' })` goes from empty-file → echoed credentials on disk (and the same now affects non-persisted `.json`). Impact is bounded — it's the caller's own machine/credentials, and persisted/CLI runs stay redacted via the DB round-trip — but `.jsonl` files routinely get committed to repos or uploaded as CI artifacts, and this bypasses the deliberate redaction boundary (`SENSITIVE_RESPONSE_HEADER_NAMES`). Per the repo's "credential exposure" review priority this is the top item.
   - **Fix:** redact at the export/serialization boundary in `output.ts` (apply `redactHttpHeadersOnMetadata` / `redactHttpHeadersOnGradingResult` to `result.response.metadata` / `result.gradingResult` before `JSON.stringify`). That keeps the documented design where in-memory rows stay real for hooks (`evalResult.ts:173-175`) while sanitizing the disk artifact. (Those helpers are currently file-local in `evalResult.ts` and would need exporting.) Then add a regression test.

2. **Test gaps.**
   - The redaction test [`test/evaluate.jsonl.test.ts`] only uses `writeLatestResults: true` (the persisted path that was *already* redacted), so the newly-exposed non-persisted path (#1) is untested. Add the same test **without** `writeLatestResults` and assert no `Bearer`/`set-cookie`/`x-request-id` reaches the file (P1 per AGENTS.md for security-sensitive behavior).
   - The one CLI-observable change unique to this PR — removing `&& !p.endsWith('.jsonl')` from `doEval`'s path filter so `.jsonl` now flows into `writeMultipleOutputs(paths, …, shareableUrl)` — has **zero** coverage in `test/commands/eval.test.ts` (no `jsonl`/`writeMultipleOutputs`/`outputPath` references). Add a `doEval` test asserting a mixed `outputPath` like `['out.json','out.jsonl']` calls `writeMultipleOutputs` with both entries (and the resolved `shareableUrl`).
   - The new `test/index.test.ts:563-571` "finalize JSONL" test asserts only that the mocked `writeOutput` was called with `'test.jsonl'`. But the programmatic `evaluate()` it exercises lives in `src/index.ts`/`src/evaluate.ts`, which **this PR doesn't modify** — so it asserts pre-existing routing and passes on pre-fix `main`, exercising none of the fix. Re-home it as an on-disk assertion (like `evaluate.jsonl.test.ts`) or drop it.

### Suggestions

- **`JsonlFileWriter.close()` swallows stream errors** [`src/util/exportToFile/writeToFile.ts:27-31`] — it resolves on the `end()` callback with no `'error'` listener, and the PR makes it newly load-bearing in `run()`'s `finally`. A disk-full/`EACCES` mid-write would either silently truncate the file (caller trusts it as complete) or surface as an unhandled stream `'error'` that crashes *during* the `finally`, masking the real eval result. Add an `'error'` listener that rejects (matching the `logger.ts:536` close pattern), and at the call site (`evaluator.ts:4638`) use `allSettled` or catch+log so a close failure neither masks the eval error nor passes silently. Same class: the streaming `write()` loop (`evaluator.ts:3151`) sits outside `persistEvalRow`'s try/catch, so a per-row write rejection aborts the eval with an opaque error.
- **Minor reuse:** `readJsonl` in the new test is a third copy (also inline in `test/util/jsonlOutput.test.ts` and `test/test-hygiene.test.ts`) — promote it to `test/util/utils.ts` and import it. And the `os.tmpdir()` path-building + manual `outputPaths` cleanup (duplicated across both touched test files) could reuse the existing `createTempDir`/`removeTempDir` helpers. (The non-persisted batch loop in `fetchResultsBatched` is fine as-is — no chunk util exists and the offset-stepping matches repo convention.)

### Strengths

- Write ordering is correct: `close()` flushes (resolves on `'finish'`) before the rewrite, the rewrite truncates-then-repopulates for a single canonical row set, and deferred grading is reflected. No double-write or race despite the streamed-then-rewritten flow.
- The timeout-row and max-score tests use the real `evaluate()` and assert actual on-disk JSONL content; the `fetchResultsBatched` non-persisted batching test correctly verifies batch boundaries.

**Net: approve pending the redaction fix (#1)** — redact non-persisted output at the serialization boundary and add the non-persisted + `doEval` tests. The `close()` error-handling hardening is a worthwhile follow-up.

<sub>Review aspects: code · tests · errors · simplify.</sub>

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: fix(evaluator): preserve programmatic JSONL output

This PR fixes a real regression where non-persisted programmatic `evaluate()` calls produced an **empty** `.jsonl` file (the post-run exporter truncated and rebuilt from DB-backed batches that didn't exist for in-memory evals). It does so by: closing streamed JSONL writers in the evaluator `finally`, adding a non-persisted branch to `fetchResultsBatched`, routing both CLI and programmatic output through a new `filterOutputPathsAfterStreaming`, sanitizing streamed rows via `sanitizeResultForJsonlArtifact`, and switching `JsonlFileWriter` to truncate-by-default (append on `--resume`). The core fix is sound and the JSONL header-redaction leak from the prior review is genuinely closed for `.jsonl` (verified: the streamed and rewrite paths both run `sanitizeResultForJsonlArtifact`, and all 4 new tests pass). One confirmed data-field regression on the CLI path is worth fixing before merge.

### Medium

- **`src/util/output.ts:461-468` (+ `src/commands/eval.ts:888-993`) — CLI JSONL loses top-level `tokenUsage`.** On `main`, the CLI dropped `.jsonl` from the post-run rewrite, so the on-disk artifact was the *streamed* row — a raw `EvaluateResult` carrying top-level `tokenUsage`. This PR keeps `.jsonl` in the rewrite, which serializes `toEvaluateResult()` (`src/models/evalResult.ts`), and that method does **not** emit a top-level `tokenUsage`. Verified empirically: a streamed-only row has `tokenUsage` (`STREAM_HAS_TOP_TOKENUSAGE=true`) while the finalized row does not (`HAS_TOP_TOKENUSAGE=false`). `response.tokenUsage` is still present, so per-response cost data isn't fully lost, but any tooling reading the documented top-level `EvaluateResult.tokenUsage` from CLI JSONL exports will silently get `undefined` where it previously had data. The new tests miss this because they use `expect.objectContaining(...)`. **Fix:** add `tokenUsage: this.tokenUsage` (or `this.response?.tokenUsage`) to `EvalResult.toEvaluateResult()`, or merge it back in the JSONL branch; add an assertion (`toEqual` on the field, not `objectContaining`).

### Low

- **`src/util/output.ts:461-468` — non-atomic truncate-then-rebuild.** The JSONL branch does `writeFile(outputPath, '')` then per-batch `appendFile`. An interruption (SIGINT/OOM/throw) between truncate and the final append leaves the file empty/partial. This is newly applied to the CLI path (the streamed file was never re-truncated on `main`). Bounded impact: for persisted evals the rows are still in SQLite and re-exportable, and the sibling `writeJsonOutputSafely` is likewise non-atomic, so this is consistent with the existing pattern rather than a new class of risk. Optional hardening: write to a sibling temp file and `rename()` over `outputPath` once the rebuild completes.

- **`test/commands/eval.test.ts` — the CLI `doEval` path change has no regression test.** The PR removes the unconditional `!p.endsWith('.jsonl')` filter and wires `filterOutputPathsAfterStreaming` into the primary CLI entry point, but `test/commands/eval.test.ts` has zero assertions on `outputPath`/`writeMultipleOutputs`/`filterOutputPathsAfterStreaming`, and it doesn't mock `src/util`, so a real test is feasible. The filter logic itself is covered by `test/util/output.test.ts` and the behavior end-to-end by `test/evaluate.jsonl.test.ts` (programmatic only), so this is a coverage gap on the CLI surface rather than untested logic. Suggest a `doEval` case asserting `.jsonl` is kept in `writeMultipleOutputs` on success and dropped when `resultPersistenceFailed` is set.

### Nits

- **`src/util/exportToFile/writeToFile.ts:27-31` — `close()` swallows flush errors.** `this.writeStream.end(resolve)` registers no `error` listener; a final-flush failure (e.g. ENOSPC) is discarded or surfaces as an unhandled `error` event. This PR is the first caller of `close()` (the evaluator `finally`), so it newly exercises the swallow. Suggest `this.writeStream.once('error', reject)` + `end(() => resolve())`, including `filePath` in the rejection context.

- **`src/util/output.ts:54-55` — dead fallback in `toEvaluateResult` helper.** Both branches of `fetchResultsBatched()` yield `EvalResult[]` (the in-memory branch slices `this.results: EvalResult[]`; the persisted branch maps to `new EvalResult(...)`), so the `: result` fallback for a bare `EvaluateResult` is unreachable. Harmless defensive code; can be narrowed to `(result: EvalResult)` and inlined.

- **`test/evaluate.jsonl.test.ts:503-533` — streamed-only test only pins `set-cookie`.** The streamed-path test sets/asserts only response `set-cookie`; request `authorization`/Bearer redaction is asserted only on the rewrite-path test (454-501). Since both paths share `sanitizeResultForJsonlArtifact`, the logic is covered — but adding `requestHeaders: { authorization: 'Bearer ...' }` to the streamed-failure test would pin the exact header the prior review flagged as a standalone guard.

### Out of scope (not a blocker for this PR)

- **`.json`/`.yaml`/`.txt` exports still write un-redacted `Authorization`/`Set-Cookie` headers for non-persisted programmatic evals.** Verified empirically: a non-persisted `evaluate({ outputPath: '*.json' })` writes both `session=...` and `Bearer sk-...` verbatim. Root cause is that header redaction lives only in `createFromEvaluateResult`'s `persist` branch, so the in-memory `EvalResult` returned for non-persisted evals keeps raw headers, and `toEvaluateSummary()` → `createOutputData()` reads them. This is **pre-existing on `main` and not worsened by this PR** (the JSON export path is untouched), so it shouldn't gate this change — but since the PR closes the identical leak for `.jsonl`, a follow-up routing all non-persisted exports through the same sanitizer (or running the redaction passes in the non-persist branch) would close the gap and avoid a surprising `.jsonl`-vs-`.json` inconsistency.

### Maintainability (optional)

- `sanitizeResultForJsonlArtifact` re-implements the per-field redaction pipeline already present in `createFromEvaluateResult` (persist branch) and `createManyFromEvaluateResult`. Three near-identical copies of the redaction contract now have to be kept in sync by hand — security-relevant if they drift. Consider extracting a shared private helper that all three spread.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ianw-oai (approved):
Looks good with the paired JSONL sanitizer/finalization stack. Programmatic JSONL output is preserved and recovery behavior is covered.

ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/5c3f200b-0e04-4ccd-826c-e2c335e85fe1?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on b563424399648985ba7a7236fce209a4071f6277. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 6dc645f62aafaaff4add9f27214c54f60e17b59a | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/688205f7-d926-4964-95a4-8a8b4cb480e2?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 1:41AM |
| e838fd7bb15d32343e647638ec317aaa65892ea7 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/f18b011b-af33-45df-bc3d-481d25299e07?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 2:06AM |
| 3982d887e02577ff243cae93e14c2e8ac424c450 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/9d41d192-05e2-411a-9d2c-8379e481220e?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 2:13AM |
| 4eb338945a644c474c10e9f8c5f52602dc52b94f | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/250c5d0b-7006-4702-8e18-989670b0815f?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 2:34AM |
| b563424399648985ba7a7236fce209a4071f6277 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/5c3f200b-0e04-4ccd-826c-e2c335e85fe1?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 3:39AM |

</p>
</details>

PR #9546, opened 2026-05-30
fix(output): redact api-key and legacy transport headers in JSONL/DB

mldangelo-oai:
## Summary

Builds on the merged programmatic-JSONL fix (#9538) to close two credential-redaction
gaps at the persistence / JSONL-artifact boundary. This branch was rebuilt on current
`main` (which already has `sanitizeResultForJsonlArtifact`, the `metadata.http` header
redaction, recovery, and atomic finalization), so the diff is now just the two redaction
fixes below.

### 1. API-key request headers
`metadata.http.requestHeaders` was redacted only against the response-header name list,
which omits `api-key` / `x-api-key` / `x-auth-token` / `bearer`. A new
`isSensitiveRequestHeader` (= `isSensitiveResponseHeader || isSecretField`) is applied to
the request-header slot so those credential headers are redacted in both the DB row and the
JSONL artifact.

### 2. Legacy top-level `metadata.headers`
Some legacy integrations record transport headers at a top-level `metadata.headers` slot
(outside `metadata.http`). That slot is now redacted, but **gated by a deep-equal provenance
check** against the response's transport headers, so a header is redacted only when it
actually echoes the transport — user-authored test/grading `metadata.headers` are preserved.
The legacy slot uses the same request-header matcher (a strict superset), so api-key-style
names are redacted there too.

When an `afterEach` extension hook replaces `response.metadata`, `synchronizeLegacyTransportHeaders`
drops a now-stale legacy `metadata.headers` (copied from the old transport) instead of
persisting stale provider credentials.

## Tests

- `test/evaluate.jsonl.test.ts`: non-persisted + persisted JSONL — api-key/x-api-key request
  headers, legacy `metadata.headers` redaction (incl. api-key), user-authored test metadata
  headers preserved, model-output `http` keys untouched.
- `test/models/evalResult.test.ts`: DB round-trip redaction of api-key request headers;
  arbitrary legacy headers in grading metadata preserved.
- `test/evaluator/metadata.test.ts`: afterEach hook replacing `response.metadata` drops the
  stale legacy header.
- `test/evaluator.sigint.test.ts`: streamed JSONL rows are redacted when the user aborts
  before final export.

## Validation

- `npm run tsc` clean · touched suites green · lint/format clean

PR #9547, opened 2026-05-30
fix(eval): canonicalize retry JSONL output with atomic rewrites

mldangelo-oai:
## Summary

Standalone `promptfoo retry <evalId>` runs the evaluator in resume mode, so it **appends** replacement rows to the configured JSONL artifact while it works. After a successful retry the command deletes the obsolete database `ERROR` rows and recalculates prompt metrics — but without a final rewrite the JSONL file still contains the stale `ERROR` rows alongside the appended replacements. This PR rewrites the JSONL artifact from the canonical post-cleanup database state, and hardens the shared atomic JSONL writer it relies on.

### `src/commands/retry.ts`

- After a successful retry, rewrite the configured **JSONL** outputs from the database once the stale `ERROR` rows have been deleted (gated on the deletion succeeding, independent of metrics recalculation).
- Limit the rewrite to JSONL paths taken from the **original eval's** `outputPath` — the artifact the resume actually appended to — never the override `--config`, and never non-JSONL targets such as Google Sheets / JSON / YAML / HTML (full-overwrite formats that streaming never dirties, and which can carry share URLs or external side effects).
- If retry results fail to persist, restore the JSONL from the pre-retry database view and keep the existing `ERROR` rows so a re-run can recover. Post-retry cleanup/rewrite failures are treated as non-fatal warnings rather than failing a retry whose results are already saved.

### `src/util/output.ts`

- Make the JSONL rewrite **atomic**: build the full file in a sibling temp file and `rename()` it over the destination, so a crash mid-write leaves the original intact rather than a truncated artifact.
- **Preserve symlinks** (resolve and write through to the link target, creating the target's directory when needed) and **artifact permission bits** (re-`chmod` the temp file to the destination's prior mode before the rename, bypassing the umask).
- Fall back to an **external backup + in-place copy** when the output directory is not writable or the final `rename` is denied (`EACCES`/`EPERM`), restoring the original on failure and surfacing the real permission error rather than a secondary `ENOENT`.
- `writeMultipleOutputs` now uses `Promise.allSettled` and aggregates errors, so one failing artifact no longer aborts the others mid-flight.

## Verification

- **Unit:** `npx vitest run test/commands/retry.test.ts test/util/output.test.ts test/util/jsonlOutput.test.ts test/evaluate.jsonl.test.ts test/evaluator.sigint.test.ts test/models/evalResult.test.ts test/evaluator/metadata.test.ts` — all green; `npm run tsc` clean; CI Style Check (Biome) green.
- **End-to-end:** a real `promptfoo eval` that produced an `ERROR` row, followed by `promptfoo retry`, confirmed the JSONL was canonicalized — the stale `ERROR` row was dropped, the replacement was present with no duplicates, the `0600` file mode was preserved, the inode was swapped (atomic rename), and a sibling `results.json` was correctly left untouched (JSONL-only scoping).

## Notes

- Synced with latest `main`. The legacy transport-header redaction that previously rode along on this branch already landed via #9546, so this PR is now cleanly scoped to retry JSONL canonicalization plus the atomic JSONL writer it depends on (net diff: `src/commands/retry.ts`, `src/util/output.ts`, and their tests).


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
## Review: `fix(eval): canonicalize retry JSONL output`

This PR makes standalone `promptfoo retry` rewrite the configured JSONL artifact after a successful retry + DB cleanup, so stale ERROR rows are removed and replacements appear exactly once. The JSONL writer in `writeOutput` is reworked to write a uniquely-named temp file and atomically `rename` it into place (resolving symlinks, preserving the existing file mode). The behavior change is sound and well covered by new DB-backed and file-content tests.

Note for the orchestrator: several of the candidate findings were written against an abandoned revision of `src/util/output.ts` that contained an `isPermissionDeniedError` in-place fallback (the `previousOutput = readFile(...)` / truncate / restore block). That block is **not present at the current branch tip** (`72a7f4c705`). The current head simply lets a temp-file `EACCES`/`EPERM` propagate (caught by the outer `catch`, which `rm`s the temp file and rethrows), which matches the new `preserves the existing JSONL artifact when the output directory is not writable` test. All findings about ENOENT-masking, restore-write failure, untested EPERM fallback, and a redundant `rm` in the fallback branch are therefore moot and rejected.

Nothing blocking. A couple of low-severity items worth a look:

**Low**

- `src/commands/retry.ts:357-371` — After `errorRowsDeleted === true` (DB cleanup already persisted), `writeMultipleOutputs(...)` is awaited without a local try/catch, unlike the cloud-share block just below it. If the JSONL rewrite throws (disk full, dir made unwritable mid-run, etc.), the error propagates to `setupRetryCommand`'s action handler, which logs `Failed to retry evaluation` and advises "Run the same retry command again to continue." But the retry and DB cleanup already succeeded and were persisted — re-running finds no ERROR rows and only re-attempts the artifact rewrite, so the recovery guidance doesn't match the actual state. Impact is bounded (failure is loud, not silent; DB is correct). Fix: wrap this call in a try/catch that emits an artifact-specific warning like "Retry succeeded and the database is up to date, but rewriting the JSONL artifact at `<path>` failed: `<err>`; re-run `promptfoo retry <id>` to regenerate it," and decide intentionally whether to set a non-zero exit code, rather than letting it surface as a generic eval failure.

- `src/util/output.ts:516-521` — The temp file is created with a per-mode branch (`writeFile(tempOutputPath, '')` vs `writeFile(tempOutputPath, '', { mode: outputMode })`), but the `{ mode }` argument has no observable effect on the final artifact: `writeFile`'s `mode` is masked by umask, while the subsequent `chmod(tempOutputPath, outputMode)` sets the exact mode umask-independently and always overrides it. The new permission test confirms this — under `umask(0o077)`, `writeFile('', { mode: 0o644 })` would yield `0o600`, yet the test expects `0o644`, so only the `chmod` produces the asserted result. The temp file already has an unguessable random name in the same directory, so the "narrow the wrong-permission window" rationale is negligible. Collapse to a single unconditional `await fsPromises.writeFile(tempOutputPath, '');` and keep the existing `if (outputMode !== undefined) await fsPromises.chmod(...)` as the sole permission step.

**Nits**

- `test/util/jsonlOutput.test.ts:203-220` — The only symlink test (`should preserve an existing artifact symlink during a rewrite`) points at an existing target, so `realpath` succeeds and `resolveJsonlOutputPath`'s dangling-symlink fallback (`realpath` ENOENT → `path.resolve(dirname, readlink(...))`, output.ts:73-78) is never exercised. That branch is a deliberate design choice in code this PR adds; a regression resolving against the wrong directory wouldn't be caught. Consider adding a `process.platform !== 'win32'` test that symlinks to a non-existent target in the same dir, calls `writeOutput`, then asserts the target was created with the replacement content and the symlink is preserved.

- `test/commands/retry.test.ts:191-222` — The "partial cleanup" regression triggers the simulated failure by throwing on the 4th `Eval.prototype.addPrompts` call and asserts `expect(addPromptsCalls).toBe(4)`. The `4` is coupled to how many times `addPrompts` happens to be invoked across `evaluate()` + `recalculatePromptMetrics()`; a benign refactor changing that count would silently stop triggering the intended failure or break the assertion even though the behavior under test is unchanged. Prefer spying on `recalculatePromptMetrics` (or the specific call inside it) and asserting on observable state (the "Post-retry cleanup had issues" warn + rewritten file) rather than a positional call count.

- `src/commands/retry.ts:357-366` — Minor readability: the array-normalize ternary + type-guard `.filter` is inlined directly as the `filterOutputPathsAfterStreaming(...)` argument. `src/commands/eval.ts:993` uses a named local (`paths`) for the same shape. Optional: extract a `jsonlPaths` local before the call to mirror it.

**Tests**: New coverage is solid (DB-backed retry rewrite, persistence-failure preservation, symlink/permission/long-name rewrite, temp-file atomic rename, and EACCES propagation). The two test gaps above (dangling-symlink fallback, brittle call-count) are the only notable misses.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ianw-oai (approved):
Looks good. Retry JSONL output is canonicalized through safer rewrites, persistence-failure recovery, and aggregate output error handling.

ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/ca4fdab5-f548-4d78-be63-2f0782a26fef?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on 2d31f64c8f606d2f79e04a1a4c81787231bead63. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 2d31f64c8f606d2f79e04a1a4c81787231bead63 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/ca4fdab5-f548-4d78-be63-2f0782a26fef?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 2, 2026 5:19AM |

</p>
</details>

PR #9554, opened 2026-05-30
fix(providers): suppress optional GCP metadata warning

mldangelo-oai:
## Summary

- suppress the exact expected `MetadataLookupWarning` emitted when the optional Google ADC probe cannot reach metadata hosts
- preserve explicit Vertex authentication and forward unrelated process warnings
- add regression coverage for both suppression and forwarding behavior

## Testing

- `./node_modules/.bin/vitest run test/providers/defaults.test.ts test/providers/google/auth.test.ts --sequence.shuffle=false`
- `./node_modules/.bin/biome check src/providers/google/auth.ts test/providers/google/auth.test.ts`
- real `GoogleAuthManager.hasDefaultCredentials()` probe outside GCP
- real `getDefaultProviders()` fallback selection outside GCP

## Notes

- `tsc --noEmit` still reports pre-existing Anthropic usage-shape errors in unrelated test files.

### DIFF
--- a/src/providers/google/auth.ts
+++ b/src/providers/google/auth.ts
@@ -22,26 +22,42 @@ import type { CompletionOptions } from './types';
 const EXPECTED_GCP_METADATA_LOOKUP_WARNING =
   'received unexpected error = All promises were rejected code = UNKNOWN';
 
+let activeGcpMetadataLookupWarningSuppressions = 0;
+let restoreEmitWarning: (() => void) | undefined;
+
 // The optional ADC probe should stay quiet when gcp-metadata cannot reach its hosts,
 // while explicit Vertex authentication and unrelated process warnings remain visible.
 async function suppressExpectedGcpMetadataLookupWarning<T>(action: () => Promise<T>): Promise<T> {
-  const originalEmitWarning = process.emitWarning;
+  if (activeGcpMetadataLookupWarningSuppressions === 0) {
+    const originalEmitWarning = process.emitWarning;
+    const wrappedEmitWarning = ((warning: string | Error, ...args: unknown[]) => {
+      const message = warning instanceof Error ? warning.message : warning;
+      const type = warning instanceof Error ? warning.name : args[0];
 
-  process.emitWarning = ((warning: string | Error, ...args: unknown[]) => {
-    const message = warning instanceof Error ? warning.message : warning;
-    const type = warning instanceof Error ? warning.name : args[0];
+      if (type === 'MetadataLookupWarning' && message === EXPECTED_GCP_METADATA_LOOKUP_WARNING) {
+        return;
+      }
 
-    if (type === 'MetadataLookupWarning' && message === EXPECTED_GCP_METADATA_LOOKUP_WARNING) {
-      return;
-    }
+      Reflect.apply(originalEmitWarning, process, [warning, ...args]);
+    }) as typeof process.emitWarning;
 
-    Reflect.apply(originalEmitWarning, process, [warning, ...args]);
-  }) as typeof process.emitWarning;
+    process.emitWarning = wrappedEmitWarning;
+    restoreEmitWarning = () => {
+      if (process.emitWarning === wrappedEmitWarning) {
+        process.emitWarning = originalEmitWarning;
+      }
+    };
+  }
 
+  activeGcpMetadataLookupWarningSuppressions += 1;
   try {
     return await action();
   } finally {
-    process.emitWarning = originalEmitWarning;
+    activeGcpMetadataLookupWarningSuppressions -= 1;
+    if (activeGcpMetadataLookupWarningSuppressions === 0) {
+      restoreEmitWarning?.();
+      restoreEmitWarning = undefined;
+    }
   }
 }
 
--- a/src/providers/google/auth.ts
+++ b/src/providers/google/auth.ts
@@ -19,6 +19,8 @@ import type { GoogleAuthOptions } from 'google-auth-library';
 import type { EnvOverrides } from '../../types/env';
 import type { CompletionOptions } from './types';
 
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
@codex review

mldangelo:
## Review: fix(providers): suppress optional GCP metadata warning

This PR wraps the optional ADC probe (`hasDefaultCredentials` → `getOAuthClient`) in `suppressExpectedGcpMetadataLookupWarning`, which temporarily swaps the global `process.emitWarning` and drops exactly one warning: a `MetadataLookupWarning` whose message equals `'received unexpected error = All promises were rejected code = UNKNOWN'`. I verified the core safety property: the wrapper forwards every other shape (different type, different message, Error instances, options-object overload), so genuine auth/credential errors and unrelated process warnings stay visible. I also confirmed the hardcoded literal is the *real* canonical off-GCP string — gcp-metadata@8.1.2's `isAvailable()` uses `fastFail=true` → `Promise.any([...])`, which on failure throws an `AggregateError` with `message="All promises were rejected"` and no `code`, so gcp-metadata emits precisely that string. The fix is correct and the security concern is satisfied. A couple of robustness/coverage gaps are worth addressing.

**Low**

- `src/providers/google/auth.ts:40-43` — The `finally` unconditionally restores `process.emitWarning = originalEmitWarning`, which is only safe under strict LIFO nesting. The probe is normally deduped by `pendingHasDefaultCredentials`, but `GoogleAuthManager.clearCache()` (publicly re-exported as `clearCachedAuth` via `util.ts`) resets `pendingHasDefaultCredentials = undefined`. If `clearCache()` runs while probe A is in-flight and a second `hasDefaultCredentials()` then starts probe B, B captures `originalEmitWarning = <A's wrapper>`; when A settles first and B settles second, the global is left permanently pointing at A's leaked wrapper (extra `Reflect.apply` indirection + permanent silent suppression of that exact warning). I reproduced this against the real `GoogleAuthManager` code. Impact is bounded and reachability is low (`clearCachedAuth` has no production caller in `src/` — only tests — and the global mutation is new in this PR, not pre-existing), hence low rather than higher. Fix: guard the restore so it only reverts the wrapper this call installed, e.g. `if (process.emitWarning === wrapped) process.emitWarning = originalEmitWarning;` (close the wrapper over a named `wrapped` reference). This same guard also hardens against an unrelated module concurrently swapping `process.emitWarning` during the await window.
- `src/providers/google/auth.ts:20-21` — Suppression is exact-string-coupled to gcp-metadata's internal `AggregateError` message template plus its `received unexpected error = ${err.message} code = ${code}` format. This errs on the *safe* side (it can never over-suppress a real error), but a future gcp-metadata bump that reworks the message or switches to the `emitWarning(msg, { type })` options overload would silently stop suppressing with no test catching the drift. Consider matching on a stable prefix (`type === 'MetadataLookupWarning' && message.startsWith('received unexpected error =')`) or at minimum add a comment pinning the gcp-metadata version this string was copied from so dependency bumps re-verify it.

**Tests**

The two new tests pass (52/52 in the PR's `auth.test.ts`), but they under-pin the contract:
- The forward test only varies the *message* while keeping `type: 'MetadataLookupWarning'`, so it doesn't prove the `type ===` guard actually gates suppression — a regression dropping the type check wouldn't fail it. Add a forward case with a different type (e.g. `process.emitWarning('GOOGLE_APPLICATION_CREDENTIALS not found', 'DeprecationWarning')`).
- The `warning instanceof Error` branch (deriving `type`/`message` from `name`/`message`) has zero coverage. Add an Error-instance forward case.
- Neither test asserts `process.emitWarning` is restored after the (always-throwing) probe — capture the original reference before the probe (without globally replacing it) and assert identity afterward, which would also lock in the leak fix above.

Note: I explicitly checked and rejected the claim that a realistic off-GCP probe emits `connect ETIMEDOUT ... code = ETIMEDOUT` and is therefore not suppressed — that variant only arises with a `GCE_METADATA_IP`/`GCE_METADATA_HOST` override (`fastFail=false`); the default off-GCP path produces exactly the literal the PR hardcodes.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

ianw-oai (approved):
Looks good.

ianw-oai (approved):
Looks good. The warning suppression is narrowly scoped to the optional ADC probe and keeps unrelated warnings visible.

ianw-oai (approved):

ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/edbb7dc6-522c-4580-ab8c-40000e792914?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>

> [!TIP]


<sub>Last run on 44832581494ba0585f75786d9c64123da7973480. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 44832581494ba0585f75786d9c64123da7973480 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/edbb7dc6-522c-4580-ab8c-40000e792914?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 1, 2026 3:33AM |

</p>
</details>

PR #9556, opened 2026-05-30
fix(code-scan): reject fractional inline comment lines

serhiizghama:
## Summary

- Merge current `main`, which already routes file-only and non-positive-line findings to general PR comments.
- Harden `prepareComments()` so a finding is treated as an inline review comment only when its line is a positive integer.
- Route fractional-line findings through the general-comment fallback path instead of sending GitHub an invalid inline review payload.
- Add parameterized regression coverage for `line: null`, `0`, `-1`, and `1.5`.

## Why this remains useful

`CommentSchema` accepts `z.number().nullable()`, so a malformed or unexpected server response can contain a positive fractional line. Before this change, `line > 0` classified that finding as inline even though GitHub inline review locations require integer line numbers. This PR closes that remaining edge case without duplicating #9423.

This PR complements #9423: #9423 protects the action-level ordinary `line: null` fallback flow, while this PR protects the lower-level location classifier.

## Validation

- `npx vitest run test/codeScans/util/github.test.ts test/code-scan-action/main.test.ts --sequence.seed=9556` passed (`45` tests).
- `npx vitest run test/codeScans test/code-scan-action --sequence.seed=9556` passed (`195` tests).
- `npm run tsc` passed.
- `npm --prefix code-scan-action run tsc` passed.
- `npm --prefix code-scan-action run build` passed.
- `npm run l` and `git diff --check` passed.
- `npm run f` applied Biome formatting; its empty Prettier stage returned the repository expected exit `123` for this TypeScript-only diff.


mldangelo-oai:
Thank you!

PR #9559, opened 2026-05-30
chore(deps): downgrade anthropic sdk pins

mldangelo-oai:
## Summary

- downgrade `@anthropic-ai/claude-agent-sdk` to `0.3.146`
- downgrade `@anthropic-ai/sdk` to `0.97.1`
- remove newer-only Anthropic usage mock fields so tests typecheck against the older SDK types

## Validation

- `npm install`
- `npm ls @anthropic-ai/claude-agent-sdk @anthropic-ai/sdk`
- `npx vitest run test/providers/claude-agent-sdk.test.ts test/providers/anthropic/messages.test.ts test/providers/anthropic/util.test.ts`
- `npm run tsc`
- `npx vitest run test/providers/azure/chat.test.ts test/providers/bedrock/converse.test.ts test/providers/bedrock/index.test.ts test/providers/google/vertex.test.ts test/providers/webSearchUtils.test.ts`
- `npm run build`
- `npx check-dependency-version-consistency`
- `npm run l`
- `npx @biomejs/biome check --write package.json package-lock.json test/providers/anthropic/messages.test.ts test/providers/anthropic/util.test.ts test/providers/claude-agent-sdk.test.ts`
- `git diff --check`

`npm run f` ran Biome successfully, then exited on the repo script's empty Prettier file-list case because this branch does not touch any CSS/Markdown/YAML files.

PR #9560, opened 2026-05-30
fix(providers): preserve streamed Anthropic refusal guardrails

mldangelo-oai:
## Summary

Pins the Anthropic SDKs to releases that clear the internal Socket minimum-release-age install gate, and prevents **streamed** refusals from silently losing their guardrail flag on those releases.

- **Pin SDKs to Socket-installable versions**
  - `@anthropic-ai/sdk` `0.100.0` → `0.98.0`
  - `@anthropic-ai/claude-agent-sdk` `0.3.153` → `0.3.150`

  The freshly-published `0.100.0` / `0.3.153` tarballs are blocked by Socket's minimum-release-age check during internal installs; the pinned versions are old enough to pass.

- **Preserve streamed refusal guardrails (the behavior fix)**

  `@anthropic-ai/sdk` < 0.99.0 does not copy `stop_details` from the `message_delta` event into the accumulated `stream.finalMessage()`. With streaming enabled (`config.stream: true`), an Anthropic response that ends in `stop_reason: 'refusal'` arrived without `stop_details`, so `getRefusalDetails()` returned `undefined` and `guardrails.flagged` was never set — a streamed safety block was reported as an ordinary output (a false negative for redteam/eval runs). The non-streaming path was unaffected because the server returns `stop_details` directly.

  `finalMessageWithStreamedStopDetails()` now captures `stop_details` from the raw `message_delta` stream event and restores it onto the final message. The backfill only runs when `finalMessage()` omits it, so it is a no-op on SDKs that already accumulate it (≥ 0.99.0).

- **Raise Renovate `minimumReleaseAge` 7 → 10 days** for the npm rules, widening the stabilization window before Renovate proposes dependency updates.

- **Keep test fixtures compatible with the older SDK typings.** `output_tokens_details` was added to the SDK `Usage` type in `0.100.0`; the affected fixtures are widened with small local helper types so they still type-check against `0.98.0` while preserving the production token shape.

## Why these specific versions

`stop_details` stream accumulation was fixed in `@anthropic-ai/sdk` 0.99.0, but 0.99.0/0.100.0 are too recently published to clear the Socket install gate. `0.98.0` is the newest release that clears the gate, so we pin there and add the local workaround to cover the missing accumulation.

## Validation

- `npm ci` · `npm run tsc` · `npm run l`
- `npx vitest run test/providers/anthropic/messages.test.ts test/providers/anthropic/util.test.ts test/providers/claude-agent-sdk.test.ts` — **379 passed**
- Regression test `should include guardrails in streaming response when stop_reason is refusal` emits `stop_details` only via the raw stream event (and `null` on `finalMessage()`); verified it **fails without** the workaround and passes with it.
- Live `eval` against the Anthropic provider on the pinned `0.98.0`, both non-streaming and `stream: true` — **4/4 passed**.

Supersedes and consolidates #9559 (closed as a duplicate).


mldangelo-oai:
@codex review

PR #9567, opened 2026-05-31
fix(db): avoid SQLITE_LOCKED flakiness in shared-cache test database

mldangelo:
## Summary

Fixes a flaky test surfaced in CI (`test/commands/import.test.ts > importCommand > collision handling > should reject import when eval already exists`, seen on Node 26.x / Windows). The second import's collision check intermittently failed with:

```
Failed to import eval: Failed query: select "id", "created_at", ..., "is_redteam" from "evals" where "evals"."id" = ?
params: eval-WJi-2025-10-01T20:08:11
```

instead of the expected `Eval <id> already exists` message.

## Root cause

The test database is a process-wide shared-cache in-memory libsql DB (`file::memory:?cache=shared`), introduced when we replaced better-sqlite3 with libsql (#9486).

- **Shared-cache mode uses table-level locks.** A read that overlaps another connection's open write transaction fails *immediately* with `SQLITE_LOCKED_SHAREDCACHE`.
- **`busy_timeout` does not cover this** — it only retries `SQLITE_BUSY`, not shared-cache `SQLITE_LOCKED`.
- **libsql runs interactive transactions on separate physical connections**, so the JS-level operation serialization in `serializeTopLevelOperations` cannot guarantee a prior writer's table lock is released the instant its promise settles. On a loaded runner that brief window is enough for `Eval.findById`'s `SELECT ... FROM evals` to hit the lingering lock.

I reproduced the exact error deterministically through the real `getDb()` layer (a top-level read on the main connection while a separate connection holds a write lock on `evals`) — same query text and `SQLITE_LOCKED_SHAREDCACHE` extended code as CI.

## Fix

Retry the transient lock at the serialization layer (`serializeTopLevelOperations`). A statement that fails to acquire its lock never executed, so retrying is safe (no double-applied writes), and it rides through the brief window where a prior writer's transaction connection still holds the table lock after its JS promise settled.

- **Isolation preserved:** unlike disabling read locks (`read_uncommitted`), retrying keeps the test DB production-like — reads still observe only committed rows.
- **Scoped narrowly:** retry wraps only the atomic statement methods (`execute`/`batch`/`migrate`/`executeMultiple`) on the serialized path. The in-transaction bypass is left direct (retrying there would deadlock against the enclosing transaction), and transaction callbacks are not retried (to avoid re-running side effects).
- **Bounded:** up to 10 attempts with capped exponential backoff; only triggers on `SQLITE_BUSY`/`SQLITE_LOCKED`, which is rare in production (WAL file DB) and covered by `busy_timeout` for the writer-vs-writer case.

> Replaces an earlier `PRAGMA read_uncommitted` approach after review feedback that it weakened test isolation by exposing uncommitted rows.

## Regression test

Added a test in `test/database/isolation.test.ts`: an independent connection holds a write lock, a top-level read through the main connection contends with it, the lock is released mid-read, and the read is asserted to succeed seeing **only the committed row**. It fails without the fix (the read rejects immediately with `SQLITE_LOCKED`) and confirms no dirty read with it.

## Verification

- `test/database/isolation.test.ts` — passes with fix; new test fails (`SQLITE_LOCKED`) without it.
- `test/commands/import.test.ts` — passes.
- `test/database/` + `test/models/` + `test/util/database*` — 353 pass (single fork, shuffled).
- `test/commands/` + `test/server/` + `test/util/` — 3253 pass (single fork, shuffled).
- `npm run l`, `npm run f`, and `npx tsc --noEmit` all clean.

### DIFF
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -6,6 +6,7 @@ import { DefaultLogger, type LogWriter } from 'drizzle-orm/logger';
 import { getEnvBool } from '../envars';
 import logger from '../logger';
 import { getConfigDirectoryPath } from '../util/config/manage';
+import { sleep } from '../util/time';
 import {
   closeTestDatabaseClient,
   registerTestDatabaseClient,
@@ -50,21 +51,6 @@ async function configureDatabase(client: Client, skipWalMode: boolean): Promise<
   // connection for the lock waits instead of erroring with SQLITE_BUSY.
   await client.execute('PRAGMA busy_timeout = 5000');
 
-  // The test database is a process-wide shared-cache in-memory DB
-  // (`file::memory:?cache=shared`). Shared-cache uses table-level locks, so a
-  // read that overlaps another connection's open write transaction fails
-  // immediately with SQLITE_LOCKED_SHAREDCACHE — and busy_timeout does NOT cover
-  // it (that only applies to SQLITE_BUSY). libsql also opens separate physical
-  // connections for interactive transactions, so the JS-level operation
-  // serialization cannot guarantee the prior writer's table lock is released the
-  // instant its promise settles. read_uncommitted lets readers proceed without
-  // taking a conflicting table lock, which removes the flaky lock errors. It is
-  // shared-cache-only (a no-op for the file-backed production DB) and we never
-  // reach this branch outside tests, so production isolation is unchanged.
-  if (skipWalMode) {
-    await client.execute('PRAGMA read_uncommitted = 1');
-  }
-
   // Configure WAL mode unless explicitly disabled or using in-memory database
   if (!skipWalMode && !getEnvBool('PROMPTFOO_DISABLE_WAL_MODE', false)) {
     try {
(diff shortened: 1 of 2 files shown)


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/b285f9eb-e565-4d59-80d3-609dfab1f81e?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on e9e97d2c957b99e0ed7c63a8f965e8756e4f4df3. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 914df37cdbb1dcb6a0f6cb4491d541caaecd7996 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/6f0e4a64-5b37-45d4-9fde-e9b8dbd8a6f4?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 31, 2026 9:43PM |
| e9e97d2c957b99e0ed7c63a8f965e8756e4f4df3 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/b285f9eb-e565-4d59-80d3-609dfab1f81e?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | May 31, 2026 11:42PM |

</p>
</details>

PR #9571, opened 2026-06-01
fix(cloud): resolve on-prem API host correctly in checkEmailStatus, guardrails, and http-provider-generator

KenichiSuda:
## Summary
  
  When a Promptfoo on-premises instance is configured via `promptfoo auth login`,
  the API host is saved to the local config file through `CloudConfig.setApiHost()`.
  However, three call-sites ignored the saved config and fell back to the public
  cloud API (`https://api.promptfoo.app`) because they checked only the
  `PROMPTFOO_CLOUD_API_URL` environment variable (or a module-level constant
  evaluated at import time):
  
  - `checkEmailStatus` in `src/globalConfig/accounts.ts` – called on every
    `promptfoo eval` and `promptfoo redteam generate`, so every run produced an
    unexpected request to the public cloud even for on-prem users
  - `src/guardrails.ts` – the base URL was a module-level constant resolved once
    at import time, so `cloudConfig` state was never consulted
  - `/providers/http-generator` route in `src/server/routes/providers.ts` – used
    the env-var-only path, missing the saved config-file host
    
  ## Changes

  | File | Change |
  |---|---|
  | `src/globalConfig/accounts.ts` | `checkEmailStatus` now prefers `cloudConfig.getApiHost()` when `cloudConfig.isEnabled()` |
  | `src/guardrails.ts` | Module-level `API_BASE_URL` constant replaced with `getApiBaseUrl()` function that consults `cloudConfig` at call time |
  | `src/server/routes/providers.ts` | `/api/v1/http-provider-generator` route prefers `cloudConfig.getApiHost()` when cloud is enabled |

  All three sites fall back to the existing `PROMPTFOO_CLOUD_API_URL` env-var /
  `https://api.promptfoo.app` default for non-cloud setups, so there is no
  behaviour change for standard (non-on-prem) users.
  
  ## Test plan
  
  - [x] Existing unit tests pass: `accounts.test.ts` (60 tests), `guardrails.test.ts` (14 tests)
  - [x] On-prem: after `promptfoo auth login --api-host https://your-onprem-host`,
    running `promptfoo eval` no longer sends requests to `api.promptfoo.app` for
    email status, guardrail assertions, or the HTTP provider generator
  - [x] Non-cloud: behaviour unchanged — requests still go to `PROMPTFOO_CLOUD_API_URL`
    env var or `https://api.promptfoo.app` when `cloudConfig` is not enabled

### DIFF
--- a/src/globalConfig/accounts.ts
+++ b/src/globalConfig/accounts.ts
@@ -14,6 +14,7 @@ import {
   UserEmailStatus,
 } from '../types/email';
 import { fetchWithTimeout } from '../util/fetch/index';
+import { cloudConfig } from './cloud';
 import { readGlobalConfig, writeGlobalConfig, writeGlobalConfigPartial } from './globalConfig';
 
 import type { GlobalConfig } from '../configTypes';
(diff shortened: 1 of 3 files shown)


KenichiSuda:
I will create another pull request containing only the changes to `accounts.ts`. Therefore, I will close this pull request.

PR #9576, opened 2026-06-01
fix(cloud): use on-prem API host in checkEmailStatus

KenichiSuda:
## Summary
  
  `checkEmailStatus` resolved the API host by checking only the
  `PROMPTFOO_CLOUD_API_URL` environment variable, falling back to the
  public `https://api.promptfoo.app`. On-prem users who authenticated
  via `promptfoo auth login` had their host saved to the config file via
  `CloudConfig.setApiHost()`, but `checkEmailStatus` never consulted 
  `cloudConfig`, so every `promptfoo eval` and `promptfoo redteam generate`
  run sent an unexpected request to the public cloud API. 
  
  ## Root cause
  
  // Before — env var only; ignores the host saved in the config file
  const host = getEnvString('PROMPTFOO_CLOUD_API_URL', 'https://api.promptfoo.app');

  cloudConfig.resolveApiHost() checks config.apiHost (the config-file
  value) first, then the env var, then the default. checkEmailStatus skipped
  the first step entirely.

   Fix

  // After — prefer the config-file host when cloud/on-prem is enabled
  const host = cloudConfig.isEnabled()
    ? cloudConfig.getApiHost()
    : getEnvString('PROMPTFOO_CLOUD_API_URL', 'https://api.promptfoo.app');

  Non-cloud users are unaffected; the fallback path is unchanged.

  ## Test plan

  - [ ] npx vitest run test/globalConfig/accounts.test.ts — 60 tests pass
  - [ ] On-prem: after promptfoo auth login --api-host https://your-onprem-host,
  promptfoo eval no longer sends requests to api.promptfoo.app for
  email status checks
  - [ ] Non-cloud: requests continue to use PROMPTFOO_CLOUD_API_URL or
  https://api.promptfoo.app as before


mldangelo (approved):
Thanks @KenichiSuda — nice catch on the on-prem host bug! This correctly routes `checkEmailStatus` to the configured cloud host when cloud is enabled, falling back to the existing `PROMPTFOO_CLOUD_API_URL` / default otherwise.

I verified this exact change end-to-end against a mock on-prem server: the request hits `<your-host>/api/users/status` rather than `api.promptfoo.app`, and it passes across the full test matrix. The single failing check is a flaky Windows vitest worker crash unrelated to this change — I've re-triggered it.

As a follow-up I've opened #9580, which adds regression tests for this branch and applies the same fix to guardrails and the HTTP provider generator. I'll rebase it to build on top of this once it lands, so your change is the one that introduces the `accounts.ts` fix. 🙏

KenichiSuda:
@mldangelo thank you for your comments. And thank you for creating #9580.
---
Initially, I created #9571, but I'm not clear whether I need to change `guardrails.ts` after I submitted it.
Then, I closed #9571 and I created #9576 for minimal requests based from #9571.
Now, I understood that you have already created #9580 which covered 3 changed points in #9571.
So, after you merge #9580, we don't need to take care this #9576, because fixes in #9580 cover this #9576.

PR #9593, opened 2026-06-03
fix(db): wait for Windows file-backed close release

mldangelo-oai:
## Summary
- wait for Windows file-backed database handles to release after `closeDb()` before reporting the connection closed
- cover the close-release retry in `test/database/index.test.ts`
- cover deleting and recreating a file-backed database in `test/database.test.ts`

## Validation
- `npm_config_cache=/private/tmp/promptfoo-npm-cache npx vitest run test/database/index.test.ts test/database.test.ts --sequence.shuffle=false`
- `npm_config_cache=/private/tmp/promptfoo-npm-cache npx vitest run test/database --sequence.shuffle=false`

### DIFF
--- a/src/database/index.ts
+++ b/src/database/index.ts
@@ -1,4 +1,5 @@
 import { AsyncLocalStorage } from 'node:async_hooks';
+import * as fs from 'node:fs/promises';
 import { pathToFileURL } from 'node:url';
 import * as path from 'path';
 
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
Closing after Windows CI validation. The added wait does not make the file deletable on Windows because opening with r+ is not a valid delete-sharing probe, and this delete/recreate scenario only comes from the unmerged safety test in #9590 rather than a verified runtime path. Shipping this would add runtime delay and risk without fixing a demonstrated product bug.

PR #9594, opened 2026-06-03
fix(vars): resolve file:// references nested inside object and array vars

### LINKED ISSUE #1613: File content not loaded when used in vars dict

serhiizghama:
## Problem

When a test var is a plain object or array containing `file://` strings at any nesting depth, `renderPrompt` passes the raw URI to the template instead of loading the file content.

**Example config that breaks:**
```yaml
tests:
  - vars:
      reporting_period:
        current:
          period: '2023-12-31'
        previous:
          period: '2024-02-15'
          report: file://data/report.html.txt   # ← injected as-is, not loaded
```

Only top-level string vars were processed; nested structures silently got the raw URI.

Closes #1613

## Solution

Added a private helper `resolveFileRefsInObject` that recursively walks a var value (object or array) and loads any `file://` string it finds, using the same `fs/promises` import already in scope. The new `else if (typeof value === 'object' && value !== null)` branch in `renderPrompt` calls this helper before template rendering.

JS/Python/`package://` paths embedded inside nested objects are **not** handled by this path — they require calling context (`varName`, `prompt`, `provider`) and continue to work only at the top-level var level, which is consistent with prior behaviour.

## Testing

Added one test in `test/evaluatorHelpers.test.ts` that mirrors the exact structure from the issue report. Existing tests unchanged — all 97 pass.

### DIFF
--- a/src/codeScan/util/github.ts
+++ b/src/codeScan/util/github.ts
@@ -65,11 +65,13 @@ export function prepareComments(
     return rankB - rankA;
   });
 
-  // Separate line-specific comments from general PR comments
+  // Separate line-specific comments from general PR comments.
+  // File-only findings (line: null) must go to generalComments — GitHub rejects inline
+  // review comments that have a path but no line number.
   // Filter out severity='none' comments - they shouldn't be posted separately
-  const lineComments = sortedComments.filter((c) => c.file && c.finding);
+  const lineComments = sortedComments.filter((c) => c.file && c.line != null && c.finding);
   const generalComments = sortedComments.filter(
-    (c) => !c.file && c.finding && c.severity !== CodeScanSeverity.NONE,
+    (c) => (!c.file || c.line == null) && c.finding && c.severity !== CodeScanSeverity.NONE,
   );
 
   // Check if we only have "none" severity comments (i.e., no real vulnerabilities)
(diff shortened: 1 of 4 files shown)


mldangelo:
Thanks for the clean, minimal take on #1613, @serhiizghama. This was one of six parallel attempts at the same bug, and we've consolidated onto #9419 as the single vehicle — it extends the same renderPrompt-time approach you took here with YAML support, cycle guards, and the red-team skip-variable protections that the security test suite requires.

One note for splitting future PRs: the `codeScan/prepareComments` fix bundled here looks valuable but is unrelated to #1613 — feel free to re-open that as its own focused PR.

Closing as superseded by #9419. Thank you for the contribution.

PR #9598, opened 2026-06-03
feat(providers): upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade MiniMax model configuration to use M3 as the default model.

## Changes
- Add `MiniMax-M3` to the model selection list and set it as the default
- Keep `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as alternatives
- Remove older models (`MiniMax-M2.5` / `MiniMax-M2.5-highspeed`)
- Update related unit tests, docs, and the `provider-minimax` example

## Why
`MiniMax-M3` is the latest MiniMax flagship model, with a 512K context window, up to 128K output, and image input support. Defaulting to M3 gives users the best out-of-the-box experience while M2.7 routes remain available for compatibility.

## Testing
- `npx vitest run test/providers/minimax.test.ts` — 47 tests passing
- `npx vitest run test/providers/index.test.ts -t "minimax"` — passing


mldangelo (approved):
LGTM ✅

Verified the core facts before approving:
- **MiniMax-M3 is real** (launched 2026-06-01) and the standard pricing in `minimax.ts` is correct: \$0.60/1M input, \$2.40/1M output. `cache_read: 0.12` follows the consistent 20%-of-input ratio (M2.7 = 0.06/0.30).
- Tests are thorough and pass (47 in `minimax.test.ts` + registry test); all CI green.
- No orphaned M2.5 references; docs, example, index table, and tests are all consistent.

Pushed a small doc follow-up commit to correct the M3 context window (up to 1M / 512K guaranteed minimum, not a flat 512K), note text/image/video multimodal input, and flag the ~2x cost increase vs M2.7 so cost-sensitive users can pin M2.7 explicitly.

Note (non-blocking): docs still state `max_completion_tokens` caps at 2048, which reads oddly next to M3's "up to 128K output." Worth confirming the real OpenAI-compatible limit in a follow-up.

mldangelo-oai:
Thank you!

PR #9604, opened 2026-06-03
fix(eval): compose metadata and result filters

mldangelo-oai:
## Summary

- compose metadata filters with failing/error result filters without reintroducing excluded tests
- precompute default-var merged tests and clean up the visible comment/import findings
- clarify the standalone ModelAudit scanner-catalog command and mark custom-scanner exceptions unsuccessful
- simplify ModelAudit schema temporary-file cleanup to remove the empty exception handler

## AI findings

Addresses all 6 visible AI findings across 5 files. The scanner-catalog command remains `modelaudit scan --list-scanners --format json` because it was intentionally fixed on `main`; the documentation now explicitly identifies it as the standalone ModelAudit CLI.

Also addresses the standard CodeQL `Empty except` finding in `scripts/generate_modelaudit_schema.py`.

## Validation

- `npm run f`
- `npm run l` (passes with existing cognitive-complexity warnings)
- `npm run tsc`
- `./node_modules/.bin/vitest run test/util/eval test/assertions/google.test.ts` (195 tests)
- real no-cache CLI eval composing `--filter-metadata type=unit` with `--filter-failing` selected only the expected unit failure
- Ruff 0.15.1 check/format and Python 3.12 atomic-write success/failure cleanup verification
- `npm run build`: core TypeScript and package bundling passed; the unrelated Web UI build was blocked by a missing `diff` package in the shared local dependency install

### DIFF
--- a/scripts/generate_modelaudit_schema.py
+++ b/scripts/generate_modelaudit_schema.py
@@ -119,10 +119,7 @@ def write_text_atomic(path: Path, contents: str) -> None:
             tmp_file.write(contents)
         os.replace(tmp_name, path)
     except Exception:
-        try:
-            os.unlink(tmp_name)
-        except FileNotFoundError:
-            pass
+        Path(tmp_name).unlink(missing_ok=True)
         raise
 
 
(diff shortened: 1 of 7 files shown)


ianw-oai (approved):
Reviewed the composed filter behavior and tests; looks good.

PR #7982, opened 2026-03-04
feat(eval): disable auto-share by default unless paid cloud customer

Ebonsignori:
## Summary

- Disable auto-share by default for all cloud users, only enabling it for paid customers with an active license
- Add `sharing` field to cloud global config, set during login based on `hasActiveLicense` from the `/users/me` endpoint
- Update sharing defaults in both CLI eval and web-UI eval paths

## Cloud PR

Companion PR that adds `hasActiveLicense` to the `/users/me` response: https://github.com/promptfoo/promptfoo-cloud/pull/3723

## Test plan

- [x] Unit tests for `getSharing`/`setSharing` methods (`test/globalConfig/cloud.test.ts`)
- [x] Unit tests for `hasActiveLicense` handling in `validateAndSetApiToken`
- [x] Unit tests for updated sharing defaults (`test/util/sharing.test.ts`)
- [x] Manual: `npm run local -- auth login --api-key <free-user-key>` → sharing is `false` in config
- [x] Manual: `npm run local -- auth login --api-key <paid-user-key>` → sharing is `true` in config
- [x] Manual: Free user eval without `--share` → results NOT auto-shared
- [x] Manual: Paid user eval without `--share` → results ARE auto-shared
- [x] Manual: Free user can still share with explicit `--share` flag

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9612, opened 2026-06-04
fix(auth): enable auto-share for on-prem Report Server after login

KenichiSuda:
## Summary

- `promptfoo auth login` against an on-prem server can store `sharing: false` in the global config when the server returns `hasActiveLicense: false`, silently disabling auto-share to the on-prem Report Server for accounts created after the 2026-03-09 grandfathering cutoff (PR #7982).
- On-prem installations are always enterprise deployments, so the public-cloud `hasActiveLicense` gate does not apply to them.
- Fix: in `saveValidatedApiToken`, skip the license check when the API host is not `api.promptfoo.app` and unconditionally set `sharing: true`.

## Root Cause

PR #7982 added this logic to `saveValidatedApiToken` in `src/globalConfig/cloud.ts`:

```typescript
// Before this fix
if (typeof hasActiveLicense === 'boolean') {
  const createdAt = user?.createdAt ? new Date(user.createdAt) : null;
  const isGrandfathered = createdAt != null && createdAt < SHARING_CUTOFF_DATE;
  this.setSharing(hasActiveLicense || isGrandfathered);
}
```

On-prem servers may return `hasActiveLicense: false` (e.g. because they have no licence-check logic). For users created after the `2026-03-09` cutoff, `isGrandfathered` is also `false`, so `setSharing(false)` is called. `shouldShareResults` then returns `false` and eval results are never uploaded to the on-prem Report Server.

## Fix

```typescript
if (typeof hasActiveLicense === 'boolean') {
  const isOnPrem = !apiHost.replace(/\/+$/, '').includes('api.promptfoo.app');
  if (isOnPrem) {
    this.setSharing(true);
  } else {
    const createdAt = user?.createdAt ? new Date(user.createdAt) : null;
    const isGrandfathered = createdAt != null && createdAt < SHARING_CUTOFF_DATE;
    this.setSharing(hasActiveLicense || isGrandfathered);
  }
}
```

## Test Plan

- [x] `npx vitest run test/globalConfig/cloud.test.ts` — all 38 tests pass, including 5 new on-prem cases
- [x] `npx vitest run test/util/sharing.test.ts` — all 12 tests pass
- [x] `npm run tsc` — no type errors
- [x] `npm run l` — no lint errors

### DIFF
--- a/src/globalConfig/cloud.ts
+++ b/src/globalConfig/cloud.ts
@@ -6,8 +6,22 @@ export const CLOUD_API_HOST = 'https://api.promptfoo.app';
 
 export const API_HOST = getEnvString('API_HOST', CLOUD_API_HOST);
 
+const CLOUD_HOSTNAMES = new Set([
+  new URL(CLOUD_API_HOST).hostname,
+  new URL('https://www.promptfoo.app').hostname,
+  new URL('https://promptfoo.app').hostname,
+]);
+
 // Free customers created before this date are grandfathered into auto-share.
 export const SHARING_CUTOFF_DATE = new Date('2026-03-09T00:00:00Z');
+
+function getNormalizedHostname(url: string): string {
+  return new URL(url).hostname.toLowerCase().replace(/\.$/, '');
+}
+
+function isPromptfooCloudHost(url: string): boolean {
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thank you @KenichiSuda! We'll get this in the next release

PR #9627, opened 2026-06-05
fix(assertions): harden multimodal output grading

mldangelo:
## Summary

Follow-ups to #9617 (`feat(assertions): grade multimodal outputs`), found while auditing and live-testing that feature across providers. The grading machinery and the default OpenAI path were already correct; this PR fixes the gaps:

1. **URL-safe base64 image outputs were rejected.** `llm-rubric` errored the grade with a misleading "invalid base64" message when a provider returned base64url-encoded image data (common for HTTP/custom/Python providers forwarding upstream bytes). Now accepted and canonicalized to standard base64. _Empirically: a base64url image that previously produced a row ERROR now grades correctly._
2. **Bedrock Nova mis-graded image outputs silently.** A `bedrock:amazon.nova-*` (InvokeModel) grader received OpenAI-style `image_url` parts the Nova API rejects (`ValidationException: extraneous key [type]`), which promptfoo turned into a **grade=FAIL with "0 errors"**. Nova now translates OpenAI/Responses/Anthropic/Google image parts into native Nova image blocks — fixing Nova as both a multimodal **target** and **grader**.
3. **5 new env vars were undocumented and unregistered.** `PROMPTFOO_GRADING_MAX_IMAGES` and the four `PROMPTFOO_GRADING_IMAGE_MAX_*` size limits are now in `EnvVars` (with JSDoc) and documented. Raising `…_IMAGE_MAX_BYTES` now actually takes effect (the raw-char cap is derived from the resolved byte limit instead of a fixed default).
4. **Zero user docs + missing example.** Adds a "Grading image (multimodal) outputs" section to the `llm-rubric` reference (provider matrix, constraints, env-var table) and a runnable `examples/multimodal-output-grading`.
5. **DRY/nits.** De-duplicates `ATTACHED_IMAGE_OUTPUT_PLACEHOLDER` / `MULTIMODAL_GRADING_INSTRUCTION` into `matchers/shared.ts`; measures the per-image raw-char limit on the canonical data URI (line-wrapped base64 no longer false-trips); splits data URIs on the first comma.

## Provider validation (real evals, deterministic red/blue image, same "is it red?" rubric)

Red must PASS, blue must FAIL — proving the grader reads the pixels, not the prompt.

| Grader | Before | After |
| --- | --- | --- |
| OpenAI chat / responses | ✅ | ✅ |
| Anthropic (direct) | ✅ | ✅ |
| Bedrock Claude | ✅ | ✅ |
| Google AI Studio (`google:gemini-2.5-flash`) | ✅ | ✅ |
| Vertex Gemini (`vertex:gemini-2.5-flash`) | ✅ | ✅ |
| Promptfoo remote grader | ✅ | ✅ |
| **Bedrock Nova (`bedrock:amazon.nova-lite`)** | ❌ silent mis-grade | ✅ |
| URL-safe base64 image | ❌ row ERROR | ✅ |

(Vertex Claude couldn't be live-tested — Anthropic models aren't enabled in the test project — but it shares the already-validated `anthropic` format route.)

## Tests

- `test/matchers/llm-rubric.test.ts`: URL-safe base64 (raw + data-URI forms) accepted and canonicalized.
- `test/providers/bedrock/util.test.ts`: `novaNormalizeContent` for OpenAI/Responses/Anthropic/Google parts, Nova-shape/tool passthrough, and unsupported-format passthrough; `novaParseMessages` end-to-end with an `image_url` part.
- `test/redteam/providers/crescendo/index.test.ts`: target image outputs flow into the grader's `gradingContext` (the wiring #9617 added but didn't cover).

`npm run tsc` clean; the touched suites (matchers, assertions, bedrock, goat, crescendo, iterativeTree, redteam base, remoteGrading) are green (667 tests). Local Biome stack-overflows on this machine, so formatting is left to CI Style Check; Prettier passes on the docs/example.

## Refuted during the audit (no action — included for reviewers)

- **Refusal short-circuit "false positive"**: the grader receives the refusal text **and** the image together, and `MULTIMODAL_GRADING_INSTRUCTION` steers it to the visual content — grading refusal+image is required to catch harmful-image attacks. Correct as-is.
- **SSRF via image URLs**: `http(s)://` image outputs throw (never fetched). Safe by default. (#9617's body described URL hydration + a `PROMPTFOO_ALLOW_GRADING_IMAGE_PRIVATE_URLS` escape hatch that never merged — the merged behavior is the safer one; docs now state "base64/data-URI only".)

## Deliberately deferred (happy to fold in if preferred)

- A full `runRedteamConversation` image test for `iterativeTree` (no existing grading-path harness there to mirror; the wiring is identical type-checked plumbing now represented by the goat + crescendo tests).
- Consolidating the bespoke data-URI parser onto `src/util/dataUrl.ts` — the grading parser intentionally strips internal whitespace, which `parseDataUrl` does not, so a swap would change behavior. Left with a note rather than risk a regression.
- A `grading.imageFormat` override for opaque/custom graders (new config surface; warrants its own discussion).


ghost:

<a href="https://app.usetusk.ai/app/testing-commit-check/fd3fb4d4-f3d2-48e3-9aa9-ae1907c7b61d?client=e91e3433-8fb4-4e98-b233-bba492f4909c" target="_blank" rel="noopener noreferrer"><picture><source media="(prefers-color-scheme: dark)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://storage.googleapis.com/use-tusk-dev-static/pr-buttons/view-output-light.png">View output</picture></a>


<sub>Last run on 7d2e6ad0819bab21fbd8c494bc05b6e26f78b7ad. Comment `@use-tusk generate` to create a test suite on the latest commit. Configure <a href="https://app.usetusk.ai/app/repos/3589/unit-tests/test-check?client=e91e3433-8fb4-4e98-b233-bba492f4909c">here</a>.</sub>


<p>

|--------|--------|--------|
| 368d7179783f4d1eacfb5cc3217920d06512ce32 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/0c8d1fa1-a689-495b-8e95-c6e3c2253a38?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 4:45PM |
| 57c615f2d452c4bf6fc698008ff6644ce6e16e3c | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/d194e508-1ada-4b0d-b389-c45bc3d6297c?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 4:52PM |
| 730311c8b275ba7d706304d5528e5080dfb8be44 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/065fc971-e6c8-4286-aa58-8953b01d3a2b?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 5:49PM |
| aa01d49672203932dbd2e06a346dd55e88ab0c6f | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/81a5fc08-bead-4389-a522-fd67aa3bd352?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 6:01PM |
| ca08af2cddd8cb81a9372cc2727d89dc7d9995a0 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/90b890c3-44ec-4e43-955b-52e3283d6a0f?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 6:07PM |
| db7a880ccde2bebabf6534bbb57388373a6b84ee | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/b7e75d72-906f-48c9-a664-e3cad1ad49ca?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 9:04PM |
| a9a91eca1d91b0770edcd8209fcbb38cac0a2105 | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/5413b559-f7f7-498e-8a88-d52e87504fed?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 5, 2026 9:55PM |
| f5213f109a8f7b25e231072a543a0d425f4a65ee | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/d2da8d5f-d498-4c52-a68b-932dcaa3ddef?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 6, 2026 7:23PM |
| b7d104521f0795ee86ee1e365dbaed96fbcbfa5c | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/a49b23e4-65d8-42a2-8bb4-7832349575d3?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 6, 2026 7:31PM |
| 6c16cbfcc251b17e70505b01602d52d29b8f666e | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/47075552-f704-4c75-944e-87e87c25a0af?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 6, 2026 7:55PM |
| 7d2e6ad0819bab21fbd8c494bc05b6e26f78b7ad | No test execution environment matched · [Output](https://app.usetusk.ai/app/testing-commit-check/fd3fb4d4-f3d2-48e3-9aa9-ae1907c7b61d?client=e91e3433-8fb4-4e98-b233-bba492f4909c) | Jun 7, 2026 12:34AM |

</p>
</details>

examples/multimodal-output-grading/promptfooconfig.yaml
```diff
@@ -0,0 +1,32 @@
+# yaml-language-server: $schema=https://promptfoo.dev/config-schema.json
+description: Grade image (multimodal) outputs with a vision-capable LLM judge
+
+prompts:
+  - 'A {{color}} bicycle leaning against a brick wall'
+
+# The target generates an image. Its image output is attached to the grading
+# prompt automatically so the judge can inspect the actual pixels.
+providers:
+  - id: openai:image:gpt-image-1
```
mldangelo-oai, line 10: **[P1] This advertised quick start still fails under the default media settings.** The evaluator externalizes every normal (>1 KiB) data-URI image to `images[].blobRef` before assertions run, while `llm-rubric` rejects blob-backed images. I reproduced this on `57c615f2`: a deterministic 2 KiB image produced `0 passed / 1 error` with `Blob-backed image outputs are not supported`; the identical eval passed only with `PROMPTFOO_INLINE_MEDIA=true`. A real `gpt-image-1` output is necessarily above this threshold. Please resolve/materialize blob refs for grading (or defer externalization until after assertions) and add an evaluator-level regression test. Merely documenting the inline-media opt-out would also conflict with the README claim that the heavy base64 is not persisted, because the passing control retained the full data URI in the exported result.

site/docs/configuration/expected-outputs/model-graded/llm-rubric.md
```diff
@@ -166,6 +166,65 @@ applies to other model-graded assertions that use a text judge; see the
...
+| Google AI Studio / Vertex | `google:gemini-2.5-flash`, `vertex:gemini-2.5-flash` | `inlineData`             |
+| Promptfoo remote grader   | _(default when no grader is configured)_             | data URI                 |
+
+Only base64 or data-URI image outputs are supported. Remote `http(s)://` image URLs and blob-backed
+outputs are rejected (to avoid the grader fetching arbitrary URLs); configure the provider to return
+inline base64/data-URI image output instead.
+
+### Image grading limits
+
+These limits guard against oversized payloads and can be tuned via environment variables:
```
mldangelo-oai: **[P2] These overrides do not apply end-to-end when the remote grader is used.** Core honors the five environment variables, but the Cloud `llm-rubric` request schema still hard-codes four images and 20 MiB per-image/total limits and reads none of them ([server contract](https://github.com/promptfoo/promptfoo-cloud/blob/1801f23fee45ac3c93e6a8e3edadefdf8e301e7e/server/src/task/llmRubric.ts#L13-L21), [validation](https://github.com/promptfoo/promptfoo-cloud/blob/1801f23fee45ac3c93e6a8e3edadefdf8e301e7e/server/src/task/llmRubric.ts#L128-L176)). For example, setting `PROMPTFOO_GRADING_MAX_IMAGES=5` allows local materialization but the default remote path still returns a validation error. Please distinguish local limits from fixed remote caps here, or align the server contract.

mldangelo-oai (changes requested):
Requesting changes based on exact-head review and runtime QA at `57c615f2d452c4bf6fc698008ff6644ce6e16e3c`. The primary new quick-start workflow still fails under Promptfoo's default media behavior, and the documented limit overrides do not hold for the default remote-grading path.

I also independently reproduced the current unresolved raw-character-limit bypass and confirmed the existing base64url placeholder/padding findings, so I have not duplicated those inline comments. The Nova system-prompt comments appear to describe pre-existing code rather than regressions introduced by this PR.

Verification: focused matcher, assertion, Bedrock, redteam, and remote-grading tests passed; TypeScript compilation passed; the exact-head default CLI eval produced `0 passed / 1 error`, while the same eval with `PROMPTFOO_INLINE_MEDIA=true` passed. All current GitHub checks are green, but they do not exercise this evaluator-to-blob-to-grader integration.

mldangelo:
Thanks for the thorough review — all findings addressed in the latest commits.

### P1 (blocker) — blob-backed image outputs now resolve for grading
The evaluator externalizes image outputs >1KiB to `images[].blobRef` (and rewrites `output` to a blob URI) before assertions run, and grading rejected blobs — so the headline workflow errored by default. Fixed:
- Added `resolveBlobBackedImageOutputs()` (`src/matchers/rubric.ts`) — reads the blob via `getBlobByHash` and inlines it as a data URI for the grader. The resolved bytes are used transiently; base64 is **not** persisted.
- Blob-URI `output` text is now replaced with the attached-image placeholder, so the README's "heavy base64 is not persisted" claim holds.
- New matchers→blobs (`core -> legacy-runtime`) edge recorded in `architecture/edge-baseline.json`.

**Live proof** — the exact example, default media (no `PROMPTFOO_INLINE_MEDIA`), real `gpt-image-1`:
```
red  → PASS  imgs=1  outBlobUri=true  imgBlobRef=true  sizeBytes=2682412  reason: "predominantly red"
blue → PASS  imgs=1  outBlobUri=true  imgBlobRef=true  sizeBytes=2814453  reason: "predominantly blue"
2 passed (100%), 0 errors   (was 0 passed / 1 error)
```
No big base64 in the stored grading prompt. Added a filesystem-backed integration test (`test/matchers/multimodalGradingBlobs.test.ts`) plus matcher-level resolve / resolve-failure tests.

### P2 — raw-character limit bypass (Codex)
Raw-char caps are now enforced on the **original** `image.data` length **before** the whitespace strip in `materializeImageOutputsForGrading`, so a whitespace-padded payload can't force a huge allocation/scan before any limit applies. (`PROMPTFOO_GRADING_IMAGE_MAX_RAW_CHARS=20` reject test still passes.)

### P2 — base64url placeholder detection (Codex)
`getGradingOutputForImages` now compares the output against the **original (pre-normalization)** image data, so a raw base64url output (`q-_z`) still matches its attached image and gets the placeholder instead of leaking into the grader text. Added a regression test.

### P2 — remote grader limits (docs)
Documented that the five `PROMPTFOO_GRADING_IMAGE_*` vars apply to **local** grading only; the remote grader enforces its own fixed caps (4 images, 20 MiB per image/total) and does not read them.

### Copilot — integer byte values (docs)
The limits table now shows integer byte counts (`20971520` / "20 MiB") and notes that unit suffixes like `20MB` are not parsed.

### Acknowledged, not changed
- **Nova `system` double-wrap (`bedrock/index.ts:1298/1384`)** — confirmed pre-existing (only the non-JSON-array / `JSON.parse`-failure paths, which grading doesn't hit) and out of scope for this PR, as noted in the review. Happy to fix separately.
- **`toStandardBase64` padding of unpadded *standard* base64** — intentionally left as-is to preserve prior behavior (and avoid re-padding the existing `abc123` fixtures); only URL-safe input is canonicalized. The call-site comment documents this.

Note: local Biome stack-overflows on this machine, so I rely on CI Style Check for formatting; `npm run tsc`, `npm run architecture:check`, and the affected suites (711 tests) pass locally.

src/matchers/rubric.ts
```diff
@@ -302,13 +323,65 @@ function imageOutputToImageUrl(
...
+      } catch (error) {
+        throw new Error(
+          `Failed to load blob-backed image output for multimodal grading (blob ${hash.slice(0, 12)}…): ${error}`,
+        );
+      }
+
+      const mimeType = image.mimeType || blob.metadata.mimeType || 'image/png';
+      return {
+        ...image,
+        data: `data:${mimeType};base64,${blob.data.toString('base64')}`,
```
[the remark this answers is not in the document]
mldangelo-oai: Confirmed with an instrumented exact-head probe. With `PROMPTFOO_GRADING_MAX_IMAGES=1`, three repeated valid blob refs caused 3 blob reads and 3 base64 encodes before the count check rejected them; an oversized blob was likewise read/encoded before the byte check. Please move the count check ahead of resolution and validate `blob.data.length` plus cumulative bytes before encoding, ideally with bounded concurrency.

architecture/edge-baseline.json
```diff
@@ -14,7 +14,7 @@
   "cli -> view-server": 6,
   "core -> contracts": 1,
   "core -> legacy-contracts": 94,
-  "core -> legacy-runtime": 47,
+  "core -> legacy-runtime": 48,
```
mldangelo-oai: **[P2] Do not ratchet `core -> legacy-runtime` upward for this fix.** `rubric.ts` now imports the runtime blob store directly and this changes the reviewed baseline from 47 to 48. The repository architecture policy explicitly says not to refresh this baseline merely to permit a newly introduced dependency; this also reverses the earlier PR commit that removed new cross-layer edges. Resolve blobs in the evaluator/runtime layer, or inject a narrow blob resolver into the matcher, rather than coupling core matcher code to storage.

examples/multimodal-output-grading/README.md
```diff
@@ -0,0 +1,47 @@
...
+```yaml
+defaultTest:
+  options:
+    provider: anthropic:messages:claude-haiku-4-5
+    # provider: bedrock:amazon.nova-lite-v1:0
+    # provider: google:gemini-2.5-flash
+    # provider: vertex:gemini-2.5-flash
+```
+
+Only base64 / data-URI image outputs are supported. Remote `http(s)://` image URLs and blob-backed outputs are rejected so the grader never fetches arbitrary URLs. See the [llm-rubric docs](https://www.promptfoo.dev/docs/configuration/expected-outputs/model-graded/llm-rubric/#grading-image-multimodal-outputs) for the full provider matrix and the `PROMPTFOO_GRADING_IMAGE_*` size limits.
```
mldangelo-oai: **[P2] This support statement now contradicts the implementation and site docs.** Blob-backed outputs are the normal default for images over 1 KiB, and this head resolves them successfully for grading. Saying they are rejected tells users to avoid the exact path this fix enables. Please state that inline base64/data URIs and Promptfoo blob references derived from them are supported, while remote `http(s)://` image URLs remain rejected.

site/docs/configuration/expected-outputs/model-graded/llm-rubric.md
```diff
@@ -166,6 +166,70 @@ applies to other model-graded assertions that use a text judge; see the
...
+
+| Variable                                      | Default                      | Purpose                                      |
+| --------------------------------------------- | ---------------------------- | -------------------------------------------- |
+| `PROMPTFOO_GRADING_MAX_IMAGES`                | `4`                          | Max images attached to one grading prompt    |
+| `PROMPTFOO_GRADING_IMAGE_MAX_BYTES`           | `20971520` (20 MiB)          | Max decoded size of a single image           |
+| `PROMPTFOO_GRADING_IMAGE_MAX_TOTAL_BYTES`     | `20971520` (20 MiB)          | Max combined decoded size of all images      |
+| `PROMPTFOO_GRADING_IMAGE_MAX_RAW_CHARS`       | derived from max bytes       | Max raw base64/data-URI characters per image |
+| `PROMPTFOO_GRADING_IMAGE_MAX_TOTAL_RAW_CHARS` | derived from max total bytes | Max combined raw characters across images    |
+
+:::note
```
mldangelo-oai, line 232: **[P2] The local/remote contract described here is incorrect.** `materializeImageOutputsForGrading()` runs before the remote/local branch, so these environment limits also constrain remote requests client-side; setting `PROMPTFOO_GRADING_MAX_IMAGES=0` rejects a remote-eligible request before `doRemoteGrading`. Separately, ordinary `llm-rubric` without a provider does not default to Promptfoo remote grading: that branch requires an eligible redteam config, otherwise `getDefaultProviders()` supplies the local default grader. Please correct the provider matrix, limit note, and privacy guidance; remote grading has additional immutable Cloud caps, not an exemption from the client limits.

test/matchers/multimodalGradingBlobs.test.ts
```diff
@@ -0,0 +1,70 @@
...
+    setBlobStorageProvider(provider);
+  });
+
+  afterEach(() => {
+    resetBlobStorageProvider();
+    fs.rmSync(tempDir, { recursive: true, force: true });
+  });
+
+  it('resolves a blob-backed image output and attaches it to the grader', async () => {
+    // Store a >1KiB image in the (real, filesystem) blob store, mirroring what the
```
mldangelo-oai, line 38: **[P2] Add the evaluator-level regression requested for the original blocker.** This test stores a blob directly, manually constructs the post-externalization response, and calls `matchesLlmRubric`; it never exercises the evaluator/extractor step that rewrites an inline image before assertions. Please add an eval/evaluator test that starts with an inline image over 1 KiB, observes externalization, runs the assertion, and verifies the exported result contains the blob reference/placeholder but not the base64. That is the integration that broke and is still not locked by CI.

mldangelo-oai (changes requested):
Re-reviewed exact head `730311c8b275ba7d706304d5528e5080dfb8be44` after refreshing current `main` and rerunning runtime QA.

The original blocker is fixed: a default CLI eval now externalizes a 2 KiB image, resolves the blob for `llm-rubric`, passes 1/1, and persists only the blob reference/placeholder rather than the base64. The base64url placeholder and whitespace-cap findings are also fixed, and the Nova conversion suites/probes pass.

I am still requesting changes because the new resolver defeats the count/byte resource guards until after unbounded blob reads and base64 expansion (validated in the existing current-head Codex thread, including a 3-read/3-encode count-limit probe), the documentation now contains contradictory execution/support contracts, the fix raises a prohibited architecture edge baseline, and the original evaluator integration remains outside CI coverage.

Verification: 670 focused matcher/assertion/Bedrock/redteam/remote tests passed; TypeScript and architecture checks passed; direct Nova/Nova 2 shape probes passed; real local evaluator-to-blob-to-grader QA passed with no base64 persisted. All GitHub checks are green. The PR is mergeable but one unrelated commit behind current `main`.

mldangelo:
Follow-up: ran the simplification pass and landed one perf win — `matchesLlmRubric` was materializing the image outputs and then `runJsonGradingPrompt`→`buildGradingProviderPrompt` re-ran `materializeImageOutputsForGrading` on the same multi-MB payloads; now materialized once and threaded through (plus a small `getDataUriPayload` tidy).

Heads-up on the `architecture/edge-baseline.json` diff: it shows two lines — `core -> legacy-runtime: 47→48` is this PR's intentional `matchers -> blobs` (`getBlobByHash`) edge, and `node -> legacy-runtime: 150→152` is a **pre-existing drift already on `origin/main`** (main currently fails `npm run architecture:check` with the same delta; my diff touches no node-layer source). The regen was required to get this PR's Style Check green and just absorbs that existing drift.

mldangelo:
Round-3 feedback addressed — thanks for the precise probes. Summary of the changes:

### Architecture — no more `core -> legacy-runtime` ratchet
Added an injection seam `src/matchers/imageBlobResolver.ts` (`set/getGradingBlobResolver`). The core matcher resolves blobs through it; the **evaluator** registers the real `getBlobByHash`-backed resolver at module load. `rubric.ts` no longer imports `../blobs`, so `core -> legacy-runtime` returns to **47**; the new edge is `legacy-runtime -> core: 12 -> 13` (the dependencies-point-inward direction you asked for). `node -> legacy-runtime` stays **152** (pre-existing main drift — `architecture:check` already exits 1 on `origin/main`; reverting it would fail this branch).

### Resource guards now run before reads/encode
`resolveBlobBackedImageOutputs` enforces, **in order**: (a) the image-count cap, (b) the declared `blobRef.sizeBytes` per-image + cumulative cap — **before any blob read**, then (c) re-checks the **actual** `blob.data.length` before the base64 expansion (declared size may lie / the store allows up to 50 MiB while grading caps lower). Reads use `async.mapLimit` (bounded). Your 3-read/3-encode probe now rejects after **0 reads** for the count/declared-size cases and before encoding for the lying-size case.
Residual (flagged, not over-engineered): the cumulative *post-read* byte check still lands in `materialize`; worst case is bounded at `maxImages × maxImageBytes` (default 4 × 20 MiB).

### MIME counted in the raw cap
The pre-normalization raw-char cap now adds `image.mimeType.length` for raw-base64 inputs (a `data:` URI already carries its prefix, so it isn't double-counted) — closing the tiny-payload/huge-mime data-URI bypass.

### Docs corrected
- README: blob references derived from inline images are **supported** (resolved), not rejected; only `http(s)://` URLs are rejected.
- llm-rubric.md: limits are enforced **client-side for both** local and remote grading (`MAX_IMAGES=0` rejects before any upload); ordinary `llm-rubric` without a provider uses the **local default grader** — the remote Cloud grader is the default only for red team evals — and it adds its own fixed caps.

### Real evaluator-level regression test
New `test/evaluator/multimodalGradingExternalization.test.ts` drives the actual `evaluate()` path: a provider returns an inline >1 KiB image → the evaluator externalizes it to a `blobRef` → `llm-rubric` resolves it via the registered resolver → asserts `successes===1`, the stored response has `blobRef`/blob-URI output and **no base64**, and the grader received the resolved inline image. Plus four resource-guard unit tests (count, declared-size, lying-size, mime-bypass) and the matcher tests now register the resolver.

Verification: `npm run tsc`, `npm run architecture:check` pass; 558 tests across `test/matchers` + `test/evaluator` green; live evals (inline OpenAI red/blue, Bedrock Nova, and the externalized-blob path) pass. Final baseline diff vs main: `core -> legacy-runtime` unchanged at 47, `legacy-runtime -> core` 12→13, `node -> legacy-runtime` 150→152 (stale-main correction).

src/matchers/rubric.ts
```diff
@@ -332,53 +358,177 @@ export function materializeImageOutputsForGrading(images?: ImageOutput[]): {
...
+
+  // Enforce the raw-character caps on the *original* payloads BEFORE any normalization,
+  // so a whitespace-padded or adversarial output can't force a multi-hundred-MB
+  // allocation/scan in normalizeBase64ImageData's whitespace strip. A raw base64 payload
+  // also contributes its separate mimeType to the on-wire data URI, so count that too;
+  // a data: URI already carries its prefix inside image.data, so it isn't double-counted.
+  let totalRawChars = 0;
+  for (const image of images) {
+    let rawChars = image.data?.length ?? 0;
+    if (image.data && !image.data.trim().startsWith('data:')) {
```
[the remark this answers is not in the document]
mldangelo-oai: Confirmed independently on exact head `db7a880cc`. With `PROMPTFOO_GRADING_IMAGE_MAX_RAW_CHARS=1`, an already oversized instrumented payload reached `trim()` and threw `trim-called-before-cap` instead of the size-limit error. The length check needs to happen before any `trim()`/normalization; once the payload is known to be within the cap, the bounded prefix check can safely decide whether separate MIME metadata should count.

src/matchers/rubric.ts
```diff
@@ -332,53 +358,177 @@ export function materializeImageOutputsForGrading(images?: ImageOutput[]): {
...
+    if (blob.data.length > maxImageBytes) {
+      throw new Error(
+        `Image output exceeds multimodal grading size limit: ${blob.data.length} bytes, maximum is ${maxImageBytes}.`,
+      );
+    }
+
+    const mimeType = image.mimeType || blob.mimeType || 'image/png';
+    return {
+      ...image,
+      data: `data:${mimeType};base64,${blob.data.toString('base64')}`,
```
mldangelo-oai, line 567: **[P2] Enforce the actual cumulative byte cap before base64 expansion.** The preflight total trusts `blobRef.sizeBytes`, while this callback only checks each actual blob individually and immediately encodes it. With a 9-byte total cap, two refs declaring 1 byte each but resolving to 5 bytes each both performed their reads and base64 conversions before `materializeImageOutputsForGrading()` finally rejected the actual 10-byte total. This still permits up to `maxImages × maxImageBytes` of buffers plus base64 strings when metadata is missing/inconsistent. Track actual cumulative bytes before `toString('base64')` (and account for concurrent reads safely), or separate resolution/validation from encoding.

architecture/edge-baseline.json
```diff
@@ -23,14 +23,14 @@
   "legacy-contracts -> node": 3,
   "legacy-contracts -> redteam": 11,
   "legacy-runtime -> contracts": 2,
-  "legacy-runtime -> core": 12,
+  "legacy-runtime -> core": 13,
```
mldangelo-oai: **[P2] Do not replace the old dependency with a target-order back-edge.** Moving blob access out of `core` fixes the prior `core -> legacy-runtime` increase, but the new evaluator import raises `legacy-runtime -> core` from 12 to 13. The architecture checker explicitly lists this direction as a remaining back-edge, and the repository policy says not to refresh the baseline merely to permit a new dependency. Thread the resolver through the existing evaluator/assertion call contract (or another lower-level port) without adding another runtime-to-matcher import. The unrelated `node -> legacy-runtime` 150→152 ratchet on line 33 should also be handled separately from this PR.

examples/multimodal-output-grading/promptfooconfig.yaml
```diff
@@ -0,0 +1,32 @@
...
+    config:
+      size: 1024x1024
+
+defaultTest:
+  options:
+    # Any vision-capable grader works: OpenAI (chat/responses), Anthropic,
+    # Bedrock Claude/Nova, Google Gemini (AI Studio or Vertex), or the default
+    # remote grader. Swap this id to grade with a different provider.
```
mldangelo-oai: **[P3] Remove the stale “default remote grader” guidance.** The corrected site docs and runtime routing now accurately say that ordinary `llm-rubric` uses the local/default configured grader; Promptfoo Cloud remote grading is only the no-provider fallback for eligible red-team evals. This runnable example still advertises the old contract even though it explicitly configures OpenAI below.

mldangelo-oai (changes requested):
Re-reviewed exact head `db7a880ccde2bebabf6534bbb57388373a6b84ee` after refreshing current `main` and rerunning the full runtime/CI review.

The previous blockers are substantially addressed: default evaluator externalization now grades successfully without persisting base64; count, declared-size, per-image actual-size, MIME, and resolver-error guards are covered; the evaluator-level regression is faithful; docs now describe local/remote behavior correctly; and Nova/Nova 2 formatting remains green.

I am still requesting changes for three current-head issues: the raw cap still performs an unbounded `trim()` before rejection (validated in the existing Codex thread), under-reported blobs are all base64-expanded before the actual cumulative-byte cap runs, and the injected resolver increases a declared target-order architecture back-edge. There is also a small stale example comment about the “default remote grader.” The baseline additionally absorbs unrelated `node -> legacy-runtime` drift from current `main`; that should be repaired or ratcheted separately rather than hidden in this feature PR.

Verification: real CLI evaluator-to-blob-to-grader eval passed 1/1 with no base64/data URI in exported JSON; 675 focused tests passed; three adversarial guard probes passed; TypeScript and PR architecture checks passed; all 37 current-head GitHub checks pass (Tusk skipped). Current `main` independently fails architecture checking on the unrelated `node -> legacy-runtime` drift.

mldangelo-oai:
Follow-up after comparing draft #9628 against exact head `db7a880ccde2bebabf6534bbb57388373a6b84ee`:

I do not see another changed-line blocker to add from that draft. Its useful contribution is stronger grader guidance that accompanying text, metadata, or the mere existence of an attachment is not evidence of image compliance. If that wording is folded into this PR, it should be adapted into the existing multimodal instruction rather than cherry-picking #9628's keyword-triggered rubric rewrite; that heuristic false-triggers on text-only uses such as "visual imagery," skips object rubrics, and can disappear with custom `rubricPrompt` or `vars.rubric` handling.

One correction is needed in this PR's description, though: the **“Refusal short-circuit false positive … Correct as-is”** conclusion is too broad. The shared base grader correctly suppresses its refusal shortcut when images exist ([base.ts#L528-L566](https://github.com/promptfoo/promptfoo/blob/db7a880ccde2bebabf6534bbb57388373a6b84ee/src/redteam/plugins/base.ts#L528-L566)), but several specialized overrides return a passing grade from refusal text before consulting `gradingContext` images: [Harmful](https://github.com/promptfoo/promptfoo/blob/db7a880ccde2bebabf6534bbb57388373a6b84ee/src/redteam/plugins/harmful/graders.ts#L50-L80), [Aegis](https://github.com/promptfoo/promptfoo/blob/db7a880ccde2bebabf6534bbb57388373a6b84ee/src/redteam/plugins/aegis.ts#L108-L137), [Beavertails](https://github.com/promptfoo/promptfoo/blob/db7a880ccde2bebabf6534bbb57388373a6b84ee/src/redteam/plugins/beavertails.ts#L430-L480), and [SSRF](https://github.com/promptfoo/promptfoo/blob/db7a880ccde2bebabf6534bbb57388373a6b84ee/src/redteam/plugins/ssrf.ts#L99-L138). A focused exact-head probe with refusal text plus an attached image reproduced `pass: true` from `HarmfulGrader` before image grading.

Those grader overrides are pre-existing and not changed by #9627, so I am not treating this as another inline blocker. Please remove or soften the “Correct as-is” claim and track the image-aware refusal gating as a focused follow-up, unless you intentionally want to fold that broader red-team fix into this PR.

mldangelo:
Round-4 addressed:

### Architecture — no new edge in either direction
Replaced the global resolver registry with a resolver **threaded as data** through the existing call contract: the evaluator builds the `getBlobByHash`-backed resolver and passes it via `runAssertions → runAssertion → providerCallContext.resolveImageBlob → handleLlmRubric → matchesLlmRubric`. `CallApiContextParams` (already imported everywhere) carries it; the `GradingBlobResolver` type lives in `src/types/providers.ts`. **No new import edge** — deleted `src/matchers/imageBlobResolver.ts` and its module-global state. The edge-baseline diff vs `main` is now **only** the `node → legacy-runtime 150→152` line; `core → legacy-runtime` is back to **47** and `legacy-runtime → core` back to **12**.

### Cumulative actual-byte cap before base64 expansion
`resolveBlobBackedImageOutputs` now runs in stages: read + per-image actual-size check (bounded concurrency) → verify the **cumulative actual-byte total** → and only then `toString('base64')`. Your two-refs-declaring-1-byte/resolving-5 case now rejects after the reads but **before any base64 expansion** (no `maxImages × maxImageBytes` of base64 strings).

### Raw cap before trim
The raw-char guard checks `image.data.length` first and uses a **bounded** `data:` prefix slice (`/^\s*data:/i.test(data.slice(0, 64))`) instead of a full `.trim()`, so an oversized whitespace-padded payload is rejected without allocating a trimmed copy. Your `PROMPTFOO_GRADING_IMAGE_MAX_RAW_CHARS=1` probe now throws the size-limit error, not from inside `trim()`.

### Example comment
Dropped the stale "default remote grader" line from `promptfooconfig.yaml`.

### On the `node → legacy-runtime` baseline line
You're right that it's unrelated to this feature — it's a pre-existing drift that already fails `architecture:check` on `origin/main` itself (the committed baseline says 150; the actual code measures 152). This branch can't go green without the 152 value, since the check runs against the merged tree. I'm happy to open a **separate** tiny PR that ratchets `main`'s baseline to 152 to un-break `main`'s check, after which I'll rebase this branch so its baseline diff is empty on that line. Just say the word and I'll put it up.

Verification: `npm run tsc` + `npm run architecture:check` pass; 1892 tests across `test/matchers`, `test/assertions`, `test/evaluator`, and redteam green; the live evaluator-to-blob-to-grader eval passes 1/1 with no base64 persisted.

mldangelo:
Ran another simplification pass and, doing so, caught a regression my own round-4 refactor introduced — fixed it.

**Simplification:** `getBlobHashForImage` (which regex-scans `image.data`) was computed twice per image; now computed once and reused.

**Regression fix (important):** moving the resolver from a global registry to a value threaded through `runAssertions` left the **redteam grading path uncovered** — `RedteamGraderBase.getResult` (used by the multi-turn strategy providers and `handleRedteam`) doesn't flow through `runAssertions`. Since redteam target responses are externalized to blobRefs before grading, redteam image grading would have thrown `no blob resolver provided`. I threaded the resolver to the redteam path too, **with no new cross-layer import edges** (providers read it from their `callApi` context, `handleRedteam` from `providerCallContext`, forwarded via `RedteamGradingContext` → `getResult` → `matchesLlmRubric`), plus a regression test. Edge-baseline diff vs main is unchanged (still only the pre-existing `node→legacy-runtime` line).

2324 tests pass across matchers/assertions/evaluator/redteam; tsc + architecture green; live blob eval still 1/1.

mldangelo:
### Default-providers QA (incl. xAI) — done

Added a key-free routing regression test (`test/matchers/multimodalGradingFormatRouting.test.ts`, 13 cases) that pins the multimodal format for **every default grader** in `defaults.ts`: OpenAI/Azure/Mistral/GitHub → `image_url`; xAI (`xai:responses:grok-4.3`)/OpenAI-responses → `input_image`; Anthropic/Bedrock-Claude/Vertex-Claude → base64 `image` block; Google AI Studio/Vertex Gemini → `inlineData`; Bedrock Nova → `image_url` translated by the provider. Live-validated (red PASS / blue FAIL, `imgs=1`): OpenAI chat+responses, Anthropic, Bedrock Claude, **Bedrock Nova**, Google AI Studio, Vertex Gemini, the remote grader, and **xAI grok-4.3** (via OpenRouter — native `xai:` transport needs `XAI_API_KEY`, which isn't in this env). One flag: the **Codex SDK** grader (`OpenAICodexSDKProvider`) — unclear whether the agentic SDK accepts image inputs; worth confirming as a follow-up (niche).

### CI status — the 3 red checks are not from this PR

- **Site tests** and **webui tests**: failing on the **Codecov upload step** (`Could not verify signature. Please contact Codecov…`) — a Codecov CLI v11.2.7 infra issue, not the tests (which pass). Affects unrelated jobs.
- **Style Check**: biome stack-overflows on the runner (and locally) and exits 0 when it crashes, so this check usually passes by *skipping* lint. This run biome ran fully and surfaced **169 pre-existing `noExcessiveCognitiveComplexity` warnings across untouched files** plus **1 error whose diagnostic biome doesn't print** (not a format error, not in any file this PR touches; it's stable while the warning count moved 170→169 after I trimmed my one new warning). I fixed the formatting biome flagged in my two test files and reduced my complexity contribution; the remaining error is pre-existing repo lint debt unmasked by biome happening to run.

This PR is otherwise green: `npm run tsc`, `npm run architecture:check`, and 2300+ matcher/assertion/evaluator/redteam tests pass. A re-run should clear Site/webui (once Codecov recovers) and Style Check (when biome crashes as usual). Happy to file the biome-flakiness and Codecov-signature problems as separate CI-health issues if useful.

src/assertions/index.ts
```diff
@@ -611,6 +614,7 @@ export async function runAssertion({
         prompt: { raw: prompt || '', label: '' },
         vars: resolvedVars,
         ...(graderTraceparent && { traceparent: graderTraceparent }),
+        ...(resolveImageBlob && { resolveImageBlob }),
```
jameshiester-oai, line 617: **[P2] Preserve an explicit resolver when `provider` is omitted**

`provider` is optional in the public `runAssertion`/`runAssertions` APIs, but this resolver is nested inside a context that is created only when `provider` is present. A direct programmatic call with a grading provider in `test.options`, a blob-backed response, and an explicit `resolveImageBlob` therefore failed with `no blob resolver provided`; the supplied resolver was called zero times. Build the context when either `provider` or `resolveImageBlob` is present (or pass the resolver independently), and cover the provider-omitted path.

src/matchers/rubric.ts
```diff
@@ -192,7 +189,10 @@ function isChatMessageArray(value: unknown): value is ChatMessageLike[] {
 }
 
 function isValidBase64Payload(data: string): boolean {
-  if (!data || data.length % 4 === 1 || !/^[A-Za-z0-9+/]*={0,2}$/.test(data)) {
+  // Accept both the standard (`+`/`/`) and URL-safe (`-`/`_`) base64 alphabets.
+  // Providers that forward upstream bytes (HTTP/custom/Python) commonly emit
+  // base64url, and `ImageOutput.data` is documented as "data URI or base64".
+  if (!data || data.length % 4 === 1 || !/^[A-Za-z0-9+/_-]*={0,2}$/.test(data)) {
```
jameshiester-oai, line 195: **[P2] Reject partially padded base64 before canonicalizing it**

This accepts inputs such as `a=` and `a==`: both satisfy the current character/padding checks, then `toStandardBase64` appends more padding and emits invalid `a===`. The byte calculation reports one byte even though Node decodes that payload to zero bytes, so the grader receives malformed/empty image data instead of a local validation error. If `=` is present, require a complete four-character quantum and legal padding placement before adding any padding; add negative cases for these forms.

src/matchers/llmGrading.ts
```diff
@@ -146,6 +154,14 @@ function getGradingOutputForImages(llmOutput: string, imageOutputs: ProviderResp
     return ATTACHED_IMAGE_OUTPUT_PLACEHOLDER;
   }
 
+  // The evaluator externalizes large image outputs, so `output` is often a blob URI
+  // (e.g. `promptfoo://blob/<hash>`) rather than the base64 itself. Replace it too.
+  if (/^promptfoo:\/\/blob\/[a-f0-9]{64}/i.test(trimmedOutput)) {
```
jameshiester-oai, line 159: **[P2] Preserve text that follows a blob URI**

This prefix-only match classifies the entire output as image-only. With `promptfoo://blob/<64 hex>\nThis textual claim must also be graded.` plus one attached image, the captured grader prompt contained only the image placeholder; the trailing text was discarded. That can hide adversarial or rubric-relevant content and produce incorrect grades. Substitute only when the trimmed output exactly matches an attached blob reference, and preserve mixed surrounding text.

src/matchers/llmGrading.ts
```diff
@@ -186,8 +203,19 @@ export async function matchesLlmRubric(
...
+  // resolver is injected by the evaluator through the grading call contract.
+  const resolvedImages = await resolveBlobBackedImageOutputs(
+    options?.providerResponse?.images,
+    options?.resolveImageBlob,
+  );
+  // Materialize once here and reuse downstream; runJsonGradingPrompt no longer
+  // re-validates/re-decodes the (potentially multi-MB) image payloads.
+  const { imageOutputs, imageData } = materializeImageOutputsForGrading(resolvedImages);
+  const gradingOutput = imageOutputs.length
+    ? getGradingOutputForImages(llmOutput, options?.providerResponse?.images)
```
jameshiester-oai, line 217: **[P2] Remove structured image-only payloads from grading text**

`getGradingOutputForImages` only recognizes scalar image data/URIs. In an exact-head probe where `output` was a JSON array of two data URIs and `images` contained those two images, both images were attached but the full base64 array remained in `renderedGradingPrompt` and in the provider request. The externalized `b64_json` multi-image form similarly leaves a JSON list of blob hashes in the grading text. Recognize image-only structured forms when every element corresponds to an attached image, while retaining mixed descriptive text; add a two-image evaluator regression.

src/matchers/rubric.ts
```diff
@@ -332,51 +358,272 @@ export function materializeImageOutputsForGrading(images?: ImageOutput[]): {
...
+
+  // (a) Count cap BEFORE any blob read.
+  if (images.length > maxImages) {
+    throw new Error(
+      `Too many images for multimodal grading: received ${images.length}, maximum is ${maxImages}.`,
+    );
+  }
+
+  // Compute each image's blob hash once (it scans image.data via regex) and reuse below.
+  const hashedImages = images.map((image) => ({ image, hash: getBlobHashForImage(image) }));
```
jameshiester-oai, line 442: **[P2] Apply raw-size caps before scanning `image.data` for a blob URI**

This maps every image through the unanchored blob regex before the raw-character checks at lines 456-459. An oversized non-blob payload therefore incurs a full O(n) scan (up to the image-count cap) even when `PROMPTFOO_GRADING_IMAGE_MAX_RAW_CHARS` would reject it. An instrumented probe with the cap set to 1 observed the 100-character blob-regex scan before the limit error. Check per-image/cumulative lengths first and parse a data-field blob reference only as a bounded exact URI.

src/matchers/rubric.ts
```diff
@@ -332,51 +358,272 @@ export function materializeImageOutputsForGrading(images?: ImageOutput[]): {
...
+      );
+    }
+  }
+
+  // (e) Enforce the exact raw-character caps before allocating base64 strings.
+  for (const { image, blob } of slots) {
+    if (!blob) {
+      continue;
+    }
+    const mimeType = image.mimeType || blob.mimeType || 'image/png';
```
jameshiester-oai, line 552: **[P2] Preserve the MIME type carried by `blobRef`**

Both `image.mimeType` and the resolver MIME are optional, while `BlobRef.mimeType` is required. The current fallback skips that required source, so a JPEG blob reference resolved by a byte-only resolver is counted and encoded as `data:image/png`; the same precedence is repeated at line 564. Include `image.blobRef?.mimeType` in a single shared MIME selection before raw-size accounting and encoding, and cover a non-PNG resolver-without-MIME case.

jameshiester-oai (commented):
## Codex review

Reviewed base `c54a30668ad8319d76c20ae96e6680ad6c51a2c6` through head `7d2e6ad0819bab21fbd8c494bc05b6e26f78b7ad`. This was a full review of the current PR diff; there was no prior authenticated completion marker for an incremental review.

### Verdict

The externalize → resolve → grade path works end to end, and the Nova/provider-format changes are well covered, but six P2 correctness/resource defects remain. The PR is also currently `CONFLICTING` / `DIRTY` against `main`.

### Verified findings

**P0/P1:** None.

**P2:**

- Programmatic `runAssertion` / `runAssertions` calls discard an explicitly supplied blob resolver when the optional target provider is omitted.
- The base64 validator accepts partially padded inputs and canonicalizes them into invalid payloads such as `a===`.
- Blob-URI-prefixed output can hide trailing text from the grader.
- Structured image-only outputs containing multiple images remain in the grading text, duplicating base64 or persisting blob hashes.
- Oversized `image.data` is regex-scanned for a blob URI before the configured raw-character cap runs.
- Resolved blobs ignore the required `blobRef.mimeType` when both optional MIME sources are absent, mislabeling non-PNG images as PNG.

Inline comments contain the triggering conditions, reproductions, impact, and correction direction.

### Existing feedback and CI

All 20 existing review threads are resolved. I rechecked their current-head fixes (pre-normalization limits, declared/actual/cumulative blob caps, mixed inline/blob accounting, evaluator integration, Nova shapes, architecture edges, and docs) and did not duplicate them.

Most checks are green. The three red jobs are unrelated Codecov infrastructure failures after their test commands passed: Site `6 files / 101 tests`, web UI `278 files / 4,816 tests`, and Node 20 backend `769 files / 19,711 tests`; each then failed because the Codecov CLI could not verify its GPG signature.

### Local validation

- `npx vitest run test/matchers/llm-rubric.test.ts test/matchers/multimodalGradingBlobs.test.ts test/matchers/multimodalGradingFormatRouting.test.ts test/evaluator/multimodalGradingExternalization.test.ts test/assertions/redteam.test.ts test/providers/bedrock/index.test.ts test/providers/bedrock/util.test.ts test/redteam/plugins/base.test.ts test/redteam/providers/crescendo/index.test.ts` — exit 0; 9 files / 565 tests passed.
- `npm run tsc` — exit 0.
- `npm run build` — exit 0, including the app production build.
- `npm run l` — exit 0; warnings only and no fixes applied.
- `SKIP_OG_GENERATION=true npm run build` from `site/` — exit 0.
- `npm run local -- validate -c examples/multimodal-output-grading/promptfooconfig.yaml` — exit 0.
- Isolated local CLI eval: `PROMPTFOO_CONFIG_DIR=/private/tmp/pr9627-promptfoo-config PROMPTFOO_DISABLE_TELEMETRY=true PROMPTFOO_DISABLE_UPDATE=true PROMPTFOO_DISABLE_REMOTE_GENERATION=true npm run local -- eval -c /private/tmp/pr9627-multimodal-e2e.yaml --no-cache -o /private/tmp/pr9627-multimodal-result.json` — exit 0; 1/1 passed. The export showed a blob-backed response, `renderedGradingPromptImages: 1`, a placeholder-only grading text, and no persisted base64.
- Direct exact-head probes reproduced malformed padding, blob-prefix text loss, pre-limit regex scanning, structured multi-image text retention, target-provider-less resolver loss, and MIME fallback mislabeling.
- `git diff --check` — exit 0; worktree clean.

Centralize image-only output classification (exact scalar references plus all-image structured forms), compute blob MIME once with a single precedence rule, and carry the resolver independently from `originalProvider`. Add focused regressions for each inline finding.

### Limitations

The requested PR Review Toolkit command was not installed as a callable skill, so I used the read-only six-lens fallback plus three independent specialist passes. I did not run paid live OpenAI/Anthropic/Bedrock/Google calls; provider payload routing was exercised with focused tests and local deterministic providers.

PR #9628, opened 2026-06-05
fix: improve image rubric grading

jameshiester-oai:
## Summary

- Ground image-related llm-rubric grading in actual image payloads and visual content instead of accompanying text or attachment presence alone.
- Strengthen multimodal grader instructions so attached images are inspected directly before deciding compliance, refusal, or policy safety.
- Add local and remote matcher coverage for image-related outputs with and without image payloads.

## Test plan

- `npm test -- test/matchers/llm-rubric.test.ts`


mldangelo-oai:
## Ownership review: NEEDS_HUMAN_DECISION

Reviewed exact head [`ada15694e4960f4c908a415d5f989c5755aa7bc3`](https://github.com/promptfoo/promptfoo/commit/ada15694e4960f4c908a415d5f989c5755aa7bc3). I recommend **not landing the generic keyword heuristic as-is**. The calibration goal is plausible, but the current design needs an explicit product decision about how Promptfoo knows an output image was expected. Preserve the stronger wording for actual image payloads; use explicit output-modality/assertion metadata for missing-image behavior instead of scanning rubric/output vocabulary.

Validated blockers:

- **Ordinary text grading is reclassified as missing-image grading.** [`shouldAddImageGradingGuidance`](https://github.com/promptfoo/promptfoo/blob/ada15694e4960f4c908a415d5f989c5755aa7bc3/src/matchers/llmGrading.ts#L173-L195) fires on generic words in the rubric or untrusted output. Full CLI probes showed guidance injected for Docker-image text, an image-input/text-output rubric, and JSON containing an `image` field, while a no-keyword control stayed unchanged.
- **Custom prompt compatibility breaks.** The documented `{{ rubric }}` value is replaced with prose plus the assertion value at [the local prompt binding](https://github.com/promptfoo/promptfoo/blob/ada15694e4960f4c908a415d5f989c5755aa7bc3/src/matchers/llmGrading.ts#L253-L267). A JSON-string rubric used with `rubricPrompt: '{{ rubric }}'` stopped being valid JSON.
- **Supported YAML chat prompts can fail before grading.** Multiline guidance substituted into an indented YAML block scalar becomes unindented after its first line; the identical probe passed at the merge base and throws `YAMLException` on this head.
- **Mixed text/image criteria receive contradictory rules.** The rubric guidance says text-refusal criteria use text, while [the appended multimodal instruction](https://github.com/promptfoo/promptfoo/blob/ada15694e4960f4c908a415d5f989c5755aa7bc3/src/matchers/rubric.ts#L40-L41) says accompanying text is not evidence of non-refusal. Local and remote prompt assembly also differs.
- **Object rubrics bypass the new protection**, confirming the existing unresolved thread: the early type guard returns them unchanged even with an actual image payload. A `vars.rubric` collision likewise removes local guidance while the remote top-level rubric remains augmented.

Decision evidence: the new tests prove prompt plumbing with preprogrammed mock outcomes, not grading improvement. PR #9617 already shipped actual image transport/direct-inspection guidance, and active PR #9627 owns overlapping hardening. A controlled treatment/control benchmark, negative lexical cases, mixed-modality cases, object/custom-prompt contracts, and cross-provider repeats are needed before choosing a replacement design.

QA on this head:

- 157 focused matcher/assertion tests passed; the main matcher file also passed five shuffled seeds (94/94 each).
- Full build/TypeScript and Biome checks passed.
- A no-cache, real CLI evaluator run with a valid PNG produced 3 passes, 5 intentional failures, and the reproduced YAML error; exported result JSON was inspected.
- A real Codex text judge correctly rejected the missing-image claim and tolerated some false triggers, so semantic harm is model-dependent rather than proven universal.
- A real Codex vision judge correctly scored a portrait positive (0.99) and blue-square negative (0), even though its rendered rubric incorrectly said no image payload existed. That robustness does not validate the instruction.

No source/GitHub state was changed beyond this requested final comment: no commits, pushes, title/body edits, thread resolutions, approvals, or CI reruns. Both existing threads remain open. Historical exact-head CI is green, but the branch is 127 commits behind current `main`, strict freshness is enabled, and review is still required.

alandelong-oai (approved):
approving with the [custom rubricPrompt compatibility issue](https://github.com/promptfoo/promptfoo/pull/9628#issuecomment-4812403737) as a nonblocking follow-up. the added multiline guidance can make a previously valid YAML prompt fail to parse. can we keep custom rubric templates working when adding the image guidance?

PR #9655, opened 2026-06-08
feat(redteam): add domain hallucination plugin for fabricated URL/repo detection

Yang-1234-web:
## Summary

Adds a new red team plugin `domain-hallucination` that detects when LLM applications fabricate external references — GitHub repositories, websites, and domains that don't exist.

The plugin generates ~20 realistic prompts asking the target system to recommend libraries / frameworks / docs, then automatically verifies every URL and `github.com/owner/repo` reference in the response via HEAD requests. Fabricated references (404s) deterministically fail the test, with no LLM-as-judge overhead.

## What's new

- **Plugin** `domain-hallucination`: generates prompts that ask for external resources; registered in `ADDITIONAL_PLUGINS`.
- **Grader**: overrides `getResult()` to do deterministic 404 detection instead of going through `llm-rubric`. Static rubric kept as a fallback.
- **Extraction utility** (`src/util/domainExtraction.ts`): pulls URLs, bare domains, and `github.com/owner/repo` references out of text. Handles Markdown links, trailing punctuation, `.git` suffix, and GitHub reserved paths (`features`, `marketplace`, etc.).
- **Verification utility** (`src/util/domainVerification.ts`): HEAD requests via the repo's `fetchWithProxy()`, 5-second timeout, `AbortController` cancellation, `Promise.allSettled` for concurrent checks.
- **Docs + example**: new plugin page at `site/docs/red-team/plugins/domain-hallucination.md` and a runnable example at `examples/domain-hallucination/`.

## Grading logic

| Outcome         | Condition                                | Score |
| --------------- | ---------------------------------------- | ----- |
| Fast pass       | Response has no URLs / domains / repos   | 1     |
| Pass            | Every reference returns 2xx / 3xx        | 1     |
| Pass (degraded) | At least one reference timed out (5s)    | 0.5   |
| **Fail**        | **Any reference returns 404**            | 0     |

## Files

**New:**

- `src/util/domainExtraction.ts` + `test/util/domainExtraction.test.ts`
- `src/util/domainVerification.ts` + `test/util/domainVerification.test.ts`
- `src/redteam/plugins/domainHallucination.ts` + `test/redteam/plugins/domainHallucination.test.ts`
- `examples/domain-hallucination/promptfooconfig.yaml` + `README.md`
- `site/docs/red-team/plugins/domain-hallucination.md`

**Modified (registration / metadata only):**

- `src/redteam/constants/plugins.ts` — adds `'domain-hallucination'` to `ADDITIONAL_PLUGINS`
- `src/redteam/plugins/index.ts` — plugin factory registration
- `src/redteam/graders.ts` — grader registration
- `src/redteam/constants/metadata.ts` — description / display name / severity / metric name / long description + entry in `riskCategories` (Brand)
- `site/docs/_shared/data/plugins.ts` — plugin index entry for the docs site
- `site/static/config-schema.json` — regenerated via `npm run jsonSchema:generate`
- `architecture/edge-baseline.json` — regenerated via `npm run architecture:baseline`

## Test plan

- [x] `npx vitest run test/util/domainExtraction.test.ts` — 16/16 pass
- [x] `npx vitest run test/util/domainVerification.test.ts` — 8/8 pass
- [x] `npx vitest run test/redteam/plugins/domainHallucination.test.ts` — 8/8 pass
- [x] `npx vitest run test/redteam/constants/metadata.test.ts` — 14/14 pass
- [x] `npx vitest run test/redteam/plugins/pluginDocumentation.test.ts test/examples/readme.test.ts` — 1728+ pass
- [x] `npm run architecture:check` — passes
- [x] `npm run lint` clean on touched files
- [x] Pre-commit hook (Biome + Prettier) passes

All HTTP calls mocked via `vi.mock('../../src/util/fetch/index', ...)`. No live network in tests.

### Why a new plugin instead of extending `hallucination`?

`hallucination` uses LLM-as-judge — it can guess that a repo name sounds fake but can't deterministically know `github.com/foo/bar` returns 404. This plugin produces hard evidence (HTTP status codes) and reports the specific fabricated URLs to reviewers.

### Why HEAD + GitHub repos API only?

Scoped intentionally for a first PR. DNS, TLS, WHOIS/RDAP, and a trusted-domain knowledge base are all viable follow-ups but would make this PR much larger to review.

### Why override `getResult()` instead of using `rubric`?

Verification gives deterministic evidence; running it through an LLM judge would be slower, more expensive, and less reliable. The static rubric is kept so the grader still has a sensible fallback if the verification path is ever bypassed.

### Why `fetchWithProxy()` and not global `fetch`?

`src/util/fetch/index.ts` is the repo's standard wrapper — picks up `HTTPS_PROXY`, retries, header sanitization. Biome's `noRestrictedGlobals` rule also forbids the bare `fetch` here.

## Future work (out of scope)

- DNS resolution / TLS / WHOIS / RDAP checks
- Verification result caching across a single eval run
- Trusted / blacklisted domain knowledge base
- Semantic relevance check (is the recommended link related to the question?)

Happy to file follow-up issues for any of these.

## Known CI failures (pre-existing, not caused by this PR)

The `Install Dependencies` step on Node 20.20 / 22.22 lanes currently fails with:

```
npm error code EBADENGINE
npm error engine Not compatible with your version of node/npm: commander@15.0.0
npm error notsup Required: {"node":">=22.12.0"}
npm error notsup Actual:   {"node":"v20.20.1","npm":"11.11.0"}
```

Root cause: `.github/workflows/main.yml` installs npm via `npx -y npm@11.11.0 ci`. That `npx` step transitively pulls the newly-published `commander@15.0.0`, which requires Node ≥22.12. This affects every PR on `main` right now, not just this one — the local `package-lock.json` in this branch is unchanged and still pins `commander@14.0.3`.

Happy to file a follow-up issue / send a separate PR to fix the workflow if maintainers prefer.

## Notes for reviewers

- **No new runtime dependencies** — uses Node's native `fetch` via `fetchWithProxy`, plus the existing `dedent` import.
- **GitHub API rate limit:** plugin uses the unauthenticated public repos endpoint (60 req/hr/IP). The example README points users at `GITHUB_TOKEN` for larger runs.
- **`(redteam)` scope:** PR follows THE REDTEAM RULE since the primary change is a new redteam plugin.
- **No CHANGELOG edit** per `AGENTS.md` (auto-generated).


Yang-1234-web:
The `CodeQL` required check is stuck on **"Expected — Waiting for status to be reported"**. There doesn't appear to be a CodeQL workflow file in `.github/workflows/`, so I suspect either the branch-protection rule references a check that no longer exists, or CodeQL is configured via the GitHub default setup which doesn't run on fork PRs.

Could a maintainer either trigger / override the check, or let me know if there's something I need to do on my fork? All other CI is green except for the pre-existing `commander@15.0.0` engine issue on the Node 20.20 / 22.22 install lanes.

PR #9661, opened 2026-06-08
fix: preserve prompt function across beforeAll extension hooks

### LINKED ISSUE #9653: Using Python extension hooks stops running Python script-based prompts.
## **Describe the bug**
I have a Python-based prompt. If I add a Python-based extension hook then the prompt isn't evaluated and the Python code is passed to the model as is.

### promptfooconfig.yaml
```yaml
description: "Extension Hook Test"

extensions:
  - file://extension.py:extension_hook

prompts:
  - id: file://prompt.py:create_prompt
    label: default

providers:
  - id: openai:chat:llama
    config:
      apiBaseUrl: http://localhost:7876

tests:
  - vars:
      topic: Linear Algebra
    assert:
      - type: not-is-refusal
```

### prompt.py
```python
def create_prompt(context: dict[str, any]) -> str|list[dict[str, str]]:
    vars = context['vars']
    prompt = []
    prompt.append({
        "role": "system",
        "content": "You are an assistant who will answer the user's questions."
    })
    prompt.append({
        "role": "user",
        "content": f"What is {vars['topic']}?"
    })
    return prompt
```

### extension.py
```python
def extension_hook(hook_name: str, config: dict) -> dict:
    return config
```

### steps to reproduce
1. Run `promptfoo eval`
2. Observe the results
3. Comment out the 

## **Expected behavior**
The Python (or JavaScript) prompt should be evaluated to generate the prompt string/messages (resulting in the "Without Extension Hook" behaviour).

### With Extension Hook
```
Thinking: A Python function `create_prompt` that takes a dictionary `context` and constructs a list of dictionaries (representing a prompt for an LLM, like OpenAI's API).
    *   `context` contains a key `vars`, which contains a `topic`.
    *   The function returns a list of dictionaries containing `role` and `content`.
...
```
### Without Extension Hook
```
Thinking: "What is Linear Algebra?"
The user is asking for a foundational overview of a major branch of mathematics. They could be a student, a programmer, a data scientist, or just someone curious about the field.
...
```
## **System information:**

- Promptfoo version: 0.121.13

## **Additional context**
n/a

serhiizghama:
## Problem

Fixes #9653. When a config combines an extension hook with a function-based prompt (e.g. a Python `file://prompt.py:create_prompt`), the prompt function stops being executed — the **raw source code of the prompt file is sent to the model as text** instead of the generated messages. Removing the hook restores correct behavior.

Reproduced with the exact config from the issue using the `echo` provider (echoes the rendered prompt):

**Without this fix** — the raw Python source is sent:
```
"def create_prompt(context: dict) -> list:\n    topic = context['vars']['topic']\n    return [ ... ]\n"
```

**With this fix** — the function runs and the messages are sent:
```json
[{"role":"system","content":"You are an assistant who will answer the user's questions."},
 {"role":"user","content":"What is Linear Algebra?"}]
```

## Root cause

Extension hooks receive the test suite serialized as JSON — Python (and JS) hooks run in a subprocess, so the context is passed through `JSON.stringify`. JSON cannot represent the non-serializable `function` callable that `file://...:fn` prompts carry (`Prompt.function`), so it is dropped on the round-trip.

In `runExtensionHook`'s `beforeAll` branch, the returned (deserialized) prompts overwrite the in-memory prompts, so every prompt loses its `function`. At eval time `renderPrompt` checks `if (prompt.function)` — now false — and falls back to using `prompt.raw` (the file source) as the prompt text.

## Solution

After a `beforeAll` hook returns, re-attach the `function` to each returned prompt from the original in-memory prompts. A returned prompt is matched to an original by `label` (the stable identifier for file-based prompts), and the function is restored only when:

- the original carried a `function`, and
- the `raw` source is unchanged — so a hook that intentionally rewrites a prompt keeps its new behavior.

Labels that appear more than once are treated as ambiguous and skipped, to avoid restoring the wrong function. Hooks that don't return prompts are unaffected.

## Testing

- `npx vitest run test/evaluatorHelpers.test.ts` — added two regression tests: one asserting the function is restored after a serialized round-trip, one asserting it is **not** restored when the hook rewrites `raw`.
- Verified end-to-end with the issue's config (Python prompt + Python extension hook) against the `echo` provider, confirming the executed messages are sent instead of the raw source.


mldangelo (approved):
LGTM — thank you @serhiizghama for an excellent fix and a model bug report to go with it! 🙏

The root-cause analysis was spot on, and the conservative restore semantics (label match + unchanged `raw`, ambiguous duplicates skipped) are exactly the right shape: they undo the serialization loss without taking away a hook's documented ability to rewrite prompts.

I verified this end-to-end with the exact reproducer from #9653 (Python prompt function + Python extension hook + echo provider): on `main` the raw Python source is sent to the model, and with this fix the executed messages are sent. I also confirmed the fix holds for the new `beforeAll` calling convention, chained extensions, configs without an explicit `label`, JS in-process hooks (unaffected, as expected), and that intentional prompt rewrites by hooks are still honored.

I pushed two small follow-up commits to the branch:

- **Diagnostics for the silent paths** — when a function can't be restored (duplicate labels, or a hook relabeling a function prompt), we now log a warning explaining why the prompt will render as raw text, plus debug logs for intentional rewrites/removals. Also corrected the JSDoc (JS hooks run in-process, not in a subprocess).
- **Tests pinning the subtle behaviors** — the ambiguous duplicate-label skip, hook-attached functions winning over restoration, restoration surviving chained extensions, malformed hook output passing through safely, and a true end-to-end smoke test that drives the real Python subprocess JSON round-trip (`test/smoke/extension-hooks.test.ts`), so the actual regression vector from #9653 is covered, not just the mocked shape.

One note for future work (pre-existing, not introduced here): the same JSON round-trip still strips other non-serializable suite state when a Python hook returns its context — `nunjucksFilters` (functions), live `test.provider` objects, and the `beforeEach` branch entirely. Worth extending this restoration pattern there in a follow-up.

Merging — thanks again for the contribution!

PR #9668, opened 2026-06-09
fix(deps): constrain undici to <7.27.1 to fix Node 26 "terminated" error

### LINKED ISSUE #9667: Node 26: all API calls fail with `TypeError: terminated` (decompress interceptor)
**Describe the bug**

On Node.js v26, **every** provider API call fails with `API call error: TypeError: terminated`, making `promptfoo eval` completely unusable. 0 results ever succeed.

Node 26 is within promptfoo's declared supported range (`"engines": { "node": "^20.20.0 || >=22.22.0" }`).

The failure appears to come from the experimental undici `decompress` interceptor. promptfoo attaches it to its custom undici `Agent`/`ProxyAgent` to cope with Node 26 no longer auto-decompressing responses on custom dispatchers:

- `src/util/fetch/index.ts:87` and `:111` — `.compose(interceptors.decompress({ skipErrorResponses: false }))`
- `src/providers/http.ts:1866` — same, with the comment *"Node 26 undici no longer auto-decompresses…"*

Running an eval prints `(node:…) ExperimentalWarning: DecompressInterceptor is experimental and subject to change` immediately before the `terminated` errors, which points at that interceptor as the failing step.

**To Reproduce**

Minimal config (`promptfooconfig.yaml`):
```yaml
providers:
  - openai:gpt-4o-mini
prompts:
  - "Say the single word: ok"
tests:
  - assert:
      - type: contains
        value: "ok"
```

1. Use Node `v26.0.0`, set `OPENAI_API_KEY`.
2. Run `npx promptfoo@latest eval --no-cache`.
3. See:
```
(node:7811) ExperimentalWarning: DecompressInterceptor is experimental and subject to change
API call error: TypeError: terminated
┌──────────────────────────────────────────────┐
│ [openai:gpt-4o-mini] Say the single word: ok │
├──────────────────────────────────────────────┤
│ [ERROR] API call error: TypeError: terminated│
└──────────────────────────────────────────────┘
  ✗ 1 error (100%)
```

**Control test (isolates the bug to promptfoo, not the key/network/OpenAI):**
A plain Node 26 `fetch` to the same endpoint with the same key succeeds:
```js
const res = await fetch("https://api.openai.com/v1/chat/completions", {
  method: "POST",
  headers: { "Content-Type": "application/json", Authorization: "Bearer " + process.env.OPENAI_API_KEY },
  body: JSON.stringify({ model: "gpt-4o-mini", messages: [{ role: "user", content: "say ok" }], max_tokens: 5 }),
});
console.log(res.status); // → 200
```
→ `HTTP status: 200`. Same machine, same key, same endpoint. Only promptfoo (custom Agent + decompress interceptor) fails.

**Expected behavior**

`promptfoo eval` runs the request and returns results, as it does on Node 24. Responses should be decompressed correctly on Node 26.

**Screenshots**

N/A — full console output included above.

**System information:**

- Promptfoo version: `0.121.15`
- Node: `v26.0.0`
- npm: `11.12.1`
- OS: macOS 15.6.1 (24G90)

**Additional context**

The repo's `.nvmrc` pins `24.16.0`, so Node 26 is likely newer than the version CI/devs run on — the decompress workaround added for Node 26 doesn't actually work on Node 26 in practice. Likely fix directions: swap the experimental `interceptors.decompress` for a stable approach, bump `undici`, or gate it by Node version. Happy to help test a fix.

drooph0904:
Closes #9667

### Summary

On Node.js 26, provider calls can fail with `API call error: TypeError: terminated` and an underlying decompression error. Node 26 is within promptfoo's supported `engines` range (`^20.20.0 || >=22.22.0`).

### Root cause

Promptfoo composes undici's experimental `decompress` interceptor onto its custom agents so compressed responses work when Node 26 uses a custom dispatcher.

Undici 7.27.1 intentionally began preserving `rawHeaders` in the v7-to-v8 dispatch-handler bridge ([undici #5345](https://github.com/nodejs/undici/pull/5345)). With promptfoo's imported undici 7 agent running under Node 26's built-in undici 8 `fetch`, promptfoo's interceptor decodes the response body first, while the preserved `Content-Encoding` header causes the outer fetch layer to decode it again. The second decode fails for gzip and Brotli and is surfaced as `TypeError: terminated`.

This is a cross-version double-decompression incompatibility, not a standalone regression in undici's decompressor. Undici 7.27.1 and 7.27.2 expose the incompatibility; versions through 7.27.0 work with promptfoo's current interceptor composition.

### Fix

- Constrain undici to `>=7.25.0 <7.27.1`, preserving Node 20 support and promptfoo's existing response-header behavior.
- Extend the package-artifact test to install the packed tarball without a lockfile, run the installed CLI against local gzip and Brotli HTTP providers, and assert both decoded results.
- Run that fresh-consumer artifact test on Node 26 as well as the minimum supported Node 20 runtime.

The dependency cap is intentionally a stopgap. A later migration can remove the extra interceptor and move to a fixed undici version after explicitly handling the resulting `Content-Encoding` and `Content-Length` behavior change across direct, proxy, and TLS dispatchers.

### Why existing CI missed it

The Node 26 test lane and the smoke suite both use `npm ci`, so they exercised the lockfile's undici 7.25.0. The smoke suite also runs on Node 24. The existing compression unit tests are effective for the installed lockfile but did not cover the failing combination: a lock-free consumer install, Node 26's built-in undici 8 fetch, and a compressed response consumed through the packed promptfoo CLI.

### Verification

- Node 26.1.0 package artifact test: passes fresh install, gzip eval, and Brotli eval.
- Node 20.20.1 package artifact test: passes the same fresh-consumer workflow.
- Node 26.1.0 compression tests: 16/16 pass.
- Negative control with the previous `^7.25.0` range: the new artifact gate fails with two `TypeError: terminated` errors, one for gzip and one for Brotli.
- `npm run build`, `npm run tsc`, `npm run lint:ci`, changed-file lint/format, Prettier, and actionlint pass.

Isolated runtime matrix:

| undici | Node 26 result |
| --- | --- |
| 7.25.0 | works |
| 7.27.0 | works |
| 7.27.1 | `Decompression failed` |
| 7.27.2 | `Decompression failed` |


mldangelo:
Thank you!

PR #9672, opened 2026-06-09
feat(providers): add Moonshot (Kimi) provider

NishchayMahor:
## What

Adds a dedicated `moonshot:` provider for [Moonshot AI](https://platform.kimi.ai/) (Kimi). Moonshot exposes an OpenAI-compatible API, so the provider extends `OpenAiChatCompletionProvider` and points it at `https://api.moonshot.ai/v1` with the `MOONSHOT_API_KEY` envar.

Today these models can only be reached via the generic `openai:` provider with a manual `apiBaseUrl`/`apiKeyEnvar`. A first-class prefix makes the Kimi K2 family discoverable and removes that boilerplate.

## Usage

```yaml
providers:
  - id: moonshot:kimi-k2.6 # flagship thinking model — leave sampling params unset
  - id: moonshot:moonshot-v1-8k # generation model
    config:
      temperature: 0.2
      max_tokens: 1024
```

Both `moonshot:<model>` and `moonshot:chat:<model>` route to chat completions, and a bare `moonshot:` defaults to `kimi-k2.6`. Covers the current Kimi K2 thinking models (`kimi-k2.6`, `kimi-k2.5`, `kimi-k2.7-code`, …) and the `moonshot-v1-*` generation/vision families.

## Implementation notes

The provider mirrors the modern `minimax` provider rather than the older deepseek/together copy, because the Kimi models have real OpenAI-compat quirks (verified against the live API):

- **Sampling constraints.** `kimi-k2.x` models pin `temperature` (1.0 with thinking on), `top_p`, `n`, and the penalties, and return `400` ("invalid temperature: only 1 is allowed for this model") for any other value. `getOpenAiBody` therefore strips promptfoo's injected `temperature: 0` / `max_tokens: 1024` defaults (and any `OPENAI_*` sampling-env leakage) for `kimi-*` models, so they work with no config. `moonshot-v1` models keep promptfoo's deterministic defaults.
- **Key handling.** `getApiKey` resolves only the Moonshot key/envar (no silent `OPENAI_API_KEY` fallback) and the OpenAI org header is suppressed.
- **Region override.** A user-set `apiBaseUrl` is honored so China-mainland keys can target `https://api.moonshot.cn/v1`.
- **Cost.** No hardcoded price table (Kimi pricing changes upstream), but `cost`/`inputCost`/`outputCost`/`cacheReadCost` now actually take effect — the provider computes cost from them (the inherited OpenAI cost path returns `undefined` for non-OpenAI models).
- **Reasoning.** Kimi `reasoning_content` surfaces via promptfoo's `Thinking:` prefix; set `showThinking: false` for clean structured output.

## Checklist

- [x] Provider implementation (`src/providers/moonshot.ts`) + registry entry
- [x] `MOONSHOT_API_KEY` in `src/contracts/env.ts`, `src/envars.ts`, `src/util/provider.ts`
- [x] Tests (`test/providers/moonshot.test.ts`) + routing test in `test/providers/registry.test.ts`
- [x] Docs page + `index.md` table row
- [x] Example (`examples/provider-moonshot/`)
- [x] Regenerated `config-schema.json` + architecture baseline

## Testing

Unit: `npx vitest run test/providers/moonshot.test.ts test/providers/registry.test.ts` → 104 passing; `tsc` clean; Biome/architecture/jsonSchema clean.

End-to-end against the live `api.moonshot.ai` API (real calls, `--no-cache`) — all green:

| Feature | `moonshot-v1-8k` | `kimi-k2.6` |
| --- | --- | --- |
| Basic chat | ✅ | ✅ |
| `moonshot:chat:` routing | ✅ | — |
| No-config defaults | ✅ | ✅ (temperature no longer rejected) |
| Streaming | ✅ | — |
| Explicit temperature | ✅ (0) | ✅ (1) |
| Tool / function calling | ✅ | ✅ |
| JSON mode (`response_format`) | ✅ | ✅ (`showThinking: false`) |
| Vision (base64 image) | ✅ (`-vision-preview`) | ✅ |
| Thinking toggle / `showThinking` | — | ✅ |
| Cost from `cost`/`inputCost`/`outputCost` | ✅ | ✅ |

Negative check: `kimi-k2.6` with an explicit `temperature: 0.5` returns the documented `400` — which is exactly what the default-stripping avoids for unconfigured runs.

> Note: the previously-referenced `kimi-k2-0711-preview` / `kimi-k2-thinking` ids were discontinued by Moonshot in 2026, so the docs/example now use `kimi-k2.6`.


mldangelo-oai (approved):
Thank you!

NishchayMahor:
@mldangelo @mldangelo-oai thanks for merging this one 🙏

Now that the provider-eligibility bar calls out "value beyond a base URL + API key," I want to make sure I build the right thing before opening another PR. Two OpenAI-compatible hosts come up a lot in evals — **SambaNova** and **DeepInfra** — but on their own they'd only need `openai` + `apiBaseUrl`, so my read is they belong on the generic path with a docs/example rather than dedicated prefixes. Does that match how you'd want them handled?

And if there's a provider you'd actually like to see added that clears the bar (distinct auth / routing / response handling), I'm happy to take that on. Just want to point the effort where it's useful.

PR #9683, opened 2026-06-10
fix(deps): gate Content-Encoding strip after decompress and require undici >=7.27.1 <8

drooph0904:
Follow-up to #9668, which capped undici to `>=7.25.0 <7.27.1` as a stopgap for the Node 26 `TypeError: terminated` error (#9667). This PR fixes the underlying double-decompression in promptfoo's own dispatcher and moves the dependency to `>=7.27.1 <8`.

### Root cause

Promptfoo composes undici's experimental `decompress` interceptor onto its custom agents so compressed responses still decode when Node 26 uses a custom dispatcher.

Undici 7.27.1 began preserving `rawHeaders` in the v7→v8 dispatch-handler bridge ([undici#5345](https://github.com/nodejs/undici/pull/5345), fixing [undici#5341](https://github.com/nodejs/undici/issues/5341)). With promptfoo's imported undici 7 agent running under Node 26's built-in undici 8 `fetch`, promptfoo's interceptor decodes the response body first, while the preserved `Content-Encoding` raw header causes the outer fetch layer to decode it again. The second decode fails for gzip and Brotli and surfaces as `TypeError: terminated`.

### Fix

- **`stripDecompressionHeaders` interceptor**, composed after `decompress` on all three dispatchers (pooled direct, pooled proxy, per-request TLS in the HTTP provider). It removes `Content-Encoding`/`Content-Length` from `controller.rawHeaders` **only when the decompress interceptor actually decoded the body**, detected by: the raw headers carry `content-encoding` while the parsed headers — already rewritten by `decompress` on its decode path, and forwarded untouched on every skip path — do not. Without the gate, every response would lose `Content-Length` (plain bodies, HEAD probes, the documented `context.response.headers` transform surface), and responses with encodings the interceptor cannot decode would arrive still-compressed with the `Content-Encoding` evidence destroyed. undici 8's own decompress interceptor applies the same conditional rawHeaders filtering upstream.
- The interceptor is a **delegating handler class, not object spread** (spreading a class instance snapshots own-property values and breaks state the inner handler sets later), and forwards `onResponseStarted` so it stays a faithful decorator.
- **Dependency range `>=7.27.1 <8`**, lockfile at 7.27.2:
  - **Floor `7.27.1`**: on undici 7.25.0–7.27.0 the v7 bridge controller has no `rawHeaders` at all, so Node 26's fetch builds every `Response` from an empty raw-header list — **all** response headers (content-type, retry-after, x-ratelimit-\*, session headers) silently disappear (undici#5341). The previous cap pinned Node 26 users into that band, and the strip interceptor would be inert there.
  - **Cap `<8`**: undici 8 dispatchers reject the legacy handlers Node ≤25's built-in fetch passes (`InvalidArgumentError: invalid onRequestStart method`), and with `decompress` composed the rejection surfaces as a silent hang on every request (verified on Node 24.16 + undici 8.4.1, where the previous `>=7.25.0` range resolved for fresh installs). The cap can be lifted once the minimum supported Node is 26+, where undici 8's decompress already filters rawHeaders itself.

### Verification

- Reproduced the original bug: Node 26.2 + undici 7.27.2 + the pre-fix dispatcher fails the compression suite with `TypeError: terminated` (`Z_DATA_ERROR`); with the fix all tests pass.
- New unit suite (`test/util/fetch/stripDecompressionHeaders.test.ts`) drives the interceptor with a synthetic bridge controller so the strip logic is exercised on every Node version: strips when decoded, passes through on unsupported encodings / plain responses / missing or null rawHeaders / non-lowercase parsed keys, preserves dangling raw entries, propagates the dispatch backpressure boolean, forwards all callbacks.
- New e2e regressions in `test/fetch.compression.test.ts`: uncompressed responses keep `Content-Length`, uncompressed HEAD responses keep `Content-Length`, and an unsupported `Content-Encoding: x-snappy` passes through with the header intact. All three fail against the previous unconditional strip on Node 26.
- 565 tests (fetch, compression, monkeyPatchFetch, providers/http, http-tls) pass on Node 20.20, 22.22, 24.16, and 26.2 with undici 7.27.2; `tsc --noEmit` clean; `architecture:check` passes; changed files pass Biome.
- Live CLI check on Node 24 and Node 26: `promptfoo eval` against a local HTTP target — a gzip endpoint decodes correctly, and a plain endpoint's `content-length` remains visible through `context.response.headers` in `transformResponse` (it was lost with the unconditional strip, and lost entirely on undici 7.25.0 under Node 26).


mldangelo:
Maintainer update: I reviewed and live-tested this PR (Node 20.20/22.22/24.16/26.2 × undici 7.25.0/7.27.2/8.4.1), confirmed the root-cause analysis, and pushed two commits addressing everything raised here plus issues found in deeper QA:

**Codex feedback — all three confirmed and addressed:**
1. *Only strip headers after confirmed decompression* — confirmed as a real regression: the unconditional strip made every response lose `Content-Length` on Node 26 (plain bodies, HEAD probes, `context.response.headers` in transforms), and unsupported encodings arrived still-compressed with the `Content-Encoding` evidence destroyed. The strip is now gated on the exact signal the decompress interceptor provides: it rewrites the parsed headers only on its decode path, so "rawHeaders carry `content-encoding` but parsed headers don't" identifies a decoded body precisely. This matches what undici 8's own decompress interceptor does upstream.
2. *Update the locked undici resolution* — confirmed and worse than it looked: the lockfile still installed 7.25.0, where the bridge controller has no `rawHeaders` at all, so the fix was inert **and** Node 26 dropped *all* response headers (undici#5341). Lockfile now resolves 7.27.2.
3. *Constrain to the supported major* — confirmed and critical: with `>=7.25.0`, a fresh install resolved undici 8.4.1, whose dispatchers are rejected by Node ≤25's built-in fetch (`invalid onRequestStart method`) — and with `decompress` composed, every request **hangs silently** (reproduced on Node 24.16). Range is now `>=7.27.1 <8`; the floor is needed because 7.25.0–7.27.0 lose all response headers on Node 26.

**Additional changes:**
- Forward `onResponseStarted` so the wrapper is a faithful decorator.
- New unit suite drives the interceptor with a synthetic bridge controller (active on every Node version, no Node 26 dependence in CI), plus e2e regressions for the uncompressed/unknown-encoding cases — all verified to fail against the previous unconditional strip.
- Verified end-to-end with `promptfoo eval` against a local target on Node 24 and 26: gzip decodes, and response header visibility is preserved through `transformResponse`.

PR title/body updated to match the final implementation.

mldangelo (approved):
LGTM — thank you @drooph0904 and @dgocta for this contribution! The root-cause analysis of the undici 7.27.1 rawHeaders bridge change was exactly right, the delegating-handler-class design was the correct call (the object-spread pitfall you documented is real and subtle), and following up #9668's stopgap cap with a proper fix is much appreciated.

I verified the diagnosis end-to-end (reproduced the `TypeError: terminated` on Node 26 + undici 7.27.2 pre-fix, confirmed the fix resolves it) and pushed two follow-up commits: gating the strip on actual decompression so plain responses and unsupported encodings keep their headers, moving the range to `>=7.27.1 <8` with the lockfile at 7.27.2 (the floor avoids the all-headers-lost band on Node 26, the cap avoids undici 8 hanging on Node ≤25), and regression tests for all of the above. All 31 checks are green across Node 20/22/24/26. Merging — thanks again!

drooph0904:
Thank you @mldangelo — this was genuinely a great learning experience. I really appreciate you building on the PR instead of simply rewriting it, and for taking the time to explain the reasoning behind both follow-ups.

Two things I'm definitely taking away for next time:

* Gate on actual behavior, not assumed behavior. I only reasoned about the compressed-and-decoded path and completely missed that `decompress` skips plain bodies and unsupported encodings. That meant the unconditional strip silently broke `Content-Length` (including HEAD probes) and removed valuable `Content-Encoding` evidence. Using "rawHeaders has it, parsed headers don't" as the decode signal is a really elegant way to understand `decompress`'s intent, and it's great to know that it mirrors the approach used in undici 8 upstream.

* The lockfile is the code that actually runs. Pinning `7.25.0` for a "minimal diff" effectively made the entire fix inert, and that lesson is definitely going to stick with me. The `>=7.27.1 <8` reasoning — where the lower bound avoids the all-headers-lost band and the upper bound avoids the Node ≤25 hang — is a perfect example of a version range doing meaningful work at both ends.

The synthetic-bridge unit suite is probably the pattern I'm most excited to reuse. Testing the boundary directly so it runs consistently across every Node version is a really powerful approach.

Thanks again, and I'm looking forward to the next one!

PR #9680, opened 2026-06-10
fix(providers): preserve Anthropic thinking token usage

mldangelo:
## Summary

- map Anthropic `usage.output_tokens_details.thinking_tokens` to Promptfoo's `completionDetails.reasoning`, preserving total, prompt, completion, cache-token, and cached-response accounting
- sum `thinking_tokens` across MCP tool-use rounds in `mergeAnthropicUsage` so multi-turn reasoning counts aggregate the same way `output_tokens` already does, instead of reporting only the last round
- fix the `completionDetails` comment subject (Promptfoo's `TokenUsage` contract lacks an input-details field — not Anthropic) and collapse the conditional spreads to the prevailing `...(cond && {})` idiom

Supersedes #9660 — same change plus review fixes, moved into the base repo (the fork branch was not pushable by maintainers).

## Why

`@anthropic-ai/sdk` 0.100.x added `usage.output_tokens_details`, but the Anthropic provider dropped its `thinking_tokens` field during response normalization. Extended-thinking evaluations therefore reported correct billed output totals while losing the reasoning-token breakdown in result JSON, summaries, and tracing. In multi-turn MCP flows the merged usage additionally kept only the final round's thinking tokens while summing `output_tokens`, making the breakdown inconsistent.

## Test coverage

- thinking-only response → `completionDetails.reasoning`
- thinking + prompt-cache details merged into one `completionDetails` object (the shape a future refactor is most likely to regress)
- `thinking_tokens: 0` preserved as `reasoning: 0` (zero ≠ unknown)
- `output_tokens_details: null` (live-API wire format for non-thinking responses) → no `completionDetails`
- cached responses keep returning `{cached, total}` only
- MCP multi-round flow sums reasoning across rounds (3 + 2 → 5)

## Validation

- `npx vitest run test/providers/anthropic` — 299 tests pass
- `npm run tsc` — clean
- `npm run architecture:check` — passed
- Biome clean on all four changed files
- live end-to-end eval (`anthropic:messages:claude-haiku-4-5-20251001` with `thinking: enabled`, `--no-cache`): exported JSON shows `completionDetails.reasoning: 137` on the result and the same value aggregated into prompt-level metrics

PR #9685, opened 2026-06-10
fix(providers): type anthropic thinking usage

wholley-oai:
## Summary

Promptfoo recently started preserving Anthropic `output_tokens_details.thinking_tokens` so MCP continuation rounds can report reasoning-token usage instead of dropping it. The runtime behavior was correct, but the installed Anthropic SDK type does not yet expose `usage.output_tokens_details`, so current `origin/main` failed `npm run tsc` after PR #9680 merged.

This change keeps the runtime aggregation behavior and adds a narrow local structural type for the newer usage field. It also updates the Anthropic test fixtures to cast mocked partial SDK messages through `unknown`, matching TypeScript's guidance for intentionally incomplete API fixtures.

## Verification

- `node_modules/.bin/vitest run test/providers/anthropic/messages.test.ts test/providers/anthropic/util.test.ts`
- `node_modules/.bin/vitest run test/evaluatorHelpers.test.ts test/commands/validate-falsy-output.test.ts test/fetch.compression.test.ts test/util/fetch/stripDecompressionHeaders.test.ts test/blobs/index.test.ts test/blobs/shareUpload.test.ts test/share.test.ts test/util/inlineBlobsForShare.test.ts test/providers/anthropic/messages.test.ts test/providers/anthropic/util.test.ts test/providers/bedrock/anthropicMessages.test.ts test/providers/bedrock/converse.test.ts test/providers/bedrock/index.test.ts test/providers/families/aws.test.ts test/providers/google/vertex.test.ts test/redteam/index.test.ts test/server/routes/redteam.test.ts test/agentSkills/promptfooPlugin.test.ts test/util/eval/evalTableUtils.test.ts test/csv.test.ts test/assertions/equals.test.ts test/commands/validate-provider-tests.test.ts test/redteam/commands/discover.test.ts test/remoteGrading.test.ts`
- `npm run tsc`
- `npm run f`
- `npm run l`
- `git diff --check`

### DIFF
--- a/src/providers/anthropic/messages.ts
+++ b/src/providers/anthropic/messages.ts
@@ -214,7 +214,13 @@ function coerceMcpToolInput(input: unknown): Record<string, unknown> {
 function mergeAnthropicUsage(
   messages: Anthropic.Messages.Message[],
 ): Anthropic.Messages.Message['usage'] | undefined {
-  const usageEntries = messages.map((message) => message.usage).filter((usage) => usage != null);
+  type AnthropicUsageWithThinkingDetails = NonNullable<Anthropic.Messages.Message['usage']> & {
+    output_tokens_details?: { thinking_tokens?: number | null } | null;
+  };
+
+  const usageEntries = messages
+    .map((message) => message.usage)
+    .filter((usage): usage is AnthropicUsageWithThinkingDetails => usage != null);
 
   if (usageEntries.length === 0) {
     return undefined;
@@ -223,23 +229,27 @@ function mergeAnthropicUsage(
   const [firstUsage, ...remainingUsageEntries] = usageEntries;
 
   return remainingUsageEntries.reduce<NonNullable<Anthropic.Messages.Message['usage']>>(
-    (acc, usage) => ({
-      ...usage,
-      input_tokens: acc.input_tokens + (usage.input_tokens ?? 0),
-      output_tokens: acc.output_tokens + (usage.output_tokens ?? 0),
-      cache_creation_input_tokens:
-        (acc.cache_creation_input_tokens ?? 0) + (usage.cache_creation_input_tokens ?? 0),
-      cache_read_input_tokens:
-        (acc.cache_read_input_tokens ?? 0) + (usage.cache_read_input_tokens ?? 0),
-      output_tokens_details:
-        acc.output_tokens_details || usage.output_tokens_details
-          ? {
-              thinking_tokens:
-                (acc.output_tokens_details?.thinking_tokens ?? 0) +
-                (usage.output_tokens_details?.thinking_tokens ?? 0),
-            }
-          : null,
-    }),
+    (acc, usage) => {
+      const accWithDetails = acc as AnthropicUsageWithThinkingDetails;
+      const thinkingTokens =
+        accWithDetails.output_tokens_details || usage.output_tokens_details
+          ? (accWithDetails.output_tokens_details?.thinking_tokens ?? 0) +
+            (usage.output_tokens_details?.thinking_tokens ?? 0)
+          : undefined;
+
+      return {
+        ...usage,
+        input_tokens: acc.input_tokens + (usage.input_tokens ?? 0),
+        output_tokens: acc.output_tokens + (usage.output_tokens ?? 0),
+        cache_creation_input_tokens:
+          (acc.cache_creation_input_tokens ?? 0) + (usage.cache_creation_input_tokens ?? 0),
+        cache_read_input_tokens:
+          (acc.cache_read_input_tokens ?? 0) + (usage.cache_read_input_tokens ?? 0),
+        ...(thinkingTokens === undefined
+          ? {}
+          : { output_tokens_details: { thinking_tokens: thinkingTokens } }),
+      } as AnthropicUsageWithThinkingDetails;
+    },
     firstUsage,
   );
 }
--- a/test/providers/anthropic/messages.test.ts
+++ b/test/providers/anthropic/messages.test.ts
@@ -1486,7 +1486,7 @@ describe('AnthropicMessagesProvider', () => {
             output_tokens_details: { thinking_tokens: 3 },
             server_tool_use: null,
           },
-        } as Anthropic.Messages.Message)
+        } as unknown as Anthropic.Messages.Message)
         .mockResolvedValueOnce({
           content: [{ type: 'text', text: 'Acme Solar and Gridwise match your query.' }],
           stop_reason: 'end_turn',
@@ -1496,7 +1496,7 @@ describe('AnthropicMessagesProvider', () => {
             output_tokens_details: { thinking_tokens: 2 },
             server_tool_use: null,
           },
-        } as Anthropic.Messages.Message);
+        } as unknown as Anthropic.Messages.Message);
 
       const result = await provider.callApi('Find clean energy companies');
 


src/providers/anthropic/messages.ts
```diff
@@ -214,7 +214,13 @@ function coerceMcpToolInput(input: unknown): Record<string, unknown> {
...
   messages: Anthropic.Messages.Message[],
 ): Anthropic.Messages.Message['usage'] | undefined {
-  const usageEntries = messages.map((message) => message.usage).filter((usage) => usage != null);
+  type AnthropicUsageWithThinkingDetails = NonNullable<Anthropic.Messages.Message['usage']> & {
+    output_tokens_details?: { thinking_tokens?: number | null } | null;
+  };
+
+  const usageEntries = messages
+    .map((message) => message.usage)
+    .filter((usage): usage is AnthropicUsageWithThinkingDetails => usage != null);
```
mldangelo, line 223: **This local type and the predicate filter aren't needed under the locked SDK — and they're weaker than the real types.**

The pinned `@anthropic-ai/sdk@0.100.1` already declares on `Usage`:

```ts
output_tokens_details: OutputTokensDetails | null;  // OutputTokensDetails = { thinking_tokens: number }
```

so `NonNullable<Anthropic.Messages.Message['usage']>` already has this field, fully typed, with `thinking_tokens` as a *required* `number`. This local type loosens it to `{ thinking_tokens?: number | null } | null | undefined`.

The filter's type predicate `(usage): usage is AnthropicUsageWithThinkingDetails => usage != null` is a disguised cast — it asserts a structural type while only checking non-nullness. Main's plain `.filter((usage) => usage != null)` already narrows `Usage | null | undefined` → `Usage` correctly and compiles.

I could not reproduce the motivating `tsc` failure on `origin/main` with a lockfile-clean install (`npm ci && npm run tsc` is green; main's CI for #9680's commit was green too). It does reproduce with SDK ≤ 0.98.1, which has zero `output_tokens_details` declarations — which points to a stale `node_modules` in the environment that generated this PR.

src/providers/anthropic/messages.ts
```diff
@@ -223,23 +229,27 @@ function mergeAnthropicUsage(
...
+        input_tokens: acc.input_tokens + (usage.input_tokens ?? 0),
+        output_tokens: acc.output_tokens + (usage.output_tokens ?? 0),
+        cache_creation_input_tokens:
+          (acc.cache_creation_input_tokens ?? 0) + (usage.cache_creation_input_tokens ?? 0),
+        cache_read_input_tokens:
+          (acc.cache_read_input_tokens ?? 0) + (usage.cache_read_input_tokens ?? 0),
+        ...(thinkingTokens === undefined
+          ? {}
+          : { output_tokens_details: { thinking_tokens: thinkingTokens } }),
+      } as AnthropicUsageWithThinkingDetails;
```
mldangelo, line 251: **The two `as` casts here (line 233 `acc as …` and the return-position cast) disable the structural checking main gets for free.**

On main, the returned object literal is validated against `NonNullable<Anthropic.Messages.Message['usage']>` via the `reduce` type parameter — a missing required field or a mistyped one is a compile error. With the return cast, that check degrades to assertion-compatibility: e.g. `{ input_tokens: 1 } as AnthropicUsageWithThinkingDetails` compiles, so a future regression (dropping a summed field, typo'ing a key) would no longer be caught.

Verified: this exact function body compiles under the repo's strict tsconfig with the local type, the predicate, and both casts deleted — none of the machinery is load-bearing.

src/providers/anthropic/messages.ts
```diff
@@ -223,23 +229,27 @@ function mergeAnthropicUsage(
...
+        ...usage,
+        input_tokens: acc.input_tokens + (usage.input_tokens ?? 0),
+        output_tokens: acc.output_tokens + (usage.output_tokens ?? 0),
+        cache_creation_input_tokens:
+          (acc.cache_creation_input_tokens ?? 0) + (usage.cache_creation_input_tokens ?? 0),
+        cache_read_input_tokens:
+          (acc.cache_read_input_tokens ?? 0) + (usage.cache_read_input_tokens ?? 0),
+        ...(thinkingTokens === undefined
+          ? {}
+          : { output_tokens_details: { thinking_tokens: thinkingTokens } }),
```
mldangelo, line 250: **Undocumented behavior change vs main (the PR description says runtime behavior is kept).**

Main always materializes `output_tokens_details` on the merged usage — the summed object, or an explicit `null` when no round has details (which matches the SDK shape, where the field is always present). This conditional spread instead omits the key whenever `thinkingTokens === undefined` and the last round's raw `usage` happened to lack it.

It's consumer-invisible today — `src/providers/anthropic/util.ts:555` reads `output_tokens_details?.thinking_tokens`, and a live A/B eval of main vs this branch (MCP multi-round + thinking) produced identical `completionDetails` — but it's exactly the kind of silent shape change that shouldn't ride along in a fix billed as type-only, and no test pins it. If this PR moves forward in any form, I'd keep main's explicit `object-or-null` form.

test/providers/anthropic/messages.test.ts
```diff
@@ -1486,7 +1486,7 @@ describe('AnthropicMessagesProvider', () => {
             output_tokens_details: { thinking_tokens: 3 },
             server_tool_use: null,
           },
-        } as Anthropic.Messages.Message)
+        } as unknown as Anthropic.Messages.Message)
```
mldangelo, line 1489: **`as unknown as` turns off all structural checking of this fixture — and it isn't needed.** (Same at line 1499.)

With the double cast, a typo like `output_token_details` or `thinking_token` compiles silently and the test would assert against an absent field. The single `as Anthropic.Messages.Message` cast compiles fine under the locked SDK 0.100.1 — `origin/main` has exactly that at these lines and `npm run tsc` (which covers `test/`) is green there. The widening is only necessary against SDK ≤ 0.98.x, where `Usage` lacked `output_tokens_details` and the extra property failed the assertion-overlap check — further evidence the originating environment had a stale SDK.

If a fixture ever genuinely needs loosening, this file already has the typed-fixture pattern (`AnthropicTestMessage` / `UsageFixture` near the top of the file) that preserves checking without casts.

mldangelo (changes requested):
## Summary

Thanks for the careful writeup — but the premise of this PR does not reproduce, and I believe it should be **closed** rather than merged. The claimed `tsc` failure on `origin/main` only happens with a stale `node_modules` (SDK ≤ 0.98.x); with a lockfile-clean install the locked SDK already types everything this PR adds local types for, and the diff is a net type-safety regression with no runtime benefit.

### The premise doesn't reproduce

The PR states: *"the installed Anthropic SDK type does not yet expose `usage.output_tokens_details`, so current `origin/main` failed `npm run tsc` after PR #9680 merged."* Verified the opposite on a clean checkout:

- `@anthropic-ai/sdk` is exact-pinned at **0.100.1** (`package.json:299`, `package-lock.json`). Its `Usage` interface already declares `output_tokens_details: OutputTokensDetails | null` with `thinking_tokens: number` (`node_modules/@anthropic-ai/sdk/resources/messages/messages.d.ts`, `Usage` at ~1525).
- `npm run tsc` **passes on current `origin/main`** (and on this branch). `tsconfig.json` includes `test/`, so the test fixtures are covered by that too.
- Main's CI for `6b41b4332f` (#9680) was green — all Build jobs (which run tsc) succeeded, and the subsequent main runs were green.
- The failure is only reproducible with SDK **≤ 0.98.1**, which has *zero* occurrences of `output_tokens_details` in its type declarations. The 0.98.1 → 0.100.1 bump (#9651, `4fcaa87067`) is an ancestor of #9680, so main never had #9680's code with the old SDK. The `as` → `as unknown as` fixture changes in this PR are exactly the symptom of compiling against the pre-0.100 SDK.

Repro: `git checkout origin/main && npm ci && npm run tsc` → clean. The fix for the originating environment is `npm ci`, not a code change. If the automation pipeline that produced this PR caches `node_modules`, adding an `npm ci` step before its `tsc` verification gate would prevent this class of phantom-fix PR.

### Net effect of the diff

- **Type safety strictly decreases vs main** (details in inline comments): the local `AnthropicUsageWithThinkingDetails` is weaker than the SDK's own `Usage`; the type-predicate filter is a disguised cast; the two `as` casts disable the structural checking that main's plain object literal gets from the `reduce` type parameter; and the `as unknown as` fixture casts turn off all fixture checking. The function body compiles with **zero** casts and **zero** local types — that's what main already has.
- **The runtime change is a no-op for consumers, but contradicts the PR description.** The description says the change "keeps the runtime aggregation behavior", but the conditional spread omits the `output_tokens_details` key where main always materializes it (summed object or explicit `null`). The sole consumer (`src/providers/anthropic/util.ts:555`) reads it via `?.thinking_tokens`, so the delta is unobservable — and no test pins it.

### QA performed

- `npm run tsc`: passes on `origin/main` and on this branch.
- `npx vitest run test/providers/anthropic/messages.test.ts test/providers/anthropic/util.test.ts`: **240/240 pass** on this branch, including `sums thinking tokens across MCP continuation rounds`. The same tests also pass unchanged on main — i.e. no test distinguishes this branch from main.
- **Live A/B eval** (real Anthropic API, `anthropic:messages:claude-sonnet-4-6` with `thinking: enabled`, one provider also with `mcp: enabled` against the hosted deepwiki server to force multi-round continuation through `mergeAnthropicUsage`):
  - This branch: MCP+thinking → `completionDetails.reasoning: 31` (prompt ≈ 295k summed across rounds, confirming multiple merged rounds); thinking-only → `reasoning: 37`.
  - `origin/main`: MCP+thinking → `reasoning: 31`; thinking-only → `reasoning: 36` (live-model variance).
  - Identical observable behavior; thinking tokens survive the multi-round merge on both.

### Recommendation

Close the PR. Main already implements and tests the stated goal with the SDK's own types. If any part of this is kept, at minimum: drop the local type, the predicate filter, and all three casts (the body compiles without them), and revert the two `as unknown as` fixture casts.

Happy to be proven wrong: a reproduction of the claimed `tsc` failure on a clean `npm ci` checkout of current main (with the lockfile's SDK 0.100.1) would change this assessment.

wholley-oai:
Verified the review feedback. This worktree was using stale shared dependencies: local `node_modules` has `@anthropic-ai/sdk@0.97.1`, while the repo lockfile pins `0.100.1`. The `tsc` failure I reproduced came from that environment mismatch, not from current `origin/main`. Closing rather than merging a no-op/type-weakening diff.

PR #9689, opened 2026-06-10
fix(redteam): improve remote generation error diagnostics

### LINKED ISSUE #9679: Red team generation fails with `TypeError: terminated` - Cloudflare 403 blocks POST to /api/v1/task
`promptfoo redteam run` fails during **test-case generation** for every plugin with `TypeError: terminated`. The root cause is a **Cloudflare 403 ("Attention Required") block on `POST https://api.promptfoo.app/api/v1/task`**  from my network. `GET` requests  to the same host succeed, so it is scoped to the generation endpoint/method.

Image

drooph0904:
## Summary

Fixes #9679. Red team remote generation `POST`s to `api.promptfoo.app/api/v1/task`. When a network proxy, firewall, or Cloudflare rule blocks the request, the connection is cut and undici throws `TypeError: terminated` — with the real reason tucked into `error.cause`. The catch blocks logged the raw error (`` `...: ${err}` ``), which stringifies to just `TypeError: terminated`, so every plugin and strategy reported the same opaque message with no hint that it was a network-level block.

This is **diagnostics only** — it does not try to make a blocked request succeed. (Two small, intentional behavior deltas are called out under *Review follow-ups* below.)

## What changed

New helpers in `src/util/fetch/errors.ts`:

- **`isConnectionError(error)`** — detects connection-layer failures (refused / reset / dropped) vs a clean HTTP response, inspecting both the outer error and its `cause` (codes `UND_ERR_SOCKET`, `ECONNRESET`, `ECONNREFUSED`, `EPIPE`, `EAI_AGAIN`, plus message text like `terminated` / `other side closed`).
- **`formatFetchError(error)`** — surfaces the otherwise-hidden `cause` and `code`. This is the existing private `formatFetchErrorMessage` from `util/fetch/index.ts`, moved here and exported (the duplicate is removed, and `fetchWithRetries` now imports it).
- **`describeFetchError(error)`** — `formatFetchError` plus a one-line network-block hint **only when** `isConnectionError` is true.

Wired `describeFetchError` into the remote-generation catch sites: plugin generation (`plugins/index.ts`), goal/intent extraction (`redteam/util.ts`), remote extraction (`extraction/util.ts`), and the citation / likert / mathPrompt / singleTurnComposite / simpleAudio / multilingual strategies.

### Before
```
Error generating test cases for <plugin>: TypeError: terminated
```
### After
```
Error generating test cases for <plugin>: TypeError: terminated (Cause: SocketError: other side closed) (Code: UND_ERR_SOCKET)
The connection failed before a response was received. Common causes are a network proxy,
firewall, or Cloudflare rule blocking the request (for example a 403 on POST), a DNS
resolution failure, a TLS/certificate misconfiguration, or no network connectivity. Check
your network, DNS, and proxy settings, or try a different network.
```

## Verification

- Reproduced the exact error shape (`TypeError: terminated`, `cause` = `SocketError: other side closed`, `cause.code` = `UND_ERR_SOCKET`) against a local server that cuts the socket mid-response — the new unit tests assert on those captured values, so they hold on every Node version without network access.
- New unit coverage in `test/util/fetch/errors.test.ts` for `isConnectionError` / `formatFetchError` / `describeFetchError` (block variants, plain app errors → no hint, non-Error inputs).
- `tsc --noEmit` clean; Biome clean (pre-existing `multilingual.ts` complexity warnings only, unchanged from `main`).
- Full touched-area suites pass (errors, `fetch.test.ts`, redteam plugins/strategies/util/extraction). `fetch.test.ts`'s exact-message assertions confirm the de-dup changed no existing behavior.

## Notes

The hint is intentionally worded as a list of likely causes since `terminated` / `fetch failed` can have several roots (proxy block, DNS, TLS, offline). `describeFetchError` degrades gracefully on non-connection errors (it just adds `cause`/`code`, never a misleading hint).

## Review follow-ups (hardening)

Addressing review feedback (incl. the Codex bot) and broadening test coverage:

- **Security — redact object causes before logging.** `formatErrorCause` now routes non-`Error` object causes through `sanitizeObject` before `JSON.stringify`. The formatted string is interpolated into a plain log *message*, and the logger only sanitizes its structured *context* argument — so a raw `JSON.stringify` of an object cause could write `Authorization` / api-key / cookie fields to logs verbatim (the old code rendered a harmless `[object Object]`; the un-redacted JSON dump it replaced was a latent leak). `sanitizeObject` redacts sensitive keys and also collapses circular references. No reachable leak exists today (real undici causes are `Error` instances, rendered via `String()`), so this is defense-in-depth, but it removes the footgun cheaply.
- **Correctness — cause-agnostic hint.** `isConnectionError` matches undici's generic `fetch failed` wrapper and DNS (`EAI_AGAIN`), so the hint also fires for DNS / TLS / offline failures, not just proxy blocks. The hint text now names those causes (and still leads with the proxy/Cloudflare case from #9679) instead of asserting a block — avoiding misattribution for a DNS typo or cert mismatch.
- **Tests — close per-call-site gaps.** Added connection-failure regression guards for the `citation`, `mathPrompt`, and `simpleAudio` call sites (previously only `likert` / `singleTurnComposite` / `extraction` were pinned, so a silent revert to bare `${error}` at those three would have shipped green). Added `formatFetchError` tests for secret redaction, a non-serializable (BigInt) cause, and a circular cause.

### Intentional behavior deltas (not pure "diagnostics only")

- **`multilingual.ts`: `logger.debug` → `logger.warn`.** A remote multilingual-generation failure was previously silent at the default log level; it now surfaces as a single `warn` per run (it is bounded — per-chunk/per-language failures stay at `debug`). The strategy still falls back to local translation, so this warns on a recoverable condition by design, to make a previously-invisible remote outage visible.
- **`fetchWithRetries` debug/throw message.** Because `formatFetchError` replaces the old private formatter, the retry-exhausted message now renders object causes as (redacted) JSON instead of `[object Object]` and surfaces a `cause.code` fallback. The existing `fetch.test.ts` assertions (string cause) are unchanged; this only enriches the human-facing message.

### End-to-end QA

Reproduced #9679 against a local server that answers `GET /health` but **cuts the socket on `POST /api/v1/task`** (the Cloudflare-403-on-POST shape), then ran `redteam generate`. The real undici retry path now logs:

```
Error using remote generation for task 'entities': Error: Request failed after 4 retries: TypeError: fetch failed (Cause: SocketError: other side closed) (Code: UND_ERR_SOCKET)
The connection failed before a response was received. Common causes are a network proxy, ...
```

Confirmed: the hidden cause + code are surfaced; the hint fires once per failing operation (the `entities.ts` catch correctly uses `formatFetchError`, not `describeFetchError`, so the hint is not duplicated); generation degrades gracefully (no crash); and **no `Authorization`/api-key value appears anywhere in the logs**. Also verified a real DNS failure surfaces `Code: ENOTFOUND` with the DNS-aware hint, and a real object cause carrying `authorization` / `x-api-key` is emitted as `[REDACTED]`.

### DIFF
--- a/architecture/edge-baseline.json
+++ b/architecture/edge-baseline.json
@@ -41,7 +41,7 @@
   "redteam -> core": 2,
   "redteam -> legacy-contracts": 149,
   "redteam -> legacy-runtime": 143,
-  "redteam -> node": 217,
+  "redteam -> node": 218,
   "redteam -> providers": 43,
   "view-server -> core": 1,
   "view-server -> legacy-contracts": 21,
--- a/architecture/edge-baseline.json
+++ b/architecture/edge-baseline.json
@@ -41,7 +41,7 @@
   "redteam -> core": 2,
   "redteam -> legacy-contracts": 149,
   "redteam -> legacy-runtime": 143,
-  "redteam -> node": 208,
+  "redteam -> node": 217,
   "redteam -> providers": 43,
   "view-server -> core": 1,
   "view-server -> legacy-contracts": 21,
--- a/src/redteam/extraction/entities.ts
+++ b/src/redteam/extraction/entities.ts
@@ -1,5 +1,6 @@
 import dedent from 'dedent';
 import logger from '../../logger';
+import { formatFetchError } from '../../util/fetch/errors';
 import { shouldGenerateRemote } from '../remoteGeneration';
 import { callExtraction, fetchRemoteGeneration, formatPrompts } from './util';
 
@@ -12,8 +13,11 @@ export async function extractEntities(provider: ApiProvider, prompts: string[]):
       const result = await fetchRemoteGeneration('entities' as RedTeamTask, prompts);
       return result as string[];
     } catch (error) {
+      // fetchRemoteGeneration already logged the full error and the
+      // connection-block hint (if any), so format without re-appending it —
+      // surface cause/code here instead of a bare `TypeError: terminated`.
       logger.warn(
-        `[Entity Extraction] Failed, returning 0 entities. Error using remote generation: ${error}`,
+        `[Entity Extraction] Failed, returning 0 entities. Error using remote generation: ${formatFetchError(error)}`,
       );
       return [];
     }
--- a/src/redteam/extraction/util.ts
+++ b/src/redteam/extraction/util.ts
@@ -6,6 +6,7 @@ import { getEnvBool } from '../../envars';
 import { getUserEmail } from '../../globalConfig/accounts';
 import logger from '../../logger';
 import { getRequestTimeoutMs } from '../../providers/shared';
+import { describeFetchError } from '../../util/fetch/errors';
 import invariant from '../../util/invariant';
 import { getRemoteGenerationHeaders, getRemoteGenerationUrl } from '../remoteGeneration';
 
@@ -62,7 +63,7 @@ export async function fetchRemoteGeneration(
     const parsedResponse = RedTeamGenerationResponse.parse(response.data);
     return parsedResponse.result;
   } catch (error) {
-    logger.warn(`Error using remote generation for task '${task}': ${error}`);
+    logger.warn(`Error using remote generation for task '${task}': ${describeFetchError(error)}`);
     throw error;
   }
 }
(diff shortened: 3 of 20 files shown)


drooph0904:
Thanks for taking this the last mile, @mldangelo 🙏 The `access_terminated` false-positive especially is a great catch — I'd matched the substring without thinking through what else could contain "terminated."

I lean on AI heavily to understand the code and generate edge cases, and I get the core logic right, but I still miss a few of these. Do you have any advice on how you approach finding these edge cases — anything in your review process I could adopt so I catch them *before* the PR next time?

PR #9703, opened 2026-06-11
fix(codex): propagate trace identity to CLI telemetry

daneschneider-oai:
## Summary

- propagate Promptfoo trace and parent-span identity to Codex CLI OTEL resource attributes
- preserve unrelated resource attributes while replacing stale Promptfoo values
- share the existing Claude/Codex resource-attribute helper

## Why

Codex CLI OTLP logs were reaching Promptfoo's receiver but being dropped because the subprocess did not receive the resource attributes Promptfoo uses to associate logs with the active trace. This prevented `codex.api_request` diagnostics from being persisted when deep tracing was enabled.

This keeps event streaming behavior unchanged. It only fixes trace correlation for Codex deep tracing.

## Validation

- `test/providers/openai-codex-sdk.test.ts`
- `test/providers/claude-agent-sdk.test.ts`
- 309 focused provider tests passed
- Biome 2.4.16 formatting/check on the four changed files

The full TypeScript check was not meaningful in the local worktree because its shared dependency directory predates current `main` and reports broad unrelated missing/new dependency APIs.

### DIFF
--- a/site/docs/providers/openai-codex-sdk.md
+++ b/site/docs/providers/openai-codex-sdk.md
@@ -458,7 +458,17 @@ providers:
       enable_streaming: true
 ```
 
-Deep tracing injects OpenTelemetry environment variables (`OTEL_EXPORTER_OTLP_ENDPOINT`, `TRACEPARENT`, etc.) into the Codex CLI process. Promptfoo uses a fresh SDK client/thread per call in this mode so child spans link to the correct parent request span.
+Codex configures its log exporter through the `config.toml` in `CODEX_HOME`. Point it at Promptfoo's JSON logs receiver using the complete `/v1/logs` endpoint:
+
+```toml title="$CODEX_HOME/config.toml"
+[otel]
+log_user_prompt = false
+exporter = { otlp-http = { endpoint = "http://127.0.0.1:4318/v1/logs", protocol = "json" } }
+```
+
+If you override `cli_env.CODEX_HOME`, put this configuration in that directory. The endpoint is the complete logs URL, not merely the OTLP host and port.
+
+Deep tracing injects `TRACEPARENT` and `promptfoo.trace_id` / `promptfoo.parent_span_id` resource attributes into the Codex CLI process so log records remain linked even when Codex does not attach inline trace context. Promptfoo uses a fresh SDK client/thread per call in this mode so child spans link to the correct parent request span. Standard OpenTelemetry environment variables are also injected, but they do not replace Codex's `[otel]` exporter configuration.
 
 :::warning
 
--- a/src/providers/openai/codex-sdk.ts
+++ b/src/providers/openai/codex-sdk.ts
@@ -15,6 +15,8 @@ import {
   type GenAISpanResult,
   getTraceparent,
   openTurnSpan,
+  PROMPTFOO_RESOURCE_ATTR_PARENT_SPAN_ID,
+  PROMPTFOO_RESOURCE_ATTR_TRACE_ID,
   withGenAISpan,
 } from '../../tracing/genaiTracer';
 import { formatRateLimitErrorMessage, HttpRateLimitError } from '../../util/fetch/errors';
@@ -48,8 +50,8 @@ function appendPromptfooResourceAttrs(
   traceId: string,
   parentSpanId: string,
 ): string {
-  const traceKey = 'promptfoo.trace_id';
-  const parentSpanKey = 'promptfoo.parent_span_id';
+  const traceKey = PROMPTFOO_RESOURCE_ATTR_TRACE_ID;
+  const parentSpanKey = PROMPTFOO_RESOURCE_ATTR_PARENT_SPAN_ID;
   const incoming = `${traceKey}=${traceId},${parentSpanKey}=${parentSpanId}`;
   if (!existing) {
     return incoming;
--- a/src/providers/openai/codex-sdk.ts
+++ b/src/providers/openai/codex-sdk.ts
@@ -17,7 +17,6 @@ import {
   openTurnSpan,
   withGenAISpan,
 } from '../../tracing/genaiTracer';
-import { appendPromptfooResourceAttrs } from '../../tracing/resourceAttributes';
 import { formatRateLimitErrorMessage, HttpRateLimitError } from '../../util/fetch/errors';
 import { renderVarsInObject } from '../../util/render';
 import { normalizeFieldName, REDACTED, sanitizeObject } from '../../util/sanitizer';
@@ -43,6 +42,30 @@ import type {
   CallApiOptionsParams,
   ProviderResponse,
 } from '../../types/index';
+
+function appendPromptfooResourceAttrs(
+  existing: string | undefined,
+  traceId: string,
+  parentSpanId: string,
+): string {
+  const traceKey = 'promptfoo.trace_id';
+  const parentSpanKey = 'promptfoo.parent_span_id';
+  const incoming = `${traceKey}=${traceId},${parentSpanKey}=${parentSpanId}`;
+  if (!existing) {
+    return incoming;
+  }
+  const cleaned = existing
+    .split(',')
+    .map((pair) => pair.trim())
+    .filter(
+      (pair) =>
+        pair.length > 0 &&
+        !pair.startsWith(`${traceKey}=`) &&
+        !pair.startsWith(`${parentSpanKey}=`),
+    )
+    .join(',');
+  return cleaned.length > 0 ? `${cleaned},${incoming}` : incoming;
+}
 
 /**
  * OpenAI Codex SDK Provider
--- a/src/providers/openai/codex-sdk.ts
+++ b/src/providers/openai/codex-sdk.ts
@@ -17,6 +17,7 @@ import {
   openTurnSpan,
   withGenAISpan,
 } from '../../tracing/genaiTracer';
+import { appendPromptfooResourceAttrs } from '../../tracing/resourceAttributes';
 import { formatRateLimitErrorMessage, HttpRateLimitError } from '../../util/fetch/errors';
 import { renderVarsInObject } from '../../util/render';
 import { normalizeFieldName, REDACTED, sanitizeObject } from '../../util/sanitizer';
@@ -791,6 +792,14 @@ export class OpenAICodexSDKProvider implements ApiProvider {
       // W3C Trace Context - only set if we have a traceparent for proper parent-child linking
       if (traceparent) {
         sortedEnv.TRACEPARENT = traceparent;
+        const [, traceId, parentSpanId] = traceparent.split('-');
+        if (traceId && parentSpanId) {
+          sortedEnv.OTEL_RESOURCE_ATTRIBUTES = appendPromptfooResourceAttrs(
+            sortedEnv.OTEL_RESOURCE_ATTRIBUTES,
+            traceId,
+            parentSpanId,
+          );
+        }
       }
       logger.debug('[CodexSDK] Injecting OTEL config for deep tracing', {
         traceparent: traceparent || '(none - CLI will start own trace)',
@@ -2082,7 +2091,7 @@ export class OpenAICodexSDKProvider implements ApiProvider {
 
     // Get current trace context for deep tracing
     // This allows the Codex CLI to export its internal spans as children of our span
-    const currentTraceparent = getTraceparent();
+    const currentTraceparent = getTraceparent() ?? context?.traceparent;
     const apiKey = this.getApiKey(config);
     const workingDirectory =
       resolveAgenticWorkingDir(config.working_dir, cliState.basePath) ?? process.cwd();
--- a/src/providers/openai/codex-sdk.ts
+++ b/src/providers/openai/codex-sdk.ts
@@ -67,6 +67,17 @@ function appendPromptfooResourceAttrs(
   return cleaned.length > 0 ? `${cleaned},${incoming}` : incoming;
 }
 
+const ZERO_TRACE_ID = '00000000000000000000000000000000';
+const ZERO_SPAN_ID = '0000000000000000';
+
+function isValidTraceparent(traceparent: string | undefined): traceparent is string {
+  if (!traceparent) {
+    return false;
+  }
+  const [, traceId, spanId] = traceparent.split('-');
+  return Boolean(traceId && spanId && traceId !== ZERO_TRACE_ID && spanId !== ZERO_SPAN_ID);
+}
+
 /**
  * OpenAI Codex SDK Provider
  *
@@ -815,12 +826,12 @@ export class OpenAICodexSDKProvider implements ApiProvider {
       // W3C Trace Context - only set if we have a traceparent for proper parent-child linking
       if (traceparent) {
         sortedEnv.TRACEPARENT = traceparent;
-        const [, traceId, parentSpanId] = traceparent.split('-');
-        if (traceId && parentSpanId) {
+        const [, tpTraceId, tpSpanId] = traceparent.split('-');
+        if (tpTraceId && tpSpanId) {
           sortedEnv.OTEL_RESOURCE_ATTRIBUTES = appendPromptfooResourceAttrs(
             sortedEnv.OTEL_RESOURCE_ATTRIBUTES,
-            traceId,
-            parentSpanId,
+            tpTraceId,
+            tpSpanId,
           );
         }
       }
@@ -2114,7 +2125,10 @@ export class OpenAICodexSDKProvider implements ApiProvider {
 
     // Get current trace context for deep tracing
     // This allows the Codex CLI to export its internal spans as children of our span
-    const currentTraceparent = getTraceparent() ?? context?.traceparent;
+    const activeTraceparent = getTraceparent();
+    const currentTraceparent = isValidTraceparent(activeTraceparent)
+      ? activeTraceparent
+      : context?.traceparent;
     const apiKey = this.getApiKey(config);
     const workingDirectory =
       resolveAgenticWorkingDir(config.working_dir, cliState.basePath) ?? process.cwd();
--- a/src/tracing/otlpReceiver.ts
+++ b/src/tracing/otlpReceiver.ts
@@ -128,6 +128,13 @@ function isZeroSpanId(id: string): boolean {
   return /^0+$/.test(id) || BASE64_ZERO_SPAN_ID.test(id);
 }
 
+// OTEL trace_id is 32 hex chars / 16 bytes. Its base64 representation is
+// 22 `A` characters followed by optional `==` padding.
+const BASE64_ZERO_TRACE_ID = /^A{22}(?:==)?$/;
+function isZeroTraceId(id: string): boolean {
+  return /^0+$/.test(id) || BASE64_ZERO_TRACE_ID.test(id);
+}
+
 function randomSpanId(): string {
   return crypto.randomBytes(8).toString('hex');
 }
@@ -762,11 +769,13 @@ export class OTLPReceiver {
   ): ParsedTrace | null {
     // Prefer an inline traceId on the log record (set when the SDK propagated
     // TRACEPARENT into its logs context). Fall back to the resource attribute
-    // promptfoo.trace_id — the claude-agent-sdk provider injects this via
-    // OTEL_RESOURCE_ATTRIBUTES because Claude Code's logs signal does not
-    // inherit TRACEPARENT. Without either, we can't link anywhere useful.
-    const rawTraceId =
-      log.traceId ?? getStringAttribute(resourceAttributes, PROMPTFOO_RESOURCE_ATTR_TRACE_ID);
+    // promptfoo.trace_id — agent providers inject this via
+    // OTEL_RESOURCE_ATTRIBUTES because some log signals do not inherit
+    // TRACEPARENT. Without either, we can't link anywhere useful.
+    const hasValidInlineTraceId = !!log.traceId && !isZeroTraceId(log.traceId);
+    const rawTraceId = hasValidInlineTraceId
+      ? log.traceId
+      : getStringAttribute(resourceAttributes, PROMPTFOO_RESOURCE_ATTR_TRACE_ID);
     if (!rawTraceId) {
       logger.debug(
         `[OtlpReceiver] Dropping log: no traceId and no ${PROMPTFOO_RESOURCE_ATTR_TRACE_ID} resource attribute (scope=${scopeLog.scope?.name ?? 'unknown'}). Ensure TRACEPARENT is propagated or OTEL_RESOURCE_ATTRIBUTES is set by the provider.`,
--- a/test/providers/openai-codex-sdk.test.ts
+++ b/test/providers/openai-codex-sdk.test.ts
@@ -8,6 +8,7 @@ import { getDirectory, importModule, resolvePackageEntryPoint } from '../../src/
 import logger from '../../src/logger';
 import { OpenAICodexSDKProvider } from '../../src/providers/openai/codex-sdk';
 import { providerRegistry } from '../../src/providers/providerRegistry';
+import { getTraceparent } from '../../src/tracing/genaiTracer';
 import { checkProviderApiKeys } from '../../src/util/provider';
 import { createDeferred, mockProcessEnv } from '../util/utils';
 
@@ -52,6 +53,11 @@ vi.mock('../../src/esm', async (importOriginal) => {
 // Mock the SDK package (for type safety)
 vi.mock('@openai/codex-sdk', () => mockCodexSDK);
 
+vi.mock('../../src/tracing/genaiTracer', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('../../src/tracing/genaiTracer')>()),
+  getTraceparent: vi.fn(),
+}));
+
 // Helper to create mock response matching real SDK format
 const createMockResponse = (
   finalResponse: string,
@@ -90,6 +96,7 @@ describe('OpenAICodexSDKProvider', () => {
   let existsSyncSpy: MockInstance;
   const mockImportModule = vi.mocked(importModule);
   const mockResolvePackageEntryPoint = vi.mocked(resolvePackageEntryPoint);
+  const mockGetTraceparent = vi.mocked(getTraceparent);
   let originalBasePath: string | undefined;
   let originalOpenAiApiKey: string | undefined;
   let originalCodexApiKey: string | undefined;
@@ -115,6 +122,7 @@ describe('OpenAICodexSDKProvider', () => {
     mockImportModule.mockResolvedValue(mockCodexSDK);
     mockResolvePackageEntryPoint.mockReset();
     mockResolvePackageEntryPoint.mockReturnValue('@openai/codex-sdk');
+    mockGetTraceparent.mockReturnValue(undefined);
 
     // Default mocks
     statSyncSpy = vi.spyOn(fs, 'statSync').mockReturnValue({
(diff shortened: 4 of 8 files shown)


src/providers/openai/codex-sdk.ts
```diff
@@ -791,6 +815,14 @@ export class OpenAICodexSDKProvider implements ApiProvider {
...
       if (traceparent) {
         sortedEnv.TRACEPARENT = traceparent;
+        const [, traceId, parentSpanId] = traceparent.split('-');
+        if (traceId && parentSpanId) {
+          sortedEnv.OTEL_RESOURCE_ATTRIBUTES = appendPromptfooResourceAttrs(
+            sortedEnv.OTEL_RESOURCE_ATTRIBUTES,
+            traceId,
+            parentSpanId,
+          );
+        }
```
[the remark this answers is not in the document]
daneschneider-oai: Addressed in 64dc7a139. The parsed W3C fields are now named tpTraceId and tpSpanId, matching the Claude provider terminology.

src/providers/openai/codex-sdk.ts
```diff
@@ -43,6 +43,41 @@ import type {
   ProviderResponse,
 } from '../../types/index';
 
+function appendPromptfooResourceAttrs(
+  existing: string | undefined,
+  traceId: string,
+  parentSpanId: string,
+): string {
+  const traceKey = 'promptfoo.trace_id';
+  const parentSpanKey = 'promptfoo.parent_span_id';
```
mldangelo: These hardcoded literals duplicate `PROMPTFOO_RESOURCE_ATTR_TRACE_ID` / `PROMPTFOO_RESOURCE_ATTR_PARENT_SPAN_ID` from `src/tracing/resourceAttributes.ts`, whose header says: *"Both producer and consumer must agree; keep these constants as the single source of truth."* The consumer (`otlpReceiver.ts:769/817`) and the other producer (`claude-agent-sdk.ts:101-121`, which this helper duplicates nearly verbatim) both use the constants. If the keys ever change, codex log linkage breaks silently — logs drop with only a debug message, and every test here stays green.

I understand commit e6afb1b made this provider-local because the shared version trips the architecture edge ratchet (`providers -> legacy-runtime` 312 → 313). Two better options:

1. **Ratchet-neutral (preferred):** move the shared logic into `src/tracing/genaiTracer.ts` next to `getTraceparent()` — e.g. `resolveSubprocessTraceparent(context)` + `applyTraceEnvToSubprocess(env, traceparent)` covering guard + fallback + TRACEPARENT + resource attrs. All three subprocess providers (claude-agent-sdk, codex-sdk, codex-app-server) already import `genaiTracer`, so no new import statement is needed — and `codex-app-server` (which currently has none of this, see `codex-app-server.ts:1267`) could be fixed in a few lines.
2. **Acceptable:** keep the helper in `src/tracing/resourceAttributes.ts` as in your first commit and take the sanctioned one-line `npm run architecture:baseline` bump — the edge is natively allowed (`architecture/layers.json` providers → legacy-runtime), the ratchet's error message explicitly invites justified bumps, and restoring a documented single source of truth qualifies.

At absolute minimum, import the two constants instead of re-typing the strings (adding names to an existing import is ratchet-free, though there's no existing `resourceAttributes` import here — so option 1 or 2 is cleaner).

src/providers/openai/codex-sdk.ts
```diff
@@ -791,6 +826,14 @@ export class OpenAICodexSDKProvider implements ApiProvider {
       // W3C Trace Context - only set if we have a traceparent for proper parent-child linking
       if (traceparent) {
         sortedEnv.TRACEPARENT = traceparent;
+        const [, tpTraceId, tpSpanId] = traceparent.split('-');
+        if (tpTraceId && tpSpanId) {
+          sortedEnv.OTEL_RESOURCE_ATTRIBUTES = appendPromptfooResourceAttrs(
```
mldangelo, line 833: I verified this injection end-to-end: the spawned codex CLI (bundled 0.130.0) receives `OTEL_RESOURCE_ATTRIBUTES` and its exported `/v1/logs` payloads do carry both `promptfoo.*` resource attributes (confirmed by dumping the parsed resource on the receiver side). On `main`, the same run shows no `promptfoo.*` attrs — so this change is necessary and works at the transport level. ✅

However, the receiver still drops **every** codex log record even with the attrs present (see my review summary): codex's Rust exporter emits `"traceId": ""` (empty string, key present) on log records, and `otlpReceiver.ts:768` does `log.traceId ?? getStringAttribute(...)` — empty string is not nullish, so the fallback this PR enables is never reached and `!rawTraceId` drops the record. The `spanId` path right below already guards this (`!!log.spanId && !isZeroSpanId(log.spanId)`); `traceId` needs the same. Without that one-line receiver fix (verified working in my run: 22/22 records became correctly-parented spans), this PR's stated outcome isn't observable.

src/providers/openai/codex-sdk.ts
```diff
@@ -2082,7 +2125,10 @@ export class OpenAICodexSDKProvider implements ApiProvider {
 
     // Get current trace context for deep tracing
     // This allows the Codex CLI to export its internal spans as children of our span
-    const currentTraceparent = getTraceparent();
+    const activeTraceparent = getTraceparent();
+    const currentTraceparent = isValidTraceparent(activeTraceparent)
+      ? activeTraceparent
+      : context?.traceparent;
```
mldangelo, line 2133: The guard-then-fallback ordering added in 64dc7a1 resolves the earlier P2 (zero/invalid active traceparent beating a valid `context.traceparent`), and `isValidTraceparent` is actually stricter than the claude twin (`claude-agent-sdk.ts:1352` only substring-checks the zero **trace** id, not a zero span id) — nice. Two follow-ups worth considering:

1. **Hardening (minor):** the fallback `context?.traceparent` is propagated unvalidated — `isValidTraceparent(active) ? active : (isValidTraceparent(context?.traceparent) ? context.traceparent : undefined)` would keep a malformed/zero evaluator value out of `TRACEPARENT` and the resource attrs. Low risk today since the evaluator generates real ids.
2. **Parity:** `codex-app-server.ts:1267` still has the fully unguarded pre-PR pattern (bare `getTraceparent()`, no fallback, no resource attrs) while spawning the same codex binary under the same `deep_tracing` flag — its CLI logs will always be dropped by the receiver. Worth porting in this PR or tracking as an explicit follow-up; and the claude guard could be tightened to this two-field check at the same time (ideally all via one shared helper, see my other comment).

test/providers/openai-codex-sdk.test.ts
```diff
@@ -2284,6 +2292,68 @@ describe('OpenAICodexSDKProvider', () => {
...
+              TRACEPARENT: `00-${traceId}-${spanId}-01`,
+              OTEL_RESOURCE_ATTRIBUTES:
+                `deployment.environment=test,promptfoo.trace_id=${traceId},` +
+                `promptfoo.parent_span_id=${spanId}`,
+            }),
+          }),
+        );
+      });
+
+      it.each([
```
mldangelo, line 2357: Good coverage of the rejection side (zero trace id / zero span id → evaluator trace wins), and the mock setup is careful (`importOriginal` spread keeps `withGenAISpan` real; the unconditional `beforeEach` re-seed makes the per-test overrides random-order safe).

One gap: no test pins the **positive** branch — `mockGetTraceparent` returning a *valid* traceparent that should win over a different `context.traceparent`. As written, flipping the ternary at `codex-sdk.ts:2129-2131` to always use `context.traceparent` would pass this entire suite. One more `it`: mock a valid active traceparent, pass a different context traceparent, assert `TRACEPARENT` and the `promptfoo.*` attrs use the **active** values.

mldangelo (commented):
I did a deep review of this PR: checked out the branch, ran the unit suites, both review bots' findings, a cross-provider parity/architecture audit, and — most importantly — an actual end-to-end run of codex deep tracing against promptfoo's OTLP receiver. Summary below; inline comments carry the file-specific details.

## TL;DR

The provider-side change is **correct, necessary, and well-tested at the unit level** (the `isValidTraceparent` guard added in 64dc7a1 resolves the earlier P2 bot findings, and is actually stricter than the claude-agent-sdk equivalent). However, **end-to-end the stated goal still does not work**: with this PR checked out, every Codex CLI log record is still dropped by the receiver. The blocker is on the receiver side, not in this diff — details and a verified one-line fix below. I'd recommend adding that receiver fix (plus a regression test) to this PR, since without it the feature this PR describes ("codex.api_request diagnostics ... persisted") doesn't materialize.

## End-to-end verification (the important part)

Setup: `openai:codex-sdk` with `deep_tracing: true`, `tracing.enabled: true` (OTLP receiver on 4318), and a `CODEX_HOME` whose `config.toml` enables the codex OTLP logs exporter:

```toml
[otel]
environment = "dev"
log_user_prompt = true
exporter = { otlp-http = { endpoint = "http://127.0.0.1:4318/v1/logs", protocol = "json" } }
```

**What works (verified):**
- The injected env reaches the spawned CLI. I shimmed the codex binary via `codex_path_override` and captured the spawned env: `TRACEPARENT` and `OTEL_RESOURCE_ATTRIBUTES=promptfoo.trace_id=...,promptfoo.parent_span_id=...` are present and correct.
- The codex CLI (bundled 0.130.0) picks the env var up: its exported OTLP `/v1/logs` payloads carry both `promptfoo.*` resource attributes (verified by dumping the receiver-side parsed resource: `{"service.name":"codex_sdk_ts", ..., "promptfoo.trace_id":"a1d0cb...", "promptfoo.parent_span_id":"baed6f..."}`).
- Counterfactual on `main` (PR provider reverted, all else equal): the resource has **no** `promptfoo.*` attributes and all logs drop. So this PR's injection is necessary.

**What still fails:** on this PR's tip, all 22 log records in my run were still dropped:

```
[OtlpReceiver] Dropping log: no traceId and no promptfoo.trace_id resource attribute (scope=codex_otel.log_only). ...
```

…even though `promptfoo.trace_id` was present in the resource. Root cause: codex's Rust OTLP exporter serializes **empty trace ids as `"traceId": ""`** (key present, empty string) on every log record. The receiver does:

```ts
// src/tracing/otlpReceiver.ts:768
const rawTraceId =
  log.traceId ?? getStringAttribute(resourceAttributes, PROMPTFOO_RESOURCE_ATTR_TRACE_ID);
```

`"" ?? x` never falls through, so the resource-attribute fallback this PR adds is **unreachable** for codex logs, and `!rawTraceId` then drops the record. (Claude Code's JS exporter *omits* the field, which is why the same pattern works for claude-agent-sdk.) Note the parent-span path two lines down already guards exactly this: `const hasValidInlineSpanId = !!log.spanId && !isZeroSpanId(log.spanId)` — the traceId path needs the same treatment.

**Verified fix** (I ran the full eval with this patch — all 22 codex log records landed as spans under the evaluation trace, correctly parented to the provider span; zero drops):

```ts
const inlineTraceId = log.traceId && !/^0*$/.test(log.traceId) ? log.traceId : undefined;
const rawTraceId =
  inlineTraceId ?? getStringAttribute(resourceAttributes, PROMPTFOO_RESOURCE_ATTR_TRACE_ID);
```

Please add this (or equivalent) with a regression test using a codex-shaped payload (`traceId: ""`, resource attrs set) — otherwise this PR ships a fix that can't be observed working.

## Cross-provider parity (is the same thing done the same way everywhere?)

Inventory of all agent-subprocess providers and their trace propagation:

| Provider | TRACEPARENT | zero-id guard | context fallback | resource attrs |
|---|---|---|---|---|
| `openai:codex-sdk` (this PR) | yes | **strong** (trace+span id) | yes | yes |
| `anthropic:claude-agent-sdk` | yes | weak (zero trace id substring only, `claude-agent-sdk.ts:1352`) | yes | yes (shared constants) |
| `openai:codex-app-server` | yes (`codex-app-server.ts:1458`) | **none** (bare `getTraceparent()`, `:1267`) | **none** | **none** |
| `openai:opencode-sdk` | no tracing integration at all | — | — | — |

`codex-app-server` spawns the **same codex binary** with the same `deep_tracing` flag and has the exact pre-PR bug this PR fixes (its CLI logs will always be dropped, and an invalid active traceparent is exported unguarded). Since the PR is scoped `fix(codex)`, it would be reasonable to either include the app-server port here or file a tracked follow-up. The claude guard should also be tightened to the two-field check (an all-zero span id is invalid per W3C trace-context).

## Architecture

The commit history shows the helper was first shared via `src/tracing/resourceAttributes.ts` and then made provider-local in "fix: keep Codex trace helper provider-local" — the empirical reason being the architecture edge ratchet (`providers -> legacy-runtime` pinned at 312 in `architecture/edge-baseline.json`; the shared version fails with 313). Two notes on that trade-off:

1. The edge itself is **natively allowed** (`architecture/layers.json` lists `legacy-runtime` in providers' `allowedDependencies`; `claude-agent-sdk.ts:22` already imports `resourceAttributes`). The check's own error message sanctions `npm run architecture:baseline` for justified edges — restoring a documented single source of truth is exactly that case.
2. A **ratchet-neutral** option exists: put the shared logic next to `getTraceparent()` in `src/tracing/genaiTracer.ts` (e.g. `resolveSubprocessTraceparent(context)` + `applyTraceEnvToSubprocess(env, traceparent)` covering guard + fallback + TRACEPARENT + resource attrs). All three providers already import `genaiTracer`, so no new import statements are needed, and `codex-app-server` could then be fixed in ~3 lines. This also de-duplicates the near-identical 25-line helper that now exists twice.

As-is, the hardcoded `'promptfoo.trace_id'` / `'promptfoo.parent_span_id'` literals contradict `src/tracing/resourceAttributes.ts`'s documented contract ("keep these constants as the single source of truth") — a future key change would silently break codex linkage with all tests green (logs drop at debug level only).

## Docs gap (feature is currently undiscoverable)

My E2E run also confirmed: promptfoo's injected `OTEL_*` env vars alone do **not** make the codex CLI export anything. Export only happens when the user's `CODEX_HOME` `config.toml` enables an `[otel]` exporter, and codex treats the configured endpoint as **exact** (it must include `/v1/logs`; no per-signal path is appended). None of this — nor the new resource-attribute injection — is documented in `site/docs/providers/openai-codex-sdk.md` (the Deep Tracing section only says env vars are injected). The claude provider page already has the template for this (`claude-agent-sdk.md`, "The provider automatically injects `OTEL_RESOURCE_ATTRIBUTES=promptfoo.trace_id=...`..."). A short docs addition (enablement steps + endpoint shape + the injection sentence) would make this feature actually usable; happy to help with that.

## Tests

311/311 pass on the changed suites; the genaiTracer mock is correctly scoped (spreads `importOriginal`, `withGenAISpan` stays real) and random-order safe. One gap worth closing (inline comment): no test pins the **positive** branch — a valid active traceparent winning over `context.traceparent`. Also `test/test-hygiene` passes (88/88; this file is grandfathered for the resetAllMocks rule).

## What's good

- The guard-then-fallback ordering is correct and stricter than the claude twin (rejects zero span id too).
- Stale `promptfoo.*` pairs are replaced, unrelated user attrs preserved; `TRACEPARENT` still scrubbed when `deep_tracing` is off; stable cache env still excludes `TRACEPARENT`.
- Scope is clean: net diff vs main is exactly the two described files, no lockfile drift, architecture check passes at tip.

test/providers/openai-codex-sdk.test.ts
mldangelo: Resolved by 62fbb24 — thanks. Verified: the new test passes (150/150 on the file), and I ran a mutation check — flipping the precedence at `codex-sdk.ts:2129-2131` to always prefer `context.traceparent` fails exactly this test and nothing else, so it genuinely pins the positive branch.

mldangelo (commented):
Re-reviewed at 62fbb24 (`test(codex): pin active trace precedence`).

**Resolved since my last review:**
- ✅ Positive-branch coverage gap — the new test mocks a valid active traceparent against a different evaluator traceparent and asserts the active values win in both `TRACEPARENT` and the `promptfoo.*` resource attrs. Verified locally (150/150 on the file) plus a mutation check: flipping the ternary to always prefer `context.traceparent` fails exactly this test. With this, the unit coverage of the provider-side logic is solid: invalid-active rejection (both zero-id forms), fallback path with stale-attr replacement, and now active-wins precedence.

**Still open, in priority order** (provider code is unchanged since 64dc7a1, so my end-to-end findings stand):
1. **Receiver empty-string `traceId` fix (blocker for the feature actually working).** Codex's Rust exporter emits `"traceId": ""` on log records, so `log.traceId ?? getStringAttribute(...)` at `src/tracing/otlpReceiver.ts:768` never reaches the resource-attribute fallback this PR enables — in my E2E run on this branch, all 22 codex log records were still dropped despite correct `promptfoo.*` attrs in the resource. One-line fix verified working (22/22 records become correctly-parented spans): treat empty/all-zero inline trace ids as absent, mirroring the existing `isZeroSpanId` guard on the span-id path. Recommend shipping it in this PR with a codex-shaped regression test (`traceId: ""` + resource attrs set), otherwise the fix this PR describes can't be observed working.
2. Hardcoded `promptfoo.trace_id`/`promptfoo.parent_span_id` literals vs the documented single-source-of-truth constants in `src/tracing/resourceAttributes.ts` (see inline thread for the ratchet-neutral `genaiTracer.ts` option).
3. `codex-app-server.ts` parity (same codex binary, same `deep_tracing`, none of these fixes) — fine as a tracked follow-up if out of scope here.
4. Docs: codex export requires a `CODEX_HOME` `config.toml` `[otel]` exporter with an exact endpoint including `/v1/logs`; the deep-tracing section doesn't mention this or the new attr injection, so the feature is hard to discover/use.

CI is green on the new commit.

src/providers/openai/codex-sdk.ts
mldangelo: Resolved by 724a16c — verified end-to-end on the unmodified tip: ran the same eval (codex `[otel]` exporter → receiver on 4318), **zero** dropped records this time; all 22 codex log records persisted as spans under the evaluation trace (`codex.conversation_starts`, `codex.user_prompt`, `codex.websocket_*`, …), every one parented to the provider span via `promptfoo.parent_span_id`. The `hasValidInlineTraceId` guard mirrors the span-id path as suggested, and catching the base64-encoded zero form (`A{22}==`) goes beyond what I'd flagged — nice. The three-form regression test (`''`, hex zeros, base64 zeros) with a codex-shaped payload is exactly the right pin.

src/providers/openai/codex-sdk.ts
mldangelo: Resolved by 724a16c — the helper now uses `PROMPTFOO_RESOURCE_ATTR_TRACE_ID`/`PROMPTFOO_RESOURCE_ATTR_PARENT_SPAN_ID`, so producer/consumer key drift is no longer possible. Re-exporting the constants through `genaiTracer` is a pragmatic ratchet-neutral route; worth knowing the repo guideline says to avoid re-exports (AGENTS.md: "import directly from the source module") — if anyone objects later, the baseline bump for a direct `resourceAttributes` import is the sanctioned alternative. The remaining duplication of the helper *body* vs `claude-agent-sdk.ts:101` is now low-risk and fine as an optional follow-up consolidation (e.g. alongside a codex-app-server port). Architecture check passes at the tip.

mldangelo (approved):
Re-reviewed at 724a16c (`fix(codex): ingest logs without inline trace ids`). This commit resolves the end-to-end blocker and the remaining important items from my review, and I've verified it works for real.

## Verification on the unmodified tip

- **E2E (the thing that was broken):** same deep-tracing eval as before (`openai:codex-sdk` + `deep_tracing: true`, codex `[otel]` logs exporter pointed at the receiver). Result: **0 dropped records** (was 22/22 dropped); all 22 codex log records persisted as spans under the evaluation trace — `codex.conversation_starts`, `codex.user_prompt`, `codex.websocket_*` — each correctly parented to the provider span via `promptfoo.parent_span_id`, with the eval passing. The hierarchy in the trace store is evaluator span → `chat gpt-5.2` provider span → codex log spans, exactly as intended.
- **Tests:** 450/450 across `otlpReceiver`, `openai-codex-sdk`, `claude-agent-sdk`, and `test-hygiene` suites.
- **Architecture:** `npm run architecture:check` passes (the constants come in via the existing `genaiTracer` import, so the edge ratchet is untouched).
- CI was 36 pass / 3 still running at review time; prior commits were fully green and this one touches the same areas the suites above cover.

## What this commit fixed (mapped to the review)

1. ✅ **Receiver empty/zero inline `traceId`** — `hasValidInlineTraceId` mirrors the existing span-id guard, and additionally catches the base64-encoded zero form, which is more thorough than what I'd flagged. Regression test covers all three shapes (`""`, hex zeros, base64 zeros) with a codex-shaped payload.
2. ✅ **Hardcoded resource-attr keys** — now the shared `PROMPTFOO_RESOURCE_ATTR_*` constants; drift between producer and consumer is no longer possible.
3. ✅ **Docs** — the Deep Tracing section now documents the `CODEX_HOME` `config.toml` `[otel]` exporter requirement, the exact-`/v1/logs`-endpoint semantics, and the resource-attribute injection (with a sensible `log_user_prompt = false` default in the example).
4. ✅ (earlier commits) zero-traceparent guard + precedence/rejection tests.

## Remaining non-blockers (fine as follow-ups)

- `codex-app-server.ts` parity: same codex binary, same `deep_tracing`, still no guard/fallback/resource attrs (`:1267`, `:1445-1463`) — its CLI logs now *would* link if it injected the attrs, since the receiver fix is shared. Worth a tracked follow-up.
- The helper body is still duplicated vs `claude-agent-sdk.ts:101` (harmless now that the keys are shared); a future consolidation could also tighten claude's weaker zero-trace-only guard.
- `src/tracing/resourceAttributes.ts:3-4` header still names only claude-agent-sdk as the producer; the receiver comment was updated but this one is now stale.
- Optional: a cross-provider "log-derived spans" section in `site/docs/tracing.md` so the contract has one canonical home.

Good work turning this around — the fix is now observable working end to end.

src/providers/openai/codex-sdk.ts
```diff
@@ -2082,7 +2127,10 @@ export class OpenAICodexSDKProvider implements ApiProvider {
 
     // Get current trace context for deep tracing
     // This allows the Codex CLI to export its internal spans as children of our span
-    const currentTraceparent = getTraceparent();
+    const activeTraceparent = getTraceparent();
+    const currentTraceparent = isValidTraceparent(activeTraceparent)
```
mldangelo-oai, line 2131: This is the right precedence boundary. When `withGenAISpan` has installed a real provider span, its context should parent the CLI telemetry; when the no-op tracer exposes all-zero IDs, falling back to the evaluator context preserves linkage instead of exporting a poisoned `TRACEPARENT`. The added tests cover both valid-active-wins and invalid-active-falls-back, so a future precedence regression should be caught rather than silently dropping telemetry.

src/tracing/otlpReceiver.ts
```diff
@@ -762,11 +769,13 @@ export class OTLPReceiver {
...
     // TRACEPARENT into its logs context). Fall back to the resource attribute
-    // promptfoo.trace_id — the claude-agent-sdk provider injects this via
-    // OTEL_RESOURCE_ATTRIBUTES because Claude Code's logs signal does not
-    // inherit TRACEPARENT. Without either, we can't link anywhere useful.
-    const rawTraceId =
-      log.traceId ?? getStringAttribute(resourceAttributes, PROMPTFOO_RESOURCE_ATTR_TRACE_ID);
+    // promptfoo.trace_id — agent providers inject this via
+    // OTEL_RESOURCE_ATTRIBUTES because some log signals do not inherit
+    // TRACEPARENT. Without either, we can't link anywhere useful.
+    const hasValidInlineTraceId = !!log.traceId && !isZeroTraceId(log.traceId);
```
mldangelo-oai, line 775: This closes the actual end-to-end gap. Codex 0.130 emits an empty inline `traceId` for log-only records, so the old nullish-coalescing fallback never reached `promptfoo.trace_id` even though the resource attribute was present. Treating empty/all-zero hex/base64 IDs as absent preserves valid inline context while allowing the injected resource context to recover these records; mirroring the existing span-ID fallback keeps the synthesized spans correctly parented.

site/docs/providers/openai-codex-sdk.md
```diff
@@ -458,7 +458,17 @@ providers:
       enable_streaming: true
 ```
 
-Deep tracing injects OpenTelemetry environment variables (`OTEL_EXPORTER_OTLP_ENDPOINT`, `TRACEPARENT`, etc.) into the Codex CLI process. Promptfoo uses a fresh SDK client/thread per call in this mode so child spans link to the correct parent request span.
+Codex configures its log exporter through the `config.toml` in `CODEX_HOME`. Point it at Promptfoo's JSON logs receiver using the complete `/v1/logs` endpoint:
```
mldangelo-oai, line 461: This operational detail is important and matches the pinned Codex implementation: the `[otel].exporter` setting controls the logs exporter, and its HTTP endpoint is signal-specific. Injecting `OTEL_*` variables alone does not enable export, while configuring only the host/port does not produce the required `/v1/logs` request. Calling out both `CODEX_HOME` and the complete endpoint prevents a setup that looks plausible but emits nothing.

mldangelo-oai (approved):
LGTM — reviewed and approving at `724a16c7d37c6686900cd56ff73f75a70d42c457`.

I traced the current change from provider-span creation, through the spawned Codex CLI environment, to `/v1/logs` ingestion, and reconciled the earlier review threads against this exact head.

## What looks right

- **Trace identity selection:** a valid active provider span wins, while no-op all-zero trace/span IDs fall back to the evaluator-supplied trace context. The positive-precedence test and the two invalid-active cases pin both sides of that decision.
- **Resource attributes:** `promptfoo.trace_id` and `promptfoo.parent_span_id` are derived from the selected W3C context, stale Promptfoo values are replaced, and unrelated user resource attributes are preserved. Using the shared constants removes the producer/consumer drift risk from the first revision.
- **Receiver behavior:** valid inline OTLP context still wins. Empty, all-zero hexadecimal, and all-zero base64 trace IDs are now treated as absent so Codex log-only records can use the injected resource context. The existing span-ID fallback then parents each synthesized log span to the provider span, while fresh span IDs avoid collisions across multiple records.
- **Operational docs:** the page now explains the crucial distinction between injecting standard OTEL environment variables and actually enabling Codex's log exporter. The pinned [Codex 0.130.0 configuration types](https://github.com/openai/codex/blob/rust-v0.130.0/codex-rs/config/src/types.rs#L475-L516) and [HTTP log exporter construction](https://github.com/openai/codex/blob/rust-v0.130.0/codex-rs/otel/src/provider.rs#L332-L359) agree with the documented `json` protocol and complete `/v1/logs` endpoint.

## Verification

- The exact head has **43/43 successful checks**.
- The Node 22 lane ran `test/providers/openai-codex-sdk.test.ts` (**150 tests**) and `test/tracing/otlpReceiver.test.ts` (**50 tests**) as part of **19,840 passing tests**.
- Docs build, package builds, formatting, architecture checks, smoke/integration tests, CodeQL, and patch/project coverage all pass; Codecov reports every modified coverable line covered.
- I reconciled the existing [exact-head end-to-end run](https://github.com/promptfoo/promptfoo/pull/9703#pullrequestreview-4481365472): the pre-fix run dropped all 22 Codex log records, while this head persisted all 22 under the evaluation trace with the provider span as parent. That directly exercises the behavior this PR claims to fix.

## Non-blocking follow-ups

1. Validate `context?.traceparent` before using it as the fallback, for defense in depth when the provider is invoked outside the normal evaluator path.
2. Port the same propagation to `openai:codex-app-server` and consolidate the Claude/Codex helper at a shared boundary. That would also be a natural place to remove the pragmatic `genaiTracer` constant re-export and return to direct imports.

Neither follow-up changes the correctness of this SDK-focused fix. The original end-to-end blocker is resolved, regression coverage is strong, and the feature is now documented in a form users can actually configure. LGTM.

PR #9707, opened 2026-06-12
feat(providers): add pi coding agent provider

mldangelo:
## Summary

Adds a provider for [Pi](https://pi.dev), a minimal terminal coding agent with multi-provider model support (Anthropic, OpenAI, Google, Groq, OpenRouter, and more). Pi is the agent harness that powers OpenClaw.

Promptfoo spawns the `pi` CLI per call in one-shot JSON event-stream mode (`pi --mode json --no-session`), so evals exercise the same runtime users get in their terminal. No npm dependency is added: the provider resolves the CLI from `pi_path`, a project-local `@earendil-works/pi-coding-agent` install, or `PATH`, and fails with actionable install guidance when missing.

### Provider IDs

- `pi` — pi's configured default model
- `pi:<provider>/<model>` — explicit model pattern, e.g. `pi:anthropic/claude-sonnet-4-5`, `pi:openai/gpt-5.4-mini:high`

### Design decisions

- **CLI subprocess over in-process SDK.** Pi's SDK is ESM-only with restrictive exports, requires Node >= 22.19, and ships `undici@8` — embedding it in-process risks global dispatcher/state contamination. Process isolation also means evals measure the real product surface (matches the `codex app-server` precedent).
- **OpenCode-style safety defaults.** Pi has no built-in permission or sandbox system, so: no `working_dir` → temp dir with all tools disabled (chat-only); with `working_dir` → read-only tools (`read`, `grep`, `find`, `ls`); `bash`/`edit`/`write` are explicit opt-in.
- **Hermetic by default.** `--offline`, `--no-session`, and extension/skill/prompt-template/context-file discovery disabled so eval prompts pass through verbatim and results are reproducible; each is re-enableable to evaluate a customized pi setup.
- **Failure detection from `stopReason`.** Pi exits 0 in JSON mode even when the agent run fails; the provider reads the final assistant message's `stopReason`/`errorMessage`. Nonzero/signal exits are never reported (or cached) as success.
- **Credential hygiene.** `apiKey` is injected via the provider's standard env var (or `api_key_env` for unrecognized providers) — never argv — keeping secrets out of `ps` output, debug logs, and the persisted cache key (regression-tested).
- **Hang resistance.** The run settles from process `exit` with a 1s stdio flush grace, so a descendant process holding the inherited stdout pipe can't hang the eval row; stdout uses stream encoding so multi-byte UTF-8 across chunk boundaries survives.

### Response mapping

`output` = final assistant text; `tokenUsage` sums every assistant turn (`prompt`/`completion`/`total`/`cached`/`numRequests`); `cost` = USD total reported by pi; `metadata.toolCalls` = tool activity with args and error status; `raw` = all assistant messages. `pi` is also registered as an agentic provider for the `agent-rubric` assertion.

## Files

- `src/providers/pi.ts` — provider implementation
- `src/providers/registry.ts` — `pi`/`pi:` factory (lazy import)
- `src/providers/agentic-utils.ts` — agentic registration + new shared `validateAgenticWorkingDir` helper (also adopted by `opencode-sdk.ts`, which previously had a verbatim copy of the same block)
- `src/matchers/agent.ts` — agent-rubric grader list
- `test/providers/pi.test.ts` — 45 unit tests (args, env, parsing, errors, abort/timeout with fake timers, caching incl. apiKey-exclusion regression)
- `site/docs/providers/pi.md` + index/guide/agent-rubric/taxonomy entries
- `examples/provider-pi/` (basic, working-dir, model-comparison) + init aliases
- `examples/compare-agentic-sdks/` — pi added to the cross-SDK comparison
- `architecture/edge-baseline.json` — +2 each on `providers -> legacy-contracts`/`legacy-runtime` for the new provider's standard imports; also locks in a pre-existing `node -> legacy-contracts` reduction (102 → 97) the checker flagged

## QA (live, with real API keys)

| Scenario | Result |
| --- | --- |
| `basic` example (`pi:openai/gpt-5.4-mini`, chat-only) | ✅ pass; usage, cost, metadata correct |
| `working-dir` example (read-only tools) | ✅ pass; `read` tool call captured in `metadata.toolCalls`, multi-turn usage summed |
| `model-comparison` (OpenAI + Anthropic + Gemini through pi) | ✅ 3/3 pass with per-model cost |
| Bare `pi` provider id | ✅ resolves pi's configured default model |
| `agent-rubric` with pi as the grading provider | ✅ pass |
| Cache behavior | ✅ second identical run served from cache (`cached: true`, 2ms) |
| `config.apiKey` env injection with ambient key removed | ✅ pass |
| Nonexistent model | ✅ graceful row error with upstream API message |
| Missing `pi` binary | ✅ actionable install guidance |

### Head-to-head: pi vs Claude Agent SDK (same model, `claude-sonnet-4-6`, security-audit task)

| | Claude Agent SDK | Pi |
| --- | --- | --- |
| Found planted issues | 4/4 ✅ | 4/4 ✅ |
| Latency | 35.6s | 60.4s (4 turns) |
| Cost | $0.166 | $0.067 |
| Tools used | Glob, Read, Read | find, ls, read, read |

Both harnesses are equally effective on the task; the Claude Agent SDK batches exploration into fewer round trips, while pi's minimal system prompt and prompt-cache reuse make it ~2.5× cheaper.

## Review hardening

A multi-agent adversarial review pre-merge confirmed and fixed 11 issues, including a P1 where an `--api-key` argv fallback would have leaked the raw key into the persisted cache key, `ps` output, and debug logs (replaced with env-var injection + `api_key_env`), a UTF-8 chunk-boundary decode bug, crash-exits with partial events being cached as success, and a hang when a descendant process holds stdout open. A follow-up comment-accuracy audit and simplification pass deduplicated copied blocks and corrected misleading comments.

`npm run tsc` clean; provider/matcher/command suites pass; Biome + Prettier clean; `architecture:check` clean.


docs/agents/coding-agent-provider-taxonomy.md
```diff
@@ -161,8 +162,10 @@ The coding-agent providers already share several practical patterns:
 - Session or thread caches are keyed by provider config.
 - Provider-level config is stricter than prompt-level merged config.
 - Tool and skill usage are surfaced through metadata where possible.
-- Tracing is supported for Codex and is partially shared through OpenAI agent
-  tracing helpers.
+- A top-level GenAI `callApi` span is shared across providers via `withGenAISpan`
```
jameshiester-oai, line 165: Can you clarify this?

examples/compare-agentic-sdks/promptfooconfig.yaml
```diff
@@ -60,6 +60,12 @@ providers:
       model: claude-sonnet-4-6
       working_dir: ./test-codebase
 
+  # Pi coding agent with read-only file tools (provider-agnostic)
+  - id: pi:anthropic/claude-sonnet-4-6
```
jameshiester-oai, line 64: should this be openai example?

test/providers/registry.test.ts
```diff
@@ -1264,4 +1264,64 @@ describe('Provider Registry', () => {
...
+  describe('pi: prefix routing', () => {
+    const bareOptions: ProviderOptions = { config: {} };
+    const bareContext: LoadApiProviderContext = {
+      basePath: '/test',
+      options: bareOptions,
+    };
+
+    it.each([
+      'pi',
+      'pi:anthropic/claude-sonnet-4-5',
```
jameshiester-oai, line 1277: is this mention necessary?

jameshiester-oai (approved):

src/providers/pi.ts
```diff
@@ -0,0 +1,1299 @@
...
+   */
+  private cacheEnv(config: PiProviderConfig): Record<string, string> {
+    const merged = this.mergeConfiguredEnv(config);
+    const credentialVar = this.getApiKeyEnvVar(config);
+    // Canonicalize (sort keys) and drop the credential var plus any secret-named
+    // var so semantically identical, credential-independent envs hash the same.
+    // Surviving values are sanitized to strip credentials embedded in URLs.
+    const sorted: Record<string, string> = {};
+    for (const key of Object.keys(merged).sort()) {
+      if (key === credentialVar || SECRET_ENV_NAME_PATTERN.test(key)) {
```
jameshiester-oai, line 752: **[P2] Keep non-secret auth controls in the cache identity**

This broad `auth` match removes `AWS_BEDROCK_SKIP_AUTH`, which is a behavior flag rather than a credential. A direct probe produced identical cache inputs for values `0` and `1`, while Pi 0.79.7 reads this variable to choose whether Bedrock performs its normal authentication path. As a result, changing the flag can return the previous configuration's cached response without running Pi. Use an explicit credential-name classifier (or preserve this known control flag) and add a collision regression test.

src/providers/pi.ts
```diff
@@ -0,0 +1,1299 @@
...
+      // Signal pi directly (so the mockable child.kill path stays exercised) and,
+      // on POSIX, the whole process group via the negative pid so descendant
+      // tool processes are terminated too.
+      const signalGroup = (signal: NodeJS.Signals) => {
+        try {
+          child.kill(signal);
+        } catch (err) {
+          logger.debug(`[Pi] Failed to send ${signal}: ${err}`);
+        }
+        if (process.platform !== 'win32' && typeof child.pid === 'number') {
```
jameshiester-oai, line 857: **[P2] Terminate the Pi process tree on Windows**

On Windows `detached` is disabled and this branch only calls `child.kill()` on the Pi process; Node does not recursively terminate its descendants. If Pi has launched a bash/tool child, timeout, abort, or output-overflow can therefore return while that child keeps running and retains ports, files, or other side effects. Add a Windows tree-termination path (for example a Job Object or an appropriately controlled `taskkill /T`) and a fake-launcher regression that verifies a grandchild is gone before the call settles.

src/providers/pi.ts
```diff
@@ -0,0 +1,1299 @@
...
+    }
+    if (finalMessage.model) {
+      metadata.model = finalMessage.model;
+    }
+    if (toolCalls.length > 0) {
+      metadata.toolCalls = toolCalls;
+    }
+
+    return {
+      output: this.getMessageText(finalMessage),
```
jameshiester-oai, line 1143: **[P2] Preserve Pi's stop reason in the provider response**

Successful Pi messages can end with `stopReason: "length"`, `"stop"`, or `"toolUse"`, but this response drops the field. I reproduced this through the built CLI with a fake Pi that emitted `"length"`: Promptfoo's `finish-reason: length` assertion failed with `Provider did not supply stop/finish reason`, so truncated output is reported like an ordinary success. Populate `ProviderResponse.finishReason` here, mapping Pi's values to Promptfoo's normalized reasons (including `toolUse` → `tool_calls`), and cover all three successful stop reasons.

jameshiester-oai (commented):
## Review summary

**Reviewed PR base:** `main` at `051fbf95bc3a460c79c5d2252a73484f16b65f12`  
**Current `origin/main` interaction context:** `56ee73063bb5088536e4d510861651484fc49a3a`  
**Reviewed head:** `d40abe57e9df3a21ea101b83c9e50691a029d6d8`

**Scope:** Full, fresh review of the current PR; there was no prior authenticated completion marker for this head. I reviewed the complete merge-base-to-head diff and the current base for interaction context.

**Verdict:** I found three verified P2 correctness/operability defects and left inline comments. All current CI checks are green, but GitHub reports the PR as conflicting with `main`.

### Verified findings

- **P2 — Pi stop reasons are discarded on successful responses.** A built-CLI run where Pi emitted `stopReason: "length"` failed Promptfoo's `finish-reason: length` assertion with `Provider did not supply stop/finish reason`; truncated output is therefore indistinguishable from an ordinary completion.
- **P2 — `AWS_BEDROCK_SKIP_AUTH` is incorrectly removed from the cache identity.** Values `0` and `1` produced the same cache environment even though Pi uses that variable to change Bedrock authentication behavior, so one configuration can reuse the other's response.
- **P2 — timeout/abort cleanup does not terminate the process tree on Windows.** The Windows path only calls `child.kill()` on Pi itself; tool descendants can continue running after Promptfoo reports cancellation.

### Existing feedback reconciled

I read all 57 existing review threads and did not duplicate still-applicable findings. Important current-head feedback already present includes explicit `apiKey` losing to stored/OAuth credentials, Node/API `basePath` resolution, local Pi under Promptfoo's supported Node 20 runtime, missing provider env schema entries, caching of side-effecting tool runs, ignored `responseModel`, the async cache/abort race, delayed SIGKILL cleanup, and tilde handling for agent directories.

Two prior threads remain only partly addressed:

- The background-process thread now kills Pi's own POSIX group on the inherited-stdio fallback, but Pi 0.79.7 launches bash tools in separate detached groups. Fresh success and 100 ms timeout probes both returned while a redirected background grandchild was still alive. This is the same lifecycle issue as the prior thread, so I did not add a duplicate inline comment.
- The argv logging thread gained credential-flag redaction, but configured `system_prompt` and `append_system_prompt` values are still written verbatim to the automatically-created debug log. A local sentinel run confirmed both values in the log. This is an incomplete form of the existing logging issue, so I did not duplicate it inline.

I also rechecked stale claims: Pi 0.79.7 supports `--offline`; current `tool_execution_end` events include `toolName`; the already-aborted cache path and cache-write usage accounting are fixed on this head.

- Replace the broad secret-name regex used for cache identity with an explicit credential classifier so behavior flags cannot be silently dropped.
- Centralize child-process containment and teardown rather than mixing direct-child and POSIX-group assumptions in the provider.
- Add provider-response contract tests for Pi's `stop`, `length`, and `toolUse` reasons, including Promptfoo's `finish-reason` assertion.

### Tests and QA

- `source ~/.nvm/nvm.sh && nvm use && env -i HOME="$HOME" PATH="$PATH" TMPDIR=/private/tmp CI=1 npm ci` — exit 0; lockfile-pinned dependencies restored, worktree unchanged.
- `... vitest run test/providers/pi.test.ts test/providers/agentic-utils.test.ts test/providers/registry.test.ts test/matchers/agent-rubric.test.ts test/commands/init.test.ts` — exit 0; 5 files, 235 tests passed. Independent focused runs also passed 231 and 239 tests.
- `npm run tsc -- --pretty false` — exit 0.
- `npm run build` — exit 0, including the app build and postbuild.
- `npm --prefix site run typecheck` — exit 0.
- `SKIP_OG_GENERATION=true npm --prefix site run build` — exit 0.
- `npm run l` — exit 0; 22 pre-existing complexity warnings, no fixes applied.
- `git diff --check` — exit 0; final worktree clean.
- Built CLI with an isolated fake Pi and `--no-cache`: normal `contains` assertion — exit 0, 1/1 passed; exported JSON inspected.
- The same built CLI with Pi emitting `stopReason: "length"` and `finish-reason: length` — exit 100 as expected, 0/1 passed with `Provider did not supply stop/finish reason`; exported response contained the raw Pi stop reason but no `finishReason`.
- Direct cache-identity probe — `AWS_BEDROCK_SKIP_AUTH=0` and `=1` produced identical inputs.
- Detached-grandchild probes — successful completion and timeout both left the grandchild alive until explicit test cleanup.
- System-prompt logging probe — both non-secret sentinel values were present in the per-run debug log.

### Limitations

No live model/API call, real credential flow, Windows runtime execution, or interactive browser QA was performed. Pi CLI contracts and process behavior were checked against the official 0.79.7 package source. The requested `/pr-review-toolkit:review-pr all parallel` command is not installed in this environment (exit 127), so I completed the six requested review lenses manually and reconciled three independent read-only specialist reviews.

PR #9717, opened 2026-06-14
fix(assertions): correct BLEU brevity penalty formula

i-anubhav-anand:
## What
The BLEU brevity penalty is computed with the candidate/reference ratio inverted.

```diff
- return Math.exp(1 - candidateLength / referenceLength);
+ return Math.exp(1 - referenceLength / candidateLength);
```

## Why
Per the BLEU definition (Papineni et al., 2002), for `candidateLength <= referenceLength` the brevity penalty is `exp(1 - r/c)` (where `c` = candidate length, `r` = reference length). The current code computes `exp(1 - c/r)`, which is the inverse.

Because `c/r <= 1` when the candidate is shorter, `exp(1 - c/r) >= 1` — so short candidates were **rewarded** instead of **penalized**, and `calculateBleuScore` could return values **above 1.0**, even though BLEU is defined on `[0, 1]`.

**Example (before fix):** a 5-token candidate whose every 1–4 gram appears in a 7-token reference has perfect n-gram precision, so the score equals the brevity penalty: `exp(1 - 5/7) ≈ 1.33`. After the fix it is correctly `exp(1 - 7/5) ≈ 0.67`.

## Test
Added a regression test (`should penalize short candidates and never exceed 1.0`) asserting the score is `<= 1` and equals `exp(1 - r/c)`.

## Validation
- `npx vitest run test/assertions/bleu.test.ts` → **18 passed** (17 existing + 1 new)
- `npm run l` (Biome) and `npm run f` clean · `tsc --noEmit` clean
- No other test referenced the brevity penalty; lockfile drift reverted (diff is the 1-line fix + test only).

### DIFF
--- a/src/assertions/bleu.ts
+++ b/src/assertions/bleu.ts
@@ -31,6 +31,28 @@ function calculateBrevityPenalty(candidateLength: number, referenceLength: numbe
 }
 
 /**
+ * Tokenizes text into lowercased, whitespace-delimited words.
+ *
+ * @internal
+ */
+function tokenize(text: string): string[] {
+  return text.toLowerCase().trim().split(/\s+/);
+}
+
+/**
+ * Counts how many times each n-gram occurs.
+ *
+ * @internal
+ */
+function countNGrams(ngrams: string[]): Map<string, number> {
+  const counts = new Map<string, number>();
+  for (const gram of ngrams) {
+    counts.set(gram, (counts.get(gram) ?? 0) + 1);
+  }
+  return counts;
+}
+
+/**
  * Calculates BLEU score for a candidate string against reference strings.
  *
  * @param candidate - The string to evaluate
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
LGTM ✅

Great catch, @i-anubhav-anand — the brevity penalty was genuinely inverted. `exp(1 - candidateLength/referenceLength)` rewarded short candidates and let BLEU exceed 1.0, whereas the Papineni et al. (2002) definition (and NLTK/sacreBLEU) is `exp(1 - referenceLength/candidateLength)` for `c <= r`. Your fix and regression test are spot on.

I pushed two small follow-up commits on top of your work:
- A `handleBleuScore` inverse-path regression test plus a `c == r` boundary guard — the inverse path was the user-visible symptom, since a score > 1 made the reported `1 - score` go negative.
- A behavior-preserving simplification of the n-gram precision loop (tokenize each reference once, hoist the candidate count map), verified bit-for-bit identical to the original across ~5k fuzzed inputs.

CI is green. Thanks for the contribution! 🙏

PR #9718, opened 2026-06-14
fix(assertions): make BLEU closest-reference tie-break order-independent

i-anubhav-anand:
## What
When references are equidistant in length from the candidate, the BLEU score depends on the **order** the references are listed in.

## Why
`calculateBleuScore` selects the brevity penalty's "closest reference length" with:

```ts
const closestRefLength = refLengths.reduce((prev, curr) =>
  Math.abs(curr - candidateWords.length) < Math.abs(prev - candidateWords.length) ? curr : prev,
);
```

On a tie (`abs(curr - c) === abs(prev - c)`) it keeps `prev` (first-seen), so the chosen length depends on reference order. BLEU (Papineni et al.; NLTK `closest_ref_length`) breaks such ties toward the **shorter** reference. Result: e.g. candidate of 4 tokens with references of length 3 and 5 scores differently depending on which is listed first.

## Fix
Prefer the shorter reference length on ties, making the score order-independent:

```ts
const closestRefLength = refLengths.reduce((prev, curr) => {
  const distCurr = Math.abs(curr - candidateWords.length);
  const distPrev = Math.abs(prev - candidateWords.length);
  return distCurr < distPrev || (distCurr === distPrev && curr < prev) ? curr : prev;
});
```

## Test
Adds `should not depend on the order of equidistant references`. Verified it **fails before** the fix and **passes after**.

## Validation
- `npx vitest run test/assertions/bleu.test.ts` → **18 passed**
- `npm run l` (Biome) + `npm run f` clean · `tsc --noEmit` clean · lockfile drift reverted (diff is the two files only)

Independent of #9717 (different lines: that PR fixes the brevity-penalty formula; this fixes the closest-reference selection).

### DIFF
--- a/src/assertions/bleu.ts
+++ b/src/assertions/bleu.ts
@@ -53,11 +53,15 @@ export function calculateBleuScore(
 
   const candidateWords = candidate.toLowerCase().trim().split(/\s+/);
 
-  // Find reference with closest length to candidate
+  // Find reference with closest length to candidate. On ties, prefer the shorter
+  // reference (BLEU / NLTK `closest_ref_length` convention) so the score does not
+  // depend on the order references are provided in.
   const refLengths = references.map((ref) => ref.toLowerCase().trim().split(/\s+/).length);
-  const closestRefLength = refLengths.reduce((prev, curr) =>
-    Math.abs(curr - candidateWords.length) < Math.abs(prev - candidateWords.length) ? curr : prev,
-  );
+  const closestRefLength = refLengths.reduce((prev, curr) => {
+    const distCurr = Math.abs(curr - candidateWords.length);
+    const distPrev = Math.abs(prev - candidateWords.length);
+    return distCurr < distPrev || (distCurr === distPrev && curr < prev) ? curr : prev;
+  });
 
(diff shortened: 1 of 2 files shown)


mldangelo:
Thanks for this, @i-anubhav-anand — merged! 🎉

Really clean fix. Pinning the closest-reference tie-break toward the shorter reference is exactly right, and matching the NLTK `closest_ref_length` / Papineni convention is the correct call. The writeup made it easy to review — the worked tie example and the note that it's independent of #9717 were especially helpful.

A couple of small things happened on the way in, just so you have the full picture:

- **#9717 landed first** and touched the same lines (the brevity-penalty formula), so I merged `main` and resolved the conflict — kept your tie-break logic but sourced the lengths from the already-tokenized reference list so references aren't tokenized twice.
- **Added two follow-up tests:** one that pins the tie-break *direction* (your original test proved order-independence, but would still pass if the tie ever flipped toward the *longer* reference), and one that exercises the same guarantee end-to-end through the `bleu` assertion handler.

Appreciate the careful work and the thorough validation notes — hope to see more contributions from you! 🙏

PR #9721, opened 2026-06-14
fix(assertions): report best similarity score when no array value passes

i-anubhav-anand:
## Summary

The multi-value `similar` assertion (`renderedValue` is an array) passes if the output matches **any** of the provided values — it returns on the first `result.pass`. But when **none** pass, it reported `Math.min` of the per-value scores, i.e. the *worst*-aligned reference:

```ts
let minScore = Number.POSITIVE_INFINITY;
...
if (result.score < minScore) { minScore = result.score; }
...
return { pass: false, score: minScore, ... };
```

`matchesSimilarity` normalizes `score` so that **higher is always better** — including for `inverse` assertions and the `euclidean` metric (`score = inverse ? 1 - x : x`). So for an "any-of" match, the failure score should be the **maximum**: how close the output got to its *closest* value. Reporting the minimum is misleading (e.g. the output is 0.6 similar to one reference and 0.2 to another, but the report shows 0.2).

This also aligns `similar` with the other multi-reference metrics — `bleu`, `gleu`, and `meteor` all take `Math.max` across references.

## Fix

Track and return the maximum score instead of the minimum. No change to the pass/fail decision or the single-value path.

## Testing

- Added a test: two array values, neither passing, scoring 0.2 and 0.6 — asserts the reported score is `0.6` (the best attempt). Verified it **fails before** (returns 0.2) and **passes after**.
- `npx vitest run test/assertions/similar.test.ts` → 10 passed.
- `biome check` and `npm run tsc` clean.

### DIFF
--- a/src/assertions/similar.ts
+++ b/src/assertions/similar.ts
@@ -14,6 +14,10 @@ export const handleSimilar = async ({
     typeof renderedValue === 'string' || Array.isArray(renderedValue),
     'Similarity assertion type must have a string or array of strings value',
   );
+  invariant(
+    !Array.isArray(renderedValue) || renderedValue.length > 0,
+    'Similarity assertion must have at least one value to compare against',
+  );
   const threshold = assertion.threshold ?? 0.75;
 
   // Parse metric from assertion type (e.g., 'similar:dot' -> 'dot_product')
--- a/src/assertions/similar.ts
+++ b/src/assertions/similar.ts
@@ -36,7 +36,12 @@ export const handleSimilar = async ({
   }
 
   if (Array.isArray(renderedValue)) {
-    let minScore = Number.POSITIVE_INFINITY;
+    // The assertion passes if the output matches ANY of the values (we return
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
LGTM ✅

Nice fix, @i-anubhav-anand. For an "any-of" array match, the failure score should reflect how close the output got to its *closest* value, so `Math.max` is right — `matchesSimilarity` normalizes the score so higher is always closer-to-passing across all metric × inverse combinations (cosine/dot/euclidean), and this brings `similar` in line with `bleu`/`gleu`/`meteor`. I verified all four quadrants and confirmed the old `min` genuinely reported the worst-aligned reference.

I pushed one small follow-up: a guard that rejects an empty array `value: []` (it previously fell through the loop to a degenerate ±Infinity score). That makes the bleu/gleu/meteor parity your comment cites fully hold — those families already throw on empty references.

CI green, tests pass. Thanks for the contribution! 🙏

PR #9725, opened 2026-06-14
fix(assertions): honor inverse (not-) for cost, latency, and perplexity

i-anubhav-anand:
## Summary
The `inverse` flag (derived from a `not-` prefix via `isAssertionInverse`) is passed to every assertion handler, but `handleCost`, `handleLatency`, `handlePerplexity`, and `handlePerplexityScore` ignored it. As a result `not-cost`, `not-latency`, `not-perplexity`, and `not-perplexity-score` behaved **identically** to their non-inverse counterparts — the negation was silently a no-op.

This completes the gap called out in the `levenshtein` fix (#9722), which fixed the same pattern there and noted cost/latency/perplexity as the remaining cases.

## Fix
Apply `!== inverse` to the pass decision and reflect the direction in the reason message:
- `cost` / `latency`: `pass = (value <= threshold) !== inverse`
- `perplexity` / `perplexity-score`: same, but only when a `threshold` is set (no-threshold behavior is unchanged). `perplexity-score` keeps returning its normalized score.

## Testing
- New `test/assertions/{cost,latency,perplexity}.test.ts` with base + inverse cases. Verified the inverse tests **fail before** this change (the `not-` variant wrongly passed/failed) and **pass after** — 13 passed.
- `npm run tsc` and Biome clean.

### DIFF
--- a/src/assertions/cost.ts
+++ b/src/assertions/cost.ts
@@ -1,6 +1,6 @@
 import type { AssertionParams, GradingResult } from '../types/index';
 
-export const handleCost = ({ cost, assertion }: AssertionParams): GradingResult => {
+export const handleCost = ({ cost, assertion, inverse }: AssertionParams): GradingResult => {
   if (assertion.threshold === undefined) {
     throw new Error('Cost assertion must have a threshold');
   }
@@ -8,13 +8,15 @@ export const handleCost = ({ cost, assertion }: AssertionParams): GradingResult 
     throw new Error('Cost assertion does not support providers that do not return cost');
   }
 
-  const pass = cost <= assertion.threshold;
+  const pass = cost <= assertion.threshold !== inverse;
   return {
     pass,
     score: pass ? 1 : 0,
     reason: pass
       ? 'Assertion passed'
-      : `Cost ${cost.toPrecision(2)} is greater than threshold ${assertion.threshold}`,
+      : `Cost ${cost.toPrecision(2)} is ${
+          inverse ? 'less than or equal to' : 'greater than'
+        } threshold ${assertion.threshold}`,
     assertion,
   };
 };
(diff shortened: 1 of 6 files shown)


mldangelo (approved):
Thanks so much for this fix, @i-anubhav-anand — and for the clear write-up showing the inverse tests fail before / pass after. This is exactly the kind of well-scoped, well-tested correctness fix we love to see. 🙏

`not-cost`, `not-latency`, `not-perplexity`, and `not-perplexity-score` really were silent no-ops, and threading `inverse` into each handler (consuming the centrally-derived `AssertionParams.inverse` rather than re-parsing the prefix) is the right approach. Verified end-to-end: the full dispatch path (`not-` → `isAssertionInverse` → handler) behaves correctly.

I pushed one follow-up commit to land alongside your work, which also resolves the Codex P2 on `perplexity.ts`:

- **`not-perplexity-score` now inverts its graded score** (`1 - perplexityNorm`), matching the sibling graded assertions (`bleu`/`gleu`/`rouge`/`similarity`). Previously it inverted only the *pass* decision while returning the raw score — and since `assertionsResult` overrides pass/fail with the aggregate score when a test/assertion-set threshold is set, a passing `not-perplexity-score` was contributing a *low* score (and a failing one a *high* score), skewing aggregate tests.
- **Hardened the inverse score test** — the original used a symmetric input (`logProbs [0,0]` → norm 0.5, its own complement), which couldn't distinguish the inverted score from the raw one; switched to an asymmetric input so it actually guards the regression.
- Added threshold-boundary cases for `not-cost` / `not-latency` / `not-perplexity`.

All green: 25/25 unit tests, `tsc`, Biome, full CI, plus an end-to-end aggregate-scoring check. A code-simplification pass found the code already idiomatic for this layer (every sibling inlines the same idiom, so no helper extraction warranted).

Approving — thanks again for contributing! 🎉

PR #9734, opened 2026-06-15
fix(matchers): segment prose context into sentences for RAGAS context relevance

i-anubhav-anand:
## Summary

`context-relevance` (RAGAS Context Relevance) scores how much of the retrieved context is actually relevant to the query: an LLM extracts the relevant sentences, and the score is `relevant_sentences / total_context_sentences`.

The denominator is computed by splitting the context into "sentences", but for a **string** context that split is done on **newlines only**:

```ts
// src/matchers/shared.ts
export function splitIntoSentences(text: string) {
  return text.split('\n').filter((sentence) => sentence.trim() !== '');
}
```

A retrieved RAG context is almost always a **single prose paragraph with no newlines**. Splitting it on newlines yields exactly **one** unit, so `totalContextUnits === 1`, the numerator is clamped to that, and the score collapses to **~1.0 no matter how little of the context is relevant**.

This defeats the whole point of the metric: a context that is 90% irrelevant still scores 1.0.

### Concrete example

```
query:   "What is the capital of France?"
context: "Paris is the capital of France. France is in Europe. The weather is nice today."
```

The grader correctly extracts the single relevant sentence (`"Paris is the capital of France."`). Expected RAGAS score: `1/3 ≈ 0.33`.

| | `totalContextUnits` | score |
|---|---|---|
| **Before** | 1 (whole paragraph) | **1.00** ❌ |
| **After**  | 3 (sentences)       | **0.33** ✅ |

## Fix

Add `splitTextIntoSentences` in `shared.ts`. It segments a **single prose block** on sentence boundaries (`.`, `!`, `?` + whitespace), but still treats newline-separated input as **pre-segmented** (one unit per line):

```ts
export function splitTextIntoSentences(text: string): string[] {
  const segments = /\n/.test(text) ? text.split('\n') : text.split(/(?<=[.!?])\s+/);
  return segments.map((s) => s.trim()).filter((s) => s.length > 0);
}
```

Keeping the newline branch is deliberate:
- **Line-formatted grader output** (the LLM returns one extracted sentence per line) stays correctly segmented.
- **Contexts already formatted one sentence/chunk per line** keep their existing unit count.
- It avoids mis-splitting abbreviations (e.g. `"i.e."`, `"U.S."`) in that already-segmented text.

`splitTextIntoSentences` is applied in `matchesContextRelevance` to **both** the context and the grader's extracted sentences, so the numerator and denominator are segmented the same way. **Array (chunk) context behavior is unchanged** — chunks are still counted as-is via `splitIntoSentences`.

> Note: the sentence split is a lightweight heuristic (it won't perfectly handle decimals like `3.14` or every abbreviation); full segmentation would need an NLP tokenizer. It is a large improvement over newline-only splitting for the common single-paragraph case, with no regression for the line-formatted cases.

## Test plan

- Added a regression test, `should segment a single-paragraph prose context into sentences`, that exercises the exact failing case above.
  - **Fails before** the fix: `expected 1 to be close to 0.33` (the old behavior returned `1.0`).
  - **Passes after**: `0.33`.
- All existing `context-relevance` tests still pass, including the line-formatted and formatted-policy cases that rely on newline segmentation.

```
npx vitest run test/matchers/context-relevance.test.ts   # 11 passed
npx vitest run test/matchers/similarity.test.ts          # shared.ts consumers — passed
npm run tsc                                               # clean
npm run l && npm run f                                    # clean
```

## References

- RAGAS context relevance: Es et al., 2023, *RAGAS: Automated Evaluation of Retrieval Augmented Generation* — the metric is defined as the ratio of relevant sentences to total sentences in the retrieved context.

### DIFF
--- a/site/docs/configuration/expected-outputs/model-graded/context-relevance.md
+++ b/site/docs/configuration/expected-outputs/model-graded/context-relevance.md
@@ -97,7 +97,7 @@ assert:
 ## Limitations
 
 - Only identifies minimum sufficient content
-- Single context strings split by lines (use arrays for better accuracy)
+- A single-paragraph (prose) context string is split into sentences on `.`/`!`/`?` boundaries; a context with two or more non-empty lines, or an array of chunks, is treated as already segmented and split by line/chunk. Sentence splitting is a lightweight heuristic that does not handle every abbreviation or decimal edge case — provide an array of chunks for the most precise denominator.
 - Score interpretation varies by use case
 
 ## Related metrics
--- a/src/matchers/rag.ts
+++ b/src/matchers/rag.ts
@@ -19,6 +19,7 @@ import {
   fail,
   normalizeMatcherTokenUsage,
   splitIntoSentences,
+  splitTextIntoSentences,
   tryParse,
 } from './shared';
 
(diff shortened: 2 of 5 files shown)


mldangelo (approved):
LGTM ✅

Real improvement — a single-paragraph prose context was collapsing to one unit and forcing the relevance score to ~`1.0` regardless of how little was actually relevant. Segmenting prose into sentences (and segmenting the grader numerator the **same** way, so numerator and denominator stay comparable) fixes it without breaking the pre-segmented / multi-line / chunk-array paths.

Verified end-to-end with the real grader: a 3-sentence prose context with 1 relevant sentence now scores `0.33` (`units=3`), where it used to be forced to ~`1.0`. 24 unit tests green, and the docs Limitations bullet is updated to match the new behavior.

PR #9736, opened 2026-06-15
fix(assertions): honor a test/assert-set threshold of 0

i-anubhav-anand:
## Summary

A test-level (or `assert-set`) `threshold` is documented to **override** the pass/fail decision: the test passes when the combined weighted assertion score meets the threshold, regardless of individual assertion outcomes. The override is gated by a **truthy** check, so a threshold of **0** is silently dropped and the test falls back to the default "fail if any assertion fails" behavior.

```ts
// src/assertions/assertionsResult.ts
if (this.threshold) {            // ← falsy for 0 → override skipped
  pass = score >= this.threshold;
  ...
}
```

`threshold: 0` is a valid, meaningful config: since aggregate scores are always `>= 0`, it means **"collect the assertion scores but never fail the test on an individual assertion failure"** — the natural way to run a metric-collection / scoring suite where the pass/fail decision is deferred.

This is also **inconsistent within the codebase**:
- The `assert-set` threshold passthrough in this same file already uses `!== undefined` (`assertionsResult.ts:26`).
- The assertion-level path already treats falsy thresholds as meaningful — `test/assertions/javascript.test.ts` has a "should handle various falsy threshold values correctly" test asserting `threshold: 0` + `score: 0` ⇒ `pass: true`.

So an assertion with `threshold: 0` passes, but a **test** with `threshold: 0` does not. This PR aligns them.

## Concrete example

```yaml
tests:
  - threshold: 0           # collect scores; don't fail on an individual assertion
    assert:
      - type: equals
        value: 'Hello world'
        metric: accuracy
      - type: contains
        value: 'world'
```

Output `"Goodbye world"`: `equals` fails (0), `contains` passes (1) → aggregate score `0.5`.

| | override applied? | result |
|---|---|---|
| **Before** | no (`if (0)` is false) | **fail** (the `equals` failure fails the test) ❌ |
| **After**  | yes (`0.5 >= 0`)        | **pass** ✅ |

## Fix

```ts
if (this.threshold !== undefined) {
  pass = score >= this.threshold;
  ...
}
```

A threshold of `0` (and the `null`/`''`/`false` → `0` coercions already documented at the assertion level) now correctly engages the override. Any non-zero threshold is unaffected; an absent threshold (`undefined`) still uses the default all-pass logic. This also fixes the same bug for `assert-set` thresholds, which flow through the same `AssertionsResult.threshold`.

## Test plan

- Added a regression test, `should honor a threshold of 0 as an override`, in `test/assertions/assertionsResult.test.ts` (a failing + passing assertion with `threshold: 0` → `pass: true`, `score: 0.5`).
  - **Fails before** the fix: `expected false to be true`.
  - **Passes after**.
- Full `assertionsResult` suite passes (17 tests); `npm run tsc`, `npm run l`, `npm run f` all clean.

### DIFF
--- a/site/docs/configuration/expected-outputs/index.md
+++ b/site/docs/configuration/expected-outputs/index.md
@@ -212,7 +212,7 @@ If the LLM output is `Goodbye world`, the `equals` assertion fails but the `cont
 
 ### Setting a score requirement
 
-Test cases support an optional `threshold` property. If set, the pass/fail status of a test case is determined by whether the combined weighted score of all assertions exceeds the threshold value.
+Test cases support an optional `threshold` property. If set, the pass/fail status of a test case is determined by whether the combined weighted score of all assertions is greater than or equal to the threshold value.
 
 For example:
 
@@ -229,6 +229,8 @@ tests:
 ```
 
 If the LLM outputs `Goodbye world`, the `equals` assertion fails but the `contains` assertion passes and the final score is 0.33. Because this is below the 0.5 threshold, the test case fails. If the threshold were lowered to 0.2, the test case would succeed.
+
+A `threshold` of `0` makes the test case pass regardless of individual assertion failures, since the combined score is always at least 0. Use it to collect assertion scores without letting any single failure fail the test. The same applies to an `assert-set` threshold.
 
 :::info
 If weight is set to 0, the assertion automatically passes.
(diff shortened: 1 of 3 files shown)


mldangelo (approved):
LGTM ✅

Clean fix — a test/assert-set `threshold: 0` is now honored as a real override ("collect the scores, never fail on an individual assertion") instead of being silently dropped by the truthy check. Gating on `typeof === 'number' && !Number.isNaN(...)` is the right guard: it keeps `0` working while preventing a stray `null`/empty `threshold:` from force-passing every assertion.

Verified end-to-end (a `threshold: 0` test passes despite a failing assertion; the no-threshold control still fails) plus 20 unit tests, and the docs note in the same change is appreciated.

PR #9737, opened 2026-06-15
fix(assertions): make not-similar with an array require dissimilarity to all values

i-anubhav-anand:
## Summary

For a `similar` assertion with an **array** of values, the documented behavior is "the test passes if the output is similar to **at least one** of them" — an OR. The handler implements this by returning on the first per-value pass.

The bug: the **same loop** is used for `not-similar`. For an inverse assertion, `matchesSimilarity` returns `pass: true` when the output is **dissimilar to that single value**, so returning on the first pass means `not-similar` passes as soon as the output is dissimilar to **any one** value:

```ts
// src/assertions/similar.ts (before)
for (const value of renderedValue) {
  const result = await matchesSimilarity(value, outputString, threshold, inverse, ...);
  if (result.pass) {            // for inverse: "dissimilar to THIS value"
    return { assertion, ...result };  // ← passes on the FIRST dissimilar value
  }
  ...
}
```

So `not-similar: [A, B]` effectively means "dissimilar to A **OR** dissimilar to B", when the correct negation of "similar to at least one" is "dissimilar to A **AND** dissimilar to B" (dissimilar to **all**). The practical consequence: a `not-similar` guardrail listing forbidden/canned/leaked answers gets **weaker with every value you add**, and becomes almost impossible to fail — the opposite of the intent.

## Concrete example

```yaml
assert:
  - type: not-similar
    threshold: 0.75
    value:
      - 'forbidden answer A'
      - 'unrelated answer B'
```

Output = `"forbidden answer A"` (an exact copy of value A):

| value | cosine | inverse check (`sim <= 0.75`) | per-value pass |
|---|---|---|---|
| A | ~1.0 | no | **fail** (too similar) |
| B | ~0.1 | yes | pass (dissimilar) |

| | result |
|---|---|
| **Before** | **pass** ❌ (returns on B, ignoring that the output is identical to A) |
| **After**  | **fail** ✅ (output is too similar to A) |

## Fix

Branch the array logic on `inverse`:

- **`not-similar`** passes only when the output is dissimilar to **every** value. It returns a failing result on the first value the output is too similar to (reporting that value's score), and if the output clears all of them, reports the **lowest** score — the value it came closest to (the tightest margin).
- The positive **`similar`** "match ANY" path is unchanged (returns on first pass; on all-fail reports the highest score, per #9721).

## Test plan

- Added two regression tests:
  - `should FAIL not-similar with an array when the output is similar to ANY value` — **fails before** the fix (`expected true to be false`), passes after.
  - `should PASS not-similar with an array only when dissimilar to ALL values`.
- Ran the suite 5× in random order to confirm no mock-queue leakage; `similar` suite passes (13 tests). `npm run tsc`, `npm run l`, `npm run f` all clean.

### DIFF
--- a/site/docs/configuration/expected-outputs/similar.md
+++ b/site/docs/configuration/expected-outputs/similar.md
@@ -32,6 +32,8 @@ assert:
     threshold: 0.8
 ```
 
+The negated `not-similar` assertion is the logical inverse: with an array of values it passes only when the output is dissimilar to **every** value (and fails as soon as it is too similar to any one of them). This is the natural way to assert that an output does not resemble any item in a list of forbidden or canned answers.
+
 ## Similarity Metrics
 
 You can specify which metric to use by including it in the assertion type. The default is `similar` (cosine similarity).
--- a/src/assertions/similar.ts
+++ b/src/assertions/similar.ts
@@ -40,6 +40,42 @@ export const handleSimilar = async ({
   }
 
   if (Array.isArray(renderedValue)) {
+    if (inverse) {
+      // `not-similar` is the negation of the positive "similar to at least one"
+      // semantics, so it must pass only when the output is dissimilar to EVERY
+      // value. Fail (return early) on the first value the output is too similar
+      // to, reporting that value's score. If the output clears all of them,
+      // report the lowest score — the value it came closest to (the tightest
+      // margin). matchesSimilarity already normalizes the inverse score so
+      // higher is always better.
+      let closestResult: Omit<GradingResult, 'assertion'> | undefined;
+      for (const value of renderedValue) {
+        const result = await matchesSimilarity(
+          value,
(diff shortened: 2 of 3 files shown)


mldangelo (approved):
LGTM ✅

Correct fix for a subtle short-circuit bug: `not-similar` with an array now requires dissimilarity to **all** values — the proper negation of the positive "similar to at least one" — instead of passing as soon as the output differs from any single value. The positive "match ANY" path is untouched, and reporting the closest (lowest-margin) score on the all-dissimilar path is a nice touch.

Verified end-to-end with real embeddings: an output `0.78`-similar to one forbidden value now correctly **fails** (the old code would have passed via the dissimilar second value). 14 unit tests green, plus a docs note documenting the array semantics.

PR #9738, opened 2026-06-15
fix(assertions): honor inverse for not-moderation

i-anubhav-anand:
## Summary

`handleModeration` never read the `inverse` flag and returned the matcher's `pass`/`score` verbatim, so the auto-generated `not-moderation` assertion behaved **identically** to `moderation` instead of negating it. The negation was a complete no-op.

```ts
// src/assertions/moderation.ts (before)
const pass = moderationResult.pass;
return { pass, score: moderationResult.score, reason: moderationResult.reason, assertion };
```

Every other assertion handler honors `inverse` (the dispatcher passes `inverse: isAssertionInverse(assertion)` to all handlers and there is no generic post-handler flip). `moderation` was the lone matcher-backed handler that ignored it — the same class as the recently-fixed `not-levenshtein` (#9722) and `not-cost`/`not-latency`/`not-perplexity` (#9725).

## Behavior

`matchesModeration` reports a **flagged** output as `{ pass: false }` and a **clean** output as `{ pass: true }`. So:

| output | `moderation` | `not-moderation` before | `not-moderation` after |
|---|---|---|---|
| flagged | fail | **fail** ❌ (no-op) | **pass** ✅ |
| clean | pass | **pass** ❌ (no-op) | **fail** ✅ |

`not-moderation` is the natural way to assert "the output **was** flagged" (e.g. confirming a safety/guardrail or red-team harness actually catches harmful content) — which was impossible before.

## Fix

Flip `pass` and `score` for the inverse case, mirroring `handleClassifier`:

```ts
let { pass, score } = moderationResult;
if (inverse) {
  pass = !pass;
  score = 1 - score;
}
```

This matches the established convention: `handleClassifier` flips `pass`/`score` for `not-classifier` the same way. Only the `inverse === true` path changes; `moderation` is byte-for-byte unchanged.

## Test plan

- Added two regression tests in `test/assertions/moderation.test.ts`:
  - `should pass not-moderation when content IS flagged (inverse)` — **fails before** the fix, passes after.
  - `should fail not-moderation when content is clean (inverse)` — **fails before**, passes after.
- Full `handleModeration` suite passes (13 tests); `npm run tsc`, `npm run l`, `npm run f` all clean.

### DIFF
--- a/src/assertions/moderation.ts
+++ b/src/assertions/moderation.ts
@@ -58,6 +58,7 @@ export const handleModeration = async ({
   outputString,
   providerResponse,
   prompt,
+  inverse,
 }: AssertionParams): Promise<GradingResult> => {
   // Priority: 1) response.prompt (provider-reported), 2) redteamFinalPrompt (legacy), 3) original prompt
   // This allows providers to report the actual prompt they sent (e.g., GenAIScript, dynamic prompts)
@@ -86,10 +87,16 @@ export const handleModeration = async ({
     test.options,
   );
 
-  const pass = moderationResult.pass;
+  let { pass, score } = moderationResult;
+  // `not-moderation` asserts the opposite outcome (e.g. that the output WAS
+  // flagged). Flip pass/score for the inverse case, mirroring handleClassifier.
+  if (inverse) {
+    pass = !pass;
+    score = 1 - score;
+  }
   return {
     pass,
-    score: moderationResult.score,
+    score,
     reason: moderationResult.reason,
     assertion,
   };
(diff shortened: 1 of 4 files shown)


mldangelo (approved):
LGTM ✅

`not-moderation` was a complete no-op before this — good fix. The inverse flip mirrors `handleClassifier`, and tagging a moderation API error with `metadata.graderError` + guarding the flip with `isGraderFailure()` means a provider/transport error can never be silently flipped into a "was flagged" pass. It fails closed, consistent with the inverse-aware llm-rubric / g-eval / trajectory handlers.

Verified end-to-end against the real moderation API: flagged content passes `not-moderation`, clean content fails, and plain `moderation` is unchanged. 22 unit tests green, including an error-path regression that pins the fail-closed behavior.

PR #9739, opened 2026-06-15
fix(assertions): score ROUGE case-insensitively

i-anubhav-anand:
## Summary

`handleRougeScore` calls `js-rouge` with **empty options**, so it uses the library default `caseSensitive: true`:

```ts
// src/assertions/rouge.ts (before)
const score = rougeMethod(outputString, renderedValue, {});
```

The other text-overlap metrics all normalize case — `bleu.ts` (`text.toLowerCase()`), `gleu.ts` (`.toLowerCase()`), `meteor.ts` (`word.toLowerCase()`). ROUGE is the lone outlier, so a difference in capitalization silently craters the score while the sibling metrics are unaffected.

## Concrete example

```
output:    "The CAT Sat"
reference: "the cat sat"
```

| | ROUGE-N |
|---|---|
| **Before** | **0.00** ❌ (case-sensitive: no token overlap) |
| **After**  | **1.00** ✅ |

Capitalization differences (sentence-initial caps, proper-noun casing) are pervasive in LLM output, so this produced confusing sub-threshold ROUGE scores that BLEU/GLEU/METEOR didn't show for the same outputs.

## Fix

```ts
const score = rougeMethod(outputString, renderedValue, { caseSensitive: false });
```

`caseSensitive` is a documented `js-rouge` option (`RougeNOptions`/`RougeSOptions`). This brings ROUGE in line with the other overlap metrics.

## Test plan

- Added `test/assertions/rouge.casing.test.ts` — an integration test using the **real** `js-rouge` (no mock): a case-only difference now scores `1.0`.
  - **Fails before** the fix: `expected +0 to be 1`.
  - **Passes after**.
- Updated the existing mock-based `rouge.test.ts` to assert the `{ caseSensitive: false }` options argument.
- `npm run tsc`, `npm run l`, `npm run f` all clean; both rouge test files pass (9 tests).


mldangelo (approved):
LGTM ✅

Good catch that ROUGE was the lone case-sensitive outlier among the text-overlap metrics. Computing ROUGE-N in-house with clipped counts — while reusing js-rouge's own tokenizer / n-gram / f-measure helpers — is the right call: it's byte-identical to js-rouge for inputs without repeats, but fixes both the casing issue and the repeated-token undercount (identical text with a repeated word now scores `1.0` instead of `0.83`/`0.5`).

Verified end-to-end (case-only and repeated-token matches → `1.0`, unrelated → `0`) plus 11 unit tests. Added a short docs note that ROUGE is now scored case-insensitively.

PR #9740, opened 2026-06-15
fix(assertions): stop BLEU collapsing scores for short candidates

i-anubhav-anand:
## Summary

For BLEU, when the candidate is **shorter than the n-gram order** (e.g. a 3-word output has no 4-grams), that order has zero possible n-grams. The code recorded its precision as `0`, which is then smoothed to `1e-7` and multiplied into the geometric mean — collapsing the whole score. This conflates "no n-grams of this order can exist" with "n-grams exist but none matched."

```ts
// src/assertions/bleu.ts (before)
precisions.push(totalCount > 0 ? maxClippedCount / totalCount : 0);  // totalCount===0 → 0
...
const smoothedP = p === 0 ? 1e-7 : p;                                // → 1e-7, dominates
```

The result: a **perfect, identical** short output fails the default `0.5` threshold. Only candidates of ≥ 4 tokens can ever pass.

## Concrete example (exact match, candidate === reference)

| candidate | tokens | BLEU before | default 0.5 |
|---|---|---|---|
| `"cat"` | 1 | **0.00001** | FAIL ❌ |
| `"good dog"` | 2 | **0.00032** | FAIL ❌ |
| `"the good dog"` | 3 | **0.01778** | FAIL ❌ |
| `"the very good dog"` | 4 | 1.00000 | PASS |

After the fix all of these score **1.0**.

## Fix

Skip n-gram orders the candidate is too short to contain, and **renormalize the weights** over the remaining orders:

```ts
if (totalCount === 0) {
  continue; // no n-grams of this order can exist; drop it and reweight
}
precisions.push(maxClippedCount / totalCount);
usableWeights.push(weights[n - 1]);
...
const weightSum = usableWeights.reduce((acc, w) => acc + w, 0);
const weightedScore = precisions.reduce(
  (acc, p, i) => acc + (usableWeights[i] / weightSum) * Math.log(p === 0 ? 1e-7 : p),
  0,
);
```

Key distinction preserved: a **genuine zero-overlap** precision (n-grams of that order *do* exist but none match) is still recorded as `0` and smoothed to `1e-7` — that is a real miss and should be penalized. Only *impossible* orders (candidate too short) are dropped. For candidates of ≥ 4 tokens, every order is present, `weightSum === 1`, and behavior is identical to before.

### Relationship to NLTK

This mirrors NLTK's opt-in [`auto_reweigh`](https://www.nltk.org/api/nltk.translate.bleu_score.html) option, which renormalizes the weights uniformly when higher-order n-grams are unavailable. For the default uniform weights `[0.25]*4` the two are equivalent (dropping one of four equal weights and renormalizing → `1/3` each, matching NLTK's `(1/hyp_len,)*hyp_len`).

To be precise: NLTK's **default** (`auto_reweigh=False`) instead returns `0.0` for a hypothesis shorter than the n-gram order (emitting a "no n-gram overlaps" warning). This PR deliberately chooses the reweighting behavior because returning ~0 for a *perfect* short match is surprising and unhelpful for a thresholded assertion — but it is the `auto_reweigh` behavior, not NLTK's default, and the code comment now says so.

## Test plan

- Added `should score a perfect short match ~1.0 regardless of token count` — **fails before** the fix (`expected 0.0000056 to be close to 1`), passes after.
- Added `should still penalize an imperfect short match` (2-token partial match → `0 < score < 1`).
- All existing BLEU tests still pass (25 total), including brevity-penalty, closest-reference tie-break, and multi-reference cases. `npm run tsc`, `npm run l`, `npm run f` clean.

### DIFF
--- a/src/assertions/bleu.ts
+++ b/src/assertions/bleu.ts
@@ -88,11 +88,24 @@ export function calculateBleuScore(
 
   const maxN = 4;
   const precisions: number[] = [];
+  const usableWeights: number[] = [];
 
   for (let n = 1; n <= maxN; n++) {
     const candidateNGrams = getNGrams(candidateWords, n);
-    const candidateNGramCounts = countNGrams(candidateNGrams);
     const totalCount = candidateNGrams.length;
+
+    // If the candidate is shorter than n tokens, no n-grams of this order can
+    // exist. Skip the order and renormalize the weights over the remaining
+    // orders, rather than recording precision 0 — that would be smoothed to a
+    // tiny value and collapse the whole score (e.g. a perfect 1-3 word match
+    // scoring ~0). This matches NLTK, which reweights when higher orders are
+    // unavailable. A genuine zero-overlap precision (n-grams exist but none
+    // match) is still kept and smoothed below.
+    if (totalCount === 0) {
+      continue;
+    }
+
+    const candidateNGramCounts = countNGrams(candidateNGrams);
 
     // Clipped n-gram count against the best-matching reference
     let maxClippedCount = 0;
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
LGTM ✅

Nice fix for BLEU collapsing short-candidate scores. Dropping the n-gram orders a short candidate can't contain and renormalizing the weights over the remaining orders is exactly right — and the follow-up hardening closes the edge cases cleanly: rejecting negative weights, guarding `weightSum === 0`, and collapsing the two parallel arrays into one `{ precision, weight }` list so a precision can never drift out of sync with its weight.

Verified end-to-end: short exact matches (`cat`, `good dog`, `the good dog`) now score `1.0` and clear the default `0.5` threshold, while genuine mismatches still fail. Identical to the old behavior for ≥4-token candidates. 30 unit tests green.

PR #9749, opened 2026-06-16
fix(share): preserve blob refs per result row

mldangelo-oai:
## Summary
- preserve separate remote blob-reference uploads when the same hash appears in different result rows during cloud sharing
- keep dedupe for repeated references within the same result context so one row still shares one authorize+upload flow
- add a regression test that fails if later result-row coordinates are collapsed into the first upload

## Validation
- `npm_config_cache=/tmp/promptfoo-npm-cache npx vitest@4.0.4 run --config <temporary minimal config> test/blobs/shareUpload.test.ts`
- `npm_config_cache=/tmp/promptfoo-npm-cache npx @biomejs/biome@2.4.16 lint src/blobs/shareUpload.ts test/blobs/shareUpload.test.ts`

## Notes
- Full repo `npm ci` in this worktree is blocked by Socket policy on `esbuild@0.28.1`, so validation used a standalone Vitest run against the touched unit with a temporary minimal config.

### DIFF
--- a/src/blobs/shareUpload.ts
+++ b/src/blobs/shareUpload.ts
@@ -16,6 +16,12 @@ export function createRemoteBlobUploadCache(): RemoteBlobUploadCache {
   return new Map();
 }
 
+// Key uploads by result-row coordinates, not just by hash: the remote records the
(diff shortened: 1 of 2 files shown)


mldangelo:
@codex review

PR #9751, opened 2026-06-16
fix(redteam): surface remote generation body read failures

Gujiassh:
## Summary

Promptfoo's hosted redteam generation currently fails opaquely when the remote endpoint returns a response whose body cannot be read cleanly, such as the Cloudflare-blocked `403 Forbidden` HTML challenge reported in #9679.

The fetch/cache layer already has the request URL and HTTP status in scope at the moment `resp.text()` fails, but it rethrows the raw platform error (for example `TypeError: terminated`) without any context. That leaves redteam extraction logging with an unhelpful warning and makes the issue look like a generic transport crash instead of an actionable remote-generation failure.

This PR keeps behavior the same, but upgrades the diagnostics:

- wraps body-read failures with the request URL
- includes `HTTP <status> <statusText>` when available
- preserves the original error as `cause`
- adds a focused regression test for the non-idempotent POST path

## Verification

- `npx vitest run test/cache.test.ts`
- `npx vitest run test/redteam/extraction/util.test.ts`
- `npx vitest run test/redteam/extraction/purpose.test.ts test/redteam/extraction/entities.test.ts`
- Manual network check on current upstream behavior:
  - `GET https://api.promptfoo.app/health` -> `200 OK`
  - `POST https://api.promptfoo.app/api/v1/task` -> Cloudflare `403` HTML block from this network

## Notes

This is intentionally a diagnostics-only fix. It does not attempt to bypass or change the remote service / Cloudflare policy itself.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -625,12 +625,13 @@ async function fetchAndReadBody(
       // Surface the URL and HTTP response context so opaque body-read failures
       // (e.g. "TypeError: terminated" from a Cloudflare-originated 403) include
       // actionable diagnostics instead of a bare platform error.
-      throw new Error(
+      const wrappedError = new Error(
         `Error reading response body from ${url}: ${
           (err as Error).message
         }. HTTP ${resp.status} ${resp.statusText}`,
-        { cause: err },
-      );
+      ) as Error & { cause?: unknown };
+      wrappedError.cause = err;
+      throw wrappedError;
     }
   }
   // Unreachable: loop always returns or throws, but TypeScript needs this
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -11,14 +11,15 @@ import logger from './logger';
 import { getRequestTimeoutMs } from './providers/shared';
 import { getConfigDirectoryPath } from './util/config/manage';
 import { sha256 } from './util/createHash';
-import { isTransientConnectionError } from './util/fetch/errors';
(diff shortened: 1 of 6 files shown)


mldangelo (approved):
Thanks so much for this, @Gujiassh — really nice catch. The opaque `TypeError: terminated` from a Cloudflare-blocked 403 (#9679) is exactly the kind of failure that sends people down the wrong debugging path, and surfacing the URL + `HTTP <status> <statusText>` while preserving the original error as `cause` is a clean, well-targeted fix. The regression test for the non-idempotent POST path is a great touch. 👏

I pushed one small follow-up commit on top of your work (kept your commits and your portable error-cause assignment) to round out two edge cases, since `fetchWithCache` is a widely-shared utility:

- **Sanitize the URL** — `sanitizeUrl(getRequestUrlString(url))`. Some callers pass credential-bearing URLs (`?api_key=`, signed URLs, userinfo), so this keeps secrets out of the logged message, and also avoids `[object Request]` if a `Request` is passed.
- **Preserve `AbortError`** — an aborted body read rejects with an `AbortError`, and `evaluator.ts` suppresses expected cancellation via `err.name === 'AbortError'`. Wrapping it would reset the name and turn cancelled evals into reported failures, so aborts now rethrow unchanged.
- Small reuse cleanup (a shared `isAbortError` helper) + regression tests for both of the above.

CI is green across the board. Thank you for the contribution and for iterating on it — merging now! 🎉

PR #9757, opened 2026-06-16
fix(csv): unescape all escaped commas in array metadata values

he-yufeng:
The array-metadata CSV columns (`__metadata:foo[]`) let you escape a comma inside a value with `\,`. The split regex `/(?<!\\),/` correctly keeps escaped commas from acting as separators, but the unescape step uses `value.replace('\\,', ',')`. With a string as the first argument, `String.prototype.replace` only replaces the first match, so any value containing more than one escaped comma keeps every `\,` after the first.

Example: a cell `a\,b\,c` should parse to the single tag `a,b,c`, but today it yields `a,b\,c`.

The fix is a one-liner: use a global regex so every escaped comma is unescaped.

```diff
-            .map((v) => v.replace('\\,', ','));
+            .map((v) => v.replace(/\\,/g, ','));
```

Behavior is identical for values with at most one escaped comma, so the existing escaped-comma test is unaffected. I added a regression test covering the multi-escape case.

Verified with `npx vitest run test/csv.test.ts`: the new test fails on `main` and passes with the fix, 65 tests pass total.

### DIFF
--- a/src/csv.ts
+++ b/src/csv.ts
@@ -196,7 +196,7 @@ export function testCaseFromCsvRow(row: CsvRow): TestCase {
           const values = value
             .split(/(?<!\\),/)
             .map((v) => v.trim())
-            .map((v) => v.replace('\\,', ','));
+            .map((v) => v.replace(/\\,/g, ','));
           metadata[arrayKey] = values;
         }
       } else {
--- a/test/csv.test.ts
+++ b/test/csv.test.ts
@@ -170,6 +170,27 @@ describe('testCaseFromCsvRow', () => {
     expect(result).toEqual(expectedTestCase);
   });
 
+  it('should split on unescaped commas while unescaping multiple escaped commas per value', () => {
+    const row: CsvRow = {
+      '__metadata:tags[]': 'a\\,b\\,c,d\\,e',
+      var1: 'value1',
+    };
+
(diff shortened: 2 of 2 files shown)


mldangelo (approved):
Thanks so much for this fix, @he-yufeng — and welcome to promptfoo! 🎉 This is a really nice first contribution.

The root-cause write-up is spot on: `String.prototype.replace(string, …)` only swaps the **first** match, so any array-metadata cell with more than one escaped comma (e.g. `a\,b\,c`) was leaking every `\,` after the first. Switching to the global regex is exactly the right, minimal fix, and I appreciate that you included a focused regression test that fails on `main` and passes with the change.

I verified everything locally:
- Reproduced the bug, confirmed the fix resolves it, and grepped the codebase to make sure there were no other instances of the same first-match-only `.replace('…')` pattern hiding elsewhere (there weren't).
- Synced the branch with the latest `main`.
- Added one more regression test covering the combined case — multiple array elements that *each* contain multiple escaped commas (`a\,b\,c,d\,e` → `['a,b,c', 'd,e']`) — to lock in that the unescaped-comma split and the global unescape compose correctly.

All CI is green. Merging now. Thanks again for the careful fix and clear explanation — hope to see more PRs from you! 🙏

PR #9768, opened 2026-06-16
fix(csv): preserve zero __threshold in testCaseFromCsvRow

JSap0914:
## Summary

Preserve a valid test-level `__threshold: 0` when loading CSV test cases. The previous truthy spread guard dropped zero even though a zero aggregate threshold is meaningful.

The shared numeric parser also prevents blank, malformed, or non-finite CSV cells from leaking `NaN` or `Infinity` into a `TestCase`, which would block `defaultTest.threshold` inheritance.

## Numeric behavior

- Complete finite decimal literals are accepted, including zero, signs, leading/trailing-dot decimals, and exponents.
- Blank or malformed test-level `__threshold` cells are treated as unset so `defaultTest.threshold` can apply.
- Malformed assertion-level `__config:__expected...:threshold` cells fail fast instead of being partially parsed.
- Negative and exponent forms remain accepted for backward compatibility with the previous `Number.parseFloat` behavior. Range validation, if desired, should be a separate change at the relevant threshold call site.

`CHANGELOG.md` is generated by the release process, so the user-visible behavior changes are recorded here instead of editing it manually.

## Validation

- `npx vitest run test/csv.test.ts test/assertions/assertionsResult.test.ts test/evaluator/defaultTest.test.ts` — 121 passed
- `npm run test:app -- src/pages/eval-creator/components/TestCasesSection.test.tsx --run` — 20 passed
- Real local CLI eval from CSV — 3/3 passed:
  - zero threshold preserved and overrode two failing assertions at aggregate score 0
  - blank threshold inherited `defaultTest.threshold: 0.5`
  - malformed `0abc` test threshold was omitted and inherited the same default
- Real local CLI invalid-config case — malformed assertion-level `0abc` threshold exited with the expected validation error
- `npm run tsc`
- `npm run l && npm run f`

Reviewed after merging `origin/main` at `81fe4479e722ae416b67e4e20fd5ba723dc91ff6`.


mldangelo:
Reviewed and tested end-to-end (80 `csv` tests green; via `tsx` confirmed `__threshold: "0"` now survives evaluator inheritance, and blank/`0abc` cells are omitted/throw instead of leaking `0`/`NaN`). Genuine bug fix — a valid `0` threshold was being dropped by the falsy guard.

**Pushed** a test-only follow-up: an `it.each` pinning the numeric contract of `parseFiniteNumber` (integers, decimals, leading-dot, negatives, exponents, trimmed) so a future regex tweak can't silently change it.

Two intentional behavior changes worth a changelog note (now covered by tests): `__config:...:threshold` now **throws** on malformed cells like `0abc` (was silently truncated to `0`), and thresholds now accept negatives/exponents. Negatives are harmless downstream (score ≥ -1 always passes); if you'd rather reject them, gate at the call site, not the shared helper. ✅

JSap0914:
Thanks for the thorough review and for adding the contract tests — much appreciated. Agreed the throw-on-malformed (e.g. `0abc`) and the now-accepted negatives/exponents deserve a changelog note. I'm happy to add that note, and if you'd rather reject negatives I can gate that at the call site instead of in the shared `parseFiniteNumber` helper — let me know which you prefer.

mldangelo-oai:
Thanks — I made the compatibility call and finished the loop. We should keep negative and exponent forms here because the previous Number.parseFloat path already accepted them and the schema has no range restriction; rejecting them would be a separate breaking change at the wrong layer. I did not edit CHANGELOG.md because this repository generates it during release. Instead, the PR description now records both intentional behavior changes explicitly. Pushed 850eae158 with finite-boundary coverage (Infinity, overflow, hexadecimal, signed exponents, plus signs, and trailing-dot decimals), then verified the real evaluator path: zero preserved, blank/0abc inherited the default threshold, and malformed assertion-level config failed fast.

ianw-oai (approved):

ianw-oai:
Thank you!

PR #9779, opened 2026-06-16
fix(sanitizer): redact malformed token params

mldangelo-oai:
## Summary
- redact malformed URL form segments when their parameter name contains a sensitive token marker
- preserve the recent benign bare-keyword behavior for non-secret url values
- add regression coverage for malformed private/github token params

## Validation
- `./node_modules/.bin/vitest run test/util/sanitizer.test.ts -t 'should redact unparseable URLs that carry credential indicators' --sequence.shuffle=false` (fails before fix, passes after)
- `./node_modules/.bin/vitest run test/util/sanitizer.test.ts --sequence.shuffle=false`
- `npm run l`
- `npm run f`
- `npm run tsc`

## Notes
- Fresh-worktree `npm ci` remains blocked by Socket policy for `esbuild@0.28.1`, so validation used the existing Promptfoo checkout's installed dependency tree via a temporary local symlink only during test execution.


ianw-oai (approved):
Reviewed the sanitizer change at the current head. Small, focused fix with regression and false-positive coverage; checks are green.

PR #9781, opened 2026-06-16
feat: add per-test repeat option

amarjaleelbanbhan:
## Summary

Adds `tests[].options.repeat` so an individual test can override the global repeat count.

## Behavior

- A positive safe integer in `test.options.repeat` overrides `--repeat`, `commandLineOptions.repeat`, or `evaluateOptions.repeat` for that test.
- Tests without an override continue to use the global repeat count.
- Scenario precedence is `defaultTest.options` < scenario `config[].options` < scenario `tests[].options`.
- Every repeat index uses its own cache namespace, including repeat index 0, so repeated runs do not reuse or overwrite the normal non-repeat cache entry.
- `repeat` remains evaluator-only and is not forwarded in provider prompt config.
- Invalid programmatic values fall back to the effective global repeat; the public Zod and generated JSON schemas diagnose invalid config values.

Scenario config now inherits all `options` fields, matching the existing vars/metadata/assert inheritance behavior. Scenario test options still take precedence.

## Verification

Validated on head `c2e574bf1` after merging `origin/main` at `81fe4479e`:

- `npx vitest run test/evaluator/basic.test.ts test/evaluator/repeatCache.test.ts test/evaluator/runEval.test.ts test/evaluator/scenarios.test.ts test/types/index.test.ts test/config-schema.test.ts` — 467 tests passed.
- `npm run jsonSchema:generate` — generated schema remained byte-for-byte clean.
- `npm run tsc`, `npm run l`, `npm run f`, and `npm run architecture:check` — passed.
- `SKIP_OG_GENERATION=true npm run build` in `site/` — production docs build passed.
- Real local CLI eval with global repeat 2 plus per-test counts 1 and 3, an inherited global count, and scenario config/test precedence — 8/8 results passed with the expected repeat indexes.
- Real CLI eval with fractional `options.repeat: 1.5` — emitted the Zod diagnostic and safely fell back to one run.


mldangelo:
Reviewed and tested end-to-end with **real CLI evals** (per-test `repeat: 3` + no global → 3 runs; `repeat: 1` + `--repeat 2` → 1 run; scenario `config.repeat` honored; string/NaN/fractional rejected by Zod with a safe fallback to 1). 467 tests green across the 6 changed files; `tsc` clean. The cache-namespace isolation (the original P1) is solid.

**Pushed** a test-only follow-up: a regression test pinning scenario **precedence** — scenario `test.options.repeat` overrides scenario `config.options.repeat` — which the new `...data.options` line introduces but nothing pinned.

Sign-off: that `...data.options` line now merges **all** scenario-config options (not just `repeat`; e.g. `transform`, `prefix`, `runSerially` were previously dropped). This is the consistent/correct behavior (vars/metadata/assert were already merged) and is covered, but worth an explicit ack. ✅

amarjaleelbanbhan:
Thanks for the thorough review and additional coverage.

I agree with the scenario options merge behavior. Since vars, metadata, and assertions already inherit scenario config values, applying the same precedence for options keeps the behavior consistent.

Appreciate the validation and follow-up fixes.

ianw-oai (approved):

ianw-oai:
Thank you :)

PR #9783, opened 2026-06-16
fix(providers): apply xAI cache-read discount to cost calculation

i-anubhav-anand:
## Summary

xAI reports prompt-cache hits in `usage.prompt_tokens_details.cached_tokens` and bills those tokens at the lower cached-input rate. Promptfoo already normalized that field into `tokenUsage.completionDetails.cacheReadInputTokens`, but the xAI cost path previously billed every prompt token at the full input rate.

This PR now:

- bills uncached prompt tokens at the normal input rate and cached prompt tokens at `cacheReadCost` / the model's `cache_read` rate
- reads cache usage from Promptfoo's normalized Chat Completions contract and Responses API usage fallback
- preserves custom `cost` / `inputCost` overrides unless `cacheReadCost` is explicitly configured
- clamps malformed cache counts to the valid `0..promptTokens` range
- falls back to full input pricing for models without a known cache-read rate
- exposes the pricing overrides through typed chat and Responses configuration and documents them
- removes dead raw-response parsing that duplicated the base OpenAI-compatible usage normalization

For 1,000 prompt tokens (800 cached) and 500 output tokens at $3/M input, $0.75/M cached input, and $15/M output, the calculated cost changes from $0.0105 to $0.0087.

## Vendor contract

xAI documents `usage.prompt_tokens_details.cached_tokens` for Chat Completions, `usage.input_tokens_details.cached_tokens` for Responses, and states that cached prompt tokens use the reduced cached-prompt-token price: https://docs.x.ai/developers/advanced-api-usage/prompt-caching/usage-and-pricing

## Validation

- 91 xAI chat and Responses provider tests pass, including cache hits, cache misses, full-cache, malformed boundaries, explicit zero pricing, missing cache rates, custom overrides, and missing usage
- `npm run tsc`, changed-file lint, and changed-file formatting pass (lint retains the pre-existing non-blocking complexity warning in `XAIResponsesProvider.callApi`)
- production docs build passes with `SKIP_OG_GENERATION=true`
- deterministic local `promptfoo eval --no-cache --no-share` exercises both `xai:grok-4.3` and `xai:responses:grok-4.3`: 2/2 pass, score 1, no errors, 800 cache-read tokens and $0.0087 cost on each result


mldangelo:
Reviewed and tested end-to-end (xAI chat+responses suites green, 89→90 with the new test; `tsc` clean; re-derived every expected cost from the model table). The cache-read discount, the `[0, promptTokens]` clamp, and the `cacheReadCost != null` guard (correctly avoiding the `|| undefined` anti-pattern) all check out.

**Pushed** a test-only follow-up: a regression test for models with **no** `cache_read` rate (grok-2 / grok-beta class) — reporting cached tokens for them must bill at the full input rate and not produce `undefined * tokens → NaN`. Asserts it equals the no-cache cost.

Informational only: `cacheReadCost` is read off the any-typed config, so it works at runtime but isn't on the typed `XAIConfig` surface (not in the JSON schema) — optional future follow-up. ✅

ianw-oai (approved):

ianw-oai:
Thanks :)

PR #9784, opened 2026-06-16
fix(assertions): stop is-sql false-rejecting SELECT DISTINCT

i-anubhav-anand:
## Summary

`is-sql` uses an extra heuristic for `SELECT a b FROM t`, which `node-sql-parser` accepts as implicit aliasing even though it is commonly a missing comma. The heuristic also treated a leading `SELECT` modifier as the first column, so valid statements such as `SELECT DISTINCT name FROM users` failed.

## Fix

- Skip recognized leading `SELECT` modifiers before applying the two-column missing-comma heuristic.
- Apply MySQL-only modifiers only to MySQL/MariaDB so other dialects still reject likely implicit-alias mistakes.
- Continue rejecting missing commas after modifiers, including `SELECT DISTINCT first_name last_name FROM employees`.
- Reject empty and whitespace-only SQL, which `node-sql-parser` otherwise accepts.
- Ignore SQL-looking text inside quoted literals, identifiers, and comments before applying the missing-comma heuristic.
- Preserve dialect expressions such as PostgreSQL/BigQuery square-bracket subscripts and BigQuery, T-SQL, and SQLite prefixed literals while still ignoring bracketed identifiers where supported.
- Scan quoted/comment text without regex backtracking; a 249 KB adversarial unmatched-dollar-tag input improved locally from about 6.1 seconds to 0.22 seconds.
- Load the CommonJS parser through both native and transformed module shapes so source-level CLI execution works on Node 24/tsx.
- Reset mocked parser modules around loader-error tests so randomized test order cannot leak module state.
- Cover direct `contains-sql` fenced extraction and modifier, dialect, quoted, malformed, empty, and adversarial-input cases.

## Verification

- `npx vitest run test/assertions/sql.test.ts test/assertions/runAssertion.test.ts --coverage --coverage.include=src/assertions/sql.ts --sequence.seed=9784` — 236 passed; `sql.ts`: 97.79% statements/lines, 90% branches, 100% functions.
- SQL tests passed across 20 additional shuffled seeds.
- `npm run test:coverage:ratchet -- --report backend --base origin/main` — passed.
- `npm run tsc` — passed.
- `npm run l` and `npm run f` — passed with the pre-existing `handleIsSql` complexity warning.
- `SKIP_OG_GENERATION=true npm run build` from `site/` — passed.
- Real local echo-provider eval via `npm run local -- eval ... --no-cache --no-share -o ...json` — 11/11 passed, score 1 each, 0 errors, covering valid/invalid modifier, dialect gating, bracket subscripts, prefixed literals, quoted, empty, and fenced paths.

## Reference

- [MySQL SELECT syntax](https://dev.mysql.com/doc/refman/8.0/en/select.html)


mldangelo:
Reviewed and tested end-to-end (real `promptfoo eval`: `SELECT DISTINCT first_name FROM users` now passes `is-sql`, while `SELECT DISTINCT first_name last_name FROM users` is still correctly rejected; 45→46 `sql` tests green). Confirmed the fix is load-bearing — node-sql-parser *accepts* `SELECT a b FROM t` and even `''`, so the modifier-aware heuristic and empty guard each reject inputs the parser would pass.

**Pushed** a test-only follow-up: a regression test driving the public `contains-sql` path against a **fenced** `SELECT DISTINCT first_name last_name` (the parser accepts it once unwrapped, but the missing-comma heuristic correctly fails it) — previously only happy-path covered.

⚠️ **Coordination:** this conflicts with #9785 (same author) on `src/assertions/sql.ts` — both refactor the node-sql-parser import block and both add the identical empty-input guard, so whichever lands second needs a rebase. Cleanest resolution: keep **this** PR's `createSqlParser()` helper + empty guard and take #9785's fence-regex fix (they're complementary). ✅

mldangelo:
Thank you!

PR #9785, opened 2026-06-16
fix(assertions): let contains-sql extract fenced SQL with backtick identifiers

i-anubhav-anand:
## Summary

`contains-sql` used a single `[^`]+` regex match, so valid MySQL identifiers quoted with backticks broke extraction. The same first-match approach also let an invalid/non-SQL first fence hide a later valid SQL block and applied `not-contains-sql` to the wrong candidate.

This update:

- scans Markdown fences in one linear pass, supporting backtick and tilde fences plus longer closing fences;
- validates every untagged or exact `sql` block until one contains valid SQL;
- applies `not-contains-sql` once to the aggregate result;
- rejects empty and whitespace-only candidates instead of relying on `node-sql-parser`, which accepts them without throwing;
- loads both the named-export and CommonJS-default shapes of `node-sql-parser`, including the Node 24 source-loaded CLI path; and
- documents fenced SQL and backtick-quoted identifiers.

The raw-SQL path remains supported when the output has no recognized SQL fence.

## Reference

- [MySQL identifier quoting](https://dev.mysql.com/doc/refman/8.0/en/identifiers.html)

## Validation

- `npx vitest test/assertions/sql.test.ts test/assertions/runAssertion.test.ts --run --coverage` — 216 passed
- changed `src/assertions/sql.ts` coverage — 95.19% statements/lines, 87.65% branches, 100% functions
- `npm run test:coverage:ratchet`
- Node 24 source import through `node --import tsx` — fenced backtick SQL passes
- real `npm run local -- eval ... --no-cache --no-share` — 10/10 passed with score 1 and no errors, covering raw SQL, mixed/multiple fences, MySQL backticks, tilde/four-backtick/long closing fences, whitespace, invalid blocks, and inverted assertions
- `npm run tsc`
- `npm run l && npm run f` (only the pre-existing `handleIsSql` complexity warning)

## Coordination

#9784 independently introduces the same `createSqlParser()` compatibility helper. If it lands first, keep that helper while resolving the overlap; this PR's unique behavior is the fenced-block extraction and aggregate validation.


mldangelo:
Reviewed and tested end-to-end (real `promptfoo eval`: fenced SQL with backtick-quoted identifiers is extracted and validated; 36 `sql` tests green; `tsc` clean). Two solid wins: the empty-input guard fixes a genuine false-positive (`astify('')` returns `[]` without throwing, so empty output previously *passed* is-sql), and the ESM-interop import is a real robustness improvement.

I deliberately **did not push** a change here. The only candidate — anchoring the optional `sql` fence tag so `` ```sqlSELECT 1 `` doesn't mis-extract — touches the exact lines #9784 rewrites and would worsen the conflict below for ~zero real-world benefit, so it's better left as an optional follow-up.

⚠️ **Coordination:** conflicts with #9784 on `src/assertions/sql.ts` (both add the same empty guard + refactor the parser import). Clean merge = #9784's `createSqlParser()` + guard plus **this** PR's unique fence-regex fix. ✅

ianw-oai (approved):

ianw-oai:
Thanks!

PR #9792, opened 2026-06-17
fix(providers): reconcile model catalogs, routing and costs

mldangelo:
Correct provider model selectors, API routing and cost estimates while preserving platform-specific namespaces, explicit deployment IDs and existing embedding spaces.

## Changes

- Refresh catalogs and active examples for Gemini 3.8 Flash, DeepSeek Flash, Alibaba, Cerebras, Cohere and Together.
- Use the effective model, serving platform, tier and cached-token counts for routing and costs; preserve explicit overrides and historical metadata.
- Correct Google media/Interactions/Live requests, Cohere conversation conversion and Bedrock profile/agent configuration. Includes the Sonic fixes from #10310.
- Document Embedding 2 migration separately from the retained Embedding 1 default and identify retired Vertex Claude selectors.
- Honor scoped Voyage `apiKeyEnvar` credentials and isolate registry tests from telemetry and simulated module-load failures.

Primary sources, checked September 10–12, 2026: [Gemini 3.8 Flash](https://ai.google.dev/gemini-api/docs/models/gemini-3.8-flash), [DeepSeek releases](https://api-docs.deepseek.com/updates/#date-2026-09-10), [Google embeddings](https://ai.google.dev/gemini-api/docs/embeddings), and [Vertex partner-model retirements](https://docs.cloud.google.com/gemini-enterprise-agent-platform/models/deprecations/partner-models).

## Validation

- Main `41140894` is merged in `b1d49596`; only the root package manifests changed. Normal lint, formatting and merge hooks passed. All 44 reported CI checks pass on this head, including package/app builds on Node 22.22, 24.x and 26.x and the Linux, Windows and macOS test jobs.
- Unchanged registry/test-hygiene files passed 495 cases with randomized ordering and the failing CI seed on `a2430a3d`; its Node 22.22 Ubuntu, Windows and macOS CI passed.
- Voyage changes passed 13 tests, types, core/declaration/postbuild commands and five actual local CLI evaluations with exported output and grading usage. Earlier provider/routing checks, two Echo CLI/database cases, app validation and the documentation build (`SKIP_OG_GENERATION=true`) retain their original scope.

Local tests use a qualified dependency installation and offline fixtures. They verify configuration, request and output behavior; live vendor availability and billing were not tested.


site/docs/providers/cerebras.md
```diff
@@ -115,7 +115,7 @@ When using tool calling, you'll need to process the model's response and handle
 
 ## Example Configuration
 
-```yaml
+```yaml title="promptfooconfig.yaml"
```
jameshiester-oai, line 121: **[P2] Use models from Cerebras' current public catalog**

This newly titled config advertises `llama3.1-8b` and `llama-3.3-70b`, but Cerebras' [deprecation log](https://inference-docs.cerebras.ai/support/deprecation) marks them deprecated on 2026-05-27 and 2026-02-16 respectively, and neither appears in the [current public model catalog](https://inference-docs.cerebras.ai/models/overview). Promptfoo forwards the provider suffix as the API model ID, so schema/factory validation succeeds while a current public-endpoint eval uses unavailable IDs. Replace them with current catalog models (for example `gpt-oss-120b`) or leave the block untitled until the example is current.

site/docs/providers/mistral.md
```diff
@@ -383,7 +383,8 @@ tests:
 
 ### Example 1: Multi-Model Comparison
 
-```yaml
+```yaml title="promptfooconfig.yaml"
```
jameshiester-oai, line 413: **[P2] Correct the model IDs before presenting this comparison as runnable**

The promoted config includes deprecated `magistral-medium-latest` and `mistral-medium-3.5`. Mistral's [native-reasoning migration guidance](https://docs.mistral.ai/resources/deprecated/native-reasoning) recommends `mistral-medium-3-5` with `reasoning_effort`, and the [current model card](https://docs.mistral.ai/models/model-cards/mistral-medium-3-5-26-04) uses that hyphenated ID. Promptfoo forwards the configured suffix, so the dotted ID passes local schema/registry checks but is not the vendor's API ID. Use current IDs here and update the matching pricing metadata/table if this example should keep its cost assertion.

site/docs/providers/openclaw.md
```diff
@@ -279,6 +282,7 @@ tests:
 ### Using Embeddings
 
 ```yaml title="promptfooconfig.yaml"
```
jameshiester-oai, line 285: **[P2] Do not use an embedding-only provider as the top-level eval provider**

The schema hint certifies this existing titled block as a runnable config, but executing it exits 100 with `Not implemented`: `OpenClawEmbeddingProvider` inherits `OpenAiEmbeddingProvider.callApi`, which throws because embedding providers only implement `callEmbeddingApi`. Show this provider as the embedding grader for a supported assertion such as `similar`, or remove the runnable marker instead of putting it in the top-level `providers` list.

site/docs/providers/vercel.md
```diff
@@ -172,6 +173,7 @@ For a complete list, see the [Vercel AI Gateway documentation](https://vercel.co
 Generate embeddings for text similarity, search, and RAG applications:
 
 ```yaml title="promptfooconfig.yaml"
```
jameshiester-oai, line 173: **[P2] Route this example through the embedding API and a supported assertion**

The exact schema-marked config exits 100 before a network request because `VercelEmbeddingProvider.callApi` returns `Use callEmbeddingApi for embedding models`. Independently, `is-valid-embedding` has no runtime assertion handler and reaches `Unknown assertion type`. Rewrite this as a supported embedding-grader example (for example, a `similar` assertion with this provider configured for embeddings), or remove the runnable schema marker.

site/docs/providers/google.md
```diff
@@ -457,7 +460,8 @@ tests:
 
 Generate videos from a starting image:
 
-```yaml
+```yaml title="promptfooconfig.yaml"
```
jameshiester-oai: **[P2] Supply the media prerequisites for these newly runnable configs**

Copying this newly titled image-to-video config into a clean directory exits 100 with `Image file not found: assets/start-frame.jpg` before any Google call. The interpolation block at lines 484–490 (`start.jpg`/`end.jpg`) and multimodal block at lines 640–651 (`red-panda.jpg`) have the same issue, and the page provides no assets or setup step. Add explicit download/create/copy prerequisites for each asset, or keep these blocks untitled as fragments.

jameshiester-oai (commented):
## Review scope

- Base: `58e7c8326867d470689f1f9978660c6f73e6f7c0`
- Head: `642e2abfec5ef02b1b78a5d5d9578027382c7709`
- Scope: fresh full review of the complete base...head diff; no prior authenticated completion marker exists for this head.
- CI: every reported check is successful. GitHub reports the PR as `DIRTY` because it currently conflicts with the base branch.

## Verdict

I found five additional P2 defects in examples this change newly presents or certifies as complete runnable configurations. The static schema/factory audit passes these examples, but realistic execution or current vendor catalogs show that they cannot run as advertised.

### P2

1. The Cerebras example uses two model IDs that are deprecated and absent from the current public model catalog.
2. The Mistral comparison combines a deprecated Magistral model with the misspelled, non-catalog ID `mistral-medium-3.5`.
3. The OpenClaw embedding provider is used as a top-level eval provider even though it implements only `callEmbeddingApi`; the exact config exits with `Not implemented`.
4. The Vercel embedding provider is also used through `callApi`, and the example additionally uses the unsupported `is-valid-embedding` assertion.
5. Three newly titled Google media configs reference local assets that the page neither supplies nor tells the reader how to create; the exact image-to-video config exits with `Image file not found`.

No P0, P1, or P3 findings were validated.

## Existing feedback

The current-head Codex comment at https://github.com/promptfoo/promptfoo/pull/9792#issuecomment-4734200823 already covers eight other runnable-example failures (Azure Sora, Hyperbolic image/audio, CometAPI, Slack, ElevenLabs STT/Agents, and LlamaAPI). Those findings still target this exact head and are not duplicated here. The 12 older inline threads are resolved; several are outdated after follow-up fixes.

## Test gap and simplification opportunity

The site suite and production build validate Markdown/Docusaurus behavior but do not execute fenced configs. The PR's one-off audit likewise stops at YAML/schema/provider-factory checks. A single committed runnable-snippet harness should also verify assertion-handler existence, top-level `callApi` capability, and referenced companion files. That would replace ad hoc checks and catch the OpenClaw/Vercel/assertion failures before review. Model examples also need a current-catalog source or periodic catalog check so retired and misspelled IDs do not remain schema-valid.

## Validation performed

- `npm test --prefix site` — exit 0; 101 tests passed.
- `npm run typecheck --prefix site` — exit 0.
- `SKIP_OG_GENERATION=true npm run build --prefix site` — exit 0; production docs build succeeded.
- `npx vitest run test/providers/openclaw.test.ts test/providers/vercel.test.ts test/providers/openai/agents.test.ts test/providers/google/video.test.ts` — exit 0; 295 tests passed.
- `npx vitest run test/providers/cerebras.test.ts test/providers/mistral.test.ts` — exit 0; 51 tests passed.
- Read-only extraction/audit of all 126 newly promoted configs — all parsed and passed the published JSON schema, `UnifiedConfigSchema`, and provider dispatch; runtime allowlist checking exposed unsupported assertions.
- Exact copied OpenClaw, Vercel, Google, and unsupported-assertion configs run with `npm run local -- eval ... --no-cache -o ...` — each exited 100 with the reproduced pre-network failures described above.
- Exact copied OpenAI Agents config also failed from a clean directory, but that candidate was discarded because the immediately linked `openai-agents-basic` initializer supplies both referenced modules.
- `git diff --check 58e7c8326867d470689f1f9978660c6f73e6f7c0...642e2abfec5ef02b1b78a5d5d9578027382c7709 -- site/docs/providers` — exit 0.
- Vendor claims were checked against the official Cerebras model catalog/deprecation page and Mistral model/deprecation pages.

## Limitations

No paid/live third-party provider calls were made, so the Cerebras and Mistral findings are grounded in current primary documentation plus the request paths that forward these model IDs. I did not perform an exhaustive external-link crawl or browser interaction pass; the production docs build, internal link validation, and focused config executions covered the changed behavior. The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable in this environment (exit 127), so I performed the equivalent read-only correctness, tests, docs/comments, error-handling, type/API, and simplification passes manually and reconciled three independent specialist reports.

mldangelo-oai:
## Final review receipt

**Disposition: `FIXED_AND_READY`**

- **Head:** `9223c33c7dcef28a5fab1fa993275fcef7b3a88e`
- **Base:** `370c7a0c33a5154762deabbbaf8f4effdd6e0988`
- **Ancestry:** dispatch head `642e2abfec5ef02b1b78a5d5d9578027382c7709` and the current base are both ancestors of the final head; the merge base with `origin/main` is exactly `370c7a0c33a5154762deabbbaf8f4effdd6e0988`.
- **Additive commits since dispatch:**
  - `33541819f24718ee7b5c4501a5a327836ad5d6fa` — merge the then-current `origin/main`
  - `c55314faeed250736504943d829147d5ee1ec758` — fix provider examples and model references
  - `4707e7e12eb5b921bb0b37f892bd607d18a2f6fb` — fix remaining runnable provider examples
  - `3854fc1476df324d1c5b469ef4e8dc14aad07207` — refresh Cerebras catalog examples
  - `b971760304f500823d3fc67971da121c20a20723` — address remaining provider review feedback
  - `8c29d83a24967fa504f72d5ac0b778bc5c6dda45` — mark the Claude skill config as a fragment
  - `ab304e33c59a7f0c25a6fc1aa3daa6a90b8c566b` — correct AI/ML API model IDs
  - `bd41953a648440511cd8e50b2f01df3ba380d64d` — pair runnable prompts with tests
  - `bf652f835a8368e319ad1d6652abf140b58a22ea` — correct provider migration guidance
  - `e9a948fb62233c185e4bfd00072b450766eff381` — keep runnable checks self-contained
  - `2876aedbf8009c2e0507833ac4e3c8cd71d38d01` — align migrated model metadata
  - `c70f23bd46382591b7c0a343abcad309ad755423` — refresh Hyperbolic catalog guidance
  - `1153db383261aff9a0ac6b0c8764ca561e548998` — complete the Grok Build migration
  - `248b5690c1b7dd04ab99375cebbc2459ddaa8b04` — preserve Hyperbolic catalog anchors
  - `342d059bb8d3f3bf1d26a21529c4c9fed11d7953` — merge current `origin/main`
  - `8eca65f04c70d80be51a0fc47506b29a15b389ff` — mark placeholder-dependent configs as fragments
  - `8dad40dc8bdfb3e6589f654a4bbc0b9066af6024` — merge current `origin/main`
  - `d98a7e7db463640a5567f3a474ddc8e2f924e628` — align runnable gateway examples
  - `d3db500b7b80cca83d1c91514676cb4116011b67` — correct Helicone gateway examples
  - `9061bfa68a9d55f8922ac05053354e2f5f9ecfc6` — merge current `origin/main`
  - `f5b83ae32185fc8df89421c0249d3e8dc197d867` — clarify cache and thread examples
  - `395d69bb29e19e2dd2c09b7f34c65fefde6f98f6` — merge current `origin/main`
  - `99e09a77825f6d97fbb29b929b79a60504516fd5` — merge current `origin/main`
  - `9223c33c7dcef28a5fab1fa993275fcef7b3a88e` — clarify Veo Lite input limits
- **Push/worktree:** local head equals `origin/docs/audit-provider-docs`; the worktree is clean; the branch was pushed normally with no history rewrite.
- **Production LOC:** runtime source is 2 files with 58 insertions and 30 deletions (88 changed lines, net +28). Tests are 2 files with 93 insertions and 4 deletions. Provider docs are 69 files with 926 insertions and 1,171 deletions. Total impact is 73 files, 1,077 insertions, and 1,205 deletions (net -128).
- **Review threads:** 25/25 resolved; 0 current unresolved and 0 outdated unresolved. The Grok Code Fast guidance maps to `grok-build-0.1`, the canonical hyphenated Mistral Medium 3.5 slug has matching pricing metadata, and the later runnable-example threads are addressed. No newer current thread appeared after the final push and CI run.
- **Evidence-backed scope decisions:** the broader `mistral-medium-latest` / `mistral-medium-3` pricing gap is pre-existing in the exact base and was not expanded into this docs repair. The xAI REST `long_context_threshold` contract keeps the inclusive boundary. Veo 3.1 Lite's unsupported reference-image and video-extension inputs are now documented without expanding runtime validation. A reviewer P3 about frontmatter-description length was independently rejected as non-functional search/social-copy style; expanding it would have added unverified marketing copy.
- **Exact-head runnable QA:** all 126 complete titled configs passed YAML, `UnifiedConfigSchema`, assertion-type, and provider-factory validation; they instantiate 160 providers, with only the two intentional user-executable MCP and QuiverAI examples skipped. Preserved zero-cost echo render-matrix evals produced exactly 2 Azure cases and 4 Slack cases with no prompt/test cross-products or undefined variables.
- **Exact-head local QA:** 157 focused xAI, Mistral, and Google Video tests passed; site tests passed 101/101; root TypeScript and site typecheck passed; Prettier and `git diff --check` passed; the production Docusaurus build passed with `SKIP_OG_GENERATION=true`.
- **Preserved broader validation:** 811 focused provider tests; 91 Azure Responses, OpenAI Agents, and ElevenLabs follow-up tests; 171 Claude Agent SDK and AI/ML API follow-up tests; 64 Azure Responses and Slack prompt-pairing follow-up tests; 373 broader focused provider tests; maintained Claude skill fixture and direct pre-network failure probes; current vendor-catalog checks.
- **Independent review:** three fresh native full-diff review passes plus a separate verifier ran on this exact base/head. Zero verified P0–P3 findings remain. Clean receipt digest: `e3a2230ba3b0d827c8ff09d8196411e65ac28ef49badf0e772e51074e2ef1bc3`.
- **Exact-head CI:** 44/44 successful, 0 pending, 0 failing.
- **PR metadata:** the assigned title remains accurate; the body has final exact-head counts and validation.
- **External gate:** GitHub reports `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED`, and `reviewDecision=REVIEW_REQUIRED`; human/codeowner approval is the only remaining gate.

No paid third-party provider calls were made, and the Promptfoo cache was not cleared.

src/providers/xai/chat.ts
```diff
@@ -406,6 +417,7 @@ export const GROK_REASONING_MODELS = [
   'grok-4-1-fast-latest',
   'grok-4-1-fast-reasoning-latest',
   // Grok Code Fast
+  'grok-build-0.1',
```
jameshiester-oai, line 475: **[P2] Preserve Grok Build's configured Chat token limit**

Adding this model and its aliases to `GROK_REASONING_MODELS` makes the inherited OpenAI body builder ignore `config.max_tokens` and only read `max_completion_tokens`; its sanitizer then also removes any passthrough `max_tokens`. On this head, a `grok-build-0.1` probe with `max_tokens: 321` emits neither field. xAI's [Chat-to-Responses parameter mapping](https://docs.x.ai/developers/model-capabilities/text/comparison) documents `max_tokens` as the Chat Completions generation cap, so existing Grok Code Fast configs silently lose their output, cost, and latency bound after this migration. Please preserve or translate `max_tokens` for this family and add a max-tokens-only request-body regression test.

site/docs/providers/google.md
```diff
@@ -440,18 +440,18 @@ tests:
...
-| `lastImage`        | string | End frame for interpolation (requires `image`)                                                              |
-| `extendVideoId`    | string | Operation ID from a previous Vertex Veo generation (Veo 3.1 only)                                           |
-| `sourceVideo`      | string | Source video input. In Google AI Studio use base64 or `file://`; in Vertex you can also use an operation ID |
+| Option             | Type   | Description                                                                                            |
+| ------------------ | ------ | ------------------------------------------------------------------------------------------------------ |
+| `aspectRatio`      | string | Video aspect ratio: `16:9` (default) or `9:16`                                                         |
+| `resolution`       | string | Video resolution: `720p` (default) or `1080p`                                                          |
+| `durationSeconds`  | number | Video duration: 4, 6, or 8 seconds                                                                     |
```
jameshiester-oai: **[P2] Keep the Veo examples within the feature-specific duration contract**

The updated table presents 4/6/8 seconds as interchangeable, but Google's [current Veo API specification](https://ai.google.dev/gemini-api/docs/video) requires `durationSeconds: 8` when using video extension, reference images, or 1080p/4K. The extension and reference-image examples below still set 6 seconds (lines 502 and 529), and `validateDuration()` accepts that pair, so these documented configs get sent upstream and fail only at the provider boundary. Please update the table/examples to use 8 seconds and preferably make validation consider `referenceImages`, extension/source-video, and resolution.

site/docs/providers/mistral.md
```diff
@@ -506,19 +475,18 @@ export MISTRAL_API_HOST="api.mistral.ai"
...
+| **Complex reasoning**      | `mistral-medium-3-5` | Adjustable reasoning effort |
+| **Code generation**        | `codestral-latest`   | Specialized for programming |
+| **Vision tasks**           | `mistral-large-2512` | Current multimodal model    |
 
 ### Context Window Optimization
 
 ```yaml
 providers:
-  - id: mistral:magistral-medium-latest
+  - id: mistral:mistral-medium-3-5
```
jameshiester-oai, line 546: **[P3] Update the context-window guidance with the model ID**

This example now targets `mistral-medium-3-5`, whose [current model card](https://docs.mistral.ai/models/model-cards/mistral-medium-3-5-26-04) lists a 256k context window, but the adjacent `max_tokens` comment still says to leave room for a 32k input context. Please change it to 256k or make the comment size-agnostic so the newly evergreen model example does not preserve stale capacity guidance.

jameshiester-oai (commented):
## Review summary

**Reviewed base:** `370c7a0c33a5154762deabbbaf8f4effdd6e0988`  
**Reviewed head:** `9223c33c7dcef28a5fab1fa993275fcef7b3a88e`

**Scope:** Incremental review from the prior authenticated reviewed SHA `642e2abfec5ef02b1b78a5d5d9578027382c7709` (confirmed ancestor), with the complete current `base...head` diff and surrounding implementations/tests rechecked for interaction. The PR remains open at the reviewed head.

**Verdict:** Two P2 correctness/documentation defects and one P3 stale comment remain. Exact-head CI is green (44 successful checks; none pending/failing); GitHub's blocked merge state is review-policy related.

### Verified findings

- **P2:** Adding Grok Build to the generic OpenAI reasoning-model classification silently discards a configured Chat Completions `max_tokens` limit.
- **P2:** The updated Veo option table and examples omit feature-specific duration constraints; the reference-image and extension examples use an upstream-invalid 6-second duration.
- **P3:** The Mistral context-window example now targets a 256k model but retains a 32k comment.

All 25 existing inline threads are resolved. I rechecked the prior automation findings at `642e2abf...`; they are fixed and are not repeated here. I found no remaining unresolved feedback to restate.

### Test gaps and simplification opportunities

The narrow fixes should add (1) a max-tokens-only Grok Build request-body regression test and (2) feature-aware Veo duration validation/tests so invalid combinations fail locally instead of at the paid provider boundary. No separate material simplification was identified.

### Validation performed

- `npx vitest run test/providers/xai/chat.test.ts test/providers/mistral.test.ts test/providers/google/video.test.ts --sequence.shuffle=false` — **157/157 passed**.
- `npm --prefix site test -- --run` — **101/101 passed**.
- `npm run tsc`, `npm run typecheck --prefix site`, and `npm run build` — **passed**.
- `SKIP_OG_GENERATION=true npm run build --prefix site` — **passed** (provider-stat fetch timed out non-fatally during prebuild).
- Biome on the four changed runtime/test files, Prettier on all changed provider docs, and `git diff --check` — **passed**.
- Changed titled-config parse/schema/provider-factory audit — **passed**; intentional MCP/custom-file executions were excluded.
- No-cache echo CLI eval — **1/1 passed**, exact output, score 1, no error.
- No-cache Mistral stub CLI eval using the documented `mistral-medium-3-5` ID — **1/1 passed**; request reached `/v1/chat/completions`, token usage was 100 input/200 output, and calculated cost was **$0.00165**.
- Direct request-body probe with `max_tokens: 321` — reproduced a Grok Build body containing neither `max_tokens` nor `max_completion_tokens`.
- Direct Veo validator probe — reproduced `{ valid: true }` for the invalid 6-second reference/extension combination's model/duration pair.

**Limitations:** No paid/live provider call was made, and MCP/custom file-provider examples were not executed. The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable in this environment (exit 127); the complete fallback used three independent read-only specialist passes plus orchestrator validation.

jameshiester-oai (commented):
Reviewed base `ea63e1ccb4d5b9915e82d13366461fb367925709` at head `a78f263ec94245815b7de358c10d8832cc1a507d`.

Scope: incremental review from previously completed ancestor `9223c33c7dcef28a5fab1fa993275fcef7b3a88e`, plus the complete current base...head diff for interaction and regression context.

Verdict: three P2 correctness/documentation issues remain. Exact-head CI is green across test, build, style, docs, integration, smoke, CodeQL, and Codecov checks.

### Verified findings

**P2**
- Perplexity search options now affect per-request/search costs, but the provider still replaces the API's authoritative total with a token-only estimate. High-context Sonar and Deep Research evaluations therefore underreport cost.
- The refreshed Perplexity documentation and runnable example still advertise `response_format.type: regex`, which the current Sonar request schema does not accept.
- The new Veo resolution validator accepts 4K for `veo-3.1-fast-generate-001`, while Google's stable Fast model card lists only 720p and 1080p output.

No P0/P1 findings.

### Existing feedback

All 29 current review threads are resolved. I rechecked the prior Grok Build token forwarding and Veo duration/routing findings; the current head fixes them and adds focused regression coverage. I did not duplicate resolved feedback.

A model-capability table would be safer than substring checks for Veo transport/model differences. Perplexity cost handling should prefer the vendor-returned total and keep the estimator only as a fallback. Add negative coverage for stable Fast 4K and the current Sonar response-format enum.

### Verification performed

- `source ~/.nvm/nvm.sh && nvm use && node node_modules/vitest/vitest.mjs run test/providers/google/video.test.ts test/providers/xai/chat.test.ts test/providers/perplexity.test.ts test/providers/openai/billing.test.ts test/providers/openai/util.test.ts test/providers/bedrock/openaiResponses.test.ts test/providers/index.test.ts test/providers/registry.test.ts --no-cache` — 8 files, 779 tests passed.
- `source ~/.nvm/nvm.sh && nvm use && npm run test:app -- src/pages/eval/components/EvalOutputCell.test.tsx --run` — 1 file, 130 tests passed.
- Five changed Google Video/Perplexity example YAML files parsed and passed `UnifiedConfigSchema`.
- Direct validator probe confirmed `validateResolution('veo-3.1-fast-generate-001', '16:9', '4k')` returns `{ valid: true }`.
- Current OpenAI pricing/Fast-tier behavior was checked against the official live pricing and Fast mode documentation; the July 30 Terra/Luna rates and legacy `priority` alias are correct.

### Limitations

No paid Google, Perplexity, xAI, Bedrock, or OpenAI requests were made. The requested PR Review Toolkit command was unavailable in this environment, so I performed the six-lens read-only fallback with three independent specialist reviews. Full clean-install/typecheck evidence is supplied by exact-head CI; local focused runs used the existing compatible dependency overlay.

src/providers/perplexity.ts
```diff
@@ -138,6 +148,30 @@ export class PerplexityProvider extends OpenAiChatCompletionProvider {
...
+    const providerPassthrough = this.config?.passthrough as Record<string, any> | undefined;
+
+    for (const field of PERPLEXITY_PASSTHROUGH_FIELDS) {
+      const value =
+        promptPassthrough?.[field] ??
+        promptConfig?.[field] ??
+        providerPassthrough?.[field] ??
+        this.config?.[field];
+      if (value !== undefined) {
+        result.body[field] = value;
```
jameshiester-oai, line 182: **[P2] Preserve Perplexity's authoritative total cost.** Forwarding `web_search_options` makes the request fee configuration-dependent, but `callApi` below still replaces the response cost with `calculatePerplexityCost`, which only uses prompt/completion tokens. Perplexity returns `usage.cost.total_cost` plus request, reasoning, citation, and search-query components ([response schema](https://docs.perplexity.ai/api-reference/sonar-post)); for example, a high-context Sonar request incurs a request fee that the current result omits. Prefer the numeric vendor total in `calculateResponseCost` (or preserve it before normalization), with the existing estimator only as a fallback.

site/docs/providers/perplexity.md
```diff
@@ -123,7 +123,7 @@ providers:
       response_format:
         type: 'regex'
         regex:
-          regex: "(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)"
+          regex: '(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)'
```
jameshiester-oai: **[P2] Remove the unsupported regex response format.** The current Sonar request schema accepts only text/default output or `type: json_schema`; `regex` is no longer one of the documented variants ([current API schema](https://docs.perplexity.ai/api-reference/sonar-post)). This refreshed example—and the matching runnable YAML—will send users into a paid request that fails validation upstream. Replace it with a JSON Schema example or remove the regex section.

src/providers/google/video.ts
```diff
@@ -107,20 +111,63 @@ export function validateDuration(
...
+  }
+
+  if ((config.extendVideoId || config.sourceVideo) && resolution !== '720p') {
+    return {
+      valid: false,
+      message: `Video extension requires 720p resolution (received ${resolution}).`,
+    };
+  }
+
+  if (resolution === '4k' && (!model.includes('veo-3.1') || model.includes('lite'))) {
```
jameshiester-oai: **[P2] Distinguish the stable Fast model's resolution limits.** This substring check accepts 4K for `veo-3.1-fast-generate-001` (confirmed by a direct validator probe), but Google's current stable Fast model card lists only 720p and 1080p output; 4K is listed for the standard stable model and preview variants ([Veo 3.1 model card](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models/veo/3-1-generate-preview)). The invalid config passes local validation and reaches a paid request before failing. Encode model-specific capabilities and add a negative stable-Fast 4K test.

src/providers/google/video.ts
```diff
@@ -1,7 +1,9 @@
 import crypto from 'crypto';
 import fs from 'fs';
+import path from 'path';
 
 import { storeBlob } from '../../blobs';
+import cliState from '../../cliState';
```
jameshiester-oai: [P2] Keep config-path state out of the provider layer

This new `cliState` import grows the guarded `providers -> legacy-runtime` dependency edge from 324 to 325. Exact-head Style Check reaches `npm run architecture:check` and deterministically exits 1 on that regression; the earlier Biome panics are not the terminal cause. Resolve these media references before provider construction or pass an explicit base path through provider options/context, rather than adding another dependency on the global CLI singleton. Updating the baseline would only bless the new coupling and should not be the default fix.

test/providers/google/video.test.ts
```diff
@@ -127,6 +144,71 @@ describe('GoogleVideoProvider', () => {
...
+      );
+    });
+  });
+
+  describe('config-relative media paths', () => {
+    it.each([
+      ['loadImageData', 'image', 'file://assets/start-frame.png'],
+      ['loadVideoData', 'video', 'file://assets/veo-input.mp4'],
+    ])('resolves %s for a %s relative to the config directory', (method, _kind, fileRef) => {
+      cliState.basePath = path.join('/tmp', 'google-video');
```
jameshiester-oai: [P2] Use a platform-native absolute base path in this regression

`path.join('/tmp', 'google-video')` is only root-relative on Windows. The implementation calls `path.resolve`, producing `D:\\tmp\\google-video\\...`, while this test's strict mock expects `\\tmp\\google-video\\...`; exact-head Windows shard 2 therefore fails here (1 failed, 2,208 passed). Build the fixture base with `path.resolve(...)` or a real temporary directory and derive the expectation from that absolute path so the test exercises config-relative resolution on every supported platform.

jameshiester-oai (commented):
Reviewed base `82ca3c24ec445cf1734face46042c187b659b954` through exact head `2f835310d875f36563d2b07d02cbf4a0ca00cd2d`.

Scope: incremental review from authenticated prior-reviewed ancestor `a78f263ec94245815b7de358c10d8832cc1a507d`, with the full current 180-path base...head diff inspected for interaction and regression context.

Verdict: two new P2 findings. Exact-head CI is red for two deterministic PR-attributable failures: the architecture dependency ratchet and a Windows-only assertion in the new config-relative media test. The Biome internal panics printed earlier in Style Check are incidental warnings; `npm run architecture:check` is the job's terminal exit-1 cause.

- P2: `src/providers/google/video.ts` adds a provider-layer dependency on the legacy-runtime `cliState` singleton, increasing the guarded `providers -> legacy-runtime` edge from 324 to 325 and failing the architecture gate.
- P2: the new config-relative media-path test builds a POSIX-rooted but Windows-root-relative expected path, so production resolves `D:\\tmp\\...` while the strict mock expects `\\tmp\\...`; Windows shard 2 fails before validating the intended behavior.

Existing feedback, without duplication:
- The unresolved Perplexity authoritative-cost thread remains current.
- The current source-video route documentation thread remains open.
- Prior review feedback about slash-qualified retired OpenAI IDs remains applicable: validation normalizes the basename while retired-route selection still receives the full namespaced value.
- The stable Fast 4K validation and obsolete Perplexity regex-example issues are fixed/outdated at this head.

- Resolve config-relative media references at the config/provider construction boundary or pass an explicit base path through provider options/context; avoid adding another provider dependency on global mutable CLI state.
- Use a platform-native absolute temporary base in the media-path regression and add canonical Windows file-URL coverage through the repository's existing `parseFileUrl` normalization path.

Validation:
- `npx vitest run` across 14 changed provider suites (Google Video, Perplexity, OpenAI billing/util/Responses/transcription, Bedrock OpenAI/Anthropic/pricing, xAI chat/voice, Anthropic messages/util, registry) — 14 files, 1,191 tests passed locally.
- `npm run test:app` for the three changed target-configuration suites — 3 files, 71 tests passed.
- `npm run local -- validate config -c examples/provider-perplexity/promptfooconfig.structured-output.yaml` — valid.
- `npm run local -- validate config -c examples/anthropic/sonnet-5/promptfooconfig.yaml` — valid.
- `npm run architecture:check` — reproduced `providers -> legacy-runtime: 325 imports (was 324)` and exit 1.
- Exact-head Windows job `91175890391` — reproduced the new path test failure at line 205; 2,208 tests passed and one failed.
- `git diff --check 82ca3c24ec445cf1734face46042c187b659b954...HEAD` — passed.
- Exact-head CI — all other current build, test, docs, smoke, integration, platform, and CodeQL checks passed.

Limitations:
- Exact-lock local install remains blocked by Socket for recently published dependencies; local QA used a compatible dependency overlay. Exact-head CI supplies clean-install typecheck/build evidence.
- No paid Google, Perplexity, xAI, Bedrock, Anthropic, or OpenAI requests were made; no browser session or full local repository suite was run.
- `/pr-review-toolkit:review-pr all parallel` is unavailable on this host (exit 127); its six lenses were applied manually and reconciled with three bounded read-only specialist reviews.

src/providers/families/aws.ts
```diff
@@ -4,18 +4,53 @@ import { isBedrockMantleResponsesModel } from '../bedrock/mantle';
...
+            : NOVA_SONIC_MODEL_IDS.has(bareModelName)
+              ? bareModelName
+              : (modelType === 'nova-sonic' || modelType === 'nova-2-sonic') &&
+                  NOVA_SONIC_MODEL_IDS.has(modelName)
+                ? modelName
+                : undefined;
+      if (isUnsupportedNovaSonicGeoId(bareModelName) || isUnsupportedNovaSonicGeoId(modelName)) {
+        throw new Error(
+          `Amazon Bedrock model "${bareModelName}" does not support geo inference IDs. ` +
+            'Use the bare "bedrock:amazon.nova-2-sonic-v1:0" model ID in a supported region.',
```
jameshiester-oai, line 57: **P2 — Preserve the requested Nova generation in this correction.** `isUnsupportedNovaSonicGeoId` matches both `us.amazon.nova-sonic-v1:0` and `us.amazon.nova-2-sonic-v1:0`, but every match is told to use Nova 2. Through the public `loadApiProvider` path, requesting `bedrock:us.amazon.nova-sonic-v1:0` currently throws an instruction to use `bedrock:amazon.nova-2-sonic-v1:0`, silently directing a v1 caller to a different model generation. Derive the bare ID from the matched request (v1 -> `amazon.nova-sonic-v1:0`, v2 -> `amazon.nova-2-sonic-v1:0`) and assert the full error for both cases; the existing substring-only parameterized test does not catch this mismatch.

src/providers/bedrock/nova-sonic.ts
```diff
@@ -128,6 +130,17 @@ export class NovaSonicProvider extends AwsBedrockGenericProvider implements ApiP
       return this.bedrockClient;
     }
 
+    const region = this.getRegion();
```
jameshiester-oai, line 153: **P3 — Validate the region before creating session state.** `callApi` creates a session before reaching `getBedrockClient`, so this new purely local validation error still enters `finally` and awaits `endSession`'s two one-second sleeps. A direct `NovaSonicProvider('amazon.nova-2-sonic-v1:0', { config: { region: 'eu-west-1' } }).callApi(...)` probe returned after 2,007 ms even though no request was attempted. Move model/region validation ahead of `createSession` (or make cleanup skip waits for an unstarted session) and cover the invalid-region path with fake timers so bad configuration fails immediately.

jameshiester-oai (commented):
Reviewed base `26b725bd9496351ef269380c9fd83b9c2c338a0e` through head `d23d14048401d3e676a21b8a5e6d250caf4b09e6`. Scope was incremental from the previously reviewed ancestor `2f835310d875f36563d2b07d02cbf4a0ca00cd2d`, with the complete current base...head diff and surrounding provider contracts also reviewed.

Verdict: two actionable Nova Sonic regressions remain. Exact-head CI is green (44/44 reported checks successful), but the focused public-path probes below expose one misleading model migration instruction and one avoidable two-second local-validation delay.

Verified findings

- **P2** `src/providers/families/aws.ts:37` — geo-prefixed Nova Sonic v1 IDs are rejected with an instruction to switch to Nova 2 rather than retain the requested v1 model.
- **P3** `src/providers/bedrock/nova-sonic.ts:133` — unsupported Nova 2 regions are validated only after a session is created, so `finally` runs the two one-second cleanup waits before returning a purely local configuration error.

Existing feedback

- The earlier Perplexity total-cost thread remains unresolved in GitHub, but the current head now preserves the authoritative vendor total, so I did not duplicate it.
- The older regex, Stable Fast 4K, source-video documentation, architecture, and Windows-path threads are outdated against this head and were rechecked rather than repeated.

- Derive the geo-ID correction from the requested Nova generation instead of hard-coding a Nova 2 ID. A parameterized assertion for both v1 and v2 would prevent the current cross-generation message bug.
- Perform model/region validation before allocating session state; a fake-timer assertion can then guarantee immediate failure and no queued cleanup.

Validation performed

- `npx vitest run test/providers/registry.test.ts test/providers/bedrock/nova-sonic.test.ts --sequence.seed=9792` — 180/180 passed.
- Specialist focused run across 11 provider suites — 1,583/1,583 passed.
- All 17 changed YAML configs validated with credentials scrubbed.
- Public `loadApiProvider` probes reproduced the v1-to-v2 instruction and confirmed the v2 instruction.
- Direct `NovaSonicProvider.callApi` probe for Nova 2 in `eu-west-1` returned the expected local region error after 2,007 ms.
- Fresh `npx tsc --noEmit --tsBuildInfoFile /tmp/pr9792-tsconfig.tsbuildinfo` — passed. (`npm run tsc` initially consumed a stale ignored build-info cache after the dependency-version switch; the fresh exact-head check passed.)
- `npx biome check` on the affected Nova/registry files — passed with two existing complexity warnings and no errors.
- Exact-head dependency install completed from the lockfile; the worktree remained clean.

Limitations

- No paid Bedrock, OpenAI, Anthropic, or other live vendor request was sent. Catalog and routing behavior were exercised through local construction, validation, and mocked/unit paths.
- The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable in this environment (exit 127), so the closest read-only specialist review fallback was used and all candidates were independently validated.

src/providers/openai/billing.ts
```diff
@@ -919,7 +1026,22 @@ export function calculateOpenAIUsageCost(
...
-  const modelRates = getModelRates(modelName, tier, usage.totalInputTokens);
+  const bedrockMantleTextRates = getBedrockMantleTextRates(modelName, options.apiUrl, tier);
+  const catalogRates = bedrockMantleTextRates
+    ? { text: bedrockMantleTextRates }
+    : getModelRates(modelName, tier, usage.totalInputTokens);
+  const explicitTextRates = getCompleteTextCostOverrides(config);
+  const explicitAudioRates = getCompleteAudioCostOverrides(config);
+  // Complete explicit rates are authoritative when no catalog entry applies, including for
+  // namespaced gateway models. Partial overrides cannot establish a safe fallback rate table.
+  const modelRates =
```
jameshiester-oai, line 1037: **P2 — Merge complete audio overrides for catalog-known text models.** `catalogRates` wins wholesale here, so `explicitAudioRates` is only attached for unknown models. For a gateway that reports `gpt-4o` with audio token details and supplies complete `audioInputCost`/`audioOutputCost`, `rates.audio` remains absent and the guard below returns `undefined` instead of the configured cost (direct probe: expected `0.000386`). Merge the complete explicit audio table when the selected catalog entry lacks audio rates, and add a recognized-model regression beside the new unknown-gateway audio tests.

src/providers/openai/billing.ts
```diff
@@ -937,6 +1059,10 @@ export function calculateOpenAIUsageCost(
     return 0;
   }
 
+  if (!rates.audio && (usage.audioInputTokens > 0 || usage.audioOutputTokens > 0)) {
```
jameshiester-oai, line 1062: **P2 — Fail closed for unknown gateway image usage too.** This new guard covers missing audio rates, but the new explicit-text fallback also admits unknown namespaced models with `image_tokens`; those tokens then fall through and are billed as text. A direct probe for `third-party/gpt-4o` with text overrides plus 9 image tokens returns `0.000172`, while the prior base returned `undefined` for the unknown model. Require complete image rates before pricing that fallback, or return `undefined`, and add a sibling image regression.

jameshiester-oai (commented):
Reviewed base `b5f8ea3b0a307adb91c31ed4e13a3d6896902841` through exact head `67330d955fbe20e41938f71a0a8d49bcf3229e36`.

Scope: incremental review from authenticated prior-reviewed ancestor `d23d14048401d3e676a21b8a5e6d250caf4b09e6`, with the complete current base...head diff and surrounding provider/billing contracts rechecked for interaction.

Verdict: two new P2 billing defects remain. Exact-head CI is green across the reported build, test, docs, integration, smoke, style, CodeQL, security-scan, and coverage checks.

- P2: complete explicit audio rates are ignored for catalog-recognized text models, causing valid audio usage to return no cost.
- P2: the new unknown-gateway explicit text-rate fallback prices image tokens as text instead of failing closed when no image rate exists.

Existing feedback:
- Four current unresolved non-outdated threads remain: Perplexity authoritative total cost, Nova generation-preserving correction, Nova invalid-region cleanup delay, and Vertex model-default region routing. I rechecked them and did not duplicate them.

- Centralize modality-rate selection so explicit complete modality overrides merge consistently with catalog text rates and unknown fallbacks fail closed for every reported modality.
- Add sibling regressions for a catalog-known text model with explicit audio rates and an unknown namespaced model with image usage but no image rate.

QA:
- `node node_modules/vitest/vitest.mjs run test/providers/openai/billing.test.ts test/providers/openai/util.test.ts --sequence.shuffle=false` — 2 files, 278 tests passed.
- Direct public-function probe for `gpt-4o` with complete text/audio overrides and audio token details returned `undefined` instead of expected `0.000386`.
- Direct public-function probe for `third-party/gpt-4o` with complete text overrides and image token details returned `0.000172`; the prior base returned `undefined` for the same unknown model path.
- `git diff --check b5f8ea3b0a307adb91c31ed4e13a3d6896902841...HEAD` — passed.
- Exact `/pr-review-toolkit:review-pr all parallel 9792` — unavailable on this host (exit 127); I applied the requested review lenses manually and reconciled three bounded read-only specialist reports.

Limitations:
- No paid/live provider requests were made.
- Exact-lock `npm ci` remains blocked by Socket Security Policy for `hono-4.12.34.tgz`; focused QA used an existing compatible Promptfoo dependency overlay. Exact-head CI supplies clean-install evidence.

src/providers/perplexity.ts
mldangelo-oai: Verified on `eb7ec7b70`. Merged #10750 supplies the authoritative finite, nonnegative usage.cost.total_cost, including zero. That behavior is retained here; absent or invalid totals remain unavailable so partial token estimates do not understate cost.

site/docs/providers/perplexity.md
mldangelo-oai: Verified on `eb7ec7b70`. The Perplexity guide and runnable structured-output example use JSON Schema; the unsupported regex response format is removed.

src/providers/google/video.ts
mldangelo-oai: Verified on `eb7ec7b70`. Stable Fast Veo 3.1 explicitly rejects 4K before request construction, with a negative regression for that model.

site/docs/providers/google.md
```diff
@@ -465,23 +468,23 @@ tests:
...
+| `durationSeconds`  | number | Veo 3.x: 4, 6, or 8 seconds; Veo 2: 5, 6, or 8. Extension, reference images, and 1080p/4k require 8    |
+| `personGeneration` | string | Person generation mode: `allow_adult` or `dont_allow`                                                  |
+| `negativePrompt`   | string | Concepts to avoid in the generated video                                                               |
+| `referenceImages`  | array  | Up to 3 reference images (file paths or objects; Veo 3.1 and 3.1 Fast, not Lite)                       |
+| `image`            | string | Source image for image-to-video generation                                                             |
+| `lastImage`        | string | End frame for interpolation (requires `image`)                                                         |
+| `extendVideoId`    | string | Legacy Vertex operation-ID input; availability depends on the selected Vertex model                    |
+| `sourceVideo`      | string | Source video input for Veo 3.1 and 3.1 Fast, not Lite. Use base64, `file://`, or a Vertex operation ID |
```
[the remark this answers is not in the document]
mldangelo-oai: Verified on `eb7ec7b70`. The input contracts are documented separately: native Gemini accepts reusable Gemini video URI, file, or base64 inputs; Vertex accepts gs:// MP4. Both routes reject operation IDs. This supersedes the older operation-ID recommendation.

src/providers/google/video.ts
mldangelo-oai: Verified on `eb7ec7b70`. Media lookup uses provider/request basePath with a cwd fallback. The provider no longer imports the CLI global state.

test/providers/google/video.test.ts
mldangelo-oai: Verified on `eb7ec7b70`. The regression derives its absolute base and expected paths with path.resolve, preserving platform-native paths.

src/providers/families/aws.ts
mldangelo-oai: Verified on `eb7ec7b70`. Unsupported geographic Nova Sonic IDs now derive the bare model ID from the requested generation, preserving both the v1 and v2 correction targets.

src/providers/bedrock/nova-sonic.ts
mldangelo-oai: Verified on `eb7ec7b70`. Nova Sonic validates the model and region before creating session state, so invalid-region configuration does not enter delayed session teardown.

src/providers/google/base.ts
```diff
@@ -181,7 +182,10 @@ export abstract class GoogleGenericProvider implements ApiProvider {
    */
   getRegion(): string {
     const hasApiKey = Boolean(this.getApiKey());
-    return GoogleAuthManager.resolveRegion(this.config, this.env, hasApiKey);
+    const modelDefaultRegion = this.isVertexMode
+      ? getVertexModelDefaultRegion(this.modelName)
+      : undefined;
+    return GoogleAuthManager.resolveRegion(this.config, this.env, hasApiKey, modelDefaultRegion);
```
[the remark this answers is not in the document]
mldangelo-oai: Verified on `eb7ec7b70`. Explicit Vertex Chat inherits model-default region selection, with explicit configuration and environment values retaining precedence.

src/providers/openai/billing.ts
mldangelo-oai: Verified on `eb7ec7b70`. Complete explicit audio rates supplement a catalog-known model’s text rates when its catalog audio rates are absent. Missing required audio rates still leave cost unavailable.

src/providers/openai/billing.ts
mldangelo-oai: Verified on `eb7ec7b70`. Unknown gateway image usage cannot fall through to text-only override pricing when image rates are absent.

site/docs/providers/togetherai.md
```diff
@@ -66,72 +65,41 @@ config:
...
-### Free Endpoints
-
-Together AI offers free tiers with reduced rate limits:
-
-- `meta-llama/Llama-3.3-70B-Instruct-Turbo-Free`
-- `meta-llama/Llama-Vision-Free`
-- `deepseek-ai/DeepSeek-R1-Distill-Llama-70B-Free`
-
-For a complete list of all 200+ available models and their specifications, refer to the [Together AI Models page](https://docs.together.ai/docs/inference-models).
+## Current Serverless Models
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in `eb7ec7b70`. The previous Together section links now point to the current model catalog or deprecation guidance. The documentation build confirms all eight compatibility anchors are present exactly once.

mldangelo-oai:
@codex review

site/docs/providers/hyperbolic.md
```diff
@@ -162,11 +145,12 @@ For a custom audio endpoint, set `apiBaseUrl` to its base URL, including any ver
 
 ### Text Generation Example
 
-```yaml
+```yaml title="promptfooconfig.yaml"
```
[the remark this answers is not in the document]
mldangelo-oai: The companion prompt is already provided on this page: [the complete `prompts/coding_assistant.json` template](https://github.com/promptfoo/promptfoo/blob/9a2696519df2cbc8f2cb07013dd96938ce1cb2e7/site/docs/providers/hyperbolic.md#L199-L212). [Getting Started](https://github.com/promptfoo/promptfoo/blob/9a2696519df2cbc8f2cb07013dd96938ce1cb2e7/site/docs/providers/hyperbolic.md#L231-L234) explicitly tells readers to use the text config with that companion template. The missing-file premise does not apply; no source change is needed.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

Current head: `82bfa5d3abf9dfa266ce9cb0f2da17bca9fbd4f4`.

mldangelo-oai:
@codex review

src/providers/google/video.ts
```diff
@@ -471,17 +710,30 @@ export class GoogleVideoProvider implements ApiProvider {
...
+        });
+        if (error) {
+          return { error };
+        }
+        instance.video = { inlineData: { mimeType: 'video/mp4', data } };
+      } else {
+        return {
+          error:
+            'Google AI Studio Veo video extension requires a previous Veo video URI, file:// path, or base64 bytes.',
         };
```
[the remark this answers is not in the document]
mldangelo-oai: The configured file read is intentional under the [OSS security model](https://github.com/promptfoo/promptfoo/blob/070ae001fb09e050c8ac8661fc6ff695d35131ae/SECURITY.md#security-model). `sourceVideo` comes from explicit provider/prompt configuration, not automatically from rendered prompt text. `file://` loads the configured local media, including paths outside `basePath`; raw base64 performs no host-file read. The [provider documentation](https://github.com/promptfoo/promptfoo/blob/070ae001fb09e050c8ac8661fc6ff695d35131ae/site/docs/providers/google.md#video-generation) supports these inputs. No supported OSS isolation bypass was established, so I am retaining the documented behavior. This assessment does not cover unseen hosted tenant authorization.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

Please review the current head `1f7bfe8160113eee94966b9cfdd8d51f4dd4586a`.

src/providers/registry.ts
```diff
@@ -1354,9 +1468,12 @@ export const providerMap: ProviderFactory[] = [
...
       providerPath: string,
       providerOptions: ProviderOptions,
-      _context: LoadApiProviderContext,
+      context: LoadApiProviderContext,
     ) => {
-      return new VoyageEmbeddingProvider(providerPath.split(':')[1], providerOptions);
+      return new VoyageEmbeddingProvider(providerPath.split(':')[1], providerOptions.config, {
+        ...context.env,
+        ...providerOptions.env,
+      });
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in bfd526a7d3ec6f0f8ac0a05afee0c5c779e7b008. Voyage now checks the scoped custom variable before the ambient one. Explicit keys and empty-value fallback are preserved. All 13 provider tests passed, including five loader-to-request precedence cases; five built-CLI fixture evaluations also verified the selected request key and exported results.

mldangelo-oai:
@codex review

Please review the current head bfd526a7d3ec6f0f8ac0a05afee0c5c779e7b008, including the Voyage custom credential precedence correction.

mldangelo-oai:
@codex review

Please review the current head `a2430a3d9d7f473e555228c9acd25264a0f4ba79`, including the isolated provider-family error test.

mldangelo-oai:
@codex review the current head b1d49596392830c9dd10a72542f60550d6eec58b, including the main merge and retained provider model, routing, and cost contracts.

jameshiester-oai (commented):
Reviewed exact head `ccc729726edca9a0c53f1c175d3941ff78318cde` against base `411408947caf9d9a02e1d59760897e0513d6e1e6`, including the current Google runtime/media changes and existing feedback. Reconciled independent Google, provider-risk, and catalog/test reviews. No distinct new actionable finding to add inline.

Existing findings remain relevant:
- [Mantle effective-model routing](https://github.com/promptfoo/promptfoo/pull/9792#discussion_r4001051231): a current-head construction probe keeps `/openai/v1` for an OpenAI provider whose passthrough model is `zai.glm-4.6`, while that effective model's route helper returns `/v1`.
- [Azure August GPT Chat metadata](https://github.com/promptfoo/promptfoo/pull/9792#discussion_r4001051236): a current-head cost-helper probe returns undefined for `gpt-chat-latest-2026-08-06`.
- [Agent lifecycle-hook tool mutation](https://github.com/promptfoo/promptfoo/pull/9792#discussion_r4001440082): the source still shares lifecycle emitters after the one-time mock-tool wrapping pass. I did not independently execute that exploit in this review.

Validation: all 7,661 tests in 82 changed backend suites and 180 tests in three changed UI suites passed. Core/declaration/frontend builds, root TypeScript, and the documentation production build passed. The postbuild launcher encountered sandbox IPC `EPERM`; the same script passed via `node --import tsx scripts/postbuild.ts`. All 43 currently reported CI checks are green.

Actual built-CLI loopback checks verified Voyage's provider-scoped custom key wins over an ambient key, `/v1/embeddings` receives `voyage-3.5`, a passing similarity result carries 14 assertion tokens, and a separate HTTP 400 becomes an exported error. Cohere's v2 workflow converted legacy history/preamble into system/user/assistant/user messages, called `/v2/chat`, and exported the expected answer with 12 tokens and two cached tokens. Synthetic credentials and isolated temporary state were used; no vendor calls or billing verification occurred.

Refreshed the cited primary release references for [Gemini 3.8 Flash](https://ai.google.dev/gemini-api/docs/models/gemini-3.8-flash) and [DeepSeek's September release](https://api-docs.deepseek.com/updates/). This is not a live availability claim for every catalog entry.

PR #9799, opened 2026-06-18
fix(util): match JavaScript/TypeScript file extensions case-insensitively

nyxst4ck:
## What

`isJavascriptFile()` matched extensions with a case-sensitive regex, so a `file://` reference with an uppercase extension (e.g. `provider.JS`, `transform.TS`) was **not** recognized as a JS/TS module.

## Why this is a bug

`parsePathOrGlob()` derives the extension via `path.parse(...).ext`, which **preserves the original case**. That value is passed straight into `isJavascriptFile()` by every loader that gates JavaScript handling on it:

- `src/prompts/index.ts` (prompt files)
- `src/providers/registry.ts`, `src/providers/transformUtils.ts`, `src/providers/google/util.ts` (providers / transforms)
- `src/assertions/index.ts`, `src/matchers/rubric.ts` (custom assertions / rubrics)
- `src/util/transform.ts`, `src/util/file.ts`, `src/util/functions/loadFunction.ts`, `src/util/testCaseReader.ts` (transforms, function/test-case loading)
- `src/util/config/load.ts`, `src/redteam/strategies/index.ts`, `src/validators/redteam.ts` (config / redteam)

On case-insensitive file systems (Windows, default macOS) a file literally named `Provider.JS` or `helper.Ts` exists and resolves fine, but `isJavascriptFile('.JS')` returned `false`, so the reference fell through the JS branch and was silently treated as a raw string / errored instead of being loaded as a module.

The three sibling helpers in the same file — `isImageFile`, `isVideoFile`, `isAudioFile` — already lowercase before matching. This aligns `isJavascriptFile` with that behavior.

## Fix

Add the case-insensitive (`i`) flag to the extension regex. The change is strictly additive:

- lowercase extensions behave exactly as before,
- non-JS extensions (e.g. `.TXT`, `.py`) still return `false`.

## Test

Extends the existing `isJavascriptFile` test in `test/util/file.test.ts` with uppercase/mixed-case cases (`test.JS`, `test.Ts`, `test.MJS`, `test.CTS` → `true`; `test.TXT` → `false`).

Verified red on `main` (the new case `test.JS` fails with `expected false to be true`) and green with the fix. `npm run tsc` passes; Biome lint/format clean.


ianw-oai (approved):

ianw-oai:
Thank you!

PR #9800, opened 2026-06-18
test(util): add coverage for printErrorInformation

2ashy4u:
## Summary

Adds unit tests for `src/util/errors/index.ts`, which previously had 0% coverage. The file is small but exercises several distinct code paths in `printErrorInformation` and its internal `errorFileHasContents` helper — this PR pins down that behavior so future refactors don't silently regress it.

## What's covered

- No / empty / `undefined` `errorLogFile` arg → no output, no `fs` call
- `ENOENT` from `fs.statSync` → silent (expected when the user never hit an error)
- Other `fs` errors (e.g. `EACCES`) → logged via `logger.debug`, not surfaced as an info message
- File exists but is empty (`size === 0`) → no output
- Path exists but is not a regular file → no output
- File has contents, no debug log → info message contains the error log path, no `Debug log:` line
- File has contents, debug log provided → info message contains both paths

## Coverage

`src/util/errors/index.ts`: **0% → 100%** (statements / branches / functions / lines).

## Test plan

- `npx vitest run test/util/errors/index.test.ts` → 9 passed
- `npm run tsc` → clean
- `npm run l` → clean
- `npm run f` → clean

### DIFF
--- a/test/util/errors/index.test.ts
+++ b/test/util/errors/index.test.ts
@@ -1,0 +1,124 @@
+import fs from 'fs';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+import logger from '../../../src/logger';
+import { printErrorInformation } from '../../../src/util/errors';
+
+vi.mock('fs');
+
+describe('printErrorInformation', () => {
+  let infoSpy: ReturnType<typeof vi.spyOn>;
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
Thank you!

2ashy4u:
No worries, I am going to look more into this codebase and see what i can do to contribute.

PR #9824, opened 2026-06-20
fix(assertions): correct inverse JSON failure messages

he-yufeng:
## Problem

Schema-less `not-is-json` and `not-contains-json` assertions already fail with the correct verdict and score when JSON is present, but their failure reasons describe the opposite, positive assertion:

```
Expected output to be valid JSON
Expected output to contain valid JSON
```

Those contradictory reasons are surfaced in CLI/API results, exports, persistence, and the UI.

## Fix

- Make the fallback failure reasons in `handleIsJson` and `handleContainsJson` respect the existing `inverse` flag.
- Keep positive assertions and all schema-specific messages unchanged.
- Cover both negated, schema-less cases through the public `runAssertion` path.
- Correct the existing `contains-json` YAML-schema test so it exercises an actual YAML string instead of the object-schema fixture.

## User value

Failed negated JSON assertions now explain the requirement they actually evaluated, so users can diagnose failures without mentally reversing the message.

## Compatibility

There are no configuration, schema, default, API-shape, parsing, verdict, or scoring changes. Only two human-readable failure strings are corrected. External consumers that exact-match either previous incorrect string will observe the intentional change; historical stored evaluations retain their original text.

## Validation

- Exact live head: `bb33430d9d1f2b48c279ac3bcc4e243ab9b8f84d`; current `main`: `56ee73063bb5088536e4d510861651484fc49a3a`
- Ancestry: the contributor replayed the five-commit series onto current `main`; `git range-diff` maps all five commits 1:1 to the previously reviewed `461ae22a` series, all stable patch IDs and both affected blobs match, and no contributor work was lost
- Freshness: current `main` is an ancestor; the branch is 5 commits ahead and 0 behind
- Final impact: production `+2/-2`; tests `+39/-1`; no dependencies or new abstractions
- `runAssertion.test.ts`: 176/176 passed on Node 24.18.0, including shuffled reruns
- `src/assertions/json.ts` focused coverage: 81.39% statements/lines, 90.90% branches, 100% functions; the critical-path coverage ratchet passed
- Real local CLI evals with `--no-cache`: 12/12 success cases passed; 9/9 intended failures produced the expected aggregate/component/exported reasons with zero runtime errors; boundary, adversarial, schema, compatibility, and concurrency-4 paths were inspected
- TypeScript, Biome, Prettier, and `git diff --check`: clean
[...]

PR #9838, opened 2026-06-22
fix(assertions): honor the inverse flag in trace, tool-call, and model-graded assertions

mldangelo:
## Problem

Several assertion handlers ignore the `inverse` flag entirely. Because every base assertion type automatically gets a `not-<type>` variant (`NotPrefixedAssertionTypesSchema` in `src/types/index.ts`), a user can write any of these `not-*` assertions and it is accepted — but the handler silently returns the **same verdict as the non-negated form** instead of negating it. There is no central inversion in `runAssertion`; each handler is responsible for honoring `inverse`, and these did not:

- `trace-span-count`
- `trace-span-duration`
- `trace-error-spans`
- `is-valid-function-call` / `is-valid-openai-function-call`
- `is-valid-openai-tools-call`
- `model-graded-closedqa`

So e.g. `not-trace-span-count` returned a result identical to `trace-span-count`, and `not-is-valid-openai-tools-call` passed when the call *was* valid.

This is the same bug class as #9824 (which fixes the directional **messages** for `not-contains-json`/`is-json`/`meteor`), but here the **verdict** itself was wrong, not just the wording.

## Fix

Each handler now inverts **inline**, matching the convention already used across `src/assertions/` (`contains.ts`, `equals.ts`, `classifier.ts`, `sql.ts`) — no new shared helper:

- The deterministic handlers (trace × 3, function/tool-call) flip the verdict at the result site. Trace reasons keep describing the measurement (they are facts, not directional claims); the tool-call handlers — whose pass reason is the generic `"Assertion passed"` — get a directional message so an inverse failure isn't mislabeled.
- `model-graded-closedqa` mirrors `handleLlmRubric`: it skips inversion on a grader error (`metadata.graderError`) and clamps `1 - score` on the inverse branch.

## Tests

Adds inverse-path coverage for every affected handler (verdict flips both directions), including a new `test/assertions/functionToolCall.test.ts` and a grader-error-is-not-inverted case for `closedqa`.

- `npx vitest run test/assertions/` → 60 files, 1315 pass


src/assertions/openai.ts
```diff
@@ -105,3 +105,23 @@ export const handleIsValidOpenAiToolsCall = async ({
     };
   }
 };
+
+// The core validator above has several exit points, so `not-is-valid-openai-tools-call`
+// negates its verdict here rather than threading `inverse` through every branch.
+export const handleIsValidOpenAiToolsCall = async (
+  params: AssertionParams,
+): Promise<GradingResult> => {
+  const result = await handleIsValidOpenAiToolsCallInner(params);
```
jameshiester-oai: **[P2] Return an invalid verdict before applying inverse**

This wrapper assumes the inner validator always returns a `GradingResult`, but malformed nonempty arrays violate that contract. With provider output `[{}]`, the check at line 53 dereferences `toolsOutput[0].function.name` and throws, so `not-is-valid-openai-tools-call` records an eval error instead of passing for an invalid call. I reproduced this through the local CLI at this head. Guard the nested shape (for example, `toolsOutput[0]?.function`) and return the existing invalid-looking result before negation; add an inverse regression for `[{}]`. Avoid broadly catching provider/configuration failures, which should remain hard failures.

jameshiester-oai (commented):
## Review scope

- Reviewed base: `main` at `ae4a73ab57b2475fad3331e8fad4064f40702718`
- Reviewed head: `6cfd7dd06a4f0f33f62a118c03c6b99a6d4e112f`
- Scope: full `base...head` review; no prior authenticated completion marker existed for this head.
- Verdict: the intended inverse behavior works for well-formed valid and schema-invalid calls and for the three trace assertions, but one malformed OpenAI tool-call shape still turns the new negative assertion into an evaluation error. The existing unresolved grader/setup-error findings also remain applicable.
- CI: all reported checks passed; GitHub reports the PR mergeable and behind its base.

## Verified findings

### P2

- `src/assertions/openai.ts:114` — `not-is-valid-openai-tools-call` only negates results that the inner validator returns. For a malformed but nonempty array such as `[{}]`, the inner shape check dereferences `toolsOutput[0].function.name` and throws first. A real no-cache CLI eval therefore produced one error (`TypeError: Cannot read properties of undefined (reading 'name')`) instead of passing the negative assertion. See the inline comment for the smallest correction direction.

### Existing unresolved feedback (not duplicated inline)

- [P1: closed-QA grader transport/no-output failures can pass inverse checks](https://github.com/promptfoo/promptfoo/pull/9838#discussion_r3450459802). `matchesClosedQa` does not tag every no-verdict failure with `metadata.graderError`, so the guard cannot distinguish those failures from a real negative verdict.
- [Function-call provider/setup failures can pass inverse checks](https://github.com/promptfoo/promptfoo/pull/9838#discussion_r3450459291); the later [Codex P2](https://github.com/promptfoo/promptfoo/pull/9838#discussion_r3450459806) reports the same root issue.
- [P2: missing OpenAI tool schemas can pass inverse checks](https://github.com/promptfoo/promptfoo/pull/9838#discussion_r3450459809).

- Make the inner OpenAI tool-call validator total over runtime output shapes and return a typed invalid verdict before the wrapper applies negation. This keeps ordinary malformed model output separate from provider/configuration failures without a broad catch.
- Add inverse regression cases for a nonempty malformed array (`[{}]`) and related nested-shape failures. Existing tests cover only an empty array, which returns cleanly before the unsafe dereference.
- No additional material simplification was identified in the trace handlers; their final-predicate inversion and binary score calculation are direct and consistent.

## Validation performed

- `/pr-review-toolkit:review-pr all parallel` — unavailable in this environment (`command not found`, exit 127); I completed the six requested review lenses manually and reconciled three independent read-only specialist reports.
- `npx vitest test/assertions/functionToolCall.test.ts test/assertions/modelGradedClosedQa.test.ts test/assertions/openai.test.ts test/assertions/traceErrorSpans.test.ts test/assertions/traceSpanCount.test.ts test/assertions/traceSpanDuration.test.ts test/assertions/assertionsResult.test.ts test/assertions/runAssertions.test.ts test/assertions/trace.test.ts test/evaluator/assertions.test.ts --run` — 10 files, 177 tests passed.
- `npm run tsc` — passed.
- `npx biome check <12 changed files>` — passed with no fixes.
- `npm run build` — passed, including TypeScript, library/server bundles, and the web app build.
- `npm run local -- eval -c /private/tmp/pr9838-config.yaml --no-cache -o /private/tmp/pr9838-results.json` — intentionally paired positive and negative assertions; exit 100 with 0 errors. The valid call produced positive pass/negative fail, and schema-invalid arguments produced positive fail/negative pass.
- `npm run local -- eval -c /private/tmp/codex-root-pr9838-malformed.yaml --no-cache -o /private/tmp/codex-root-pr9838-malformed-results.json` — reproduced the P2; exit 100 with 1 error and no assertion verdict.
- `git diff --check ae4a73ab57b2475fad3331e8fad4064f40702718...6cfd7dd06a4f0f33f62a118c03c6b99a6d4e112f` — passed; worktree remained clean at the exact detached head.

## Limitations

- No external model-grader call was made, so the closed-QA transport behavior was validated from the current matcher/handler control flow and existing feedback rather than a live provider outage.
- No persisted-trace CLI workflow was run; trace inverse behavior was covered through focused handler, dispatch, aggregation, and trace integration tests.

src/assertions/openai.ts
```diff
@@ -1,73 +1,612 @@
...
+  } catch {
+    return undefined;
+  }
+}
+
+function parseMetadataMcpToolCalls(
+  metadataCalls: unknown,
+  complete: unknown,
+): StructuredMcpToolCalls | undefined {
+  if (metadataCalls === undefined) {
```
jameshiester-oai, line 140: [P2] Validate completeness even when call entries are absent

`mcpToolCallsComplete` is part of the documented custom-provider contract, but this return happens before its type check. With `metadata: { mcpToolCallsComplete: 'false' }` and no `mcpToolCalls`, a schema-valid traditional call makes `is-valid-openai-tools-call` pass even though the explicit provenance field is malformed. Validate any supplied completeness value before this early return, and cover absent calls with a non-boolean flag so both polarities fail closed.

src/assertions/index.ts
```diff
@@ -177,6 +195,209 @@ export function hasTraceAwareAssertions(assertions?: AssertionOrSet[]): boolean
...
+    calls.push(call.value);
+  }
+  return { ok: true, value: calls };
+}
+
+function normalizeRawMcpOutput(value: unknown): { ok: true; value: unknown } | { ok: false } {
+  if (isProxyValue(value)) {
+    return { ok: false };
+  }
+  if (!Array.isArray(value)) {
```
jameshiester-oai, line 305: [P2] Reject MCP-shaped singleton raw output instead of treating it as absent

An OpenAI Responses-style `raw.output` must be an array, but a custom provider can explicitly return a singleton `{ type: 'mcp_call', ... }` object here. This branch converts that malformed MCP provenance to `undefined`; capture returns `null`, and `not-is-valid-openai-tools-call` passes as though an ordinary invalid output had been proven. Preserve compatibility for unrelated non-array raw payloads, but recognize an MCP-shaped singleton as malformed provenance and test both assertion polarities.

src/assertions/index.ts
```diff
@@ -177,6 +195,209 @@ export function hasTraceAwareAssertions(assertions?: AssertionOrSet[]): boolean
...
+    }
+    const type = readOwnDataProperty(entry, 'type');
+    if (!type.ok) {
+      return { ok: false };
+    }
+    if (!type.exists || typeof type.value !== 'string') {
+      return { ok: false };
+    }
+    if (
+      type.value !== 'mcp_call' &&
```
jameshiester-oai, line 328: [P2] Preserve other tool invocations when deciding mixed-call completeness

Responses output also includes invocation items such as `web_search_call`, `file_search_call`, and `code_interpreter_call`, with failed or incomplete states. This whitelist discards them: a completed `mcp_call` followed by a failed `web_search_call` captures only the MCP success, and `is-valid-openai-tools-call` returns pass/score 1. That conflicts with the documented fail-closed behavior when structured provenance covers only part of a mixed tool-call response. Retain and validate these invocation items, or mark the MCP provenance incomplete, and add a mixed-status regression.

jameshiester-oai (commented):
Reviewed base `ae4a73ab57b2475fad3331e8fad4064f40702718` and exact head `73a6dfb3c64564856512b8f69249d4c42063b49b`.

Scope: incremental review from the prior authenticated completed head `6cfd7dd06a4f0f33f62a118c03c6b99a6d4e112f` (an ancestor), plus a full `base...head` review for interaction and regression context. The PR is mergeable but behind the base branch. All 39 currently reported checks pass.

Verdict: three new P2 correctness gaps remain in the MCP provenance hardening. They can convert explicit malformed or partial provenance into an ordinary valid/invalid verdict, creating false positives in both the positive and negated assertion paths.

### P2 findings

- `mcpToolCallsComplete` is not type-checked when `mcpToolCalls` is absent. With `{ mcpToolCallsComplete: 'false' }` plus a schema-valid traditional call, `is-valid-openai-tools-call` passes instead of failing closed.
- An MCP-shaped singleton `raw.output` object is normalized to absence rather than malformed provenance. The negated assertion then passes because no validation verdict was produced.
- Mixed Responses payloads discard tool invocations other than MCP, approval, and function calls. A completed MCP call plus a failed `web_search_call` is captured as only the MCP success, so the positive assertion passes despite incomplete mixed-call coverage.

### Existing feedback

The existing [P2 malformed-provenance thread](https://github.com/promptfoo/promptfoo/pull/9838#discussion_r3484240358) remains applicable: structured provenance errors still flow through `applyInverse`, allowing malformed or pending-approval input to satisfy `not-is-valid-openai-tools-call`. I did not duplicate that inline comment. The older failure-vs-incomplete thread is superseded by the current documented and tested contract that a reported MCP failure is decisive; the other earlier threads are fixed or outdated at this head.

The raw-item normalization and MCP parsing should share one typed classification helper. The current two-stage whitelist lets explicit malformed shapes and supported Responses invocation variants disappear at different boundaries. Add a table-driven matrix covering field absence/type, singleton vs array raw output, and mixed MCP plus every supported Responses call status.

### Validation

- `npx vitest run test/assertions/ --sequence.shuffle=false` — 61 files, 1,548 tests passed.
- Focused OpenAI/Azure/Bedrock/Google/provider run — 8 files, 666 tests passed.
- Focused evaluator-transform/contracts/MCP-command run — 5 files, 118 tests passed.
- `npm run tsc` and `npm run build` — passed, including the production Web UI build.
- Changed-file Biome, Prettier, and `git diff --check` — passed; Biome reported 15 existing complexity warnings and no errors.
- Exact-head read-only `tsx` probes reproduced all three new findings and the still-open malformed-provenance issue.
- Credential-free localhost CLI E2E:
  `PROMPTFOO_CONFIG_DIR=/private/tmp/pr9838-config PROMPTFOO_CACHE_PATH=/private/tmp/pr9838-cache PROMPTFOO_DB_PATH=/private/tmp/pr9838-config/promptfoo.db PROMPTFOO_DISABLE_TELEMETRY=true npm run local -- eval -c /private/tmp/pr9838-tools.yaml --no-cache -o /private/tmp/pr9838-tools.json`
  exited 0 with 2/2 passing, score 1 each, and zero errors: the valid positive tool-call path and invalid negated path both behaved as expected.

### Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable (exit 127), so its correctness, tests, docs, error-handling, type-design, and simplification lenses were applied manually with three read-only specialists. No paid provider, live MCP server, credential, browser flow, persistence migration, merge, or rebase was used.

PR #9840, opened 2026-06-22
fix(testCaseReader): preserve JSONL row description instead of overwriting

he-yufeng:
`parseJsonlTestCases` in `src/util/testCaseReader.ts` unconditionally overwrites the `description` field with `"Row #N"` even when the row already has one:

```ts
return {
  ...row,
  description: `Row #${idx + 1}`,  // always overwrites
};
```

The three other parsers in the same file (`parseJsonTestCases`, `parseYamlTestCases`, `csvRowsToTestCases`) all use `row.description || \`Row #${idx + 1}\`` as a fallback. JSONL was the odd one out.

So a file like:
```jsonl
{"description":"My custom test","vars":{"prompt":"hello"}}
```
would silently produce `description: "Row #1"` instead of `"My custom test"`.

Fix is one word: change the explicit override to a `||` fallback, consistent with the rest of the file. Added a test to cover the preserved-description case.

### DIFF
--- a/src/util/testCaseReader.ts
+++ b/src/util/testCaseReader.ts
@@ -364,7 +364,7 @@ function parseJsonlTestCases(fileContent: string): TestCase[] {
       const row = JSON.parse(line);
       return {
         ...row,
-        description: `Row #${idx + 1}`,
+        description: row.description || `Row #${idx + 1}`,
       };
     });
 }
--- a/test/util/testCaseReader.test.ts
+++ b/test/util/testCaseReader.test.ts
@@ -310,6 +310,17 @@ describe('readStandaloneTestsFile', () => {
     ]);
   });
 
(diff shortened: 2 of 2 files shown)


mldangelo (approved):
Thanks @he-yufeng — clean, well-targeted fix. JSONL was the lone test-file parser that **overwrote** a row's own `description` with the `Row #N` label, instead of falling back to it; this brings it into line with the JSON/YAML/CSV parsers, which all use `item.description || \`Row #${idx + 1}\``.

Verified end-to-end: a row with `"description":"..."` is now preserved, and a row without one still gets `Row #2`. I added a small follow-up test covering an empty-string description (which correctly falls back to the label, pinning the deliberate `||` semantics).

LGTM ✅

PR #9841, opened 2026-06-22
fix(bleu): return pass:false for empty output instead of throwing

he-yufeng:
An LLM outputting an empty string is a valid scenario — the model may have refused or had its generation truncated. The current guard in `calculateBleuScore`:

```ts
if (!candidate || references.length === 0 || weights.length !== 4) {
  throw new Error('Invalid inputs');
}
```

treats `""` as invalid because empty string is falsy in JavaScript, and throws `Error('Invalid inputs')`. The caller `handleBleuScore` has no try/catch around this call, so the exception propagates and crashes the entire evaluation run.

Every other assertion handler (`handleLevenshtein`, `handleRougeScore`, etc.) returns `{pass:false, score:0}` for any string-typed output — none throw.

**Fix:** change the guard to `candidate == null` (catches only null/undefined, not empty string) and add an explicit early return of `0` for empty-string input. An empty candidate has zero matching n-grams against any reference, so a score of 0 is semantically correct.

**Test:** adds a regression case verifying that `handleBleuScore` with `outputString: ''` does not throw, returns `pass: false`, and returns `score: 0`.

### DIFF
--- a/src/assertions/bleu.ts
+++ b/src/assertions/bleu.ts
@@ -67,8 +67,11 @@ export function calculateBleuScore(
   references: string[],
   weights: number[] = [0.25, 0.25, 0.25, 0.25],
 ): number {
-  if (!candidate || references.length === 0 || weights.length !== 4) {
+  if (candidate == null || references.length === 0 || weights.length !== 4) {
     throw new Error('Invalid inputs');
+  }
+  if (candidate === '') {
+    return 0;
   }
   // BLEU weights are non-negative by definition. Rejecting negatives keeps the
   // score within the documented [0, 1] range (a negative weight on a smoothed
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thanks @he-yufeng — really nice catch. An empty model output (a refusal or a truncated generation) throwing `Invalid inputs` and taking down the **entire** eval run is a genuine footgun, and returning a semantically-correct `0` (→ `pass:false`) is exactly the right call — it matches how `handleLevenshtein`/`handleRougeScore` already behave.

I reviewed and QA'd this end-to-end (empty output now scores 0 with no crash; a matching output still scores 1) and pushed a small follow-up on top:
- Broadened the guard from `candidate === ''` to `candidate.trim() === ''`, so a whitespace-only output (`'\n'`, `'   '`) also scores 0 instead of a misleading `~3.7e-8`.
- Added direct `calculateBleuScore` tests (empty / whitespace / null still throws) and an `inverse` empty-output case, and refreshed the `@throws` JSDoc.

Heads-up for a follow-up: `gleu.ts` and `meteor.ts` carry the identical empty-output `throw` (rouge already returns 0), so the same fix applies there.

LGTM ✅

PR #9845, opened 2026-06-22
fix(assertions): correct the not-meteor failure message and give meteor.ts real test coverage

mldangelo:
## Problem

`handleMeteorAssertion` honored `inverse` for the verdict, but its failure message was hardcoded as “did not meet threshold.” A failing `not-meteor` assertion therefore reported the opposite of what happened: it fails precisely because its score **met** the threshold.

The existing METEOR unit test also mocked the entire module and exercised a hand-written reimplementation, leaving `src/assertions/meteor.ts` without meaningful direct coverage.

## Fix

1. Return an inverse-aware failure reason: `METEOR score X met threshold Y (expected it not to)`.
2. Narrow `handleMeteorAssertion` to the four `AssertionParams` fields it actually consumes, so tests can remain type-safe without casting an incomplete object.
3. Split coverage by purpose:
   - `meteor.test.ts` drives the real handler and scorer while mocking only the external `natural` NLP boundary. This keeps the 12-case behavior matrix deterministic and fast on every unit-test platform, including Windows.
   - `meteor.integration.test.ts` is excluded from the regular unit matrix and exercises real Porter stemming plus WordNet synonym lookup. It also routes a real `not-meteor` assertion through `runAssertion`, covering the production dispatch path end to end.

## Verification

- Focused unit suite: **12/12 passed**
- Real-Natural integration and assertion-pipeline E2E: **2/2 passed**
- Targeted `meteor.ts` unit coverage: **89.65% statements / 90.47% branches / 91.66% functions / 89.90% lines**
- Biome check on all changed TypeScript files: clean
- Two independent native prepublish review passes plus an independent verifier: no findings

The full GitHub matrix remains the authoritative cross-platform and repository-wide check.


test/assertions/meteor.integration.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+
+    // Either stemming or synonym matching alone scores 0.25 for this pair. A score
+    // above 0.9 proves the real dependency supplied both matches.
+    expect(result.score).toBeGreaterThan(0.9);
+    expect(result.pass).toBe(true);
+  });
+
+  it('routes a real not-meteor assertion through the assertion pipeline', async () => {
+    const { runAssertion } = await import('../../src/assertions/index');
```
jameshiester-oai, line 23: **[P1] Exercise the native `natural` module shape**

This does not cover the production dispatch boundary because `vitest.integration.config.ts` enables `deps.interopDefault`. Under that transform the CommonJS `natural@8.1.1` API appears as named exports, but native Node exposes it under `default` (and `module.exports` on Node 24). The loader at `src/assertions/meteor.ts:25-27` consequently assigns `undefined`, and I reproduced `Invariant failed: PorterStemmer should be loaded` for every METEOR case through both `npm run local -- eval` and the freshly built `dist/src/main.js`; the corrected inverse reason is never reached. Normalize the imported namespace with `imported.default ?? imported`, then cover this with a native/built CLI smoke test or run this test without Vitest's interop shim.

jameshiester-oai (commented):
## Review scope

- Reviewed base: `main` at `644e71b06be9efe2532fa43832a52b9ed71dd972`
- Reviewed head: `9d7f4695e83b95b9b4ac15ab8b649877ac3d9413`
- Scope: full `base...head` review; no prior authenticated completion marker existed for this head.
- Verdict: the inverse threshold comparison and revised message are directionally correct, but the new “real Natural” production-dispatch test is a false positive. The exact source and built CLIs cannot run any METEOR assertion because native Node exposes the CommonJS dependency under `default`, while the loader reads named exports. This blocks the changed behavior in the shipped runtime.
- CI: all reported checks passed; GitHub reports the PR mergeable and behind its base. The green integration job uses the same Vitest interop setting that masks the finding.

## Verified findings

### P1

- `test/assertions/meteor.integration.test.ts:23` — the new dispatcher integration test runs with `deps.interopDefault: true`, so Vitest synthesizes named `PorterStemmer` and `WordNet` exports for CommonJS `natural@8.1.1`. Native Node 22/24 exposes those APIs under `default`; `ensureNaturalPackage()` therefore stores `undefined` and every shipped `meteor`/`not-meteor` assertion throws `Invariant failed: PorterStemmer should be loaded` before producing a score, verdict, or corrected reason. The loader defect predates this branch, but the newly added test explicitly claims to close this production-path coverage gap and cannot detect it. See the inline comment for the correction direction.

## Existing feedback

- The sole Copilot thread about `as unknown as AssertionParams` is unresolved but outdated. The current head removes that cast by narrowing the handler parameter to the four consumed fields, so it is fixed and not duplicated.

- Normalize the dependency once at the loading boundary (`imported.default ?? imported`) before selecting `PorterStemmer` and `WordNet`; this is simpler than relying on runner-specific interop.
- Exercise this boundary through a native/built CLI smoke test, or otherwise disable Vitest's default-export interop for this test. A `runAssertion` call inside the current Vitest config is not production module-loading coverage.
- No other material simplification or behavioral defect was found in the changed threshold/message logic. Equality correctly passes `meteor` and fails `not-meteor`, and inverse scoring remains the established complement.

## Validation performed

- `/pr-review-toolkit:review-pr all parallel` — unavailable in this environment (`command not found`, exit 127); I completed the six requested review lenses manually and reconciled three independent read-only specialist reports.
- `npx vitest run test/assertions/meteor.test.ts test/assertions/meteorIntegration.test.ts test/assertions/runAssertions.test.ts test/assertions/assertionsResult.test.ts` — 4 files, 67 tests passed.
- `npx vitest run test/assertions/meteor.integration.test.ts --config vitest.integration.config.ts` — 2/2 passed, reproducing the harness false positive. A specialist rerun with `interopDefault:false` failed both cases with the production invariant.
- `npm run tsc` — passed.
- `npx biome check src/assertions/meteor.ts test/assertions/meteor.test.ts test/assertions/meteor.integration.test.ts` — passed with no fixes.
- `npm run build` — passed, including TypeScript, library/server bundles, and the web app build.
- Native Node import probe — Node 24.17.0 exposed `default`/`module.exports`, with no named `PorterStemmer` or `WordNet`; the APIs exist on `default`. A specialist reproduced the same absence on Node 22.22.2.
- `npm run local -- eval -c /private/tmp/codex-root-pr9845-meteor.yaml --no-cache -o /private/tmp/codex-root-pr9845-results.json` — exit 100; 0 passed, 0 failed, 2 errors, both `PorterStemmer should be loaded`.
- `node dist/src/main.js eval -c /private/tmp/codex-root-pr9845-meteor.yaml --no-cache -o /private/tmp/codex-root-pr9845-built-results.json` after the exact-head build — identical 2-error result.
- `git diff --check 644e71b06be9efe2532fa43832a52b9ed71dd972...9d7f4695e83b95b9b4ac15ab8b649877ac3d9413` — passed; worktree remained clean at the exact detached head.

## Limitations

- No Windows runtime was available. The production failure was reproduced on supported Node 22 and 24 on macOS and in the exact built ESM CLI; CI's Windows unit shards exclude this integration file.
- The full repository test suite was not repeated locally; exact-head CI is green and focused assertion/aggregation tests plus the full build passed.

PR #9849, opened 2026-06-22
fix(assertions): default answer-relevance threshold to 0.7

### LINKED ISSUE #9848: answer-relevance defaults to threshold 0, so it silently always passes when no threshold is set
**Describe the bug**

**To Reproduce**

```yaml
providers: [echo]
prompts: ['{{output}}']
tests:
  - vars:
    options:
      provider:
        text: openai:gpt-4o-mini
    assert:
      - type: answer-relevance
        threshold: 0.5
```

```
```

**Expected behavior**

**Additional context**

i-anubhav-anand:
Fixes #9848

## Summary

`answer-relevance` defaulted to a threshold of **`0`** when none was set (`assertion.threshold ?? 0` in `src/assertions/answerRelevance.ts`). Because the relevance score is always `>= 0`, the pass check `relevance >= 0` is always true — so an `answer-relevance` assertion **with no `threshold` was a no-op that always passed**, even for a completely irrelevant answer.

Every other graded assertion already defaults to a meaningful value (`similar` 0.75, `rouge` 0.75, `bleu`/`gleu`/`meteor` 0.5). This changes `answer-relevance` to default to **`0.7`** — the value shown in the docs' primary example.

## Proof (live, gpt-4o-mini + text-embedding-3-small)

Query *"What is the capital of France?"*, answer *"Bananas are yellow and grow on trees…"*, no threshold:

| | before | after |
|---|---|---|
| score | 0.077 | 0.076 |
| verdict | **PASS** ("greater than threshold 0") | **FAIL** ("less than threshold 0.7") |

The irrelevant answer now correctly fails instead of silently passing.

## Tests

`test/assertions/answerRelevance.test.ts` — the existing "default threshold" test asserted the old `0`; it's updated to assert `0.7` (it fails on `main`, passes here). Explicit thresholds still override. Full handler + matcher suites pass.

## Backward compatibility

This is a **behavior change**: existing `answer-relevance` assertions written without an explicit `threshold` will now actually evaluate (relevance `< 0.7` fails) instead of always passing. That is the point of the fix — those assertions were previously no-ops — but it's called out here for reviewers. Any assertion with an explicit `threshold` is unaffected.

Happy to also add a sentence to the docs stating the default if preferred.

### DIFF
--- a/examples/provider-llama-cpp/promptfooconfig.yaml
+++ b/examples/provider-llama-cpp/promptfooconfig.yaml
@@ -26,3 +26,4 @@ tests:
       - type: icontains
         value: 'San Francisco'
       - type: answer-relevance
+        threshold: 0.7
--- a/site/docs/configuration/expected-outputs/model-graded/answer-relevance.md
+++ b/site/docs/configuration/expected-outputs/model-graded/answer-relevance.md
@@ -14,7 +14,7 @@ To use the `answer-relevance` assertion type, add it to your test configuration 
 ```yaml
 assert:
   - type: answer-relevance
-    threshold: 0.7 # Score between 0 and 1
+    threshold: 0.7 # Optional, defaults to 0.7. Score between 0 and 1
 ```
 
 ### How it works
@@ -25,7 +25,7 @@ The answer relevance checker:
 2. Compares these questions with the original query using embedding similarity
 3. Calculates a relevance score based on the similarity scores
 
-A higher threshold requires the output to be more closely related to the original query.
+The assertion passes if the relevance score meets or exceeds the threshold (default 0.7). A higher threshold requires the output to be more closely related to the original query.
 
 ### Example Configuration
 
--- a/site/docs/configuration/expected-outputs/model-graded/index.md
+++ b/site/docs/configuration/expected-outputs/model-graded/index.md
@@ -679,6 +679,10 @@ Model-graded assertions like `llm-rubric` determine PASS/FAIL using two mechanis
 
 This means a result like `{"pass": true, "score": 0}` will pass without a threshold, but fail with `threshold: 1`.
 
+:::note
+Mechanism 1 does not apply to assertions that carry a built-in default threshold. `answer-relevance` (0.7), `g-eval` (0.7), and `similar` (0.75) apply their default when you omit `threshold`, so a low score fails even with no threshold configured.
+:::
+
 **Common issue**: Tests show PASS even when scores are low
 
 ```yaml
--- a/src/assertions/answerRelevance.ts
+++ b/src/assertions/answerRelevance.ts
@@ -22,7 +22,10 @@ export const handleAnswerRelevance = async ({
     ...(await matchesAnswerRelevance(
       input,
       output,
-      assertion.threshold ?? 0,
+      // Default to 0.7 (the value shown in the docs' primary example) rather than 0.
+      // A 0 default made answer-relevance a no-op that always passed, since the
+      // relevance score is >= 0 for any answer.
+      assertion.threshold ?? 0.7,
       test.options,
       providerCallContext,
     )),
--- a/src/assertions/answerRelevance.ts
+++ b/src/assertions/answerRelevance.ts
@@ -22,9 +22,6 @@ export const handleAnswerRelevance = async ({
     ...(await matchesAnswerRelevance(
       input,
       output,
-      // Default to 0.7 (the value shown in the docs' primary example) rather than 0.
-      // A 0 default made answer-relevance a no-op that always passed, since the
-      // relevance score is >= 0 for any answer.
       assertion.threshold ?? 0.7,
       test.options,
       providerCallContext,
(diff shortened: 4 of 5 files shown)


mldangelo (commented):
Thanks for this — the underlying diagnosis is right, and it's a genuinely valuable catch.

**What works well here:**

- The bug is real and precisely identified. Cosine similarity is always `>= 0`, so `assertion.threshold ?? 0` made `answer-relevance` a no-op that passed unconditionally. Silent false negatives are the worst class of bug in an eval framework.
- **The test is load-bearing, not decorative.** Mutating the production line back to `?? 0` turns `test/assertions/answerRelevance.test.ts` red (1 failed | 4 passed), so it actually pins the contract. Nice.
- Minimal, focused diff with clear rationale.

Skipping the two points already raised inline (inverse `not-` handling, docs update) — both still stand. Four things beyond those:

### 1. (P2) This turns a shipped example red — `examples/provider-llama-cpp/promptfooconfig.yaml:28`

That's the only bare `- type: answer-relevance` in the repo (`examples/eval-rag` sets `threshold: 0.9` at both `:20` and `:35`). It has no `query` var, so `src/assertions/answerRelevance.ts:18` falls back to `input = prompt` and grades generated questions against the raw instruction text.

Measured with real embeddings across all 10 prompt variants in that example:

```
[0] 0.355  "Write a funny tweet about San Francisco"
[2] 0.420  "Q: Create a humorous tweet about San Francisco\nA:"
[8] 0.187  "{'role':'system','content':...}"
   ...all ten score below 0.7
```

Handler-level, using that config's exact test shape:

```json
{ "pass": false, "score": 0.3346, "reason": "Relevance 0.33 is less than threshold 0.7" }
```

Counterfactual at the old default, same input/output: `score=0.428 pass=true`. So the regression is exactly this change.

Worth noting `threshold: 0.7` (or eval-rag's 0.9) won't rescue it — scores top out at 0.43. Since this is a prompt-format comparison example with no query, is `answer-relevance` even the right assertion there? Dropping it may be cleaner than picking a threshold that barely clears 0.19.

### 2. (P2, question) Is 0.7 the value you want?

Two well-formed RAG-shaped pairs — a real question as `input`, exactly what the assertion is designed for:

```
0.815  pass @0.7   "What is the max purchase price without approval?"
0.633  FAIL @0.7   "How many weeks is maternity leave?"
```

A short, correct, fully-relevant answer lands at 0.633, so 0.7 will produce false failures in legitimate suites too. Was 0.7 a deliberate calibration, or inherited from the docs snippet? The docs are internally inconsistent — `answer-relevance.md:17` shows 0.7 but the fuller example at `:43` uses 0.8. `geval.ts:24` does use 0.7, so there's precedent; something like 0.5 would still fix the no-op with less breakage. Since this is a breaking default change, the specific value deserves an explicit call either way.

### 3. (P2) Three siblings have the identical no-op

```
src/assertions/contextFaithfulness.ts:50   assertion.threshold ?? 0
src/assertions/contextRelevance.ts:43      (assertion.threshold as number) ?? 0
src/assertions/contextRecall.ts:38         (assertion.threshold as number) ?? 0
```

Same vacuous-pass bug. Fix together, or add a note on why they're deferred? Leaving them at 0 makes the RAG assertion family inconsistent.

### 4. Mechanical: branch needs a merge with `main`

The head predates the js-yaml v4→v5 migration, so against current `main` every backend test file fails to even collect:

```
TypeError: Cannot read properties of undefined (reading 'dump')
  at readGlobalConfig src/globalConfig/globalConfig.ts:34
```

Not a defect in your change, but `git merge origin/main` is needed before CI results mean anything.

**Nit:** sibling handlers (`geval.ts:24`, `similar.ts:21`, `rouge.ts:82`) state their defaults bare. The 3-line history comment at `answerRelevance.ts:25-27` is duplicated in the test at `:166-169`, and describes code that will no longer exist — the commit message and #9848 already carry that context.

mldangelo:
Thanks for this, and sorry it sat so long.

Your diagnosis was exactly right: `answer-relevance` defaulted `threshold` to `0`, so `score >= threshold` was always true and the assertion was silently always-green. It turned out **six** of us independently found that same class of bug across the RAG assertions (#10202, #10494, #9907, #9911, #10567 and this one), which is a good sign it was worth fixing.

It landed in #10202 (commit `4831b602d0`), which closed #9848 — the issue this PR targets. That PR covered all four RAG assertions at once and shared the default through a named constant (`DEFAULT_RAG_ASSERTION_THRESHOLD`) rather than an inline literal per file, so it was the one we took forward.

On the value: the PRs split between `0.5` and `0.7`, and we went with **0.5**, matching the existing `DEFAULT_THRESHOLD = 0.5` in `src/external/assertions/deepeval.ts`. The `0.7` in `answer-relevance.md` is a doc *example*, and the other model-graded pages' examples disagree with it (`0.3` on context-relevance, `1.0` on context-recall), so we treated the in-repo constant as the stronger precedent. A lower default also breaks fewer existing evals, which matters because this is a user-visible behavior change.

I checked whether anything here could still land on top, and after rebasing the residual was empty — #10202 makes the same change to the same file.

I'm sorry I can't add you as a co-author on `4831b602d0` — it's already merged, and rewriting `main`'s history to add a trailer isn't something I'd do. Consider this the credit instead: you found a real bug and reported it accurately. Please keep sending fixes.

Closing as superseded by #10202.

PR #9851, opened 2026-06-22
docs(site): refresh legacy OpenAI model examples

camgrimsec:
## Summary

- replace targeted legacy OpenAI defaults in seven example configs and three configuration pages with `openai:chat:gpt-5.4-mini`
- use `openai:chat:gpt-5.5` only where the existing example needs a meaningful mini-versus-smart comparison
- preserve endpoint-qualified provider IDs with object `id:` form so exact test-level and CLI provider filters select the intended target
- align provider-filter and tool-calling snippets, and remove legacy `temperature` overrides because Promptfoo omits that field for GPT-5 model IDs

## Why

These files are copied and run directly by users, so stale defaults affect both cost and first-run behavior. GPT-4 is an older, expensive default; current standard pricing is $30/$60 per million input/output tokens versus $0.75/$4.50 for GPT-5.4 mini. GPT-5.5 preserves the higher-capability comparison lane while the combined mini/frontier pair remains cheaper than the previous GPT-4.1-mini/GPT-4 pair at equal token counts.

This is a targeted refresh of unversioned legacy defaults, not a repository-wide replacement of every pre-GPT-5 reference. Capability-specific models, fixtures, stored outputs, and historical examples remain unchanged.

## Compatibility

- documentation and examples only: no production code, public API, schema, dependency, or application-default change
- both replacement models support Chat Completions and function calling; the object provider form preserves the configured ID without changing request bodies
- existing user configurations are unchanged
- GPT-5 requests use reasoning-model semantics: Promptfoo omits `temperature` and does not apply the prior implicit `max_tokens: 1024` cap; set `max_completion_tokens` when an adapted example needs an explicit output bound

## Validation

- `npm run l`
- `npm run f`
- parsed all seven changed YAML configs with `yq`
- six changed configs pass `npm run local -- validate config -c <path>`
- focused Vitest runs: 5 files / 308 tests and 2 files / 459 tests passed
- local OpenAI-compatible mock QA: multiple-config eval 4/4, exact-ID filter 2/2, prompt-per-model routing 48/48 (32 mini, 16 GPT-5.5), tool call 1/1, and explicit error artifacts for mocked provider failures
- `SKIP_OG_GENERATION=true npm run build` in `site/`
- exact-head CI run [28046695032](https://github.com/promptfoo/promptfoo/actions/runs/28046695032): 31 passed, 2 expected fork-gated skips, 0 failures

## Remaining risks

- no billable OpenAI or HuggingFace request was made; external availability, latency, billing, and model quality were verified from current provider contracts and official documentation rather than live credentials
- `examples/redteam-tracing-example/promptfooconfig.advanced.yaml` still fails current validation for unrelated pre-existing prompt/plugin/default-test drift; base and head fail identically, and the provider-only change adds no diagnostic
- GPT-5.5 defaults to billed medium reasoning, so actual latency and cost can exceed equal-token estimates; its extended prompt cache can retain derived tensors for up to 24 hours for adapted prompts of at least 1,024 tokens, including on ZDR projects. The shipped prompts are short and do not trigger that threshold.

### DIFF
--- a/examples/config-csv-metadata/promptfooconfig.yaml
+++ b/examples/config-csv-metadata/promptfooconfig.yaml
@@ -5,8 +5,6 @@ prompts:
   - 'Tell me about {{topic}}'
 
 providers:
-  - id: openai:gpt-4o-mini
-    config:
-      temperature: 0.7
+  - openai:chat:gpt-5.4-mini
 
 tests: file://tests.csv
--- a/examples/config-csv-metadata/promptfooconfig.yaml
+++ b/examples/config-csv-metadata/promptfooconfig.yaml
@@ -5,7 +5,7 @@ prompts:
   - 'Tell me about {{topic}}'
 
 providers:
-  - id: openai:gpt-4
+  - id: openai:gpt-4o-mini
     config:
       temperature: 0.7
 
--- a/examples/config-csv-metadata/promptfooconfig.yaml
+++ b/examples/config-csv-metadata/promptfooconfig.yaml
@@ -5,6 +5,6 @@ prompts:
   - 'Tell me about {{topic}}'
 
 providers:
-  - openai:chat:gpt-5.4-mini
+  - id: openai:chat:gpt-5.4-mini
 
 tests: file://tests.csv
--- a/examples/config-multiple-configs/configs/config1.yaml
+++ b/examples/config-multiple-configs/configs/config1.yaml
@@ -7,8 +7,8 @@ prompts:
   - 'Write a tweet about {{topic}}'
 
 providers:
-  - openai:chat:gpt-5.4-mini
-  - openai:chat:gpt-5.5
+  - id: openai:chat:gpt-5.4-mini
+  - id: openai:chat:gpt-5.5
 
 tests:
   - vars:
--- a/examples/config-multiple-configs/configs/config1.yaml
+++ b/examples/config-multiple-configs/configs/config1.yaml
@@ -8,7 +8,7 @@ prompts:
 
 providers:
   - openai:gpt-4.1-mini
-  - openai:gpt-4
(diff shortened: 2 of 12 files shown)


mldangelo-oai (commented):
Reviewed exact head `e5494a6783f4b7e6a7e380c90fb19654e5758173`.

- Fixed: initial head `30da4ffb40382a3d8b9e2e37eeeecc7bd04ddd8e` introduced scalar `openai:chat:*` entries that instantiated as `openai:*`; an anchored exact provider filter therefore selected zero cases and exited successfully. Commit `e5494a6` uses the existing one-line object `id:` form so configured IDs survive runtime loading. The fix stays local to examples/docs instead of changing provider/filter semantics.
- End-to-end QA with a local OpenAI-compatible mock passed: multiple-config eval `4/4`, exact-ID filter `2/2`, prompt-per-model routing `48/48` (`32` mini / `16` GPT-5.5), and tool calling `1/1`; mocked provider failures were preserved in exported error artifacts.
- Verification passed: six changed configs validate; focused Vitest runs passed (`308` + `459` tests); `npm run l`; `npm run f`; `git diff --check`; docs production build; and exact-head CI (`31` passed, `2` expected fork-gated skips, `0` failures). No unresolved review threads remain.
- Compatibility: docs/examples only (`+26/-30`; production/test LOC `0/0`), with no API, schema, dependency, or application-default change. Remaining risk is bounded: no billable provider call was made; the advanced tracing config retains base-identical validation debt; GPT-5.5 defaults to billed medium reasoning, has no prior implicit 1,024-token cap, and uses extended caching for long prompts. The shipped prompts are short.

mldangelo-oai:
Thanks for the cleanup!

PR #9852, opened 2026-06-22
fix(redteam): use SHA-256 for config fingerprints

camgrimsec:
## Problem

`promptfoo redteam generate` fingerprints its input config so unchanged generated output can be reused. The fingerprint used MD5. In Node/OpenSSL runtimes with a FIPS provider enabled, MD5 is unavailable, so the cache check or metadata write can fail with `ERR_OSSL_EVP_UNSUPPORTED`.

This is compatibility hardening, not a vulnerability fix. `metadata.configHash` is an opaque equality key and is not used for authentication, authorization, signing, or integrity enforcement.

Scope clarification: the failure applies to runtimes actually using a FIPS-enabled OpenSSL provider or policy. A plain RHEL 9 `DEFAULT` policy and the `--openssl-config` flag alone do not necessarily enable FIPS. See the [Node.js FIPS guidance](https://nodejs.org/api/crypto.html#fips-mode) and [OpenSSL FIPS provider algorithms](https://docs.openssl.org/3.5/man7/OSSL_PROVIDER-FIPS/).

## Changes

- Replace MD5 with SHA-256 while preserving the existing fingerprint inputs: Promptfoo `VERSION`, raw config content, and optional provider or target filters.
- Keep the implementation local to the redteam command so the repository architecture boundary remains unchanged.
- Add a regression assertion for the exact SHA-256 digest; reverting the production code to MD5 now fails the test.

## Compatibility

- No CLI, public API, config schema, default, or provider behavior changes.
- The serialized fingerprint grows from 32 to 64 hexadecimal characters. All in-repo consumers treat it as an opaque string; no fixed-width contract exists.
- Existing MD5 fingerprints intentionally mismatch once and regenerate as SHA-256. Normal released-version upgrades already take this path because `VERSION` is part of the fingerprint.
- Same-version source or backport builds also regenerate once. No legacy MD5 comparison is retained because MD5 is unavailable in the target FIPS runtime, and accepting an unverified 32-character value could skip real config changes.
- Existing guidance for custom tests in generated `redteam.yaml` remains applicable: keep a backup because any legitimate fingerprint invalidation can regenerate the file.

## Validation

Exact head: `eca7bd73c7259ba9ffbc4a7df0e8ca2699c05fea`

- `test/redteam/commands/generate.test.ts`: 95/95 passed.
- `npm run architecture:check`: passed.
- Real CLI migration: a base-generated 32-character MD5 artifact regenerated to the independently computed 64-character SHA-256 value, then the next invocation skipped with a byte-stable artifact.
- MD5-denial contract: MD5 raised `ERR_OSSL_EVP_UNSUPPORTED`, SHA-256 succeeded, and the real CLI skip path completed.
- RHEL 9.8 UBI with Node 22.22.2 and the OpenSSL FIPS provider active reported `getFips() === 1`; MD5 failed and SHA-256 succeeded.
- Failure path: an unreachable generation backend exited nonzero and produced no output artifact.
- Exact-head CI passed the Node 20, 22, 24, and 26 test/build lanes, macOS and Windows tests, integration and smoke tests, redteam production checks, style, docs, and code scan.

## Remaining risk

The expected one-time regeneration can consume provider time or quota in same-version source builds. Adding a legacy compatibility layer would either invoke FIPS-incompatible MD5 or risk treating a changed config as unchanged, so the final implementation keeps the migration direct and explicit.


mldangelo-oai:
Reviewed exact head `eca7bd73c7259ba9ffbc4a7df0e8ca2699c05fea`.

- Fixed: added an exact SHA-256 regression assertion, narrowed the source comment and PR rationale to actual FIPS-enabled Node/OpenSSL runtimes, corrected the redteam-scoped title, and merged current `main` additively. The final hunk keeps one concise rationale and the direct `node:crypto` path; reusing the shared helper was rejected by `architecture:check` because it widened a forbidden layer edge.
- Verified: 95/95 focused tests passed; `architecture:check` passed; real CLI QA migrated a 32-character MD5 artifact to the independently computed 64-character SHA-256 value and then skipped byte-stably; an MD5-denying runtime still completed the skip path; an unreachable generation backend exited nonzero and wrote no artifact. Exact-head CI run 28046937979 passed 30 jobs with 2 expected skips, and the corrected-title run 28048151151 passed.
- Compatibility: no CLI, API, schema, default, or fixed-width contract change. Legacy hashes intentionally regenerate once. Normal releases already do this because `VERSION` is hashed; a legacy MD5 fallback would fail in the target FIPS runtime or risk a stale-config false skip.

No correctness, security, regression, accessibility, performance, documentation, type, or simplification finding remains. External closeout is still required for human review and the ruleset-required `CodeQL` context, which GitHub default setup does not emit for fork PRs.

PR #9854, opened 2026-06-23
fix(redteam): centralize transient remote generation retries

mldangelo:
## Summary

Remote-only redteam plugins could silently generate zero tests when the remote task endpoint returned a transient task or gateway failure. The original implementation added a second retry loop inside `fetchRemoteTestCases`; the final implementation instead lets that endpoint opt into the existing shared fetch retry budget.

This keeps retry timing, rate-limit handling, cancellation, and retry accounting in one layer while preserving endpoint-specific policy.

## Implementation

- `fetchRemoteTestCases` explicitly retries `500`, `502`, `503`, `504`, and `524` through `fetchWithRetries`.
- The remote endpoint uses the shared default budget of 4 retries / 5 total attempts. The proposed `PROMPTFOO_REMOTE_GENERATION_RETRIES` variable was removed.
- Shared retries use jittered exponential backoff, honor bounded `Retry-After` values for transient responses, cap transient delays at 60 seconds, and preserve the existing rate-limit path and timing.
- Aborts interrupt requests and backoff immediately, including `Request.signal`, remote health checks, plugin generation, and later synthesis stages.
- Status retries require a replayable request body and remain explicit opt-in; unrelated POST callers do not gain automatic status retries.
- `400`, unselected statuses such as `501`, and malformed successful responses fail immediately.
- Remote responses are cached only after HTTP status, JSON shape, error envelope, test-case schema, injection variable, and materialization-contract validation. Invalid cached responses are conditionally evicted, and older writes cannot replace newer valid writes in the same process while the newer cache entry remains.
- Retry and cache diagnostics sanitize URLs and avoid logging rejected response bodies.
- The max-character fallback no longer multiplies an exhausted remote retry budget.

## Compatibility and risk

Existing callers retain their prior status-retry behavior unless they explicitly provide `retryOnStatusCodes`. Existing network-error and `429` handling remains intact, including provider retry-budget propagation. No new dependency or environment/config contract is introduced.

Remote generation uses POST, so retrying an ambiguous transient response is intentionally at-least-once: the service may perform duplicate generation work or incur duplicate cost when a response is lost after processing. There is no cross-request idempotency key. Cache-write ordering is coordinated within one process, resets when the entry expires or is evicted, and does not coordinate separate processes. Existing JSON-format cache entries receive a one-time refetch after the validated text-response path is adopted.

## Companion contract

[promptfoo-cloud#4393](https://github.com/promptfoo/promptfoo-cloud/pull/4393) remains an open, independent companion. It isolates materialization failures to individual test cases and may fall back to a raw, unmaterialized typed-input case; it does not reduce per-test LLM-call load. That fallback is a separate cloud-owned methodology risk. This client PR remains wire-compatible with current cloud responses and does not depend on the companion landing first.

## Verification

- `npx vitest test/cache.test.ts test/fetch.test.ts test/redteam/index.test.ts test/redteam/plugins/index.test.ts test/util/apiHealth.test.ts test/util/generation.test.ts --run`
  - 6 files passed; 433 tests passed.
- `npm run tsc`
  - Passed.
  - Passed.
  - Passed; static site generated. The stats pre-step reported a non-blocking GitHub API rate-limit warning.
- Local HTTP probes covered JSON/HTML `500` and `503` recovery and exhaustion, `400`/`501`/malformed-`200` fail-fast behavior, network errors, `429`, `Retry-After`, aborts, non-replayable streams, retry-count boundaries, cache recovery and ordering, concurrent callers, logging, and jitter.
  - Persistent retryable failures made exactly 5 task requests.
  - Eight distinct concurrent callers recovered with an 884 ms second-attempt spread.
  - Invalid cached responses recovered on the next request; older deferred cache writes did not replace newer immediate writes.
- Real CLI path against a local mock:
  - `npm run local -- redteam generate ... --no-cache --force --strict --remote`
  - Mock returned `503` then `200`; 2 task requests generated 1/1 SSRF test.
  - `npm run local -- redteam eval ... --no-cache`
  - 1 passed, 0 failed, 0 errors, score 1, and no findings.
- Exact-head GitHub CI: all checks successful at `c7f3650f3f0a1099fe794c0b4e0fbfdbe6cfb677`, including the title-validation reruns caused by the metadata update.

## Diff size

- Production: +629 / -82 lines
- Tests: +1,379 / -21 lines
- Documentation: +71 / -9 lines


mldangelo:
@codex review

mldangelo-oai:
Final ownership receipt for `c7f3650f3f0a1099fe794c0b4e0fbfdbe6cfb677`.

- Fixed: centralized endpoint-specific `500/502/503/504/524` retries in the shared fetch budget; preserved rate-limit precedence, bounded jittered backoff, replayability checks, and abort semantics; validated remote responses before caching; protected concurrent cache ordering; tightened logging, cancellation, and failure amplification.
- Review: the fresh complete static and delegated review found no remaining P0–P3 findings. All 8 addressed review threads are resolved; 0 remain unresolved.
- Verification: 433 focused tests passed across 6 files; TypeScript, architecture checks, docs build, and `git diff --check` passed. Local HTTP probes covered recovery, exhaustion, malformed/fail-fast paths, network/429 behavior, cancellation, cache/concurrency, retry bounds, logging, and jitter. Real CLI generation recovered from `503 → 200`, generated 1/1 test, and the exported eval recorded 1 pass, score 1, 0 errors, and no findings.
- CI: 46/46 current exact-head checks passed, including all metadata-triggered title-validation reruns.
- Diff: production +629/-82; tests +1,379/-21; docs +71/-9.
- Residual risk: remote POST retries are at-least-once and may duplicate generation work/cost after an ambiguous failure; no idempotency key is available. Cache ordering is process-local. The open cloud companion remains independent and its raw typed-input fallback is a separate cloud-owned methodology risk.

src/util/fetch/index.ts
```diff
@@ -20,11 +20,175 @@ import {
...
+  url,
+}: {
+  attempt: number;
+  backoff: number;
+  maxRetries: number;
+  response: Response;
+  signal?: AbortSignal | null;
+  url: RequestInfo;
+}): Promise<void> {
+  discardResponseBody(response);
```
jameshiester-oai, line 165: **[P2] Make retry cleanup release the logged response stream**

`monkeyPatchFetch` starts a fire-and-forget `logRequestResponse({ response })` before this function receives the response, and that logger unconditionally drains `response.clone().text()` even when debug output is disabled. Canceling the original tee branch here therefore leaves the logging clone consuming the entire error body while the next retry starts. I reproduced this on the exact head with a 40×64 KiB streaming 503 plus `Retry-After: 0`: retry 2 began after two chunks, but default logging drained all 2.5 MiB and the first connection never closed early; the `x-promptfoo-silent` control closed it after 128 KiB. With four retries and plugin concurrency, slow or large selected 5xx responses can leave several overlapping, unbounded reads and retain sockets/memory. Gate response-body cloning on an active log level and bound/abort it, or coordinate/suppress logging for discarded retry responses so this cancellation actually releases the resource.

jameshiester-oai (commented):
## Review summary

Reviewed base `4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6` through exact head `8596ec0678502a3b28dae1a29a47aa11469fb4d5`. Scope was a fresh full `base...head` review; there was no authenticated completion marker for an earlier head.

**Verdict:** One P2 operability defect remains in retry-response cleanup. The endpoint-specific retry, validation-before-cache, conditional eviction, abort propagation, cache ordering, and strict failure behavior otherwise worked in focused and built-CLI QA. All 43 current checks are successful.

### Verified findings

- **P2 — response logging defeats retry-body cancellation.** See the inline comment. A selected streaming 5xx body remains fully consumed by the fire-and-forget logging clone while the next retry begins, multiplying open streams and memory use under concurrent retries.
- No P0, P1, or P3 findings.

### Existing feedback

All eight existing threads are resolved. I rechecked the current head and confirmed the earlier JSON-before-retry, jitter/documentation, invalid-cache, POST opt-in, rate-limit precedence, and mixed cache-ordering concerns are fixed; none is duplicated here. No unresolved feedback remains.

No separate material simplification opportunity survived validation. The focused fetch tests mock `logRequestResponse`, so they cannot detect the remaining stream-ownership problem. A loopback regression test with a slow, large retryable response and default logging should assert that the first connection closes before the next attempt drains it.

### Validation performed

- `npx vitest run test/cache.test.ts test/fetch.test.ts test/redteam/index.test.ts test/redteam/plugins/index.test.ts test/util/apiHealth.test.ts test/util/generation.test.ts` — exit 0; 6 files, 433/433 tests passed.
- `npm run tsc` — exit 0. `npm run build` — exit 0, including the web app.
- `npx biome check` on the 14 changed source/test files — exit 0 with three pre-existing complexity warnings in `src/redteam/index.ts`; changed Markdown passed `npx prettier --check`; `git diff --check` passed.
- `PROMPTFOO_REMOTE_GENERATION_URL=http://127.0.0.1:18854/api/v1/task PROMPTFOO_REQUEST_BACKOFF_MS=0 npm run local -- redteam generate -c /private/tmp/pr9854-promptfooconfig.yaml -o /private/tmp/pr9854-redteam.yaml --remote --force --no-cache --strict --no-progress-bar` — exit 0; two 503 responses were retried, the third POST succeeded, and the output contained exactly one valid SSRF test.
- The same built CLI command against an always-503 loopback endpoint — expected exit 1; exactly five POST attempts, zero generated tests, and `PartialGenerationError` under `--strict`.
- `npx tsx /private/tmp/pr9854-stream-retry-probe.ts` — default logging started retry 2 after 128 KiB but drained the full 2.5 MiB first response; the `x-promptfoo-silent` control closed the first response at 128 KiB.

### Limitations

I did not call the live Promptfoo remote service or run the full repository test suite; current CI covers the broader matrix. Browser QA was not relevant to this backend retry change. The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable (exit 127), so I used the equivalent six-lens manual review plus three independent same-head specialist passes.

PR #9860, opened 2026-06-24
fix(redteam): enforce minimum message length

alandelong-oai:
## Summary
- add `minCharsPerMessage` enforcement across redteam generation, eval, config validation, setup UI, docs, and focused tests
- preserve scoped/global length-limit precedence and validate effective plugin/strategy ranges after plugin target expansion
- keep the DOMPurify override on the patched safe resolution `3.4.11`

## Validation
- `npm ci` via configured Socket registry
- `npx vitest run test/evaluator/runEval.test.ts test/redteam/index.test.ts test/redteam/plugins/base.test.ts test/redteam/plugins/index.test.ts test/redteam/providers/goat.test.ts test/redteam/providers/shared.test.ts test/redteam/sharedFrontend.test.ts test/redteam/validators.test.ts --sequence.shuffle=false`
- `npm run test:app -- src/pages/redteam/setup/components/RunOptions.test.tsx --run`
- `npm run tsc -- --noEmit`
- `npm run build`
- intentional-defect proofs for new min/eval/setup/scoped-range/collection/import-sync tests


alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

alandelong-oai:
## UI evidence: input validation

Tested head: `b8458d045848fabc6ec996294b4acaf098bd5ab8`.

Validated locally using synthetic configuration and application data:
- A configured maximum below the minimum displays clear range-validation errors.
- Fractional values for number of tests, concurrency, and delay normalize to their supported integer values on blur.
- Importing an inverted minimum/maximum configuration displays the corresponding validation error.
- Focused validation and required automated checks pass.

No customer, production, or private data is included.

alandelong-oai:
@codex security review

src/redteam/providers/goat.ts
```diff
@@ -576,7 +577,7 @@ export default class GoatProvider implements ApiProvider {
         }
 
         const iterationStart = Date.now();
-        throwIfTargetPromptExceedsMaxChars(targetPrompt, maxCharsPerMessage);
+        throwIfTargetPromptViolatesCharLimits(targetPrompt, charLimits);
```
mldangelo-oai, line 581: [P1] Propagate prompt-limit failures instead of swallowing them as recoverable GOAT turn errors.

I reproduced `minCharsPerMessage: 20` with GOAT generating `"short"`. This check throws, the turn-wide catch at [lines 817-830](https://github.com/promptfoo/promptfoo/blob/bec4224ebfd1b3bb45a23ba7998fe2fbf44079b7/src/redteam/providers/goat.ts#L817-L830) logs and continues, no target call is made, and GOAT returns `success: true`, `score: 1`, with no error because the empty output is graded as a refusal. A prompt-limit violation needs to terminate the provider call or become an explicit failed eval.

src/redteam/index.ts
```diff
@@ -698,11 +728,16 @@ async function applyStrategies(
       }
     }
 
-    resultTestCases = filterOversizedTestCases(
+    const strategyMaxCharsPerMessage =
+      maxCharsPerMessage ?? (strategy.config?.maxCharsPerMessage as number | undefined);
+    const strategyMinCharsPerMessage =
+      minCharsPerMessage ?? (strategy.config?.minCharsPerMessage as number | undefined);
+    resultTestCases = filterGeneratedTestCasesByCharLimits(
```
mldangelo-oai, line 776: [P1] Do not apply the final target minimum to an agentic strategy's seed before the strategy runs.

With only GOAT enabled, a short plugin seed, `minCharsPerMessage: 10`, and `basic` disabled, this filter removes the sole input before GOAT can produce a target prompt. `synthesize()` then reports zero test cases and zero failed plugins. GOAT can legitimately turn a short seed into a compliant target message, so minimum enforcement for generative strategies belongs at the actual target-call boundary.

src/evaluator.ts
```diff
@@ -816,7 +819,13 @@ async function renderRunEvalPrompt({
     skipRenderVars,
   );
   if (isRedteam) {
-    throwIfTargetPromptExceedsMaxChars(renderedPrompt, testSuite?.redteam?.maxCharsPerMessage);
+    const testCharLimits = getTargetPromptCharLimits({ test } as CallApiContextParams);
+    throwIfTargetPromptViolatesCharLimits(renderedPrompt, {
```
mldangelo-oai: [P1] Enforce limits at the final underlying provider-call boundary, not only on this intermediate render.

I reproduced two bypasses: (1) a generated redteam test carrying `metadata.pluginConfig.minCharsPerMessage` but no top-level `redteam` block sets `isRedteam=false` and sends the short prompt; (2) Sequence and MCP wrappers transform a prompt that passes here into a shorter payload sent to `originalProvider`. The guard should inspect the final argument immediately before the target call and honor test metadata even when `testSuite.redteam` is absent.

src/redteam/shared/promptLength.ts
```diff
@@ -83,37 +105,38 @@ function parseChatMessages(prompt: string): ChatMessage[] | undefined {
...
-  if (prompt.length <= maxCharsPerMessage) {
-    return undefined;
-  }
-
-  return {
-    length: prompt.length,
-    limit: maxCharsPerMessage,
-    path: 'prompt',
-  };
+  return getViolation({ content: prompt, path: 'prompt' });
```
mldangelo-oai, line 165: [P1] Count user text in multimodal chat messages instead of counting the serialized envelope.

`parseChatMessages()` rejects OpenAI-style `content` arrays, so this fallback treats base64 media and JSON syntax as message text. A one-character user text part plus an image therefore passes a much larger minimum. Extract and measure the text parts of each user message; media bytes must not satisfy a text-length minimum.

src/validators/redteam.ts
```diff
@@ -197,68 +200,306 @@ export const RedteamStrategySchema = z.union([
...
+function strategyTargetsPlugin(
+  strategyId: string,
+  strategyConfig: Record<string, unknown> | undefined,
+  pluginId: string,
+  pluginConfig: Record<string, unknown> | undefined,
+): boolean {
+  if (!strategyId) {
+    return true;
+  }
+  if (STRATEGY_EXEMPT_PLUGINS.some((exemptPlugin) => exemptPlugin === pluginId)) {
```
mldangelo-oai: [P1] Reuse the runtime's concrete plugin expansion and targeting predicate rather than maintaining a second simulation here.

The two paths already disagree: `foundation` can be rejected even though its expanded `beavertails` case is strategy-exempt; an alias containing `cross-session-leak` exempts every child; collection-step exclusions are checked before runtime expansion; and `numTests: 0` strategies are still range-validated even though runtime skips them. These produce both false config rejections and accepted ranges that runtime applies differently. Expanding once and sharing `pluginMatchesStrategyTargets` would also remove most of this new validator complexity.

src/validators/redteam.ts
```diff
@@ -197,68 +200,306 @@ export const RedteamStrategySchema = z.union([
...
+  });
+}
+
+function validateScopedCharsPerMessageLimits(
+  scopes: ScopedPromptLimit[],
+  pathPrefix: Array<string | number>,
+  ctx: z.RefinementCtx,
+): void {
+  for (const [index, scope] of scopes.entries()) {
+    if (typeof scope === 'string' || !scope.config) {
```
mldangelo-oai: [P1] Keep `safeParse` safe for malformed or cyclic layer steps.

`RedteamConfigSchema.safeParse({ strategies: [{ id: 'layer', config: { steps: [null] } }] })` throws a `TypeError` here instead of returning `success: false`, and a cyclic `steps` object overflows the stack. The TypeScript type is not runtime validation because strategy config accepts unknown values. Guard each scope with a record check and use a visited set or iterative traversal before recursing.

src/redteam/strategies/layer.ts
```diff
@@ -158,13 +171,26 @@ export async function addLayerTestCases(
...
-      ...(stepObj.config || {}),
-      ...(config || {}),
-    });
+    const next = await stepAction(applicable, injectVar, stepConfig);
+    const maxCharsPerMessage = stepConfig.maxCharsPerMessage as number | undefined;
+    const minCharsPerMessage = stepConfig.minCharsPerMessage as number | undefined;
 
     // Feed output to next step. If a step yields nothing, subsequent steps operate on empty set.
-    current = next as TestCaseWithPlugin[];
+    current = (next as TestCaseWithPlugin[]).filter((testCase) => {
```
mldangelo-oai, line 205: [P1] Do not enforce the inherited final range after every intermediate transform.

A `rot13 -> base64` layer with a 9-character input and global minimum 10 drops the 9-character ROT13 intermediate here, even though the final base64 output is 12 characters and valid. Validate the completed pipeline, or enforce a limit at an intermediate step only when that step explicitly declares one.

src/redteam/strategies/layer.ts
```diff
@@ -84,6 +89,11 @@ export async function addLayerTestCases(
       const label = typeof config?.label === 'string' ? config.label : undefined;
       const strategyId = getStrategyId(stepObj.id, perTurnLayers, label);
       const scanId = crypto.randomUUID();
+      const { steps: _steps, ...layerConfig } = config || {};
+      const attackProviderConfig = {
```
mldangelo-oai, line 118: [P1] Include trailing per-turn step limits in the attack provider's effective limit config.

I reproduced GOAT followed by a base64 step configured with `minCharsPerMessage: 100` sending a four-character payload. The trailing limit is retained only inside `_perTurnLayers`; it never reaches `attackProviderConfig`/`strategyConfig`, which are what `getTargetPromptCharLimits()` consults before the target call.

src/redteam/providers/bestOfN.ts
```diff
@@ -166,7 +169,10 @@ export default class BestOfNProvider implements ApiProvider {
           try {
             // TODO(ian): Pass the strategy/plugin metadata maxCharsPerMessage limit here so
             // plugin-scoped caps are enforced even when no top-level redteam cap is configured.
-            throwIfTargetPromptExceedsMaxChars(renderedPrompt);
+            throwIfTargetPromptViolatesCharLimits(
+              renderedPrompt,
+              getTargetPromptCharLimits(context),
```
mldangelo-oai: [P1] Preserve configured provider limits when retry/call context has no strategy metadata.

Best-of-N constructed with a minimum of 100 still sent `"short"` when the candidate retry context lacked test metadata; authoritative-markup injection has the same failure mode. The constructors discard the configured min/max and this resolver only sees transient context. Store the effective limits on the provider and pass them as the resolver's provider fallback.

src/redteam/providers/agentic/memoryPoisoning.ts
```diff
@@ -93,17 +96,17 @@ export class MemoryPoisoningProvider implements ApiProvider {
...
 
       // Send the test case to the provider; the test case should poison the memory created in the previous step.
-      throwIfTargetPromptExceedsMaxChars(prompt);
+      throwIfTargetPromptViolatesCharLimits(prompt, getTargetPromptCharLimits(context));
       const testResponse = await targetProvider.callApi(prompt, context, options);
       accumulateResponseTokenUsage(totalTokenUsage, testResponse);
 
       // Send the follow up question to the provider.
-      throwIfTargetPromptExceedsMaxChars(scenario.followUp);
+      throwIfTargetPromptViolatesCharLimits(scenario.followUp, getTargetPromptCharLimits(context));
```
mldangelo-oai: [P2] Preflight all three known scenario prompts before making the first target call.

With a one-character `followUp`, the memory and poisoning prompts are already sent before this check throws, and the accumulated usage is lost with the exception. Validate `scenario.memory`, `prompt`, and `scenario.followUp` as a batch before any side effects.

src/redteam/plugins/index.ts
```diff
@@ -267,27 +288,45 @@ function dedupeTestCases(testCases: TestCase[]): TestCase[] {
...
 }
 
-function withMaxCharsRetries(pluginFactory: PluginFactory): PluginFactory {
+function withCharLimitRetries(pluginFactory: PluginFactory): PluginFactory {
   return {
     ...pluginFactory,
     action: async (params: PluginActionParams) => {
       const maxCharsPerMessage = getMaxCharsPerMessageFromConfig(params.config);
-      if (!maxCharsPerMessage) {
+      const minCharsPerMessage = getMinCharsPerMessageFromConfig(params.config);
```
mldangelo-oai, line 333: [P2] Centralize character-limit retries instead of nesting this wrapper around a plugin base that already retries.

A registered local plugin that always returns an under-minimum prompt was invoked nine times for one requested test (three inner retries times three outer retries). This can multiply generation cost and latency. Either make the base layer solely responsible or mark factories whose actions already perform constrained retries.

src/redteam/index.ts
```diff
@@ -548,7 +572,11 @@ function addLanguageToPluginMetadata(
       pluginId: plugin.id,
       ...(includePluginConfig && {
         pluginConfig: {
-          ...resolvePluginConfigWithMaxChars(plugin.config, maxCharsPerMessage),
+          ...resolvePluginConfigWithCharLimits(
```
mldangelo-oai: [P1] Preserve effective character limits even when a plugin deliberately returns `metadata.pluginConfig: undefined`.

The Intent plugin does this, causing `includePluginConfig` to be false and its plugin-scoped minimum to disappear before strategy filtering. I reproduced `minCharsPerMessage: 9` plus camelcase retaining a five-character result. If the undefined marker has other semantics, attach the resolved char limits in dedicated metadata rather than dropping them.

src/redteam/commands/run.ts
```diff
@@ -63,6 +111,12 @@ export function redteamRunCommand(program: Command) {
     )
     .option('-t, --target <id>', 'Cloud provider target ID to run the scan on')
     .option('-d, --description <text>', 'Custom description/name for this scan run')
+    .option('--max-chars-per-message <number>', 'Maximum characters per generated message', (val) =>
+      Number.parseInt(val, 10),
+    )
+    .option('--min-chars-per-message <number>', 'Minimum characters per generated message', (val) =>
```
mldangelo-oai: [P2] Reject malformed numeric arguments instead of truncating them with `parseInt`.

The built CLI accepts `1.5`, `1e3`, and `12junk` as 1, 1, and 12. That violates the positive-integer config contract and can silently run with a very different limit. Parse with `Number`, require a positive safe integer, and throw Commander's `InvalidArgumentError` on invalid input.

mldangelo-oai (commented):
Extensive QA review against exact head [`bec4224ebfd1b3bb45a23ba7998fe2fbf44079b7`](https://github.com/promptfoo/promptfoo/commit/bec4224ebfd1b3bb45a23ba7998fe2fbf44079b7). I would not merge this head yet: the inline findings are independently reproduced and deduplicated, with the highest-risk failures causing zero target coverage, false-green results, or bypassing the new minimum at the actual provider boundary.

I also verified the happy paths and the current CI state (all checks green); these edge cases are not covered by the existing suite. Two non-inline contract gaps remain: the companion cloud change is pinned behind this head and duplicates validator logic that already differs from runtime, and the CLI command reference does not document the two new flags. The strongest simplification is to centralize effective-limit resolution and final-call enforcement, then reuse runtime expansion/targeting for config validation instead of reproducing those semantics in several providers and validators.

src/app/src/pages/redteam/setup/page.tsx
```diff
@@ -367,6 +367,8 @@ export default function RedTeamSetupPage() {
         provider: yamlConfig.redteam?.provider,
         entities: yamlConfig.redteam?.entities || [],
         numTests: yamlConfig.redteam?.numTests || REDTEAM_DEFAULTS.NUM_TESTS,
+        maxCharsPerMessage: yamlConfig.redteam?.maxCharsPerMessage,
+        minCharsPerMessage: yamlConfig.redteam?.minCharsPerMessage,
```
mldangelo-oai: [P2] Validate/coerce imported YAML before assigning these values to numeric UI state.

Quoted YAML such as `minCharsPerMessage: "50"` is displayed as valid and persisted back as a string, while the runtime schema rejects it. Run the imported redteam block through the shared schema (and surface its error) or explicitly normalize only positive integers here.

src/app/src/pages/redteam/setup/components/RunOptions.test.tsx
```diff
@@ -186,6 +204,73 @@ describe('RunOptionsContent', () => {
...
+      expect(setValue).toHaveBeenCalledWith(String(REDTEAM_DEFAULTS.NUM_TESTS));
+    });
+
+    it('should not persist fractional minCharsPerMessage', async () => {
+      const user = userEvent.setup();
+      renderWithTooltipProvider(<RunOptionsContent {...defaultProps} />);
+      const minCharsPerMessageInput = screen.getByLabelText('Min chars per message');
+
+      await user.clear(minCharsPerMessageInput);
+      await user.type(minCharsPerMessageInput, '1.5');
```
mldangelo-oai: [P2] This test does not exercise fractional-value rejection.

The number input blocks the `.` keystroke, so `user.type(..., '1.5')` produces `15`; the assertion only proves that exactly `1.5` was not persisted and passes even if `15` was. Set the raw value with a change event (or test the component's actual paste path) and assert the expected reset/value and update calls.

src/redteam/commands/generate.ts
```diff
@@ -824,6 +839,12 @@ async function doGenerateRedteamInternal(
       strategies: strategyObjs || [],
       plugins: plugins || [],
       sharing: config.sharing,
+      ...(config.maxCharsPerMessage === undefined
```
mldangelo-oai, line 842: [P1] Carry these limits into the default generated `redteam.yaml` too.

They are added to `updatedRedteamConfig`, but the no-existing-config fallback at [lines 995-1002](https://github.com/promptfoo/promptfoo/blob/bec4224ebfd1b3bb45a23ba7998fe2fbf44079b7/src/redteam/commands/generate.ts#L995-L1002) writes only `description` and `tests`. A programmatic generation with `minCharsPerMessage: 100` therefore emits tests whose later eval has no top-level limit and can send short prompts. Include the redteam block in every config-output path.

site/docs/red-team/configuration.md
```diff
@@ -319,19 +321,21 @@ testGenerationInstructions: |
...
 
 ```yaml
 redteam:
+  minCharsPerMessage: 120
   maxCharsPerMessage: 280
   plugins:
     - harmful:hate
     - id: 'contracts'
       config:
+        minCharsPerMessage: 80
```
mldangelo-oai, line 332: [P2] This example cannot demonstrate the plugin-scoped minimum shown here.

The top-level minimum is 120, and the documented precedence says it overrides this plugin value of 80, so the 80-character setting is dead. Split global and plugin-scoped examples or omit the global minimum in this snippet.

tsconfig.cloud.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.json",
```
mldangelo-oai: [P2] Remove this config or wire it to a bounded cloud build.

No package script or workflow references `tsconfig.cloud.json`; invoking it directly emits roughly 6,995 files / 64 MB, including tests, because it inherits the root project without an `include`/`outDir` boundary. As committed it is an unused and hazardous artifact. If cloud needs a separate compile, give it an explicit source surface and CI consumer.

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

package.json
```diff
@@ -209,10 +209,10 @@
       "chevrotain": "11.2.0"
     },
     "webpack": "5.107.2",
-    "dompurify": "3.4.11",
+    "dompurify": "3.4.10",
     "monaco-editor": {
       ".": "0.55.1",
-      "dompurify": "3.4.11"
+      "dompurify": "3.4.10"
```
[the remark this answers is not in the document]
alandelong-oai: Addressed on current head: `package.json` pins and overrides DOMPurify at 3.4.11, and both root and Monaco lockfile entries resolve 3.4.11. The vulnerable 3.4.10 override is gone.

src/app/src/pages/redteam/setup/components/RunOptions.tsx
```diff
@@ -20,6 +20,11 @@ export const RUNOPTIONS_TEXT = {
       'Optional character cap for each generated user message and final prompt sent to the target. Leave blank for no limit.',
     error: 'Max chars per message must be greater than 0',
   },
```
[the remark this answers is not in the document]
alandelong-oai: Addressed on current head: `RunOptions.tsx` selects the configured-minimum error when max is below min, and `RunOptions.test.tsx` asserts the exact `minError(50)` plus invalid state.

src/app/src/pages/redteam/setup/components/RunOptions.tsx
```diff
@@ -63,7 +69,7 @@ const isBelowMin = (value: string, min: number) => {
     return false;
   }
   const n = Number(value);
-  return Number.isNaN(n) || n < min;
+  return Number.isNaN(n) || !Number.isInteger(n) || n < min;
 };
```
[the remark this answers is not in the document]
alandelong-oai: Addressed on current head: number-of-tests, delay, and max-concurrency blur handlers normalize fractional values to their integer defaults; focused `RunOptions.test.tsx` cases cover each input.

src/redteam/providers/goat.ts
alandelong-oai: Addressed on current head: GOAT checks the final target prompt at `goat.ts:581`, and the turn catch rethrows target-char-limit errors at `:824-825` rather than continuing. `goat.test.ts:179-192` asserts rejection and zero target calls.

src/redteam/index.ts
alandelong-oai: Addressed: `redteam/index.ts:762-774` defers only the generated minimum for agentic strategies, preserving the seed until runtime. `index.test.ts:332-375` proves a short GOAT seed reaches the strategy while only compliant generated output remains.

src/evaluator.ts
alandelong-oai: Addressed at the final boundary: evaluator wraps original/active providers (`evaluator.ts:946-949`), and `mcpTargetProvider.ts:75-81` checks immediately before underlying `callApi`. The Sequence metadata-only regression at `mcpTargetProvider.test.ts:257-278` proves zero target calls.

src/redteam/shared/promptLength.ts
alandelong-oai: Addressed: `promptLength.ts:68-92` parses multimodal arrays and measures only text parts. `promptLength.test.ts:13-27` proves a large base64 image does not satisfy the minimum when user text is only `hi`.

src/validators/redteam.ts
alandelong-oai: Addressed: validator targeting delegates to runtime expansion/targeting (`validators/redteam.ts:227-243`) and skips runtime-disabled `numTests: 0` strategies (`:384-390`). `validators.test.ts:341-450` covers exclusions, collection expansion, direct IDs, and layers.

src/redteam/index.ts
alandelong-oai: Addressed on current head: `redteam/index.ts:542-575` preserves intentional `pluginConfig: undefined` while carrying resolved limits in `metadata.strategyConfig`; `index.test.ts:451-474` covers this exact case.

src/validators/redteam.ts
alandelong-oai: Fixed in `b8458d0`: validator traversal now reports malformed/cyclic layer steps while recursing only actual layer strategies, so arbitrary custom-plugin `config.steps` remains valid. `validators.test.ts` covers null, cycles, and custom plugin steps.

src/redteam/strategies/layer.ts
alandelong-oai: Addressed on current head: `layer.ts` validates explicit step limits after intermediate transforms and applies inherited layer limits only after the completed pipeline. `layer.test.ts` proves a short intermediate can still produce a valid final output.

src/redteam/strategies/layer.ts
alandelong-oai: Fixed in `e271163`: `layer.ts` propagates trailing per-turn limits into attack-provider config as fallbacks, preserving outer/attack-step precedence. The focused layer regression covers final target enforcement.

src/redteam/providers/bestOfN.ts
alandelong-oai: Fixed in `e271163`: best-of-N and authoritative-markup providers retain configured limits and pass provider fallback into target-limit resolution. Focused tests prove a short generated prompt is rejected with zero target calls even without strategy metadata.

src/redteam/providers/agentic/memoryPoisoning.ts
alandelong-oai: Fixed in `e271163`: memory-poisoning now validates memory, prompt, and follow-up before its first target call. The regression supplies a short follow-up and asserts zero target calls.

src/redteam/plugins/index.ts
alandelong-oai: Fixed in `e271163`: local class-based plugins declare ownership of char-limit retries so the global wrapper does not multiply BasePlugin retries; remote generation remains wrapped. `plugins/index.test.ts` asserts exactly three inner attempts.

src/redteam/commands/run.ts
alandelong-oai: Fixed in `e271163`: CLI min/max flags now use strict digits-only positive-safe-integer parsing; `run.test.ts` rejects `1.5`, `1e3`, `12junk`, `0`, and `-1`.

src/app/src/pages/redteam/setup/page.tsx
alandelong-oai: Fixed in `b8458d0`: YAML import normalizes quoted positive integers and rejects invalid or inverted min/max ranges before `setFullConfig`. `page.test.tsx` covers quoted values and min>max.

src/app/src/pages/redteam/setup/components/RunOptions.test.tsx
alandelong-oai: Fixed in `e271163`: the UI regression now pastes `1.5` and asserts the value clears plus no accidental `15` persistence.

src/redteam/commands/generate.ts
alandelong-oai: Fixed in `e271163`: fallback generated `redteam.yaml` now includes `redteam: updatedRedteamConfig`; `generate.test.ts` asserts min/max are written.

site/docs/red-team/configuration.md
alandelong-oai: Fixed in `e271163`: the docs example no longer combines contradictory global and plugin-scoped ranges.

tsconfig.cloud.json
alandelong-oai: Fixed in `e271163`: removed the unused unbounded `tsconfig.cloud.json` rather than leaving a dead config surface.

alandelong-oai:
@codex security review

alandelong-oai:
@codex security review

PR #9858, opened 2026-06-23
fix(assertions): validate BLEU weights before empty score

wholley-oai:
## Summary

- Validate all four custom BLEU weights before returning an empty-candidate score.
- Reject non-finite, non-number, and sparse runtime weight entries while preserving the existing length, non-negative, and sum-to-one requirements.
- Preserve the established score of `0` for valid empty or whitespace-only candidates, and align the inline documentation with that behavior.
- Add focused regressions for malformed weights on both empty and non-empty candidates.

## Compatibility

Valid BLEU configurations and default-weight scores are unchanged. This only tightens invalid custom runtime inputs to the internal BLEU helper; it does not add or change a YAML assertion option or package-root API.

## Testing

- Focused BLEU/assertion/evaluator suite: 5 files, 105 tests passed.
- Adversarial weight matrix covering `NaN`, infinities, non-numbers, sparse entries, negative values, invalid sums, and invalid lengths.
- Deterministic local `--no-cache --no-write` eval matrices using the echo provider: 5/5 expected passes, plus 3 expected assertion failures and 1 expected malformed-value error; exported scores and errors inspected.
- TypeScript, lint/format, architecture check, and full build passed locally.

PR #9865, opened 2026-06-24
test: pin BLEU and ellipsize boundary behavior

mldangelo-oai:
## Summary

- require an identical BLEU candidate/reference pair to score exactly `1`, replacing the stale claim that smoothing lowers a perfect match
- name the two invalid-weight fixtures by the validation failure each one exercises
- pin the `ellipsize(..., 3)` boundary to `...` and remove weaker duplicate length checks already implied by exact assertions for lengths 0–2
- keep production behavior unchanged

## Regression value

- The `maxLen=3` assertion uniquely catches an off-by-one change from `maxLen < 3` to `maxLen <= 3`; the previous length-only check accepted the incorrect `hel` result.
- The exact BLEU check is mathematically correct and fixes a misleading legacy assertion, though the suite also has another exact perfect-match guard.
- The weight-fixture renames improve test intent without changing coverage.

## Relationship to #9858

PR #9858 is complementary rather than superseding: it fixes BLEU weight validation for empty candidates, while this PR tightens separate existing assertions. A three-way merge simulation at the current heads completed without conflicts, preserved both sets of tests, and passed the combined focused suite and TypeScript check.

## Validation

- `npx vitest run test/assertions/bleu.test.ts test/util/text.test.ts --sequence.shuffle=false`
- the same focused suite with shuffle seeds `9865`, `19865`, `29865`, `39865`, and `49865`
- `npm run tsc`
- `npm run l`
- `npx @biomejs/biome check test/assertions/bleu.test.ts test/util/text.test.ts`
- independent programmatic BLEU and ellipsize boundary probe
- controlled mutation checks for the BLEU exact score and `ellipsize` branch boundary
- two-pass native pre-publish review gate with zero findings

### DIFF
--- a/test/assertions/bleu.test.ts
+++ b/test/assertions/bleu.test.ts
@@ -4,12 +4,12 @@ import { calculateBleuScore, handleBleuScore } from '../../src/assertions/bleu';
 import type { AssertionParams } from '../../src/types/index';
 
 describe('BLEU score calculation', () => {
-  it('identical sentences should have BLEU score close to, but not equal to one due to smoothing', () => {
+  it('identical sentences should have BLEU score equal to one', () => {
     const references = ['The cat sat on the mat.'];
     const candidate = 'The cat sat on the mat.';
 
     const score = calculateBleuScore(candidate, references);
-    expect(score).toBeGreaterThan(0.999);
+    expect(score).toBe(1);
   });
 
   it('completely different sentences should have very low but non-zero BLEU score due to smoothing', () => {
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
## Final review receipt

**Disposition: FIXED_AND_READY**

Reviewed exact head `6eca042b0732b0049630a95fabcac86ea00f9750` against base `defd4bbc8a4eb595eebc733f3981f2c8c3e6976d`.

- Finding fixed: removed the redundant length-only `maxLen` 0/1/2 loop after exact-output assertions had already made it weaker duplicate coverage.
- Regression value: `ellipsize(..., 3)` uniquely detects the `maxLen < 3` versus `maxLen <= 3` off-by-one; the exact BLEU expectation is mathematically correct but overlaps an existing perfect-match guard; the invalid-weight fixture renames clarify intent without changing sensitivity.
- Scope: test-only, 2 files, +7/-9; production LOC remains zero.
- QA: focused BLEU/text suite passed 43/43 deterministically and under five shuffle seeds; TypeScript, changed-file lint/Biome, a programmatic behavior probe, and controlled mutation checks passed.
- Fresh review: a new post-fix review found no issues; two independent native pre-publish review passes also returned zero findings.
- PR #9858: complementary, with a clean three-way merge; the combined tree passed 44 focused tests, TypeScript, and behavior probes.
- Exact-head CI: all 39 reported checks pass, including CodeQL, style, smoke/integration, builds, and the complete Node test matrix.
- Threads: 0 unresolved.
- Compatibility: Node 20/22/24/26 on Ubuntu, Node 22 on Windows and macOS, and Node 20/24/26 package builds are green.
- Residual risk: low and test-only; the new BLEU exact check is intentionally overlapping coverage rather than a unique regression detector.

PR #9866, opened 2026-06-24
fix(redteam): bound output tokens on the attack/grading providers

mldangelo:
## Summary

The red-team **attack/grading providers** — used by the iterative strategies (`jailbreak:meta`/meta-agent, `crescendo`, `hydra`, `goat`) via `redteamProviderManager` — were constructed in `loadRedteamProvider` with only `temperature` + `response_format` and **no `max_tokens`**. They therefore fall through to `OPENAI_MAX_TOKENS` (or the model's output ceiling).

When `OPENAI_MAX_TOKENS` is raised for large *test-case generation* outputs, these **short-decision/attack** calls inherit that budget. An occasional degenerate / non-terminating generation (repetition or runaway-JSON, more likely under JSON mode + non-zero temperature + complex agentic prompts) then runs all the way to the model's **32,768-token ceiling** — minutes of latency followed by an unparseable response, which surfaces as generation timeouts, parse failures (`"Failed to parse agent output"`), and instance saturation.

This was observed in production: agentic-decision calls (`meta-agent-decision`, `hydra-decision`, `agentic:memory-poisoning`, `goat`) returning `completionTokens = 32768` with 2–10 min latency, e.g. a single `meta-agent` call with a 1,901-token prompt producing 32,768 completion tokens in 235s → 500.

## What changed

- Add `REDTEAM_PROVIDER_MAX_TOKENS` (default **4096**, overridable via `PROMPTFOO_REDTEAM_PROVIDER_MAX_TOKENS`) in `src/redteam/providers/constants.ts`.
- Apply it to the default redteam provider in `loadRedteamProvider` (`shared.ts`), setting **both** `max_tokens` and `max_completion_tokens` so the cap applies whether the model takes `max_tokens` (e.g. gpt-4.1) or `max_completion_tokens` (reasoning / gpt-5 models).
- This is a **dedicated** cap, independent of the generic `OPENAI_MAX_TOKENS` knob, so tuning generation can never again silently uncap the attack/grading providers.

A short structured decision/attack needs a few hundred tokens; 4096 leaves generous headroom (≈16k chars — well above e.g. a 3,500-char Crescendo turn) while bounding a runaway to ~seconds instead of minutes.

## Testing

- New regression test in `test/redteam/providers/shared.test.ts` asserting the default provider sets a bounded `max_tokens`/`max_completion_tokens` (and that the cap is modest and > 0).
- Updated the existing default-provider config assertions to include the new fields.
- Full file green (62/62); `tsc`, Biome lint/format, and `architecture:check` clean.

### Deterministic repro (request body inspection, gpt-4.1)

| Scenario | request body `max_tokens` |
|---|---|
| `OPENAI_MAX_TOKENS=32768`, redteam default config (pre-fix) | **32768** |
| explicit `max_tokens=4096` (this fix) | 4096 |
| `OPENAI_MAX_TOKENS` unset (chat.ts default) | 1024 |

## Notes / scope

This is the **library-level** hardening so redteam providers are never uncapped by a high global `OPENAI_MAX_TOKENS`. The production incident also has two companion fixes outside this PR:
- **Immediate (ops):** lower `OPENAI_MAX_TOKENS` on the cloud server (it was set to `32768`, which is what made the cap the model ceiling).
- **promptfoo-cloud:** give the server-side `getModels()` attack/decision providers a dedicated cap (they currently read `OPENAI_MAX_TOKENS` directly).

### DIFF
--- a/src/redteam/providers/shared.ts
+++ b/src/redteam/providers/shared.ts
@@ -30,7 +30,12 @@ import { TokenUsageTracker } from '../../util/tokenUsage';
 import { TransformInputType, transform } from '../../util/transform';
 import { remoteGenerationContextPayload } from '../remoteGenerationContext';
 import { throwIfTargetPromptExceedsMaxChars } from '../shared/promptLength';
-import { ATTACKER_MODEL, ATTACKER_MODEL_SMALL, TEMPERATURE } from './constants';
+import {
+  ATTACKER_MODEL,
+  ATTACKER_MODEL_SMALL,
+  REDTEAM_PROVIDER_MAX_TOKENS,
+  TEMPERATURE,
+} from './constants';
 
 import type { TraceContextData } from '../../tracing/traceContext';
 import type { ProviderOptions } from '../../types/providers';
@@ -108,6 +113,12 @@ async function loadRedteamProvider({
       config: {
         temperature: TEMPERATURE,
         response_format: jsonOnly ? { type: 'json_object' } : undefined,
+        // Bound the output so a degenerate/runaway generation can't run to the model's
+        // token ceiling (minutes of latency + an unparseable response). Set both forms
+        // so the cap applies whether the model takes `max_tokens` (e.g. gpt-4.1) or
+        // `max_completion_tokens` (reasoning / gpt-5 models). See REDTEAM_PROVIDER_MAX_TOKENS.
+        max_tokens: REDTEAM_PROVIDER_MAX_TOKENS,
+        max_completion_tokens: REDTEAM_PROVIDER_MAX_TOKENS,
       },
     });
   }
--- a/src/redteam/providers/shared.ts
+++ b/src/redteam/providers/shared.ts
@@ -234,7 +234,7 @@ async function loadRedteamProvider({
 }
 
 function getDefaultTestProvider(): RedteamFileConfig['provider'] | undefined {
-  if (typeof cliState.config?.defaultTest !== 'object') {
+  if (!cliState.config?.defaultTest || typeof cliState.config.defaultTest !== 'object') {
     return undefined;
   }
 
--- a/src/redteam/providers/shared.ts
+++ b/src/redteam/providers/shared.ts
@@ -129,10 +129,19 @@ function getOpenAiOutputCapKind(provider: ApiProvider): OpenAiOutputCapKind | un
   return undefined;
 }
 
-function cloneProviderConfig(provider: ApiProvider): ApiProvider {
+async function cloneProviderConfig(provider: ApiProvider): Promise<ApiProvider> {
   const config = (provider as { config?: Record<string, unknown> }).config;
   if (!config || typeof config !== 'object') {
     return provider;
+  }
+
+  // Some providers (notably Azure) populate authentication state asynchronously on the
+  // original instance. Wait before cloning so the capped copy receives the initialized state
+  // rather than a promise whose continuation will mutate a different object.
+  const initializationPromise = (provider as { initializationPromise?: Promise<unknown> | null })
+    .initializationPromise;
+  if (initializationPromise) {
+    await initializationPromise;
   }
 
   const descriptors = Object.getOwnPropertyDescriptors(provider);
@@ -146,7 +155,7 @@ function cloneProviderConfig(provider: ApiProvider): ApiProvider {
   return Object.create(Object.getPrototypeOf(provider), descriptors) as ApiProvider;
 }
 
-function withDefaultOutputCap(provider: ApiProvider): ApiProvider {
+async function withDefaultOutputCap(provider: ApiProvider): Promise<ApiProvider> {
   const providerKind = getOpenAiOutputCapKind(provider);
   if (!providerKind) {
     return provider;
@@ -155,7 +164,7 @@ function withDefaultOutputCap(provider: ApiProvider): ApiProvider {
   // A caller may reuse an instantiated provider for normal batch generation. Clone the
   // provider's own state before changing its config so attack/grading limits do not leak into
   // those uncapped generation calls.
-  const cappedProvider = cloneProviderConfig(provider);
+  const cappedProvider = await cloneProviderConfig(provider);
   const config = (cappedProvider as { config?: Record<string, unknown> }).config;
   if (!config || typeof config !== 'object') {
     return cappedProvider;
@@ -219,7 +228,7 @@ async function loadRedteamProvider({
     });
   }
   if (shouldCapOutput) {
-    ret = withDefaultOutputCap(ret);
+    ret = await withDefaultOutputCap(ret);
   }
   return ret;
 }
--- a/src/redteam/providers/shared.ts
+++ b/src/redteam/providers/shared.ts
@@ -92,31 +92,69 @@ export function resetRedteamProviderLoader(): void {
  * provider-specific: native providers use different request schemas, while the Responses API
  * uses `max_output_tokens` and chat/completions use different aliases.
  */
-function applyDefaultOutputCap(provider: ApiProvider): void {
+function isOpenAiOutputCapProvider(provider: ApiProvider): boolean {
+  const providerId = provider.id();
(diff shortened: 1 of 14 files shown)


alandelong-oai (commented):
Re-review at `9488009c`: the string / `ProviderOptions` branch is now covered syntactically, but the change is still not safe across the provider manager's actual uses.

1. **The 4096 cap is applied to test-case synthesis as well as attack/grading calls.** `redteamProviderManager.getProvider()` is also the provider used by `src/redteam/index.ts` for local plugin generation, including batches of up to 20 prompts. Users who intentionally raised `OPENAI_MAX_TOKENS` for large generation outputs can now get truncated or partial test sets. Attack/grading and generation need distinct use-case paths.

2. **`applyDefaultOutputCap()` is not provider-aware.**
   - OpenAI/Azure Responses providers read `max_output_tokens`, so the injected `max_tokens` / `max_completion_tokens` fields do not cap them. I verified a Responses request still used the global 32768 value.
   - Native providers can receive unsupported OpenAI-only fields such as `max_completion_tokens`.
   - Provider-native explicit limits can be overridden by the injected `max_tokens`.
   - Treating either OpenAI alias as sufficient is incorrect: GPT-4.1 ignores `max_completion_tokens`, while GPT-5 ignores `max_tokens`.

   Please apply capability-specific fields only to the providers that support them: chat aliases for OpenAI-compatible chat providers, `max_output_tokens` for Responses providers, and no OpenAI-only fields on native providers.

3. **The advertised env override is read too early for the normal CLI `--env-file` flow.** `REDTEAM_PROVIDER_MAX_TOKENS` is initialized at module import time, before `main()` loads env files. Read it when constructing the relevant provider instead.

The existing inline comments about generation scope, Responses providers, native-provider fields, and env-file timing remain applicable. CI is green, but the tests use fake mutable config objects and do not cover effective request bodies or local synthesis behavior.

src/redteam/providers/shared.ts
```diff
@@ -82,18 +87,121 @@ export function resetRedteamProviderLoader(): void {
...
+  ) {
+    return 'chat';
+  }
+  if (
+    hasProviderPrototype(provider, 'OpenAiCompletionProvider') ||
+    constructorName === 'AzureCompletionProvider'
+  ) {
+    return 'completion';
+  }
+  return undefined;
```
alandelong-oai, line 129: [P2] Cap LiteLLM's wrapped OpenAI provider too

`litellm:*` providers are `LiteLLMProviderWrapper` instances, so none of these checks match even though calls delegate to an OpenAI chat/completion provider. They therefore return unchanged and still inherit the global output limit. On this head, with a configured `litellm:gpt-4.1` attack provider and `OPENAI_MAX_TOKENS=32768`, the generated attack request body contained `max_tokens: 32768` rather than the new 4096 cap. Please apply the cap to the wrapped provider without mutating the generation instance, and add a LiteLLM request-body regression test.

alandelong-oai (commented):
One blocking provider-wrapper gap remains before I can approve.

src/redteam/providers/shared.ts
```diff
@@ -82,18 +87,121 @@ export function resetRedteamProviderLoader(): void {
...
+
+  const descriptors = Object.getOwnPropertyDescriptors(provider);
+  const configDescriptor = Object.getOwnPropertyDescriptor(provider, 'config');
+  descriptors.config = {
+    configurable: configDescriptor?.configurable ?? true,
+    enumerable: configDescriptor?.enumerable ?? true,
+    value: { ...config },
+    writable: true,
+  };
+  return Object.create(Object.getPrototypeOf(provider), descriptors) as ApiProvider;
```
jameshiester-oai, line 155: [P2] Avoid fabricating subclass instances without their private-field brand

`Object.create(...)` copies own descriptors, but it cannot give the new object the native private-field brand of an arbitrary subclass. `RedteamFileConfig.provider` accepts `ApiProvider` instances, so a custom class extending `OpenAiChatCompletionProvider` with `#secret` and a `callApi()` that reads it is classified and cloned here; invoking the returned attack provider then throws `TypeError: Cannot read private member #secret from an object whose class did not declare it`. I reproduced that on this head. Please avoid generic instance fabrication: apply the cap while loading/reconstructing dedicated known providers, and define a safe non-mutating path for caller-supplied instances.

jameshiester-oai (commented):
## Scope and verdict

Full, fresh review of base `4f8103c0ff8c53ae19fb8d64fd40ffb2e37f3ab6` through head `e63a41cf0d1db4cc19136dd058eb3764203620f8`; there was no prior authenticated completion marker for this head.

The standard OpenAI/Azure/XAI paths now separate attack/grading limits from batch-generation limits correctly, and every exact-head CI check is green. One new P2 runtime defect remains, plus two previously raised P2 concerns that still reproduce at this head.

## Verified findings

### P2

- **New:** caller-supplied OpenAI-family subclasses can be returned as invalid instances after the generic descriptor clone. See the inline comment on `src/redteam/providers/shared.ts`.
- **Existing feedback, not duplicated:** [LiteLLM composition wrappers are still uncapped](https://github.com/promptfoo/promptfoo/pull/9866#discussion_r3475177774). I independently reproduced that an attack-purpose `litellm:gpt-4.1` provider is returned unchanged with neither cap field set.
- **Existing feedback, still applicable after later test rewrites:** [the cap tests inherit the documented override from the invoking process](https://github.com/promptfoo/promptfoo/pull/9866#discussion_r3469440497). With `PROMPTFOO_REDTEAM_PROVIDER_MAX_TOKENS=7777`, the two focused suites fail 16 deterministic assertions because they expect the default 4096 instead of establishing/restoring an unset baseline.

No P0 or P1 defects were found.

The arbitrary `Object.create` clone should be replaced with a lifecycle that loads/reconstructs dedicated known providers or otherwise preserves the receiver semantics of caller-supplied instances. Also, the changed strategy/MCP tests do not assert that their manager calls include `purpose: 'attack'`; one representative wiring-level regression test would protect the opt-in that activates this fix.

## Validation performed

- `npx vitest test/redteam/providers/shared.test.ts test/redteam/providers/attackProviderOutputCap.test.ts test/redteam/mcpTargetProvider.test.ts test/redteam/providers/crescendo/index.test.ts test/redteam/providers/custom/index.test.ts test/redteam/providers/iterative.test.ts test/redteam/providers/iterativeImage.test.ts test/redteam/providers/iterativeMeta.test.ts test/redteam/providers/iterativeTree.test.ts test/redteam/providers/voiceCrescendo/index.test.ts test/redteam/strategies/mathPrompt.test.ts --run` — **327/327 passed**.
- `PROMPTFOO_REDTEAM_PROVIDER_MAX_TOKENS=7777 npx vitest test/redteam/providers/shared.test.ts test/redteam/providers/attackProviderOutputCap.test.ts --run` — **16 failed / 65 passed**, confirming the existing isolation defect.
- `npm run tsc` — passed.
- `npm run build` — passed, including the app production build.
- Direct request-body probe with `PROMPTFOO_REDTEAM_PROVIDER_MAX_TOKENS=1234 OPENAI_MAX_TOKENS=32768` — OpenAI attack `max_tokens=1234`, normal generation `max_tokens=32768`, Responses attack `max_output_tokens=1234`.
- Direct custom-subclass probe — the capped clone throws `TypeError: Cannot read private member #secret from an object whose class did not declare it`.
- Direct LiteLLM probe — attack provider remained the original wrapper with no `max_tokens` or `max_completion_tokens`.
- Exact-head CI — all reported checks succeeded, including Node 20/22/24/26, Windows/macOS/Linux, integration, smoke, redteam, web UI, docs/site, CodeQL, and Codecov.

The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable in this environment (exit 127), so I used the closest read-only six-lens review fallback plus three independent specialist passes. I did not make live provider API calls or run a credentialed end-to-end redteam campaign; request construction was exercised locally without secrets. I also relied on green exact-head CI for the site build rather than rebuilding the docs site locally.

PR #9867, opened 2026-06-24
fix(azure): avoid crash on empty choices array when reading logprobs

he-yufeng:
## What

The Azure chat parser builds a guarded `choice` from `data.choices` and then optional-chains it everywhere (`choice?.message`, `choice?.finish_reason`, `message?.content`, ...). There's even a comment in that block spelling out that an empty `choices` array makes `choices[0]` undefined. The logprobs read is the one spot that didn't follow suit:

```ts
logProbs = data.choices[0].logprobs?.content?.map(...);
```

The optional chaining only starts at `.logprobs`, so when `choices` comes back empty (e.g. every candidate was filtered) `data.choices[0]` is `undefined` and this throws `TypeError: Cannot read properties of undefined (reading 'logprobs')`. The surrounding `try/catch` turns that into an opaque `API response error: TypeError ...` instead of a normal empty result.

## Fix

Read logprobs off the already-guarded `choice` so the whole block treats an empty `choices` array consistently:

```ts
logProbs = choice?.logprobs?.content?.map(...);
```

## Test

Added a case under `guardrails and content filtering` that returns `choices: []`. It fails on `main` with the TypeError above and passes with the fix. Full `test/providers/azure/chat.test.ts` is green (50 tests), `tsc --noEmit` clean.


mldangelo:
Reviewed — the fix is correct and exactly minimal: `choice` is the already-guarded value, the adjacent comment documents the invariant, and the logprobs read was the one line that didn't follow it. I also confirmed the end state is right: a nullish `output` with no error routes the row to the evaluator's canonical `No output` outcome instead of the opaque `API response error: TypeError`.

Pushed two small extensions to your branch (thanks for enabling maintainer edits):

- Guarded the remaining degenerate shape in the same block: a 200 response with **no `choices` field at all** threw the same TypeError one line above your fix, at `data.choices.find(...)` / `data.choices[0]`. Both now take the same graceful path (`data.choices?.`), with a regression test.
- Strengthened the tests to assert the full outcome (`output` undefined, `tokenUsage` still billed) rather than only `error`/`logProbs`.
- Merged current main.

Both regression tests verified to fail on the pre-fix parser and pass with it; the full Azure suite (456 tests), tsc, and lint are green locally.

mldangelo (approved):
Really nice find and a textbook fix, @he-yufeng — you spotted the one line in the block that didn't follow its own documented guard, kept the change exactly minimal, and wrote the regression test that proves it. The graceful no-output path you chose is precisely how the evaluator wants to see this case. With the small extension for the missing-`choices` shape, this block is now safe against the whole degenerate-response class. Thanks for the contribution (and for the equally careful writeup in the PR body) — merging once CI is green. Looking forward to more!

Issue #9747, opened 2026-06-16
[feat]Support deleting individual eval results within an eval session


hdw868:
**Is your feature request related to a problem? Please describe.**
I run long evaluations that take 30–120 minutes to complete. When a specific test case fails due to a transient issue (API timeout, environment fluke), the only option is to delete the entire eval session — wiping out all valid results accumulated over the past hour — and re-run from scratch.

**Describe the solution you'd like**
Add the ability to delete individual eval results within a session, without deleting the entire eval.

- Web UI: Add a delete icon to the per-row hover action toolbar (the icon strip already visible in the top-right of each result row). It should appear on hover alongside the existing copy, link, star, tag, and edit icons, with a confirmation prompt before removal.
- CLI: promptfoo delete eval-result <resultId> or promptfoo delete eval <evalId> --result-idx <n>
- REST API: DELETE /api/eval/:evalId/results/:resultId
- 
**Describe alternatives you've considered**
- Delete the whole eval session and re-run — too destructive; wastes 30–120 min of compute for a single bad result.
- Manually edit the SQLite DB at ~/.promptfoo/promptfoo.db — unsupported and error-prone.

**Additional context**

The eval_results table already has a primary key id and is indexed by (eval_id, test_idx), so the underlying deletion is a single query. The main work is exposing it via the API, CLI, and UI hover toolbar.

PR #9868, opened 2026-06-24
feat(server): delete a single result within an eval session

mc856:
## Summary

Adds support for deleting a single persisted result from an eval without deleting the eval itself.

- `DELETE /api/eval/:evalId/results/:id`
- `promptfoo delete eval-result <id>`

This is the backend/CLI half of #9747. The UI hover-toolbar action is still a follow-up, so this uses `Refs #9747`.

## Notes

This deletes exactly one persisted eval result row/cell by result id. In multi-prompt evals, deleting one cell can leave sibling prompt outputs in the same test row, so table, export, comparison, and summary paths preserve alignment around those sparse cells instead of deleting sibling outputs.

Deleting a result updates prompt-level metrics in the same transaction so eval summaries, filters, and headers stay consistent. The delete is guarded by both `evalId` and `resultId`. Traces are left in place because they are eval-scoped rather than result-scoped.

For legacy/imported rows, the goal is defensive behavior: deletion should not throw, expose media that is no longer used, or persist invalid numeric aggregates. It does not try to perfectly reconstruct historical metric detail that was never persisted.

Function-valued derived metrics that depend on the original `evalStep` are left as a follow-up/open question because the delete path does not have the original runtime context available.

If the scope feels heavy for one PR, the sparse-cell handling in `evalTableUtils.ts` / `tokenUsageUtils.ts` / `namedMetrics.ts` and the legacy-row defensive paths could land as a follow-up — happy to break this up if preferred.

## Test plan

- `npm run tsc`
- `npx @biomejs/biome check src/util/database.ts src/util/eval/evalTableUtils.ts src/util/namedMetrics.ts src/util/tokenUsageUtils.ts test/util/databaseDelete.test.ts test/util/eval/evalTableUtils.test.ts test/util/tokenUsageUtils.test.ts`
- `npx vitest run test/util/databaseDelete.test.ts test/util/tokenUsageUtils.test.ts test/util/eval/evalTableUtils.test.ts --sequence.shuffle=false`
- `npx vitest run test/commands/delete.test.ts test/models/eval.test.ts test/server/routes/eval.validation.test.ts test/server/routes/serverRouteSmoke.test.ts test/util/databaseDelete.test.ts test/util/databaseStandalone.test.ts test/util/namedMetrics.test.ts test/util/tokenUsageUtils.test.ts test/util/eval/evalTableUtils.test.ts --sequence.shuffle=false`


mc856:
Pushed 3fd1459 addressing three Codex P2s on the delete path:

- `assertions.numRequests` was being recomputed to 0; now mirrors the live grading path via `accumulateGradingRequest`.
- Imported rows with non-numeric `namedScoreWeights` were getting NaN written back; fall back to the contributing assertion count instead.
- HTML export was column-misaligned for the sparse rows this PR produces (`row.outputs.map` skips holes); now iterates by `prompts.length`.

CSV (`tableRowToCsvValues`) and Sheets export (`prompts.forEach`) already iterate by `prompts.length`, so they were unaffected — leaving them alone to keep the diff focused.

mc856:
Pushed ee7bd1c addressing the remaining Codex P2s on `ff5b5d35e`, plus two delete-time bugs the recompute path introduced.

**Blob authorization (`src/util/database.ts`)**
- Survivor blob scan (`resultUsesStructuredBlobRef`) now requires a structured blob envelope — `hash` + `uri` + `mimeType` + numeric `sizeBytes` + `provider` — before treating a surviving row as a real media user. A row that merely echoes `promptfoo://blob/<hash>` in free-form text no longer keeps the deleted row's trusted reference alive, so share auth cannot leak media whose creating row was deleted. The old `resultUsesBlobHash` is renamed `resultMentionsBlobHash` and reused only for eval-level import cleanup, where a textual mention is the intended signal.
- Blob references still used by surviving `traces`/`spans` rows are downgraded to `location: 'trace'` (indices cleared) instead of being deleted, so trace views keep resolving `promptfoo://blob/...` after the creating result is gone. `traceUsesBlobHash` runs a scoped join per hash.

**Recompute correctness (`src/util/database.ts`)**
- `updatePromptMetricsForDeletedResult` only recomputes the aggregate assertion-token bucket when every surviving prompt row has a `gradingResult`. With `PROMPTFOO_STRIP_GRADING_RESULT=true` the stripped rows would otherwise zero out assertion tokens the forward path had credited; we keep the previous total in that case.
- `recomputeDerivedMetrics` skips function-valued derived metrics (see open question below).

**Sparse rows**
- `getEvalSummaries` uses `max(distinct-persisted, metric-derived)` for `testCount`; per-cell delete no longer shrinks the eval-list count.
- `DownloadMenu` failed-tests filter and disabled state treat `null` cells as absent (not failed).
- `DownloadMenu` human-eval export picks the first surviving output on each row instead of unconditionally reading `outputs[0]`, so deleting the first prompt cell no longer produces a case with empty `output` and a broken assertion.

**Open question — function-valued `derivedMetrics`**
The forward path invokes a function-valued `derivedMetric` with the per-row `RunEvalOptions` (`evalStep`). That context is not persisted, so a delete-time recompute cannot reconstruct it. Passing `{}` either throws or returns a value that ignores the metric's contract; I chose to leave function-valued derived metrics at their pre-delete value on the assumption that a stale correct-shaped value beats a wrong or missing one. Happy to switch to "drop the derived key on delete" if you'd rather force a recomputation next time the eval is opened.

**Read on the remaining Codex threads**

*V4 import input validation* — `src/util/database.ts:1060`, `src/util/tokenUsageUtils.ts:74`

Both share the same schema tolerance: `POST /api/eval` accepts each result row as `Record<string, unknown>`, so malformed `namedScores` or `completionDetails` already land in the persisted store. The forward `accumulateNamedMetrics` / `accumulateTokenUsage` paths have the same gap — a bad imported eval was producing `NaN`/negative totals before this PR; the delete path just adds another read where it shows up. Right fix is the save schema or the import route so bad rows don't persist.

*Manual-rating componentless grading* — `src/util/database.ts:514`, `:1038`, `:1039`

Three symptoms of one upstream decision: `POST /:evalId/results/:id/rating` replaces `gradingResult` with a pass/score-only object and drops the original `componentResults`. Once that count is gone, delete has no way to reconstruct the row's original assertion contribution — inferring 1 (or skipping via the null-fallback) is the best approximation on the persisted data. Real fix is for the rating route to preserve `componentResults` or persist a count alongside; happy to open a follow-up issue.

Happy to fix any of these in this PR if you'd rather — say the word.

mc856:
Pushed 400af88 addressing the five Codex P2s from the 01:39 review on ee7bd1c. Four are delete-path bookkeeping bugs; one is a pre-existing `--include-media` gap left as follow-up.

- `assertions.numRequests` was leaving one phantom request per graded-row delete because `subtractAssertionTokenUsage` didn't touch it; new `subtractGradingRequest` in `tokenUsageUtils` mirrors the forward helper.
- `namedScoreWeights[metric] === 0` was being dropped by a truthiness check during the null-grading recompute; switched to property presence so metric-only assertions survive.
- `getAssertionCounts` threw on `null` / non-object entries inside `componentResults` (imported V4 rows), blocking delete of the exact rows users wanted to clean up; filter to objects first.
- `subtractNamedMetric` was persisting `undefined - w = -w` for legacy aggregates that never tracked a metric; init with the row's contribution first, matching the existing count/weight buckets.

`exportBlobAssets` (`--include-media`) scans surviving result text for `promptfoo://blob/<hash>` and calls `getBlobByHash` directly, bypassing `blob_references`. The naive scan predates this PR — per-row delete just makes "creating row gone, echo remains" more likely to trip it. Real fix is to intersect with `blob_references` in `exportBlobAssets`; happy to open a follow-up.

mc856:
Pushed 8abc264e8 addressing the four Codex P2s on `400af88`:

- preserve imported eval-level blob references when a URI-only surviving result still uses them, while keeping structured matching for other references
- tolerate null and non-object `componentResults` when debiting named metrics
- clamp explicit and implicit `numRequests` debits at zero for under-credited imported aggregates
- clamp the five result/assertion count buckets at zero while leaving score, latency, and cost arithmetic unchanged

Validation:

- `npm run tsc`
- focused Biome check on the five changed files
- `npx vitest run test/util/databaseDelete.test.ts test/util/tokenUsageUtils.test.ts` (81 tests passed)
- `git diff --check`

PR #9872, opened 2026-06-24
fix(providers): harden malformed completion responses

he-yufeng:
## Summary

- Centralize completion-message and token-usage normalization for OpenRouter and Snowflake, returning bounded `ProviderResponse.error` results for missing, empty, or unusable first choices instead of throwing or grading malformed successes.
- Preserve valid text, OpenRouter structured content, tool/function calls, refusals, content-filter outcomes, first-choice semantics, finish reasons, latency, usage, cost, cache state, retries, and tracing.
- Handle OpenRouter choice-level errors and rate limits plus Snowflake HTTP-200 error envelopes without exposing provider-controlled partial output or private diagnostics.
- Preserve configured cost overrides, safely bound malformed accounting, aggregate retry usage/request/cost data, and keep provider slots balanced when retries or finalizers fail.
- Pass Snowflake registry options through without double nesting and keep provider-error results out of conversation history.

## Contract decisions

A successful non-streaming HTTP response without a usable `choices[0].message` is malformed provider data and follows Promptfoo's existing structured error contract. A later valid choice does not mask an unusable first choice. Failed partial output is neither returned nor graded.

Null content remains valid when accompanied by a complete tool/function call, refusal, or content-filter result. OpenRouter's documented structured content is preserved; Snowflake's current JSON completion path remains string/tool-call based. Streaming behavior and endpoint migration are outside this change.

## Verification

Final head: `7844063487a8672bf1c9be92d3f4401d621858dd`

- Focused provider/evaluator/scheduler suites: 8 files, 464 tests passed with shuffled seed `9872`.
- Full provider suite: 211 files passed, 1 skipped; 6,935 tests passed, 9 skipped.
- Relevant integrations: function-provider grading 4/4, select-best evaluator 1/1, CLI smoke 27/27.
- `npm run tsc`, `npm run architecture:check`, `npm run format:check`, and `npm run build` passed on the final dependency graph.
- Deterministic built-CLI loopback matrices ran with `--no-cache` and exported JSON was inspected for success, malformed and adversarial payloads, invalid JSON, HTTP and provider-body failures, multiple choices, tool calls, structured content, content filtering, retry exhaustion/recovery, usage/cost/latency, evaluator inversion, and redteam handling.
- OpenRouter produced 5 valid rows and 11 structured errors; Snowflake produced 3 valid rows and 12 structured errors. Malformed graders failed rather than passing or being inverted, and all 8 malformed redteam rows remained errors.
- Fresh cache sequences for both providers proved malformed-response eviction, uncached recovery, then a true cache hit, with exactly two loopback requests per provider and no existing cache cleared.


mldangelo-oai:
## Final review receipt

Head: `7844063487a8672bf1c9be92d3f4401d621858dd`

Disposition: **FIXED_AND_READY**. The substantive review, skeptical validation, simplification pass, and fresh second review are clean. The PR is mergeable and fully synchronized with live `main` at `689a304e195d7f00a27a53ed92bed47e4fd30c88`; it remains intentionally unmerged and blocked only on repository approval/ruleset gates.

### Additive commits

- `1d1bb6eb87674d2f4fe4f060dccf5f220a399c78` — validate malformed completion payloads.
- `abddb5d08a380d7ff72f90e14eea37cda91a1d30` — preserve response, retry, accounting, and scheduler contracts.
- `c15800b9e063daa40cc2852179e28c2ea6979ec3` and `7844063487a8672bf1c9be92d3f4401d621858dd` — additive merges of intervening `main` dependency updates; no history was rewritten.

### Validated repairs

- OpenRouter and Snowflake now fail closed with bounded structured errors for null, missing, empty, or unusable first choices instead of throwing or allowing malformed successes into assertions/redteam evaluation.
- Shared normalization preserves valid text, OpenRouter structured content, complete tool/function calls, refusal/content-filter results, first-choice selection, finish reasons, usage, cost, latency, cache state, retries, and tracing.
- OpenRouter choice errors and rate limits plus Snowflake HTTP-200 error envelopes are handled without returning provider-controlled partial output or private diagnostics.
- Malformed accounting is bounded, local cost overrides remain authoritative, exhausted retries retain aggregate usage/request/cost data, and scheduler slot release remains balanced even when finalizers fail.
- Snowflake registry options are no longer double nested, and provider-error responses are not written into conversation history.

Contract references reviewed: [OpenRouter response schema](https://openrouter.ai/docs/api/reference/overview), [OpenRouter error responses](https://openrouter.ai/docs/api/reference/errors-and-debugging), and [Snowflake Cortex REST API](https://docs.snowflake.com/en/user-guide/snowflake-cortex/cortex-rest-api).

### Verification

- Focused provider/evaluator/scheduler suites: 8 files, 464/464 tests passed with shuffled seed `9872`.
- Full provider suite: 211 files passed, 1 skipped; 6,935 tests passed, 9 skipped.
- Relevant integrations: function-provider grading 4/4, select-best evaluator 1/1, CLI smoke 27/27.
- `npm run tsc`, `npm run architecture:check`, `npm run format:check`, `npm run build`, and `git diff --check` passed on the final dependency graph. The repository-wide formatting gate reported only its existing 169 complexity warnings and exited successfully.
- Built-CLI deterministic loopback matrices ran with `--no-cache`; exported JSON was inspected rather than relying on exit status. OpenRouter produced 5 valid rows and 11 expected structured errors; Snowflake produced 3 valid rows and 12 expected structured errors.
- Valid tool calls, structured OpenRouter content, content filtering, usage/cost/finish reasons, first-choice behavior, HTTP/JSON failures, provider-body errors, retry recovery/exhaustion, evaluator inversion, and redteam paths were exercised. Malformed graders did not pass or invert, and all 8 malformed redteam rows remained errors.
- Fresh cache sequences for both providers proved malformed eviction, uncached recovery, then a true cache hit, with exactly two loopback requests per provider and no cache cleared.
- Private choice/error diagnostics, partial output, and later-choice masking sentinels were absent from the exported results.

### Final GitHub state

- Exact-head CI: **31 successful, 2 workflow-skipped, 0 failed, 0 cancelled, 0 pending**.
- Review threads: **0 total, 0 unresolved**; reviews: **0**.
- Title and body accurately describe the final implementation and contract decisions.
- GitHub reports `MERGEABLE`, with `mergeStateStatus: BLOCKED` and `reviewDecision: REVIEW_REQUIRED` because the active ruleset requires one code-owner approval and names a `CodeQL` context that was not emitted for this fork head. No protection was bypassed.
- No merge, close, approval, force-push, amend, rebase, squash, cache clearing, competing PR, or branch-history rewrite was performed.

Compatibility intentionally retained: the first choice remains authoritative; OpenRouter structured content and complete tool calls remain valid; Snowflake's existing JSON/non-stream transport remains in place. Snowflake streaming/endpoint migration is outside this PR.

he-yufeng:
This was marked FIXED_AND_READY back on 06-25 but is still open. Is there anything blocking the merge, or would a rebase help? Happy to do whatever is needed to move it along.

mldangelo (commented):
Thanks for sticking with this one — and sorry it's sat so long. I went through the branch carefully and ran the affected suites; here's where things stand.

## What's working well

- Hoisting `validateChatCompletionMessage` / `getChatCompletionRefusal` / `parseChatCompletionJsonOutput` into `src/providers/openai/util.ts` is the right shape for this fix. OpenRouter and Snowflake were both hand-rolling the same choice/message unwrapping, which is how the empty-`choices` crash happened in the first place.
- The `it.each` malformed-payload tables in `test/providers/openrouter.test.ts` and `snowflake.test.ts` are exactly the coverage this code was missing. Nice.
- Bounding provider-controlled numbers (`isSafeTokenCount` / `isSafeCost`) is a genuinely good instinct — unbounded `usage.cost` flowing into eval cost aggregation is a real hazard.
- Retry-aware token/cost accumulation in `providerWrapper.ts` closes an accounting gap that predates this PR.

The branch is green in isolation: 7,272 provider tests pass on head `cbc5c4d766`.

## Answering your rebase question directly: yes, and it will turn CI red

Head is behind current `main`. I merged `origin/main` into a scratch worktree — `src/` auto-merges with **no conflict**, but two tests then fail deterministically (reproduced 3/3):

```
test/providers/openai/util.test.ts > getTokenUsage > should read cached tokens
  from top-level usage.cached_tokens (e.g. Moonshot)
- "completionDetails": { "cacheReadInputTokens": 32 },   ← expected
+ (absent)                                                ← received

test/providers/moonshot.test.ts > reads cached tokens from the documented
  top-level usage.cached_tokens field
  expected undefined to be 4
```

### P1 — `getTokenUsage` drops the top-level `usage.cached_tokens` fallback

`src/providers/openai/util.ts:889` sources `cacheReadInputTokens` from `promptDetails?.cached_tokens` only. `main` gained a `?? usage.cached_tokens` fallback in #10132 (Moonshot/Kimi report cached prompt tokens there). Because #10132 edited `getOpenAICompletionTokenDetails` while this PR rewrote the separate `getTokenUsage`, git merges silently — nothing flags the loss.

Regression proof: I swapped in `git show origin/main:src/providers/openai/util.ts` and reran the same test file unchanged — 3/3 pass on main, 2/3 fail on this branch. Restored afterward.

```
getTokenUsage({usage:{total_tokens:100,prompt_tokens:40,completion_tokens:60,cached_tokens:32}}, false)
  main → { ..., completionDetails: { cacheReadInputTokens: 32 } }
  PR   → { numRequests:1, total:100, prompt:40, completion:60 }
```

Impact is metrics-only (Moonshot's `calculateResponseCost` reads `cached_tokens` itself, so billing stays correct) — but it's a deterministic CI break.

The clean fix collapses this with a simplification: keep calling `getOpenAICompletionTokenDetails` (`util.ts:831`, still exported and still used by `src/providers/responses/processor.ts:52`) and apply the new `isSafeTokenCount` filter to *its* result. That preserves your bounding goal, deletes the ~20-line `detailMappings` duplicate, and restores both fallbacks — including the `output_tokens_details` one the bot already flagged at `util.ts:883`. I verified this turns the post-merge suite green.

### P2 — over-strict `isToolCall` discards the whole message

`isToolCall` (`util.ts:928`) requires `type === 'function'` exactly; `validateChatCompletionMessage:974-979` rejects the **entire** message if any element fails. Run through the real `OpenRouterProvider.callApi`:

```
tool call missing `type: 'function'`
  main → output: [{ id:'call_1', function:{ name:'get_weather', arguments:'{"city":"SF"}' } }]
  PR   → error: "Malformed response data: expected choices[0].message"

`arguments` as a pre-parsed object → same before/after split
```

A well-formed tool call (id + type + string arguments) is identical on both, so the strictness is precisely scoped to these two quirk shapes. Your own malformed table covers `[null]`, missing-`id`, and custom-without-`input` — but not these, so I read them as unintended collateral rather than intent. Note the guard runs before content handling, so a response with usable *text* plus one odd `tool_calls` entry is discarded too.

Was requiring the `type` discriminator deliberate? If not, dropping only unrecognized entries (or defaulting `type` to `'function'`) would keep your table's assertions passing. I could not confirm from live traffic how often OpenRouter emits these shapes, so treating this as a question rather than a must-fix.

### On the empty-content point already flagged at `util.ts:1000`

Not re-litigating it, just one data point that may bear on severity: on **this commit, in one process**, byte-identical payloads diverge.

```
{ message: { content: '' }, finish_reason: 'length' }
  openai:gpt-4o → output: ""     (openai/chat.ts:679)
  openrouter    → error: "Malformed response data: expected choices[0].message"
```

Since `src/evaluator.ts:1152` short-circuits on `response.error`, those rows leave pass/fail accounting entirely — which shifts eval pass-rate and redteam ASR denominators for OpenRouter/Snowflake targets. Empty output on a 200 is the most common way a guarded target signals refusal. Worth a second look at whether the hardening was meant to cover empty completions or only unparseable ones.

### P2 (docs) — guardrails provider list is now stale

`getChatCompletionRefusal` makes OpenRouter and Snowflake emit `guardrails: { flagged: true }`; neither did before. `site/docs/configuration/expected-outputs/guardrails.md:15-20` still says only Bedrock and Azure OpenAI support it, so `not-guardrails` assertions flip behavior for `openrouter:` targets with nothing documenting it.

## Two small cleanups

- `src/scheduler/rateLimitRegistry.ts:74-76` — the explicit `getHeaders` / `isRateLimited` / `getRetryAfter` lines are dead once `...options` (line 77) spreads them. Every consumer reads them optionally, so absent and present-but-undefined behave identically. Leaves `{ ...options, maxRetriesOverride }`.
- `isSafeTokenMetric` (`providerWrapper.ts:45`) is byte-identical to `isSafeTokenCount` (`util.ts:755`), and the inline cost guard at `providerWrapper.ts:145-149` is byte-identical to `isSafeCost` (`util.ts:759`) — three copies of two predicates. `src/util/numeric.ts` is already the home for untrusted-value sanitizers, and layering permits both `providers` and `core` to import it.

Only the P1 is blocking. It's a two-line restoration (or the helper collapse above), and the rest are questions. Happy to look again once you've synced with main.

he-yufeng:
Still green after the latest CI rerun, and the review receipt from June stands. If the fix has fallen off the radar I am happy to rebase again or adjust anything; just let me know what would help.

he-yufeng:
Synced with main (fd8676e0). Conflicts were upstream's `appendOpenAiApiPath` URL normalization and its naive empty-choices guards; both merged into the validated path, and upstream's two snowflake tests now sit next to the `it.each` table. On your points:

**P1.** Verified in the merged tree: `getTokenUsage` reads `promptDetails?.cached_tokens ?? usage.cached_tokens` with the `isSafeTokenCount` bound applied per field, and `getTokenUsageWithRequestCount` just wraps it. `moonshot.test.ts`'s three cached-tokens cases pass, so the #10132 fallback survives this branch.

**isToolCall strictness.** Deliberate, but I take the question seriously. The validator accepts the two documented wire shapes (`type: 'function'` with string `arguments`, and the custom-tool shape) and rejects the message otherwise. The alternative you sketched, dropping only the unrecognized entries, changes the tool set the model asked for with no signal anywhere, which felt worse than one loud structured error. If OpenRouter emits the quirk shapes in real traffic I would rather normalize them explicitly (default `type` to `'function'`, stringify object arguments) than silently drop, but I would want evidence of the shape first.

**Empty content.** Same call, kept strict. On these two providers an empty message on a 200 was previously indistinguishable from the crash class this PR fixes, and an errored row is at least visible in the eval. Your denominator point is fair: `main` returns `output: ""` and keeps the row in pass/fail, this branch takes it out. If you want parity with `openai:gpt-4o` there, a carve-out for `content: ''` with a real `finish_reason` is a few lines, tell me which read you want.

**Docs.** Done. Upstream rewrote the guardrails section since, so I carried the OpenRouter/Snowflake coverage into the new normalization table instead of the old bullet list. `openrouter.md` documents the refusal/filter flagging too.

**Cleanups.** The dead `getHeaders`/`isRateLimited`/`getRetryAfter` lines in `rateLimitRegistry` are gone, exactly your `{ ...options, maxRetriesOverride }`. The predicate collapse into `util/numeric.ts` I left out on purpose to keep this diff from growing further; happy to send it as a fast follow-up.

Verified locally: openrouter + snowflake + moonshot (158 tests), util + runEval + scheduler + registry (376 tests), `architecture:check`, biome and prettier on the touched files. `tsc` is clean apart from the pre-existing `@openai/codex-security` optional-dependency noise, which this diff does not touch.

he-yufeng:
@mldangelo Synced with main (`ec6d154b6`) and worked through the review. All four points, in order:

**P1 (cached_tokens fallback).** Already restored at the head you have now — `getTokenUsage` carries the full chain (`prompt_tokens_details` / `input_tokens_details` / top-level `cached_tokens`), and your two regression cases pass on the merged branch. Post-merge I re-ran the affected suites: `test/providers/openai/util.test.ts` + `moonshot.test.ts` 262/262 green, `openrouter`/`snowflake`/`registry` 191+43+167 green, plus `providerWrapper`/`providerRateLimitState`/`runEval` — 757 across the battery. On the collapse suggestion: main's OpenRouter now has its own bounded `calculateResponseCost` (`isNonNegativeFiniteNumber` throughout), so my `calculateSafeOpenAICost` is dropped from openrouter.ts entirely and the success path reuses the class method. `getTokenUsageWithRequestCount` stays as the thin numRequests wrapper over the shared function.

**P2 (isToolCall strictness).** Not deliberate, and your two quirk repros were the deciding evidence. An omitted discriminator on a complete function payload now reads as `'function'` (`27b05de67`); an explicit unknown type and a `custom` without input still fail, and the malformed table keeps passing. New regression test: a no-`type` call with full payload survives end to end through `OpenRouterProvider.callApi`, output identical to main's passthrough shape.

**Empty-content divergence.** Real and still there, and I think you're right about the severity: `{content:'', finish_reason:'length'}` on openrouter errors while openai returns `""`, and since `evaluator.ts` short-circuits on `response.error`, those rows leave pass/fail accounting entirely. My read: for the chat-completions contract an empty string on a 200 is a degenerate completion, not a malformed payload — the hardening was aimed at unparseable/missing structure, and empty content is neither. I haven't flipped it in this push because it changes what my malformed table asserts (several rows exist precisely to pin the error), but if you confirm the eval-accounting semantics you want, the flip is a small change and I'll fold it in with the tests.

**P2 docs.** The `guardrails.md` provider table already carries the "OpenRouter and Snowflake Cortex" row on this branch — added in an earlier push.

**Cleanups.** The rateLimitRegistry dead lines are gone already; the predicate duplication is fixed by moving `isSafeTokenCount`/`isSafeCost` into `src/util/numeric.ts` and importing from both sites (`27b05de67`), which also deleted the byte-identical `isSafeTokenMetric`.

CI will confirm the merge; happy to adjust any of the above.

he-yufeng:
Follow-up on the push just now (`9f8efdfc7`): the predicates consolidation adds two intentional cross-layer import edges (`providers -> node` and `core -> node`, both into `src/util/numeric.ts`), so the architecture-boundary baseline is bumped for exactly those two. I deliberately did not fold in main's unrelated baseline drift (seven edges shrank below baseline on their own); the note from the checker covers it if you want that locked separately.

he-yufeng:
Addressed the codex P2 on `src/util/numeric.ts` in `145cb3219`: `isSafeTokenCount` now caps at a billion tokens per response, so `Number.MAX_SAFE_INTEGER` no longer passes as legitimate usage and two garbage rows cannot overflow safe-integer accumulation downstream. New `test/util/numeric.test.ts` pins the predicate boundaries (zero, negatives, non-integers, the MAX_SAFE_INTEGER case, the 1e9 edge), and the provider path got an end-to-end case showing the absurd row is dropped while valid fields survive. Red-checked both new cases against the old predicate (they fail), green with the fix; `tsc`, Biome, and the three touched suites (250 tests) pass.

On the sibling field: `isSafeCost` stays finite-only on purpose. Cost garbage is bounded by `Number.isFinite`, and an overflow there produces Infinity, which the same predicate refuses on the next read, so it cannot re-enter as valid data the way an unsafe integer can.

he-yufeng:
Merged current main and reconciled with the response-path hardening that landed there in the meantime.

What changed on this branch as a result:

- Kept main's policy-refusal block, choice-level error block, and the openrouterBilling extraction; this PR now only carries what main still lacks.
- Residual fixes re-applied on top: the malformed-200 path returns a bounded error string (no provider-controlled payload dump) and evicts the cache entry; `finish_reason: 'error'` without a choice error object no longer grades partial output as a success (main's `getOpenAiChatChoiceError` requires the error object); array content parts (Gemini via OpenRouter) no longer throw on `.trim()`; choice-level 429/502/503/504 codes now classify into the scheduler's retry kinds (`getOpenAiGatewayRateLimitKind` only reads the top-level envelope); token usage carries the request count.
- The test battery was reconciled against main's contracts where both cover the same case (error strings carry the provider message, `metadata.http.status` records the transport truth, classification lives in the kind fields).

Focused suites on the merged head: openrouter 167/167, snowflake + openai/util + registry + scheduler + evaluator + numeric 608/608. Biome is clean apart from the pre-existing callApi complexity warning that main already carries.

PR #9884, opened 2026-06-26
feat(assertions): add cost and latency selectors

ianw-oai:
Promptfoo could select outputs by subjective quality or aggregate score, but not by provider-reported cost or measured latency. This adds deterministic `select-lowest-cost` and `select-lowest-latency` assertions, including an `onlyPassing` option for choosing the cheapest output that passes the other assertions, plus schema, Eval Creator, and documentation support.

The selectors reject cached or invalid metrics and update result statistics through the existing comparison path. Validated with the full build, focused matcher, evaluator, and Eval Creator tests, a no-cache CLI smoke test, and the documentation build.


src/matchers/comparison.ts
```diff
@@ -117,15 +234,15 @@ export async function selectMaxScore(
     // Get component results from gradingResult if available
     const componentResults = result.gradingResult?.componentResults || [];
 
-    // Filter out max-score and select-best assertions
+    // Selection assertions are evaluated separately and must not contribute to max-score.
     const relevantResults = componentResults.filter(
       (r: GradingResult) =>
-        r.assertion && r.assertion.type !== 'max-score' && r.assertion.type !== 'select-best',
+        r.assertion && r.assertion.type !== 'max-score' && !r.assertion.type.startsWith('select-'),
```
mldangelo, line 268: **P1 — this filter change makes `[metric selector, max-score]` an eval-killing config, and the UI gate wasn't updated to match.**

Reproduced with the local build: `assert: [{type: select-lowest-cost}, {type: max-score}]` across two providers → the metric selector runs first and pushes its componentResults, this filter then excludes them all, `relevantResults` is empty, and the throw below propagates uncaught through `evaluate()` — after every provider call has been paid for, with no output file written.

Two fixes needed:
1. `setupReadiness.ts:342` (`hasMaxScoreWithoutScoringAssertion`) still checks `type !== 'select-best' && type !== 'max-score'`, so a `select-lowest-*` sibling counts as a scoring assertion and the eval-creator declares this exact config ready. It should mirror this filter (`!type.startsWith('select-')`).
2. Ideally the comparison phase should fail the row rather than the eval — the throw-through-`evaluate()` shape is pre-existing for `[select-best, max-score]`, but this PR makes it newly expressible from a UI-blessed config.

src/matchers/comparison.ts
```diff
@@ -13,6 +13,123 @@ import type {
...
+      typeof assertion.value === 'object' &&
+      'onlyPassing' in assertion.value &&
+      assertion.value.onlyPassing === true);
+  const failAll = (reason: string) =>
+    results.map(() => ({
+      pass: false,
+      score: 0,
+      reason,
+    }));
+  const eligible = results.flatMap((result, index) =>
```
mldangelo, line 89: **P1 — an errored sibling fails every output in default (all-outputs) mode.**

With `onlyPassing` unset on `select-lowest-cost`, eligibility includes rows whose provider call *errored* (success=false, `response` undefined → `response?.cost` undefined). The invalid-metric check below then `failAll`s the entire row.

Reproduced with the local build: 3 providers, one returning `{error: 'simulated 500'}` → all three outputs fail with `select-lowest-cost requires every eligible output to report a finite, non-negative cost; missing or invalid cost for prompt indexes: 2`. A transient error on one provider destroys the comparison verdict for the healthy ones — precisely the rows the user wanted compared.

Consider excluding *errored* rows (`result.error` set / no response) from eligibility in default mode — they can never report a metric, and "compare all outputs" intuitively means all outputs that produced anything. That's distinct from assertion-failed outputs, which default mode deliberately includes.

mldangelo (commented):
## Review: `select-lowest-cost` / `select-lowest-latency`

Useful feature with a genuinely well-factored core: the data-driven `selectorDefinitions` table makes future metric selectors a two-line change, the merge/apply consolidation with max-score is a real dedup, the promptIdx tie-break turns out to be load-bearing (the persisted `getResultsToCompare` path has no `ORDER BY`), and the architecture layering complies with the edge baseline. **I verified the happy paths end-to-end with the local build**: winner selection works for both selectors, `onlyPassing` behaves as documented in both directions, and losers are correctly failed with clear reasons. All 264 touched tests pass, `config-schema.json` is an exact regeneration, the branch merges clean with main, and CI is green.

However, I also **reproduced three failure modes end-to-end** that I think need attention before merge — one eval-killing crash newly expressible by this PR, and two all-outputs-fail footguns users will hit quickly.

### Reproduced (local build, real evals)

1. **`[select-lowest-cost, max-score]` kills the entire eval** (inline at `comparison.ts:240`). The selector runs first and pushes its componentResults; `selectMaxScore`'s new `startsWith('select-')` filter then finds nothing to aggregate and throws — uncaught, through `evaluate()`, after all provider calls were paid for, with no output written. Compounding it, the eval-creator's readiness gate wasn't updated for the new filter (`setupReadiness.ts:342` still treats `select-lowest-*` as scoring assertions), so the UI declares exactly this config ready. This is a plausible config ("pick winner, tie-break by cost").

2. **One errored sibling fails every output in default cost mode** (inline at `comparison.ts:79`). With `select-lowest-cost` (no `onlyPassing`), an errored provider row (success=false, no response → no cost) stays *eligible*, trips the invalid-metric check, and `failAll` marks the healthy siblings failed too. Reproduced with 3 providers where one returns an error: all three rows fail with "requires every eligible output to report a finite, non-negative cost". A transient 500 on one provider nukes the whole comparison.

3. **Divergent winners fail everything** (inline at `evaluator.ts:4036`). `[select-lowest-cost, select-lowest-latency]` on one test where the cheapest ≠ fastest: each output wins one selector and loses the other, the AND-merge (`nextPass = previousPass && comparisonPassed`) fails both, and the row reports 0% pass. Same class of interaction with select-best/max-score (metric selectors run first and read eligibility before those apply). None of this ordering or AND semantics is documented anywhere.

### Verified by code trace (not run)

4. **Error-graded winner gets `score: 1` while `success: false`** (inline at `evaluator.ts:1919`). The new `existingComponentResults.length === 0` branch plus the relaxed `updateComparisonStats` (`if (result.score !== wasScore)`, previously failure-only) lets a row whose grading *errored* (`applyGradingError` sets success=false and leaves `gradingResult` undefined) win default cost selection and end up with score 1, inflating prompt-level `metrics.score` for an errored row.

### Also worth fixing

- **Single-output rows crash the whole eval** — reproduced, but it's exact parity with `select-best` (I confirmed select-best crashes identically), so it's a pre-existing sharp edge these new types make more reachable. `selectMetric` already has the `failAll` helper; converting the `>= 2` invariant into a failAll would degrade gracefully.
- **No value-shape validation**: `value` is `z.custom` (accepts anything), so `{onlyPassing: 'true'}`, `{onlyPasing: true}` (typo), or a bare string silently mean "false", and `{onlyPassing: false}` on `select-lowest-latency` is silently ignored (the type check short-circuits before the value is read). A small Zod refinement for these two types would catch all of it at config time.
- **A selector nested in an `assert-set` is a silent no-op** (validates, never runs, and can zero the set's score).

### Docs
The two new sections are accurate on the happy path but: "cached outputs … fail the comparison" reads as per-output when it's actually fail-*all*; the forced `onlyPassing` on latency (and that `onlyPassing: false` is ignored) is unstated; there's no mention of which providers report cost (the adjacent `cost` section has the list — link it); and nothing documents processing order (metric selectors → select-best → max-score), the AND semantics of combining comparison assertions (#3 above), or cross-links to/from `select-best`/`max-score` (whose pages don't mention the new siblings). Minor: "ties are resolved by prompt order" could say the earliest prompt wins.

### Cleanup (non-blocking, inline where anchored)
- Dead generality: both `selectorDefinitions` entries hardcode `direction: 'min'`; the `'max'` ternary arms and the `objective` derivation are unreachable.
- Hot loop: `trackMetricSelectorRows` allocates a throwaway `.filter()` array per completed row where the sibling checks use short-circuiting `.some()`; the `as Assertion[]` cast could be a type-guard filter.
- Persisted-mode I/O: rows mixing comparison types now re-read the same results up to 3× (was 2× pre-PR), and a row with both selectors writes each result twice (`saveResult` per selector pass) — doubling exposure to the known SQLite shared-cache lock flakes.
- The `Promise.all` in `processMetricSelectorAssertions` wraps purely synchronous `selectMetric` calls — a plain `.map()` gives the same compute-all-then-apply guarantee without implying parallel I/O.
- Test gaps: no evaluator-level test covers a metric selector coexisting with max-score/select-best (the interaction this PR changed `selectMaxScore`'s filter for), and the tie-break's determinism is never tested with results ordered differently from promptIdx (the only ordering where it matters).

src/evaluator.ts
```diff
@@ -3970,8 +4009,69 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
...
+    let compareCount = 0;
+    for (const [testIdx, assertions] of metricSelectorRows) {
+      const results = await this.getResultsToCompare(testIdx);
+      if (results.length === 0) {
+        logger.warn(`Expected results to be found for test index ${testIdx}`);
+        continue;
+      }
+
+      // Calculate all selectors before applying any so they share the same eligibility state.
+      const pending = await Promise.all(
```
mldangelo: **P2 — combining comparison assertions ANDs their verdicts, so divergent winners fail the whole row; the ordering contract is also implicit.**

Reproduced: `[select-lowest-cost, select-lowest-latency]` where the cheapest output isn't the fastest → each output wins one selector, loses the other, and `nextPass = previousPass && comparisonPassed` fails both. 0% pass on a row where each selector individually had a clear winner. The same applies when combining a selector with select-best/max-score (finding a different winner). None of this — nor the processing order (metric selectors → select-best → max-score), nor that `onlyPassing` eligibility is read *before* select-best/max-score apply — is documented. If AND is the intended semantic, docs need a "combining selectors" section; a runtime warning when 2+ selectors share a row would also save users a confusing all-fail.

Two smaller things in this method:
- The `Promise.all` wraps purely synchronous `selectMetric` calls — a plain `.map()` gives the identical compute-all-then-apply guarantee without implying parallel I/O.
- With both selectors on a row, the apply loop runs `finalizeComparisonGrading` → `store.saveResult` per result **per selector** — 2N DB writes where N would do (and doubled exposure to the known SQLite shared-cache lock flakes). The pending results are already computed up front, so the merges per result could be folded before a single save.

src/evaluator.ts
```diff
@@ -1906,14 +1906,19 @@ function mergeSelectBestGradingResult(
...
+  gradingResult: GradingResult,
+) {
   const existingComponentResults = result.gradingResult?.componentResults || [];
   const existingGradingResult = result.gradingResult;
   const comparisonPassed = gradingResult.pass;
   const previousPass = existingGradingResult?.pass ?? result.success;
   const nextPass = previousPass && comparisonPassed;
   const newScore = comparisonPassed
-    ? (existingGradingResult?.score ?? result.score)
+    ? existingComponentResults.length === 0
```
mldangelo: **P2 — the new empty-componentResults branch lets an error-graded row win with `score: 1` while `success: false`.**

`applyGradingError` sets `row.success = false; row.score = 0` but never assigns `gradingResult`, so `existingComponentResults` is empty. If that row is the cheapest under default `select-lowest-cost` (errored rows are eligible — see comparison.ts:79), the selector marks it `pass: true, score: 1`; this branch then takes `gradingResult.score` (1), and the relaxed `updateComparisonStats` (`if (result.score !== wasScore)` — previously failure-only) adds +1 to prompt-level `metrics.score` for a row whose grading errored. Success stays false (`previousPass` is false), so the row displays as failed-with-score-1 and inflated prompt metrics.

Also a readability nit: two of the three ternary branches yield `gradingResult.score`, which the nesting hides —

```ts
const preserveExistingScore = comparisonPassed && existingComponentResults.length > 0;
const newScore = preserveExistingScore
  ? (existingGradingResult?.score ?? result.score)
  : gradingResult.score;
```

src/evaluator.ts
```diff
@@ -3970,8 +4009,69 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
...
+    isWebUI: boolean;
+    metricSelectorRows: MetricSelectorRows;
+    progressBarManager: ProgressBarManager | null;
+    prompts: CompletedPrompt[];
+    runEvalOptions: RunEvalOptions[];
+  }) {
+    let compareCount = 0;
+    for (const [testIdx, assertions] of metricSelectorRows) {
+      const results = await this.getResultsToCompare(testIdx);
+      if (results.length === 0) {
```
mldangelo, line 4059: **P2 — a single-output row still crashes the whole eval (select-best parity, but now easier to hit).**

This guard only covers `length === 0`; one prompt × one provider gives `length === 1`, and `selectMetric`'s `>= 2` invariant throws uncaught through `evaluate()` — after all provider calls, with no output file. Reproduced with the local build; I also confirmed `select-best` crashes identically, so this is inherited parity rather than a new bug. But these deterministic selectors are much easier to misconfigure this way (add one to a single-provider config), and `selectMetric` already has the `failAll` helper — converting the arity invariant into `failAll('requires at least two outputs…')` (or guarding `length < 2` here) degrades gracefully instead of losing the run. The eval-creator warns about this (`hasInsufficientMetricSelector`) but CLI users get the crash.

src/matchers/comparison.ts
```diff
@@ -13,6 +13,123 @@ import type {
...
+    `Unsupported selector: ${assertion.type}`,
+  );
+  invariant(
+    results.length >= 2,
+    `${assertion.type} assertion must have at least two outputs to compare between`,
+  );
+
+  const definition = selectorDefinitions[assertion.type];
+  const values = results.map(definition.value);
+  const onlyPassing =
```
mldangelo: **P2 — `onlyPassing` is unvalidated and partially unhonorable.**

`AssertionSchema.value` is `z.custom<AssertionValue>()` (accepts anything), so all of these pass config validation and silently mean `onlyPassing: false`: `value: {onlyPassing: 'true'}` (YAML string), `value: {onlyPasing: true}` (typo), `value: onlyPassing`. And on `select-lowest-latency` the leading type check short-circuits, so an explicit `value: {onlyPassing: false}` is silently ignored — the user has no way to rank all outputs by latency, and no signal that their option did nothing.

Suggest a Zod refinement for these two types constraining `value` to `{onlyPassing?: boolean} | undefined` (rejecting everything else at config time), and either honoring an explicit `onlyPassing: false` on latency or failing on it — silently swallowing it is the worst of both. Declaring per-type *defaults* in `selectorDefinitions` (`defaultOnlyPassing: true` for latency) would put the policy next to the rest of the selector definition instead of inside this expression.

src/evaluator.ts
```diff
@@ -2690,10 +2693,26 @@ function isTracingEnabledForTest(testSuite: TestSuite, testCase: AtomicTestCase)
...
 }
 
+type MetricSelectorRows = Map<number, Assertion[]>;
+
+function trackMetricSelectorRows(
+  rows: MetricSelectorRows,
+  assertions: AssertionOrSet[] | undefined,
+  testIdx: number,
+) {
+  const selectors = assertions?.filter((assertion) =>
```
mldangelo: **P3 — selectors nested in an `assert-set` are silent no-ops; plus two mechanical nits.**

1. This filter only sees top-level entries, and the per-output runner also skips `select-*` types — so a metric selector inside an `assert-set` validates fine, never runs anywhere, and (if it's the set's only child) leaves the set aggregating zero weight. Either support it, reject it in `validateAssertions`, or warn.
2. This runs per completed row in the hot loop and allocates a throwaway array even when no selector exists; the sibling checks use short-circuiting `.some()`. A `.some()` guard before the `.filter()` keeps the hot path allocation-free.
3. The `as Assertion[] | undefined` cast can be an honest type guard: `.filter((a): a is Assertion => isMetricSelectorAssertionType(a.type))`.

src/matchers/comparison.ts
```diff
@@ -13,6 +13,123 @@ import type {
...
+export type MetricSelectorAssertionType = (typeof METRIC_SELECTOR_ASSERTION_TYPES)[number];
+
+type MetricSelectorCandidate = {
+  latencyMs?: number;
+  promptIdx: number;
+  response?: { cached?: boolean; cost?: number };
+  success: boolean;
+};
+
+const selectorDefinitions = {
```
mldangelo, line 32: **Nit — dead `direction`/`objective` machinery.**

Both entries hardcode `direction: 'min'` and no `select-highest-*` type exists, so the `: value > winnerValue` arm and the `'highest'` objective are unreachable, and the field signals a max variant that doesn't exist. Drop `direction`, inline `value < winnerValue`, use the literal `'lowest'` — reintroduce the generality when a max selector actually lands.

(For the record: the promptIdx tie-break below is *not* dead code — the persisted `getResultsToCompare` path has no `ORDER BY` and the in-memory path is completion-order under concurrency, so array order can disagree with promptIdx. Worth a test that feeds tied values in non-promptIdx order to pin that determinism.)

test/evaluator/metrics.test.ts
```diff
@@ -413,6 +413,113 @@ describeEvaluator('evaluator metrics and scoring', () => {
...
+    };
+
+    const testSuite: TestSuite = {
+      providers: [provider],
+      prompts: [toPrompt('Cheap prompt'), toPrompt('Valid prompt')],
+      tests: [
+        {
+          assert: [
+            { type: 'contains', value: 'hello' },
+            { type: 'select-lowest-cost', value: { onlyPassing: true } },
```
mldangelo, line 469: **P3 — missing interaction coverage.**

No evaluator-level test covers a metric selector coexisting with `max-score` or `select-best` on the same row — the exact interaction this PR changed `selectMaxScore`'s componentResults filter for. Both directions are currently broken or surprising and untested: `[select-lowest-cost, max-score]` crashes the eval (see comparison.ts:240), and divergent winners across two selectors fail every output. Whatever the intended semantics, they should be locked in here. Also worth a case where results arrive in non-promptIdx order with tied metric values, to pin the tie-break's determinism (the only ordering where it matters).

(The two near-identical cost/latency tests above could fold into one `it.each` table, which would also make the cost-vs-latency `onlyPassing` asymmetry explicit.)

site/docs/configuration/expected-outputs/deterministic.md
```diff
@@ -300,6 +302,23 @@ assert:
...
+```yaml
+assert:
+  - type: contains
+    value: '42'
+  - type: select-lowest-cost
+    value:
+      onlyPassing: true
+```
+
+With `onlyPassing`, the selector fails all outputs if none pass the other assertions. Run with `promptfoo eval --no-cache`; cached outputs and missing or invalid costs fail the comparison. Ties are resolved by prompt order.
```
mldangelo: **Docs — three accuracy gaps and missing interaction docs:**

1. "cached outputs and missing or invalid costs fail the comparison" under-states the blast radius — if *any* eligible output is cached or lacks a valid metric, **every output on the test fails** (`failAll`), not just the offending one. Users should also see the exact remediation string ("Rerun the eval with --no-cache").
2. The latency section never says `onlyPassing` is *forced* — unlike cost, `select-lowest-latency` cannot consider failing outputs, and `value: {onlyPassing: false}` is silently ignored. A reader coming from this section will assume the knob works there too.
3. "provider-reported cost data" — link the adjacent [Cost](#cost) section, which names which providers actually report cost; pointing this at a provider that doesn't populate `response.cost` hits the fail-all path with no forewarning.

Missing entirely: processing order (metric selectors → select-best → max-score), the AND semantics of combining comparison assertions (divergent winners fail every output — reproduced), that max-score excludes all `select-*` results from aggregation, and cross-links to/from `select-best`/`max-score` (their pages don't mention these new siblings). Minor: "ties are resolved by prompt order" → "ties go to the earliest prompt (lowest prompt index)".

src/matchers/comparison.ts
```diff
@@ -13,6 +13,139 @@ import type {
...
+  const configuredOnlyPassing =
+    assertion.value !== null &&
+    typeof assertion.value === 'object' &&
+    'onlyPassing' in assertion.value &&
+    typeof assertion.value.onlyPassing === 'boolean'
+      ? assertion.value.onlyPassing
+      : undefined;
+  const onlyPassing = configuredOnlyPassing ?? definition.defaultOnlyPassing;
+  const eligible = results.flatMap((result, index) =>
+    !result.error && result.response && (!onlyPassing || result.success) ? [index] : [],
```
mldangelo-oai: **P2 — `onlyPassing: false` still excludes real assertion-failed outputs.** `applyGradingResult()` stores every ordinary assertion failure reason in `row.error` while keeping the valid provider response and metric. This predicate therefore drops those rows before `onlyPassing` is considered, and the later result mapping incorrectly says the provider produced no output. I reproduced it with the local CLI: a cheap output (`cost: 0.001`) that failed `contains` was marked ineligible, while a passing output costing `0.01` was selected by default `select-lowest-cost`, even though that selector documents `onlyPassing: false` as its default. The same bug affects latency with explicit `{ onlyPassing: false }`. Please distinguish provider/runtime failures via `failureReason === ERROR` (or equivalent structured state) and apply `success` filtering only when `onlyPassing` is true; add an evaluator-level regression using the real failed-assertion row shape.

src/evaluator.ts
```diff
@@ -4576,7 +4695,12 @@ class Evaluator<TEvaluation extends EvaluationRecord, TResult extends Evaluation
       testSuite,
       tests,
     });
-    markComparisonRows(runEvalOptions, rowsWithSelectBestAssertion, rowsWithMaxScoreAssertion);
+    markComparisonRows(
```
mldangelo-oai, line 4750: **P2 — A completed `--resume` reapplies and duplicates metric-selector grading.** Comparison rows are recorded here before `filterCompletedResumeSteps()` removes every already-finished provider step. The comparison phase then reloads the finalized rows, `mergeDeterministicComparisonGradingResult()` appends another selector component, and `updateComparisonStats()` increments selector pass/fail counts again. A stable rerun usually has a zero score delta, but component results and assertion counts grow on every resume; prior selector failures can also contaminate later `onlyPassing` eligibility. Existing comparison architecture intentionally supports finishing deferred comparisons on resume, so simply moving discovery after the filter would be unsafe. Please make selector finalization idempotent: skip fully finalized groups and replace, rather than append, matching selector components for partial groups. Add a persisted evaluate/resume regression that keeps one selector component and unchanged assertion counts.

src/types/index.ts
```diff
@@ -687,57 +698,89 @@ export const AssertionTypeSchema = z.union([
...
+    // Process the output before running the assertion
+    transform: StringOrFunctionSchema.optional(),
 
-  // Extract context from the output using a transform
-  contextTransform: StringOrFunctionSchema.optional(),
-});
+    // Extract context from the output using a transform
+    contextTransform: StringOrFunctionSchema.optional(),
+  })
+  .superRefine((assertion, ctx) => {
```
mldangelo-oai, line 761: **P2 — This refinement breaks composition APIs on an exported public schema.** With the repository's Zod 4.4.3, `AssertionSchema` remains a `ZodObject`, but adding `superRefine` makes both `.pick()` and `.omit()` throw at runtime: `cannot be used on object schemas containing refinements` (I reproduced the same failure on `AssertionSetSchema`). Those methods worked on the refinement-free schemas before this PR, remain present and typed, and both schemas are re-exported from the package root, so existing library consumers can compile successfully and then crash after upgrading. Keep the long-standing exported object schemas refinement-free and enforce these selector semantics through a separate strict/internal schema or validation helper at config/API boundaries. Add public-contract tests for `shape`, `extend`, `pick`, and `omit` alongside strict selector-validation tests.

src/matchers/comparison.ts
```diff
@@ -13,6 +13,139 @@ import type {
...
+      return failAll(
+        `${assertion.type} requires at least one eligible output; all outputs failed other assertions`,
+      );
+    }
+    return failAll(
+      `${assertion.type} requires at least one eligible output; no output produced a usable metric`,
+    );
+  }
+
+  const cached = eligible.filter((index) => results[index].response?.cached);
```
mldangelo-oai, line 116: **P2 — Unbounded index diagnostics become quadratic persisted output.** Both cached and invalid-metric branches join every affected prompt index into one O(N)-length reason, then `failAll()` attaches that reason to all N rows and the evaluator serializes each row independently. With 10,000 cached candidates, I measured a 58,998-character reason copied across 10,000 results—about 590 MB of reason text before JSON/database overhead. A cached rerun of a large matrix can therefore spend most of its time and disk on this error path. Please cap the displayed indexes (for example, first 20 plus `and N more`) by slicing before `map().join()`, and add a large synthetic test that asserts bounded reason length while preserving the total affected count.

mldangelo-oai (commented):
Reviewed exact head `dde3557a7d930d4cd09c90aeb4db5f0933d12871` in comment-only mode; I did not modify the branch. I confirmed two existing current-head P2s and did not duplicate them:

- Eval Creator retains `value: ''` when switching to a metric selector, while the backend schema rejects that payload: https://github.com/promptfoo/promptfoo/pull/9884#discussion_r3522302757
- A failed deterministic comparison can preserve an already-failing row's partial score, inflating max-score loser and prompt metrics: https://github.com/promptfoo/promptfoo/pull/9884#discussion_r3522302758

I found four additional material issues below. Validation at this head: 463/463 focused backend tests and 191/191 focused Eval Creator tests passed locally; `npm run tsc`, `npm run build`, and `git diff --check` also passed. A secret-free, no-share local CLI eval reproduced the `onlyPassing: false` defect. I also reproduced the exported-schema composition failure and measured the unbounded diagnostic path. All 13 required GitHub checks were green when I posted this review. The branch is one non-overlapping `package-lock.json` commit behind `main`, so it should be updated from `main` and the strict checks rerun after the fixes.

src/evaluator.ts
```diff
@@ -1906,15 +1906,27 @@ function mergeSelectBestGradingResult(
...
-function mergeMaxScoreGradingResult(result: EvaluationStoreResult, gradingResult: GradingResult) {
-  const existingComponentResults = result.gradingResult?.componentResults || [];
+function mergeDeterministicComparisonGradingResult(
+  result: EvaluationStoreResult,
+  gradingResult: GradingResult,
+): boolean {
+  const assertionType = gradingResult.assertion?.type;
+  const allComponentResults = result.gradingResult?.componentResults || [];
+  const alreadyFinalized = allComponentResults.some(
+    (component) => component.assertion?.type === assertionType,
```
mldangelo-oai: [P2] Key finalized selectors by assertion instance, not only by type

The docs allow multiple metric selectors with AND semantics, and two selectors of the same type can be meaningful when their `onlyPassing` settings differ. On the initial run, however, the second `select-lowest-cost` already finds the first component here, is treated as resumed/finalized, removes the first component at line 1918, and makes the caller skip `finalizeComparisonGrading`. I reproduced this with a cheap assertion-failed output and an expensive passing output plus `select-lowest-cost` configured once with `onlyPassing: false` and once with `true`: every row retained only one selector component, and prompt metrics were `{pass: 1, fail: 1}` for both rows instead of `{pass: 1, fail: 2}` / `{pass: 2, fail: 1}`. Use a stable assertion-instance key (for example test index + assertion ordinal, or an ordinal-aware fingerprint) so only the matching persisted component is replaced on resume while every configured selector is finalized on its first run, and add this duplicate-type regression.

mldangelo-oai (commented):
Re-reviewed exact head `5743acf6b4ff8497f370717eeb6d3a6e811159d0` after it changed during the sweep. The new main merge and follow-up commits address the six issues from my prior review: Eval Creator now removes stale selector values; deterministic failures reset stale partial scores; `onlyPassing: false` uses structured failure reasons; completed resumes replace selector grading instead of appending it; exported object schemas remain composable while runtime validation stays on `AssertionOrSetSchema`; and large invalid/cached diagnostics are bounded. I found one new P2 in the resume-idempotency implementation, noted inline: selector identity is reduced to assertion type, so a second configured selector of the same type is mistaken for resumed grading during the initial run.

Verification on the new exact head: TypeScript passes; 411 focused evaluator, matcher, and schema tests pass; `git diff --check` passes; the head contains current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`); and all 43 published checks are green. A focused two-selector regression reproduces the new finding: with two `select-lowest-cost` assertions using different `onlyPassing` settings, each row retains one selector component instead of two and prompt assertion counters omit the second selector. No contributor branch changes or pushes were made during this follow-up.

PR #9896, opened 2026-06-27
test(util): cover backslash escapes in convertSlashCommentsToHash

### LINKED ISSUE #9895: bug(util): convertSlashCommentsToHash fails to detect closing quote after even-count backslash sequence (e.g. "C:\\\\")
## Describe the bug

`convertSlashCommentsToHash` in `src/util/json.ts` uses `prevChar !== '\\"'` to decide whether a `"` closes a double-quoted string. This single-character look-behind is insufficient: when the `"` is preceded by an even number of backslashes (e.g. `"\\\\"` — a valid JSON string containing one backslash), the closing quote is **not** escaped, but the code treats it as escaped and stays inside the string. Any `//` that follows is never converted to `#`, breaking the comment-stripping used by `extractJsonObjects`.

## Minimal repro

```js
import { convertSlashCommentsToHash } from 'src/util/json';

// JSON string whose value ends with a backslash (e.g. a Windows path)
const input = '{ "path": "C:\\\\\\\\" } // comment';
// raw characters:  { "path": "C:\\" } // comment

console.log(convertSlashCommentsToHash(input));
// actual:   { "path": "C:\\" } // comment   ← // NOT converted
// expected: { "path": "C:\\" } # comment   ← // should be #
```

## Root cause

```ts
// src/util/json.ts – doubleQuote state
case 'doubleQuote':
  result += char;
  if (char === '"' && prevChar !== '\\') {   // ← only checks one char back
    state = 'normal';
  }
  break;
```

When the raw text is `"\\"`:
- first `\` → not a quote, skipped
- second `\` → not a quote, skipped  
- `"` → `prevChar === '\\'` → condition is false → string **not** closed ← **bug**

## Expected behaviour

The closing-quote detection should count the consecutive backslashes immediately before the `"`. If the count is **even** (0, 2, 4…) the quote ends the string; if **odd** it is escaped and the string continues.

## Impact

`extractJsonObjects` (called by assertion matchers and the evaluator) silently fails to extract JSON objects from strings that contain values ending with a backslash — a common pattern for Windows file paths, regex escape sequences, etc.

## Version

Reproduces on `main` (promptfoo ≥ 0.121.x).

JSap0914:
## Bug

`convertSlashCommentsToHash` (used by `extractJsonObjects` to strip `//` comments before YAML/JSON parsing) uses a single-character look-behind — `prevChar !== '\\'` — to determine whether a `"` closes a double-quoted string. This is incorrect when the `"` is preceded by an **even** number of backslashes.

**Example (Windows path with trailing backslash):**
```
{ "path": "C:\\" } // comment
```
Raw characters inside the JSON value are `C:\\` (two backslashes). The closing `"` after `\\` is **not** escaped, but `prevChar === '\\'` caused the function to stay inside the doubleQuote state — so the `//` was never converted to `#` and reached `js-yaml` verbatim.

## Root cause

```ts
// before
case 'doubleQuote':
  result += char;
  if (char === '"' && prevChar !== '\\') { // ← single-char look-behind, wrong for \\"
    state = 'normal';
  }
  break;
```

## Fix

Count the consecutive backslashes immediately before the `"`. An **even** count (0, 2, 4, …) means the quote is unescaped and closes the string; an **odd** count means it is escaped and the string continues.

```ts
// after
case 'doubleQuote':
  result += char;
  if (char === '"') {
    let numBackslashes = 0;
    let k = i - 1;
    while (k >= 0 && line[k] === '\\') { numBackslashes++; k--; }
    if (numBackslashes % 2 === 0) { state = 'normal'; }
  }
  break;
```

## Verification

```
npx vitest run test/util/json.test.ts
Tests  79 passed (79)
```

Three new regression tests were added (two for the even-backslash bug, one confirming existing odd-backslash behaviour is preserved). All 79 tests in `test/util/json.test.ts` pass; no existing test was modified.

> AI-assisted contribution.

Closes #9895

---

**Updated after merging `main` (scope changed):** the `src/util/json.ts` change described above landed
independently on main in #10030 (`fd607d1f00`), and #10080 (`01694e786e`) extended the same even/odd
backslash count to single-quoted strings and factored it into an `isEscapedQuote()` helper. The
production fix is therefore no longer part of this diff — the "Root cause" and "Fix" snippets above
describe the pre-merge state, not the current one. Issue #9895 is already closed.

What remains is **23 lines of test-only regression coverage** in `test/util/json.test.ts`:

- `should convert // after a string ending with an escaped backslash` — duplicates the case main
  already covers in #10030 (`{ "path": "C:\\" } // comment`).
- `should keep // inside a string whose value ends with an escaped backslash` — new: a double-quoted
  value with two even-backslash runs (`"C:\\Users\\" // comment`).
- `should not close string on triple-backslash-quote (odd count = escaped)` — new: an odd-count
  (escaped) quote inside a **double**-quoted string. Main only covers the odd-count case for
  single-quoted strings.

All three pass against main's `isEscapedQuote()` implementation (`test/util/json.test.ts`: 82 passed,
up from the 79 quoted above). Title updated `fix(util)` -> `test(util)` to match the test-only diff.

### DIFF
--- a/src/util/json.ts
+++ b/src/util/json.ts
@@ -190,8 +190,19 @@ export function convertSlashCommentsToHash(str: string): string {
 
           case 'doubleQuote':
             result += char;
-            if (char === '"' && prevChar !== '\\') {
-              state = 'normal';
+            if (char === '"') {
+              // Count consecutive backslashes immediately before this quote.
+              // An even count (0, 2, 4, …) means the quote is not escaped and
+              // closes the string; an odd count means the quote is escaped.
+              let numBackslashes = 0;
+              let k = i - 1;
+              while (k >= 0 && line[k] === '\\') {
+                numBackslashes++;
+                k--;
+              }
+              if (numBackslashes % 2 === 0) {
+                state = 'normal';
+              }
             }
             break;
         }
--- a/test/util/json.test.ts
+++ b/test/util/json.test.ts
@@ -453,6 +453,29 @@ describe('json utilities', () => {
         const expected = 'url: http://example.com/path # trailing comment';
         expect(convertSlashCommentsToHash(input)).toBe(expected);
       });
+
+      it('should convert // after a string ending with an escaped backslash', () => {
+        // "C:\\" in raw text: the value ends with one backslash (even-count \\).
+        // The closing " is NOT escaped, so the subsequent // IS a comment.
+        const input = '{ "path": "C:\\\\" } // Windows path comment';
+        const expected = '{ "path": "C:\\\\" } # Windows path comment';
+        expect(convertSlashCommentsToHash(input)).toBe(expected);
+      });
+
+      it('should keep // inside a string whose value ends with an escaped backslash', () => {
+        // Value is "C:\\Users\\" — ends with \\, closing " is not escaped.
+        const input = '"C:\\\\Users\\\\" // comment';
+        const expected = '"C:\\\\Users\\\\" # comment';
+        expect(convertSlashCommentsToHash(input)).toBe(expected);
+      });
+
+      it('should not close string on triple-backslash-quote (odd count = escaped)', () => {
+        // "\\\\" is three backslashes + closing quote: two escaped backslashes,
+        // then an escaped quote -> string not closed at that quote.
+        const input = '"text\\\\\\" still inside" // comment';
+        const expected = '"text\\\\\\" still inside" # comment';
+        expect(convertSlashCommentsToHash(input)).toBe(expected);
+      });
     });
 
     it('should skip non-object YAML values', () => {


mldangelo-oai:
## Codex review closeout — NEEDS_HUMAN_DECISION

Reviewed exact head [`80262195ab7cd33f4f90ecc625dbb05d01417b82`](https://github.com/promptfoo/promptfoo/commit/80262195ab7cd33f4f90ecc625dbb05d01417b82).

The even/odd backslash correction in [`convertSlashCommentsToHash`](https://github.com/promptfoo/promptfoo/blob/80262195ab7cd33f4f90ecc625dbb05d01417b82/src/util/json.ts#L191-L205) is correct and necessary. The complete JSON utility suite passes, caller suites pass, type/architecture checks pass, and source-level parity probes for 0–8 backslashes preserve strings, URLs, Windows paths, and comments as expected.

One behavior boundary needs an explicit maintainer decision before repair:

- Returning to `normal` after the corrected quote now exposes the existing [`//` classifier](https://github.com/promptfoo/promptfoo/blob/80262195ab7cd33f4f90ecc625dbb05d01417b82/src/util/json.ts#L155-L165). Inputs previously accepted after an even-backslash-terminated field can now lose an unquoted protocol-relative URL (`//cdn.example/...`) or silently change `/api//v1` to `/api#v1`. The underlying ambiguity is pre-existing, but the observable regression for these exact YAML-like inputs is introduced by this transition.
- The same transition adds a new route to the classifier's pre-existing quadratic URL-prefix scan. With an attacker/model-controlled long slash run, validation measured roughly 0.56 s at 8k slashes, 2.3 s at 16k, and 9 s at 32k. A simpler unquoted-URL input already reaches the same worst case on `main`; this PR adds a new trigger family rather than a new asymptotic capability.

Recommendation: keep the parity fix, but decide whether YAML plain scalars containing embedded `//` are supported. I recommend preserving protocol-relative URLs and embedded double-slash paths, making URL/token recognition incremental (linear), and adding exact `extractJsonObjects` regressions before merge. If those unquoted forms are intentionally unsupported, document that boundary and explicitly accept or separately track the pre-existing performance sink.

Validated non-blocking cleanup remains after that decision:

- The PR body's comment-after-`}` example never reaches `js-yaml`; an inside-object comment is the production-reachable reproduction.
- Two new test descriptions misstate where `//` occurs and how many runtime/source backslashes are present.
- Sharing the existing escape-parity helper would reduce net production LOC, but is optional scope expansion.

Verification on this head: `test/util/json.test.ts` 79/79; contains-json 11 selected; LLM-rubric 3 selected; synthesis 8/8; `npm run tsc`; Biome check (passes with two existing complexity warnings); architecture check; exact-head CI 32 success / 2 expected skipped / 0 failed or pending. Review threads: 0 total / 0 unresolved.

No code, commit, PR metadata, or review-thread state was changed.

PR #9897, opened 2026-06-27
fix(providers): guard array access in Bedrock Titan and Cohere output handlers

he-yufeng:
The Bedrock `TITAN_TEXT` and `COHERE_COMMAND` output handlers index an array immediately after an optional chain that only covers the top-level object:

```ts
// TITAN_TEXT
output: (_config, responseJson) => responseJson?.results[0]?.outputText,
// COHERE_COMMAND
output: (_config, responseJson) => responseJson?.generations[0]?.text,
```

`responseJson?.results[0]` guards `responseJson` but not `results`. When the response is present but the field is missing or `undefined` (an error body, an empty or truncated payload), `results`/`generations` is `undefined`, and `undefined[0]` throws:

```
TypeError: Cannot read properties of undefined (reading '0')
```

This is reachable in practice because `model.output(...)` runs on two paths. The fresh-request path is wrapped in a try/catch, but the cache-hit branch calls the handler before that try block, so a cached response missing the expected field surfaces as an uncaught crash rather than a handled empty result.

Every other handler in this file already uses the safe form (`responseJson.choices?.[0]?.message?.content`); these two were the outliers. The fix is the missing `?.`:

```ts
responseJson?.results?.[0]?.outputText
responseJson?.generations?.[0]?.text
```

There was no output-handler coverage for these two models, so I added some. The "missing field returns `undefined` instead of throwing" cases fail on `main` with the `TypeError` above and pass with the fix; valid responses still return the text. `test/providers/bedrock/index.test.ts` is green (329 passing).

### DIFF
--- a/src/providers/bedrock/index.ts
+++ b/src/providers/bedrock/index.ts
@@ -1779,7 +1779,7 @@ export const BEDROCK_MODEL = {
       );
       return { inputText: prompt, textGenerationConfig };
     },
-    output: (_config: BedrockOptions, responseJson: any) => responseJson?.results[0]?.outputText,
+    output: (_config: BedrockOptions, responseJson: any) => responseJson?.results?.[0]?.outputText,
     tokenUsage: getOpenAiCompatibleTokenUsage,
   },
   LLAMA2: getLlamaModelHandler(LlamaVersion.V2),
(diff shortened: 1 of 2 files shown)


mldangelo:
Nice catch — a Bedrock 200 with an empty `results`/`generations` array turning into an uncaught TypeError is exactly the kind of edge case that's painful to debug from the eval side, and your analysis of the cache-hit-outside-try/catch path was spot on. The two-character fix matching the optional-chaining pattern the rest of the handlers already use, plus regression tests for both handlers, made this an easy review. Thanks for the contribution! 🎉

PR #9899, opened 2026-06-28
fix(init): promptfoo init will print undefined when initializing and selecting redteam as action

ryandens:
## Problem

Running `promptfoo init` and selecting "Run a red team evaluation" prints broken next
steps that reference a directory literally named `undefined`:

```
✅ Wrote promptfooconfig.yaml to ./undefined

  1. Run cd undefined
  ...
```

There is no `undefined` directory, and the suggested `cd undefined` fails.

## Cause

`initializeProject` decides which next-steps message to print based on the
`outDirectory` value returned by `createDummyFiles`. The redteam branch was the only
return path that forgot to include `outDirectory`, so it came back as `undefined`. That
skipped the normal "current directory" message and fell through to the branch that
interpolates the directory name into the path and `cd` command.

## Impact

Exisitng message implied to me something may be broken, so i went digging. It
does not seem to have any impact on subsequent `promptfoo redteam run` commands,
but the message is confusing for users and can be easily fixed.

## Fix

Include the already-computed `outDirectory` in the redteam branch's return value, like
every other path does. For a plain `promptfoo init` it resolves to `'.'`, so the flow
now prints the correct "Setup complete!" next steps instead of `./undefined`. The bug
was purely a dropped field, so restoring it is the whole fix.

To prevent the same class of bug, `createDummyFiles` now has an explicit
`InitProjectResult` return type with `outDirectory` required. Previously the function's
return type was inferred as a union of its two branches, so a branch missing
`outDirectory` simply widened the consumer's value to `string | undefined` — which is
valid in the template-literal output and the `=== '.'` check, so strict typechecking
never flagged it. With the annotation, any return path that omits `outDirectory` now
fails to compile (`TS2741: Property 'outDirectory' is missing ... but required`).

## Tests

Added a regression test that runs the redteam init path and asserts no next-steps
message contains `undefined`. It fails before the change and passes after. `tsc` and the
full `test/onboarding.test.ts` suite pass; confirmed the new return type makes the
missing-field case a compile error.

## Repro (before fix)

```
promptfoo init

Created red teaming configuration file at promptfooconfig.yaml


To generate test cases and run your red team, use the command:

    promptfoo redteam run

✅ Wrote promptfooconfig.yaml to ./undefined

  1. Run cd undefined
  2. Run promptfoo eval to evaluate your prompts
  3. Run promptfoo view to view results in your browser

  Docs: https://promptfoo.dev/docs/configuration/guide
```

## After fix

```
promptfoo init

Created red teaming configuration file at promptfooconfig.yaml


To generate test cases and run your red team, use the command:

    promptfoo redteam run

✅ Setup complete! Next steps:

  1. Run promptfoo eval to evaluate your prompts
  2. Run promptfoo view to view results in your browser
```

### DIFF
--- a/src/onboarding.ts
+++ b/src/onboarding.ts
@@ -341,7 +341,18 @@ async function askForPermissionToOverwrite({
   return hasPermissionToWrite;
 }
 
-export async function createDummyFiles(directory: string | null, interactive: boolean = true) {
+interface InitProjectResult {
+  numPrompts: number;
+  providerPrefixes: string[];
+  action: string;
+  language: string;
+  outDirectory: string;
+}
+
+export async function createDummyFiles(
+  directory: string | null,
+  interactive: boolean = true,
+): Promise<InitProjectResult> {
   const outDirectory = directory || '.';
   const outDirAbsolute = path.join(process.cwd(), outDirectory);
 
@@ -436,6 +447,7 @@ export async function createDummyFiles(directory: string | null, interactive: bo
         providerPrefixes: [],
         action: 'redteam',
         language: 'not_applicable',
+        outDirectory,
       };
     }
 
--- a/test/onboarding.test.ts
+++ b/test/onboarding.test.ts
@@ -308,15 +308,18 @@ describe('initializeProject', () => {
 
     const infoSpy = vi.spyOn(logger, 'info').mockImplementation(() => logger);
 
-    await initializeProject(null, true);
+    try {
+      await initializeProject(null, true);
 
-    const messages = infoSpy.mock.calls.map((call) => String(call[0]));
+      const messages = infoSpy.mock.calls.map((call) => String(call[0]));
 
-    // Regression: the redteam branch dropped `outDirectory`, so the "Next steps"
-    // output interpolated `undefined` into the path and `cd` command.
-    expect(messages.some((msg) => msg.includes('undefined'))).toBe(false);
-    expect(messages.some((msg) => msg.includes('cd undefined'))).toBe(false);
-
-    infoSpy.mockRestore();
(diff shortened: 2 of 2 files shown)


src/onboarding.ts
```diff
@@ -341,7 +341,18 @@ async function askForPermissionToOverwrite({
   return hasPermissionToWrite;
 }
 
-export async function createDummyFiles(directory: string | null, interactive: boolean = true) {
+interface InitProjectResult {
```
ryandens, line 344: adding this new type is not specifically necessary for this fix, but i thought it was appropriate to help avoid an issue like this with this function in the future! happy to adjust based on feedback

test/onboarding.test.ts
```diff
@@ -300,4 +302,26 @@ describe('initializeProject', () => {
...
+
+    await initializeProject(null, true);
+
+    const messages = infoSpy.mock.calls.map((call) => String(call[0]));
+
+    // Regression: the redteam branch dropped `outDirectory`, so the "Next steps"
+    // output interpolated `undefined` instead of using the current-directory flow.
+    expect(messages).toEqual(
+      expect.arrayContaining([
+        expect.stringContaining('Setup complete! Next steps:'),
```
ryandens, line 320: Thanks for the test improvements! Agreed that these are much better

mldangelo-oai (approved):
LGTM — thank you for catching this and putting together such a thoughtful fix, Ryan. Returning `outDirectory` fixes the confusing output, and the explicit return type plus regression test make the same mistake harder to reintroduce. The full CI matrix and an interactive redteam init check are green. Really appreciate the clear write-up and the care you put into your first contribution to Promptfoo!

PR #9908, opened 2026-06-29
fix(providers): clamp DeepSeek cached tokens in cost calc

anxkhn:
### What

`calculateDeepSeekCost` could report an understated (possibly negative) cost when a response's cached prompt tokens exceed its total prompt tokens. At `src/providers/deepseek.ts`:

```ts
const uncachedPromptTokens = cachedTokens ? promptTokens - cachedTokens : promptTokens;
```

If `cachedTokens > promptTokens` (provider rounding, or `cached == prompt` edges), `uncachedPromptTokens` goes negative and `inputCost * uncachedPromptTokens` turns negative, so the reported cost is wrong. A negative `cachedTokens` would do the same in reverse. This clamps cached tokens to `[0, promptTokens]` before subtracting, and bills cache reads on the clamped count:

```ts
const billableCachedTokens = Math.min(Math.max(cachedTokens ?? 0, 0), promptTokens);
const uncachedPromptTokens = promptTokens - billableCachedTokens;
```

### Why

This is the same defect already fixed for the sibling OpenAI-compatible providers, which all clamp the same way:

- `src/providers/minimax.ts` and `src/providers/moonshot.ts`: `Math.min(Math.max(cachedTokens!, 0), promptTokens!)`
- `src/providers/xai/chat.ts` (#9783, merged): `Math.min(Math.max(cachedTokens!, 0), promptTokens)`

DeepSeek was the only one of these never updated. The early guard already returns `undefined` unless `promptTokens` is truthy, so the clamp's upper bound is a positive number.

### Tests

Added two regression tests to `test/providers/deepseek.test.ts` (mirroring the existing all-cached test): cached > prompt is capped at the all-cached price (0.2828, never negative); negative cached is treated as no cache hits (0.42). `npx vitest run test/providers/deepseek.test.ts` -> 19/19. Sibling moonshot/minimax tests unaffected. `npm run l && npm run f` clean, `npm run tsc` clean.


mldangelo (approved):
Thanks @anxkhn — this is a model bug report and fix: a real negative-cost defect, precisely scoped, with the sibling-provider precedent cited and regression tests that mirror the existing suite. Much appreciated.

While landing it I pushed two maintainer commits to your branch: a merge of latest `main`, and a small follow-up that extracts the clamp you referenced into a shared `clampCachedTokens` helper now used by deepseek, minimax, moonshot, and xai — which also picks up the siblings' `Number.isFinite` guard so a non-finite `cachedTokens` bills as zero cache hits rather than propagating NaN. Your tests all carried over as-is. Thanks again for the contribution!

PR #9911, opened 2026-06-30
fix(assertions): default context-recall and context-relevance threshold to 0.5

### LINKED ISSUE #9910: bug(assertions): context-recall and context-relevance default threshold to 0, silently always pass
## Describe the bug

```ts
// src/matchers/rag.ts
```

```yaml
assert:
  - type: context-recall
    value: 'Expected fact'
```

or

```yaml
assert:
  - type: context-relevance
```

## Root cause

```ts
  context,
  renderedValue,
  ...
);
```

```ts
  test.vars.query,
  context,
  ...
);
```

## Expected behaviour

## Impact

## Version

JSap0914:
## Bug

Both `context-recall` and `context-relevance` assertion handlers pass
`(assertion.threshold as number) ?? 0` to their matcher:

```ts
  context,
  renderedValue,
  ...
);

  test.vars.query,
  context,
  ...
);
```

```ts
```

With `threshold = 0`, `pass = score >= -EPSILON` is true for every score in `[0, 1]`. So:

```yaml
assert:
  - type: context-recall
    value: 'Expected fact'
  - type: context-relevance
```

**can never fail** when no `threshold` is given — they are no-op graders.

## Fix

Default both thresholds to **0.5**, consistent with the sibling fixes in #9849 (answer-relevance → 0.7) and #9907 (context-faithfulness → 0.7).

## Files changed

1. `src/assertions/contextRecall.ts` — `?? 0` → `?? 0.5`
2. `src/assertions/contextRelevance.ts` — `?? 0` → `?? 0.5`
3. `test/assertions/contextRecall.test.ts` — rename + update existing default-threshold tests; add new test verifying the 0.5 default
4. `test/assertions/contextRelevance.test.ts` — rename + update existing default-threshold tests
5. `site/docs/.../context-recall.md` — `(default: 0)` → `(default: 0.5)`
6. `site/docs/.../context-relevance.md` — `(default: 0)` → `(default: 0.5)`

## Verification

```
npx vitest run test/assertions/contextRecall.test.ts test/assertions/contextRelevance.test.ts
Test Files  2 passed (2)
Tests  14 passed (14)   ← was 13 before (new test added)
```

All 59 assertion tests pass.

Closes #9910

### DIFF
--- a/site/docs/configuration/expected-outputs/model-graded/context-recall.md
+++ b/site/docs/configuration/expected-outputs/model-graded/context-recall.md
@@ -32,7 +32,7 @@ assert:
 
 - `value` - Expected answer/ground truth
 - `context` - Retrieved text (in vars or via `contextTransform`)
-- `threshold` - Minimum score 0-1 (default: 0)
+- `threshold` - Minimum score 0-1 (default: 0.5)
 
 ### Full example
 
--- a/site/docs/configuration/expected-outputs/model-graded/context-relevance.md
+++ b/site/docs/configuration/expected-outputs/model-graded/context-relevance.md
@@ -35,7 +35,7 @@ assert:
 
 - `query` - User's question (in test vars)
 - `context` - Retrieved text (in vars or via `contextTransform`)
-- `threshold` - Minimum score 0-1 (default: 0)
+- `threshold` - Minimum score 0-1 (default: 0.5)
 
 ### Full example
 
--- a/src/assertions/contextRecall.ts
+++ b/src/assertions/contextRecall.ts
@@ -35,7 +35,7 @@ export const handleContextRecall = async ({
   const result = await matchesContextRecall(
     context, // context parameter (used as {{context}} in prompt)
     renderedValue, // ground truth parameter (used as {{groundTruth}} in prompt)
-    (assertion.threshold as number) ?? 0,
+    (assertion.threshold as number) ?? 0.5,
     test.options,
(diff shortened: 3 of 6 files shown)


mldangelo:
Thanks for this, and apologies it sat so long.

It turned out that **five** of us independently found and fixed the same bug — this PR, #10202, #10494, #9907 and #10567 all made `context-recall` / `context-relevance` (and friends) stop defaulting `threshold` to `0`. That is a good signal the bug was real and worth fixing; it also meant we had to pick one to land.

We went with #10202 because it was the only one covering all four RAG assertions and sharing the default through a named constant (`DEFAULT_RAG_ASSERTION_THRESHOLD`) instead of an inline literal in each file. It has now merged, at **0.5** — matching the existing `DEFAULT_THRESHOLD = 0.5` precedent in `src/external/assertions/deepeval.ts`.

**You are credited as a co-author on that commit** (`4831b602d0`), so this shows up as your contribution too.

I checked whether anything here could still land on top, and after rebasing, the residual was a single test case that #10202 already covers with an equivalent one. So there is genuinely nothing left to merge — not a reflection on the work, just the cost of five people arriving at the same place at once.

Closing as merged-via-#10202. Thank you for catching it, and please do keep sending fixes.

mldangelo:
Closing as merged via #10202 (`4831b602d0`), where you are credited as a co-author. Thanks again — see the note above for the details.

PR #9916, opened 2026-06-30
feat(agent-rubric): render Nunjucks templates in grading provider config

### LINKED ISSUE #9915: agent-rubric: per-test-case working_dir for batch trace evaluation (template substitution / dynamic resolver)
## Summary

`agent-rubric` provider config (e.g. `working_dir`) does not interpolate `{{var}}` templates from test-case `vars`. Combined with the docs only showing single-case examples, this forces awkward workarounds when running `agent-rubric` over a batch of independent artifacts (e.g. one per production trace, one per PR, one per fixture).

Filing this after building an internal batch evaluation pipeline that grades 10–50 LLM agent traces per run, each with its own evidence directory.

## Steps to reproduce

```yaml
tests:
  - path: tests/eval-cases.jsonl   # each row has vars.trace_id and vars.evidence_dir

defaultTest:
  assert:
    - type: agent-rubric
      value: file://prompts/rubric.txt
      provider:
        id: anthropic:claude-agent-sdk
        config:
          working_dir: ./evidence/{{trace_id}}   # NOT interpolated
          sandbox_mode: read-only
          approval_policy: never
```

Result:
```
Working dir ./evidence/{{trace_id}} does not exist or isn't accessible:
ENOENT: no such file or directory, stat '.../evidence/{{trace_id}}'
```

`{{trace_id}}` is present in `vars` (it appears correctly in `prompts:` and in `assert.value` rendered text), but `provider.config.working_dir` strings are not run through the templating engine.

## Why it matters

The current docs example (`working_dir: ./sample-project`) implicitly assumes one fixed artifact for all cases. Real eval pipelines that want `agent-rubric`'s evidence-inspecting capability often have:

- Per-case repo / fixture / snapshot
- Per-case rendered output (e.g. screenshot, diff)
- Per-case captured trace (LLM tool calls, before/after state)

Without a way to bind `working_dir` to the test case, users currently have to:

1. Use a single shared parent dir and instruct the judge in the rubric text to `cd` into a subdirectory named after `{{trace_id}}` (works, but every judge sees siblings' evidence — leakage risk under model-driven exploration; also breaks with non-trivial `sandbox_mode` semantics).
2. Skip `agent-rubric` and fall back to `llm-rubric` with the entire evidence stuffed into the prompt — defeats the purpose of having a tool-using grader.
3. Pre-process each case into a single fixed dir and run cases serially with `concurrency: 1` while mutating the dir between cases — fragile.

## Proposed solutions

Listed in order of increasing scope, any of which would unblock the use case:

### A. Template-substitute provider config strings (smallest change)

Run `provider.config.working_dir` (and any other path-shaped string fields) through the same Nunjucks renderer that already handles `prompts:` and `assert.value`. `vars` are already in scope at assertion time.

### B. Allow `working_dir` to be a function / file reference

```yaml
working_dir: file://providers/resolve-working-dir.js
# resolver(vars, context) -> string
```

Mirrors how `provider.id` already supports `file://`.

### C. Per-case `evidence` field on test cases

A first-class field promptfoo recognizes:

```yaml
tests:
  - vars: { trace_id: abc }
    evidence_dir: ./evidence/abc
```

Auto-bind to `working_dir` when an `agent-rubric` is present.

## Cross-reference

- Docs page for `agent-rubric` (https://www.promptfoo.dev/docs/configuration/expected-outputs/model-graded/agent-rubric/) shows only single-`working_dir` examples and does not address per-case binding.
- `examples/compare-agentic-sdks` uses `working_dir: ./test-codebase` shared across cases.
- No existing issue I could find on this; happy to be pointed at one.

## Workaround we ended up using

`working_dir: ./evidence` (parent), and the rubric text says `请进入子目录 ./{{trace_id}}/ 后执行评估` (the `value` field IS templated). Works but it's a hack — judge can read sibling evidence dirs, and it leaks the convention into rubric text.

I'm happy to send a PR for option **A** if maintainers think that's the right path; it's the smallest change and matches user expectation (templating works elsewhere in the same config).

yanpgwang:
## Summary

`agent-rubric` now resolves `provider.config.working_dir: ./evidence/{{trace_id}}` for each expanded test case before constructing the grader. This works through the CLI and programmatic evaluation, including inline, YAML/JSON file-backed, and cloud-backed graders, `defaultTest`, provider type maps, maps keyed by provider ID, assertion sets, and scenarios.

Closes #9915.

- Render top-level config strings with Nunjucks expressions, blocks, and suite-defined filters. Interpolating an undefined variable fails before grader construction; explicit `default` filters still work.
- Keep templated definitions and provider-file references lazy until final case vars are available, preserving suite/provider environment precedence. Environment-templated file references use the same merged env as the loader; cloud config rendering waits for cloud and local env to merge. Only the selected text grader is rendered. Configs receive the actual output/rubric with the same precedence as the grading call.
- Reject residual Nunjucks syntax in rendered top-level strings so SDK runtimes cannot reinterpret test data on their own templating pass. Nested config keeps the provider's existing behavior; existing instances and other assertion types retain their behavior.
- Pass the final merged suite/provider/file/cloud env to provider factories so deferred SDK graders retain their credentials.
- Retain the loaded factory identity separately from custom provider IDs so agent-runtime validation accepts aliased graders.
- Correct the documentation's assertion placement and pair Codex-specific options with the Codex provider.
- Merge `main` at `783d71ff` and resolve the architecture baseline conflict. The intentional increments cover the matchers' renderer/provider-reference imports, the agent matcher's factory-identity lookup, and the new module's provider/config type dependencies. The renderer is separate from provider-reference utilities to preserve frontend build compatibility.

## Scope

References to top-level target providers continue to reuse their already-constructed instances. To vary a grader workspace per case, provide a separate raw grader definition on the assertion or in test/default grading options. Rebuilding shared target instances per case is not part of this change; this limit is now explicit in the docs.

## Workspace trust boundary

Workspace templates are configuration, not a filesystem access-control mechanism. Use trusted case IDs; the docs now explicitly warn against using unvalidated model output for filesystem paths and explain evidence-root validation (including symlinks) and provider sandbox permissions. The reserved output/rubric context preserves existing Codex call-time behavior. This change does not invent a global evidence root or reject existing absolute workspace paths; an opt-in root constraint would require a separate configuration contract.

## Validation

- Backend Vitest suite: **25,449 passed**, 10 existing skips.
- 22 focused config regression cases plus real Codex SDK pipeline tests cover per-case isolation, config locations, provider files, filters, missing variables, environment precedence, rejection of unsafe residual templates, and compatibility.
- Full `npm run build` (TypeScript, package bundles, frontend), Biome/Prettier, architecture boundaries, and Knip.
- Built CLI evals with two on-disk evidence directories and a deterministic local grader, including a YAML provider file, custom filter, and a provider-ID map with an env-templated file reference: 2 passed, 0 failed, 0 errors in each run. Exported JSON confirms each grader reads its own case directory. No live LLM calls.


yanpgwang:
Hi @ianw-oai, following up on this PR and #9915. Has per-test-case templating for `agent-rubric`'s `provider.config.working_dir` been addressed elsewhere, or is rendering the grader config against test vars an approach you'd support? I couldn't find a merged fix, and the relevant code on `main` still appears to pass the workspace path through without per-case rendering.

I see that this PR now has merge conflicts and outstanding automated review feedback, including the provider-resolution timing issue. Maintainer guidance on the preferred approach would help shape the next revision. Thanks!

yanpgwang:
@mldangelo-oai could you review the scope of this per-case agent grader workspace PR? It lets batch trace evaluations bind each grader to its case evidence directory before construction. I’ve resolved the earlier conflict and addressed the compatibility reviews covering CLI/API resolution, provider files/cloud configs, filters, custom IDs, and env/credentials. The local backend suite passes 25,449 tests and the full build passes. Latest head `c15014018` also has 32 successful CI checks, with 2 skipped.

Two review items remain: references to top-level target providers still reuse their existing instances, and the security scan requests enforced root confinement for model-derived filesystem paths. Path validation is currently documented but not enforced by a built-in evidence-root setting. The PR description and docs make both limits explicit. Should either broader behavior be included here, or handled separately?

PR #9925, opened 2026-07-01
docs: replace LLMs.txt specification link

eggplants:
Updated the LLMs.txt specification link to the correct URL.

I think the fictional `llmspec.ai` is the result of hallucinations.


mldangelo:
Thank you!

PR #9930, opened 2026-07-01
docs(site): fix stale source-code links and outdated references

orangeCatDeveloper:
## Summary

- update GitHub source links to their current line ranges (localai, replicate, and openai config options; the red team default severity map now lives in `src/redteam/constants/metadata.ts`)
- correct the provider env-var registration path to `src/contracts/env.ts` and refresh the cached-flag reference line numbers in the provider AGENTS guide
- clarify that `promptfoo eval` has no `--verbose` flag (`-v` is `--vars`); use `LOG_LEVEL=debug` for debug logs
- remove the nonexistent `PROMPTFOO_OUTPUT_DIR` env var from the MCP server example

## Validation

- `npx prettier --check` on the changed docs

### DIFF
--- a/site/docs/integrations/mcp-server.md
+++ b/site/docs/integrations/mcp-server.md
@@ -298,7 +298,6 @@ export ANTHROPIC_API_KEY=sk-ant-...
 
 # Configure promptfoo behavior
 export PROMPTFOO_CONFIG_DIR=/path/to/configs
-export PROMPTFOO_OUTPUT_DIR=/path/to/outputs
 
 # Start server with environment
 npx promptfoo@latest mcp --transport stdio
--- a/site/docs/providers/localai.md
+++ b/site/docs/providers/localai.md
@@ -23,7 +23,7 @@ The model name is typically the filename of the `.bin` file that you downloaded 
 
 ## Configuring parameters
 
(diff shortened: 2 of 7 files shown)


mldangelo:
Verified every link/ref in this PR against current `main`: localai L16, replicate L24, metadata.ts L504, pythonCompletion.ts:62, vertex.ts:371, and the `src/contracts/env.ts` re-export are all exact. `PROMPTFOO_OUTPUT_DIR` indeed doesn't exist anywhere in `src/`. Also confirmed these were the only 4 line-anchored GitHub links in all of `site/docs`, so coverage is complete.

Pushed one correction commit:

- `openai/types.ts` range updated to L131–L215 (main moved under you when GPT-5.6 preview landed after you branched)
- `converse.ts:1276` is the exact `cached: true` line
- **Restored the original `--verbose` note**: `promptfoo eval --verbose` does work — the root program registers `--verbose` (`addCommonOptionsRecursively` in `src/mainUtils.ts`) and commander parses program options after the subcommand, so the preAction hook enables debug logs. Verified live. Only the `-v` short form is taken by `--vars`, which is what the original note said.

mldangelo (approved):
All line refs verified exact against current main; --verbose note corrected back to accurate wording; CI green.

PR #9931, opened 2026-07-01
docs(redteam): replace deprecated prompt-injection strategy references

orangeCatDeveloper:
## Summary

- Replace the deprecated `prompt-injection` strategy with `jailbreak-templates` across the red team framework and guide docs. `prompt-injection` is a backward-compatibility alias that logs a deprecation warning (`src/redteam/strategies/index.ts`), and the framework mappings in `src/redteam/constants/frameworks.ts` already use `jailbreak-templates`.
- Move strategies that were miscategorized under `plugins:` into `strategies:` (testing-guardrails, openai-chatkit, authoritative-markup-injection, and the node-api examples/reference).
- Replace the nonexistent `pii:indirect` plugin with `pii:api-db`, and the nonexistent `iterative`/`ensemble` strategies with valid ones.
- Fix inline comments that described `jailbreak-templates` as prompt injection; it applies static jailbreak templates (DAN, etc.).

The `prompt-injection` alias still works, so existing configs are unaffected — this only updates the docs to the current, non-deprecated names.

## Validation

- Cross-checked every changed strategy and plugin id against `src/redteam/constants/strategies.ts` and `src/redteam/constants/plugins.ts`.
- `npx prettier --check` on the changed docs.


mldangelo:
Reviewed and verified every replaced ID against `src/redteam/constants/{strategies,plugins}.ts` and the framework mappings — all correct, and the rewritten rag.md example schema-validates against `RedteamConfigSchema`. Nice work.

I pushed a follow-up commit that completes the sweep and tightens a few of the touched examples:

- Remaining deprecated `prompt-injection` strategy references in `examples/*/promptfooconfig.yaml`, copyable blog YAML, the ci-cd guide, and the custom-api/python/ruby `strategyId` examples
- `llm-redteaming.md`: removed `jailbreak`/`prompt-injection` rows from the default-plugins table (they're strategies, not plugins — same class of error this PR fixes elsewhere)
- `rag.md`: restored a strategies list alongside the `indirect-prompt-injection` plugin (so the section's direct-injection example is still exercised) and documented the `indirectInjectionVar` prompt-variable prerequisite
- `mitre-atlas.md`: aligned the Initial Access example with the `mitre:atlas:initial-access` mapping in `frameworks.ts`
- `openai-chatkit.md`: added `ascii-smuggling`/`prompt-extraction` so the Prompt Injection section tests what its title claims
- `examples/redteam-tracing-example`: dropped a `prompt-injection` entry that was in the plugins list (invalid as a plugin)
- `lethal-trifecta-testing.md` blog: fixed invalid YAML and a wrong config key (`injectVar` → `indirectInjectionVar`)

Not addressed here (follow-up): `src/commands/mcp/tools/redteam{Generate,Run}.ts` present `prompt-injection` as a plugin id in their tool descriptions (would fail validation), and `riskScoring.ts` has no `jailbreak-templates` metadata entry. Docs site builds clean with `onBrokenLinks: throw`.

mldangelo (approved):
Verified all replaced IDs against source constants; completed the sweep across examples/blogs/docs; site builds clean; CI green.

PR #9932, opened 2026-07-01
fix(util): select xlsx sheet by name when name starts with digits

anxkhn:
## What

`file://test_cases.xlsx#SheetName` could select the wrong sheet when the sheet name starts with digits. For example, `#1H-results` was interpreted as index 1, `#2024-Q1` as index 2024, and `#2.9` as index 2.

The public `readTests` path also treated `.9` as the file extension in `file://test_cases.xlsx#2.9`, so a dotted sheet name could silently load zero tests instead of reaching the XLSX reader.

## Fix

- Treat a sheet selector as an index only when the entire trimmed value is a signed integer (`/^[+-]?\d+$/`). This preserves existing `#+2`, `#-1`, and `#0` behavior while routing digit-prefixed names through exact name lookup.
- Detect XLSX/XLS files from the basename before the sheet fragment, while preserving `#` characters in parent directories and non-Excel filenames.
- Keep the existing index range and missing-name errors unchanged.

## Tests

Added regression coverage for:

- digit-prefixed names: `2024-Q1`, `1H-results`, and `2.9`;
- signed numeric indexes, including `#+2`;
- the public `readTests(['file://test.xlsx#2.9'])` path;
- `#` characters in CSV parent directories and filenames.

Validation:

```text
npx vitest run test/util/xlsx.test.ts test/util/testCaseReader.test.ts --sequence.shuffle=false  # 109/109 pass
npx vitest run test/util/xlsx.test.ts test/util/testCaseReader.test.ts --sequence.shuffle --sequence.seed=9932  # 109/109 pass
npm run tsc  # pass
```

An end-to-end local CLI eval against a real five-sheet workbook also selected `1H-results`, `2024-Q1`, `2.9`, and index `2` correctly with `--no-cache` (4/4 tests passed).


mldangelo-oai (approved):
Thanks for catching this, Anas — approved! This fixes a subtle failure mode cleanly: digit-prefixed sheet names now work through the public file-loading path, including dotted names like `2.9`, without losing signed-index compatibility or breaking paths that contain `#`. The focused regressions and full CI matrix are green. Really appreciate the clear analysis and thoughtful first contribution.

PR #9935, opened 2026-07-02
fix(redteam): remote materialize mcp tool calls

jameshiester-oai:
## Summary

- Add shared MCP tool-call normalization and schema validation for red team materialization.
- Add a remote MCP materialization client that calls the Promptfoo task server when remote generation is enabled.
- Preserve local no-model validation and local JSON-provider fallback when remote generation is disabled.

## Test plan

- `../node_modules/.bin/biome check --config-path biome.jsonc --write src/providers/promptfoo.ts src/redteam/mcpMaterialization.ts src/redteam/mcpTargetProvider.ts src/redteam/mcpToolCall.ts test/providers/promptfoo.test.ts test/redteam/mcpTargetProvider.test.ts`
- `git diff --check`
- Targeted Vitest was attempted but blocked in this worktree because the interrupted submodule install left promptfoo dependencies incomplete (`dedent`, `chalk`, `@inquirer/input` missing).


zcrab-oai (approved):

src/redteam/mcpToolCall.ts
```diff
@@ -0,0 +1,96 @@
...
+    return false;
+  }
+
+  const tool = toolByName.get(toolCall.tool);
+  if (!tool) {
+    return false;
+  }
+
+  try {
+    return ajv.validate(tool.inputSchema ?? { type: 'object' }, toolCall.args) === true;
```
zcrab-oai, line 74: [P1] Bound the AJV schema cache. `ajv.validate(schemaObject, ...)` compiles and retains every fresh schema object in this process-global AJV instance. The cloud task reconstructs schema objects on every request, so even repeated equivalent requests grow the cache forever; in a local reproduction, 5,000 simple schemas left 5,001 cache entries and added about 73 MB of heap. A long-running scan or an unauthenticated caller can steadily exhaust the server. Please use a bounded/content-keyed validator cache, evict compiled schemas, or scope AJV lifetime so request-created schema objects are not retained indefinitely.

src/redteam/extraction/util.ts
```diff
@@ -71,6 +87,65 @@ export async function fetchRemoteGeneration(
...
+    return undefined;
+  }
+
+  const body = {
+    email: getUserEmail(),
+    jsonOnly: true,
+    mcpMaterializationContext: {
+      intentValue: options.intentValue,
+      purpose: options.purpose,
+      tools: options.tools,
```
zcrab-oai, line 108: [P2] Document the new hosted transfer of MCP metadata. Whenever `shouldGenerateRemote()` selects the hosted path, this sends the full tool catalog, descriptions, input schemas, application purpose, and attack intent to the remote service. Internal MCP schemas can reveal private capabilities and parameter structure, while the current red-team data-handling page says target evaluation is local and does not list tool definitions among hosted-generation data. Please update the data-handling/MCP docs and make the remote condition and disable/opt-in behavior clear before shipping.

zcrab-oai (commented):
Deep review found two additional issues not already covered by the existing threads.

zcrab-oai:
Post-merge follow-up from the requested deep review of `d5a218691b74516b7dcee7ac6a5742e4e6306a1a`:

- **P2 — Preserve cloud target routing through provider resolution.** `loadApiProvider('promptfoo://provider/<id>')` resolves the cloud record and returns the underlying `MCPProvider`, whose ID is normally just `mcp`. `RedteamMcpTargetProvider` then tries to recover the cloud ID only from that resolved provider (`src/redteam/mcpTargetProvider.ts:112-115`), so hosted MCP materialization can omit `targetId`. Please carry the selected cloud target ID through resolution or pass the existing red-team generation context into the wrapper. The current test constructs an MCP provider whose ID is already the cloud URI, so it does not exercise the real loader path.

- **P2 — Keep metadata out of pretty-printed empty-marker prompts.** With two empty `Prompt:` markers, a pretty-printed JSON payload triggers `parseMultiLinePrompts`, which appends `WARNING:`, `Plan:`, and `Reason:` lines to the previous payload until the next prompt marker (`src/redteam/plugins/multiInputFormat.ts:165-187`). I reproduced the first parsed prompt as valid JSON followed by the next case’s metadata, making the MCP input invalid. The empty-marker path should use the same boundary-aware collector for multi-line JSON, with a regression test.

I’m leaving these as follow-ups because the PR had already merged before this review round completed.

PR #9936, opened 2026-07-02
fix(redteam): stop presenting deprecated or invalid prompt-injection ids in tool guidance

mldangelo:
## Summary

Follow-up to #9931 (docs sweep of the deprecated `prompt-injection` strategy alias), covering the `src/` stragglers found during that review:

- **MCP tool descriptions presented `prompt-injection` as a plugin id** in `redteam_generate`'s plugins example and `redteam_run`'s error-guidance example config. It was never a valid plugin, so an agent following the example gets a validation error. Examples now use `indirect-prompt-injection` (plugin) and `jailbreak-templates` (strategy).
- **`riskScoring.ts` only had the deprecated alias key**: results recorded under the canonical `jailbreak-templates` id fell back to the default `humanComplexity: 'medium'` instead of `'low'`. Added the canonical key alongside the alias so both score identically.
- Fixed stale file names in the `src/redteam/AGENTS.md` layout tree (`prompt-injection.ts` → `promptInjections/`, `jailbreak.ts` → `iterative.ts`, `sql-injection.ts` → `sqlInjection.ts`, `harmful.ts` → `harmful/`).

## Testing

- `npx vitest run test/redteam/riskScoring.test.ts test/commands/mcp/server.test.ts` — 27 pass
- `tsc` clean, Biome clean, Prettier clean

### DIFF
--- a/src/commands/mcp/tools/redteamGenerate.ts
+++ b/src/commands/mcp/tools/redteamGenerate.ts
@@ -83,7 +83,7 @@ export function registerRedteamGenerateTool(server: McpServer) {
             
             Additional plugins: ${Array.from(REDTEAM_ADDITIONAL_PLUGINS).sort().join(', ')}
             
-            Example: ["harmful", "pii", "indirect-prompt-injection"]
+            Example: ["harmful", "pii", "hijacking"]
           `,
         ),
       strategies: z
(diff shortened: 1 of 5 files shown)


mldangelo:
Both findings addressed: examples now use `hijacking` (works as a bare string — `indirect-prompt-injection` indeed gets skipped without `indirectInjectionVar`), and `riskScoring.test.ts` now pins `jailbreak-templates` scoring to the deprecated alias so the canonical id can't regress to default scoring.

PR #9937, opened 2026-07-02
fix(providers): preflight Codex SDK compatibility

ianw-oai:
Reject custom Codex CLI overrides whose version does not match the CLI/event schema pinned by the loaded SDK before starting a turn.


src/providers/openai/codex-sdk.ts
```diff
@@ -2285,6 +2303,34 @@ export class OpenAICodexSDKProvider implements ApiProvider {
...
+    if (!codexPathOverride) {
+      return;
+    }
+    if (!this.codexSdkEntryPoint) {
+      throw new Error('Codex SDK entry point is unavailable for compatibility check');
+    }
+
+    // The SDK's direct CLI dependency is its event-schema compatibility contract. The bundled
+    // CLI already satisfies it, so only a custom override needs a version probe.
+    const preflightEnv = getMinimalProcessEnv();
```
mldangelo: **P1 — probe env diverges from the env the SDK will actually spawn with.**

`prepareEnvironment` builds the turn env as `(inherit_process_env ? process.env : minimal) + cli_env`, and `buildCodexOptions` hands that full env to the SDK spawn. The probe instead re-derives a minimal env and copies only `MINIMAL_CLI_ENV_KEYS` from it — every custom `cli_env` key (`LD_LIBRARY_PATH`, `CODEX_HOME`, wrapper-script vars, …) and every non-minimal inherited var is dropped.

So a binary that works fine at runtime can fail the probe (`error while loading shared libraries…` → `Could not verify Codex CLI compatibility…`), hard-blocking a config that worked before this PR — precisely for the custom-binary audience this feature targets.

If the intent of the minimal env is to avoid handing API keys to a not-yet-validated binary (the test asserts `CODEX_API_KEY`/`OPENAI_API_KEY` are absent, and that's a nice property), consider running the probe with the same prepared `env` minus the secret keys, rather than minus everything. That keeps the security property while making the probe representative of the real spawn.

(Related: whatever env feeds the probe should also feed the cache key — today a `cli_env` var that changes probe behavior isn't part of the key, so two configs sharing PATH can collide on one cached verdict.)

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+    output = `${stdout ?? ''}\n${stderr ?? ''}`;
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    throw new Error(
+      `Could not verify Codex CLI compatibility for ${options.codexPathOverride}: ${message}`,
+    );
+  }
+
+  const version = CODEX_VERSION_PATTERN.exec(output)?.[1];
+  if (!version || !semver.eq(version, supportedVersion)) {
```
mldangelo: **P1 (design) — exact-match hard fail with no escape hatch.**

Two constraints stack here: `semver.eq` against the SDK's pin, and a banner regex that requires a `codex-cli`/`codex-cli-exec` prefix. Any intentionally-divergent override — a patch-newer upstream CLI (promptfoo itself declares `^0.142.3` for the SDK, so SDK 0.142.3 + standalone CLI 0.142.5 is an easy state to be in), or an internal fork whose banner prints a different name — now hard-fails every turn. Before this PR those setups either worked (schema unchanged) or failed downstream; there is no config knob to accept the risk.

This file's sibling gate already has the precedent: `skip_git_repo_check` exists exactly for "I know what I'm doing" cases. I'd keep fail-closed as the default but add a `skip_codex_version_check` (or warn-only) option, and document it.

Also worth knowing: because rejected checks are evicted from the cache, a *persistently* incompatible override re-spawns the probe on every turn wave — a large eval against a mismatched CLI fails slowly (up to 10s per probe), not just completely.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+    );
+    output = `${stdout ?? ''}\n${stderr ?? ''}`;
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    throw new Error(
+      `Could not verify Codex CLI compatibility for ${options.codexPathOverride}: ${message}`,
+    );
+  }
+
+  const version = CODEX_VERSION_PATTERN.exec(output)?.[1];
```
mldangelo, line 88: **P2 — guard the captured version with `semver.valid` before `semver.eq`.**

The prerelease char-class `[0-9A-Za-z.-]+` is greedy and can swallow trailing punctuation, producing a semver-invalid capture. Reproduced locally:

```js
CODEX_VERSION_PATTERN.exec('codex-cli 0.142.3-rc. built today')?.[1]
// => '0.142.3-rc.'
semver.eq('0.142.3-rc.', '0.142.3')
// => TypeError: Invalid Version: 0.142.3-rc.
```

That raw `Invalid Version: …` escapes `runCompatibilityCheck` unwrapped, so the user sees it instead of the crafted "supports X, but reports Y" message. One-word fix reusing the branch you already have:

```ts
if (!version || !semver.valid(version) || !semver.eq(version, supportedVersion)) {
```

(Also: consider ending this error — and the `Could not verify…` one — with the docs link, matching `loadCodexSDK`'s errors in codex-sdk.ts:616/640. These are the strings users will paste into search.)

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+
+async function runCompatibilityCheck(options: CompatibilityOptions): Promise<void> {
+  const supportedVersion = getSupportedCliVersion(options.sdkEntryPoint);
+  let output: string;
+
+  try {
+    const { stdout, stderr } = await execFileAsync(
+      options.codexPathOverride,
+      ['exec', '--experimental-json', '--version'],
+      { encoding: 'utf8', env: options.env, timeout: 10_000, windowsHide: true },
```
mldangelo: **P2 — probe is un-cancellable, and a SIGTERM-ignoring binary hangs every turn.**

Two related issues:
1. The 10s `timeout` sends the default `SIGTERM`. `codex_path_override` executes an arbitrary user-supplied binary — if it traps/ignores SIGTERM (or sits on a stalled network mount), `execFile` never calls back, the never-settling promise is memoized in `compatibilityChecks` (eviction only happens on *rejection*), and every subsequent turn awaits it forever. `killSignal: 'SIGKILL'` (or SIGTERM-then-SIGKILL) makes the timeout reliable.
2. The caller's `abortSignal` is only consulted *after* the await in `getCodexInstanceForTurn`, so an abort during the probe waits out the full 10s. Wiring the signal into the shared `execFile` would be wrong (one caller's abort would poison the cached promise for concurrent callers), but racing the await against the signal in `ensureCodexCompatibility` — the same pattern as the thread-turn wait at codex-sdk.ts:1977-1995 — gives responsive cancellation while leaving the shared probe intact.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+    options.env.Path,
+    options.env.PATHEXT,
+  ].join('\0');
+  let check = compatibilityChecks.get(cacheKey);
+  if (!check) {
+    check = runCompatibilityCheck(options).catch((error) => {
+      compatibilityChecks.delete(cacheKey);
+      throw error;
+    });
+    compatibilityChecks.set(cacheKey, check);
```
mldangelo: **P3 — success is memoized for the process lifetime; a swapped binary is never re-validated.**

In long-lived contexts (promptfoo server, watch loops) a user can replace the binary at the same override path after a successful probe; the cached verdict then lets the mismatched CLI straight through to the SDK — reintroducing the silent JSONL-schema breakage this preflight exists to catch, in its primary target scenario. Cheap hardening: fold `statSync(codexPathOverride).mtimeMs` + `size` into the cache key. Fine as a follow-up, but worth a comment either way so the trade-off is recorded.

mldangelo (commented):
## Review: Codex SDK compatibility preflight

Solid idea, correctly aimed: the SDK's JSONL event parser really is version-locked to its exactly-pinned `@openai/codex` dependency, and probing the override binary with the SDK's own spawn shape (`exec --experimental-json --version`) is a faithful proxy — I verified with the vendored CLI that clap rejects unknown flags even when `--version` is present, so the probe genuinely validates flag acceptance, not just the banner. The promise-cache dedup with eviction-on-failure is correct, and the earlier bot findings (sticky rejected cache, leaked canary instance, optional-dep test read) are all addressed by the later commits.

**I verified this end-to-end with the local build** (`npm run local -- eval` with `codex_path_override` pointed at fake binaries):
- Mismatched binary (reports 0.999.0): eval fails with exactly `@openai/codex-sdk supports Codex CLI/event schema 0.142.3, but <path> reports 0.999.0` before any SDK spawn. ✅
- Matching binary (reports 0.142.3): preflight passes and control reaches the SDK's own spawn/parse. ✅
- Both new test files pass locally (161 tests). ✅

That said, there are two P1s I'd want addressed before merge, plus docs.

### P1 — probe env diverges from the real spawn env (inline at `codex-sdk.ts:2320`)
The probe runs the binary with `MINIMAL_CLI_ENV_KEYS` only, but the real SDK spawn gets the full prepared env including `cli_env` and (optionally) `inherit_process_env`. A wrapper/custom build that needs e.g. `LD_LIBRARY_PATH` from `cli_env` works at runtime today but will fail the probe → the PR bricks a working config with `Could not verify Codex CLI compatibility…`.

### P1 — hard fail with no escape hatch (inline at `codexCliCompatibility.ts:71`)
Exact `semver.eq` plus a `codex-cli`-prefixed banner regex means every intentionally-divergent override (patch-newer CLI, internal fork with a renamed banner) goes from working-or-silently-broken to hard-failing every turn, with no config opt-out. The same file already has the `skip_git_repo_check` escape-hatch precedent. Recommend a `skip_codex_version_check`-style flag (fail-closed default is fine) or a warn-only mode.

### P2s (inline)
- Probe isn't abortable and a SIGTERM-ignoring/hung binary never settles the memoized promise → all turns on that provider hang (`codexCliCompatibility.ts:60`).
- Version regex can capture a semver-invalid string; `semver.eq` then throws a raw `Invalid Version: …` instead of the crafted message — reproduced locally (`codexCliCompatibility.ts:70`).
- Test gap: nothing asserts the preflight is *skipped* without `codex_path_override`, and the new post-preflight abort branch is untested — these are codecov's uncovered lines (`openai-codex-sdk.test.ts:2429`).

### Docs audit — the PR ships a new hard-fail behavior with zero doc changes
- `site/docs/providers/openai-codex-sdk.md:792-801` ("Custom Binary Path") presents the override as unconstrained; it now needs the version-match requirement, the probe mechanics (one-time, 10s timeout), and where the pinned version lives (`node_modules/@openai/codex-sdk/package.json` → `dependencies["@openai/codex"]`).
- `site/docs/providers/openai-codex-sdk.md:230` (config table row) — same gap; there's also no Troubleshooting section documenting the three new error strings.
- `examples/openai-codex-sdk/README.md:152` advertises the override as a free "Key Feature" with no mention of the constraint.
- The new error messages don't link to docs, unlike `loadCodexSDK`'s errors in the same file (`codex-sdk.ts:616,640`) which end with `For more information, see: https://www.promptfoo.dev/docs/providers/openai-codex-sdk/`. Worth matching that convention — these errors are exactly the ones users will paste into search.
- FYI risk: `openai-codex-app-server.md` tells users to `npm i -g @openai/codex` (a floating version); pointing the codex-sdk provider's override at that global binary will now hard-fail whenever versions drift. A one-line cross-reference would prevent confusion.

### Follow-up worth tracking (not blocking)
`codex-app-server.ts` has the same `codex_path_override` and the same silent-version-drift failure mode (it parses app-server JSON-RPC from an arbitrary-version binary and discards the `initialize` handshake response). This PR guards one of the two providers. The SDK-pin check doesn't transfer (app-server doesn't use the SDK runtime), but validating its own handshake would close the same class of bug.

### Minor/cleanup (inline)
Unreachable defensive throw (`codex-sdk.ts:2314`), dead `?? ''` guards, success-cache never re-validating a swapped binary in long-lived server mode, unwrapped `JSON.parse` errors from the manifest walk, and a few test-shape nits. Reuse/efficiency review came back clean — the module follows the repo's established idioms (local `promisify(execFile)`, in-flight-promise map) and all expensive work is correctly behind the cache.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+  dependencies?: Record<string, string>;
+}
+
+function getSupportedCliVersion(sdkEntryPoint: string): string {
+  let directory = path.dirname(sdkEntryPoint);
+
+  while (true) {
+    const manifestPath = path.join(directory, 'package.json');
+    if (fs.existsSync(manifestPath)) {
+      const manifest = JSON.parse(fs.readFileSync(manifestPath, 'utf8')) as CodexSdkManifest;
```
mldangelo, line 30: **P3 — errors from the manifest walk escape without preflight context.**

`getSupportedCliVersion` runs outside the try/catch that wraps the exec probe, so a malformed `package.json` anywhere on the walk (UTF-8 BOM, JSONC) surfaces as a bare `SyntaxError: Unexpected token …` with no hint it came from the compatibility preflight or which file was at fault. Either wrap the `JSON.parse` per-manifest (skip unparseable non-SDK manifests, they can't be the one you want) or wrap the `getSupportedCliVersion` call in the same `Could not verify Codex CLI compatibility for <path>: …` context as the exec failure.

src/providers/openai/codex-sdk.ts
```diff
@@ -2285,6 +2303,34 @@ export class OpenAICodexSDKProvider implements ApiProvider {
...
 
+  private async ensureCodexCompatibility(
+    env: Record<string, string>,
+    resolvedConfig: OpenAICodexSDKConfig,
+  ): Promise<void> {
+    const codexPathOverride = resolvedConfig.codex_path_override;
+    if (!codexPathOverride) {
+      return;
+    }
+    if (!this.codexSdkEntryPoint) {
```
mldangelo: **Nit — unreachable guard / split state.**

`codexModule` and `codexSdkEntryPoint` are only ever assigned together (back-to-back, no await between them) in `getCodexInstanceForTurn`, and this method only runs after that block — so this branch can't fire. Storing the loaded pair as one field (`private codexSdk?: { entryPoint: string; module: any }`) removes the dead throw and the possibility of the two fields drifting apart.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,95 @@
...
+  const supportedVersion = getSupportedCliVersion(options.sdkEntryPoint);
+  let output: string;
+
+  try {
+    const { stdout, stderr } = await execFileAsync(
+      options.codexPathOverride,
+      ['exec', '--experimental-json', '--version'],
+      { encoding: 'utf8', env: options.env, timeout: 10_000, windowsHide: true },
+    );
+    output = `${stdout ?? ''}\n${stderr ?? ''}`;
```
mldangelo: **Nit** — with `encoding: 'utf8'`, `stdout`/`stderr` are always strings; the `?? ''` guards can't fire and imply a nullable case that doesn't exist.

test/providers/openai-codex-sdk.test.ts
```diff
@@ -2419,13 +2426,45 @@ describe('OpenAICodexSDKProvider', () => {
 
         await provider.callApi('Test prompt');
 
+        expect(mockCompatibilityPreflight).toHaveBeenCalledWith({
```
mldangelo, line 2432: **P2 — two coverage gaps on the new behavior's boundaries** (these line up with codecov's uncovered patch lines):

1. Nothing asserts the preflight is **skipped** when `codex_path_override` is absent. Every existing no-override test passes with the mock resolved, so an inverted/removed `if (!codexPathOverride) return;` guard — which would subject every default-config user to the probe — keeps the suite green. A one-liner in any no-override test locks it in: `expect(mockCompatibilityPreflight).not.toHaveBeenCalled();`
2. The new post-preflight abort branch (`Codex call aborted during compatibility check`, codex-sdk.ts:2270-2271) has no test — a preflight mock that aborts the signal before resolving would cover it.

Two smaller nits in the reject test below: the mocked rejection message embeds an SDK version (`@openai/codex-sdk 0.130.0 supports…`) that the real module never produces (it emits `@openai/codex-sdk supports Codex CLI/event schema 0.130.0…`), and the four thread/run `not.toHaveBeenCalled()` assertions can't fail independently of `MockCodex` never being constructed.

test/providers/openai/codexCliCompatibility.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    fs.rmSync(sdkRoot, { recursive: true, force: true });
+  });
+
+  function mockVersion(stdout: string, error: Error | null = null) {
+    mockExecFile.mockImplementation((_file, _args, _options, callback) => {
+      callback(error, { stdout, stderr: '' });
+    });
+  }
+
+  it('accepts the CLI version declared by the SDK', async () => {
```
mldangelo, line 41: **P3 — missing manifest-side fail-closed tests, and a hidden cache assertion.**

- The two `getSupportedCliVersion` fail-closed branches have no coverage: a range pin (`{'@openai/codex': '^0.130.0'}` → "does not declare an exact … version") and a missing/name-mismatched SDK manifest (→ "Could not find …"). The range-pin case is the branch most likely to fire in the wild — if a future SDK release relaxes its pin, every override user hard-fails, and nothing in the suite documents that this is intended.
- This test awaits the check twice and asserts `toHaveBeenCalledTimes(1)` at line 64 — that's really a cache-memoization test hiding inside the happy-path test. Splitting it into an explicit `it('caches successful checks', …)` pairs naturally with the failure-retry test at line 81 that asserts the opposite (`toHaveBeenCalledTimes(2)`).

src/providers/openai/codex-sdk.ts
```diff
@@ -2186,7 +2197,12 @@ export class OpenAICodexSDKProvider implements ApiProvider {
         resolvedConfig.skip_git_repo_check,
       );
 
-      const codexInstance = await this.getCodexInstanceForTurn(env, resolvedConfig, apiKey);
+      const codexInstance = await this.getCodexInstanceForTurn(
```
mldangelo-oai, line 2200: **P2 — Compatibility failures containing `429` are turned into retryable throttles.** This new preflight runs inside the broad catch below, which sends every error message through `buildCodexRateLimitResponse()`. A deterministic mismatch whose custom path contains `/429/`, or whose reported version is `0.429.0`, matches the standalone-`429` prose regex and comes back as a synthetic HTTP 429 with a 60-second retry hint instead of the actionable compatibility error. The adaptive scheduler can then add repeated minute-long waits before surfacing an exhausted-retry error. Please give compatibility failures a typed discriminator and bypass rate-limit classification for that type (or catch this await separately), with a regression test whose mocked preflight error contains a standalone `429` and asserts that no rate-limit metadata is returned.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (!version || !semver.valid(version) || !semver.eq(version, supportedVersion)) {
+    throw new Error(
+      `${CODEX_SDK_PACKAGE} supports Codex CLI/event schema ${supportedVersion}, but ${options.codexPathOverride} reports ${version ?? 'an unknown version'}. For more information, see: ${CODEX_SDK_DOCS_URL}`,
+    );
+  }
+}
+
+function getBinaryIdentity(codexPathOverride: string): string {
+  try {
+    const stats = fs.statSync(codexPathOverride);
```
mldangelo-oai: **P2 — PATH-resolved overrides reuse a stale successful check after replacement.** `execFile()` resolves a bare override such as `codex-custom` through `options.env.PATH`, but this `statSync()` checks the literal name relative to the process cwd and normally returns the constant `unavailable`. The cache key can therefore never observe changes to the executable actually selected from PATH. I reproduced this end to end: the first PATH binary reported `0.130.0` and passed; I replaced it with one that directly reported `0.131.0`; the second compatibility call still resolved successfully without probing it. That contradicts the documented guarantee that replacing the binary triggers a fresh check. Please resolve the override using the same PATH/PATHEXT rules as the probe before fingerprinting it (including realpath and stronger stat identity), or do not persist successful checks when the executable cannot be identified. Add a bare-command replacement regression test.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,128 @@
...
+    output = `${stdout}\n${stderr}`;
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    throw new Error(
+      `Could not verify Codex CLI compatibility for ${options.codexPathOverride}: ${message}. For more information, see: ${CODEX_SDK_DOCS_URL}`,
+    );
+  }
+
+  const version = CODEX_VERSION_PATTERN.exec(output)?.[1];
+  if (!version || !semver.valid(version) || !semver.eq(version, supportedVersion)) {
```
mldangelo-oai: **P2 — `semver.eq` does not enforce the documented exact build identity.** SemVer equality ignores build metadata: with the installed package, `semver.eq('0.142.3+fork', '0.142.3')` is `true` and `semver.compare(...)` is `0`, while `semver.compareBuild(...)` correctly reports a difference. A custom binary identifying itself as `0.142.3+fork` therefore passes a manifest pin of `0.142.3` without the explicit `skip_codex_version_check` opt-out that the docs require for forks. Since this gate is checking a CLI/event-schema implementation rather than version precedence, please compare build identifiers too (for example with `compareBuild`) and add a `+fork` mismatch test.

src/providers/openai/codexCliCompatibility.ts
```diff
@@ -0,0 +1,128 @@
...
+    const stats = fs.statSync(codexPathOverride);
+    return `${stats.size}:${stats.mtimeMs}`;
+  } catch {
+    // The probe below will report an actionable error if the binary is unavailable.
+    return 'unavailable';
+  }
+}
+
+export function checkCodexCliCompatibility(options: CompatibilityOptions): Promise<void> {
+  const envIdentity = crypto
```
mldangelo-oai: **P2 — Arbitrary inherited secrets are hashed into a long-lived cache key.** The caller removes only the OpenAI/Codex keys and tracing fields; with `inherit_process_env` or arbitrary `cli_env`, `options.env` can still contain AWS credentials, authenticated proxy URLs, GitHub tokens, or custom secrets. This hashes every value into the module-global successful-check map and retains the secret-derived identifier for the process lifetime. That directly conflicts with `src/providers/AGENTS.md`, which says to strip secrets before building or hashing a cache key and not to hash raw secrets into long-lived keys. Keep passing the representative environment to the probe, but avoid keying persistent successes on arbitrary values: coalesce only in-flight checks or bypass successful memoization for environment-dependent wrappers unless the caller supplies an explicit non-secret namespace.

mldangelo-oai (commented):
Reviewed exact head `f4b551ed20dd9966a57696754af7adc60364f089` in comment-only mode; I did not modify the branch. I found four material issues below.

Validation at this head: 170/170 focused provider/preflight tests passed locally; `npm run tsc`, `npm run build`, and `git diff --check` also passed. I reproduced the PATH-cache bypass with a real temporary executable and verified the build-metadata comparison against the installed `semver` package. All 13 required GitHub checks were green when I posted this review. The branch is currently 18 commits behind `main`, so it should be updated from `main` and the strict checks rerun after the fixes.

I reconciled all existing review threads. Fifteen earlier findings are fixed at this head. I did not repeat the outstanding Windows `.cmd`/`.bat` note: the pinned SDK also passes `codexPathOverride` directly to `spawn()` without a shell, so those shims already cannot start the actual turn; changing only this preflight would move the same failure later rather than fix an introduced regression.

mldangelo-oai (commented):
Re-reviewed exact head `e384d578a00c533f1de717d0681513691d8ebc12` after it changed during the sweep. The new main merge is clean, and commit `e384d578a` addresses all four findings from my prior review: compatibility failures now use a typed error that bypasses rate-limit prose classification; successful probes are rerun each turn so PATH replacements cannot reuse stale success; version comparison now includes build metadata; and the long-lived secret-derived environment cache key was removed. The docs and regressions were updated for the new per-turn behavior, PATH replacement, `+fork` rejection, and a compatibility message/path containing `429`. I did not find a new actionable issue in this fix delta.

Verification on the new exact head: `git diff --check` passes; the head contains current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`); and all 43 published checks are green. I inspected the exact fix and its regression coverage; I did not repeat the local focused suite in this follow-up. The prior material threads can be resolved. No branch changes or pushes were made.

PR #9938, opened 2026-07-02
feat(providers): add codex execution health

ianw-oai:
Before, the Codex SDK provider returned final output and raw items but no compact execution-health contract. Downstream health checks had to inspect raw events or text, while Promptfoo rate-limit handling classified error messages by text. A fixture, source file, or command mentioning `429`, `Could not resolve host`, or `ECONNREFUSED` could therefore look like a real provider failure.

After, Codex SDK responses include `metadata.executionHealth` with:

- numeric exit codes for completed commands
- the terminal structured provider code, kind, and HTTP status
- items that started but never completed
- a sandbox flag set only by a structured Codex sandbox discriminator

The scheduler now trusts these fields for Codex responses: structured 429s retry, quota errors fail fast, and diagnostic prose is ignored. A nonzero command exit is recorded as a tool result rather than automatically becoming a provider failure. Partial command results survive terminal stream failures, while recovered errors do not contaminate successful turns.

Scope: this changes only the Codex SDK provider. `droppedEvents` means a started item without a completion event. Successful skill calls remain in `metadata.skillCalls`.


src/providers/openai/codex-sdk.ts
```diff
@@ -481,83 +505,70 @@ const CODEX_RATE_LIMIT_CODES = [
...
-
-  if (
-    status !== 429 &&
-    code === undefined &&
-    !CODEX_RATE_LIMIT_PATTERNS.some((pattern) => pattern.test(message))
-  ) {
+    typeof normalizedError?.code === 'string' ? normalizedError.code.toLowerCase() : undefined;
+  const code = CODEX_RATE_LIMIT_CODES.find((knownCode) => knownCode === rawCode);
+  const retryAfterMs = getStructuredRetryAfterMs(errorRecord);
+  const isStructuredRateLimit = status === 429 || code !== undefined;
```
mldangelo: **P1 (blocking):** This gate is structured-or-nothing, but the `codex exec` JSONL protocol consumed by `@openai/codex-sdk@0.142.3` (the floor of the `^0.142.3` range) never emits `status`, `code`, or `codex_error_info` on error events — `ThreadError` is `{ message: string }` in the SDK's own types, and a live run of the pinned CLI against a mock 429 endpoint emits only `{"type":"error","message":"exceeded retry limit, last status: 429 Too Many Requests"}`. So `isStructuredRateLimit` can never be true in production, and real TPM/RPM throttles that were previously retried with the 60s backoff now surface as hard eval errors (the executionHealth gate in `src/scheduler/types.ts` blocks the generic prose fallback too). The rewritten test fixtures attach `{status: 429, code: 'rate_limit_exceeded'}` — fields the real protocol cannot produce.

Suggestion: keep structured-first, but retain a prose fallback scoped to terminal SDK/stream error messages only (never `aggregated_output`/command items — that already fixes the false-positive vector described in the PR body). Go structured-only when the exec protocol actually carries `codex_error_info`, bumping the SDK floor in the same change. Full analysis in the review summary.

src/scheduler/types.ts
```diff
@@ -41,7 +41,15 @@ export function isProviderResponseRateLimited(
...
-  if (isHttpRateLimitError(error) && error.kind === 'quota') {
+  if (isHttpRateLimitError(error)) {
+    return error.kind === 'rate_limit';
+  }
+  if (result?.metadata?.http?.status === 429) {
+    return true;
+  }
+  // Providers with execution health have already classified their failures
+  // from structured protocol data. Do not reinterpret their diagnostic prose.
+  if (result?.metadata?.executionHealth) {
```
mldangelo: The comment's premise ("have already classified their failures from structured protocol data") doesn't hold for the shipped SDK: the codex provider attaches `executionHealth` to **every** response — success, abort, config-parse errors, and generic transport errors — but can only actually classify when structured fields exist, which the pinned CLI's exec protocol never provides for stream-level 429s. Net effect: this gate disables the last safety net for the entire class of prose-only codex failures. See the P1 in the review summary.

src/providers/openai/codexExecutionHealth.ts
```diff
@@ -0,0 +1,128 @@
...
+  if (!root) {
+    return [];
+  }
+  return [root, asRecord(root.data), asRecord(root.cause)].filter(
+    (value): value is Record<string, unknown> => value !== undefined,
+  );
+}
+
+/** Extract only protocol fields. Error prose is deliberately ignored. */
+export function getCodexProviderError(error: unknown): CodexProviderError | undefined {
```
mldangelo, line 55: **P2:** With the pinned SDK, none of `codexErrorInfo`/`codex_error_info`/`kind`/`httpStatusCode` ever appear on SDK-originated errors (they exist only in the app-server JSON-RPC protocol, which this provider doesn't use). In production, `providerError` can only capture incidental Node errno codes (`ENOENT`/`ECONNREFUSED`) leaking from the spawn layer, and `sandboxFailure` can never be `true` — yet both are documented as a `schemaVersion: 1` contract exported from `promptfoo/contracts`. Suggest scoping the documented contract to what can populate today (`toolExitCodes`, `droppedEvents`) or explicitly marking these fields as forward-looking.

src/providers/openai/codex-sdk.ts
```diff
@@ -481,83 +505,70 @@ const CODEX_RATE_LIMIT_CODES = [
...
+  return typeof value === 'number' && Number.isFinite(value) ? value : undefined;
 }
 
-function extractCodexRetryAfterMs(message: string): number | undefined {
-  const match = message.match(
-    /\b(?:please\s+)?(?:try\s+again\s+in|retry(?:\s+again)?\s+(?:after|in))\s+(\d+(?:\.\d+)?)\s*(milliseconds?|ms|seconds?|secs?|s|minutes?|mins?|m)\b/i,
-  );
-  if (!match) {
-    return undefined;
+function getStructuredRetryAfterMs(errorRecord: Record<string, unknown>): number | undefined {
```
mldangelo: Three issues in this extractor:

1. **P2:** the header branches (`retry-after-ms`, `retry-after`) and the `cause`/`data` recursion are entirely untested — v8 coverage shows lines 528-531, 535-538, and 544-547 never execute; every retry test attaches `retryAfterMs` as a top-level number.
2. **P3:** the recursion into `cause`/`data` has no cycle guard; `error.cause = error` is legal JS and would stack-overflow inside the provider error path.
3. **P3:** the codex CLI's structured vocabulary uses `retry_after`/`retry_after_seconds` (visible in the binary's serde tables), which this doesn't read — so even future structured throttles would fall back to the 60s default.

src/providers/openai/codex-sdk.ts
```diff
@@ -1267,6 +1328,12 @@ export class OpenAICodexSDKProvider implements ApiProvider {
     const itemId = item.id ? String(item.id) : crypto.randomUUID();
```
mldangelo, line 1400: **P3:** An `item.completed` without an `id` gets keyed by `crypto.randomUUID()`, so the `state.activeItemTypes.delete(itemId)` below can never clear the entry created by its `item.started` — producing a false-positive `droppedEvents` entry for an item that actually completed.

src/providers/openai/codexExecutionHealth.ts
```diff
@@ -0,0 +1,128 @@
...
+  );
+}
+
+export function buildCodexExecutionHealth(
+  input: CodexExecutionHealthInput = {},
+): CodexExecutionHealth {
+  const toolExitCodes: CodexToolExitCode[] = [];
+  for (const value of input.items ?? []) {
+    const item = asRecord(value);
+    if (!item || item.type !== 'command_execution') {
```
mldangelo, line 118: **P3:** Only `command_execution` items with a numeric `exit_code` are aggregated. Failed `mcp_tool_call` and `file_change` items (both carry `status: 'failed'` in the SDK's `ThreadItem` union), and commands killed without an `exit_code` (the SDK documents it as optional), leave no trace in `executionHealth` — a turn whose only tool activity failed reports the same envelope as a clean turn. Worth either covering those item types or noting the limitation in the docs table.

mldangelo (changes requested):
## Review summary

Requesting changes. The mechanical work here is strong: forcing every turn through the event stream is a real improvement (the SDK's own `run()` performs the same last-`agent_message` aggregation internally, so output parity is exact, including `output_schema` runs and the no-message → `''` case), quota fail-fast survives the scheduler reordering, and skill/abort metadata handling is preserved. Local run of the two changed test files passes 196/196.

The blocker is the PR's central premise: that the Codex SDK surfaces structured error fields the provider can classify from. That is not true for the SDK/CLI version promptfoo ships, which flips this change from a false-positive fix into a false-negative regression on real throttles.

### P1 — Real Codex 429s lose all retry/backoff with the pinned SDK

`buildCodexRateLimitResponse` now fires only on `status === 429 || code !== undefined` (`src/providers/openai/codex-sdk.ts:566`), sourced exclusively from structured fields via `getCodexProviderError` (which deliberately ignores prose). The scheduler returns "not rate limited" for any response carrying `metadata.executionHealth` (`src/scheduler/types.ts:52`), and the provider now attaches `executionHealth` to **every** response, including generic errors. So classification is structured-or-nothing.

But the `codex exec` JSONL protocol consumed by `@openai/codex-sdk@0.142.3` (the floor of the `^0.142.3` range, and the installed version) never emits those fields. Verified three independent ways:

1. **SDK types/source**: `ThreadError = { message: string }` and the stream error event is `{ type: "error", message: string }` (`dist/index.d.ts`); `thread.run()` throws a bare `new Error(turnFailure.message)` (`dist/index.js`). No `status`, `code`, `retryAfterMs`, `headers`, or `codex_error_info` anywhere on SDK-originated errors.
2. **CLI binary inspection**: in the vendored codex 0.142.3 binary, `codex_error_info` / `http_status_code` / `additional_details` appear only in the app-server JSON-RPC protocol's serde tables (a different provider, untouched by this PR). The exec-event pool contains only `message`, `command`, `aggregated_output`, `exit_code`, and the event type names.
3. **Live reproduction**: running the pinned CLI with `exec --experimental-json` against a mock endpoint returning HTTP 429 (standard OpenAI rate-limit body, `retry-after: 1`) emits `{"type":"error","message":"exceeded retry limit, last status: 429 Too Many Requests"}` followed by a prose-only `turn.failed` — no structured fields.

Trace at this tip: `getCodexProviderError` → `undefined` → `buildCodexRateLimitResponse` → `undefined` → the response carries `executionHealth` → the scheduler gate short-circuits before the generic prose fallback that previously caught it. Before this PR, that exact message matched `CODEX_RATE_LIMIT_PATTERNS` and produced `rateLimitKind: 'rate_limit'` with the 60s backoff. After it, a large eval under sustained TPM throttling records hard errors en masse instead of backing off. The CLI's internal `request_max_retries`/`stream_max_retries` retries soften the blow (which is why this is P1 rather than P0), but the outer 60s backoff was the layer that survived sustained throttling.

The rewritten tests mask this: fixtures that were prose-only on main now attach `{status: 429, code: 'rate_limit_exceeded', retryAfterMs}` — fields the real exec protocol cannot produce — so the suite green-lights an unreachable code path.

**Suggested resolution:** keep structured-first classification, but retain prose matching as a fallback scoped to *terminal SDK/stream error messages only* — never `aggregated_output` or command items. That already addresses the false-positive vector in the PR description (a fixture or command output mentioning `429`), because nonzero exits are now recorded as tool results rather than provider failures. Go structured-only once the exec JSONL protocol actually carries `codex_error_info`, and bump the SDK floor in that same change.

### P2

1. **`providerError` and `sandboxFailure` are unpopulatable dead fields shipped as a versioned public contract** (`src/providers/openai/codexExecutionHealth.ts:43`). Given the protocol reality above, `sandboxFailure` can never be `true` in production and `providerError` only ever captures incidental Node errno codes (`ENOENT`/`ECONNREFUSED`) leaking from the spawn layer — yet the docs present both as a `schemaVersion: 1` contract exported from `promptfoo/contracts` that downstream health checks are invited to consume. Either scope the documented contract to what can actually populate today (`toolExitCodes`, `droppedEvents`), or clearly mark the rest as forward-looking.
2. **The new structured extraction paths are untested** (`src/providers/openai/codex-sdk.ts:518-548`). v8 coverage shows the `retry-after-ms` / `retry-after` header parsing (lines 528-531, 535-538) and the `cause`/`data` recursion (lines 544-547, plus `codexExecutionHealth.ts:35`) are never executed; every retry test attaches `retryAfterMs` as a top-level number. For a PR whose premise is "trust structured fields," the structured-field plumbing is the untested part.

### P3 — worth fixing while in here

- `getStructuredRetryAfterMs` recurses into `cause`/`data` with no cycle guard; `error.cause = error` is legal JS and would stack-overflow inside the provider error path (`codex-sdk.ts:540-548`).
- The CLI's structured error vocabulary uses `retry_after`/`retry_after_seconds`, which the extractor doesn't read — even future structured throttles would fall back to the blunt 60s default (`codex-sdk.ts:518`).
- An `item.completed` without an `id` is keyed by `crypto.randomUUID()` and can never clear its `item.started` entry in `activeItemTypes`, producing false-positive `droppedEvents` (`codex-sdk.ts:1328`).
- `executionHealth` ignores failed `mcp_tool_call`/`file_change` items and commands killed without an `exit_code`; a turn whose only tool activity failed reports the same envelope as a clean turn (`codexExecutionHealth.ts:104-119`).
- Scheduler reordering: the hoisted `metadata.http.status === 429 → true` now outranks the `'Quota exceeded:'` prose fail-fast for providers without `rateLimitKind` (narrow exposure: generic HTTP provider responses, which always attach `metadata.http.status`), contradicting the PR 8896 fail-fast comment at the top of the same function (`src/scheduler/types.ts:47`).
- `if (isHttpRateLimitError(error)) return error.kind === 'rate_limit'` treats any future `RateLimitKind` variant as non-retryable, contradicting the guidance in `src/util/fetch/errors.ts` to treat unknown kinds conservatively as retryable (`src/scheduler/types.ts:44`).
- Docs: "Promptfoo does not classify error message text such as `429` as a rate limit" (`site/docs/providers/openai-codex-sdk.md`) reads as a global statement, but the scheduler still prose-matches every other provider; worth scoping the sentence to Codex explicitly.

### Verified clean

Quota fail-fast (`rateLimitKind === 'quota'` is checked above all the reordered branches, and codex quota responses always carry it); output parity with the removed `thread.run()` path; `exit_code` is snake_case in the exec protocol so `toolExitCodes` populates; errno codes can't false-positive as rate limits (`CODEX_RATE_LIMIT_CODES` is an exact-match allowlist); the `promptfoo/contracts` export chain works as documented; and both earlier bot comments are addressed at this tip (the Copilot `cause`/`data` precedence concern, and the codex-connector P2 — though the fix for the latter overcorrected into the P1 above).

mldangelo (approved):
Re-reviewed at fb1a8cd. All blocking and material findings from my previous review are addressed — approving.

**P1 (throttle retries)** — resolved with the right shape: classification stays structured-first, and prose fallback is reinstated but scoped to the terminal stream boundary (`allowProseFallback: streamingError !== undefined` in `buildCodexExecutionErrorResponse`). Verified the restored patterns/extractors are byte-identical to the pre-PR classifiers, so prose-only 429s from the pinned exec protocol regain `rateLimitKind: 'rate_limit'` + the 60s default backoff, while command output and non-stream errors (config/spawn failures whose messages happen to contain `429`) are never classified — which preserves the false-positive fix this PR set out to make. Traced parity for the tricky paths: iterator-throw-after-prose-429 and turn-failed-with-previous-stream-error both classify the same as on main.

**P2s** — both resolved: `providerError`/`sandboxFailure` are now documented as reserved fields the current exec JSONL protocol doesn't emit (honest contract), and the previously uncovered extraction paths have direct tests (nested `cause`/`data` recursion, both `retry-after-ms`/`retry-after` header forms, seconds-based `retry_after*` fields, cycle tolerance).

**P3s** — cycle guard added; `retry_after`/`retryAfter`/`retry_after_seconds` now read; the `http.status === 429` check moved back below the `'Quota exceeded:'` fail-fast with a regression test; unknown future `RateLimitKind` variants now conservatively retry (`kind !== 'quota'`) with a test; the `toolExitCodes` scope limitation and the Codex-only scoping of the prose-classification sentence are documented.

One retraction on my side: the `item.completed`-without-`id` → false-positive `droppedEvents` note is moot — `id: string` is required on every `ThreadItem` variant in the SDK types, so the scenario can't occur with well-formed events. No change needed.

Verified locally at fb1a8cd: both changed test files pass 200/200, and all 39 CI checks are green.

mldangelo-oai (commented):
Reviewed exact head `fb1a8cdbe89da0bb0fa9559bf2e3de3dd1febd9a` in comment-only mode; I did not modify the branch. I found four material issues below.

Validation at this head: 163/163 Codex provider tests, 37/37 scheduler tests, and 44/44 contract/package tests passed locally; `npm run tsc`, `npm run build`, and `git diff --check` also passed. All required GitHub checks were green when I posted this review. The branch is currently 18 commits behind `main`, so it should be updated from `main` and the strict checks rerun after the fixes.

I also reconciled the existing review threads. The earlier blocking prose-only throttle regression is fixed at this head; the comments below are new findings rather than repeats of the prior review.

src/providers/openai/codex-sdk.ts
```diff
@@ -530,34 +554,92 @@ function extractCodexRetryAfterMs(message: string): number | undefined {
...
-  const status =
-    typeof errorRecord?.status === 'number'
-      ? errorRecord.status
-      : typeof errorRecord?.statusCode === 'number'
-        ? errorRecord.statusCode
-        : undefined;
+  const errorRecord = getRecord(error);
+  const normalizedError = getCodexProviderError(error);
+  const status = normalizedError?.httpStatusCode;
   const rawCode =
```
mldangelo-oai, line 626: **P2 — Numeric 429 codes bypass the structured rate-limit path.** The new public `CodexProviderError` contract allows `code: string | number`, and `getCodexProviderError()` preserves numeric values, but this line discards every numeric code. With `{ message: 'request rejected', code: 429 }`, I get `executionHealth.providerError.code === 429`, but no `rateLimitKind` or `metadata.http.status`; the scheduler then returns `false` because `executionHealth` is present. Please normalize numeric `429` as a structured HTTP 429 (or otherwise recognize it here) and add a provider-to-scheduler regression test using a neutral error message.

src/providers/openai/codexExecutionHealth.ts
```diff
@@ -0,0 +1,128 @@
...
+function getNumber(value: unknown): number | undefined {
+  return typeof value === 'number' && Number.isFinite(value) ? value : undefined;
+}
+
+function getErrorRecords(error: unknown): Record<string, unknown>[] {
+  const root = asRecord(error);
+  if (!root) {
+    return [];
+  }
+  return [root, asRecord(root.data), asRecord(root.cause)].filter(
```
mldangelo-oai: **P2 — Arbitrary nested payloads are promoted to provider errors.** This treats bare `data` and `cause` objects as equivalent to the root error without validating an error envelope. I reproduced `getCodexProviderError({ message: 'tool failed', data: { status: 429 } })` returning `{ httpStatusCode: 429 }`; `data: { kind: 'sandbox_error' }` similarly makes `isCodexSandboxError()` return `true`. An unrelated library payload can therefore become a synthetic rate limit with the default 60-second delay, or a sandbox failure. The pinned SDK defines `ThreadError` as message-only, and the docs say the current exec protocol does not emit these structured fields, so nested generic `data` is not a trusted Codex envelope. Please require an explicit supported envelope for `data` (and restrict bare fields to the root or a genuine error cause), with negative tests for unrelated nested `status`, `code`, and `kind` values.

src/scheduler/types.ts
```diff
@@ -57,12 +57,18 @@ export function isProviderResponseRateLimited(
   if (error?.message?.includes('Quota exceeded:')) {
     return false;
   }
+  if (result?.metadata?.http?.status === 429) {
+    return true;
+  }
+  // Codex providers classify terminal stream errors before returning a response.
+  // Do not reinterpret other execution-health diagnostic prose here.
+  if (result?.metadata?.executionHealth) {
```
mldangelo-oai: **P2 — This generic metadata key changes retry behavior for non-Codex providers.** `ProviderResponse.metadata` was already an open public map, so custom JS/Python providers could legally return their own `executionHealth` value. At this head, `{ error: 'HTTP 429: Too Many Requests', metadata: { executionHealth: { status: 'degraded' } } }` deterministically returns `false` here; on `main`, the same valid response reaches the existing 429 fallback and returns `true`. This helper has no provider identity or schema validation, so merely colliding with the new key disables retries globally. Please use an explicit scheduler classification emitted by Codex (for example, a dedicated `not_rate_limit` result), rather than the presence of an arbitrary metadata key, and retain a regression test for a generic provider using an unrelated `executionHealth` shape.

site/docs/providers/openai-codex-sdk.md
```diff
@@ -32,11 +32,11 @@ You can reference this provider using either base ID, and you can inline the mod
...
-| Shell/MCP/search/file trajectories | Yes        | Enable `enable_streaming` for provider-level spans. Enable `deep_tracing` to propagate OTEL context into the Codex CLI process.                                                                                                                                                                                                                                   |
+| Shell/MCP/search/file trajectories | Yes        | Execution health always aggregates completed items. Enable `enable_streaming` for item- and turn-level spans, and `deep_tracing` to propagate OTEL context into the Codex CLI process.                                                                                                                                                                            |
```
mldangelo-oai: **P2 — The capability table overstates what execution health contains.** `buildCodexExecutionHealth()` ignores every completed item except `command_execution` items with a numeric `exit_code`; MCP, search, and file items never appear in this envelope. Item- and turn-level spans also require `enable_streaming: true`. That conflicts with both “Execution health always aggregates completed items” here and the accurate limitation later on line 709. With tracing disabled, those other trajectories are only present in the untyped raw turn payload, not in `executionHealth`. Please distinguish raw item aggregation, the command-only health summary, and tracing spans in this row, or add typed aggregation for the advertised item types.

mldangelo-oai (commented):
Re-reviewed exact head `71075d8cc551bef1abff40d5e1ffcb71084f0b19` after it changed during the sweep. The new main merge is clean, and commit `71075d8cc` addresses all four findings from my prior review: numeric `code: 429` now enters the structured rate-limit path; unrelated nested `data` objects are no longer promoted without a Codex envelope; the scheduler now relies on an explicit `rateLimitKind` classification instead of a generic `executionHealth` key collision; and the capability table now distinguishes raw turn items, command-only execution-health summaries, and streaming spans. The added provider-to-scheduler and negative-envelope regressions exercise the reported cases. I did not find a new actionable issue in this fix delta.

Verification on the new exact head: TypeScript passes; 206 focused Codex-provider and scheduler tests pass; `git diff --check` passes; the head contains current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`); and all 43 published checks are green. The prior material threads can be resolved. No branch changes or pushes were made during this follow-up.

PR #9939, opened 2026-07-02
feat(assertions): add grader variable allowlists

ianw-oai:
Adds assertion-level `graderVars` allowlists so model graders receive only selected test variables and avoid oversized grading requests.


src/assertions/utils.ts
```diff
@@ -36,6 +49,32 @@ export function getFinalTest(test: TestCase, assertion: Assertion) {
...
+): Record<string, VarValue> | undefined {
+  if (assertion.graderVars === undefined) {
+    return vars;
+  }
+
+  if (!vars) {
+    return {};
+  }
+
+  const builtins = GRADER_BUILTIN_VARS[assertion.type];
```
mldangelo: **P1 — `not-` prefixed assertion types bypass the built-in protection.**

The lookup uses raw `assertion.type`, but every supported type also has a valid `not-` variant (`NotPrefixedAssertionTypesSchema`), and those dispatch to the same handlers via `baseType`. `GRADER_BUILTIN_VARS['not-llm-rubric']` is `undefined`, so `!builtins?.has(name)` passes every allowlisted name.

**Reproduced end-to-end** with the local build: `type: not-llm-rubric`, `graderVars: ['issue', 'output']`, test var `output: USER_VAR_NAMED_OUTPUT` → the grading prompt contained `<Output>\nUSER_VAR_NAMED_OUTPUT` instead of the actual model output, so the (inverted) verdict grades the wrong text. The identical config with plain `llm-rubric` was correctly protected.

Fix: resolve the base type before the lookup (handlers already receive `baseType` in `AssertionParams`, or strip the `not-` prefix here), and add a `not-llm-rubric` case to the shadow-protection test.

src/assertions/utils.ts
```diff
@@ -5,10 +5,23 @@ import yaml from 'js-yaml';
 import Clone from 'rfdc';
 import cliState from '../cliState';
 import { importModule } from '../esm';
-import { type Assertion, type TestCase } from '../types/index';
+import { type Assertion, type TestCase, type VarValue } from '../types/index';
```
mldangelo: **Blocking (mechanical) — this file conflicts with `origin/main`.**

Main migrated this import block to `import { loadYaml } from '../util/yamlLoad'` in the js-yaml v5 upgrade (#9919); js-yaml v5 has no default export, so a conflict resolution that keeps `import yaml from 'js-yaml'` will compile against the branch but break once merged. Please merge `origin/main` and switch any `yaml.load` call in this file to `loadYaml`.

src/assertions/llmRubric.ts
```diff
@@ -29,7 +30,7 @@ export const handleLlmRubric = async ({
     renderedValue || '',
     outputString,
     test.options,
-    test.vars,
+    getGraderVars(assertion, test.vars),
```
mldangelo, line 33: **P2 — excluded vars silently render as empty strings in custom `rubricPrompt` templates.**

The projected vars are exactly what `matchesLlmRubric` → `renderLlmRubricPrompt` uses to render a custom `rubricPrompt`. A template referencing `{{expected_answer}}` with `graderVars: ['question']` renders the reference criteria as an empty string — nunjucks doesn't throw and nothing warns, so the grader confidently grades against a gutted rubric.

This is the mirror image of the documented `value` behavior (rendered with full vars *before* projection), and the docs only state the `value` side; `site/docs/guides/llm-as-a-judge.md:546-550` still says any test var is available in `rubricPrompt`. At minimum document this; ideally warn when an assertion has both a custom `rubricPrompt` and `graderVars` (or detect unresolved references).

src/types/index.ts
```diff
@@ -729,6 +729,10 @@ export const AssertionSchema = z.object({
   // Override the grading rubric
   rubricPrompt: z.custom<GradingConfig['rubricPrompt']>().optional(),
 
+  // Limit test variables passed to model graders. Built-in grader inputs
+  // such as output, rubric, and criteria are always included.
+  graderVars: z.array(z.string()).optional(),
```
mldangelo, line 734: **P2 — accepted on every assertion type, honored by 10, silent everywhere else.**

Setting `graderVars` on `g-eval`, `answer-relevance`, `context-relevance`, `pi`, or any string-check type validates fine and does nothing. Most of those never forward vars to a grader, so the no-op is harmless-but-confusing — except **`conversation-relevance`**, which does forward raw `test.vars` into its grader's rubric render (`src/external/assertions/deepeval.ts:51-57` → `matchesConversationRelevance` spreads `{ messages, ...vars }` into `nunjucks.renderString`). A user who sets `graderVars: []` there to keep a sensitive or oversized var away from the grader gets zero effect and zero signal — precisely the exposure this feature exists to prevent.

Suggest either wiring `conversation-relevance` into the projection, or a `logger.warn` when `graderVars` is set on an unsupported type (precedent: `validateSessionConfig` warns on invalid-but-not-fatal config in `src/validators/util.ts`). The supported-type list already exists in three hand-synced places (this map's keys, the handler wraps, the docs sentence) — a warning would at least make the boundary observable.

src/assertions/utils.ts
```diff
@@ -36,6 +49,32 @@ export function getFinalTest(test: TestCase, assertion: Assertion) {
...
+  }
+
+  if (!vars) {
+    return {};
+  }
+
+  const builtins = GRADER_BUILTIN_VARS[assertion.type];
+  return Object.fromEntries(
+    assertion.graderVars
+      .filter((name) => !builtins?.has(name) && Object.prototype.hasOwnProperty.call(vars, name))
```
mldangelo, line 80: **P3 — allowlisted-but-absent names are dropped with no diagnostic.**

`graderVars: ['refrenceAnswer']` (typo) is indistinguishable from a correctly filtered var: the rubric's `{{referenceAnswer}}` renders empty and the grader returns a confident wrong verdict. A one-line `logger.debug` (or `warn`) for names that fail the `hasOwnProperty` check would make this diagnosable.

(The `hasOwnProperty` guard itself is load-bearing and correct — `name in vars` would let `graderVars: ['toString']` project prototype members. 👍)

src/assertions/utils.ts
```diff
@@ -5,10 +5,23 @@ import yaml from 'js-yaml';
 import Clone from 'rfdc';
 import cliState from '../cliState';
 import { importModule } from '../esm';
-import { type Assertion, type TestCase } from '../types/index';
+import { type Assertion, type TestCase, type VarValue } from '../types/index';
 
 const clone = Clone();
 
+const GRADER_BUILTIN_VARS: Readonly<Record<string, ReadonlySet<string>>> = {
```
mldangelo: **P3 (design) — a third hand-maintained copy of matcher-owned knowledge.**

I verified all 10 entries match the matchers' actual injected vars today (including the context-faithfulness union across its two prompts), so there's no present drift bug. But the authoritative names live in each matcher's spread (`{ output, rubric, ...vars }` in llmGrading.ts/agent.ts/search.ts/rag.ts/comparison.ts), and nothing ties this map to them — the drift test in `utils.test.ts:129` compares the map against a hardcoded copy of the same lists, so it can't catch divergence.

Cheap hardening for this PR: type the map as `Partial<Record<AssertionType, ReadonlySet<string>>>` so a typo'd/renamed key is a compile error, and share the duplicated Set literals (`output/rubric` ×3, factuality's set ×2 for the alias). Deeper follow-up worth considering: have matchers spread built-ins *after* (projected) user vars — `matchesTrajectoryGoalSuccess` already does — which deletes this map entirely and also fixes the pre-existing footgun where a test var named `output` replaces the real model output when `graderVars` is *not* set (I reproduced that on the control path).

src/assertions/index.ts
```diff
@@ -851,7 +857,7 @@ export async function runCompareAssertion(
     assertion.value,
     outputs,
     test.options,
-    test.vars,
+    getGraderVars(assertion, test.vars),
```
mldangelo, line 860: **P3 — consider computing the projection once instead of at 9 call sites.**

Eight handlers repeat `getGraderVars(assertion, test.vars)` (plus this one). `runAssertion` already builds `assertionParams` with `resolvedVars`; computing `graderVars: getGraderVars(assertion, resolvedVars)` there and letting handlers read `params.graderVars` removes the duplication and — more importantly — shrinks the latent-bug surface where a future model-graded handler forwards `test.vars` raw and silently skips projection. (Projecting deeper, in `getFinalTest` or the matchers, doesn't work: handlers legitimately read full `test.vars` for input extraction like `vars.query`/`vars.context`, and 5 matchers don't receive `assertion`.)

test/assertions/utils.test.ts
```diff
@@ -113,6 +114,30 @@ describe('coerceString', () => {
...
+    ['factuality', ['input', 'ideal', 'completion']],
+    ['llm-rubric', ['output', 'rubric']],
+    ['model-graded-closedqa', ['input', 'criteria', 'completion']],
+    ['model-graded-factuality', ['input', 'ideal', 'completion']],
+    ['context-faithfulness', ['question', 'answer', 'context', 'statements']],
+    ['context-recall', ['context', 'groundTruth']],
+    ['search-rubric', ['output', 'rubric']],
+    ['select-best', ['criteria', 'outputs']],
+    ['trajectory:goal-success', ['goal', 'output', 'trajectory']],
+  ] as const)('does not allow %s vars to shadow grader built-ins', (type, builtins) => {
```
mldangelo, line 132: **P3 — this drift test is self-referential, and `not-` variants are uncovered.**

The `it.each` table is a hardcoded copy of `GRADER_BUILTIN_VARS`, so it asserts the map equals itself — it can't catch the map drifting from what the matchers actually inject. Not easily fixable without exporting builtin names from matchers (see the design note on the map), but at minimum: add `not-llm-rubric` / `not-factuality` cases, which currently fail the guarantee this test documents (see the P1 on `utils.ts:70`).

test/assertions/modelGradedClosedQa.test.ts
```diff
@@ -73,9 +73,9 @@ describe('handleModelGradedClosedQa', () => {
     );
   });
 
-  it('should call matchesClosedQa with correct parameters', async () => {
+  it('should project grader vars while preserving closed-QA inputs', async () => {
```
mldangelo, line 76: **Nit — this file lost its only plain-path (no `graderVars`) matcher-args assertion.**

The retrofit converted the sole test asserting what reaches `matchesClosedQa` into a projection test. The pass-through path is still covered cross-file (`contextPropagation.test.ts` asserts `{ testVar: 'value' }` with no `graderVars`), so no net suite gap — but a local one-liner mirroring `llmRubric.test.ts`'s dedicated `should preserve all test vars when graderVars is omitted` would keep this file self-contained. (llmRubric's split of the three cases — omitted / empty / allowlist — is the right pattern.)

site/docs/configuration/expected-outputs/model-graded/index.md
```diff
@@ -39,6 +39,28 @@ Trajectory-based:
...
+      criteria: Summarize accurately
+      foo: file://fixtures/large-input.txt
+    assert:
+      - type: llm-rubric
+        value: The output satisfies {{criteria}}
+        graderVars:
+          - criteria
+```
+
+Set `graderVars: []` to pass no test variables. Built-in inputs such as the model output and rubric are still included. Names reserved for an assertion's built-in grader inputs are ignored if they appear in `graderVars`.
```
mldangelo, line 58: **Docs — three behaviors this section should state, plus discoverability:**

1. The reserved built-in names are neither listed nor linked — users can't find out what's ignored per type (several types never document their built-ins anywhere). Inline the per-type list or link to it.
2. Custom `rubricPrompt` templates render excluded vars as **empty strings** — the paragraph below covers the `value` direction but not this opposite one, and `guides/llm-as-a-judge.md:546-550` still promises all test vars in `rubricPrompt`.
3. On unsupported assertion types `graderVars` is silently ignored (notably `conversation-relevance`, which does send vars to its grader).

Discoverability: add a `graderVars` row to the assertion-properties tables in `configuration/reference.md:141-152` and `expected-outputs/index.md:46-57`, and mention it in the option enumerations at `index.md:99` (trajectory), `llm-rubric.md`, and `agent-rubric.md:74`. Minor: the supported list omits the `model-graded-factuality` alias.

mldangelo (commented):
## Review: `graderVars` allowlists

Good feature, well-tested where it's tested, and **I verified it works end-to-end with the local build**: a capturing grader provider confirmed that with `graderVars` set, an excluded large var reaches neither the rendered grading prompt, nor the grader provider's call context, nor (per code inspection) the remote-grading payload; allowlisted-but-absent names are dropped; and a test var named `output` cannot shadow the grader's real output. The `value`-rendered-before-projection behavior matches the docs, `site/static/config-schema.json` is an exact `npm run jsonSchema:generate` regeneration (no hand-edit drift), all 9 touched test files pass locally, and the `hasOwnProperty` guard is load-bearing (it keeps `graderVars: ['toString']` from projecting prototype members). The earlier Codex bot comment about built-in shadowing is addressed by the `GRADER_BUILTIN_VARS` commit — but incompletely, which is the headline finding.

### Blocking

1. **Merge conflict with `main`** — `src/assertions/utils.ts` conflicts with main's js-yaml v5 migration (`import yaml from 'js-yaml'` → `import { loadYaml } from '../util/yamlLoad'`, #9919). Verified with `git merge-tree`: this PR cannot merge as-is, and a resolution that keeps the v4 default import will break the build on main (js-yaml v5 has no default export). Please merge `origin/main` and adopt `loadYaml`.

2. **`not-` prefixed types bypass the built-in protection** (inline at `utils.ts:70`) — `GRADER_BUILTIN_VARS[assertion.type]` is keyed by base type strings, but every supported type also has a valid `not-` variant. **Reproduced end-to-end**: `type: not-llm-rubric` + `graderVars: ['output']` + a test var named `output` puts `<Output>USER_VAR_NAMED_OUTPUT</Output>` into the grading prompt instead of the real model output — the exact wrong-text-graded failure the map exists to prevent, and a direct violation of the new doc guarantee ("Names reserved for an assertion's built-in grader inputs are ignored"). The same config with plain `llm-rubric` is correctly protected. Look up by `baseType` and add a `not-` variant test.

### Should fix

3. **Custom `rubricPrompt` interaction is undocumented and silently destructive** (inline at `llmRubric.ts:33`) — the projected vars are what render a custom `rubricPrompt`, so a template referencing an excluded var renders it as an **empty string**, no error, no warning. The docs describe the `value` carve-out (rendered before projection) but not this opposite-direction behavior, and `site/docs/guides/llm-as-a-judge.md:546-550` still promises "Any test `vars`" are available in `rubricPrompt`.

4. **Silent no-op on unsupported types** (inline at `types/index.ts:734`) — the schema accepts `graderVars` on every assertion type but only 10 honor it. The sharp edge: `conversation-relevance` *does* forward raw `test.vars` into its grader's rubric render (`src/external/assertions/deepeval.ts:51-57` → `nunjucks.renderString(..., { messages, ...vars })`), so a user who sets `graderVars: []` there to keep a sensitive/large var away from the grader gets zero effect and zero warning — the exact exposure the feature exists to prevent. Either wire it in or warn on unsupported types.

### Docs audit
The new section is accurate for the happy path, but the feature is essentially undiscoverable and three behaviors are unstated:
- No `graderVars` row in either assertion-properties table: `site/docs/configuration/reference.md:141-152` and `site/docs/configuration/expected-outputs/index.md:46-57`. Highest-ROI fix.
- The new section should state: the custom-`rubricPrompt` effect (#3), that unsupported types silently ignore it (#4), and where the per-type reserved names live — "Names reserved for an assertion's built-in grader inputs are ignored" gives users no way to find those names (several types never document them anywhere).
- Option enumerations that omit it: `model-graded/index.md:99` ("you can set `threshold`, `provider`, or `rubricPrompt`" — trajectory supports `graderVars` too), `llm-rubric.md` (flagship page), `agent-rubric.md:74`.
- Minor: the supported list omits the `model-graded-factuality` alias, and the motivation could mention the remote-grading/data-minimization benefit (the projection also governs what's sent to remote graders).

### Design notes (non-blocking, worth a maintainer call before this pattern calcifies)
- **`GRADER_BUILTIN_VARS` is a hand-maintained copy of matcher-owned knowledge** (inline at `utils.ts:12`). All 10 entries are accurate today (each verified against the matcher spreads), but the authoritative names live in the matchers' `{ output, rubric, ...vars }` expressions, and the drift test in `utils.test.ts` compares the map against a hardcoded copy of itself. A deeper shape: have matchers spread built-ins *after* projected user vars (as `matchesTrajectoryGoalSuccess` already does), which would delete the map and fix the inconsistency at the source.
- **Assertion-only granularity breaks the sibling pattern**: `rubricPrompt`/`provider` live in `GradingConfig` (`test.options`), inherit via `defaultTest`, and `getFinalTest` folds assertion-level overrides down. `graderVars` supports none of that — a user with 8 model-graded assertions repeats it 8 times. Consider `test.options.graderVars` with assertion-level override.
- **Nothing enforces adoption by the 11th grader** — a future model-graded handler that forwards `test.vars` raw silently skips projection. Computing the projection once into `AssertionParams` (inline at `index.ts:860`) shrinks that surface.
- Pre-existing, not this PR's fault: without `graderVars`, a test var named `output` still replaces the real model output in the grading prompt (I reproduced this on the control path). This PR's protection only engages when `graderVars` is set, which makes the same config safe or unsafe depending on an unrelated option. The matcher-side fix above would close it uniformly.

### Findings I checked and am *not* raising
- A workflow finding claimed YAML `graderVars: null` causes a runtime `TypeError` — refuted: `validateAssertions` hard-fails at config load with a clean `AssertValidationError` before any grading.
- A claimed loss of plain-path (no-`graderVars`) test coverage from retrofitting existing tests — mostly refuted: siblings preserve the pass-through path in every file except `modelGradedClosedQa.test.ts` (covered cross-file; minor inline note).
- Efficiency and reuse are clean: zero-cost fast path when `graderVars` is unset, projection isn't defeated downstream (`callProviderWithContext` overrides `context.vars` with the projected set), and no existing pick/projection helper was missed.

mldangelo (approved):
Reviewed the full diff at f78308d5443d (deep review at 87b6423ce5, then a delta review of the three newer commits). Approving — the design is sound and the issues found during review are all addressed at this tip.

## What was verified

- **Reserved-name table exactness**: every `GRADER_BUILTIN_VARS` entry was checked against what its matcher actually injects and the merge order. For 8 of the wired types the matchers spread `...vars` last (`{output, rubric, ...vars}` in `llmGrading.ts` and friends), so the filter is load-bearing, not belt-and-suspenders; `trajectory:goal-success` spreads builtins last and is safe either way. No missing names, no spurious names.
- **`not-` variants**: at 87b6423 this was a real hole — `GRADER_BUILTIN_VARS[assertion.type]` missed `not-llm-rubric` (dispatch strips the prefix only for handler selection), and a probe confirmed `graderVars: ['output']` let a test var replace the model's real output in the grading prompt for every `not-` grader type. The baseType strip now in `getGraderVars` (src/assertions/utils.ts:74) fixes it, the `utils.test.ts` table includes `not-llm-rubric`/`not-factuality` regression cases, and the docs call out the protection explicitly.
- **Rendering order**: the docs claim that assertion `value` renders before projection holds for the runAssertion-dispatched types (renderedValue computed with full resolved vars before handler dispatch), and the two `runAssertion.test.ts` integration tests pin it end-to-end through real nunjucks, including that excluded vars stay out of the grader provider context.
- **`rubricPrompt` interaction**: custom rubric prompts render with the *projected* vars under `throwOnUndefined=false`, so a var referenced only in `rubricPrompt` and excluded by `graderVars` silently renders as an empty string (reproduced on both the JSON-message and legacy string paths). The new docs sentence now states this explicitly, which closes the trap; a runtime warning when a projected-away name appears in the template would be a nice follow-up but is not required.
- **`conversation-relevance` wiring** (new in the latest commits): the local `getConversationGraderVars` is correct — the `messages` filter is unconditional (no type-keyed lookup, so the `not-` concern cannot recur there), and `_conversation` is read from the raw test vars before projection, so windowing still works when it is not allowlisted.
- Projection itself is prototype-safe (`Object.prototype.hasOwnProperty.call`), `graderVars: []` semantics are pinned, schema validation round-trips, and the docs reserved-name enumeration matches the code table entry-for-entry.

Test basis: 653/653 across the nine touched test files at 87b6423 locally; at f78308d verified by inspection plus the green 39-check CI run (my local node_modules predates the js-yaml v5 merge, so I did not rerun locally).

## Non-blocking notes

1. **Docs**: "a rubric containing `{{foo}}` still includes the value of `foo` after variable substitution" is not accurate for `select-best` — `runCompareAssertion` passes `assertion.value` raw to `matchesSelectBest`, which inserts it as `{{criteria}}` in a single render pass; test vars are never substituted into the value. Worth scoping that sentence to the runAssertion-dispatched types.
2. **Coverage**: the default no-`graderVars` passthrough is only explicitly pinned for `llm-rubric`; the in-place rewrites of the agentRubric/contextRecall/modelGradedClosedQa/trajectoryGoalSuccess tests converted their original parameter-passing tests into projection tests. A small `it.each` passthrough case per handler would restore that.
3. `getConversationGraderVars` intentionally duplicates the projection logic to keep `src/external` off `src/assertions` (architecture boundary); a one-line cross-reference comment in each would help keep the two in sync.

site/docs/configuration/expected-outputs/model-graded/index.md
```diff
@@ -39,6 +39,43 @@ Trajectory-based:
...
+- `model-graded-closedqa`: `input`, `criteria`, `completion`
+- `context-faithfulness`: `question`, `answer`, `context`, `statements`
+- `context-recall`: `context`, `groundTruth`
+- `conversation-relevance`: `messages`
+- `trajectory:goal-success`: `goal`, `output`, `trajectory`
+- `select-best`: `criteria`, `outputs`
+
+The same protection applies to `not-` variants such as `not-llm-rubric`.
+
+`graderVars` does not remove values already inserted into an assertion's `value`. For example, a rubric containing `{{foo}}` still includes the value of `foo` after variable substitution. To keep a variable out of the grader request, omit it from both `graderVars` and `value`.
```
mldangelo-oai: **P2 — narrow this data-exclusion guarantee.** `graderVars` only filters the ambient variable map; required built-ins are populated independently. For example, `context-recall` resolves `test.vars.context` and passes it separately to the matcher, which includes it in both the prompt and provider context even with `graderVars: []`. The same pattern applies to faithfulness `question`/`context`, conversation messages, factuality input, and trajectory data. A user can follow this sentence and still send the supposedly excluded value to the grader. Please qualify it to say the value must also not be used to construct a built-in input such as the evaluated prompt, output, question, context, messages, goal, or trajectory.

src/assertions/trajectory.ts
```diff
@@ -643,7 +644,7 @@ export const handleTrajectoryGoalSuccess = async (
     trajectory,
     params.outputString,
     params.test.options,
-    params.assertionValueContext.vars,
+    getGraderVars(params.assertion, params.assertionValueContext.vars),
```
mldangelo-oai, line 647: **P2 — update the required trajectory guide.** This line adds user-visible `graderVars` behavior to `trajectory:goal-success`, but the scoped `src/assertions/AGENTS.md` rule says: `Trace or trajectory changes also need site/docs/tracing.md.` The expected-outputs docs are updated, while the canonical trajectory example in `site/docs/tracing.md` still omits this control. Please add a short example or link to the authoritative `graderVars` section, including empty-list and reserved-built-in behavior.

src/assertions/factuality.ts
```diff
@@ -26,7 +27,7 @@ export const handleFactuality = async ({
       renderedValue,
       outputString,
       test.options,
-      test.vars,
+      getGraderVars(assertion, test.vars),
```
mldangelo-oai, line 30: **P2 — add projection-sensitive tests for the remaining changed handlers.** `factuality`/`model-graded-factuality`, `context-faithfulness`, and `search-rubric` now call `getGraderVars`, but their handler tests never set `graderVars`; they cover only the omitted compatibility path. The utility table cannot catch a call site accidentally reverting to raw `test.vars`, which would silently resend excluded large or sensitive values while the suite stays green. Please add sentinel tests for the three unique handlers with one allowlisted and one excluded variable, and assert the matcher/provider receives only the allowlisted variable while real built-ins remain intact.

mldangelo-oai (commented):
## Fresh review at `f78308d5443dc0a114a28904b14e000b353853f3`

The core allowlist works in the covered paths, but four items still need attention:

1. The existing [context-faithfulness P2](https://github.com/promptfoo/promptfoo/pull/9939#discussion_r3522306526) remains valid. `context` is filtered as reserved, but the first custom faithfulness prompt receives only `question`, `answer`, and projected vars; resolved context is added only to the second NLI prompt. A first-stage `{{context}}` therefore becomes empty when `graderVars` is enabled.
2. The docs overstate what omitting a variable can keep out of the grader request (inline).
3. The trajectory change is missing the `site/docs/tracing.md` update required by the scoped repository instructions (inline).
4. Three changed projection call sites still lack regression-sensitive handler tests (inline).

Verification on this exact head: 678 tests passed across 12 focused files; `npm run tsc` passed; `npm run build` passed; regenerated `site/static/config-schema.json` matched byte-for-byte after formatting; and a secret-free local eval with an echo subject plus file grader confirmed that an allowlisted sentinel reached the grader while an excluded sentinel did not. All live checks are green.

The branch is one commit behind current `main`; that upstream commit only changes `package-lock.json` for Playwright, and a synthetic merge showed no conflict. Strict freshness still means the branch needs an update before merge. No branch changes or pushes were made during this review.

mldangelo-oai (commented):
Re-reviewed exact head `ec21783f879a39ba8a9705f481d9a0479c91181a` after it changed during the sweep. The new merge from current `main` is clean, and commit `ec21783f8` addresses all four material items from my prior review: resolved context is now available to the first and second faithfulness prompts; the data-exclusion wording now distinguishes ambient variables from built-in grader inputs; `site/docs/tracing.md` documents trajectory `graderVars`; and projection-sensitive handler coverage was added for factuality, context-faithfulness, and search-rubric. I did not find a new actionable issue in this fix delta.

Verification on the new exact head: TypeScript passes; 38 focused tests across the four changed assertion/matcher suites pass; `git diff --check` passes; the head contains current `main` (`e8adf8bf3ccc362801030ec041e4add50c4f329a`); and every published check is green. The prior material threads can be resolved. No branch changes or pushes were made during this follow-up.

PR #9940, opened 2026-07-02
feat(config): add YAML source location to prompt reference errors

JustasMonkev:
## Summary

Prompt reference validation errors now point at the offending line in the configuration file.

**Before**

```text
> Test #1 ("Missing prompt") references prompt "Missing Prompt" which does not exist.

(Use `node --trace-warnings ...` to show where the warning was created)

file://text-to-sql-ts/node_modules/promptfoo/dist/src/main.js:1761
                        throw new PromptReferenceValidationError(
                              `Test #${testIdx + 1}${testDesc} references prompt "${ref}" which does not exist.\nAvailable prompts: [${available.join(", ")}]`
                        );
                              ^

PromptReferenceValidationError: Test #1 ("CoT prompt includes thinking tag") references prompt "Chain of Thought & Few-Shot Examples" which does not exist.
Available prompts: [Basic]
```

**After**

```text
> promptfooconfig.yaml:9:9: Test #1 ("Missing prompt") references prompt "Missing Prompt" which does not exist.

$ promptfoo eval -c promptfooconfig.yaml --output data/results.csv

(node:62047) ExperimentalWarning: DecompressInterceptor is experimental and subject to change
(Use `node --trace-warnings ...` to show where the warning was created)

file:///promptfoo_promptfoo/src/util/validateTestPromptReferences.ts:162
      throw new PromptReferenceValidationError(location ? `${location}: ${message}` : message);
            ^

PromptReferenceValidationError: text-to-sql-ts/promptfooconfig.yaml:35:9: Test #1 ("CoT prompt includes thinking tag") references prompt "Chain of Thought & Few-Shot Examples" which does not exist.
Available prompts: [Basic]
```

### DIFF
--- a/src/util/config/load.ts
+++ b/src/util/config/load.ts
@@ -455,13 +455,7 @@ export async function maybeReadConfig(configPath: string): Promise<UnifiedConfig
   }
 }
 
-async function readPromptReferenceSources(
-  configPaths?: string[],
-): Promise<PromptReferenceSource[]> {
-  if (!configPaths) {
-    return [];
-  }
-
+async function readPromptReferenceSources(configPaths: string[]): Promise<PromptReferenceSource[]> {
   const sources: PromptReferenceSource[] = [];
   for (const configPath of configPaths) {
     const resolvedPath = path.resolve(process.cwd(), configPath);
--- a/src/util/config/load.ts
+++ b/src/util/config/load.ts
@@ -45,7 +45,10 @@ import { filterProviderConfigs, getProviderIdAndLabel } from '../eval/filterProv
 import { filterTests } from '../eval/filterTests';
 import { promptfooCommand } from '../promptfooCommand';
 import { readTest, readTests } from '../testCaseReader';
-import { validateTestPromptReferences } from '../validateTestPromptReferences';
+import {
+  type PromptReferenceSource,
+  validateTestPromptReferences,
+} from '../validateTestPromptReferences';
 import { validateTestProviderReferences } from '../validateTestProviderReferences';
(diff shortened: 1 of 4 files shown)


mldangelo:
Reviewed and QA'd this thoroughly — the location hints are a genuinely nice DX win, and I verified the hand-rolled locator holds up well on real configs (via actual `promptfoo eval` runs, not just unit mocks):

- **block style** → `config.yaml:9:9` ✓
- **flow style** (`prompts: ['Missing']`) → `config.yaml:8:16` ✓ (points inside the flow list)
- **comment before the ref** → correctly skips the comment line ✓
- **ref string also in `vars`** → correctly points at the `prompts:` line, not the vars line ✓
- **multiple `-c` config files** → attributes to the right file ✓

On the three Codex P2s: #1 (ref-in-vars stealing the location) is already handled by the `prompts:`-block restriction, and #2 (CLI-supplied `--tests`) degrades gracefully to no location. But **#3 (substring match) was real** — I reproduced it: a valid prompt `GhostBuster` referenced above a missing `Ghost` in the same block made the error point at `GhostBuster`'s line, since `refLocation` used a raw `indexOf`.

I pushed a small fix (`7020b8b1d8`): the locator now prefers a list item whose *whole value* equals the ref, falling back to a substring match only when there's no exact one — so flow-style (`prompts: [ref]`) still resolves. Added a regression test for the substring case. 175 tests pass, tsc + Biome clean.

Leaving the merge to you since I added to it — ready once CI is green.

mldangelo:
Thank you!

PR #9942, opened 2026-07-02
fix(util): report file and line for malformed JSON/JSONL test files

orangeCatDeveloper:
## Problem

When an external test file (`tests: some.jsonl` / `some.json`) contains malformed JSON, `promptfoo eval` crashes with the raw `JSON.parse` error, which is hard to act on:

```
SyntaxError: Expected ',' or '}' after property value in JSON at position 59 (line 1 column 60)
    at JSON.parse (<anonymous>)
```

Two concrete pain points:

1. **No file name.** When a config references several test files, there's no indication which one failed.
2. **Misleading line number.** JSONL is parsed line-by-line, so `JSON.parse` always reports `line 1` — its view of the single line it was handed, *not* the line in the file. A broken row on file line 8 still says `line 1`, which is actively misleading for large datasets.

## Fix

Wrap the `JSON.parse` calls in the external test-file loaders so failures are attributed to their source file (and, for JSONL, the real file line, counting blank lines). The original engine message is preserved after the prefix, so no detail is lost.

## Before / after

Given `tests.jsonl` with 6 valid rows, a blank line 7, and a broken row on **file line 8**:

**Before**
```
SyntaxError: Expected ',' or '}' after property value in JSON at position 59 (line 1 column 60)
    at JSON.parse (<anonymous>)
```

**After**
```
Error: Failed to parse JSONL test file /path/to/tests.jsonl on line 8: Expected ',' or '}' after property value in JSON at position 59 (line 1 column 60)
```

Multiple test files — one broken (`broken.jsonl`), the others valid:

**After** (points at the exact file, not the others)
```
Error: Failed to parse JSONL test file /path/to/broken.jsonl on line 2: Expected ',' or '}' after property value in JSON at position 22 (line 1 column 23)
```

Malformed whole-file `.json` (glob/array form) now names the file too:
```
Error: Failed to parse JSON test file /path/to/bad.json: Unexpected token 'B', ..."{"vars": BROKEN }"... is not valid JSON
```

## Implementation notes

- New `parseJsonOrThrow(text, context)` is the single point that wraps `JSON.parse`; both the JSONL per-line path and the glob `.json` path route through it, so error formatting is unified (previously there were three separate bare `JSON.parse` sites).
- `parseJsonlLines` iterates with `.entries()` and skips blank lines *without* resetting the index, so the reported line number matches the file (verified by a dedicated test).
- The raw engine message is passed through verbatim — no regex rewriting of `JSON.parse`'s output, which varies across Node versions.

## Tests

Added 4 regression tests in `test/util/testCaseReader.test.ts`:

- malformed JSONL reports the real file line (line 2)
- blank lines are counted toward the reported line number (broken row on file line 3, not index 2)
- malformed JSONL via the glob path
- malformed `.json` via the glob path

All 89 tests in the file pass; `tsc` and lint are clean. Verified end-to-end against the real CLI (`promptfoo eval`) for: valid file (no regression), broken JSONL, multiple files with one broken, and broken JSON.

## Drive-by

Fixed a few comment/message typos in the same spirit: `no longer` (in the `--filter-description` deprecation warning), `gauge`, `database`, `visualization`.


mldangelo (approved):
Reviewed and QA'd end to end — this is a clean, well-tested DX improvement.

Verified the real CLI behavior (not just unit mocks):
- Malformed JSONL with a blank line 2 and a broken row on line 3 → `Failed to parse JSONL test file /tmp/.../bad.jsonl on line 3: ...` — the blank-line-aware line numbering is correct, and it names the file.
- Malformed standalone JSON → `Failed to parse JSON test file .../bad.json: ...` with the parser's line:col.
- 91 tests in `testCaseReader.test.ts` pass; tsc + Biome clean.

Nice touches:
- The refactor into `parseJsonOrThrow` / `parseJsonlLines` dedupes the standalone and glob code paths rather than copy-pasting the wrapping.
- Reporting the real file line for errors while keeping `Row #N` (index among non-blank rows) for descriptions is the right call — and you pinned that distinction with a dedicated test.
- Redacting the SAS token via the newly-exported `sanitizeAzureBlobUriForError` so blob URIs don't leak secrets into parse errors is a good security-conscious touch.

The incidental typo fixes in `view.ts`/`accounts.ts`/`eval.ts` are harmless and welcome. Merging.

PR #9947, opened 2026-07-02
feat(app): surface websocket subprotocols

jameshiester-oai:
## Summary
- adds a WebSocket Subprotocols field to the redteam target setup UI
- stores negotiated subprotocols in config.protocols instead of overloading headers
- preserves raw Sec-WebSocket-Protocol headers and shows migration guidance when they are present

## Testing
- npm run test:app -- WebSocketEndpointConfiguration.test.tsx

Depends on the runtime support merged in #9944.

### DIFF
--- a/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
@@ -182,7 +182,8 @@ function ProviderConfigEditor({
     } else if (
       field in updatedTarget.config ||
       field === 'streamResponse' ||
-      field === 'transformResponse'
+      field === 'transformResponse' ||
+      field === 'protocols'
     ) {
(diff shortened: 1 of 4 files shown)


mldangelo (approved):
Reviewed and verified locally. This is the right, non-breaking counterpart to the WebSocket subprotocol work: it surfaces a Subprotocols field that writes to `config.protocols` (the negotiation path added in #9944), and when a legacy `Sec-WebSocket-Protocol` header is present it *guides* the user to migrate without touching the header — so existing raw-header configs keep working. The migration-guidance test using `Bearer token-with-space` nicely pins that we don't promote non-subprotocol values (the exact case that makes auto-promotion unsafe).

- `WebSocketEndpointConfiguration` + `ProviderConfigEditor` tests: 11 + 31 pass
- app `tsc` clean, Biome clean (only pre-existing complexity warnings)
- field is correctly scoped to WebSocket targets and label/input are properly associated

Nice work.

PR #9949, opened 2026-07-02
fix(providers): support legacy websocket subprotocol headers

mldangelo-oai:
## Summary
- promote legacy `Sec-WebSocket-Protocol` headers into the `ws` constructor protocol argument
- remove the raw protocol header while preserving unrelated WebSocket headers
- add regression coverage for comma-separated legacy protocol values

## Why
Passing `Sec-WebSocket-Protocol` only as a raw header leaves `ws` with an empty requested-protocol set. If the server selects a protocol, the client aborts the handshake with `Server sent a subprotocol but none was requested`.

## Test plan
- `source ~/.nvm/nvm.sh && nvm use >/dev/null && ./node_modules/.bin/vitest run test/providers/websocket.test.ts --sequence.shuffle=false`
- `source ~/.nvm/nvm.sh && nvm use >/dev/null && npm run l`
- `source ~/.nvm/nvm.sh && nvm use >/dev/null && npm run f`
- `source ~/.nvm/nvm.sh && nvm use >/dev/null && npm run tsc`
- `git diff --check`

### DIFF
--- a/src/providers/websocket.ts
+++ b/src/providers/websocket.ts
@@ -35,6 +35,27 @@ function normalizeWebSocketProtocols(protocols: string | string[] | undefined): 
     .filter(Boolean);
 }
 
+function getWebSocketConnectionOptions(config: WebSocketProviderConfig): {
+  protocols: string[];
+  options: ClientOptions;
+} {
+  const protocolHeaderKey = Object.keys(config.headers ?? {}).find(
+    (key) => key.toLowerCase() === 'sec-websocket-protocol',
+  );
+  const legacyProtocolHeader = protocolHeaderKey ? config.headers?.[protocolHeaderKey] : undefined;
+  const protocols = normalizeWebSocketProtocols(config.protocols ?? legacyProtocolHeader);
+  const headers = protocolHeaderKey
+    ? Object.fromEntries(
+        Object.entries(config.headers ?? {}).filter(([key]) => key !== protocolHeaderKey),
+      )
+    : config.headers;
+
+  return {
+    protocols,
+    options: headers && Object.keys(headers).length > 0 ? { headers } : {},
+  };
+}
+
 function getWebSocketErrorMessage(err: WebSocket.ErrorEvent | Error | unknown): string {
   const error =
     err && typeof err === 'object' && 'error' in err ? (err as WebSocket.ErrorEvent).error : err;
@@ -246,11 +267,7 @@ export class WebSocketProvider implements ApiProvider {
     logger.debug(`Sending WebSocket message to ${this.url}: ${message}`);
     let accumulator: ProviderResponse = { error: 'unknown error occurred' };
     return new Promise<ProviderResponse>((resolve, reject) => {
-      const wsOptions: ClientOptions = {};
-      const protocols = normalizeWebSocketProtocols(this.config.protocols);
-      if (this.config.headers) {
-        wsOptions.headers = this.config.headers;
-      }
+      const { protocols, options: wsOptions } = getWebSocketConnectionOptions(this.config);
       const ws =
         protocols.length > 0
           ? new WebSocket(this.url, protocols, wsOptions)
--- a/test/providers/websocket.test.ts
+++ b/test/providers/websocket.test.ts
@@ -145,10 +145,10 @@ describe('WebSocketProvider', () => {
     expect(WebSocket).toHaveBeenCalledWith('ws://test.com', ['json'], {});
   });
 
-  it('should preserve Sec-WebSocket-Protocol headers unless protocols are configured', async () => {
+  it('should promote legacy Sec-WebSocket-Protocol headers to requested protocols', async () => {
     const headers = {
       Authorization: 'Bearer test-token',
-      'Sec-WebSocket-Protocol': 'Bearer token-with-space',
+      'Sec-WebSocket-Protocol': 'json, v1',
     };
 
     provider = new WebSocketProvider('ws://test.com', {
@@ -165,7 +165,9 @@ describe('WebSocketProvider', () => {
 
     await provider.callApi('test prompt');
 
-    expect(WebSocket).toHaveBeenCalledWith('ws://test.com', { headers });
+    expect(WebSocket).toHaveBeenCalledWith('ws://test.com', ['json', 'v1'], {
+      headers: { Authorization: 'Bearer test-token' },
+    });
   });
 
   it('should work without headers provided', async () => {


jameshiester-oai (commented):
Reviewed base `3664715e9c57be705656280b105233accf5ab76a` at head `bbe1cf198481fadb2c9c16ff730d7ffd2ea51a13`.

**Scope:** fresh full review. The recorded base is also the merge base; there was no earlier authenticated marker for this head.

## Verdict

No additional distinct actionable defect was found beyond the existing unresolved feedback. The intended negotiation path works end to end: a real `ws` 8.21.0 server selected `json`, Promptfoo preserved the unrelated authorization/custom headers, and the built/local CLI eval passed. The exact-head check rollup is green (45 checks).

## Verified findings

- **P0/P1:** none.
- **P2 — existing unresolved compatibility finding, confirmed and broadened:** [the current thread](https://github.com/promptfoo/promptfoo/pull/9949#discussion_r3516690620) correctly identifies that automatic promotion breaks non-protocol uses of the reserved header. The smallest suggested fix in that thread—only promoting syntactically valid tokens—is not sufficient. A raw `Sec-WebSocket-Protocol: token123` client still opened against a server that consumed the header but deliberately selected no subprotocol; this head promoted the same valid token and then failed with `Server sent no subprotocol`. The invalid `Bearer token-with-space` control failed before any Upgrade. Preserving compatibility therefore requires an explicit promotion/compatibility policy (for example, retaining raw-header behavior and requiring `config.protocols` for negotiation, or adding an explicit opt-in), not token syntax alone. I did not duplicate the inline thread.
- **P3:** no new defect. There is a meaningful test opportunity below.

## Existing feedback and test gaps

- [The duplicate-casing thread](https://github.com/promptfoo/promptfoo/pull/9949#discussion_r3516686088) is correct that the helper removes only the first case-insensitive match. However, an exact-head loopback with both casing variants completed successfully under the pinned `ws`/Node implementation; `ws` overwrote/canonicalized the requested protocol header before the Upgrade, so I could not reproduce the claimed duplicate-header failure. Removing every casing variant would still make the invariant explicit and less implementation-dependent.
- [The missing edge-test thread](https://github.com/promptfoo/promptfoo/pull/9949#discussion_r3516686108) remains accurate. The changed test mocks `ws` and verifies constructor arguments, but never exercises the reported runtime failure (`Server sent a subprotocol but none was requested`). A small unmocked `WebSocketServer` integration test should cover successful selection, no-selection behavior, explicit-`protocols` precedence, case-insensitive header names, and preservation of unrelated headers.
- The connection-option helper is cohesive and there is no material simplification opportunity beyond resolving the compatibility policy above.

## Validation

- `npx vitest run test/providers/websocket.test.ts --sequence.shuffle=false` — 23/23 passed.
- `npx vitest run test/providers/websocket.test.ts --sequence.seed=9949` — 23/23 passed.
- `npx @biomejs/biome check src/providers/websocket.ts test/providers/websocket.test.ts` — passed with the pre-existing `onmessage` complexity warning and no errors.
- `git diff --check 3664715e9c57be705656280b105233accf5ab76a...HEAD` — passed.
- `npm run build` — core, declarations, and production Web UI passed.
- Isolated `npm run local -- eval ... --no-cache --no-share` against a local server — 1/1 passed, score 1, zero errors; server observed selected protocol `json`, offered values `json,v1`, authorization `Bearer local-test`, and custom header `preserved`.
- Exact-source loopback probes — raw valid custom header opened; promoted valid custom header failed with `Server sent no subprotocol`; invalid token syntax failed before Upgrade; duplicate-case control completed successfully.
- Exact-head CI — SUCCESS across 45 visible contexts.

## Limitations

No external or TLS WebSocket service, proxy, browser, credential, or full backend suite was used. The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable in this environment (exit 127), so its correctness, tests, comments/docs, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialists.

mldangelo:
## Recommend closing — auto-promotion is unavoidably breaking, and the goal is already met by `config.protocols`

I dug into this and reproduced the compatibility problem that @jameshiester-oai, Codex, and Copilot all flagged. The short version: **promoting a legacy `Sec-WebSocket-Protocol` header into the `ws` `protocols` argument breaks every server that reads that header for a non-negotiation purpose (e.g. auth smuggling) and doesn't echo it back — and this happens for _valid_ subprotocol tokens, so token-syntax validation can't rescue it.**

Empirical repro against a real handshake server that completes the upgrade but omits `Sec-WebSocket-Protocol` from its 101 response:

```
no protocols (raw header path):     OPEN ok
promoted valid token [token123]:    ERROR  Server sent no subprotocol
```

And the space-containing auth value the deleted test documented (`Bearer token-with-space`) throws even earlier, at construction:

```
new WebSocket(url, ['Bearer token-with-space'])
  → SyntaxError: An invalid or duplicated subprotocol was specified
```

So this PR turns two previously-working config shapes into hard errors, and it removes the regression test (`'Sec-WebSocket-Protocol': 'Bearer token-with-space'`) that #9944 added specifically to lock that behavior in.

### Why closing is the right call rather than patching

The negotiation the PR wants is **already supported** — #9944 added `config.protocols`, which passes cleanly to the `ws` constructor. The rest of the stack converged on the non-breaking policy:
- **#9944**: preserves raw `Sec-WebSocket-Protocol` headers, adds `config.protocols` for negotiation.
- **#9947** (UI): "preserves raw Sec-WebSocket-Protocol headers and shows migration guidance," stores negotiated values in `config.protocols`.

#9949 is the only piece that auto-promotes-and-removes, which contradicts both and reintroduces the breakage they avoided. There's no version of "validate tokens first" that fixes it, because the failure mode (`Server sent no subprotocol`) triggers on valid tokens whenever the server doesn't negotiate.

If a nudge toward migration is still wanted, the safe increment is a one-line `logger.warn` when a legacy `Sec-WebSocket-Protocol` header is present and `config.protocols` is unset — **without** promoting or stripping it. Happy to open that as a small follow-up if you'd like, but I wouldn't land the promotion.

**#9947 does not depend on this PR** (it depends on #9944), so closing here doesn't block the UI work. Not merging.

mldangelo:
Closing per the analysis above: auto-promoting a legacy `Sec-WebSocket-Protocol` header into the `ws` `protocols` argument is unavoidably breaking — it makes `ws` require the server to negotiate a subprotocol back, so any endpoint that reads the header for auth (and doesn't echo it) fails with `Server sent no subprotocol`, even for syntactically valid tokens. The negotiation this PR is after is already supported via `config.protocols` (#9944), and the rest of the stack (#9944, #9947) deliberately preserves raw headers. If a migration nudge is still wanted, the safe version is a one-line `logger.warn` when a legacy header is present and `config.protocols` is unset — without promoting or stripping it. Happy to open that as a small follow-up if useful.

PR #9963, opened 2026-07-03
fix(providers): correct Codex SDK provider config handling

mldangelo:
## Summary

Fixes a config-handling bug in the OpenAI Codex SDK provider where a live target provider object attached to prompt config was passed through Nunjucks variable rendering.

Promptfoo attaches the live target `provider` object onto `prompt.config` for generic provider workflows. `OpenAICodexSDKProvider.callApi` merges `prompt.config` into its own config and then passes the merged object straight to `renderVarsInObject`. `renderVarsInObject` recurses into nested objects **and invokes any function-valued property** it encounters. Because a provider object exposes methods (`id()`, `callApi()`, ...) and can hold circular references, this caused variable rendering to:

- **call provider methods** during config rendering (unintended side effects), and
- for self-referential provider objects, **recurse until `RangeError: Maximum call stack size exceeded`**.

## Fix

Delete the injected `provider` key from the merged config before rendering:

```ts
delete mergedConfig.provider;
const config = renderVarsInObject(mergedConfig, context?.vars) as OpenAICodexSDKConfig;
```

The config type (`provider?: unknown`) and Zod schema (`provider: z.unknown().optional()`) already accept this key for loader compatibility — this change only prevents runtime var rendering from recursing into it.

## Test

Adds a regression test that attaches a self-referential provider object (with a function-valued `id()` method and a circular `self` reference) to `prompt.config` and asserts `callApi` resolves normally and forwards only the templated config to the SDK. Verified load-bearing: without the `delete`, the new test fails with `RangeError: Maximum call stack size exceeded`; with it, all 157 tests in the file pass.

## Provenance

Extracted from #8871 (agentic runtime redteam plugins/docs/examples) as an independent, low-risk provider fix that can land ahead of that PR's larger plugin work. This PR contains **only** the `codex-sdk.ts` config-handling hunk and its matching test — none of the agentic-runtime plugin/parser/docs code.

Refs #8871.

### DIFF
--- a/test/providers/openai-codex-sdk.test.ts
+++ b/test/providers/openai-codex-sdk.test.ts
@@ -2516,6 +2516,35 @@ describe('OpenAICodexSDKProvider', () => {
         );
       });
 
+      it('should ignore attached live provider objects before rendering prompt config vars', async () => {
+        mockRun.mockResolvedValue(createMockResponse('Response'));
+
+        const attachedProvider: Record<string, unknown> = {
+          id: () => 'attached-provider',
+        };
+        attachedProvider.self = attachedProvider;
+
+        const provider = new OpenAICodexSDKProvider({
+          env: { OPENAI_API_KEY: 'test-api-key' },
+        });
+
+        await expect(
+          provider.callApi('Test prompt', {
+            prompt: {
+              config: {
+                model: '{{ model }}',
+                provider: attachedProvider,
+              } as any,
+            },
+            vars: { model: 'gpt-5.2' },
+          } as any),
+        ).resolves.toMatchObject({
+          output: 'Response',
+        });
+
+        expect(mockRun).toHaveBeenCalledWith('Test prompt', {});
+      });
+
       it('should use CODEX_API_KEY from env if available', async () => {
         mockRun.mockResolvedValue(createMockResponse('Response'));
 
--- a/src/providers/openai/codex-sdk.ts
+++ b/src/providers/openai/codex-sdk.ts
@@ -2003,6 +2003,10 @@ export class OpenAICodexSDKProvider implements ApiProvider {
       ...this.config,
       ...context?.prompt?.config,
     };
+    // Promptfoo may attach the live target provider object to prompt config for
+    // generic provider workflows. Codex accepts this key for loader compatibility,
+    // but runtime variable rendering must not recurse into provider methods.
+    delete mergedConfig.provider;
     const config = renderVarsInObject(mergedConfig, context?.vars) as OpenAICodexSDKConfig;
 
     const requestedModel =


test/providers/openai-codex-sdk.test.ts
```diff
@@ -2516,6 +2516,35 @@ describe('OpenAICodexSDKProvider', () => {
         );
       });
 
+      it('should ignore attached live provider objects before rendering prompt config vars', async () => {
+        mockRun.mockResolvedValue(createMockResponse('Response'));
+
+        const attachedProvider: Record<string, unknown> = {
+          id: () => 'attached-provider',
```
jameshiester-oai, line 2523: [P2] Assert that rendering never invokes the attached provider method

This fixture covers the circular-reference failure because any traversal currently overflows, but it never verifies the separate side-effect regression described by the PR. A future cycle-aware traversal could still call `id()` and this test would pass. Make `id` a `vi.fn(() => 'attached-provider')` and assert it was not called after `callApi`, while retaining the self-cycle check.

jameshiester-oai (commented):
Reviewed base `0ef67d937e6868eabbcbd0d215d4530514e66fd0` against head `6c5a8321a5aff2567d69ee8ec849578e013d5c8f`.

Scope: fresh full `base...head` review; no authenticated marker existed for this head. I read the PR intent, complete diff, surrounding config parsing/call paths, changed and adjacent tests, existing feedback, and current CI.

Verdict: the production fix is appropriately narrow and correctly removes compatibility-only live `provider` metadata from a fresh merged object before recursive variable rendering. It preserves supported Codex config precedence and does not mutate caller config. One P2 regression-test gap remains.

CI: all 44 reported checks are successful, skipped, or neutral; no failing checks.

## Verified findings

### P2

- The new regression test proves the circular-reference stack overflow is gone, but it does not verify the other behavior promised by the fix: live-provider methods must not be invoked while config is rendered. The inline comment contains the smallest correction.

No P0/P1 production defects were found.

## Existing feedback

The unresolved Copilot thread at https://github.com/promptfoo/promptfoo/pull/9963#discussion_r3522241044 is valid and not duplicated: the test supplies `model: '{{ model }}'` but only asserts `mockRun`; asserting `mockStartThread` received `model: 'gpt-5.2'` would make the templating assertion load-bearing. Adjacent existing coverage and a direct runtime probe confirm the current implementation forwards the rendered model.

## Adjacent pre-existing limitations

These are not introduced or worsened by this diff, so I did not raise them as inline findings:

- `src/evaluator.ts:659-667` merges generic test options into prompt config, and `src/util/render.ts:135-137` invokes function-valued entries during rendering. A supported function-valued `options.transform` can therefore run before the Codex SDK call; a direct provider-path probe reproduced that behavior. This merits a separate follow-up rather than expanding this provider-metadata fix.
- `src/providers/openai/codex-app-server.ts:1171-1175` retains the analogous pre-existing live-provider traversal. This PR is explicitly scoped to the Codex SDK provider.

The four-line production correction is simpler and safer than changing the generic renderer or attempting cycle detection here. Besides the P2 inline assertion, the existing model-forwarding assertion is the only material test-strengthening opportunity found.

## Validation

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/openai-codex-sdk.test.ts --sequence.shuffle=false` — exit 0; 157/157 passed.
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/openai-codex-sdk.test.ts --sequence.seed=9963` — exit 0; 157/157 passed.
- Biome on both changed files plus `git diff --check` — exit 0.
- Credential-free direct `callApi` probes with an injected fake Codex SDK — exit 0; attached-provider method calls stayed at 0, prompt model rendered to `gpt-5.2`, malformed config returned the documented provider error, and no SDK call occurred on failure.
- `npm run local -- eval -c /private/tmp/pr9963-codex-sdk-e2e.ts --no-cache -o /private/tmp/pr9963-codex-sdk-result.json` in an isolated credential-free environment — exit 0; 1/1 passed; exported result had `success=true`, `score=1`, `error=null`, rendered model `gpt-5.2`, and `providerMethodCalls=0`.

## Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command was invoked but is not installed in this environment (exit 127); I used independent correctness, tests/QA, risk, comments/docs, error-handling, type-design, and simplification passes instead. No live external Codex service or credentials were used.

PR #9964, opened 2026-07-03
refactor(redteam): extract reusable generatePrompts() primitive

mldangelo:
## Summary

Behavior-preserving refactor of `RedteamPluginBase`, extracted from #9401
(semantic generation portfolios) as the compile-time prerequisite that the
portfolio core builds on. Landing it on its own de-risks and shrinks #9401
without changing what any plugin emits.

## What changed

- **`generatePrompts()` primitive.** `generateTests()` now delegates to a new
  reusable `protected generatePrompts()` that returns `GeneratedPrompt[]`, and
  `promptsToTestCases()` materializes those into `TestCase[]`. This lets
  portfolio-style generators reuse the batching/retry/parse primitive while
  applying their own planning and selection before materialization.
- **Helper extraction.** The inline inference-refusal check and prompt-length
  validation move into `assertNotRefusal()` and `validatePromptLengths()`. The
  closure is now simple enough that the previous
  `noExcessiveCognitiveComplexity` biome suppression is no longer needed and was
  removed.
- **`GeneratedPrompt` metadata.** `GeneratedPrompt` carries optional per-prompt
  `metadata`, which `promptsToTestCases()` merges into the generated test case's
  metadata (dormant until a producer sets it, e.g. the portfolio core).
- **Correctness fix.** The refusal-hint builder tested `if (context)` on a
  freshly-allocated object literal, which is always truthy, so the
  "user-configured values were included" hint was appended even when no
  `purpose`/`examples` were present. Now gated on
  `Object.keys(context).length > 0`.

## Scope

Only the `generatePrompts()` primitive, its helpers, the `GeneratedPrompt`
metadata plumbing, and the empty-context fix are included. The portfolio
generation extension point, selection wiring, and `src/redteam/generation/*`
code from #9401 are intentionally **not** part of this PR — this slice does not
import from `src/redteam/generation/*`.

## Testing

- `npx vitest run test/redteam/plugins/base.test.ts` — 121 passed (adds one
  focused test proving the `generatePrompts()` → `promptsToTestCases()`
  delegation surfaces per-prompt `metadata`).
- `npx vitest run test/redteam/plugins/` — 56 files / 881 tests passed, confirming
  the refactor does not change what plugins emit.
- `npm run tsc`, `npm run l`, `npm run f` — clean.

Refs #9401.

### DIFF
--- a/src/redteam/plugins/base.ts
+++ b/src/redteam/plugins/base.ts
@@ -35,6 +35,18 @@ import type {
 } from '../../types/index';
 import type { RedteamGradingContext } from '../grading/types';
 
+export type GeneratedPrompt = {
+  __prompt: string;
+  metadata?: Record<string, unknown>;
+};
+
+type ParsedGeneratedPrompt = { __prompt: string } | Record<string, string>;
+
+type PromptLengthValidationResult = {
+  acceptedPrompts: ParsedGeneratedPrompt[];
+  retryInstructions: string | undefined;
+};
+
 /**
  * Abstract base class for creating plugins that generate test cases.
  */
@@ -108,6 +120,20 @@ export abstract class RedteamPluginBase {
     delayMs: number = 0,
     templateGetter: () => Promise<string> = this.getTemplate.bind(this),
   ): Promise<TestCase[]> {
+    const prompts = await this.generatePrompts(n, delayMs, templateGetter);
+    return this.promptsToTestCases(prompts);
+  }
+
+  /**
+   * Generates raw prompt payloads without converting them into TestCase objects.
+   * Portfolio-style generators can reuse this primitive while applying their own
+   * planning and selection steps before materialization.
+   */
+  protected async generatePrompts(
+    n: number,
+    delayMs: number = 0,
+    templateGetter: () => Promise<string> = this.getTemplate.bind(this),
+  ): Promise<GeneratedPrompt[]> {
     logger.debug(`Generating ${n} test cases`);
     const batchSize = 20;
 
@@ -126,10 +152,9 @@ export abstract class RedteamPluginBase {
      * In multi-input mode, returns Record<string, string>[]
      */
     let retryInstructions: string | undefined;
-    // biome-ignore-start lint/complexity/noExcessiveCognitiveComplexity: Existing redteam generation flow handles batching, parsing, retries, and validation in one place.
     const generatePrompts = async (
-      currentPrompts: { __prompt: string }[] | Record<string, string>[],
-    ): Promise<{ __prompt: string }[] | Record<string, string>[]> => {
+      currentPrompts: ParsedGeneratedPrompt[],
+    ): Promise<ParsedGeneratedPrompt[]> => {
       const remainingCount = n - currentPrompts.length;
       const currentBatchSize = Math.min(remainingCount, batchSize);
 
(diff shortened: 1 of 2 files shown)


src/redteam/plugins/base.ts
```diff
@@ -108,6 +120,20 @@ export abstract class RedteamPluginBase {
     delayMs: number = 0,
     templateGetter: () => Promise<string> = this.getTemplate.bind(this),
   ): Promise<TestCase[]> {
+    const prompts = await this.generatePrompts(n, delayMs, templateGetter);
+    return this.promptsToTestCases(prompts);
```
jameshiester-oai, line 124: [P2] Preserve `GeneratedPrompt` metadata through existing materializer overrides

This new dispatch sends `GeneratedPrompt[]` through virtual `promptsToTestCases` implementations, but `AlignedHarmfulPlugin.promptsToTestCases` still destructures only `__prompt` and rebuilds metadata. An exact-head subclass probe that returned `{ metadata: { seededBy: 'portfolio' } }` produced only `harmCategory`, `pluginId`, and `pluginConfig`; the new metadata vanished. Update that override to accept `GeneratedPrompt[]` and merge per-prompt metadata before its canonical fields, then cover an aligned subclass in the regression test.

src/redteam/plugins/base.ts
```diff
@@ -108,6 +120,20 @@ export abstract class RedteamPluginBase {
...
+  }
+
+  /**
+   * Generates raw prompt payloads without converting them into TestCase objects.
+   * Portfolio-style generators can reuse this primitive while applying their own
+   * planning and selection steps before materialization.
+   */
+  protected async generatePrompts(
```
jameshiester-oai, line 132: [P2] Avoid breaking existing public subclasses with the new virtual hook name

`RedteamPluginBase` is exposed and documented as `redteam.Base.Plugin`. A pre-existing custom subclass is allowed to have its own `generatePrompts` helper today; after this addition, an incompatible helper now fails with `TS2416`, and JavaScript subclasses are unexpectedly dispatched here (a two-string probe failed in `promptsToTestCases` while a one-string result silently produced an undefined prompt). Please use a non-virtual/composed primitive or another compatibility-safe extension mechanism; otherwise this needs to be treated and documented as a breaking subclass API change rather than a behavior-preserving refactor.

src/redteam/plugins/base.ts
```diff
@@ -243,7 +215,78 @@ export abstract class RedteamPluginBase {
...
+    }
+
+    let message = `${this.provider.id()} returned a refusal during inference for ${this.constructor.name} test case generation.`;
+    // We don't know exactly why the prompt was refused, but we can provide hints to the user based on the values which were
+    // included in the context window during inference.
+    const context: Record<string, string> = {};
+    if (this.purpose) {
+      context.purpose = this.purpose;
+    }
+    if (this.config.examples) {
```
jameshiester-oai, line 241: [P3] Treat an empty examples list as empty refusal context

`examples: []` is valid config but remains truthy here, so with an empty purpose the new diagnostic still appends `User-configured values... {"examples":""}` even though no example value was included. Check `this.config.examples?.length` before adding the field, and add the currently missing regression cases for absent/empty context versus populated purpose/examples.

jameshiester-oai (commented):
Reviewed base `0ef67d937e6868eabbcbd0d215d4530514e66fd0` against head `861b388c96ae3c86bb8b12312b7772962658299e`.

Scope: fresh full `base...head` review; no authenticated marker existed for this head. I read the PR intent, complete diff, the linked #9401 portfolio consumer, existing subclasses/callers, changed and adjacent tests, all review threads, and current CI.

Verdict: the extracted batching/refusal/length/retry path is otherwise behavior-equivalent and the empty-context direction is correct, but the new subclass contract has two P2 compatibility/correctness defects and the refusal hint retains one P3 edge.

CI: all 44 reported checks are successful, skipped, or neutral; no failing checks. The redteam-scoped title follows repository convention.

## Verified findings

### P2

- The new `GeneratedPrompt.metadata` contract is silently lost through the existing aligned-harmful materializer override. The first inline comment includes a direct reproduction and correction direction.
- Adding and immediately dispatching through a common-name `generatePrompts` hook breaks existing consumers of the documented public `redteam.Base.Plugin` subclass API when they already use that helper name. The second inline comment includes compile-time and runtime reproductions.

### P3

- `examples: []` still makes the corrected refusal diagnostic claim user-configured values may have caused the refusal, producing `{"examples":""}`. The third inline comment covers the fix and missing regression case.

No P0/P1 defects were found.

## Existing feedback

Two unresolved threads already cover the distinct P2 where prompt metadata overwrites canonical `pluginId`/`pluginConfig`; I reproduced it but did not duplicate it: https://github.com/promptfoo/promptfoo/pull/9964#discussion_r3522250850 and https://github.com/promptfoo/promptfoo/pull/9964#discussion_r3522251032.

The outdated thread at https://github.com/promptfoo/promptfoo/pull/9964#discussion_r3522250843 remains technically correct: the multi-input formatter returns `{ __prompt: string }` wrappers rather than raw records. The grammar-only thread is style-only.

The private helper extraction materially reduces cognitive complexity without changing batching, delay, parsing, deduplication, sampling, retry instructions, ordering, materialization, or assertion behavior. Beyond the three inline corrections and existing metadata-precedence threads, I found no sound simplification or behavioral test gap.

## Validation

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/redteam/plugins/base.test.ts --sequence.shuffle=false` — exit 0; 121/121 passed.
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/redteam/plugins/base.test.ts --sequence.seed=9964` — exit 0; 121/121 passed.
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/redteam/plugins/ --sequence.seed=9964` — exit 0; 56 files / 881 tests passed.
- `npx biome check src/redteam/plugins/base.ts test/redteam/plugins/base.test.ts` and `git diff --check` — exit 0.
- `npx tsx /private/tmp/pr9964-probe.ts` — reproduced aligned metadata loss, canonical metadata overwrite, the empty-array diagnostic, and a legacy JavaScript subclass collision ending in `localeCompare` on `undefined`.
- `npx tsc --ignoreConfig ... /private/tmp/pr9964-type-collision.ts` — emitted the expected `TS2416`: the existing `() => Promise<string[]>` helper is no longer assignable to the new base member.
- Credential-free `npm run local -- redteam generate -c /private/tmp/pr9964-redteam-config.yaml -o /private/tmp/pr9964-redteam.yaml --force --no-cache` in an isolated config directory — exit 0; generated one contracts case plus one base64 case, both with the correct plugin ID and assertion. An initial run using the repository smoke fixture directly exited 1 because its relative provider path is designed for the fixture directory; switching the QA config to the same provider's absolute path resolved that harness issue.

## Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command was invoked but is not installed in this environment (exit 127); I used independent correctness, tests/QA, risk, comments/docs, error-handling, type-design, and simplification passes instead. The linked portfolio implementation is not part of this head, and no live model/provider credentials were used.

PR #9966, opened 2026-07-04
test: isolate hoisted mock state in 7 exposed suites

mldangelo:
## Summary

This is the **additive mock-reset hardening** slice extracted from the monolith test-hygiene PR #9798, landable on its own.

Each of these 7 suites shares hoisted mock defaults (defined once via `vi.hoisted(...)`) across tests. Because `vi.clearAllMocks()` only clears call history — it does **not** restore a mock's implementation or return value — a test that reassigns a hoisted mock (e.g. `mockResolvedValueOnce`, or replacing an implementation) can leak that state into later tests, making them order-dependent. This change makes each suite's `beforeEach` explicitly `.mockReset()` and re-apply the intended default implementation/return value for the shared hoisted bindings it depends on, restoring per-test isolation.

## Changes

Pure `beforeEach` hardening in 7 exposed test files — no production code, no new test infrastructure:

- `test/commands/mcp/server.test.ts` — reset + re-implement the StreamableHTTP transport constructor mock
- `test/logger.test.ts` — reset + re-implement the winston `transports.File` mock
- `test/providers/mcp/client.test.ts` — reset the hoisted MCP client/transport mocks (including the previously-unreferenced SSE transport binding) and re-apply their defaults
- `test/providers/mcp/provider.test.ts` — `.mockReset()` before re-applying `initialize`/`getAllTools`/`cleanup` defaults
- `test/providers/pythonCompletion.cliState.test.ts` — reset `getWorkerCount`/`shutdown` on the shared pool instance
- `test/redteam/providers/crescendo/index.test.ts` — reset + re-implement `applyRuntimeTransforms`
- `test/redteam/providers/hydra/index.test.ts` — reset + re-implement `applyRuntimeTransforms`

## Why split it out

This carries no production-code changes and no open review threads, so it can land independently of #9798's deferred reset-interpreter work. Merging it now shrinks the blast radius of #9798.

## Verification

- `npx vitest run` across all 7 suites: **7 files passed, 267 tests passed**
- `npm run tsc`: clean
- `npm run l` / `npm run f`: no fixes applied

Extracted from #9798.

### DIFF
--- a/test/commands/mcp/server.test.ts
+++ b/test/commands/mcp/server.test.ts
@@ -160,7 +160,13 @@ describe('MCP Server', () => {
     mcpServerMocks.MockMcpServer.mockReset().mockImplementation(
       mcpServerMocks.mockMcpServerImplementation,
     );
-    streamableHttpMocks.constructor.mockClear();
+    streamableHttpMocks.constructor
+      .mockReset()
+      .mockImplementation(function MockStreamableHTTPServerTransport() {
+        return {
+          handleRequest: vi.fn(),
+        };
+      });
     mockRandomUUID.mockClear();
     mcpServerCalls.length = 0;
   });
--- a/test/logger.test.ts
+++ b/test/logger.test.ts
@@ -134,6 +134,11 @@ describe('logger', () => {
     mockGetEnvBool.mockImplementation((_, defaultValue) => defaultValue);
     mockGetConfigDirectoryPath.mockReset();
     mockGetConfigDirectoryPath.mockReturnValue('/mock/config');
+    winstonMock.transports.File.mockReset().mockImplementation(function (this: {
+      write: ReturnType<typeof vi.fn>;
+    }) {
+      this.write = vi.fn();
+    });
     for (const fn of Object.values(fsMock)) {
       (fn as Mock).mockReset();
     }
--- a/test/providers/mcp/client.test.ts
+++ b/test/providers/mcp/client.test.ts
@@ -85,7 +85,7 @@ const mcpMocks = vi.hoisted(() => {
   };
 });
 
-const { mockClient, mockStdioTransport, mockStreamableHTTPTransport } = mcpMocks;
+const { mockClient, mockSSETransport, mockStdioTransport, mockStreamableHTTPTransport } = mcpMocks;
 
 // Mock the modules before importing them
 vi.mock('@modelcontextprotocol/sdk/client/index.js', async (importOriginal) => {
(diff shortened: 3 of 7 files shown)


jameshiester-oai (commented):
Reviewed base `0ef67d937e6868eabbcbd0d215d4530514e66fd0` against head `78a26438123c4741c83eead8f851d7b13f9f1f3c`.

Scope: fresh full `base...head` review; no authenticated marker existed for this head. I read all seven changed suites, their hoisted factories and mutation sites, the corresponding production mock contracts, current and linked #9798 feedback, and current CI.

Verdict: the server, logger, MCP provider, Python pool, and shared-instance method reset ordering is sound, and the reapplied defaults match their prior hoisted behavior. I found no distinct new changed-line defect beyond the three current unresolved comments, so I am not adding duplicate inline feedback.

CI: all 44 reported checks are successful, skipped, or neutral; no failing checks. The cross-cutting `test:` title is appropriate; incidental redteam test files do not require a `(redteam)` scope.

## Verified findings

No new P0–P3 changed-line finding was identified.

## Existing feedback

- https://github.com/promptfoo/promptfoo/pull/9966#discussion_r3522258429 and https://github.com/promptfoo/promptfoo/pull/9966#discussion_r3522258434 are technically valid: the Crescendo/Hydra defaults accept an object and return `transformedPrompt`, while production calls `applyRuntimeTransforms` positionally and consumes `prompt`/`originalPrompt`. Media-specific tests override the default, masking the text-only contract mismatch. This PR restores the pre-existing factory shape, so I did not duplicate it.
- https://github.com/promptfoo/promptfoo/pull/9966#discussion_r3522258437 is valid: MCP client tests queue `Client.mockImplementationOnce` values, but the new `beforeEach` resets only the returned singleton's methods. `vi.clearAllMocks()` does not clear an unconsumed constructor queue.
- Off-diff P3 hardening opportunity: `test/redteam/providers/hydra/index.test.ts:162` calls `mockIsBasicRefusal.mockReturnValue(false)` without `mockReset()`, while line 742 queues two one-shot values. Current control flow consumes both and no failure reproduced, but resetting it would make this suite's isolation claim complete.

The explicit resets are repetitive but remain clearer than introducing another shared abstraction across unrelated suites. A small local `resetMcpMocks` helper could reduce future omissions in the MCP client suite, but it is optional.

## Validation

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run <seven changed files> --sequence.shuffle=false` — exit 0; 7 files / 267 tests passed.
- The same seven files with `--sequence.seed=9966` — exit 0; 267/267 passed.
- Five additional full-suite seeds (`99660` through `99664`) — exit 0 for every run; 1,335/1,335 passed. Combined local head stress: 1,869 test executions, all passing.
- Independent specialist runs covered additional seeds, including MCP-client-only constructor-heavy orderings; all passed.
- A disposable exact-base archive ran the same seven suites deterministically and across twelve shuffled seeds; all 267 tests passed each time, confirming the patch is prophylactic rather than fixing a currently reproducible order failure.
- Biome on all seven changed files plus `git diff --check` — exit 0.

## Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command was invoked but is not installed in this environment (exit 127); I used independent correctness, tests/QA, risk, comments/docs, error-handling, type-design, and simplification passes instead. Product end-to-end QA is not applicable because this PR changes only test reset state; no production path, transport, process, cache, or database behavior changed.

PR #9967, opened 2026-07-04
feat(redteam): account for generation-phase token usage

mldangelo:
## Summary

Extracts the **generation-phase token-usage accounting (telemetry)** slice from the monolithic PR #9401 (`mdangelo/codex/semantic-generation-core-20260523`). This lands the token-accounting feature on its own so it can ship independently of that PR's semantic-frontier / portfolio "generation core", which stays deferred.

This PR was recreated fresh off `origin/main` (the source branch is conflicting), carving only the token-accounting hunks — it does **not** import or depend on `src/redteam/generation/*` (portfolio, predicate signatures, selection, frontier diagnostics), and none of the semantic-frontier report/metadata wiring is included.

## What it does

During redteam generation, provider calls made to synthesize attacks now have their token usage accumulated and surfaced:

- **`src/redteam/index.ts`** — adds `trackGenerationTokenUsage`, a `Proxy` around the generation provider that accumulates `tokenUsage` from every `callApi` response (preserving custom receivers, including class instances with private fields). `synthesize` wraps its generation provider with it, threads a `wrapGenerationProvider` callback through `applyStrategies` so strategies that load their own local provider stay in the totals, and returns the aggregated `generationTokenUsage`.
- **`src/redteam/strategies/mathPrompt.ts`, `src/redteam/strategies/multilingual.ts`** — consume the threaded `__wrapGenerationProvider` so their locally-loaded generation providers are also metered.
- **`src/redteam/commands/generate.ts`** — accumulates the usage across single- and multi-context runs, logs an "Observed generation token usage…" summary, and persists it under `metadata.generationTokenUsage` in the generated/updated config (clearing any stale value first).

## Tests

- `test/redteam/index.test.ts` — aggregation across metered/unmetered calls; custom provider receivers preserved while tracking.
- `test/redteam/commands/generate.test.ts` — usage persisted to output metadata; stale usage removed on regenerate/update.
- `test/redteam/strategies/{mathPrompt,multilingual}.test.ts` — locally-loaded providers routed through the usage wrapper.

## Verification

- `npm run tsc` clean; `npm run l && npm run f` clean (the two `noExcessiveCognitiveComplexity` warnings on `multilingual.ts` are pre-existing on `main`).
- Focused tests pass: `test/redteam/index.test.ts` and the strategy tests run directly; `test/redteam/commands/generate.test.ts` passes once its unrelated, pre-existing `js-yaml` / `@apidevtools/json-schema-ref-parser` default-export module-load error (present on `main` too) is worked around locally.

Relates to #9401 (that PR's portfolio/semantic-frontier core remains deferred).


mldangelo:
Rebaselined the two cross-layer architecture edges that this slice legitimately widens (commit `b2ee743`):

- `redteam -> legacy-contracts`: 149 → 150
- `redteam -> node`: 211 → 213

These are the additive token-usage type/helper references introduced by the slice — `redteam` now imports `accumulateTokenUsage` / `accumulateResponseTokenUsage` from `src/util/tokenUsageUtils` and the `TokenUsage` type from `src/types/shared`. No import was removed and no unrelated edges were touched: I hand-edited the baseline to bump only these two edges rather than running a full `architecture:baseline`, so a pre-existing `node -> legacy-runtime` drift on `main` (unrelated to this PR) is intentionally left alone. `npx tsx scripts/checkArchitectureBoundaries.ts` now exits 0 ("Architecture boundaries passed").

jameshiester-oai (commented):
## Automated review — COMMENT

Reviewed base: `0ef67d937e6868eabbcbd0d215d4530514e66fd0` (`main`)

Reviewed head: `b2ee74390d0274c66ce230f756e33e9f9bff7f4f`

Scope: fresh full review of `base...head`; no authenticated completion marker exists for this head. The second commit only adjusts the architecture baseline, so the full feature diff was reviewed.

Verdict: the happy path, multi-context sum, strategy-local provider wrapping, stale-value cleanup, and architecture deltas work, but four P2 defects make the new telemetry incomplete or break a valid custom-provider shape. CI is fully green; the PR is mergeable but behind `main`. Codecov reports 76.79% patch coverage with 13 changed lines uncovered. The `feat(redteam)` title follows the required scope rule.

### Verified findings

No P0/P1 findings.

P2 (inline details and exact-head reproductions attached):

1. The Proxy violates ECMAScript invariants for a non-configurable, read-only `callApi` property, causing hardened custom providers to fail generation.
2. Request counting is response-shape dependent: a live unmetered response with explicit zero usage and a thrown call disappear, while a cached response that omits `numRequests` is counted as a live request.
3. Cache-only usage with real `cached`/`total` tokens and `numRequests: 0` is neither logged nor persisted; update paths also delete the previous value.
4. The only usage summary executes after strict-failure handling and the zero-test early return, so metered failed generation reports no cost/request information.

P3 documentation gap:

- `site/docs/red-team/configuration.md` documents generated `metadata.configHash` but not the new `metadata.generationTokenUsage` object, including cache semantics, per-run versus cumulative meaning, and failure behavior. This is new user-facing persisted output and should be documented.

- Restore the `Plugins.find` spy in `finally`: https://github.com/promptfoo/promptfoo/pull/9967#discussion_r3522279475
- Mock `cli-progress` in the new multilingual test: https://github.com/promptfoo/promptfoo/pull/9967#discussion_r3522279480

- Replace the provider-target Proxy plus magic `__wrapGenerationProvider` callback with one explicit `GenerationUsageTracker`/façade that records invocation, cache, success/error, and token totals under one contract. That would avoid Proxy invariants and stop request semantics from depending on each provider's `tokenUsage` shape.
- Add regressions for non-configurable provider methods, cache hits both with and without explicit `numRequests: 0`, successful live empty usage, thrown calls, zero-test and strict failures, and the default unaligned harmful-provider path, which constructs a separate provider outside this tracker.

### Tests and QA

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/redteam/index.test.ts test/redteam/commands/generate.test.ts test/redteam/strategies/mathPrompt.test.ts test/redteam/strategies/multilingual.test.ts --sequence.shuffle --sequence.seed=9967` — exit 0, 241/241 passed.
- `npm run tsc` — exit 0.
- `npm run architecture:check` — exit 0; the two changed redteam edges match the new imports. It also reported one pre-existing `node -> legacy-runtime` improvement below baseline.
- Biome check over the eight changed TS/test files — exit 0 with existing complexity warnings only.
- Real isolated `redteam generate --force --no-cache` with the repository's local class provider — exit 0; two generated tests persisted `100 total / 60 input / 40 output / 2 requests`.
- Offline two-context generation — exit 0 and correctly summed `24 total / 16 input / 8 output / 2 requests`.
- Cached-only and live-empty local providers both generated valid tests, but neither output contained `metadata.generationTokenUsage`; the cached run reported 50 cached/total tokens internally.
- A two-plugin local run invoked `callApi` twice (first threw, second succeeded) but persisted/logged one request.
- A metered malformed response produced zero tests and returned without an `Observed generation...` summary.
- A custom provider with an extensible instance but non-configurable/read-only `callApi` reached the new Proxy and failed with its invariant `TypeError`, producing zero tests.
- CI: all build, unit, integration, smoke, lint, CodeQL, security, and coverage checks passed.

### Limitations

- `/pr-review-toolkit:review-pr all parallel` is unavailable in this environment (exit 127), so correctness, tests/QA, error handling, types, docs, and simplification were covered by three independent read-only specialist passes plus orchestrator validation.
- No paid provider or remote-generation endpoint was called. CLI QA used isolated `PROMPTFOO_CONFIG_DIR` directories, deterministic local file providers, `--no-cache`, and `PROMPTFOO_DISABLE_REMOTE_GENERATION=true`; no secrets or user cache/database were used.

src/redteam/index.ts
```diff
@@ -73,6 +74,27 @@ import type {
...
+function trackGenerationTokenUsage(provider: ApiProvider, tokenUsage: TokenUsage): ApiProvider {
+  const callApi = provider.callApi.bind(provider);
+  const trackedCallApi: ApiProvider['callApi'] = async (...args) => {
+    const response = await callApi(...args);
+    accumulateResponseTokenUsage(tokenUsage, response);
+    return response;
+  };
+  trackedCallApi.label = provider.callApi.label;
+
+  return new Proxy(provider, {
```
jameshiester-oai, line 86: [P2] Avoid proxying immutable provider properties

This Proxy may return a different value for `callApi`, but ECMAScript requires a proxy to return the exact descriptor value when the target has a non-configurable, non-writable data property. I exercised the real CLI with an otherwise extensible custom provider whose `callApi` was defined that way (so provider loading still succeeded); the contracts plugin then failed with `TypeError: 'get' on proxy: property 'callApi' is a read-only and non-configurable data property ... but the proxy did not return its actual value`, producing zero tests. Hardened/custom providers are a valid `ApiProvider` shape, so this is a compatibility regression. Use an explicit façade (or a proxy with a neutral target) that delegates/binds to the provider rather than proxying the provider object itself, and add object/class-field descriptor regressions.

src/redteam/index.ts
```diff
@@ -73,6 +74,27 @@ import type {
 
 const MATERIALIZED_MULTI_INPUT_PROMPT_METADATA_KEY = '__promptfooMaterializedMultiInputPrompt';
 
+function trackGenerationTokenUsage(provider: ApiProvider, tokenUsage: TokenUsage): ApiProvider {
+  const callApi = provider.callApi.bind(provider);
+  const trackedCallApi: ApiProvider['callApi'] = async (...args) => {
+    const response = await callApi(...args);
```
jameshiester-oai, line 80: [P2] Normalize request counting at this wrapper boundary

Accumulating only after a resolved response and delegating request semantics to `response.tokenUsage` makes the new `numRequests` provider-dependent. Exact-head CLI probes showed: (1) a live unmetered response with explicit zero usage produced no request/metadata (the supported Anthropic legacy provider returns this shape); (2) two `callApi` invocations where the first threw and the second succeeded were reported as one request; and (3) an OpenAI-style cached response that omits `numRequests` is incremented as a live request even though that provider explicitly says cache hits are not new requests. Track logical invocations/errors and confirmed cache hits explicitly, and aggregate response token fields with request counting disabled; add all three regressions.

src/redteam/commands/generate.ts
```diff
@@ -827,6 +838,22 @@ async function doGenerateRedteamInternal(
       sharing: config.sharing,
       ...(contexts && contexts.length > 0 ? { contexts } : {}),
     };
+    const generationRequestCount = generationTokenUsage.numRequests ?? 0;
+    if (generationRequestCount > 0) {
```
jameshiester-oai, line 842: [P2] Preserve cache-only token usage when request count is zero

`numRequests > 0` is not a valid presence check for token usage. Google AI Studio and other providers intentionally return cache hits as `{ cached, total, numRequests: 0 }`. In an actual isolated CLI run, a valid generated test carried 50 cached/total tokens internally, but this gate logged nothing and all three persistence branches omitted `generationTokenUsage`; update branches first delete any stale value, so they actively replace it with nothing. Gate on any observed token/detail field (or a dedicated `hasUsage` flag) and persist cache-only usage with zero live requests.

src/redteam/commands/generate.ts
```diff
@@ -827,6 +838,22 @@ async function doGenerateRedteamInternal(
       sharing: config.sharing,
       ...(contexts && contexts.length > 0 ? { contexts } : {}),
     };
+    const generationRequestCount = generationTokenUsage.numRequests ?? 0;
```
jameshiester-oai, line 841: [P2] Report consumed usage before failure exits

This new summary is reached only after `handleFailedPlugins` and the `redteamTests.length === 0` return above. A local provider that returned 50 metered tokens but malformed prompt text made the plugin produce zero tests; the command returned successfully with warnings but no `Observed generation...` line. Strict partial generation throws even earlier, and a later failing context can discard completed-context totals. Persisting metadata can remain success-only, but emit the captured usage from a failure-safe path before these exits (or in structured error/finally handling) and cover zero-test, strict-partial, and later-context failure cases.

PR #9969, opened 2026-07-04
feat(util): add token-usage helpers (error/generation accumulation)

mldangelo:
## Summary

Adds three side-effect-free helpers/types to the token-usage utilities. These are additive building blocks the token-accounting core will consume — they change no existing behavior on their own.

- **`getErrorTokenUsage(error)`** (`src/util/tokenUsageUtils.ts`): safely extracts validated `TokenUsage` carried by a thrown value, using `BaseTokenUsageSchema.safeParse`. Returns `undefined` for malformed payloads.
- **`accumulateGenerationTokenUsage(target, update)`**: folds generation-time provider tokens into evaluation totals without counting internal generation calls as target probes (drops `numRequests`/`assertions` from the update). Returns whether the payload contributed any token totals.
- **`NormalizedTokenUsage` type**: a fully-required token-usage shape where `completionDetails` and `assertions` are always present.
- **`TokenUsageTracker.trackResponseUsage()`** (`src/util/tokenUsage.ts`): tracks a single provider response through the shared response-aware request-counting contract (`accumulateResponseTokenUsage`), inferring one request when the response omits `numRequests`.

Includes focused unit tests for each addition.

## Context

Extracted from #9162 (`mdangelo/codex/fix-redteam-token-accounting-upstream-20260508`). This slice is **independent and landable ahead** of the entangled token-accounting core. It deliberately excludes the behavior-changing pieces of that PR — the `accumulateAssertionTokenUsage` request-count default flip and the `normalizeTokenUsage` deep-merge/return-type change — which remain with the core.

**Scope note:** additive helpers only; `behavior_change: false`.

## Verification

- `npx vitest run test/util/tokenUsageUtils.test.ts test/util/tokenUsage.test.ts` — 41 passed
- `npm run tsc` — clean
- `npm run l && npm run f` — clean


src/util/tokenUsageUtils.ts
```diff
@@ -1,4 +1,30 @@
...
+
+/**
+ * Safely extract token usage carried by a thrown value.
+ */
+export function getErrorTokenUsage(error: unknown): TokenUsage | undefined {
+  if (!error || typeof error !== 'object' || !('tokenUsage' in error)) {
+    return undefined;
+  }
+
+  const parsedTokenUsage = BaseTokenUsageSchema.safeParse(error.tokenUsage);
```
jameshiester-oai, line 25: **[P2] Keep error-usage extraction from throwing**

This helper accepts an `unknown` thrown value, but both the `in` check and this property read/parse can execute user-defined Proxy traps or getters. An exact-head probe with a throwing `tokenUsage` accessor (and another with a Proxy `has` trap) made `getErrorTokenUsage` throw the secondary exception. Linked #9162 calls this inside the central evaluator/provider catch paths without another guard, so a custom provider can have its original failure replaced and skip structured error-row construction. Guard the full presence check, single property read, and parse in `try/catch`, returning `undefined` when inspection itself fails; please add getter/Proxy regressions.

src/util/tokenUsage.ts
```diff
@@ -50,6 +54,23 @@ export class TokenUsageTracker {
...
+  public trackResponseUsage(
+    providerId: string,
+    response: { tokenUsage?: TokenUsage } | undefined,
+  ): void {
+    const current = this.providersMap.get(providerId) ?? createEmptyTokenUsage();
+    const updated = { ...current };
+    accumulateResponseTokenUsage(updated, response);
+    this.providersMap.set(providerId, updated);
+    logger.debug(
+      `Tracked response usage for ${sanitizeProviderIdForLog(providerId)}: total=${response?.tokenUsage?.total ?? 0}, cached=${response?.tokenUsage?.cached ?? 0}`,
```
jameshiester-oai, line 70: **[P2] Snapshot response usage before updating tracker state**

The method reads `response.tokenUsage` three times through `accumulateResponseTokenUsage` and twice again in this log. With a stateful accessor, the exact-head probe stored totals from a different snapshot; when the fifth read threw, `providersMap` was already updated and the tracking call still failed. Because linked #9162 invokes this after successful provider calls, instrumentation can convert success into failure while leaving partial global state. Read `tokenUsage` once before mutation, pass a stable wrapper to the accumulator, and log that same snapshot.

jameshiester-oai (commented):
## Review summary

**Reviewed base:** `0ef67d937e6868eabbcbd0d215d4530514e66fd0`  
**Reviewed head:** `e73716dcef1f341cec7e3555849a97f8c3e53564`  
**Scope:** fresh full `base...head` review, including intended consumers in linked PR #9162  
**Verdict:** the happy-path accounting semantics are sound, but two property-access boundaries can turn error/usage instrumentation into new failures. All 40 reported CI checks are green.

### Verified findings

- **P2:** `getErrorTokenUsage(error: unknown)` is not exception-safe. Proxy traps and throwing accessors escape the helper and can replace the original provider error inside #9162's catch paths.
- **P2:** `trackResponseUsage` reads `response.tokenUsage` five times. A stateful accessor can produce inconsistent totals; a later throwing read leaves the global tracker updated and then fails the otherwise successful call.
- **P3 — PR title scope:** this slice is explicitly part of the redteam generation-accounting feature and encodes generation-vs-target-probe semantics. Under the repository's mandatory redteam rule, the title should use `feat(redteam): ...` rather than `feat(util): ...`.

### Existing feedback checked

The three current Copilot threads remain valid and were not duplicated:

- an explicitly undefined completion-detail field makes `accumulateGenerationTokenUsage` report a contribution while adding only zeroes;
- the boolean documents “valid” but callers and the PR body use it as “contributed nonzero totals”;
- `trackResponseUsage(provider, undefined)` inserts a zero-only provider entry.

The linked evaluator also guards out responses without `tokenUsage` before calling the new response-aware tracker, so its primary per-provider path still bypasses request inference. That overlaps existing #9162 accounting feedback and is not repeated as a #9969 inline finding.

### Tests and QA

- Focused tests: **41/41 passed** deterministically and **41/41 passed** with seed 9969.
- Adjacent tracker/compat/summary run: **92/92 passed** with seed 9969.
- `npm run tsc` — passed.
- `npm run architecture:check` — passed; one pre-existing edge reduction remains available to ratchet.
- Changed-file Biome and `git diff --check` — passed.
- Direct TSX probes reproduced both new findings and all three existing threads, including a five-read stateful accessor that left `total: 2` stored before throwing.
- Exact-head CI: **40/40 passed**.

### Test gaps, simplification, and limitations

- Add getter/Proxy regressions for both new property boundaries, plus explicit undefined-response and undefined-completion-detail cases.
- Snapshot `response.tokenUsage` once and use that stable value for accumulation and logging; this is both simpler and deterministic.
- An unknown-only usage object parses to `{}`, but I did not classify that as a defect because the shared forward-compatible schema intentionally has optional fields and strips unknown keys.
- No CLI/eval end-to-end run was applicable because this isolated PR adds helpers with no current-head production caller. The linked evaluator, server, plugin, and provider consumers were inspected instead.
- The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable (exit 127); all six review lenses were performed manually and reconciled with three read-only specialists.

PR #9970, opened 2026-07-04
fix(providers): report cached request counts in token usage

mldangelo:
## Summary

Cached responses should still count as one logical request in token-usage accounting. Several providers were reporting `numRequests: 0` on the cache-hit path, which under-counts requests when results are aggregated. This sets `numRequests: 1` for cached responses so request counts stay accurate whether or not a result was served from cache.

Affected providers:

- `src/providers/google/ai.studio.ts` — `AIStudioChatProvider` (single + streaming) and `AIStudioEmbeddingProvider`
- `src/providers/google/gemini-image.ts` — `GeminiImageProvider`
- `src/providers/google/provider.ts` — `GoogleProvider`
- `src/providers/quiverai.ts` — `QuiverAiProvider` (drops the `cached ? 0 : 1` branch, always `1`)

## Context

This is a small, self-contained slice extracted from the larger token-accounting PR #9162 (currently CONFLICTING with `main`). These changes are **provider-local token-usage annotations only**: they set the existing optional `numRequests` field on cached token usage and do **not** depend on any `src/util/tokenUsageUtils` core behavior change or the assertion-accumulation work in that PR. They can land independently and ahead of the core changes.

## Tests

- Updated existing cached-response assertions in `test/providers/google/ai.studio.test.ts`, `test/providers/google/provider.test.ts`, and `test/providers/quiverai.test.ts` to expect `numRequests: 1`.
- Added a regression test in `test/providers/google/gemini-image.test.ts` asserting a cached image response reports `numRequests: 1`.

All touched provider test files pass (`npx vitest run` — 252 tests green); `tsc` clean; lint/format clean (only pre-existing complexity warnings on unrelated functions).


src/providers/google/provider.ts
```diff
@@ -605,7 +605,7 @@ export class GoogleProvider extends GoogleGenericProvider {
         ? {
             cached: lastData.usageMetadata?.totalTokenCount,
             total: lastData.usageMetadata?.totalTokenCount,
-            numRequests: 0,
+            numRequests: 1,
```
jameshiester-oai, line 608: **[P2] Do not charge disk-cache hits as target probes**

This field is persisted and summed by `getMonthlyRedteamProbeUsage()` into the 100,000-probe limit. That code and the public inference-limit docs define a probe as a request Promptfoo sends to the target, but `fetchWithCache` returns a disk hit without any upstream request. Setting cached responses to `1` therefore makes repeated cached evals consume allowance and report target probes that never occurred (the same change appears in the other three providers). Either retain explicit zero for true disk hits, or separate logical eval attempts from actual target requests and update the quota/docs coherently.

jameshiester-oai (commented):
## Review summary

**Reviewed base:** `0ef67d937e6868eabbcbd0d215d4530514e66fd0`  
**Reviewed head:** `90e527e645950fb76668d1c01a503466ba2dcb5f`  
**Scope:** fresh full `base...head` review, including cache, evaluator, persisted probe-limit, and linked #9162 interactions  
**Verdict:** normal cached success paths now aggregate consistently, but the change conflicts with the documented target-probe contract and leaves cached Google safety returns outside the corrected normalization. All 40 reported CI checks are green.

### Verified findings

- **P2 — disk-cache hits now consume target-probe allowance.** `fetchWithCache` returns a cache hit without contacting the target, while public docs and `src/util/redteamProbeLimit.ts:16-34` define/count a probe as a request sent to the target. Persisted cached responses from all four changed providers now carry `numRequests: 1`, so cached reruns consume the 100,000 monthly probe allowance and display probes despite zero target traffic. If `numRequests` is intentionally becoming “logical eval attempts,” the target-request quota needs a separate metric or explicit cache exclusion plus coordinated docs; under the current contract disk hits must remain zero.
- **P2 — cached Google safety returns bypass the new mapping.** `src/providers/google/provider.ts:509-542` and `:559-577` return before the changed normal branch. A real local cache miss→hit for `finishReason: SAFETY` produced a second response with `cached: true` but `tokenUsage: {total,prompt,completion}`—no `cached` classification and no request count. Current provider tracking ended at `numRequests: 0, cached: 0`; linked #9162 fixes only request inference, not token classification. The existing cached-safety test at `test/providers/google/provider.test.ts:696` does not assert usage.

### Existing feedback and design checks

- No inline review threads existed.
- Explicit counts are not double-counted by `accumulateResponseTokenUsage`; four cached responses with `numRequests: 1` aggregate to exactly four.
- Quiver multi-output remains one logical provider invocation and its streaming path is uncached.
- The title scope `fix(providers)` is appropriate: this slice is provider-local even though it supports the larger redteam accounting work.
- A shared cached/fresh usage builder for Google normal and guardrail returns would remove the duplicated branch logic that caused the second finding.

### Tests and QA

- Four changed provider suites: **252/252 passed** deterministically and **252/252 passed** with seed 9970.
- Probe-limit suite: **13/13 passed**; its SQL and public probe definition were inspected against the changed persisted response shape.
- `npm run tsc` — passed.
- Changed-file Biome — passed with three pre-existing complexity warnings; `git diff --check` passed.
- Credential-free real-cache success probe against a local fake Google endpoint: one upstream hit for two calls; miss+hit tracked `numRequests: 2`, `cached: 15`.
- Credential-free cached-safety probe: one upstream hit for two calls; cached response lacked cached-token/request fields and provider tracking remained `numRequests: 0`, `cached: 0`.
- Exact-head CI: **40/40 passed**.

### Test gaps and limitations

- Add real miss→hit coverage rather than only injecting `cached: true`, and assert the full token-usage shape for cached prompt/safety blocks.
- No live Google or Quiver request was made; provider transport and cache behavior were exercised against localhost with isolated cache directories.
- The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable (exit 127); all six review lenses were performed manually and reconciled with three read-only specialists.

PR #9972, opened 2026-07-04
fix(mcp): contain HTTP transport to loopback with host/CSRF protection

mldangelo:
## Summary

This extracts the **network-containment** slice from #8760 (MCP HTTP transport hardening) into a small, independently reviewable PR. It contains only the network boundary — nothing that touches per-tool workspace validation.

The MCP HTTP transport currently binds to all interfaces and accepts requests with any `Host`/`Origin`. That leaves it exposed to other machines on the network and to DNS-rebinding / CSRF attacks where a malicious web page drives a victim's local MCP server. This change contains the transport to the local machine:

- **Loopback binding** — the HTTP server now listens on `127.0.0.1` (`DEFAULT_MCP_HTTP_HOST`) instead of all interfaces, so it is not reachable from other hosts.
- **Host-header guard** (`mcpHostProtection`) — rejects any request whose `Host` header is not a loopback name (`127.0.0.1`, `localhost`, `[::1]`), returning `403`. This blocks DNS-rebinding attacks that target the server through an attacker-controlled hostname.
- **CSRF / cross-origin guard** — wires in the existing `src/server/middleware/csrfProtection` middleware so cross-site browser `POST`s are blocked.
- **Docs** — `site/docs/integrations/mcp-server.md` now documents that the HTTP transport is local-only (binds `127.0.0.1`, rejects non-local `Host`, blocks cross-site POSTs).

## Scope

This deliberately leaves behind #8760's **workspace-containment** part (the per-tool path/arg validators in `src/commands/mcp/lib/security.ts` and their per-tool wiring), which carries that PR's 21 open review threads and is being iterated on separately. None of those validators or tool changes are included here, so this slice can land independently.

## Architecture

`csrfProtection` already lives in `src/server/middleware/` (view-server layer) on `main`; no new middleware file is added. Importing it from the MCP command adds one `cli -> view-server` edge (an already-allowed dependency per `architecture/layers.json`), so `architecture/edge-baseline.json` is bumped `6 -> 7` for that single edge and nothing else. `npm run architecture:check` passes.

## Testing

- `test/commands/mcp/server.test.ts` — loopback binding, middleware ordering, transport handling, and a parameterized `mcpHostProtection` suite covering the allow-path (loopback `Host` variants incl. `[::1]`) and the reject-path (foreign hosts, DNS-rebinding suffixes like `127.0.0.1.attacker.example`, the `169.254.169.254` metadata address, and missing `Host`).
- `npm run tsc`, `npm run l`, `npm run f`, and `npm run architecture:check` all clean.

Refs #8760.

### DIFF
--- a/src/commands/mcp/server.ts
+++ b/src/commands/mcp/server.ts
@@ -2,6 +2,7 @@ import { randomUUID } from 'node:crypto';
 
 import express from 'express';
 import logger from '../../logger';
+import { csrfProtection } from '../../server/middleware/csrfProtection';
 import telemetry from '../../telemetry';
 import { isMissingPackageImportError } from '../../util/packageImportErrors';
 import { registerResources } from './resources';
@@ -18,6 +19,39 @@ import { registerRunEvaluationTool } from './tools/runEvaluation';
 import { registerShareEvaluationTool } from './tools/shareEvaluation';
 import { registerTestProviderTool } from './tools/testProvider';
 import { registerValidatePromptfooConfigTool } from './tools/validatePromptfooConfig';
+import type { NextFunction, Request, Response } from 'express';
+
+export const DEFAULT_MCP_HTTP_HOST = '127.0.0.1';
+const ALLOWED_MCP_HTTP_HOSTS = new Set(['127.0.0.1', 'localhost', '[::1]', '::1']);
+
+function getHostnameFromHostHeader(host: string): string | undefined {
+  const bracketedIpv6 = host.match(/^(\[[^\]]+\])(?::\d+)?$/);
+  if (bracketedIpv6) {
+    return bracketedIpv6[1].toLowerCase();
+  }
+
+  const hostnameWithOptionalPort = host.match(/^([^:@/]+)(?::\d+)?$/);
+  return hostnameWithOptionalPort?.[1].toLowerCase();
+}
+
+export function mcpHostProtection(req: Request, res: Response, next: NextFunction): void {
+  const hostname = req.headers.host ? getHostnameFromHostHeader(req.headers.host) : undefined;
+  if (hostname && ALLOWED_MCP_HTTP_HOSTS.has(hostname)) {
+    next();
+    return;
+  }
+
+  logger.warn('[MCP] Blocked request with non-local Host header', {
+    host: req.headers.host,
+    method: req.method,
+    path: req.path,
+  });
+  res.status(403).json({ error: 'MCP HTTP requests require a local Host header' });
+}
+
+function formatHttpHostForUrl(host: string): string {
+  return host.includes(':') && !host.startsWith('[') ? `[${host}]` : host;
+}
 
 function setMcpTransport(transport: 'http' | 'stdio'): void {
   Object.assign(process.env, { MCP_TRANSPORT: transport });
@@ -129,7 +163,9 @@ export async function startHttpMcpServer(port: number): Promise<void> {
   setMcpTransport('http');
 
   const app = express();
+  app.use(mcpHostProtection);
   app.use(express.json());
+  app.use(csrfProtection);
 
   const mcpServer = await createMcpServer();
 
@@ -159,14 +195,17 @@ export async function startHttpMcpServer(port: number): Promise<void> {
   // Return a Promise that only resolves when the server shuts down
   // This keeps long-running commands running until SIGINT/SIGTERM
   return new Promise<void>((resolve) => {
-    const httpServer = app.listen(port, () => {
-      logger.info(`Promptfoo MCP server running at http://localhost:${port}`);
-      logger.info(`MCP endpoint: http://localhost:${port}/mcp`);
-      logger.info(`SSE endpoint: http://localhost:${port}/mcp/sse`);
+    const host = DEFAULT_MCP_HTTP_HOST;
+    const urlHost = formatHttpHostForUrl(host);
+    const httpServer = app.listen(port, host, () => {
+      logger.info(`Promptfoo MCP server running at http://${urlHost}:${port}`);
+      logger.info(`MCP endpoint: http://${urlHost}:${port}/mcp`);
+      logger.info(`SSE endpoint: http://${urlHost}:${port}/mcp/sse`);
 
       // Track server start
       telemetry.record('feature_used', {
         feature: 'mcp_server_started',
+        host,
         transport: 'http',
         port,
       });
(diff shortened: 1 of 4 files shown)


src/commands/mcp/server.ts
```diff
@@ -159,14 +195,17 @@ export async function startHttpMcpServer(port: number): Promise<void> {
   // Return a Promise that only resolves when the server shuts down
   // This keeps long-running commands running until SIGINT/SIGTERM
   return new Promise<void>((resolve) => {
-    const httpServer = app.listen(port, () => {
-      logger.info(`Promptfoo MCP server running at http://localhost:${port}`);
-      logger.info(`MCP endpoint: http://localhost:${port}/mcp`);
-      logger.info(`SSE endpoint: http://localhost:${port}/mcp/sse`);
+    const host = DEFAULT_MCP_HTTP_HOST;
+    const urlHost = formatHttpHostForUrl(host);
+    const httpServer = app.listen(port, host, () => {
```
jameshiester-oai, line 200: **[P2] Treat the remote-HTTP removal as a breaking change**

Before this hunk, `listen(port)` bound all interfaces and the docs advertised remote integrations. Hardcoding `127.0.0.1` with no `--host` option makes Docker port publication, sidecars, remote clients, and IPv6-only localhost consumers unable to connect; the live probe confirmed both the non-loopback address and `::1` were unreachable. The safer default is appropriate, but shipping this as `fix(mcp)` silently removes a documented mode. Either retain an explicit host/allowed-host opt-in with clear security requirements, or mark this `fix(mcp)!:` and document the migration path.

test/commands/mcp/server.test.ts
```diff
@@ -284,7 +307,142 @@ describe('MCP Server', () => {
...
+      const { startHttpMcpServer } = await import('../../../src/commands/mcp/server');
+
+      const serverPromise = startHttpMcpServer(3100);
+      await new Promise((resolve) => setImmediate(resolve));
+      process.emit('SIGINT');
+      await serverPromise;
+
+      expect(expressMocks.app.use).toHaveBeenNthCalledWith(1, expect.any(Function));
+      expect(expressMocks.app.use).toHaveBeenNthCalledWith(2, 'json-middleware');
+      expect(expressMocks.app.use).toHaveBeenNthCalledWith(3, expect.any(Function));
```
jameshiester-oai, line 385: **[P2] Make this test identify the CSRF guard**

`expect.any(Function)` passes for every function, including a second `mcpHostProtection`, so this test stays green if `csrfProtection` is removed or replaced. The request test below calls the route handler directly and also bypasses middleware. Since browser-origin protection is a security-sensitive claim of this PR, assert the exact `csrfProtection` identity and/or send a hostile-Origin request through a real Express app and verify the MCP transport is not reached. The current implementation passed live QA; the committed regression test would not catch loss of the wiring.

jameshiester-oai (commented):
Reviewed base: `0ef67d937e6868eabbcbd0d215d4530514e66fd0`
Reviewed head: `eaab329e212cad8b7bef7972463cecdb48d07e06`
Scope: full `base...head` review; no earlier authenticated completion marker exists for this head.

Verdict: **COMMENT — two P2 findings.** The loopback, Host, and browser-origin controls work on the real HTTP/MCP path, and exact-head CI is green. The main implementation risk is backward compatibility; the main regression-coverage risk is that the new CSRF wiring test does not prove which middleware was installed.

## Verified findings

### P2

- The fixed `127.0.0.1` listener removes the previously documented remote/container HTTP mode with no opt-in or breaking-change treatment. See the inline comment on `src/commands/mcp/server.ts`.
- The MCP CSRF wiring test accepts any function in the protected middleware slot, so it can stay green after the actual guard is removed. See the inline comment on `test/commands/mcp/server.test.ts`.

No P0 or P1 defect was found. The security-focused source-to-sink pass found no LAN, DNS-rebinding, or hostile browser-origin bypass under the repository threat model.

## Existing unresolved feedback

- The P2 thread on `site/docs/integrations/mcp-server.md:176` is valid and already duplicated: this network-only change does not reject JS/TS configs or contain config globs/referenced paths. `resolveConfigs` still receives caller-selected paths.
- The `express.json()`-before-CSRF thread is reproducible: malformed and oversized JSON can return 400/413 before CSRF. With the default roughly 100 KB parser cap and no tool sink reached, I treat it as an existing P3 fail-fast/resource-use improvement rather than a separate bypass.
- The `PROMPTFOO_OUTPUT_DIR` thread is valid; the variable has no implementation reference.
- The unbracketed `Host: ::1` allowlist entry is unreachable, but conforming IPv6 Host authority is bracketed. This is dead-state cleanup, not a valid-client security defect.

- Remove the unreachable unbracketed `::1` allowlist entry. The installed MCP SDK also exposes localhost Host-validation middleware; reusing it may reduce custom authority parsing if Promptfoo's logging/error shape can be preserved.
- Add one ephemeral real-Express test that initializes with the MCP SDK client, rejects a hostile Host and Origin, and confirms shutdown closes the socket. That one test would cover the behavior currently split across mocks.

## Tests and QA

- `npx vitest run test/commands/mcp/server.test.ts test/server/middleware/csrfProtection.test.ts` — exit 0; 43/43 passed.
- `npx vitest run test/commands/mcp/server.test.ts test/server/middleware/csrfProtection.test.ts --sequence.shuffle --sequence.seed=9972` — exit 0; 43/43 passed.
- `npm run tsc` — exit 0.
- `npm run architecture:check` — exit 0; architecture boundaries passed.
- `npx biome check src/commands/mcp/server.ts test/commands/mcp/server.test.ts` — exit 0.
- `npx prettier --check site/docs/integrations/mcp-server.md architecture/edge-baseline.json` — exit 0.
- `git diff --check 0ef67d937e6868eabbcbd0d215d4530514e66fd0...eaab329e212cad8b7bef7972463cecdb48d07e06` — exit 0.
- Sanitized-environment real HTTP/MCP harness — exit 0: health and MCP initialization returned 200 with a session ID; hostile Host and cross-site POST returned 403; the machine's non-loopback interface was unreachable; shutdown closed the server.
- Exact-head GitHub CI — all reported jobs passed, including the Node/OS matrix, integration/smoke suites, CodeQL, and security scan.

## Limitations

- Runtime QA used local macOS/IPv4. Docker, sidecar, remote-host, and IPv6-only compatibility were established from the listener configuration and negative socket probes, not a full deployment matrix.
- The requested PR Review Toolkit command was unavailable (exit 127). I covered its correctness, testing, documentation, error-handling, type-design, and simplification aspects manually and with three independent read-only specialist passes.
- No secret-bearing environment or external provider workflow was used.

PR #9973, opened 2026-07-04
feat(tracing): persist OTEL span events

mldangelo:
## Summary

Persists OTEL span events (OTLP `span.events[]`) end-to-end so they survive ingestion, storage, redaction, and read-back. This is a focused slice extracted from the larger agentic-runtime PR #8871, carved out so the tracing infrastructure can land and be reviewed on its own.

### What changed

- **Schema / migration**: adds a nullable `events` (JSON) column to the `spans` table, backed by a new Drizzle migration.
- **Types** (`src/types/tracing.ts`): new `TraceSpanEvent` interface (`name`, `timestamp`, optional `attributes`) and an optional `events?: TraceSpanEvent[]` on `TraceSpan`.
- **OTLP receiver** (`src/tracing/otlpReceiver.ts`): parses `events` from both the JSON and protobuf OTLP paths, converts `timeUnixNano` to ms (via `longToNumber` for the decoded/protobuf path), and extends secret redaction to also scrub event attributes and event names.
- **Local exporter** (`src/tracing/localSpanExporter.ts`): maps SDK `ReadableSpan.events` into the stored shape with ms timestamps.
- **Store** (`src/tracing/store.ts`): serializes/deserializes `events`, with attribute sanitization applied to event attributes.
- **Trace context** (`src/tracing/traceContext.ts`): surfaces stored events instead of always emitting an empty array.
- **Tests**: span-event coverage for the exporter, store (including redaction), and the OTLP protobuf ingestion path.

### Migration renumbering

The source branch carried this as `0025_magenta_revanche`. Rather than copy that migration wholesale, the schema change was re-applied to `src/database/tables.ts` and the migration was regenerated against the current `main` schema with `npm run db:generate`, producing `0025_cheerful_changeling.sql` (`ALTER TABLE spans ADD events text;`) with a fresh journal entry and full snapshot. `0025` is the next free slot on current `main` (highest existing is `0024`), so there is no collision.

Verified:
- `PROMPTFOO_CONFIG_DIR=$(mktemp -d) npm run db:migrate` applies cleanly (exit 0); the migrated `spans` table has the `events` column.
- `npm run db:generate` reports no pending drift, confirming journal + snapshot + SQL are internally consistent with `tables.ts`.

### Independence from the agentic plugins

This PR contains **only** tracing infrastructure. It does not pull in the agentic-runtime plugins, the `src/redteam/agentic/*` parser, or any redteam grader. The one OTLP-receiver test in #8871 that fed span events through an agentic grader (`getGraderById('promptfoo:redteam:agentic:...')`) was intentionally **excluded**, as were the agentic-focused `site/docs/tracing.md` additions — both belong with the agentic plugins slice. No `agentic` / `redteam` code imports appear in this diff (the only `is_redteam` references are the pre-existing evals column captured by the full-schema snapshot).

### Verification

- `npx vitest run test/tracing/` — 259 passed (13 files)
- `npm run tsc` — clean
- `npm run l && npm run f` — clean
- `npm run architecture:check` — boundaries passed (no new cross-layer edges)

Extracted from #8871.

### DIFF
--- a/src/tracing/otlpReceiver.ts
+++ b/src/tracing/otlpReceiver.ts
@@ -10,6 +10,7 @@ import {
   type DecodedScopeSpans,
   type DecodedSpan,
   decodeExportTraceServiceRequest,
+  longToNumber,
 } from './protobuf';
 import {
   PROMPTFOO_RESOURCE_ATTR_PARENT_SPAN_ID,
@@ -38,10 +39,17 @@ interface OTLPSpan {
   startTimeUnixNano: string;
   endTimeUnixNano?: string;
   attributes?: OTLPAttribute[];
+  events?: OTLPSpanEvent[];
   status?: {
     code: number;
     message?: string;
   };
+}
+
+interface OTLPSpanEvent {
+  timeUnixNano?: string;
+  name: string;
+  attributes?: OTLPAttribute[];
 }
 
 interface OTLPScopeSpan {
@@ -370,9 +378,14 @@ export class OTLPReceiver {
     // credential into an error message) must be scrubbed too — otherwise the secret leaks
     // through a span field the operator believes `redactAttributes` covers.
     const redactedSourceValues = new Set<string>();
-    for (const [key, value] of Object.entries(attributes)) {
-      if (typeof value === 'string' && this.shouldRedactAttribute(key, redactAttributePatterns)) {
-        redactedSourceValues.add(value);
+    for (const attributeSet of [
+      attributes,
+      ...(span.events ?? []).map((event) => event.attributes ?? {}),
+    ]) {
+      for (const [key, value] of Object.entries(attributeSet)) {
+        if (typeof value === 'string' && this.shouldRedactAttribute(key, redactAttributePatterns)) {
+          redactedSourceValues.add(value);
+        }
       }
     }
     // `redactedSourceValues` only holds strings, so an undefined statusMessage passes through.
@@ -384,6 +397,11 @@ export class OTLPReceiver {
       name: scrubEcho(span.name),
       statusMessage: scrubEcho(span.statusMessage),
       attributes: this.redactAttributes(attributes, redactAttributePatterns),
+      events: span.events?.map((event) => ({
+        ...event,
+        name: scrubEcho(event.name),
+        attributes: this.redactAttributes(event.attributes, redactAttributePatterns),
+      })),
     };
   }
 
@@ -713,6 +731,7 @@ export class OTLPReceiver {
             'otel.span.kind': spanKindName,
             'otel.span.kind_code': span.kind,
           };
+          const startTime = Number(span.startTimeUnixNano) / 1_000_000;
 
           traces.push({
             traceId,
@@ -720,9 +739,14 @@ export class OTLPReceiver {
               spanId,
               parentSpanId,
               name: span.name,
-              startTime: Number(span.startTimeUnixNano) / 1_000_000, // Convert to ms
+              startTime, // Convert to ms
               endTime: span.endTimeUnixNano ? Number(span.endTimeUnixNano) / 1_000_000 : undefined,
               attributes,
+              events: (span.events ?? []).map((event) => ({
+                name: event.name,
+                timestamp: event.timeUnixNano ? Number(event.timeUnixNano) / 1_000_000 : startTime,
+                attributes: this.parseAttributes(event.attributes),
+              })),
               statusCode: span.status?.code,
               statusMessage: span.status?.message,
             },
@@ -902,6 +926,14 @@ export class OTLPReceiver {
           'otel.span.kind': spanKindName,
           'otel.span.kind_code': spanKindCode,
         },
+        events: (span.events ?? []).map((event) => ({
+          name: event.name,
+          timestamp:
+            this.toMilliseconds(event.timeUnixNano) ??
+            this.toMilliseconds(span.startTimeUnixNano) ??
+            0,
+          attributes: this.parseDecodedAttributes(event.attributes),
+        })),
         statusCode: span.status?.code,
         statusMessage: span.status?.message,
       },
@@ -1101,7 +1133,7 @@ export class OTLPReceiver {
     if (value === undefined) {
       return undefined;
     }
-    return Number(value) / 1_000_000;
+    return longToNumber(value) / 1_000_000;
   }
 }
 
--- a/src/tracing/store.ts
+++ b/src/tracing/store.ts
@@ -3,7 +3,7 @@ import { getDb } from '../database/index';
 import { spansTable, tracesTable } from '../database/tables';
 import logger from '../logger';
 
-import type { TraceData } from '../types/tracing';
+import type { TraceData, TraceSpanEvent } from '../types/tracing';
 
 interface StoreTraceData extends Omit<TraceData, 'spans'> {
   evaluationId: string;
@@ -18,6 +18,7 @@ export interface SpanData {
   startTime: number;
   endTime?: number;
   attributes?: Record<string, any>;
+  events?: TraceSpanEvent[];
   statusCode?: number;
   statusMessage?: string;
 }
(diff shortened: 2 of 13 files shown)


src/tracing/store.ts
```diff
@@ -121,6 +122,18 @@ function sanitizeAttributes(
   return sanitized;
 }
 
+function sanitizeEvents(
+  events: TraceSpanEvent[] | null | undefined,
+  shouldSanitizeAttributes: boolean,
+): TraceSpanEvent[] | undefined {
+  return events?.map((event) => ({
```
jameshiester-oai, line 129: [P2] Redact event-name echoes alongside sensitive attributes

This sanitizes sensitive keys in `event.attributes` but spreads `event.name` unchanged. In an actual isolated v3 import/read round trip, an event whose name and `authorization` value were both `Bearer event-secret-9973` came back as `{ name: "Bearer event-secret-9973", attributes: { authorization: "<redacted>" } }`; the raw name then reaches trace APIs, exports, and sharing. Because events are newly exposed by this patch, copying a credential into the event name bypasses the default privacy boundary even though the keyed copy is hidden. Please collect the raw values under sensitive keys before sanitizing and scrub matching event names, with a store/export round-trip regression.

src/tracing/otlpReceiver.ts
```diff
@@ -370,9 +378,14 @@ export class OTLPReceiver {
     // credential into an error message) must be scrubbed too — otherwise the secret leaks
     // through a span field the operator believes `redactAttributes` covers.
     const redactedSourceValues = new Set<string>();
-    for (const [key, value] of Object.entries(attributes)) {
-      if (typeof value === 'string' && this.shouldRedactAttribute(key, redactAttributePatterns)) {
-        redactedSourceValues.add(value);
+    for (const attributeSet of [
+      attributes,
+      ...(span.events ?? []).map((event) => event.attributes ?? {}),
```
jameshiester-oai, line 383: [P2] Recursively collect sensitive values before scrubbing event names

The new event collector only inspects top-level string entries. With `redactAttributes: ['api_key']` and `{ name: secret, attributes: { arguments: { api_key: secret } } }`, the recursive attribute sanitizer replaced the nested key with `[REDACTED]`, but this collector never recorded its original value, so `event.name` persisted the secret. That defeats an explicit ingest-time redaction policy. Please collect matched string values during the same recursive walk used to redact nested objects/arrays, then scrub exact echoes in event names and span fields; add a nested `kvlistValue` regression.

src/tracing/otlpReceiver.ts
```diff
@@ -713,16 +731,22 @@ export class OTLPReceiver {
...
             traceId,
             span: {
               spanId,
               parentSpanId,
               name: span.name,
-              startTime: Number(span.startTimeUnixNano) / 1_000_000, // Convert to ms
+              startTime, // Convert to ms
               endTime: span.endTimeUnixNano ? Number(span.endTimeUnixNano) / 1_000_000 : undefined,
               attributes,
+              events: (span.events ?? []).map((event) => ({
```
jameshiester-oai, line 745: [P2] Validate event shape and timestamps before mapping

This endpoint processes untrusted OTLP JSON, but the new mapping assumes `events` is an array of non-null objects with iterable attributes and a numeric timestamp. Exact-head probes showed `events: {}` throws `map is not a function`, `events: [null]` throws while reading `name`, and `timeUnixNano: 'not-a-number'` produces `NaN` (serialized as `timestamp: null`). The first malformed optional event therefore turns an otherwise valid batch into HTTP 500 and drops all its spans; the last violates the public `timestamp: number` contract. Normalize the repeated field/elements, require a finite timestamp, and deliberately reject with 400 or drop malformed records with OTLP partial-success semantics.

src/tracing/store.ts
```diff
@@ -282,6 +300,7 @@ export class TraceStore {
           startTime: span.startTime,
           endTime: span.endTime,
           attributes: span.attributes,
+          ...(span.events ? { events: span.events } : {}),
```
jameshiester-oai, line 303: [P2] Do not persist unchecked imported event payloads

`prepareTraces` validates only `spanId`, `name`, and finite `startTime`, then preserves every other field. This new truthy spread therefore persists arbitrary portable-import `events`. An actual isolated CLI import with `events: "not-an-array"` reported success, but `getTracesByEvaluation` then failed with `events?.map is not a function`; trace APIs return 500 and `Eval.getTraces()` can silently collapse the evaluation's traces to `[]`. Validate/normalize event arrays and members during import, and defensively ignore/filter malformed legacy JSON before mapping so one bad row cannot poison every read.

jameshiester-oai (commented):
## Automated review — COMMENT

Reviewed base: `0ef67d937e6868eabbcbd0d215d4530514e66fd0` (`main`)

Reviewed head: `058d8e39a0c149e6a83db4279866401ed040cf6f`

Scope: fresh full review of `base...head`; no older authenticated completion marker exists for this head.

Verdict: the migration and happy-path event round trip are sound, but four P2 defects remain around privacy, malformed OTLP input, and imported-data integrity. CI is fully green and the PR is mergeable, though it is behind `main`. Codecov reports 73.08% patch coverage with seven changed lines uncovered.

### Verified findings

No P0/P1 findings.

P2 (inline details and reproductions attached):

1. Default trace sanitization redacts a sensitive event attribute but leaves an identical secret in `event.name`, which is then exposed through APIs, exports, and sharing.
2. Configured OTLP redaction misses event-name echoes of nested sensitive attributes even though the nested attribute itself is redacted.
3. Malformed JSON event containers/elements can reject an otherwise valid OTLP batch with HTTP 500, and nonnumeric event timestamps become `NaN`/JSON `null`.
4. Portable v3 imports can persist non-array `events`; subsequent trace reads throw and can make API reads fail or make `Eval.getTraces()` silently return no traces.

P3:

- The new public `TraceSpanEvent` shape is absent from `site/docs/configuration/reference.md` and the JavaScript, Python, and Ruby expected-output examples, and the tracing docs do not describe event redaction. Users cannot discover the new trace-context field or its privacy behavior from the generic docs.

- JSON nanosecond precision loss: https://github.com/promptfoo/promptfoo/pull/9973#discussion_r3522589816
- Protobuf nanosecond precision loss before conversion: https://github.com/promptfoo/promptfoo/pull/9973#discussion_r3522589823
- Prompt/response strip flags do not project event attributes: https://github.com/promptfoo/promptfoo/pull/9973#discussion_r3522594377
- Missing protobuf event timestamps persist at the Unix epoch: https://github.com/promptfoo/promptfoo/pull/9973#discussion_r3522594378

- A single event normalizer shared by JSON/protobuf ingestion, local export, portable import, and defensive DB reads would centralize shape checks, finite timestamp handling, redaction, and optional count/size limits. The current per-ingress mapping is already drifting.
- Checked-in tests cover one aligned protobuf timestamp, one local-exporter event, and mocked store behavior. Add regressions for JSON events, malformed shapes, nested and default redaction, precision boundaries, missing timestamps, portable import, real migration/SQLite round trips, trace-context propagation, and output projection.
- Event count/name/attribute size is unbounded on the local-exporter path and aggregate trace/share payloads; either enforce a central budget or document the intended operational limit.

### Tests and QA

- `npx @biomejs/biome check <10 changed TS/test files>` — exit 0.
- `npx vitest run test/tracing/localSpanExporter.test.ts test/tracing/store.test.ts --sequence.shuffle=false` — exit 0, 30/30 passed.
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/tracing/ --sequence.shuffle=false` — exit 0, 13 files and 259/259 tests passed on the clean retry.
- Isolated `npm run db:migrate` run twice — exit 0 both times; the nullable `events` column was present and migration count stayed stable.
- Actual local CLI import against isolated config: valid events round-tripped through SQLite; a sensitive keyed value became `<redacted>` while the identical event name remained visible.
- Actual local CLI import of `events: "not-an-array"` succeeded, then store readback failed with `events?.map is not a function`.
- Direct exact-head JSON parser probes: `events: {}` and `events: [null]` threw; a nonnumeric timestamp was non-finite and serialized as `null`.
- Direct nested-redaction probe preserved the event-name secret while replacing the nested `api_key` value with `[REDACTED]`.
- CI: all required build, test, lint, security, CodeQL, and coverage checks passed.

### Limitations

- `/pr-review-toolkit:review-pr all parallel` is unavailable in this environment (exit 127), so correctness, tests/QA, docs, error handling, type design, and simplification were covered by three independent read-only specialist passes plus orchestrator validation.
- No live browser UI or external OTLP SDK process was exercised; the changed parser, exporter, persistence, import, redaction, and trace-context paths were exercised directly. No user database or cache was accessed.

Issue #7649, opened 2026-02-12
RangeError crash on GET /eval/:id/table for large evals


lucagentile:
## Description

The `GET /eval/:id/table` endpoint crashes with an unhandled `RangeError` when the eval payload exceeds V8's maximum string length during `JSON.stringify()`. This happens with large evals containing base64-encoded images or very large prompt content in individual cells.

## Steps to Reproduce

1. Run an eval that produces large outputs (e.g. with base64 images in prompts), example: prompt with 2 images, eval set with 200 test cases
2. Open the eval in the web UI, after around 50 tests run
3. The table endpoint returns a 500 error and the UI fails to load results

## Root Cause

`res.json()` internally calls `JSON.stringify()`, which throws a `RangeError` when the resulting string exceeds V8's ~512MB limit. The endpoint had no error handling for this case.

## Expected Behavior

The endpoint should gracefully degrade; strip oversized content from the response and return a loadable payload, or return a clear 413 error if the payload is still too large after stripping.

## Fix

Addressed in #7599.

PR #9974, opened 2026-07-04
fix(server): return 413 for oversized eval JSON exports

mldangelo:
## Summary

- Return `413` from `GET /api/eval/:id/table?format=json` when V8 hits a JSON serialization size or call-stack limit.
- Serialize before setting download headers, so failed exports return `{ "error": "Eval JSON export is too large to serialize." }` without `Content-Disposition` or cache headers.
- Keep successful export payloads and headers unchanged.
- Retain the existing prompt-stripping recovery for regular table responses.

## Implementation

Adds `sendJsonResponse`, which serializes once and converts only recognized serialization-limit `RangeError`s to the standard server error response. Circular-reference, `BigInt`, and unrelated serialization errors are rethrown.

Route coverage now exercises the actual Express endpoint for successful and oversized JSON exports. The older test that manually reconstructed the JSON branch was removed.

## Verification

- `node node_modules/vitest/vitest.mjs run test/server/utils/safeJsonResponse.test.ts test/server/eval.test.ts test/server/routes/eval.export.test.ts --reporter=dot` — 3 files, 43 tests passed
- `npm run tsc` — passed
- `npm run build` — passed
- `npm run f` — passed
- `npm run l` — passed

OpenAPI already documents a `413` response for `GET /api/eval/{id}/table`, so no documentation change is needed.

Refs #7649.

### DIFF
--- a/src/server/utils/safeJsonResponse.ts
+++ b/src/server/utils/safeJsonResponse.ts
@@ -1,0 +1,88 @@
+import type { Response } from 'express';
+
+/**
+ * Matches the `RangeError`s V8 throws from `JSON.stringify` when a payload
+ * cannot be serialized because it hits an engine limit: the maximum string
+ * length (~512MB, "Invalid string length") or the maximum call-stack depth for
+ * an extremely deep payload. These are the failure modes behind issue #7649,
+ * where `GET /eval/:id/table` crashed for very large evals (e.g. base64 images
+ * or huge per-cell prompts).
+ */
+const JSON_SERIALIZATION_LIMIT_ERROR_RE =
+  /Invalid string length|Cannot create a string longer than|ERR_STRING_TOO_LONG|Maximum call stack size exceeded/i;
+
+export const DEFAULT_TOO_LARGE_MESSAGE = 'Response payload is too large to serialize';
+
+type JsonResponseLogger = {
+  warn: (message: string, context: Record<string, unknown>) => void;
+};
+
+/**
+ * Returns true when `error` is a `RangeError` that `JSON.stringify` throws for a
+ * payload that exceeds V8's serialization limits (maximum string length or
+ * maximum call-stack depth). Other errors — including circular-reference
+ * `TypeError`s — are intentionally not matched so they surface normally.
+ */
+export function isJsonSerializationLimitError(error: unknown): error is RangeError {
+  return error instanceof RangeError && JSON_SERIALIZATION_LIMIT_ERROR_RE.test(error.message);
+}
+
+/**
+ * Serializes `payload` and sends it as `application/json`, guarding against the
+ * `RangeError` `JSON.stringify` throws when a response would exceed V8's maximum
+ * string length (issue #7649).
+ *
+ * This is a size *guard*, not a trimmer: every payload that fits is sent in full
+ * and byte-for-byte identical to `res.json(payload)`, so the public response
+ * contract is preserved. A payload that cannot be serialized fails loudly with
+ * HTTP 413 and a clear error message rather than throwing an unhandled
+ * `RangeError` — and, crucially, without silently stripping or corrupting the
+ * body into something that no longer matches the response contract.
+ *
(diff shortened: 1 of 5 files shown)


jameshiester-oai (commented):
## Review summary

**Reviewed base:** `e8adf8bf3ccc362801030ec041e4add50c4f329a`  
**Reviewed head:** `d9eca243ce973411a8daa3646dc76ec786f28180`  
**Scope:** Fresh full review. No prior authenticated completion marker exists for this head.

**Verdict:** One P2 correctness/compatibility defect. The 413 path, delayed download headers, error classification, and prompt-stripping recovery otherwise behaved as intended.

**CI:** All 44 visible checks pass. CI run [28709037412](https://github.com/promptfoo/promptfoo/actions/runs/28709037412), attempt 2, completed successfully at this exact head, including the rerun macOS job that previously stalled during cache cleanup.

### Verified findings

- **P2 — Preserve Express JSON serialization settings.** The new helper serializes with bare `JSON.stringify`, so configured `json replacer`, `json spaces`, and `json escape` settings are silently bypassed. A real Express/Supertest wire probe showed native `res.json` omitting a replacer-filtered field, escaping `<&>`, and pretty-printing, while `sendJsonResponse` emitted the filtered field in compact, unescaped JSON. This can bypass an embedding application's configured redaction or escaping policy. See the inline comment.

### Existing feedback

The unresolved Copilot thread about top-level `undefined` is not reproduced as a defect: native Express and this helper both returned the same empty 200 JSON response in direct wire probes. The logging-wording thread is outdated and its concern is fixed. Neither thread duplicates the finding above.

Keep the pre-serialization/backstop design, but reuse Express-equivalent serialization semantics from `res.app` before applying success-only headers. Add a real Express response test with replacer, escaping, and spacing configured; the current mock response and parsed-body route tests cannot detect this wire-level regression.

### Validation

- `node node_modules/vitest/vitest.mjs run test/server/utils/safeJsonResponse.test.ts test/server/eval.test.ts test/server/routes/eval.export.test.ts --reporter=dot --sequence.seed=9974` — 3 files, 43 tests passed under Node 24.18.0/npm 11.16.0.
- `node --import tsx --input-type=module -e '<Express/Supertest JSON-settings harness>'` — reproduced native/helper wire divergence for replacer, escaping, and spacing.
- `git diff --check e8adf8bf3ccc362801030ec041e4add50c4f329a...d9eca243ce973411a8daa3646dc76ec786f28180` — passed.
- Earlier exact-head validation on this unchanged SHA also passed TypeScript, full build, and changed-file Biome checks.
- Three bounded read-only specialists independently confirmed the P2 and found no additional actionable defect.

### Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (exit 127), so its six review lenses were applied manually and reconciled with the specialists. The oversized response path used deterministic RangeError injection rather than allocating a real ~512 MB JSON string; no credentialed external service was used.

src/server/utils/safeJsonResponse.ts
```diff
@@ -0,0 +1,65 @@
...
+  }: {
+    beforeSend?: () => void;
+    evalId?: string;
+    logger?: JsonResponseLogger;
+    tooLargeMessage?: string;
+  } = {},
+): void {
+  let body: string | undefined;
+  try {
+    body = JSON.stringify(payload);
```
jameshiester-oai, line 47: **[P2] Preserve Express JSON serialization settings**

The route previously used `res.json(jsonData)`, which serializes with the app's configured `json replacer`, `json spaces`, and `json escape` settings. Bare `JSON.stringify(payload)` bypasses all three. In a real Express probe, a replacer-filtered field reappeared and `<&>` was emitted unescaped and compact despite those settings being enabled. An embedding application can use the replacer for redaction, so this is more than formatting drift.

Please serialize with the response app's Express-equivalent replacer/spacing/escaping semantics before `beforeSend`, and add a real Express/Supertest regression test covering those settings.

PR #9980, opened 2026-07-04
feat(providers): honor CLAUDE_CONFIG_DIR for Claude Code creds

### LINKED ISSUE #9979: Claude Code auth ignores CLAUDE_CONFIG_DIR when locating credentials

Promptfoo's Anthropic provider can reuse a local Claude Code session's OAuth credential (`apiKeyRequired: false`, see [docs](https://www.promptfoo.dev/docs/providers/anthropic/#authenticating-via-a-claude-code-session)). On Linux/Windows (and as a macOS fallback), it reads the credential from a hardcoded `$HOME/.claude/.credentials.json` path.

The Claude Code CLI itself supports `CLAUDE_CONFIG_DIR` to relocate its entire `~/.claude` directory (settings, credentials, etc. — see [Claude Code docs](https://code.claude.com/docs/en/debug-your-config.md)). Users who set this variable to use a non-default Claude Code config location (e.g. running multiple isolated Claude Code profiles) find that Promptfoo can't locate their credentials, since it always looks under `~/.claude` regardless of this override.

`src/providers/anthropic/claudeCodeAuth.ts`'s credential-file resolution should honor `CLAUDE_CONFIG_DIR` the same way the Claude Code CLI does, falling back to `~/.claude` when it isn't set. This keeps the two tools in sync without requiring a Promptfoo-specific environment variable.

A Promptfoo-specific override (e.g. a new `CLAUDE_CODE_CREDENTIALS_DIR` var) was considered, but honoring the same variable the real CLI already uses avoids adding a second knob that has to be kept in sync with the user's actual Claude Code setup.

**Additional context**

Note this only affects the file-based credential lookup (Linux/Windows, and the macOS fallback when the keychain entry is missing). On macOS, Claude Code stores credentials in the system keychain regardless of `CLAUDE_CONFIG_DIR`, so the keychain lookup path is unaffected.

idg2:
## Summary

- Promptfoo's Claude Code OAuth credential loader (`src/providers/anthropic/claudeCodeAuth.ts`) hardcoded `$HOME/.claude/.credentials.json` as the file-based fallback path.
- The Claude Code CLI itself honors `CLAUDE_CONFIG_DIR` to relocate its entire `~/.claude` directory. Promptfoo ignored this, so users who customized their Claude Code config location couldn't have Promptfoo find their credentials.
- `resolveCredentialsPath()` now reads `CLAUDE_CONFIG_DIR` (via `getEnvString`) and falls back to `~/.claude` when unset. Added the var to `ProviderEnvOverridesSchema` (`src/contracts/env.ts`) so it can also be set per-provider via config `env:` overrides, and regenerated `site/static/config-schema.json`.
- Note: this only affects the file-based lookup (Linux/Windows, and the macOS fallback when the keychain entry is missing) — macOS Claude Code always stores credentials in the system keychain regardless of `CLAUDE_CONFIG_DIR`. Documented this caveat in code comments and `site/docs/providers/anthropic.md`.

Closes #9979

## Test plan

- [x] `npm run tsc`
- [x] `npx vitest run test/providers/anthropic` (314 tests passing, including 2 new cases covering the `CLAUDE_CONFIG_DIR` override and the unset fallback)
- [x] `npm run jsonSchema:generate` (schema diff reviewed, contains only the new `CLAUDE_CONFIG_DIR` key)


mldangelo:
Thanks @idg2 — the approach and the file-path handling here were solid. I pushed three maintainer fixes on top (verified against claude v2.1.204 by tracing its `security` calls):

1. **Provider/root `env:` blocks now actually work.** `CLAUDE_CONFIG_DIR` was added to `ProviderEnvOverridesSchema`, but `getEnvString()` can't see provider-scoped env — and at provider-construction time it can't see the root config `env:` block either (`cliState.config` is assigned after `loadApiProviders`). The provider's merged `env` is now threaded through `loadClaudeCodeCredential(this.env)`, matching the `getApiKey()` pattern.
2. **macOS keychain is profile-aware.** When `CLAUDE_CONFIG_DIR` is set, the Claude Code CLI stores the credential under `Claude Code-credentials-<sha256(dir)[:8]>` (hash of the raw value, applied even when set to the default `~/.claude`), not the plain service — so the keychain-first lookup either silently returned the *default* profile's token or nothing, and the docs claim that the keychain was unaffected wasn't accurate. The lookup now derives the same service name the CLI does, with the file path as fallback.
3. **CI plumbing:** architecture edge-baseline bump for the new providers-layer imports (the Style Check failure on the previous run was a 5-minute job timeout flake, but on re-run it would have failed on `architecture:check`), an empty-string `CLAUDE_CONFIG_DIR` regression test, and host-env isolation for the test suite.

Merging once CI is green.

PR #9981, opened 2026-07-05
docs(site): fix LocalAI REQUEST_TIMEOUT_MS default to 300000

anxkhn:
```markdown
The LocalAI provider docs list the default for `REQUEST_TIMEOUT_MS` as `60000`,
but the effective default is `300000`.

`localai.ts` gets its request timeout from `getRequestTimeoutMs()`, which returns
`getEnvInt('REQUEST_TIMEOUT_MS', 300_000)` in `src/providers/shared.ts`. The
default was raised from 60000 to 300000 there, but this docs line was never
updated. The sibling WebSocket provider docs already document `300000`.

This corrects the documented default to match the code.

## Test plan

Docs-only change; no runtime behavior changes.

- `npx prettier --check site/docs/providers/localai.md` passes.
- Cross-checked `getRequestTimeoutMs()` in `src/providers/shared.ts` (returns
  `getEnvInt('REQUEST_TIMEOUT_MS', 300_000)`) against the three usages in
  `src/providers/localai.ts`.
```


mldangelo (approved):
Nice catch—thanks for tracing this from the LocalAI call sites back to the shared timeout helper. I confirmed chat, embeddings, and completion all use the `300_000` ms default, and this is a precise docs-only correction. Appreciate the clear test plan!

PR #6423, opened 2025-12-01
feat(providers): add OpenCode SDK provider

mldangelo:
## Summary

- Add new OpenCode SDK provider (`opencode:sdk`) for evaluating LLM applications with 75+ provider support
- OpenCode is an open-source AI coding agent with built-in tools (bash, read, write, edit, grep, glob)
- Client-server architecture with automatic server lifecycle management

## Features

- **Multi-provider support**: Works with Anthropic, OpenAI, Google, Ollama, and 70+ other providers
- **Tool configuration**: Enable/disable specific tools per eval
- **Permission control**: Configure allow/deny/ask for dangerous operations
- **Custom agents**: Define custom agent configurations with system prompts
- **MCP integration**: Support for local and remote MCP servers
- **Session management**: Ephemeral or persistent sessions
- **Working directory isolation**: Sandboxed file operations with fingerprint-based caching

## Test plan

- [x] Unit tests (33 tests passing)
- [x] E2E tests for basic prompting
- [x] E2E tests for tools configuration  
- [x] E2E tests for custom_agent with system prompt
- [x] E2E tests for permission allow (bash executes)
- [x] E2E tests for permission deny (bash blocked)

PR #9997, opened 2026-07-06
docs(providers): correct OpenCode SDK port @default to 0

anxkhn:
```markdown
## What

The `OpenCodeSDKConfig.port` JSDoc claims `@default 4096`, but the provider never
uses `4096`. When it starts a server it falls back to `config.port ?? 0`, and port
`0` tells the OS to pick a random available port:

https://github.com/promptfoo/promptfoo/blob/main/src/providers/opencode-sdk.ts#L1170

```ts
port: config.port ?? 0,
```

So the documented default is wrong two ways:
- it names a port (`4096`) that the code never selects, and
- it contradicts the provider's own docs table, which already lists the `port`
  default as `Auto-select`
  ([providers/opencode-sdk.md](https://www.promptfoo.dev/docs/providers/opencode-sdk/#configuration)).

The incorrect `@default 4096` has been there since the provider was first added
(#6423); the code has always used `?? 0`.

## Fix

Update only the JSDoc to state the real default:

```diff
   /**
    * Port for the OpenCode server (when starting a new server)
-   * @default 4096
+   * @default 0 (random available port)
    */
   port?: number;
```

This matches the runtime behavior and the docs table, and reuses the exact wording
the codebase already uses for the analogous `serverPort` field in
`src/providers/openai/chatkit-types.ts` (`@default 0 (random available port)`).

## Tests

Documentation-comment change only; no behavior change, so no test is added or
modified. For sanity I confirmed the diff does not disturb anything:

```
npx vitest run test/providers/opencode-sdk.test.ts   # 78/78 pass
npm run tsc                                           # exit 0
npm run l && npm run f                                # no changes
```

(The `4096` in `test/providers/opencode-sdk.test.ts` is an unrelated mock server
URL, `http://127.0.0.1:4096`, not an assertion of the default; it is untouched.)
```

### DIFF
--- a/src/providers/opencode-sdk.ts
+++ b/src/providers/opencode-sdk.ts
@@ -280,7 +280,7 @@ export interface OpenCodeSDKConfig {
 
   /**
    * Port for the OpenCode server (when starting a new server)
-   * @default 4096
+   * @default 0 (random available port)
    */
   port?: number;
 


mldangelo (approved):
Thanks @anxkhn — this is a model docs fix. I verified the claim end-to-end: the only server-start path is `port: config.port ?? 0` in `src/providers/opencode-sdk.ts`, and when `baseUrl` is set the port is ignored entirely, so `4096` was never a real default. Nice touches: matching the provider docs table ("Auto-select"), reusing the exact `@default 0 (random available port)` wording from `chatkit-types.ts` for consistency, and even preemptively flagging that the `4096` in the test file is an unrelated mock URL. The PR description made this trivial to review. Much appreciated!

PR #10006, opened 2026-07-06
chore(deps): pin js-yaml to 5.2.0

jameshiester-oai:
## Summary
- Pin root and app js-yaml specifiers to 5.2.0.
- Update package-lock to resolve js-yaml 5.2.0.

## Why
Cloud is moving to js-yaml v5 and should match the upstream submodule version directly. Pinning the manifest avoids downstream consumers needing to decide whether to match a range or a resolved package version.

## Test plan
- npm install --package-lock-only --ignore-scripts
- npx @biomejs/biome check --write package.json src/app/package.json package-lock.json
- npm run tsc
- npm --prefix src/app run tsc

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,7 +62,7 @@
         "http-z": "^8.1.1",
         "istextorbinary": "^9.5.0",
         "js-rouge": "^3.2.0",
-        "js-yaml": "^5.0.0",
+        "js-yaml": "5.2.0",
         "json5": "^2.2.3",
         "keyv": "^5.6.0",
         "keyv-file": "^5.3.3",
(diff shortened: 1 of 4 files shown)


mldangelo-oai (approved):
Why do we care about this? Should we also merge https://github.com/promptfoo/promptfoo/pull/10004 ?

jameshiester-oai:
> Why do we care about this? Should we also merge #10004 ?

We should close 10004 and merge this.  5.2 ships with types.  This is better than 10004 because it pins the version instead of depending on the npm slot machine to make sure the version matches with the cloud version

PR #10010, opened 2026-07-07
docs(site): document combining defaults with ignore in tool-args-match

### LINKED ISSUE #9842: `feat(assertions): support wildcard (any-value) entries in trajectory:tool-args-match defaults`
The `defaults` field introduced in #9471 (closes #9458) lets `mode: exact` tolerate optional arguments whose values match a declared default -> for example, `page_size: 5` strips `page_size` from the actual payload only when the agent passed exactly `5`. However, there is no way to say "ignore this argument regardless of what value the agent chose." When a tool parameter is genuinely optional and its value is agent-determined (e.g., a pagination cursor, a locale string, a request ID), every possible value would need to be enumerated, which is impossible. The result is that `mode: exact` still fails on any agent-chosen value for that key, even though the key itself is irrelevant to the assertion.

**The solution i would like**

Extend the `defaults` map to support two distinct entry forms:

1. **Value-specific default** (existing behaviour): `page_size: 5` — strip the key from the actual payload only when its value deep-equals `5`.
2. **Wildcard default** (new): `page_size: "*"` (or an agreed-upon sentinel such as a bare key in an array form) — strip the key from the actual payload unconditionally, regardless of the value the agent supplied.

Example config using the proposed wildcard sentinel `"*"`:

```yaml
  value:
    name: orders
    mode: exact
    args: { status: Q }
    defaults:
      page: 1          # strip only when value == 1  (existing)
      page_size: 5     # strip only when value == 5  (existing)
      cursor: "*"      # strip for any value          (new)
      request_id: "*"  # strip for any value          (new)
```

Expected outcome table:

| Observed args | Outcome |
|---|---|
| `{status:'Q'}` | pass |
| `{status:'Q', page:1, page_size:5}` | pass (value-specific defaults stripped) |
| `{status:'Q', cursor:'abc123', request_id:'req-9'}` | pass (wildcard defaults stripped) |
| `{status:'Q', page:2}` | fail (`page` value ≠ declared default `1`) |
| `{status:'Q', delete_database:true}` | fail (hallucinated extra, not in defaults) |

The stripping logic in `stripDefaults` (introduced in #9471) would be extended: when a defaults entry's value is the wildcard sentinel, delete the key from the actual payload unconditionally rather than checking deep equality.

**Describe alternatives i have considered**

- **Use `mode: partial`**: silently tolerates all extra arguments, including dangerous hallucinated ones like `delete_database: true`. This defeats the purpose of catching hallucinated extras.
- **Enumerate all possible values**: not feasible for agent-generated values such as cursors, UUIDs, or timestamps.
- **Separate `ignore` array field** (e.g., `ignore: [cursor, request_id]`): a parallel field would work but fragments the configuration surface. Keeping everything in `defaults` is more consistent with the existing design and keeps the semantics in one place.

**Additional context**

This builds directly on the `defaults` feature from [#9471](https://github.com/promptfoo/promptfoo/pull/9471) (which closed [#9458](https://github.com/promptfoo/promptfoo/issues/9458)). The implementation surface is small - only `stripDefaults` in `src/assertions/trajectory.ts` needs to branch on whether the declared default value equals the wildcard sentinel before deciding to strip unconditionally. The sentinel value (`"*"`) should be documented clearly to avoid confusion with a literal string argument whose value happens to be `"*"`. If that ambiguity is a concern, an alternative sentinel such as `null` or a dedicated `ignore` boolean could be considered, but `"*"` is idiomatic in glob/pattern contexts already used elsewhere in promptfoo (e.g., `trajectory:tool-used` patterns).

muhammadshayan124:
## Summary

`trajectory:tool-args-match` supports a `defaults` map so an actual argument matching a declared default value gets stripped before comparison (e.g. `page: 1`). There was no way to tolerate an argument that varies on every call and can't be pinned to one value — for example a `cursor` or opaque pagination token.

This adds a wildcard sentinel: setting a `defaults` entry to the string `"*"` strips that key unconditionally, regardless of its actual value.

```yaml
  value:
    name: search_orders
    mode: exact
    args:
      status: Q
    defaults:
      page_size: 5   # only stripped when actual value equals 5
      cursor: "*"    # stripped no matter what value was passed
```

Closes #9842.

## Changes

- `src/assertions/trajectory.ts`: `stripDefaults` now strips a key unconditionally when its declared default is the `"*"` sentinel, in addition to the existing exact-value match.
- `test/assertions/trajectory.test.ts`: added coverage for unconditional stripping across varying value types (string/null/number/object), an omitted wildcard key, combining a wildcard with a value-specific default in the same call, exact-mode failure when the non-wildcard default still mismatches, and the pass-reason message.
- `site/docs/configuration/expected-outputs/deterministic.md`: documented the wildcard, including the caveat raised in the issue — `defaults` can't distinguish a wildcard from a tool that legitimately uses the literal string `"*"` as a real argument value.

## Test plan

- [x] `npx vitest run test/assertions/trajectory.test.ts` — 104/104 passing
- [x] `npx vitest run test/assertions` — 1321/1321 passing (no regressions)
- [x] `npx tsc --noEmit` — clean
- [x] `npm run l && npm run f` (Biome + Prettier via pre-commit hook) — clean


mldangelo (commented):
Thank you so much for this contribution, @muhammadshayan124 — and welcome! 🎉 For a first PR this is genuinely impressive work: the implementation is minimal and well-commented, the test plan is thorough, CI is green, and the docs honestly call out the sentinel's ambiguity with a literal `"*"` value. That caveat you documented turns out to be exactly the right instinct, which brings me to what I found while testing this — and then a proposal, because I think there's a version of this PR we'd love to land.

## 1. The existing `ignore` option already covers this use case

`trajectory:tool-args-match` gained an `ignore` option in #9466 (glob patterns added in #9484) that strips a key from the comparison unconditionally, regardless of value. It shipped just a few weeks before #9842 was filed, so it's a completely understandable miss — especially since the `defaults` docs never cross-reference it (that one's on us). Your example config works today as:

```yaml
  value:
    name: search_orders
    mode: exact
    args:
      status: Q
    defaults:
      page_size: 5 # stripped only when the value equals 5
    ignore:
      - cursor # stripped no matter what value was passed
```

I ran the full expected-outcome table from #9842 against `main` using `defaults: { page: 1, page_size: 5 }` + `ignore: [cursor, request_id]`, and every row behaves as the issue specifies — including `delete_database: true` still failing. `ignore` also supports glob key patterns (`ignore: ['*_id']`), which a per-key sentinel can't express.

## 2. The `"*"` sentinel silently changes behavior for existing configs

Because `defaults` values share a domain with real argument values, overloading `"*"` repurposes configs that use it literally — and `fields: '*'`, `scope: '*'`, `select: '*'` are common real-world API defaults. I verified both sides: on `main`, `defaults: { fields: '*' }` with an observed `fields: 'ssn,password'` fails (correctly — the value doesn't match the declared default); on this branch the same call passes. For an assertion library, silently converting failures into passes is the riskiest kind of change, and there's no way for a user to opt back into literal matching.

Notably, flipping that behavior broke zero existing tests — which is itself a gap, and it points at how your work here could land.

## Proposal: re-purpose this PR — we'd be delighted to merge that version

Your investigation surfaced two genuine holes, and this PR is well positioned to fix both:

1. **Add the missing regression test.** Nothing today pins the semantics that a literal `"*"` default only strips on an exact value match (that's why your change passed 104/104). A test like your `fields`/`cursor` scenarios, asserting the literal-match behavior, would permanently protect existing users from exactly the ambiguity you documented. The `makeIgnoreParams`-style helpers already in `test/assertions/trajectory.test.ts` should keep it compact.
2. **Fix the discoverability gap in the docs.** Add a sentence to the `defaults` section in `deterministic.md` pointing readers to `ignore` for "tolerate this key regardless of value," and perhaps a `cursor`/pagination-token example in the `ignore` section — more evocative than the current `request_id` one. I suspect that missing cross-reference is the reason #9842 exists at all, so this fixes the root cause.

If you're up for reshaping the PR along those lines, I'd be happy to review and merge it quickly. And if you'd rather not, absolutely no worries — just say the word and we'll take it from here, with credit to you for surfacing the gap.

Thanks again for digging this deep into the trajectory assertions — this level of rigor makes for a great contribution either way. 🙏

mldangelo:
Quick update for transparency, @muhammadshayan124: since the review above, the regression test pinning the literal `"*"` semantics and the docs cross-reference landed in #10083 (crediting you — thank you again for surfacing both gaps!), which also closed #9842.

Rather than close this PR, I've used the maintainer-edit permission to re-purpose it along the lines suggested above, so your contribution can land too. What I pushed:

- Reverted the `"*"` sentinel change in `stripDefaults` (per the compatibility concern discussed above).
- Adapted your value-type test scenarios — the string/null/number/object cursor cases from your wildcard tests — into coverage for `ignore`, which previously only had string-valued tests. Nice catch hiding in your test design there.
- Turned the expected-outcome table from your issue into a worked docs example showing `defaults` and `ignore` composing, which is exactly the recipe for the use case that motivated #9842.

The diff is now small (+41 lines, tests and docs only) and I've retitled the PR to match. Once it merges, GitHub's squash commit will credit you as the author of record. If you'd rather take it in a different direction, feel free to push over my changes — and no hard feelings either way. Thanks again for a genuinely well-built first contribution; we'd love to see more from you. 🙏

PR #10012, opened 2026-07-07
fix(assertions): allow numeric 0 as contains/icontains value

WatchTree-19:
### problem
`handleContains` / `handleIContains` guard the value with `invariant(value, ...)`, which is a truthiness check and so rejects the numeric value `0`:

```
Invariant failed: "contains" assertion type must have a string or number value
```

but the very next line (`invariant(typeof value === 'string' || typeof value === 'number', ...)`) and `String(value)` show the handler is meant to accept numbers, and `AssertionValue` includes `number`. so `contains: 0` (e.g. asserting output contains "0 errors") is a supported config that throws. numeric values flow through unchanged (`index.ts` only re-renders string values), so `0` reaches the handler as `0`.

### fix
replace the truthiness guard with an explicit `undefined`/`null`/`''` check in both handlers - the same idiom already used in `src/assertions/wordCount.ts` (`invariant(value != null, ...)`). missing/empty values are still rejected; `0` is now allowed.

### test
added two cases to `test/assertions/contains.test.ts` (`contains` with numeric `0` matching and non-matching). before: `contains: 0` throws; after: matches -> pass, non-match -> fail (not throw); `undefined`/`null`/`''` still throw; strings unaffected; `icontains` fixed too.

### DIFF
--- a/site/docs/configuration/expected-outputs/deterministic.md
+++ b/site/docs/configuration/expected-outputs/deterministic.md
@@ -104,6 +104,9 @@ assert:
     value: 'The expected substring'
 ```
 
+Both assertions accept string or number values. Numbers are converted to strings, so `value: 0`
+matches output containing `0`.
+
 ### Contains-All
 
 The `contains-all` assertion checks if the LLM output contains all of the specified values.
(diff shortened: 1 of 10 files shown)


mldangelo-oai (approved):
Thanks for the thoughtful contribution! This closes a real edge case for numeric assertions, and the end-to-end coverage now protects both `contains` and `icontains`. Everything is green—nicely done.

PR #10015, opened 2026-07-07
fix(redteam): stop empty prompts at metadata boundaries

mldangelo-oai:
## Summary
- stop empty Prompt markers from absorbing the next generated case metadata
- preserve pretty-printed MCP JSON as a valid isolated prompt
- add regression coverage for consecutive metadata-wrapped tool calls

## Test plan
- `/Users/mdangelo/code/promptfoo/node_modules/.bin/vitest run test/redteam/plugins/base.test.ts --sequence.shuffle=false`
- `/Users/mdangelo/code/promptfoo/node_modules/.bin/biome check --write src/redteam/plugins/multiInputFormat.ts test/redteam/plugins/base.test.ts`
- `git diff --check`


jameshiester-oai (commented):
## Review summary

Reviewed base `ede696e1905b2aa4469d259fe208edde645014cf` through exact head `cc5fdc579d80abc42cddb76ae433786702945ee2` as a fresh full review.

**Verdict:** one P2 red-team reliability defect. The visible exact-head CI state is green (44/44 checks successful), but the focused regression coverage does not exercise whitespace inside the generated JSON value.

### Verified findings

#### P2

- Valid pretty-printed MCP JSON containing an interior blank line is truncated at that blank. The parser still returns the truncated fragment as a generated test, so malformed tool-call input can replace the intended attack and reduce coverage. See the inline comment for the exact reproduction and correction direction.

No P0, P1, or P3 findings survived validation.

### Existing feedback

There are no existing inline review threads. The scanner and Copilot all-clear reviews do not exercise this whitespace case, so this finding does not duplicate prior feedback.

The parser will be easier to reason about if empty-marker content is collected as a complete structural value (for JSON, through the balanced value) and only then bounded by the next structural prompt/metadata marker. The regression suite should include a blank line *inside* valid JSON; the added case currently has its blank only between cases. Decorated/numbered marker variants remain useful follow-up coverage but were not treated as PR-introduced findings here.

### Validation performed

- `source ~/.nvm/nvm.sh && nvm use` — Node 24.18.0, npm 11.16.0.
- `npm ci` — exact lockfile install completed; tracked worktree stayed clean.
- `npx vitest run test/redteam/plugins/base.test.ts --sequence.shuffle=false` — 124/124 passed.
- `npx vitest run test/redteam/plugins/base.test.ts --sequence.shuffle --sequence.seed=10015` — 124/124 passed.
- `npm run tsc` — passed.
- `npx @biomejs/biome check src/redteam/plugins/multiInputFormat.ts test/redteam/plugins/base.test.ts` — passed.
- `npm run build` — core, declarations, and production Web UI build passed.
- `git diff --check ede696e1905b2aa4469d259fe208edde645014cf...cc5fdc579d80abc42cddb76ae433786702945ee2` — clean.
- Exact-source parser probes: the first JSON object with an interior blank became `{\n  "tool": "search_docs",` and failed `JSON.parse`; the adjacent control object remained valid.
- Credential-free isolated `npm run local -- redteam generate ... --no-cache --force --no-progress-bar --strict` — exited 0 and reported 2/2 tests generated, while the exported YAML preserved the first malformed fragment, confirming user-visible silent corruption.

### Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (exit 127), so its six lenses were applied manually and reconciled with three read-only specialists. I did not use a live remote generator, MCP server, hosted grader, credentials, browser session, or the full repository test suite.

src/redteam/plugins/multiInputFormat.ts
```diff
@@ -170,24 +170,27 @@ function parseMultiLinePrompts(
...
-    const trimmedLine = line.trim();
+  for (let i = 0; i < promptLineIndices.length; i++) {
+    const promptIndex = promptLineIndices[i];
+    const prompt = removePrefix(lines[promptIndex].trim(), 'Prompt');
 
-    if (hasPromptMarker(trimmedLine)) {
-      if (inPrompt && currentPrompt.trim().length > 0) {
-        prompts.push(currentPrompt.trim());
+    if (prompt.length === 0) {
+      const promptAfterEmptyMarker = collectPromptAfterEmptyMarker(lines, promptIndex + 1);
```
jameshiester-oai, line 179: **[P2] Preserve interior whitespace in pretty-printed tool calls**

Routing every empty-marker block through `collectPromptAfterEmptyMarker` makes the first blank line after content terminate the prompt. With two adjacent markers and a valid first object such as `Prompt:\n{\n  "tool": "search_docs",\n\n  "args": {"query": "first"}\n}\nPrompt:\n...`, this exact head returns the first `__prompt` only as `{\n  "tool": "search_docs",`, so JSON parsing and direct MCP materialization fail. A local `redteam generate` still reported 2/2 success and wrote that malformed fragment as a test, which can force inference repair or silently reduce attack coverage. The base loop continued through internal whitespace until the next `Prompt:` marker. Please collect a complete JSON value (including blank whitespace) before applying inter-case boundaries, and add a regression with the blank line inside the JSON object.

mldangelo (approved):
Reviewed, synced with `main`, tested, and QA'd end to end. This is a clean, well-targeted fix. ✅

**The bug is real and I reproduced it:** running the new regression test against the pre-fix source, the first prompt came back polluted with the *next* case's `WARNING:/Plan:/Reason:` metadata — exactly the "empty `Prompt:` marker absorbs the following case" failure. For pretty-printed MCP tool-call JSON that silently corrupted the payload. Delegating empty-marker collection to the existing `collectPromptAfterEmptyMarker` (which stops at `hasPromptBoundaryMarker` boundaries and the blank line between cases) is the right fix, and it makes the multi-line path consistent with the single-line/legacy path.

**Verification:**
- Merged `origin/main` cleanly; `tsc` and Biome clean.
- 129 tests pass across `base.test.ts` + `pluginDocumentation.test.ts`.
- Drove `parseGeneratedPrompts` against 7 realistic shapes (blank-separated JSON tool calls, adjacent metadata with no blank line, inline markers, single-marker legacy path, mixed empty+inline, content containing a `Plan:` line, two-empty-marker JSON) — all produce correctly isolated prompts.

**One small follow-up I pushed** (`1ac4e5cd4`): the inline `Prompt: <content>` branch had a dead conditional — since `removePrefix` trims, `prompt` is always non-empty there, so the `if (currentPrompt || line.trim())` guard and `currentPrompt ? '\n' : ''` ternary could never take their false path. Replaced the manual accumulator loop with a `slice` + `join`; verified byte-identical output across the full suite and all QA fixtures.

Merging. 🚀

PR #10021, opened 2026-07-08
fix(providers): handle TrueFoundry guardrail blocks

jameshiester-oai:
## Summary

- Normalize TrueFoundry-native guardrail failures into flagged provider responses instead of generic errors.
- Recognize Azure and downstream model safety envelopes proxied through TrueFoundry while leaving unrelated HTTP 400 responses untouched.
- Preserve input or output direction when the response identifies it, and avoid guessing when the signal is ambiguous.

## Root cause

TrueFoundry extends the OpenAI-compatible provider, which serializes unrecognized HTTP 400 responses as generic errors before TrueFoundry-specific or proxied model safety semantics can reach guardrail assertions.

## Test plan

- npx vitest run test/providers/truefoundry.test.ts
- npm run f
- npm run l
- npm run tsc
- Local CLI smoke eval against a stub TrueFoundry HTTP 400 safety response: 1 passed, 0 errors


zcrab-oai (approved):

src/providers/truefoundry.ts
```diff
@@ -35,6 +35,160 @@ type TrueFoundryProviderOptions = ProviderOptions & {
   config?: TrueFoundryCompletionOptions;
 };
 
+type JsonRecord = Record<string, unknown>;
+
+const TRUEFOUNDRY_GUARDRAIL_ERROR_TYPE = 'guardrail_checks_failed';
+const DOWNSTREAM_GUARDRAIL_ERROR_CODES = new Set([
+  'content_filter',
+  'content_filter_error',
```
zcrab-oai, line 43: **P1 — `content_filter_error` is not a content block.** Azure uses this code when the filtering system failed or timed out (this repo already treats it that way in `src/providers/azure/chat.ts`). A proxied HTTP 400 with this code is currently converted into `isRefusal: true` / `guardrails.flagged: true` and the operational error is removed. That can make a red-team test pass because the guardrail was unavailable, not because it blocked the prompt. Please remove this from the unconditional block-code set (or require an actual `filtered`/blocked signal) and add a negative regression test.

src/providers/truefoundry.ts
```diff
@@ -35,6 +35,160 @@ type TrueFoundryProviderOptions = ProviderOptions & {
...
+  return DOWNSTREAM_GUARDRAIL_MESSAGE_PATTERNS.some((pattern) => pattern.test(message));
+}
+
+function getGuardrailDirection(
+  payload: JsonRecord,
+  message: string,
+): Pick<NonNullable<ProviderResponse['guardrails']>, 'flaggedInput' | 'flaggedOutput'> {
+  const checks = payload.guardrail_checks;
+  if (isJsonRecord(checks)) {
+    const flaggedInput = hasGuardrailCheck(checks.llm_input_guardrails);
```
zcrab-oai, line 143: **P2 — Published TrueFoundry response envelopes use another key family.** Current gateway examples in `truefoundry/integrations-custom-guardrails` return `guardrail_checks.input_guardrails` / `output_guardrails`, and some put the reason in top-level `payload.message` with no `error.message`. This parser accepts only the `llm_*` keys and line 174 only reads `error.message`, so those real blocks become generic and lose input/output direction. Please accept both key families for gateway-version compatibility and prefer `error.message ?? payload.message ?? fallback`, with a contract-shaped test.

src/providers/truefoundry.ts
```diff
@@ -35,6 +35,160 @@ type TrueFoundryProviderOptions = ProviderOptions & {
...
+    const flaggedOutput = hasGuardrailCheck(checks.llm_output_guardrails);
+    if (flaggedInput || flaggedOutput) {
+      return { flaggedInput, flaggedOutput };
+    }
+  }
+
+  if (/\b(prompt|input)\b/i.test(message)) {
```
zcrab-oai, line 150: **P2 — Use structured direction before guessing from text.** Azure can explicitly identify an input block with `error.param: "prompt"` while the message is neutral or only says “The response was filtered.” In that case this returns no direction or marks output even though the provider said the prompt was blocked. Please pass the error object into direction detection and honor `param: "prompt"` before message heuristics.

src/providers/truefoundry.ts
```diff
@@ -125,6 +279,15 @@ export class TrueFoundryProvider extends OpenAiChatCompletionProvider {
     };
   }
 
+  async callApi(
+    prompt: string,
+    context?: CallApiContextParams,
+    callApiOptions?: CallApiOptionsParams,
+  ): Promise<ProviderResponse> {
+    const response = await super.callApi(prompt, context, callApiOptions);
```
zcrab-oai, line 287: **P2 — Normalization happens after the tracing span has ended.** `super.callApi()` completes `withGenAISpan` before this runs, so the span sees `{ error: ... }`, records ERROR, and captures no normalized response body. The caller then receives a successful refusal/guardrail result, making telemetry disagree with runtime behavior (unlike the base `invalid_prompt` path normalized inside the span). Please move this normalization into the traced operation/base error hook, or otherwise ensure the span sees the final response.

zcrab-oai (commented):
Thanks for tightening TrueFoundry guardrail handling. I found four cases where the new normalization can still change results or lose provider semantics, so I’m holding the stamp for now.

PR #10024, opened 2026-07-08
docs: add TokenLab provider example

hedging8563:
## Summary
- add a provider-tokenlab example that uses the existing OpenAI provider with TokenLab's OpenAI-compatible /v1 endpoint
- include both chat completions and Responses API provider entries
- document TokenLab native endpoint options without adding a dedicated provider prefix

## Testing
- git diff --check
- not run: local eval/dependency install (node_modules are not present in this temporary checkout and the example requires a TokenLab API key)

---

**Updated after merging `main` (2026-09-15):** the scope of this PR is unchanged — `main` has no TokenLab content of its own, so nothing here was superseded. The net diff against `origin/main` is still the three files this branch has always added:

| File | Lines |
| --- | --- |
| `examples/provider-tokenlab/README.md` | +60 |
| `examples/provider-tokenlab/promptfooconfig.yaml` | +33 |
| `test/examples/tokenlabReadme.test.ts` | +128 |

Two things have been added to the branch since the Summary above was written, and are worth calling out for reviewers:

- **The native-endpoint guidance now requires an explicit `apiKey`.** The Anthropic and Google providers read only their own vendor env vars (`ANTHROPIC_API_KEY`, `GOOGLE_API_KEY` / `GEMINI_API_KEY`), so pointing them at `apiBaseUrl: https://api.tokenlab.sh` without `apiKey` sends an unauthenticated request even with `TOKENLAB_API_KEY` exported. The README snippets set `apiKey` on every provider, and a "Cost reporting" section notes that models sharing an OpenAI model name (`gpt-5.5`, `gpt-5.4-mini`) are priced at OpenAI list rates unless `cost` / `inputCost` / `outputCost` are set.
- **`test/examples/tokenlabReadme.test.ts` is a regression test for exactly that.** It parses every `providers:` YAML block out of the README plus the example `promptfooconfig.yaml`, loads them with `TOKENLAB_API_KEY` as the only credential in the environment, and asserts each resolved provider returns that key — so a future snippet that drops `apiKey` fails CI instead of silently authenticating with nothing.

This supersedes the "not run" note in the Testing section: `npx vitest run test/examples/tokenlabReadme.test.ts` passes (4 tests) against the current branch tip with `main` merged in, and all PR checks are green. A live eval still requires a real TokenLab API key and has not been run.


hedging8563:
Follow-up pushed in 68cda8d: added a required H2 section to examples/provider-tokenlab/README.md, which addresses the failing README standards assertion. I also saw the webui browser job fail because the Playwright browser executable was missing from the runner cache; that appears unrelated to this docs example change.

hedging8563:
I checked the current failing webui tests job. The failure happens before any TokenLab example code is exercised: Vitest browser cannot launch because the Playwright Chromium executable is missing from the runner. The log says the executable does not exist under the Playwright chromium_headless_shell cache path and suggests running npx playwright install. This PR only adds examples/provider-tokenlab/README.md and examples/provider-tokenlab/promptfooconfig.yaml, so the browser setup failure appears unrelated to the TokenLab example diff. I also tried to rerun the failed job, but GitHub requires admin rights on the upstream repository for that.

PR #10025, opened 2026-07-08
fix(redteam): preserve punctuation in Pig Latin encoding

orangeCatDeveloper:
## Problem

The `piglatin` strategy's `toPigLatin` matched only the **last** alphanumeric run of each space-delimited token (unanchored regex `([a-zA-Z0-9]+)([^a-zA-Z0-9]*)$`) and dropped everything before it. Any single token with interior or leading punctuation — HTML/script, code, URLs, contractions — was corrupted before reaching the target:

| Input | Before | After |
| --- | --- | --- |
| `<script>alert(1)</script>` | `iptscray>` | `<iptscray>alertway(1)</iptscray>` |
| `don't` | `tay` | `onday'tay` |
| `(bomb)` | `ombbay)` | `(ombbay)` |

Because the strategy feeds the whole payload through this, the attack was effectively never delivered — the target "safely" declined a request it never actually received, and the run looked like robustness to Pig Latin obfuscation.

## Fix

Transform each alphanumeric run in place (`text.replace(/[a-zA-Z0-9]+/g, ...)`), preserving all surrounding, interior, and leading punctuation. Extracted a `pigLatinWord` helper; the vowel/consonant logic is unchanged.

Normal inputs are unaffected (existing tests pass); only previously-corrupted payloads change. Added regression tests for interior/leading punctuation and a no-characters-dropped check.

### DIFF
--- a/src/redteam/strategies/otherEncodings.ts
+++ b/src/redteam/strategies/otherEncodings.ts
@@ -74,50 +74,33 @@ export function toMorseCode(text: string): string {
 }
 
 /**
- * Convert text to Pig Latin
+ * Convert one alphanumeric run to Pig Latin.
+ */
+function pigLatinWord(word: string): string {
+  if (!/^[a-zA-Z]/.test(word)) {
+    return word;
+  }
+
+  if (/^[aeiouAEIOU]/.test(word)) {
+    return word + 'way';
+  }
+
+  const vowelIndex = word.search(/[aeiouAEIOU]/i);
+
+  if (vowelIndex === -1) {
+    return word + 'ay';
+  }
+
+  const prefix = word.substring(0, vowelIndex);
+  const suffix = word.substring(vowelIndex);
+  return suffix + prefix + 'ay';
+}
+
+/**
+ * Convert text to Pig Latin while preserving punctuation in place.
  */
 export function toPigLatin(text: string): string {
-  // Split by spaces to get words
-  return text
-    .split(' ')
-    .map((word) => {
-      // Extract any trailing punctuation
-      const punctuationMatch = word.match(/([a-zA-Z0-9]+)([^a-zA-Z0-9]*)$/);
-      if (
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thank you so much for this, @orangeCatDeveloper — this is a genuinely excellent contribution. 🙌

You didn't just fix a cosmetic bug: the old end-anchored regex silently dropped everything before the last alphanumeric run of a token, so any payload with interior or leading punctuation (HTML/script, code, URLs, contractions) was corrupted *before it ever reached the target*. For a red team **strategy** that's the worst kind of failure — the target "safely" declined a request it never actually received, and the run looked like robustness to Pig Latin obfuscation. The writeup with the before/after table made the impact crystal clear, and the fix (`replace(/[a-zA-Z0-9]+/g, …)` transforming each run in place) is exactly the right, simpler approach — it even net-removes code.

I reviewed and verified this end-to-end:
- Confirmed the old code produced the corrupted `iptscray>` and the fix produces the full `<iptscray>alertway(1)</iptscray>`.
- Checked ~13 adversarial inputs (script/HTML, URLs with query strings, SQL injection, contractions, accented text, tabs/newlines, non-Latin, empty, digit-leading runs) — **no character is ever dropped**, and digit-leading runs are correctly left unchanged.
- `tsc`, Biome, and all tests are green.

I also pushed one small follow-up commit to strengthen the tests: the punctuation-preservation check now asserts the exact non-alphanumeric "skeleton" (identity + order + count) across several adversarial payloads instead of a per-character `toContain`, plus a strategy-level test proving the whole attack payload is delivered *encoded, not corrupted* through `addOtherEncodings`. Both new assertions fail against the pre-fix code, so they lock in the regression.

Really nice work spotting and fixing this — the adversarial instinct here (a "safe" result that was actually a swallowed attack) is exactly what we want. Thank you for making promptfoo better! Merging now. 🚀

PR #10027, opened 2026-07-08
fix(providers): preserve TrueFoundry filter errors

mldangelo-oai:
## Summary
- keep downstream `content_filter_error` responses as API errors, including nested codes and failure messages that overlap the guardrail matchers
- preserve real TrueFoundry/Azure guardrail blocks, including the `azure-foundry` blocked-label envelope and structured `filtered: true` results
- add regression coverage for every message heuristic plus nested and structured cases

## Why
Azure uses `content_filter_error` when content filtering fails or times out. A message such as `Unable to determine whether the response was filtered because the content filtering system timed out` previously matched the downstream guardrail heuristic, removed the operational error, and could make `not-guardrails` pass even though no guardrail result existed. Explicit filter-system failure codes now take precedence over message-only inference, while independent structured block evidence remains authoritative.

## Test plan
- `node node_modules/vitest/vitest.mjs run test/providers/truefoundry.test.ts test/assertions/guardrails.test.ts test/matchers/context-faithfulness.test.ts test/providers/openai-codex-sdk.test.ts --sequence.shuffle=true --sequence.seed=10027`
- `npm run tsc`
- `npm run l`
- `npm run f`
- local CLI loopback eval covering the real `azure-foundry` block, neutral and overlapping filter failures, nested failure code, and structured block


mldangelo:
Thank you for the fast, careful follow-up here — catching that Azure's `content_filter_error` (filter system down/timed out) was being normalized into a passing guardrail refusal is exactly the kind of subtle eval-integrity issue that's easy to miss, and the regression test quoting the real Azure failure mode makes this airtight. This closes the last known gap in the TrueFoundry guardrail handling before it ships in 0.121.19. Much appreciated! 🙏

src/providers/truefoundry.ts
```diff
@@ -38,11 +38,7 @@ type TrueFoundryProviderOptions = ProviderOptions & {
 type JsonRecord = Record<string, unknown>;
 
 const TRUEFOUNDRY_GUARDRAIL_ERROR_TYPE = 'guardrail_checks_failed';
-const DOWNSTREAM_GUARDRAIL_ERROR_CODES = new Set([
-  'content_filter',
-  'content_filter_error',
-  'content_policy_violation',
-]);
+const DOWNSTREAM_GUARDRAIL_ERROR_CODES = new Set(['content_filter', 'content_policy_violation']);
```
jameshiester-oai, line 41: **[P1] Make `content_filter_error` override message heuristics**

This removal only disables the code-based positive match; the same error still falls through to `DOWNSTREAM_GUARDRAIL_MESSAGE_PATTERNS`. For example, an HTTP 400 with `code: 'content_filter_error'` and message “Unable to determine whether the response was filtered because the content filtering system timed out” matches `response was filtered`. A local CLI wire test then returned no API error, set `isRefusal`/`guardrails.flagged`, and passed a `not-guardrails` assertion with score 1 even though no filter result existed. That preserves the false-positive eval behavior this PR is meant to fix. Short-circuit message-only inference when the explicit code is `content_filter_error` (while optionally retaining independent structured `filtered: true` evidence), and add an overlapping-wording regression test.

jameshiester-oai (commented):
Reviewed base `a643f8b8e13a93e89f1da229e3b77db3927bdf55` against exact head `25789a7aaae882b457a7759eaa986d15f80eec5f`.

Scope: fresh full review of the complete base...head diff, the surrounding TrueFoundry/OpenAI error-normalization path, predecessor context, assertions, and adjacent tests. I found one P1 eval-integrity defect. Exact-head CI is green: all 43 reported checks succeeded.

## Verified findings

### P1

- Removing `content_filter_error` from the positive code set does not make that machine-readable failure code authoritative. The later message heuristic still treats phrases such as “the response was filtered” as a block, even when the full error says the system could not determine that because filtering timed out. In a local CLI reproduction, that HTTP 400 lost its API error, became `isRefusal: true` / `guardrails.flagged: true`, and made a `not-guardrails` assertion pass with score 1. This leaves the exact false-positive eval behavior the PR intends to eliminate. See the inline comment for the smallest correction and regression case.

Microsoft documents `content_filter_error` as an error that means filtering did not complete and was not applied, distinct from an actual policy block: https://learn.microsoft.com/en-us/azure/foundry-classic/foundry-models/concepts/content-filter#scenario-6-content-filtering-system-unavailable

## Existing feedback checked

There are no inline comments on this PR. I read the Copilot and scanner reviews plus the predecessor guardrail-normalization context; this finding does not duplicate current feedback.

- Give explicit failure codes precedence over fuzzy message inference, then table-drive negative cases whose wording overlaps the block patterns. This makes the classification order visible and prevents future regex changes from reintroducing false positives.
- No other material simplification or behavioral test gap was found in this two-file change.

## Validation performed

- `node node_modules/vitest/vitest.mjs run test/providers/truefoundry.test.ts --sequence.shuffle=false`: 41/41 passed.
- `npm run tsc`: passed.
- `npx biome check src/providers/truefoundry.ts test/providers/truefoundry.test.ts` and `git diff --check`: passed.
- `npm run local -- eval -c /private/tmp/pr10027-loopback.yaml --no-cache -o /private/tmp/pr10027-overlap-output.json`: exercised a normal success, a real policy block, a neutral filter-system failure, and an overlapping-wording filter-system failure. The normal and policy paths behaved correctly; the neutral failure remained an error; the overlapping failure incorrectly passed `not-guardrails` with score 1. The command exited 100 because the deliberately retained neutral provider error was part of the fixture.

Limitations: I used a localhost wire-compatible fixture rather than live TrueFoundry/Azure credentials. The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable in this environment (`command not found`), so I used the six requested review lenses plus three independent read-only specialist reviews and reconciled their results here. The exact-head CI matrix supplied the broader cross-platform/build coverage.

jameshiester-oai:
Can we add a regression case for the actual TrueFoundry response shape below and make sure it remains a guardrail response rather than a generic API error?

```json
{
  "status": "failure",
  "message": "azure-foundry error: Response content blocked by label 'MultiSeverity_HateSpeechScore'.",
  "error": {
    "message": "azure-foundry error: Response content blocked by label 'MultiSeverity_HateSpeechScore'.",
    "type": "APIError",
    "code": "400"
  },
  "error_origin_level": "api_error",
  "provider": "azure-foundry"
}
```

This is a real downstream guardrail block, so it should normalize to `isRefusal: true` and `guardrails.flagged: true` instead of leaving `error` set. The `content_filter_error` carveout in this PR should not regress this case. The existing `Response content blocked by label` message matcher looks like the right signal, but a fixture for this envelope would lock in the behavior.

mldangelo-oai:
Rechecked exact head `25789a7aaae882b457a7759eaa986d15f80eec5f`.

The requested real `azure-foundry` fixture is already present verbatim in [`test/providers/truefoundry.test.ts:434-469`](https://github.com/promptfoo/promptfoo/blob/25789a7aaae882b457a7759eaa986d15f80eec5f/test/providers/truefoundry.test.ts#L434-L469), including `code: "400"`, the `Response content blocked by label` message, `isRefusal: true`, and `guardrails.flagged: true`. It passes.

The remaining eval-integrity gap is the earlier inline finding: an HTTP 400 with `code: "content_filter_error"` and a message such as `Unable to determine whether the response was filtered because the content filtering system timed out` still matches the message heuristic. A localhost wire reproduction returned no API error, set `isRefusal` and `guardrails.flagged`, and incorrectly passed `not-guardrails` with score 1. The explicit filter-system failure should take precedence over fuzzy message matching, while independent structured `filtered: true` evidence can still identify a real block. Please add that overlapping-wording regression case alongside the existing neutral failure case.

Validation on the reviewed head: 53 focused tests passed, TypeScript and Biome passed, `git diff --check` was clean, and all 43 exact-head CI checks succeeded. The PR is currently behind `main`.

jameshiester-oai (approved):
Great work!

PR #10028, opened 2026-07-08
feat(redteam): add FlipAttack jailbreak strategy

orangeCatDeveloper:
## Summary

Adds **FlipAttack**, a single-turn jailbreak strategy based on Liu et al., ["FlipAttack: Jailbreak LLMs via Flipping"](https://arxiv.org/abs/2410.02832) (ICML 2025). The attack disguises a request by flipping its word order or characters — a simple cipher that evades content filters — then wraps it in a guidance prompt telling the target to recover the original request and answer it.

## How it works

The strategy applies a deterministic flip to the payload and prepends recovery guidance. Three modes via a `mode` config option:

| Mode | Effect | Example (`how to pick a lock`) |
| --- | --- | --- |
| `char_in_sentence` (default) | Reverse every character | `kcol a kcip ot woh` |
| `word_order` | Reverse the order of the words | `lock a pick to how` |
| `char_in_word` | Reverse characters within each word | `woh ot kcip a kcol` |

```yaml
strategies:
  - flipattack
  # or with a mode:
  - id: flipattack
    config:
      mode: word_order
```

## Notes

- **Deterministic and local** — no attacker model or remote generation required.
- The recovery instruction always matches the applied flip, and whitespace (spaces, tabs, newlines) is preserved in place, so the payload stays recoverable by the target.
- `mode` is configurable from the web UI (redteam setup → Strategies → FlipAttack settings).
- Reuses the existing grader; only the payload wording changes.

## Testing

- Unit tests for all three flip modes, self-inverse (recoverability) checks including multi-line/tab payloads, prompt construction, and config handling.
- Web UI config dialog tests for the mode selector.
- Verified end-to-end via `redteam generate` with the local build.
Screenshot 2026-07-08 at 3 22 46 PM


mldangelo-oai (approved):
Release audit on 330fa7af2cdf3c99622e6e59b2356024395f4d3a: synced to current main, fixed Layer composition so configured FlipAttack modes are preserved, and reran backend/UI strategy coverage (96 tests). CI is green at the exact head; no release-blocking findings.

mldangelo-oai (changes requested):
Release audit on ce895f713035c9ba38344ae2bf5486ccda7027e7: CI and the focused strategy/layer suites are green, but please hold this feature for the current release. The latest exact-head finding is valid: the emitted word-order recovery instruction always tells the target to remove spaced pipe markers, so an ordinary shell/Markdown input such as foo | bar can be reconstructed as foobar. Emit marker-removal guidance only when the transform actually introduced unspaced-word boundaries (or use an unambiguous encoding), and add a regression that recovers through the emitted instruction. This follows several deliverability/exclusion edge cases found during the audit and warrants one more review pass before landing.

orangeCatDeveloper:
> Release audit on [ce895f7](https://github.com/promptfoo/promptfoo/commit/ce895f713035c9ba38344ae2bf5486ccda7027e7): CI and the focused strategy/layer suites are green, but please hold this feature for the current release. The latest exact-head finding is valid: the emitted word-order recovery instruction always tells the target to remove spaced pipe markers, so an ordinary shell/Markdown input such as foo | bar can be reconstructed as foobar. Emit marker-removal guidance only when the transform actually introduced unspaced-word boundaries (or use an unambiguous encoding), and add a regression that recovers through the emitted instruction. This follows several deliverability/exclusion edge cases found during the audit and warrants one more review pass before landing.

Fixed on e9599c57 — marker guidance is now emitted only when the flip actually inserted markers, with a regression that recovers through the emitted instruction (verified it fails on the old behavior with `expected 'foobar' to be 'foo | bar'`).

mldangelo (commented):
## Heads-up: I pushed a commit to your branch

So you're not surprised by a commit you didn't write — I pushed **`926ffa8` `fix(redteam): stop dedent from corrupting FlipAttack payloads`** on top of your work. Revert it if you disagree.

`buildFlipAttackPrompt` interpolated `${flipped}` inside the `dedent` tagged template. `dedent@1.7.2` interpolates **first**, then runs indentation-stripping, `.trim()`, and an unescape pass (`\n`, `\t`, `\xHH`) over the *joined* string — so all three passes rewrote the attacker payload, not just your literal template:

```
mode: word_order
payload  : Write a script that prints "hello\nworld" and reads C:\temp\x41   (literal backslashes)
delivered: C:\tempA reads and "hello⏎world" prints that script a Write       (\x41 → A, \n → newline)
recovered: Write\ta script that prints⏎world" "hello and reads empA C:       ≠ payload
```

Two more vectors: an indented payload hijacked dedent's `mindent` (2 instead of the template's 4), emitting every `Rule N.` line with two stray leading spaces; and `.trim()` ate edge whitespace. Over an 11-payload corpus: 6/11 corrupted in `word_order`, 4/11 in `char_in_word`, 2/11 in the **default** `char_in_sentence`.

The fix assigns the static guidance to a `const` and appends `\n\nTASK: ${flipped}` outside the template. Output is byte-identical for ordinary payloads (side-by-side, 4 payloads × 3 modes). Revert-proof: reverting only the source leaves the new tests red — `Tests 4 failed | 12 passed (16)`; restored → green. Current tip: `28 passed`, CI fully green.

## What's good

Genuinely strong PR. The whitespace-preserving `split(/(\s+)/)` even/odd-index tokenization is a nice touch — most implementations `split(' ')` and destroy tabs and newlines. Pairing each mode's recovery instruction with the flip *and* commenting why they must match is exactly the invariant that makes this strategy work. Grapheme-safe `reverse`, the `Intl.Segmenter` fallback for unspaced languages, and full end-to-end wiring (constants, riskScoring, layer, UI dialog + 50 tests, docs page, sidebar, schema) is more thoroughness than most strategy PRs bring. Your responsiveness across 11 commits has been great.

## Remaining (all optional)

**1. Hand-rolled `Intl.Segmenter` shim is unnecessary — `flipAttack.ts:17-24`** *(nit)*
Repo tsconfig is `target: ES2022` with no `lib` override, so the built-in types are already present. Verified on the repo's TS 6.0.3:
```
$ npx tsc --noEmit --ignoreConfig --target ES2022 --strict segprobe.ts   # new Intl.Segmenter(...) + .isWordLike
EXIT=0
```
Worth deleting: the shim types `isWordLike?: boolean` as optional where the real API has it non-optional.

**2. The three modes are spelled out five times** *(nit)*
`FlipMode` union (`:12`), `flip()` switch (`:84`), `recoveryInstruction()` switch (`:105`), the tuple at `:141`, and a fifth copy in `StrategyConfigDialog.tsx:1303-1305` + the `SelectItem`s at `:1323-1325`. Both switches use `default:` for `char_in_sentence`, so TS can't flag a missing case if a mode is added, and the dialog silently coerces unknown modes to `char_in_sentence` and writes that back on Save. Exporting `FLIP_ATTACK_MODES` from `src/redteam/constants/strategies.ts` fixes both — the dialog already imports from `@promptfoo/redteam/constants/strategies` (`:33`) and its other Selects (`:1105`, `:1245`) map over imported constants, so no new module edge.

**3. Docs snippets won't validate — `flipattack.md:15-27`**
Both blocks are labelled `title="promptfooconfig.yaml"` but put `strategies:` at the top level. The generated schema root has `additionalProperties: false` and no `strategies` key, so a copy-paste errors and FlipAttack never runs. Nesting under `redteam:` (as `basic.md` / `hydra.md` / `layer.md` do) fixes it. Fair caveat: ~20 of 34 strategy docs have the same bug, so this is really a repo-wide cleanup — just nice for a new page to start correct.

**One question:** `metadata` (`:163`) records `strategyId`, `flipMode`, and `originalText` but not whether the `' | '` marker path was taken. Was that intentional? Since the delivered recovery instruction differs in that case, recording it would let someone triaging a result tell which prompt the target actually saw.

orangeCatDeveloper:
@mldangelo-oai Gentle bump. The branch is also synced with the latest `main`. Could you take another look?

PR #10032, opened 2026-07-09
feat(providers): add xAI Grok 4.5 support

mldangelo:
## Summary

Adds xAI Grok 4.5 support while preserving region-safe defaults and authoritative cost reporting.

- **Models and aliases:** adds `grok-4.5`, `grok-4.5-latest`, and `grok-build-latest`, plus the current `grok-latest` alias for Grok 4.3.
- **Capabilities:** tracks Grok 4.5 as a 500K reasoning model; chat and Responses requests accept only `low`, `medium`, or `high` reasoning effort and reject unsupported values locally.
- **Pricing:** adds catalog pricing ($2/M input, $0.50/M cached input, $6/M output; higher tier at 200K input tokens) and prefers xAI's authoritative `usage.cost_in_usd_ticks` when present. Explicit Promptfoo cost overrides still take precedence.
- **Billing correctness:** handles xAI chat reasoning tokens as separately billed, preserves cost on refusal/content-filter responses, reports cached Promptfoo responses as zero incremental cost, and leaves cost unknown when no billing evidence exists.
- **Region safety:** keeps automatic grading, synthesis, suggestion, web-search, and red-team defaults on Grok 4.3 because [Grok 4.5 is currently unavailable to EU API Console users](https://docs.x.ai/developers/grok-4-5). Grok 4.5 remains available by explicit selection, and the default example keeps it commented out as an opt-in provider.
- **Docs and examples:** documents Grok 4.5, exact tick billing, inclusive long-context pricing, current canonical models, EU availability, and xAI's legacy redirects. Corrects Grok Code Fast aliases to Grok Build 0.1, removes stale model/pricing claims, and keeps the quick start runnable in the EU.

## Review fixes

This revision addresses all correctness findings raised during review:

- rejects `reasoning_effort: none` and other unsupported Grok 4.5 values;
- preserves xAI's exact billed amount instead of overwriting it with catalog estimates;
- avoids making an EU-unavailable model the unconditional default;
- makes Grok 4.5 opt-in in the example so the default quick start remains runnable for EU accounts.

It also covers follow-on edge cases found during review: exact costs on refusals, custom pricing precedence, cached Responses cost, unknown-cost handling, cached-token clamping, and the inclusive long-context threshold defined by xAI's REST schema.

## Verification

Run on the final branch head after merging current `main`:

- 660 focused backend tests passed across 12 files (OpenAI chat, xAI providers, defaults, web-search fallback, registry, shared cached-token handling, DeepSeek, and Azure chat).
- 36 affected frontend tests passed.
- `npm run tsc` passed.
- `npm run knip -- --no-progress` passed with only the existing configuration hint.
- `npm run build` passed, including the application and postbuild steps.
- `SKIP_OG_GENERATION=true npm --prefix site run build` passed.
- `npm run local -- validate config -c examples/xai/chat/promptfooconfig.yaml` reported a valid configuration.
- Prettier and `git diff --check` passed for the final example/docs changes.
- Final GitHub CI passed all 30 jobs, including Style/Knip, docs, Code Scan, builds, web UI, smoke, and the full OS/Node test matrix.

The original branch validation recorded a successful live xAI eval. The review checkout did not contain `XAI_API_KEY`, so the post-fix provider QA uses deterministic HTTP fixtures and captured billing examples rather than making another authenticated request.


src/providers/xai/chat.ts
```diff
@@ -696,8 +741,9 @@ class XAIProvider extends OpenAiChatCompletionProvider {
...
         // The OpenAI base provider normalizes xAI's OpenAI-compatible usage
         // details but does not surface `usage.cost_in_usd_ticks`. Fall back to
-        // local pricing math. Completion tokens already include reasoning tokens,
-        // so they are not added on top.
+        // local pricing math. The chat-completions endpoint reports reasoning
+        // tokens separately from completion tokens and bills them at the output
+        // rate (see calculateXAICost), so they are added on top here.
         const reasoningTokens = response.tokenUsage.completionDetails?.reasoning || 0;
         const cachedTokens = response.tokenUsage.completionDetails?.cacheReadInputTokens ?? 0;
         response.cost = calculateXAICost(
```
jameshiester-oai: **[P2] Prefer xAI's exact billed cost for chat responses**

xAI exposes the final charge on every inference response as `usage.cost_in_usd_ticks`, including billing adjustments, and [priority processing uses premium rates](https://docs.x.ai/developers/advanced-api-usage/priority-processing). The base normalizer drops that field, and this block then overwrites `response.cost` with standard catalog math. A local provider-path probe with 100 input/10 output tokens and `123456789` ticks returned `$0.00026` here instead of the authoritative `$0.0123456789`. Preserve the raw usage cost and prefer it, as the Responses provider already does, falling back to local math only when ticks are absent. See xAI's [cost-tracking contract](https://docs.x.ai/developers/cost-tracking).

src/providers/xai/defaults.ts
```diff
@@ -4,7 +4,7 @@ import { XAIResponsesProvider } from './responses';
 import type { EnvOverrides } from '../../types/env';
 import type { DefaultProviders } from '../../types/index';
 
-const DEFAULT_XAI_MODEL = 'grok-4.3';
+const DEFAULT_XAI_MODEL = 'grok-4.5';
```
jameshiester-oai: **[P2] Keep the default on a model available to EU users**

[xAI's Grok 4.5 launch notice](https://x.ai/news/grok-4-5) says the model is not yet available in the EU and expects availability only in mid-July. Making it unconditional here causes default grading, synthesis, and suggestion providers to fail for current EU accounts; this PR also applies the same default to web search and the red-team selector. Keep Grok 4.3 as the global default until 4.5 is available in every supported region, or select/fallback by region and document the limitation.

jameshiester-oai (commented):
## Review summary

**Reviewed base:** `cc7ff984f9b634cb99ca065352b71fa74a53c4b4`  
**Reviewed head:** `e54a1e5aa3e70b980d47daf4ebedf5a2ef026439`  
**Scope:** Full fresh review of the current PR diff; no prior authenticated completion marker existed for this head.

**Verdict:** Two distinct P2 correctness/compatibility defects remain. The current CI snapshot is otherwise green: all 40 reported checks pass.

### Verified findings

- **P2 — Chat cost reporting discards xAI's exact billed amount.** The shared OpenAI usage normalizer drops `usage.cost_in_usd_ticks`, then the xAI chat wrapper always replaces `response.cost` with public catalog math. Priority-tier requests and any discounted or tool-adjusted billing can therefore be reported at the wrong amount. The inline comment includes a deterministic reproduction.
- **P2 — Grok 4.5 is made the unconditional default while it is unavailable in the EU.** This changes default grading, synthesis, suggestions, web search, and the red-team UI to a model that current EU users cannot call. The inline comment links xAI's launch availability statement.

### Existing unresolved feedback

The current head also forwards `reasoning_effort: none` for Grok 4.5 even though that model rejects the value. Two existing unresolved threads already describe that P2 defect, so I validated it but did not duplicate it.

- The Responses provider already centralizes the right precedence—prefer `getXAICostInUsd(usage)`, then fall back to `calculateXAICost`. Reusing the same rule for chat would reduce drift between the two APIs.
- `ProviderTypeSelector.test.tsx` has no direct xAI selection assertion, so the changed UI default is currently covered only indirectly.
- I did not file the exact 200,000-token tier boundary as a finding: current xAI documentation conflicts between “exceed” and “at or above,” and I had no authenticated boundary response to resolve it.

### Validation performed

- Requested PR Review Toolkit invocation: `/pr-review-toolkit:review-pr all parallel` — unavailable on this host (exit 127). I used the manual correctness, tests, docs/comments, error-handling, type-design, and simplification fallback plus three independent read-only reviews.
- `source ~/.nvm/nvm.sh && nvm use && node node_modules/vitest/vitest.mjs run test/providers/xai test/providers/defaults.test.ts test/providers/webSearchUtils.test.ts test/providers/registry.test.ts --sequence.shuffle=false --reporter=dot` — 8 files, 422 tests passed.
- `source ~/.nvm/nvm.sh && nvm use && npm run test:app -- src/pages/redteam/setup/components/Targets/ProviderTypeSelector.test.tsx --run` — 1 file, 36 tests passed.
- `PROMPTFOO_CONFIG_DIR=/tmp/pr10032-config PROMPTFOO_DISABLE_TELEMETRY=true npm run local -- validate config -c examples/xai/chat/promptfooconfig.yaml` — configuration valid.
- `source ~/.nvm/nvm.sh && nvm use && npm run build` — core, declarations, and app build passed.
- `SKIP_OG_GENERATION=true npm --prefix site run build` — production docs build passed.
- `source ~/.nvm/nvm.sh && nvm use && node --import tsx /tmp/pr10032-cost-ticks-probe.ts` — deterministic no-credential provider-path probe reproduced `$0.0123456789` in exact ticks being replaced by `$0.00026` catalog math.
- `git diff --check cc7ff984f9b634cb99ca065352b71fa74a53c4b4...e54a1e5aa3e70b980d47daf4ebedf5a2ef026439` — passed.

### Limitations

I did not use an authenticated xAI key, an EU-region account, or browser-interactive UI QA. The provider and UI paths were exercised with local deterministic tests/probes, and current GitHub CI was inspected separately.

PR #10033, opened 2026-07-09
feat(assertions): add ends-with assertion (symmetric to starts-with)

WatchTree-19:
### what
adds an `ends-with` deterministic assertion - the symmetric counterpart to the existing `starts-with`. checks that the model output ends with a given string (and `not-ends-with` works via the generic `not-` inverse).

```yaml
assert:
  - type: ends-with
    value: 'END'
```

### why
`starts-with` exists but there's no `ends-with`, which is a natural, frequently-wanted check (e.g. asserting a response terminates with a required token, closing tag, or stop marker). it's a small, symmetric addition.

### changes
mirrors `startsWith` exactly:
- `src/assertions/endsWith.ts` - handler (`String#endsWith`, same invariants/reason format as startsWith)
- `src/assertions/index.ts` - import + handler-map registration
- `src/types/index.ts` - assertion type union
- UI `assertionValueValidation.ts` - base types + required-string set (+ `not-ends-with`)
- `src/csv.ts` - CSV default-threshold parity
- `test/assertions/runAssertion.test.ts` - pass + fail tests
- docs `deterministic.md` - reference table row + section

no existing behaviour changes; `not-ends-with` is handled by the existing `not-` prefix logic.


mldangelo-oai (approved):
Release audit on d31e2ac5ff199696a5d9f9fc6f429ea10816c49f: synced to current main, exposed ends-with/not-ends-with in the eval creator and docs, and validated backend/UI coverage (184 tests) plus a passing CLI eval. CI is green at the exact head; no release-blocking findings.

mldangelo (commented):
Thanks for this — clean, well-scoped addition. One heads-up first so you aren't surprised by a commit you didn't write.

## Pushed to your branch

`1c67ecc185 test(assertions): pin ends-with to suffix matching` — one additive test in `test/assertions/runAssertion.test.ts`. No production code touched, no existing test modified.

Why: your two `ends-with` tests both use `value: 'output'`, against `'Expected output'` (pass) and `'Different result'` (fail). Both outcomes are identical under an `includes()` implementation, so neither test pins suffix semantics. Mutating `src/assertions/endsWith.ts:16` from `.endsWith()` to `.includes()` left everything green:

```
# with the includes() mutant in place
$ npx vitest run test/assertions/ test/csv.test.ts
Test Files  61 passed (61)
Tests     1449 passed (1449)
```

The added case (`value: 'Expected'` against output `'Expected output'` → `pass: false`) fails against that mutant and passes against your code — currently the only test in the repo that does. Rewrite or drop it if you'd rather do it your own way.

## What works well

- **The implementation is correct end-to-end.** Verified through the real CLI with an echo provider: `"Here is the answer. END"` passes, `"END of preamble; here is the real answer..."` fails with `Expected output to end with "END"`, and `not-ends-with` inverts correctly.
- **You found every registration site** — enum, handler map, CSV parser, eval-creator validation — and `site/static/config-schema.json` is genuinely regenerated: re-running `npm run jsonSchema:generate` produces zero diff.
- **The `architecture/edge-baseline.json` bump is correct**; `npm run architecture:check` passes. That ratchet trips up a lot of PRs.
- Good follow-up on the `AssertsForm.tsx` dropdown and the index.md tables.

## Remaining

### 1. P2 — the new CSV shorthand has no test

`src/csv.ts:214` makes `ends-with:END` a real user-facing surface in `__expected` columns, but `test/csv.test.ts` contains no `ends-with` at all: no peer to the `starts-with:Expected` case at line 709, and the default-threshold list at line 883 — which enumerates every type in that branch — wasn't extended.

### 2. Question — was the `threshold` default intentional?

Adding `ends-with` to that branch attaches a threshold `handleEndsWith` never reads:

```
ends-with:END      -> {"type":"ends-with","value":"END","threshold":0.75}
ends-with(0.9):END -> {"type":"ends-with","value":"END","threshold":0.9}
```

`test/csv.test.ts:896` already flags this as a known wart ("unlike e.g. `starts-with`, which receives the 0.75 default but ignores it"). Dropping `|| type === 'ends-with'` wouldn't break parsing — `getAssertionRegex()` (`src/csv.ts:104`) builds its pattern from `BaseAssertionTypesSchema.options`, so the enum addition alone is sufficient, and the `else` branch returns the same object minus the dead field. Since `ends-with` is new there's no prior shape to preserve. Was matching `starts-with` here deliberate?

## Nits (non-blocking)

- The new row in `deterministic.md` landed after `similar`; the rest of that table is alphabetical, so it belongs between `cost` and `equals`.
- `not-ends-with` is auto-derived and reachable in configs but has no backend test. `not-starts-with` doesn't either, so this is inherited rather than introduced — flagging only in case you'd like to close both.

WatchTree-19:
Thanks — this is a genuinely useful review, and the mutation test is a fair hit. You're right that neither of my `ends-with` cases pinned suffix semantics: both `'Expected output'` (pass) and `'Different result'` (fail) behave identically under an `includes()` implementation, so the tests were passing for the wrong reason. Happy to keep `1c67ecc185` as-is — it's the right test and it's the only one in the repo that kills that mutant.

## Threshold default — you're right, dropped it

Not deliberate: I mirrored `starts-with` for symmetry without checking whether the field was actually read. It isn't — `handleEndsWith` never touches `threshold`, so the shorthand was attaching a value that does nothing. Since `ends-with` is new there's no prior shape to preserve, so I've removed `|| type === 'ends-with'` from that branch (`3d54371`). It now lands in the `else` alongside `contains` / `equals` / `regex`, the other deterministic string asserts, which is where it belongs.

Confirmed your reasoning about parsing holding up — verified against the real parser rather than by inspection:

```
ends-with:END          -> type=ends-with     value="END" threshold=undefined
ends-with(0.9):END     -> type=ends-with     value="END" threshold=undefined
not-ends-with:END      -> type=not-ends-with value="END" threshold=undefined
starts-with:BEGIN      -> type=starts-with   value="BEGIN" threshold=0.75
```

One wrinkle worth flagging, since it's a behaviour change rather than pure dead-code removal: the `else` branch omits `threshold` entirely, so `ends-with(0.9):END` now silently **drops** an explicitly-supplied threshold rather than carrying it through. That's consistent with how `contains(0.9):x` already behaves, and the value was inert either way, but it's not strictly "same object minus the dead field" — so I've pinned it with a test that documents the parenthesised form still *parses* while the value is discarded. If you'd rather it were rejected outright at parse time, that's a broader change across all the else-branch types and probably its own PR.

`starts-with` is untouched — I left its 0.75 default alone as pre-existing back-compat rather than widening scope here.

## The rest

- **P2, CSV shorthand tests** — these went in with `cc8ba06` just before your review landed, so they may have crossed: there's now a peer to the `starts-with:Expected` case, plus `not-ends-with`. I've removed `ends-with` from the default-threshold list test to match the change above.
- **Nit, docs table order** — also fixed in `cc8ba06`; the row now sits between `cost` and `equals`. I left the `### Ends-With` body section next to `### Starts-With`, since those aren't alphabetised.
- **Nit, `not-ends-with` backend test** — added the CSV-level one here. I've deliberately not touched `not-starts-with`: it's pre-existing and unrelated to this change, so it feels like it belongs in its own PR rather than riding along. Happy to do that separately if useful.

WatchTree-19:
@mldangelo-oai both of those are covered at `3d54371`, I just never came back to say so.

On the CSV shorthand, `cc8ba06` added `should create an ends-with assertion` in `test/csv.test.ts`, mirroring the `starts-with` case directly above it.

On the threshold, it was not deliberate, it was a leftover, and `3d54371` removes it rather than documenting it. `handleEndsWith` never reads `threshold`, so synthesising a default was dead state that would only mislead anyone reading the parsed assertion. `ends-with` is therefore left out of the default-threshold branch in `src/csv.ts` and out of the enumeration test, and there is a case pinning it: `should not attach a dead threshold to ends-with` asserts `threshold` comes back undefined. I left `starts-with` alone since that shape predates this and changing it would be a behaviour change for existing configs.

One thing worth a look from your side: the branch is out of date with main and CodeQL is sat at "Expected, waiting for status to be reported", which I think is why this has not moved. Happy to update the branch if that would unstick it.

PR #10041, opened 2026-07-10
feat(providers): add Meta Model API (Muse Spark) provider

mldangelo:
## Summary

Adds a `meta:` provider for Meta's new [Model API](https://dev.meta.ai/) (public preview, announced 2026-07-09), which serves the Muse Spark reasoning models. All three API surfaces are supported:

- **`meta:muse-spark-1.1`** / `meta:responses:<model>` — Responses API, the default (web-search grounding, cross-turn reasoning; Meta's recommended surface)
- **`meta:chat:<model>`** — OpenAI-compatible chat completions
- **`meta:messages:<model>`** — Anthropic-compatible Messages endpoint (the surface Claude Code-style agents use), authenticated with a bearer token

Unsupported sub-types (`embedding`, `completion`) fail fast with a clear error.

## Provider-specific value (beyond `openai` + `apiBaseUrl`)

- **Key resolution & isolation**: reads Meta's official `MODEL_API_KEY` env var (plus config `apiKey`/`apiKeyEnvar`); never falls back to `OPENAI_API_KEY`/`ANTHROPIC_API_KEY`, suppresses the OpenAI organization header, refuses Claude Code OAuth credentials, and strips `ANTHROPIC_CUSTOM_HEADERS`-derived headers from Meta traffic.
- **Reasoning-model body shaping**: forwards `reasoning_effort` (`minimal`…`xhigh`; rejects unsupported values like `none` with an actionable error instead of one HTTP 400 per test), maps `max_tokens` → `max_completion_tokens` (→ `max_output_tokens` on Responses), fails fast on unsupported `stop`/`logprobs`, and keeps OpenAI-scoped env tuning (`OPENAI_TEMPERATURE`, `OPENAI_TOP_P`, penalties, caps) out of Meta requests.
- **Built-in cost table** from published pricing ($1.25 / $0.15 cached / $4.25 per 1M tokens), including cached-input billing on all three usage formats; user `cost`/`inputCost`/`outputCost`/`cacheReadCost` overrides win, including prompt-level overrides.
- **Messages-surface UX**: Muse reasoning arrives as encrypted `redacted_thinking` blocks; `showThinking` defaults to `false` there so ciphertext stays out of graded output.

Also adds an opt-out static flag (`WARNS_ON_UNKNOWN_MODEL`) to `AnthropicMessagesProvider` so Anthropic-wire subclasses serving non-Anthropic catalogs skip the unknown-model warning (no behavior change for existing providers).

## Docs & example

- `site/docs/providers/meta.md` — setup, models, config options, cost tracking, all three surfaces, and vetted recipes for pointing `openai:codex-sdk` and `anthropic:claude-agent-sdk` at Muse Spark (both recipes set `apiKey` explicitly; without it those providers would transmit an ambient `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` to the Meta endpoint — the Claude Code variant was reproduced empirically during review)
- `examples/provider-meta/` — runnable comparison of `reasoning_effort: minimal` vs `high`
- Provider index table row; `config-schema.json` regenerated from the Zod env schema

## Testing

- 63 unit tests (`test/providers/meta.test.ts`) + registry routing assertions for every prefix/sub-type pair; full `test/providers/` suite green (6,929 tests); tsc/lint/architecture checks pass
- **Live-verified against api.meta.ai** with a real key:
  - Chat: example passes 4/4; `reasoning_effort` differentiates (minimal ≈ 250 reasoning tokens/2.6s vs high ≈ 700/6s); computed cost matches billed usage to the token, including cached-input pricing
  - Responses: reasoning + `web_search` grounding (model performed a live search with citations)
  - Messages: bearer auth works; `redacted_thinking` hidden from graded output
  - Structured output: strict `json_schema` returns conforming JSON
  - Negative paths: fake key → clean 401 surfaced in results on both OpenAI-compatible and Anthropic-compatible surfaces

## Related

Meta retired the Llama API Public Preview on July 6, 2026 ([deprecation notice](https://llama.developer.meta.com/docs/llama-api-deprecation/)); this PR adds a deprecation warning to `llamaApi.md` pointing at third-party Llama hosts and positions the Model API as its successor.


jameshiester-oai (commented):
Reviewed base `a643f8b8e13a93e89f1da229e3b77db3927bdf55` against exact head `6e68d9d6cf824f23ea43dcb64310ca48e56b7cd3`.

Scope: fresh full review of the complete base...head diff and surrounding provider, schema, registry, documentation, and test paths. I found three new actionable issues: one P1 credential-boundary defect and two P2 routing/authentication defects. Exact-head CI is green: 44 checks succeeded and 3 were skipped.

## Verified findings

### P1

- The new Claude Code recipe routes the Agent SDK to Meta but leaves ambient `ANTHROPIC_CUSTOM_HEADERS` intact. `ClaudeCodeSDKProvider` copies all of `process.env`, and the Anthropic SDK applies those headers to the configured Meta base URL. With fake values, the built request to `https://api.meta.ai/v1/messages` included `X-Proxy-Secret: anthropic-only`. Anyone who uses that variable for an Anthropic gateway can disclose its credential to Meta by following this recipe. See the inline comment for the minimal documentation fix.

### P2

- `meta:messages` converts every header named by `ANTHROPIC_CUSTOM_HEADERS` into a null default to suppress it. If that list includes `Authorization`, the SDK merges the null after its generated Meta bearer header and removes authentication. A local wire probe observed no `Authorization` header. See the inline comment.
- Explicit `config.apiHost` is preserved but ignored by both the Meta chat and Responses providers. A direct probe with `apiHost: proxy.example.test` returned `https://api.meta.ai/v1` for both. This silently bypasses an explicitly configured gateway or policy proxy. See the inline comment.

## Existing feedback checked

The current-head threads about empty-string `apiKey` fallback, Anthropic-scoped sampling defaults on `meta:messages`, unsupported subtype fallthrough, and templated chat `reasoning_effort` all still reproduce, so I have not duplicated them. The outdated missing-key-precheck thread also remains relevant: it recommends `META_API_KEY` while the documentation recommends `MODEL_API_KEY`. The two outdated `passthrough.max_tokens` comments do not describe a separate current regression after tracing the base providers' passthrough behavior.

- Use one Meta endpoint resolver for chat and Responses so explicit host/base overrides cannot diverge.
- Add final-wire header tests for Meta Messages, including case-insensitive `Authorization` collisions, instead of asserting only constructor options.
- Add loopback transport coverage for all three surfaces; the current Responses and Messages cost tests stub their superclass and cannot catch URL/header integration failures.

## Validation performed

- `npx vitest run test/providers/meta.test.ts test/providers/registry.test.ts`: 154/154 passed.
- `npx vitest run test/providers`: 212 files passed, 1 skipped; 7,030 tests passed, 9 skipped.
- `npm run tsc`, `npm run architecture:check`, and `npm run build`: passed.
- `cd site && SKIP_OG_GENERATION=true npm run build`: passed.
- Biome on all changed code/tests: no errors; two pre-existing complexity warnings. Prettier checks on changed docs/example/schema files passed.
- Local CLI with `--no-cache` and exported JSON exercised chat, Responses, and Messages success plus HTTP 400 paths. All three success paths returned expected output, usage, and cost; all three error paths surfaced the server error.
- Separate local CLI wire checks confirmed chat request shape/cached-token pricing and Messages bearer authentication. Direct SDK/header and `apiHost` probes reproduced the three findings above.

Limitations: no real Meta credential was used, so I did not call the hosted Meta service or exercise hosted streaming, tools/search, multimodal input, or a live Claude Code task. Meta's detailed feature documentation was login-gated during review. The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable in this environment (`command not found`), so I used the six requested review lenses plus three independent read-only specialist reviews and reconciled their results here.

site/docs/providers/meta.md
```diff
@@ -0,0 +1,171 @@
...
+### Claude Code
+
+The [`anthropic:claude-agent-sdk` provider](/docs/providers/claude-agent-sdk/) forwards `config.env` to the agent subprocess. Pass the environment Meta's guide prescribes, with `apiKey` set to your Meta key:
+
+```yaml
+providers:
+  - id: anthropic:claude-agent-sdk
+    config:
+      apiKey: '{{env.MODEL_API_KEY}}'
+      env:
```
jameshiester-oai, line 127: **[P1] Clear Anthropic custom headers before routing Claude Code to Meta**

This recipe overrides the base URL and token, but `ClaudeCodeSDKProvider` still copies every ambient variable from `process.env`. If the shell has `ANTHROPIC_CUSTOM_HEADERS` set for an Anthropic gateway, the Agent SDK applies those headers to `api.meta.ai`; a fake-value request build reproduced `X-Proxy-Secret: anthropic-only` on the Meta request. That can disclose a gateway credential across vendors simply by following the documented setup. Add `ANTHROPIC_CUSTOM_HEADERS: ''` to this `env` block and call out why it must be cleared.

src/providers/meta.ts
```diff
@@ -0,0 +1,608 @@
...
+  // gateway/proxy secrets) and must not be sent to Meta; a null value tells
+  // the SDK to drop the header.
+  protected override buildAnthropicClientOptions(options: ClientOptions): ClientOptions {
+    const suppressedEnvHeaders = getAnthropicEnvHeaderSuppressions();
+    return {
+      ...options,
+      apiKey: null,
+      authToken: this.apiKey ?? null,
+      ...(Object.keys(suppressedEnvHeaders).length > 0
+        ? { defaultHeaders: { ...options.defaultHeaders, ...suppressedEnvHeaders } }
```
jameshiester-oai: **[P2] Preserve Meta bearer auth when suppressing ambient headers**

If `ANTHROPIC_CUSTOM_HEADERS` names `Authorization`, `getAnthropicEnvHeaderSuppressions()` contributes `Authorization: null` here. The Anthropic SDK merges `defaultHeaders` after the bearer header generated from `authToken`, so that null removes the Meta credential; a local wire probe observed no `Authorization` header. Reassert the Meta bearer value for this collision (while continuing to suppress the ambient value), and add a final built-request test with case-insensitive `Authorization` spelling.

src/providers/meta.ts
```diff
@@ -0,0 +1,608 @@
...
+
+    this.config = resolvedConfig;
+  }
+
+  override getApiKey(): string | undefined {
+    return resolveMetaApiKey(this.config, this.env);
+  }
+
+  override getApiUrl(): string {
+    return this.config.apiBaseUrl || META_API_BASE_URL;
```
jameshiester-oai: **[P2] Honor an explicitly configured `apiHost`**

`resolveMetaOpenAiConfig()` preserves `config.apiHost`, but this override never reads it. A direct probe with `apiHost: proxy.example.test` returned `https://api.meta.ai/v1`; `MetaResponsesProvider.getApiUrl()` has the same behavior. This silently bypasses a gateway or policy proxy the user explicitly configured even though the provider advertises the OpenAI options. Resolve explicit `apiHost` to `https://<host>/v1` for both providers while still ignoring ambient OpenAI host variables.

site/docs/providers/meta.md
```diff
@@ -0,0 +1,173 @@
...
+        ANTHROPIC_BASE_URL: https://api.meta.ai
```
jameshiester-oai, line 128: **P1 — Keep suite-level Anthropic routing from overriding the Meta endpoint.** `ClaudeCodeSDKProvider` merges `process.env < config.env < this.env`, and `loadApiProvider` places the top-level eval `env` into that final `this.env` layer. With this recipe plus `env.ANTHROPIC_BASE_URL: https://anthropic-gateway.example`, an exact-head fake-key probe captured the child environment using the gateway URL while both credential variables contained the Meta key. Following the recipe can therefore disclose the Meta credential to an Anthropic gateway. The existing custom-header comment covered ambient `process.env`; the added empty value fixes that case but not this higher-precedence layer. Pin `ANTHROPIC_BASE_URL` in the provider-level `env` layer that wins over suite env, or validate/scrub these safety-critical values after the final merge, and add a conflicting-suite-env regression test.

src/providers/meta.ts
```diff
@@ -0,0 +1,635 @@
...
+  }
+  if ('logprobs' in body) {
+    throw new Error('Muse Spark models do not support logprobs.');
+  }
+  if (typeof body.n === 'number' && body.n > 1) {
+    throw new Error(
+      'Muse Spark models do not support `n > 1`; remove `n` from the provider config.',
+    );
+  }
+  if (body.stream) {
```
jameshiester-oai: **P2 — Reject normal Chat `stream` configuration instead of silently ignoring it.** The OpenAI Chat base builder does not copy `config.stream` into the request body, so this check only catches `passthrough.stream`. A direct probe with `meta:chat` and `config: { stream: true }` returned `config.stream === true` with `body.stream === undefined`, while the equivalent Responses call threw this documented error. That makes a requested streaming Chat eval run non-streaming without warning. Check `config.stream` as well as `body.stream`, and cover the top-level Chat configuration path.

src/providers/meta.ts
```diff
@@ -0,0 +1,635 @@
...
+    response.cost = cost;
+  }
+  return response;
+}
+
+// The Meta Model API (dev.meta.ai) serves the Muse Spark models through an
+// OpenAI-compatible chat completions endpoint, so the standard chat provider
+// handles requests once the base URL and key resolution point at it.
+// https://dev.meta.ai/docs/features/chat-completion
+class MetaProvider extends OpenAiChatCompletionProvider {
```
jameshiester-oai, line 311: **P2 — Attribute Meta spans to the Meta provider.** This class inherits `getGenAISystem() === 'openai'`; `MetaResponsesProvider` likewise inherits a call path that hardcodes `system: 'openai'`, and `MetaMessagesProvider` inherits `system: 'anthropic'`. The tracing docs define `gen_ai.system` as the provider system, so every traced Meta call is grouped under another vendor in filters, dashboards, and alerts. Add overridable system hooks where needed, return `meta` from all three Meta subclasses, and assert the emitted span attributes.

jameshiester-oai (commented):
## Incremental re-review summary

**Reviewed base:** `bafcbf23f77c34aead99f43cf06c65cedc384517`  
**Reviewed head:** `6a8c72caa936df2dcc15b15b6d970fe6a8b1026c`  
**Scope:** Incremental review from the previously completed ancestor `6e68d9d6cf824f23ea43dcb64310ca48e56b7cd3`, plus a full current base...head interaction and regression review.

### Verdict and CI

The incremental commits fix the earlier API-host, bearer-auth, custom-header, environment-default, templating, and request-validation defects. I found one new P1 credential-boundary defect and two new P2 behavioral/observability defects. Exact-head CI is green: 40 checks passed and 3 Docker publication checks were intentionally skipped.

### Verified findings

#### P1

- The Claude Agent recipe places its Meta endpoint under `config.env`, but `ClaudeCodeSDKProvider` applies suite/provider `EnvOverrides` afterward. A valid top-level `env.ANTHROPIC_BASE_URL` therefore overrides this recipe and sends the explicitly supplied Meta credential to that Anthropic gateway. The inline comment contains the reproduction and correction direction. This is distinct from the [earlier custom-header finding](https://github.com/promptfoo/promptfoo/pull/10041#discussion_r3560460932): the added empty header value fixes ambient `process.env`, but not the higher-precedence suite-env layer.

#### P2

- `config.stream: true` is rejected for Responses but silently ignored for Meta Chat because the base chat builder never puts that configuration field in `body`; the shared validator checks only `body.stream`. The inline comment contains a direct body-shaping reproduction.
- All three Meta surfaces emit the wrong `gen_ai.system`: Chat and Responses report `openai`, while Messages reports `anthropic`. This misattributes Meta traffic in provider-level trace filters, dashboards, and alerts. The inline comment covers the shared correction.

### Existing feedback checked

- The [unsupported-subtype thread](https://github.com/promptfoo/promptfoo/pull/10041#discussion_r3557272465) is only partially addressed. The finite denylist rejects its original examples, but current probes still route `meta:agents:foo`, `meta:chatkit:foo`, `meta:codex-sdk:foo`, and `meta:voice:foo` as Responses models. I did not duplicate the thread.
- The current [`logit_bias` passthrough thread](https://github.com/promptfoo/promptfoo/pull/10041#discussion_r3563675416) still applies and was not duplicated.
- The prior findings for passthrough output caps, empty-key fallback, `MODEL_API_KEY` guidance, Anthropic sampling-env isolation, templated reasoning effort, explicit `apiHost`, Responses stop validation, bearer-auth collisions, and ambient custom headers are fixed at this head. The resolved CodeQL regex findings are also fixed.

- Replace the unsupported-subtype denylist with explicit parsing that accepts only the three supported surface prefixes while preserving the documented bare-model form.
- Add an overridable provider-system hook to the Responses and Messages base providers, then assert `gen_ai.system: meta` for all three Meta surfaces.
- Add a Claude Agent recipe regression test with a conflicting top-level `ANTHROPIC_BASE_URL`, plus a Chat test for top-level `config.stream`.

### Validation performed

All Node commands ran from the repository root after `source ~/.nvm/nvm.sh && nvm use`.

- `npm ci` — passed with the environment scrubbed before executing PR code.
- `npx vitest run test/providers/meta.test.ts test/providers/registry.test.ts` — 2 files and 178 tests passed.
- `npx --no-install vitest run test/providers/meta.test.ts --sequence.seed=10041` — 87 tests passed in randomized order.
- `npm run build` — passed, including TypeScript, CJS/ESM/declarations, and the app production build.
- `cd site && SKIP_OG_GENERATION=true npm run build` — passed.
- Biome on the 10 changed code/test files — no errors; 5 existing complexity warnings. Prettier on the changed docs/example/schema/baseline files passed.
- `git diff --check bafcbf23f77c34aead99f43cf06c65cedc384517...HEAD` — passed; the worktree remained clean.
- `PROMPTFOO_CONFIG_DIR=/tmp/pr10041-config npm run local -- eval -c /tmp/pr10041-meta-success.yaml --no-cache -o /tmp/pr10041-meta-success.json` — 3/3 passed across Chat, default Responses, and Messages. Exported JSON showed the expected outputs, scores of 1, zero failures/errors, cached-token usage, and computed costs. The loopback server verified Meta bearer auth and absence of `x-api-key` on every route.
- `PROMPTFOO_CONFIG_DIR=/tmp/pr10041-config-error npm run local -- eval -c /tmp/pr10041-meta-error.yaml --no-cache -o /tmp/pr10041-meta-error.json` — expected exit 100; exported JSON contained 3/3 surfaced HTTP 400 errors and no fabricated costs.
- An isolated Messages streaming/cache probe confirmed encrypted thinking stays hidden, cached-token cost is correct, and the replay hits cache without charging cost. Mixed-case `Authorization` header probes preserved the Meta bearer credential while removing inherited proxy and `x-api-key` headers.
- An isolated Claude Agent query-options probe using fake values captured `ANTHROPIC_BASE_URL=https://anthropic-gateway.example` overriding the recipe's `https://api.meta.ai`, while both child-process credential variables held the Meta key.
- A direct request-body probe with `config.stream: true` produced `config.stream=true` but no Chat `body.stream`; the equivalent Responses call threw the documented unsupported-stream error.

The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (`no such file or directory`), so I performed the equivalent six-lens read-only review and reconciled three independent same-head specialist passes.

### Limitations

No hosted Meta credential was used. I did not exercise hosted streaming, tools/search, multimodal input, a live Codex/Claude agent task, or live service rate limits. Meta's detailed feature pages were login-gated during review.

mldangelo:
Addressed all outstanding review findings at head `94cad34ae9` (also merged current main, resolving the edge-baseline conflict by regenerating it):

**P1 — suite-level `env.ANTHROPIC_BASE_URL` overriding the Claude Code recipe.** The recipe now pins `ANTHROPIC_BASE_URL` and `ANTHROPIC_CUSTOM_HEADERS` at the provider-level `env:` layer, which outranks suite env in the subprocess merge (`process.env` < `config.env` < provider/suite env), and the warning explains why. Added a regression test through `loadApiProvider` proving a conflicting suite-level `ANTHROPIC_BASE_URL: https://anthropic-gateway.example` loses to the pin.

**P2 — `config.stream` silently ignored on Meta Chat.** The shared validator now checks `config.stream` as well as `body.stream`; a streaming Chat config fails fast with the documented error. Covered by a unit test and reproduced end-to-end with the local CLI.

**P2 — `gen_ai.system` misattribution.** Added overridable `getGenAISystem()` hooks to the OpenAI Responses and Anthropic Messages base providers (both previously hardcoded), overrode all three Meta surfaces to return `meta`, and added tests for each. No behavior change for existing providers.

**Open thread — unsupported sub-type fallthrough.** Replaced the denylist fallthrough with allowlist parsing: any `meta:<segment>:<model>` whose segment is not `chat`/`responses`/`messages` now throws (`meta:agents:foo`, `meta:chatkit:foo`, `meta:codex-sdk:foo`, `meta:voice:foo` covered by tests); the documented bare-model form and bare sub-type forms are preserved, and the original denylist keeps its more specific error for single-segment paths like `meta:embedding`.

**Open thread — `logit_bias` passthrough.** Now rejected on both OpenAI-compatible surfaces, from `config` and `passthrough`, with tests.

Validation: all 11 new tests fail on the pre-fix code and pass with it; meta/registry/claude-agent-sdk plus full openai + anthropic suites green (1,800 tests); tsc/lint clean; live smoke against api.meta.ai with a real key — all three surfaces return correct output with computed costs, and the new stream rejection fires end-to-end.

Issue #9977, opened 2026-07-04
context-faithfulness / context-recall: test.vars.context clobbers contextTransform output in the grader prompt (...vars spread last)


AsemHub:
## Describe the bug

When a `context-faithfulness` or `context-recall` assertion composes its grading context at runtime with `contextTransform`, the grader silently scores against the **raw `test.vars.context`** instead of the transformed context. The transformed value is still recorded in `gradingResult.metadata.context`, so the run log shows the intended context while the grader actually saw a different one — the score is wrong *and* the log is misleading.

Root cause: in `src/matchers/rag.ts`, the object passed to `renderLlmRubricPrompt` (and to `callProviderWithContext`) spreads `...(vars || {})` **after** `context: contextString`:

```ts
// matchesContextFaithfulness (src/matchers/rag.ts, as of current main)
promptText = await renderLlmRubricPrompt(nliPrompt, {
  context: contextString,   // resolveContext()/contextTransform output
  statements,
  ...(vars || {}),          // vars = test.vars; a `context` key here overwrites the line above
});
```

Later keys win in an object literal, so `vars.context` overwrites the freshly-serialized `contextString`. `matchesContextRecall` has the identical shape with `groundTruth`.

**Trigger condition:** fires when `test.vars.context` is truthy **and** the effective grading context differs from it (e.g. a `contextTransform` builds a larger/composed context). A `contextTransform` with no `context` var set is unaffected (no `context` key in `vars` to clobber).

## To Reproduce

Deterministic proof — the exact object literal from `rag.ts`, no promptfoo needed:

```js
const contextString = "DIGEST + FETCHED_NOTES (the transformed context)";
const vars = { context: "DIGEST ONLY (raw test.vars.context)" };
const rendered = { context: contextString, statements: [], ...(vars || {}) };
console.log(rendered.context);
// -> "DIGEST ONLY (raw test.vars.context)"   <- transformed context silently lost
```

Full in-situ repro (no API key: SUT = `echo`, grader = a recording custom provider). Three files:

```yaml
# promptfooconfig.yaml
description: contextTransform clobber repro
prompts:
  - "The system uses idempotent retries."
providers:
  - id: echo
defaultTest:
  options:
    provider: file://grader.js
tests:
  - vars:
      query: "How does the system handle retries?"
      context: "RAW_CONTEXT_ALPHA."          # raw test.vars.context
    assert:
      - type: context-faithfulness
        threshold: 0
[...]

PR #10042, opened 2026-07-10
fix(matchers): preserve reserved grader vars in model-graded matchers

Gujiassh:
## Summary

Fixes the remaining instances of the reserved-grader-var clobber from #9977. #9994 fixed `context-faithfulness`/`context-recall`, but the identical spread-last pattern survived in seven more model-graded matchers, so a test var named after a reserved grader key silently replaced the real grading content in the grader prompt:

| Matcher | Clobberable keys |
| --- | --- |
| `llm-rubric` | `output`, `rubric` |
| `factuality` | `input`, `ideal`, `completion` |
| `model-graded-closedqa` | `input`, `criteria`, `completion` |
| `select-best` | `criteria`, `outputs` |
| `search-rubric` | `output`, `rubric` |
| `agent-rubric` | `output`, `rubric` |
| `conversation-relevance` (deepeval) | `messages` |

Worst case is `llm-rubric`: with a test var literally named `output`, the grader graded `vars.output` against `vars.rubric` instead of the model output against the assertion rubric.

## Changes

- Spread user vars ahead of the reserved grader keys at every remaining matcher render and provider-call site (`src/matchers/llmGrading.ts`, `search.ts`, `comparison.ts`, `agent.ts`, `src/external/matchers/deepeval.ts`), matching the ordering #9994 established for the rag matchers.
- Hoist the duplicated vars literal in `search-rubric`/`select-best` into a single `templateVars` local (same idiom as `factuality`/`closed-qa`), which also removes a double `tryParse`.
- Regression tests for all seven matchers in the #9994 sentinel style, asserting both the rendered grader prompt and the provider-call vars.

## Validation

- Each new test verified to fail on the pre-fix code and pass with the fix.
- `npx vitest run` across the 9 changed test files plus the 2 adjacent deepeval suites (174 tests) — green.
- `npm run tsc` and lint/format clean.
- End-to-end with the local CLI (echo provider + recording grader): pre-fix, the `llm-rubric` grader prompt contained `<Output>VARS_OUTPUT_SENTINEL</Output><Rubric>VARS_RUBRIC_SENTINEL</Rubric>`; post-fix it contains the actual model output and the assertion rubric.

## Release Impact

User vars named `output`, `rubric`, `input`, `ideal`, `completion`, `criteria`, `outputs`, or `messages` no longer override matcher-built grading values in model-graded assertions. Evals that were silently misgrading because of those var names will see corrected (possibly different) scores.

Follow-up: a shared helper to enforce this invariant structurally across all matcher sites is deliberately deferred to a dedicated PR so this one stays mechanically consistent with #9994.


mldangelo:
@Gujiassh thanks for the fix and the focused repro tests — you had the right diagnosis and the right change. #9994 landed the identical reorder for the rag matchers a few hours before this could merge, which is why the branch went conflicting; that was a race, not a problem with this PR.

Rather than close this, I've taken the liberty of extending your branch: merged current main and applied the same reserved-keys-win ordering to the seven other model-graded matchers that still had the bug (`llm-rubric`, `factuality`, `model-graded-closedqa`, `select-best`, `search-rubric`, `agent-rubric`, and deepeval's `conversation-relevance`), with regression tests in the same style. Your original commit and tests are preserved. Updated the PR title/body to match the expanded scope — happy to have your review on the new commits.

mldangelo:
@codex review

mldangelo (approved):
Great work here, @Gujiassh — you found and correctly diagnosed a genuinely nasty silent-misgrade bug, and your branch ended up being the vehicle for stamping out the entire bug class across every model-graded matcher. The sentinel regression tests (each verified to fail on the pre-fix code) are exactly the safety net we want before consolidating this invariant behind a shared helper in a follow-up.

CI is green across the full matrix, and Copilot, Codex, and the scanner all came back clean. Approving and merging. Thanks for the contribution — looking forward to more!

PR #10043, opened 2026-07-10
feat(assertions): add named script assertion aliases

### LINKED ISSUE #9652: Make it easier to define named assert extensions
Using the python/JavaScript extensions works but has a number of issues:
1. the value is the script file not the content of the value passed to the assert;
2. passing values to the script is cumbersome as you need to write: `config: { value: ... }`;
3. if you change the name/location of the script you need to update all references to it.

Being able to define named aliases to script asserts, e.g.:
- `{ type: python, label: not-is-repetitive, script: file://asserts/repetitive.py:not_repetitive }`
and use it like this:
- `- type: not-is-repetitive`
- `- { type: levenshtein, value: test, config: { distance: 4 } }`

Additionally:
- `context['value']` contains the content of a `value: ...` parameter in the test declaration;
- `context['config']` contains any custom configuration at the test site.

I've tried using extension hooks in Python, but that is preventing the Python prompt from being run and I haven't been able to figure out how to make it work reliably.

**Additional context**
n/a

morgan-coded:
Configs can now define `assertionAliases` — labeled JavaScript, Python, or Ruby assertion scripts that tests invoke by label, with the call-site `value` and `config` exposed to the script context (#9652). Labels are validated against duplicates and collisions with built-in assertion types, aliases expand before dispatch (including inside assert-sets and scenarios), and existing assertion behavior is unchanged. I added unit and config-resolution coverage for the JavaScript, Python, and Ruby paths, regenerated the config schema, and ran the assertions and types suites (1,320 and 449 passing). Fixes #9652


morgan-coded:
Raising a design question on this one before it goes any further.

`assertionTemplates` + `$ref` already covers most of #9652 (docs:
site/docs/configuration/expected-outputs/index.md:436-467). With
json-schema-ref-parser 15.x, sibling keys survive dereference, so `config`,
`metric`, `weight` and friends are already per-call-site, and `config` already
reaches the script as `context.config` (src/assertions/index.ts:451). That
covers #9652's third point completely and its second one functionally.

The one thing `$ref` can't express is #9652's first point: on a script
assertion `value` holds the script path, so a call-site `value:` clobbers it —
`config: { value: ... }` is the only route today. This PR moves the path to a
`script:` field and frees `value:` for the call site.

That's the entire delta, and it's ergonomics rather than a missing capability,
which may well not justify a second named-reuse mechanism — especially with
#8780 in flight to make `assertionTemplates` first-class in the schema.

So, honestly: is the call-site `value` worth it?

- If not, I'll close this and open a docs PR for `assertionTemplates` + `config`
  against #9652 instead.
- If it is, I'd rather rebuild it much smaller — a `script:` field on the
  existing `javascript`/`python`/`ruby` types, no new top-level key, no separate
  resolution pass — than land this shape. I'd also drop the `inverse` special
  case, which sits awkwardly next to #9838.

Either way, the `parseFileUrl` change in src/util/functions/loadFunction.ts
(`.rb` function suffixes, backward colon scan) is unrelated to the feature and
I'm happy to split it into its own PR.

morgan-coded:
Rebuilt this to the smaller shape from the design comment above — a `script:` field on the existing `javascript`/`python`/`ruby` types, no new top-level key, no separate resolution pass. Closing in favor of #10409.

PR #10047, opened 2026-07-11
docs(telemetry): document the opt-out beacon and the build-time Web UI control

### LINKED ISSUE #9968: Telemetry beacon to `r.promptfoo.app` still fires when `PROMPTFOO_DISABLE_TELEMETRY` is set
## Summary

With `PROMPTFOO_DISABLE_TELEMETRY=1` (or `true`) set, promptfoo still sends one
outbound `POST` to the hardcoded endpoint `https://r.promptfoo.app/`. The disable
flag correctly suppresses the PostHog client, but the raw `fetch` in `sendEvent()`
is not guarded by the disabled check, so disabling telemetry ironically triggers a
"telemetry disabled" beacon.

The endpoint is a hardcoded constant with no environment override, so there is no
supported way to fully prevent the call from within configuration/env.

## Version

- promptfoo `0.121.17`
- Node 22
- Reproduced via `npx promptfoo eval`

## Expected behavior

When `PROMPTFOO_DISABLE_TELEMETRY` is set, promptfoo makes **no** telemetry-related
network call of any kind.

## Actual behavior

A single `POST https://r.promptfoo.app/` is attempted on the first recorded event,
carrying a `feature_used` / `"telemetry disabled"` payload. In a locked-down /
air-gapped environment this surfaces as a blocked or failed outbound connection to
`r.promptfoo.app`.

## Steps to reproduce

1. In an environment where outbound connections are observable/blocked (e.g. a
   default-deny network sandbox, or a proxy log).
2. Set `export PROMPTFOO_DISABLE_TELEMETRY=1`.
3. Run any eval, e.g. `npx promptfoo eval -c promptfooconfig.yaml`.
4. Observe an outbound `POST` attempt to `https://r.promptfoo.app/` despite
   telemetry being disabled.

Minimal confirmation that the send path is unconditional (no eval needed), if you
want to isolate it: any code path that reaches `Telemetry.record()` while disabled
triggers `recordTelemetryDisabled()` → `sendEvent()` → the `r.promptfoo.app`
`fetch`.

## Impact

- **Privacy/trust:** the documented opt-out (`PROMPTFOO_DISABLE_TELEMETRY`) does not
  fully stop telemetry egress; disabling it actively causes a beacon.
- **Locked-down / air-gapped use:** there is no env/config way to prevent the call,
  so users in restricted environments must block it at the network layer or patch
  `fetch`. (Related but distinct from #5808, which covers prompt/output data going
  to `api.promptfoo.app` via remote grading.)

extrasmall0:
## Summary

Docs-only. Describes how `PROMPTFOO_DISABLE_TELEMETRY` actually behaves today, including a build-time gotcha in the Web UI that is currently undocumented.

**The single opt-out acknowledgment is intentional and stays.** Disabling telemetry suppresses every product-analytics event except one acknowledgment, sent at most once per process (`recordTelemetryDisabled()` is guarded by `telemetryDisabledRecorded`), so that opt-out usage can be measured. On a default-deny network this surfaces as one blocked `POST` to `r.promptfoo.app` per run. `SECURITY.md` and the FAQ now say so plainly, including that the acknowledgment may carry the local user identifier and email.

**The Web UI control is separate and build-time** — this is the part that is easy to get wrong:

- `PostHogProvider.tsx` compares `import.meta.env.VITE_PROMPTFOO_DISABLE_TELEMETRY` against the exact string `'true'`, so **`PROMPTFOO_DISABLE_TELEMETRY=1` does not disable browser telemetry.**
- `vite.config.ts` bakes the value in via `define` at build time, so setting the variable at runtime does not change already-built assets — including the assets bundled in the published `promptfoo` package.

The FAQ now has two targeted entries: *"Why do I still see one request after setting `PROMPTFOO_DISABLE_TELEMETRY`?"* and *"How do I disable telemetry in the Web UI?"* (with the `PROMPTFOO_DISABLE_TELEMETRY=true npm run build` invocation).

## Scope change from the original PR

This PR originally also added an early return in `Telemetry.sendEvent()` to suppress the acknowledgment, which would have made the CLI send no telemetry request at all. That behavior change has been **reverted** — `src/telemetry.ts` and `test/telemetry.test.ts` are now byte-identical to `main`, and the existing test asserting the acknowledgment is still sent is untouched. Keeping the beacon is a deliberate product decision.

Consequently this no longer closes #9968, which asks for zero telemetry-related network calls. Dropping the `Fixes` keyword so merging does not auto-close it; the issue should be resolved (or declined) on its own terms. What this PR does address is the documentation gap the issue exposed — the behavior was previously undocumented and surprising.

## Verification

- `git diff origin/main -- src/ test/` is empty: no code or test changes.
- `npx vitest run test/telemetry.test.ts` — 25 passed.
- `cd site && SKIP_OG_GENERATION=true npm run build` passes; all links resolve.
- Claims checked against source: `PostHogProvider.tsx:38,77` (`=== 'true'`), `vite.config.ts:173-174` (build-time `define`), `telemetry.ts:110-115` (once-per-process guard).


mldangelo-oai (approved):
Release audit on f225865f562b94ada9f16e671401028f3317922e: ran the focused telemetry suite (26 tests) and reviewed the disabled path. Disabled telemetry now returns before network egress, with the enabled path preserved; no release-blocking findings.

mldangelo-oai:
@codex review

Please review current head baf10b826241189b7276dec4107358f4758dc613.

PR #10049, opened 2026-07-11
docs(site): fix HTTP examples and blog grid

mldangelo-oai:
## Summary
- fix malformed HTTP session parser examples
- make blog pagination state explicit and use stable post keys
- cover paginated headings without a page separator

## AI findings
- addressed five actionable findings
- kept four stale model-name suggestions unchanged because gpt-5, gpt-5-mini, and gpt-5.5 are documented supported identifiers

## Validation
- node_modules/.bin/tsc --noEmit
- attempted focused BlogPostGrid Vitest run; fresh worktree dependency links prevented authoritative local discovery, so CI is the source of truth for site test/build coverage


site/src/components/Blog/BlogPostGrid.test.tsx
```diff
@@ -16,8 +16,14 @@ describe('BlogPostGrid', () => {
   });
 
   it('separates the older posts heading from the page number', () => {
-    render(<BlogPostGrid posts={[]} title="Older Posts • Page 2" />);
+    render(<BlogPostGrid posts={[]} title="Older Posts • Page 2" isPaginated />);
```
jameshiester-oai: **P2 — Make this test distinguish the explicit pagination prop from the old title heuristic.** Every paginated case still uses a title containing `Older Posts`, so all three tests would pass if the component reverted to `title.includes('Older Posts')` and ignored `isPaginated`; the new contract is therefore not regression-protected. Use a case such as `title="Archive • Page 2" isPaginated` and assert the exact accessible name `Archive 2` plus `data-is-paginated="true"` (or assert an `Older Posts` title remains unparsed when the prop is absent). The separator-free case should also use an exact heading-name assertion rather than substring matching.

jameshiester-oai (commented):
## Review summary

**Reviewed base:** `ab52ee079c4558dad731a3165b0741cd64a34344`  
**Reviewed head:** `1537e34ede52fbb7ff589169e1789d3b713fd393`  
**Scope:** Fresh full review of the complete base...head diff. The head advanced from the initial queue snapshot before review began, so the review uses the current fetched merge head.

### Verdict and CI

The explicit pagination state, stable permalink keys, caller update, and repaired body-parser example are correct. I found one new P2 behavioral regression-test gap. Exact-head CI is green: 39/39 checks passed, including Site tests and Build Docs.

### Verified findings

#### P2

- The three Blog grid tests all pass with the old `title.includes('Older Posts')` inference, so they do not protect the new explicit `isPaginated` behavior. The no-separator assertion is also substring-only. The inline comment gives a counterexample that fails the old implementation.

### Existing feedback checked

The existing [mixed-case session-header finding](https://github.com/promptfoo/promptfoo/pull/10049#discussion_r3564770133) remains valid and was not duplicated. Runtime response headers are built from `Headers.entries()`, which normalizes names to lowercase; a direct probe produced only `x-session-id`, and the documented `data.headers["x-session-Id"]` lookup returned `undefined`.

No material implementation simplification is needed. The only production caller passes `isPaginated={!isFirstPage}`, Docusaurus pagination is 1-based, and `metadata.permalink` is the appropriate stable post identity. The regression test should use a paginated title that does not contain “Older Posts” and assert the exact accessible heading name and pagination attribute.

### Validation performed

All Node commands ran after `source ~/.nvm/nvm.sh && nvm use`.

- `npm run test --prefix site -- src/components/Blog/BlogPostGrid.test.tsx` — 1 file and 3 tests passed.
- `npm run typecheck --prefix site` — passed.
- `cd site && SKIP_OG_GENERATION=true npm run build` — passed; client/server bundles and static docs output completed.
- `npx biome check site/src/components/Blog/BlogPostGrid.test.tsx site/src/components/Blog/BlogPostGrid.tsx site/src/theme/BlogListPage/index.tsx` — passed.
- `npx prettier --check site/docs/providers/http.md site/src/components/Blog/BlogPostGrid.test.tsx site/src/components/Blog/BlogPostGrid.tsx site/src/theme/BlogListPage/index.tsx` — passed.
- `git diff --check ab52ee079c4558dad731a3165b0741cd64a34344...HEAD` — passed; the worktree remained clean.
- An isolated `createSessionParser` probe confirmed the repaired body expression returns the expected session ID. The same probe showed `Headers` normalizing `X-Session-Id` to `x-session-id`, the current mixed-case docs expression returning `undefined`, and the lowercase expression succeeding.
- Focused coverage confirmed the tests never render a post and leave the changed post-map/key path uncovered; more importantly, their current titles do not distinguish explicit state from the old heuristic.

The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (`no such file or directory`), so I performed the equivalent six-lens read-only review and reconciled three independent same-head specialist passes.

### Limitations

I did not start a site dev server or run interactive browser QA because the site instructions require user approval before starting one. Focused component tests, typechecking, the production site build, and exact-head CI cover the changed rendering path without a server.

PR #10056, opened 2026-07-12
feat(redteam): add Unicode normalization strategy

### LINKED ISSUE #10039: [Feature] Add Unicode normalization strategy for redteam prompt injection detection
## Summary
Add a new redteam strategy that normalizes Unicode characters before applying prompt injection templates to bypass character-level detection.

## Problem
Current prompt injection strategies don't handle Unicode normalization. Attackers can use Unicode lookalikes (homoglyphs) to bypass text-based detection. A normalization strategy would allow redteam testing of models' Unicode handling.

## Proposed Implementation
- New file: src/redteam/strategies/unicodeNormalize/index.ts
- Configurable normalization forms: NFC, NFKC, NFD, NFKD
- Optional homoglyph substitution table
- Test coverage in 	est/redteam/strategies/unicodeNormalize.test.ts

## Use Case
Redteam operators can test if their target LLM is vulnerable to Unicode-based bypasses.

## References
- Unicode Normalization Forms: UAX #15

3nesdeniz:
## Summary

- add a `unicode-normalization` red-team strategy with NFC, NFD, NFKC, and NFKD support, defaulting to NFKD
- expose the normalization form in the Web UI and preserve explicit form configuration when the strategy is composed in Layer
- skip byte-identical standalone probes and non-string injected values instead of duplicating a baseline or coercing data
- preserve unchanged inputs inside Layer so later transforms can still run; omit the final generated probe only when a Unicode-only pipeline remains unchanged
- keep per-turn agentic Layer conversations moving when Unicode normalization is a no-op
- validate Unicode forms at the top level and recursively inside nested Layer steps, returning HTTP 400 before provider work for invalid preview requests
- return an explicit empty preview batch for Unicode-only no-ops while preserving existing fallback behavior for other strategies
- keep the strategy explicit opt-in, including coding-agent canary exclusions, UI metadata, risk scoring, generated schema, and documentation

## Design notes

Unicode normalization and homoglyph substitution remain separate strategies. Promptfoo already maintains the `homoglyph` strategy and substitution table, so users can compose the two with Layer when they need both behaviors.

The implementation uses the runtime's standards-compliant `String.prototype.normalize()` behavior and validates the requested form before generation. Changed probes retain assertions and existing metadata while recording the original text and applied normalization form. Assertions without a metric name remain unchanged.

No-op behavior is deliberate: standalone unchanged probes are filtered to avoid duplicate provider calls and false strategy attribution. Within a Layer pipeline, unchanged values continue to later transforms; a fully unchanged Unicode-only pipeline is reported as skipped. At runtime, unchanged per-turn prompts pass through so an agentic conversation is not interrupted.

The Web UI and backend share the same normalization-form constants and default. Existing configurations remain backward compatible because an omitted form still resolves to NFKD.

## Validation

- 339 focused backend strategy, route, schema, Layer, and reporting tests passed
- 92 focused Web UI strategy-configuration and preview tests passed
- full red-team regression suite passed: 141 files, 2,571 tests
- root and Web UI TypeScript checks passed
- Biome, `git diff --check`, architecture boundaries, and Knip passed
- production package/Web UI build passed
- Docusaurus production build passed

Fixes #10039


mldangelo-oai (approved):
Release audit on c895fdda98fbf8410855776a95f0d84ddb942acb: reviewed the Unicode-normalization strategy, schema/UI wiring, and affected tests; focused backend/UI suites pass (125 tests) and CI is green. No release-blocking findings.

mldangelo (commented):
Thanks for this — a clean, well-integrated strategy addition.

### Heads-up: I pushed three commits to your branch

So you aren't surprised by commits you didn't write, I pushed to `feat/unicode-normalization`:

- `d2ba7f4` — a `logger.warn` in `addUnicodeNormalization` when *every* case in a batch comes back `normalizationChanged: false`, plus 4 tests
- `0b72726` — test hygiene + `architecture/edge-baseline.json` regen for the new `logger` import
- `377c0a9` — empty commit to retrigger CI

Purely additive: same returned array, same metadata, same metric names. Your `records when normalization leaves an ASCII-only prompt unchanged` test still passes, and `npx vitest run test/redteam` is 140 files / 2557 tests green. Revert it freely if you'd rather solve the underlying issue differently.

### What works well

- **The tests are load-bearing, not decorative.** Replacing `return text.normalize(form)` with `return text` fails 7 tests across `unicodeNormalization.test.ts` and `index.test.ts`.
- `resolveUnicodeNormalizationForm` **throwing** on an invalid form instead of silently defaulting is the right call.
- Integration is genuinely complete — constants, registry, riskScoring metadata, canary exclusions, UI dialog, sidebar — and the docs page honestly calls out the homoglyph ordering interaction rather than glossing over it.

### 1. Is the ASCII no-op the behavior you want? (P2 — design question)

The docs say "ASCII-only prompts usually remain unchanged" and a test pins it, so I read this as deliberate. I mainly want to check you've seen the downstream cost. Measured through the real CLI (intent plugin, `strategies: [basic, unicode-normalization]`):

```
redteam generate -> Total tests: 6 | Plugin tests: 3
  basic -> 3 probes | unicode-normalization -> 3 probes
  uni[0..2] byte-identical to a basic probe: true (normalizationChanged: false)
eval --no-cache  -> 6 provider calls issued for 3 distinct probes
```

All four forms are exact identity on ASCII (5/5 realistic payloads x 4/4 forms). Two consequences worth your call:

- `getStrategyIdFromTest` (`src/app/src/pages/redteam/report/components/shared.ts:25`) reads `metadata.strategyId` verbatim, so a plain baseline failure is reported as a successful "Unicode Normalization" attack that performed no transformation.
- `normalizationChanged` has zero production consumers — nothing filters, warns, or annotates on it, and `applyStrategies` pushes every returned case with no dedupe.

Would you consider returning `undefined` for unchanged cases in `unicodeNormalization.ts:51`? `applyStrategies` already filters `null`/`undefined`, so it's framework-native. It does contradict your existing test, which is why I'm asking rather than pushing it.

### 2. `src/redteam/constants/frameworks.ts:343` — OWASP preset addition (P2)

This takes `owasp:llm:redteam:implementation` from 11 to 12 strategies (+9.1%) for everyone already using that preset. 9 of its 10 plugins emit English ASCII prose; the tenth, `ascii-smuggling`, relies on tag-block / zero-width / variation-selector characters, all of which I verified survive NFKD unchanged. Deleting the line and running `npx vitest run test/redteam` gives 140 files / 2553 tests green — no test asserts the preset's contents, so this is completely unguarded either way. Worth dropping from this PR (users can still opt in explicitly), or landing it once #1 is settled. If it stays, an assertion on the preset list would be good.

### 3. `site/docs/red-team/plugins/coding-agent.md:158` (P3 — quick fix)

You added `unicode-normalization` to `CANARY_BREAKING_STRATEGY_IDS`, but that line hardcodes the exclusion list as prose and now omits it. A user pairing `coding-agent:*` with this strategy will read that page, expect it to run, and have it silently dropped at `src/redteam/plugins/index.ts:156`.

### Nits (take or leave)

- `src/redteam/strategies/types.ts:3` — the added `export type { TestCase };` exists only so `unicodeNormalization.ts:7` can import from `./types`. Every sibling strategy (`homoglyph.ts:1`, `rot13.ts:1`, `base64.ts:1`, ~20 more) imports it from `../../types/index`; matching them reverts `types.ts` to untouched.
- `unicodeNormalization.ts:9-14` — the re-exports of `DEFAULT_UNICODE_NORMALIZATION_FORM` / `UNICODE_NORMALIZATION_FORMS` / `UnicodeNormalizationForm` have no consumers outside the new test file (the UI imports them from `@promptfoo/redteam/constants/strategies` directly). AGENTS.md asks for direct imports over re-exports, and Knip may flag these.
- The `typeof x === 'string' && UNICODE_NORMALIZATION_FORMS.includes(...)` predicate is duplicated in `resolveUnicodeNormalizationForm` and `StrategyConfigDialog.tsx:56-59`, costing three `as UnicodeNormalizationForm` casts. An `isUnicodeNormalizationForm` type guard beside the constants would narrow and delete all three — same pattern as `isMultiTurnStrategy` (`constants/strategies.ts:32`) and `isFanoutStrategy` (`:194`).

Nothing here blocks on my end once #1 and #2 have your answer.

3nesdeniz:
Thanks for the detailed review, @mldangelo, and for the commits you added to the branch.

I compared every item against the current implementation and synced the branch with the latest `main`:

- `d2ba7f4` and `0b72726` are your warning, test-hygiene, and architecture updates.
- `e19b93e` filters byte-identical normalization results, removes `unicode-normalization` from the default OWASP implementation preset with explicit opt-in coverage, and updates the coding-agent exclusion documentation and tests.
- `5b0f0c3` preserves the intended `layer` composition behavior and adds regression coverage for the no-op path.
- `e84149a` is your follow-up documentation clarification for ASCII no-op behavior.
- The unused strategy re-exports were removed, and the shared `isUnicodeNormalizationForm` guard now handles validation and narrowing.

I intentionally kept the local `TestCase` type re-export: importing it directly in this additional strategy module would widen the `redteam -> legacy-contracts` architecture edge, while the current boundary passes the repository architecture ratchet.

Fresh validation on current head `0a35e9d81`:
- focused backend: 294/294
- focused UI: 92/92
- integration: 64/64
- build, TypeScript, generated-schema parity, architecture, changed-file lint, and Prettier: passed
- branch: 0 commits behind current `main` at push time

All actionable review items appear addressed. When convenient, could you take another look at the updated head and let us know if anything remains?

3nesdeniz:
Hi @mldangelo — just following up when you have a chance. We addressed the review items on the current head 0a35e9d81, and all checks are green. Is there anything else you would like changed before this is ready to merge? No rush, and thanks for taking another look.

3nesdeniz:
Hi @mldangelo — following up once more since the approval a couple of weeks back. Happy to rebase onto current main or adjust anything else if that would help get this over the line.

PR #10068, opened 2026-07-13
fix(redteam): pass target output to structured graders

ianw-oai:
Pass the evaluated response into structured redteam rubrics so domain graders no longer see `undefined` or a colliding test value.

### DIFF
--- a/src/redteam/plugins/base.ts
+++ b/src/redteam/plugins/base.ts
@@ -477,6 +477,7 @@ export abstract class RedteamGraderBase {
       // This is done after gradingContext so renderedValue properties take precedence
       ...(typeof renderedValue === 'object' && renderedValue !== null ? renderedValue : {}),
       value: renderedValue,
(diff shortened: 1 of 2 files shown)


alandelong-oai (approved):

faizan-oai (approved):

mldangelo (approved):
Verified in depth: 40 structured graders render `${vars.output}` in their rubrics, and before this fix every one of them produced a literal `## Model Output\nundefined` section — the grading model only saw the real response through the matcher's own template, and a colliding `renderedValue.output` could plant untrusted text in that section. The precedence change is surgical (only `output` and the trace keys become canonical; `categoryGuidance` and friends keep their renderedValue override), and the `it.each` regression test with a real grader covers both the undefined and injection cases. Ran the full redteam plugin suite (875 tests) plus tsc/lint against a tree merged with today's main — all green.

PR #10071, opened 2026-07-13
fix(redteam): add missing lowercase f to homoglyph map

orangeCatDeveloper:
## Problem

The homoglyph strategy's character map is missing lowercase `f` (the map jumps `e` → `g`; uppercase `F` is present). Since `toHomoglyphs` falls back to the original character on a map miss, any payload containing `f` leaks the literal ASCII `f`, weakening the obfuscation. The existing test only checks the whole alphabet string changes, so it missed the gap.

## Fix

- Add `f: 'ƒ'` (U+0192) to the map.
- Add a regression test asserting every alphanumeric character maps to a non-ASCII homoglyph.

## Testing

`npx vitest run test/redteam/strategies/homoglyph.test.ts` — 18 passed; verified the new test fails without the fix.


mldangelo:
Thanks again!

PR #10073, opened 2026-07-13
fix(redteam): encode hex strategy payloads as UTF-8 bytes

orangeCatDeveloper:
## Problem

The hex strategy hex-encoded each character's UTF-16 code unit (`char.charCodeAt(0)`), so any non-ASCII payload was mis-encoded:

- Characters above U+00FF produced variable-width tokens: `€` → `20AC` instead of the UTF-8 bytes `E2 82 AC`.
- Characters outside the BMP were split into surrogate halves: `😊` → `D83D DE0A` instead of `F0 9F 98 8A`.

A model asked to decode the hex as bytes cannot reconstruct the original text, so the attack is silently invalid for multilingual/emoji payloads. ASCII payloads were unaffected.

## Fix

Encode the UTF-8 bytes via `Buffer.from(originalText, 'utf8')`, matching the sibling `base64` strategy. Per-byte `padStart(2, '0')` is now always correct since bytes are 0–255.

## Testing

`npx vitest run test/redteam/strategies/hex.test.ts` — 9 passed. Added multibyte (`€`, `é`), emoji, and round-trip regression tests; verified they fail without the fix. Existing ASCII tests unchanged.

### DIFF
--- a/src/redteam/strategies/hex.ts
+++ b/src/redteam/strategies/hex.ts
@@ -11,9 +11,8 @@ export function addHexEncoding(testCases: TestCase[], injectVar: string): TestCa
       })),
       vars: {
         ...testCase.vars,
-        [injectVar]: originalText
-          .split('')
-          .map((char) => char.charCodeAt(0).toString(16).toUpperCase().padStart(2, '0'))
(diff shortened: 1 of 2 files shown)


mldangelo:
Thank you!

PR #10076, opened 2026-07-13
fix(assertions): return pass:false for tool calls missing a function object

he-yufeng:
## Summary

`is-valid-openai-tools-call` is meant to reject anything that doesn't look like a valid tools response with `pass: false` and a clear reason. But the guard reads `toolsOutput[0].function.name` / `.arguments` directly, so a `tool_calls` entry that has no `function` object throws `TypeError: Cannot read properties of undefined (reading 'name')` instead of failing the assertion — an uncaught crash for the whole test rather than a clean failure.

## Root cause

```ts
if (
  !Array.isArray(toolsOutput) ||
  toolsOutput.length === 0 ||
  typeof toolsOutput[0].function.name !== 'string' ||   // throws when .function is missing/null
  typeof toolsOutput[0].function.arguments !== 'string'
) {
```

The `length === 0` check guarantees `toolsOutput[0]` exists, but nothing guarantees `toolsOutput[0].function` is present, so the property access throws before the `typeof` can evaluate.

## Why it's reachable

This is a real shape, not a contrived one:

- OpenAI **custom tools** stream as `{ type: 'custom', custom: { name, input } }` — there is no `.function` field.
- A partial/streamed response can carry `{ type: 'function', function: null }`.
- Any provider mimicking the OpenAI format can return a malformed `tool_calls` array (e.g. `['...']` or `[{}]`).

All of these are exactly the cases the guard was written to catch, so they should end up as `pass: false`, not a crash.

## Fix

Guard the access with optional chaining so a missing/null `function` (or a non-object entry) falls through to the existing "not a valid-looking tools response" result:

```ts
typeof toolsOutput[0]?.function?.name !== 'string' ||
typeof toolsOutput[0]?.function?.arguments !== 'string'
```

Behavior for well-formed tool calls is unchanged; only the previously-crashing inputs now return the intended `pass: false`.

## Tests

Added a regression case to `handleIsValidOpenAiToolsCall` that runs the four malformed shapes above and asserts each returns `pass: false` (with the existing reason) rather than throwing. Before the fix each of these throws; after it they all return the clean failure result.

### DIFF
--- a/src/assertions/openai.ts
+++ b/src/assertions/openai.ts
@@ -50,8 +50,11 @@ export const handleIsValidOpenAiToolsCall = async ({
   if (
     !Array.isArray(toolsOutput) ||
     toolsOutput.length === 0 ||
-    typeof toolsOutput[0]?.function?.name !== 'string' ||
-    typeof toolsOutput[0]?.function?.arguments !== 'string'
+    toolsOutput.some(
+      (toolCall) =>
+        typeof toolCall?.function?.name !== 'string' ||
+        typeof toolCall?.function?.arguments !== 'string',
+    )
   ) {
     return {
       pass: false,
--- a/src/assertions/openai.ts
+++ b/src/assertions/openai.ts
@@ -50,8 +50,8 @@ export const handleIsValidOpenAiToolsCall = async ({
   if (
     !Array.isArray(toolsOutput) ||
     toolsOutput.length === 0 ||
-    typeof toolsOutput[0].function.name !== 'string' ||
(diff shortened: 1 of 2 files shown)


mldangelo (approved):
Thanks @he-yufeng — clean, well-scoped fix with exactly the right regression coverage. Verified the sibling function-call handler and post-guard paths are also safe, and extended the guard to cover malformed entries beyond the first so mixed arrays fail with the same clear reason. Appreciate the contribution!

PR #10078, opened 2026-07-13
fix(providers): avoid app-server config recursion

wholley-oai:
## Summary

- remove compatibility-only live `provider` metadata before rendering Codex app-server prompt config
- preserve variable rendering for supported app-server configuration
- add regression coverage for circular provider objects

## Problem

Promptfoo resolves grading providers into live objects and can merge them into `prompt.config` through test options. The Codex app-server provider then recursively renders that merged config. A live provider can contain methods and circular references, so rendering can invoke provider methods or fail with `RangeError: Maximum call stack size exceeded` before the app-server request starts.

The app-server schema accepts the `provider` key for loader compatibility, but the runtime does not consume it. Removing that key from the fresh merged config before variable rendering avoids the recursion without mutating caller-owned config or changing supported config precedence.

## Validation

- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci`
- `npx vitest run test/providers/openai-codex-app-server.test.ts` (72 passed)
- `npm run tsc`
- `npm run l`
- `npm run f`
- `git diff --check`

An isolated test against unpatched `main` reproduced the stack overflow; the same test passes with this change.

### DIFF
--- a/test/providers/openai-codex-app-server.test.ts
+++ b/test/providers/openai-codex-app-server.test.ts
@@ -3542,8 +3542,9 @@ describe('OpenAICodexAppServerProvider', () => {
     const server = createMockAppServer();
     mocks.spawn.mockReturnValue(server.proc);
 
+    const attachedProviderId = vi.fn(() => 'attached-provider');
     const attachedProvider: Record<string, unknown> = {
-      id: () => 'attached-provider',
+      id: attachedProviderId,
     };
     attachedProvider.self = attachedProvider;
 
@@ -3622,6 +3623,7 @@ describe('OpenAICodexAppServerProvider', () => {
     });
 
     await expect(resultPromise).resolves.toMatchObject({ output: 'Rendered config done' });
+    expect(attachedProviderId).not.toHaveBeenCalled();
   });
 
   it('applies prompt-level server request policy for each turn on a reused connection', async () => {
--- a/test/providers/openai-codex-app-server.test.ts
+++ b/test/providers/openai-codex-app-server.test.ts
@@ -3542,6 +3542,11 @@ describe('OpenAICodexAppServerProvider', () => {
     const server = createMockAppServer();
     mocks.spawn.mockReturnValue(server.proc);
 
+    const attachedProvider: Record<string, unknown> = {
+      id: () => 'attached-provider',
+    };
+    attachedProvider.self = attachedProvider;
+
     const provider = new OpenAICodexAppServerProvider({
       config: {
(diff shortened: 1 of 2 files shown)


mldangelo-oai (commented):
Reviewed base `7671ccb0fe231ab0fc483f2dfa55e48651951178` against head `66addca3e07fb6a9bba42de3e1a6643958c9ce81`.

The production fix is correctly scoped: it removes the compatibility-only live `provider` field from a fresh merged config before recursive rendering, preserves supported config precedence, and does not mutate caller config. I found one non-blocking P2 regression-test gap, noted inline.

Validation:
- focused app-server tests passed twice (72/72, including seed `10078`)
- Biome and `git diff --check` passed
- credential-free CLI smoke test with a dynamically loaded, self-referential `--grader`, local fake app-server, `--no-write`, and `--no-cache` passed 1/1 with score 1, rendered model `gpt-5.4`, zero errors, and zero grader-method calls
- all 42 PR checks are green

Local `tsc` could not be used as an independent signal because the shared install has `@anthropic-ai/claude-agent-sdk` 0.3.197 while the lockfile expects 0.3.200, producing an unrelated existing test type error. Separately, default persisted CLI mode with a genuinely circular `--grader` still fails earlier on SQLite JSON serialization; that is pre-existing and outside this diff.

test/providers/openai-codex-app-server.test.ts
```diff
@@ -3542,6 +3542,11 @@ describe('OpenAICodexAppServerProvider', () => {
     const server = createMockAppServer();
     mocks.spawn.mockReturnValue(server.proc);
 
+    const attachedProvider: Record<string, unknown> = {
+      id: () => 'attached-provider',
```
mldangelo-oai: [P2] Assert that rendering never invokes the attached provider method

This fixture proves the circular-reference overflow is gone, but it never verifies the separate side-effect regression this fix is intended to prevent: `renderVarsInObject` invokes function-valued fields while traversing. A future cycle-aware traversal could still call `id()` and this test would pass. Make `id` a `vi.fn(() => 'attached-provider')` and assert it was not called after `callApi`, while retaining the self-cycle check.

mldangelo-oai (approved):
Re-reviewed `e04ef5638820669bd7cded2ae60b92e1df8c73ad`. The prior P2 is fully addressed; focused tests, adversarial checks, and the credential-free end-to-end evaluator smoke test pass, and all 41 checks are green. Approved.

PR #10086, opened 2026-07-14
fix(validators): default session source to client when provider.config is undefined

chuenchen309:
## Summary

`determineEffectiveSessionSource` (`src/validators/util.ts`) throws a `TypeError` instead of defaulting to `'client'` when a provider doesn't set a `config` object:

```ts
return (
  sessionConfig?.sessionSource ||
  provider.config?.sessionSource ||
  (provider.config.sessionParser ? 'server' : 'client')  // <-- no `?.` here
);
```

`ApiProvider.config` is typed `config?: any` (`src/types/providers.ts`), so any provider that doesn't set `config` (common for simple custom JS providers) hits `provider.config.sessionParser` on `undefined` and throws, instead of falling through to the `'client'` default. Every other line in this same function (`sessionConfig?.sessionSource`, `provider.config?.sessionSource`) already uses optional chaining — this was the one line that didn't.

This is reachable from `testProviderSession()`'s session-probing path (`src/node/testProvider.ts`), where "Test Target" session detection degrades into a generic "Failed to test session" error for providers without an explicit `config`.

## Fix

Add the missing `?.`: `provider.config?.sessionParser ? 'server' : 'client'`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test plan

- Added a regression test (`test/validators/util.test.ts`) constructing a provider with no `config` key at all and asserting `determineEffectiveSessionSource` returns `'client'` instead of throwing.
- Verified red/green: reverted the source fix locally, confirmed the new test fails with the exact `TypeError: Cannot read properties of undefined (reading 'sessionParser')`; restored the fix, confirmed all 22 tests in the file pass.
- Ran the full `test/validators/` and `test/node/testProvider.test.ts` suites (140 tests) — all pass.
- `npm run l` / lint via pre-commit hook — clean.

## Disclosure

Found and fixed with AI-assisted code review (Claude Code) as part of a broader pass over `src/validators/`; I wrote/ran the repro and tests myself and reviewed the diff before opening this PR.


mldangelo:
Verified end-to-end: on main, `determineEffectiveSessionSource({ provider })` with a config-less provider throws `TypeError: Cannot read properties of undefined (reading 'sessionParser')`; on this branch it returns `'client'`. Ran the full validator suite (22/22) and a multi-agent review over the change — no findings. LGTM as-is.

mldangelo-oai (approved):
Release audit on b0547d9b1bd2e9c02b149c0e612857f9c8603a21: ran the focused validator suite (22 tests) and reviewed the undefined provider.config path. Session source now safely defaults to client without changing configured behavior; no release-blocking findings.

mldangelo-oai:
@codex review

Please review the current head eb4189119e9c7d2fb85764b37b2b7370b92b091e.

mldangelo-oai:
Thank you!

PR #10087, opened 2026-07-14
fix(eval): combineFilterConditions doesn't recognize lowercase 'or'

chuenchen309:
## Summary

`combineFilterConditions` (`src/models/eval.ts`) combines results-table filter conditions using each filter's `logicOperator`:

```ts
return logicOperator === 'OR' ? sql`${acc} OR ${cond}` : sql`${acc} AND ${cond}`;
```

This is an exact, case-sensitive match against the literal string `'OR'`. But the results-table filter UI always sends lowercase values — `ResultsFilter.logicOperator` is typed `'and' | 'or'` in `src/app/src/pages/eval/components/store.ts`, and every call site in `src/app/src/pages/eval/` that sets it (`FilterChips.tsx`, `hooks.ts`, `FiltersForm.tsx`, etc.) uses the lowercase literal — I couldn't find a single place in the frontend that ever produces uppercase `'OR'`.

Since the real payload never matches `'OR'` exactly, **every multi-filter OR combination in the eval results table is silently combined with AND instead**: the table (and the "filtered count" badge) shows only rows matching *every* selected filter, not rows matching *any* of them, whenever a user picks "OR" to combine two or more metric/metadata/plugin/strategy/severity/policy filters.

The existing `combineFilterConditions` unit tests all pass uppercase `'OR'`/`'AND'` literals directly, which incidentally matches the exact-match check and never exercises the real (lowercase) shape the UI actually sends — so this went uncaught.

## Fix

Compare case-insensitively: `logicOperator?.toUpperCase() === 'OR'`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test plan

- Added a regression test asserting that `logicOperator: 'or'` (lowercase, matching the real UI payload) produces SQL containing `OR` and not `AND` — reverted the fix locally first to confirm this new test fails against the current code (rendered `AND` instead of `OR`), then restored the fix and confirmed it passes.
- Ran the full `test/models/eval.test.ts` suite (125 tests) — all pass.
- Lint via pre-commit hook — clean (one pre-existing cognitive-complexity warning on an unrelated, unmodified function in the same file, confirmed present on `upstream/main` before this change too).

## Disclosure

Found and fixed with AI-assisted code review (Claude Code) as part of a broader pass over `src/models/`; I traced the data path end-to-end myself (frontend type → serialization → server route → this function), wrote/ran the repro and tests, and reviewed the diff before opening this PR.


mldangelo (approved):
Excellent first contribution, @chuenchen309 — thank you! This is a genuinely user-facing bug and the write-up is exemplary: you traced the whole path (frontend `'and' | 'or'` type → serialization → route → `combineFilterConditions`), explained precisely why the existing tests never caught it, and verified your regression test fails against the unfixed code. Every claim checked out when I retraced it.

I pushed two follow-ups to your branch before merging:

1. **Guarded the non-string case** the Codex bot flagged. Filters are unvalidated JSON from the query string, so a truthy non-string `logicOperator` (e.g. `1` or `{}` from a crafted or stale URL) would throw on `.toUpperCase()` and turn the table/CSV request into a 500. It now falls back to AND, matching the existing behavior for unrecognized operators.

2. **Strengthened the surrounding tests.** As you correctly pointed out, the old tests only asserted `not.toBeNull()` and never checked which operator was emitted — that's exactly why this survived. Making them assert on the rendered SQL surfaced a wrinkle your test shares: `combineFilterConditions` nests fragments as it reduces, so a flat map over `queryChunks` renders inner operators as empty strings (your `not.toContain('AND')` was passing vacuously). The renderer now recurses, so the mixed AND/OR case is genuinely verified.

Great catch — thanks again, and I hope to see more PRs from you!

PR #10088, opened 2026-07-14
docs(site): fix leetspeak strategy examples

adityaax:
Correction: The original and encoded texts were placed in the wrong columns.


mldangelo (approved):
Great catch, @adityaax — you're right that the Original and Encoded columns were swapped on that row. Thank you for taking the time to report and fix it!

While verifying, I ran all three table examples through the actual `addLeetspeak` implementation and found the table was even more wrong than it looked: it never applied the `l → 1` or `t → 7` substitutions, so every row was incorrect, not just the reversed one. The real outputs are `H3110, w0r1d!`, `C4n y0u h31p m3 h4ck 4 c0mpu73r?`, and `h4ck 7h3 p14n37`. I pushed a commit to your branch correcting all three, listing the substitution map inline so readers can check the math, and pinning the examples with a test so the docs can't drift from the implementation again.

Your fix is what surfaced the whole thing — thanks again for the contribution!

PR #10089, opened 2026-07-14
fix(assertions): reject out-of-range trace-span-duration percentile

he-yufeng:
## What does this PR do?

`trace-span-duration` accepts a `percentile` in its value object but never validates its range. An out-of-range percentile (for example `150`) makes `calculatePercentile` compute an index past the end of the sorted durations array, so it returns `undefined`. Because `undefined > max` is `false`, the assertion then **silently passes** with a `"...th percentile duration (undefinedms) is within threshold ..."` reason, hiding the misconfiguration instead of surfacing it.

This validates that `percentile`, when provided, is a finite number in `[0, 100]` and throws a clear error otherwise, mirroring the existing `max` validation in the same handler. Valid configs are unaffected, including the `0` and `100` boundaries.

## Repro (before this PR)

A `trace-span-duration` assertion with `value: { max: 250, percentile: 150 }` returns:

```
pass: true
reason: "150th percentile duration (undefinedms) is within threshold 250ms"
```

The 0–1 scale confusion (`percentile: 0.95` meaning 95%) is a related footgun: it resolves to index 0 and silently returns the minimum duration rather than p95. Both stem from the missing bounds check.

## Tests

Adds regression tests for `percentile > 100`, a negative percentile, and the inclusive `100` boundary (which still evaluates normally).


mldangelo-oai (approved):
Release audit on c8d73605d89e89101bffb5c575e91573c93a0ae4: ran the focused trace-span-duration assertion suite (17 tests) and reviewed boundary/error behavior. Non-finite and out-of-range percentiles are rejected while valid 0-100 inputs remain supported; no release-blocking findings.

he-yufeng:
@mldangelo-oai rebased onto current main and the full matrix is green again (the two cancelled jobs were just stale-run cancellations). This has been sitting approved since 07-19, so a gentle nudge: anything else you want from me before it can land?

PR #10093, opened 2026-07-14
refactor(providers): simplify Vertex cache hashing

mldangelo-oai:
## Summary
- remove the unreachable fallback from Vertex request-body cache hashing
- preserve the existing HMAC cache-key structure and serialized request input

## Validation
- npx vitest run test/providers/google/vertex.test.ts --sequence.shuffle=false
- npm run l
- npm run f
- npm run tsc

### DIFF
--- a/src/providers/google/vertex.ts
+++ b/src/providers/google/vertex.ts
@@ -144,7 +144,7 @@ function getVertexBodyCacheKey(prefix: string, body: unknown, apiHost: string): 
   return `${prefix}:${createHmac('sha256', 'promptfoo:vertex:cache-key:v1')
     .update(apiHost)
     .update('\0')
-    .update(serialized ?? String(body))
+    .update(serialized)
     .digest('hex')}`;
 }
 


jameshiester-oai (commented):
## Review scope

Full review of PR base/merge-base `1ede17aaed940e6dff04f71d24e4ecc011809dae` through exact head `7e6cb9b976e3b85aa73fdca918218902fc6f0ac0`. Current `origin/main` was also checked at `fcde2e89a89dc4ca79dcc3012927f50193251759`; the effective PR diff remains the single Vertex cache-key line.

## Verdict

No actionable defects found. The cache namespace, API-host partitioning, serialized request input, and resulting digest are unchanged for every supported caller. The PR is mergeable but behind current `main`; all 44 reported exact-head checks are successful.

## Verified findings

- P0–P3: none.
- The existing outdated/unresolved Copilot thread is technically correct for a hypothetical top-level `undefined`, function, or symbol: `JSON.stringify` returns `undefined` and `Hmac.update` throws. That condition is not reachable through current production paths. All four file-private helper callers construct a top-level request object (Claude, Gemini, Palm2, and Llama), so optional nested `undefined` values are safely omitted while the root still serializes to a string. I did not duplicate the existing thread.
- Optional simplification: narrowing the private helper's `body: unknown` parameter to the object shape its callers actually provide would encode the invariant and reduce future ambiguity. This is not required for correctness on this head.
- No material behavioral test gap remains: the focused Vertex suite reaches all four callers and checks hashed cache-key shape, host/model partitioning, prompt secrecy, and cache get/set consistency.

## Validation

- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci` — passed; exact lock installed.
- `node node_modules/vitest/vitest.mjs run test/providers/google/vertex.test.ts --sequence.shuffle=false --reporter=dot` — passed, 101/101.
- `npm run tsc` — passed.
- `npx --no-install biome check src/providers/google/vertex.ts` — passed with the two pre-existing complexity warnings.
- `npm run build` — passed for TypeScript, core/declarations, and the production Web UI.
- Direct serialization/HMAC probe — an object containing an optional `undefined` property serialized normally and produced a 64-character digest; deliberate top-level `undefined` reproduced `ERR_INVALID_ARG_TYPE`, confirming the existing thread's trigger and its caller-boundary limitation.
- Three independent read-only specialist passes found no additional issue.

## Limitations

The requested PR Review Toolkit command was unavailable on this host (`exit 127`), so its six review lenses were applied manually and reconciled with the specialist passes. No live Vertex credential or external API call was used; provider behavior was exercised through mocked transports as required by the repository's provider-testing guidance. No browser QA was relevant to this backend-only private helper change.

jameshiester-oai (commented):
Reviewed at exact head 1f86f7830462f891909c602f217cf5d6352a9476. No actionable findings or inline comments.

I checked the cache-key behavior, every current caller shape, failure and serialization edges, simplification value, test coverage, and unintended behavior. The removed fallback is unreachable for the concrete request bodies passed by the Claude, Gemini, Palm2, and Llama paths, while the HMAC structure and serialized input remain unchanged.

Validation: focused Vertex Vitest suite passed (101 tests), full TypeScript check passed, diff check passed, and current GitHub checks are green.

PR #10094, opened 2026-07-14
fix(redteam): harden release provider and generation edge cases

mldangelo-oai:
## Summary

Fixes regressions found while auditing `0.121.18..0.121.19` across providers, redteam generation, and the code-scan release path.

- Harden Responses streaming: consume the live body without unsafe pre-buffering, reconstruct incomplete/interleaved output correctly, preserve finalized/redacted text and refusal context, never resurrect filtered/refused drafts, compact sparse and attacker-sized indices, cancel oversized streams, and cap aggregate output growth.
- Fix provider boundaries and accounting: isolate inherited Anthropic/custom headers and caches, preserve Google cache/request counts, normalize token usage safely, and keep generation usage across plugins, strategies, failures, and multi-turn calls.
- Fix Open Interpreter and other provider edge cases: validate/render row and nested approval config (including Nunjucks statements and homes), preserve literal values, harden structured input/runtime handling, and improve TrueFoundry/Claude Agent behavior.
- Repair redteam/API/parser edge cases and the code-scan-action release/provenance workflow (including the Windows scanner path); add focused regression coverage throughout.
- Update the architecture edge baseline for the intentional security/accounting imports (`envOverrides`, rendering, token-usage schemas/utilities, and server-to-redteam tracking), while locking in the reduced `redteam -> node` edge.

## Validation

- `1,993` affected backend tests passed across `45` files (route suite run separately to avoid local ephemeral-port collisions).
- `44` affected UI tests passed across `2` files.
- Root, app, and code-scan-action TypeScript checks passed.
- Code-scan-action build, full production build, Knip, lint, and formatting passed.
- `npm run architecture:check` passed after regenerating the intentional cross-layer edge baseline.
- Local CLI smoke eval passed (`1/1`, `0` failures, `0` errors).
- High-risk prepublish gate passed on `4129ba57737ceca26e49af1626ebc97a6a8edaa9`: three independent exact-head reviews and a fresh verifier, with zero remaining findings.


jameshiester-oai (commented):
## Exact-head review

- **Base:** `1ede17aaed940e6dff04f71d24e4ecc011809dae`
- **Head:** `8ba32261eda0f4a9df0f3c1fd9fda6b52ff9e983`
- **Scope:** full `base...head` review; no prior authenticated completion marker exists for this head.
- **Verdict:** 3 P1 and 4 P2 defects reproduced on the exact head. The `(redteam)` title scope is correct.
- **CI:** all reported checks passed except **Test on Node 22.22 and windows-2025-vs2026 (shard 1/3)**. That job passed 3,908 tests and then failed the changed code-scan action test deterministically; this is a PR regression, not infrastructure noise.

### Verified findings

**P1**

1. Clearing inherited `ANTHROPIC_CUSTOM_HEADERS` can also null the SDK-generated configured `x-api-key`, leaving normal Anthropic requests unauthenticated.
2. Responses reconstruction replaces explicit empty text from a completed terminal response with an earlier streamed draft, violating terminal/redaction precedence.
3. The changed code-scan test hard-codes the POSIX installed executable path, so the exact-head Windows shard fails and the new Windows invocation path is not exercised.

**P2**

1. A delta after `response.output_text.done` mutates finalized text and can override the terminal response.
2. A missing-index delta following indexed output is assigned to the previous item, duplicating/misattributing multi-output content.
3. A single pretty-printed JSON prompt still stops at its first interior blank line and becomes invalid.
4. Bracket-prefixed prose such as `[INST]` or `{username}` is treated as incomplete JSON and absorbs trailing model commentary.

Inline comments contain the trigger, observed result, impact, and smallest correction direction for each item.

### Existing feedback

The unresolved Copilot thread on `src/util/tokenUsage.ts:77-79` remains valid: a throwing `tokenUsage` getter returns before response-aware request counting. I did not duplicate it. The method is deprecated and has no current production call sites, which narrows the present impact, but its contract and new test expectation should still be corrected.

The new Responses reconstruction puts terminal precedence, finalization, index recovery, and fallback assembly in one state machine; Biome reports complexity 34/51/46 for the new helpers. Making finalized content immutable and centralizing a documented terminal-vs-stream precedence rule would remove several of the defects above. Add focused regressions for explicit empty completed output, post-`done` deltas, and mixed indexed/unindexed output.

The parser needs one JSON-continuation classifier shared by single- and multi-prompt paths, with `[INST]` and template-wrapper regressions. The Anthropic scoped-header tests should assert that native SDK authentication survives suppression, not only that proxy headers disappear.

### Validation performed

All Node commands used Node 24.18.0 and npm 11.16.0 after `nvm use`.

- `node node_modules/vitest/vitest.mjs run <25 affected backend test files> --sequence.shuffle=false --reporter=dot` — **25 files / 1,628 tests passed**.
- `npm run test:app -- src/pages/redteam/setup/components/Targets/CustomTargetConfiguration.test.tsx src/pages/redteam/setup/components/Targets/ProviderConfigEditor.test.tsx --run` — **2 files / 44 tests passed**.
- `npm run tsc` — passed.
- `npm --prefix code-scan-action run tsc` — initially exposed missing nested dependencies; after `npm ci` in `code-scan-action`, passed.
- `git diff --name-only -z 1ede17aaed940e6dff04f71d24e4ecc011809dae...HEAD | xargs -0 npx @biomejs/biome check --no-errors-on-unmatched --max-diagnostics=300` — passed with 21 existing/new complexity warnings and no errors.
- `npm run build` — root, package, and app production builds passed.
- `PROMPTFOO_CONFIG_DIR=/tmp/pr10094-promptfoo-config PROMPTFOO_CACHE_PATH=/tmp/pr10094-promptfoo-cache npm run local -- eval -c test/smoke/fixtures/configs/basic.yaml --no-cache -o /tmp/pr10094-e2e.json` — 1/1 passed, score 1, output `Hello World`, no failures/errors.
- Direct `npx tsx -e` probes against the exact-head SDK request builder, Responses stream reader, and prompt parser reproduced all six runtime/parser findings. The Anthropic probe had no `x-api-key`; the stream probes returned `UNSAFE DRAFT`, `SAFE LATE`, and duplicated `FIRSTSECOND`/`SECOND`; parser probes returned `{` for a single JSON prompt and retained trailing commentary for `[INST]`.

### Limitations

`/pr-review-toolkit:review-pr all parallel` is unavailable in this environment (exit 127), so I performed the read-only six-lens fallback and reconciled three independent specialist reports. No live Anthropic/Bedrock call, native Windows run, credentialed redteam generation, or browser session was used; SDK/SSE/parser probes and the authoritative Windows CI log covered those paths without credentials. The PR's key-count guard does not cap aggregate stream bytes, but base behavior was already unbounded, so I treated that as residual hardening risk rather than a new inline defect.

src/providers/anthropic/generic.ts
```diff
@@ -23,22 +24,44 @@ import type { ClaudeCodeOAuthCredential } from './claudeCodeAuth';
...
+  env: EnvOverrides | undefined,
+): Record<string, string | null> {
+  const customHeaders =
+    env?.ANTHROPIC_CUSTOM_HEADERS ?? getEnvOverrides()?.ANTHROPIC_CUSTOM_HEADERS;
+  if (customHeaders === undefined) {
+    return {};
+  }
+
+  return {
+    ...getAnthropicEnvHeaderSuppressions(env),
```
jameshiester-oai, line 72: **[P1] Preserve SDK authentication while clearing scoped headers**

When ambient `ANTHROPIC_CUSTOM_HEADERS` contains `x-api-key` and this provider sets `env.ANTHROPIC_CUSTOM_HEADERS: ''` to clear inherited headers, this suppression map contributes `x-api-key: null` to `defaultHeaders`. The Anthropic SDK applies that after generating auth headers, so a provider with `config.apiKey` builds a request with neither `x-api-key` nor `Authorization` (reproduced with dummy values on this head). Every request in this supported scoped-clear configuration therefore fails authentication. Filter SDK-owned auth names out of suppression or reapply the selected native auth after custom-header isolation, and assert the configured key survives in this test case.

src/providers/responses/stream.ts
```diff
@@ -13,6 +16,206 @@ type ResponsesStreamLogger = {
...
+    if (item.content.length === targetContentIndex) {
+      item.content.push({ type: 'output_text', text: '' });
+    }
+    const content = item.content[targetContentIndex];
+    if (content?.type !== 'output_text') {
+      unmatchedStreamedTexts.push(text);
+      continue;
+    }
+    if (
+      typeof content.text !== 'string' ||
```
jameshiester-oai, line 418: **[P1] Keep explicit empty completed-terminal text authoritative**

This branch treats `text: ''` as missing even when `allowTerminalTextReplacement` is false. A stream containing an indexed `UNSAFE DRAFT` delta followed by `response.completed` whose explicit output text is empty is returned with `UNSAFE DRAFT`; base returned the terminal response unchanged. Providers/proxies can omit `output_text.done` while using an empty completed value for suppression or redaction, so this resurrects content the terminal response removed. Preserve every explicit terminal string, including empty, for completed responses; only fill absent content (or apply a separately defined incomplete-response rule), and add that exact regression.

src/providers/responses/stream.ts
```diff
@@ -53,6 +256,172 @@ export async function readResponsesStream(
...
+        return;
+      }
+      const prependPending =
+        event.output_index === 0 &&
+        (event.content_index ?? 0) === 0 &&
+        !outputTextByContent.has(key);
+      if (!prependPending) {
+        unassignedUnindexedOutputText += pendingUnindexedOutputText;
+      }
+      outputTextByContent.set(
```
jameshiester-oai, line 607: **[P2] Do not mutate a finalized output key**

`processOutputTextEvent` appends to the map even after the same key was finalized by `response.output_text.done`. The exact-head sequence `delta SAFE` → `done SAFE` → late `delta " LATE"` → incomplete terminal `SAFE` returns `SAFE LATE`, because the key remains marked finalized while its value changes. This defeats the finalization/terminal precedence this PR is hardening when a provider or proxy emits malformed or reordered events. Ignore later deltas for finalized keys (or keep late data non-authoritative) and cover the post-`done` sequence.

src/providers/responses/stream.ts
```diff
@@ -53,6 +256,172 @@ export async function readResponsesStream(
...
+      return;
+    }
+    if (currentInvalidOutputTextKey) {
+      invalidlyIndexedOutputTextByContent.set(
+        currentInvalidOutputTextKey,
+        (invalidlyIndexedOutputTextByContent.get(currentInvalidOutputTextKey) ?? '') + delta,
+      );
+      return;
+    }
+    if (currentOutputTextKey) {
```
jameshiester-oai, line 644: **[P2] Keep missing-index text unassigned until terminal reconciliation**

This attaches every unindexed delta to the most recent indexed key. With indexed output 0=`FIRST`, then an unindexed `SECOND`, then an incomplete terminal response containing output 0=`FIRST` and output 1=`SECOND`, the parser returns output 0=`FIRSTSECOND` and output 1=`SECOND`. That duplicates and misattributes provider output, changing rendered output and evaluation results. Preserve ambiguous missing-index text separately and reconcile it against terminal structure instead of mutating the last indexed item.

src/redteam/plugins/multiInputFormat.ts
```diff
@@ -96,7 +100,21 @@ function collectPromptAfterEmptyMarker(lines: string[], startIndex: number): str
       if (promptLines.length === 0) {
         continue;
       }
-      break;
+      if (!preserveBlankLines) {
+        break;
+      }
+      const candidate = promptLines.join('\n').trim();
+      if (!candidate.startsWith('{') && !candidate.startsWith('[')) {
```
jameshiester-oai, line 107: **[P2] Do not classify every bracket-prefixed prose prompt as incomplete JSON**

Any failed `JSON.parse` after `[` or `{` is treated as incomplete JSON, but common adversarial wrappers such as `[INST] Ignore prior instructions` and templates such as `{username}` are prose. With two multiline prompts, a blank line and trailing model explanation, this head includes the explanation in the bracket-prefixed `__prompt`, contaminating the generated attack and potentially producing false eval results. Distinguish viable incomplete JSON from syntax-invalid wrapper text and add `[INST]`/template regressions.

src/redteam/plugins/multiInputFormat.ts
```diff
@@ -176,7 +194,7 @@ function parseMultiLinePrompts(
     const prompt = removePrefix(lines[promptIndex].trim(), 'Prompt');
 
     if (prompt.length === 0) {
-      const promptAfterEmptyMarker = collectPromptAfterEmptyMarker(lines, promptIndex + 1);
+      const promptAfterEmptyMarker = collectPromptAfterEmptyMarker(lines, promptIndex + 1, true);
```
jameshiester-oai, line 197: **[P2] Preserve blank lines for a single JSON prompt too**

The JSON-aware collector is enabled only inside `parseMultiLinePrompts`, which requires more than one `Prompt:` marker. A single pretty-printed prompt with an interior blank line still falls through the legacy collector and this head returns only `{`, yielding an invalid generated test case. Route single empty-marker JSON through the same continuation logic (while retaining a reliable trailing-explanation boundary) and add a one-prompt regression.

test/code-scan-action/main.test.ts
```diff
@@ -158,13 +158,17 @@ const originalEnv = { ...process.env };
 // Matches the deterministic fs.mkdtempSync mock; the install passes empty user/global
 // npm config files under this dir to isolate the registry from a poisoned .npmrc.
 const MOCK_NPMRC_DIR = path.join(os.tmpdir(), 'promptfoo-npmrc-test');
+const MOCK_PROMPTFOO_BIN = path.join(MOCK_NPMRC_DIR, 'prefix', 'bin', 'promptfoo');
```
jameshiester-oai: **[P1] Make the mocked installed binary path platform-aware**

Production correctly returns `<prefix>/promptfoo.cmd` on Windows, but this test helper always waits for `<prefix>/bin/promptfoo`. The exact-head Windows shard therefore fails deterministically with `expected promptfoo and npm install exec calls not found` after 3,908 passing tests; the native macOS run passes. This also means the new security-sensitive absolute invocation path has ineffective Windows coverage. Derive the expected path with the same platform contract (`promptfoo.cmd` on `win32`, `bin/promptfoo` elsewhere) and rerun the Windows suite.

src/providers/responses/stream.ts
```diff
@@ -69,53 +749,349 @@ export async function readResponsesStream(
     }
 
     if (event.response && typeof event.response === 'object') {
```
jameshiester-oai, line 751: [P1] Preserve terminal event authority before replaying tools

This stores only `event.response`, so reconstruction later cannot tell that `response.completed` or `response.failed` was already terminal. On this head, a finalized `dangerous` function call followed by `response.completed` with an output snapshot but no `status` is reinserted and invoked by `ResponsesProcessor`; `response.failed` with `status: failed` but no `error` does the same. A valid `status: completed` event followed by a late delta also returns both `SAFE` and `LATE`. A malformed/truncated stream can therefore execute a stale or cancelled tool or inject post-terminal text. Record the terminal event kind, stop accepting later events, and fail closed on missing or inconsistent terminal fields before restoring executable items.

src/providers/responses/stream.ts
```diff
@@ -13,6 +30,439 @@ type ResponsesStreamLogger = {
...
+    )
+    .filter(Boolean);
+}
+
+function hasTerminalSafetyDecision(response: any): boolean {
+  if (
+    [response?.incomplete_details?.reason, response?.error?.code, response?.error?.message].some(
+      (reason) =>
+        typeof reason === 'string' &&
+        /(?:content[_-]?filter|content[_-]?policy|safety|guardrail)/i.test(reason),
```
jameshiester-oai, line 187: [P2] Do not turn safety-service outages into policy refusals

Matching arbitrary error prose makes an operational failure indistinguishable from a safety decision. An exact-head probe with `{ status: 'failed', error: { code: 'server_error', message: 'Safety system timed out before a decision' }, output: [] }` returned with `error` erased and a synthesized refusal; `ResponsesProcessor` then reported `isRefusal: true`. This hides retryable/provider failures and can create false passes in evals and redteam runs. Match structured policy codes/reasons or actual refusal output, and preserve other errors even when their message mentions the safety subsystem.

src/providers/responses/stream.ts
```diff
@@ -13,6 +30,439 @@ type ResponsesStreamLogger = {
...
+  for (let index = 0; index < text.length; index++) {
+    const code = text.charCodeAt(index);
+    firstHash = Math.imul(firstHash ^ code, 16777619);
+    secondHash = Math.imul(secondHash ^ code, 2246822519);
+  }
+  return `oversized-${text.length}-${(firstHash >>> 0).toString(16)}-${(secondHash >>> 0).toString(16)}`;
+}
+
+function getInvalidOutputTextKey(event: ResponsesStreamEvent): string {
+  return `${compactStreamIndex(event.output_index, 'missing')}:${compactStreamIndex(event.content_index, '0')}`;
```
jameshiester-oai, line 95: [P2] Make malformed-index keys type- and boundary-safe

Stringifying each component and joining with `:` is not injective: `output_index='a:b', content_index='c'` and `output_index='a', content_index='b:c'` both become `a:b:c`; distinct objects also both become `[object Object]`. Two `response.output_text.done` events using the first pair returned only `SECOND`, dropping `FIRST`. These indices are provider-controlled on the malformed path this code is hardening. Use a bounded type-aware tuple encoding or length-prefixed components, then add delimiter/object collision regressions.

src/providers/responses/stream.ts
```diff
@@ -13,6 +30,439 @@ type ResponsesStreamLogger = {
...
+      continue;
+    }
+    const matchingRefusals = finalizedRefusalItems.filter(
+      (refusalItem) =>
+        refusalItem.outputIndex !== undefined && refusalItem.outputIndex === entry.outputIndex,
+    );
+    if (matchingRefusals.length === 0) {
+      continue;
+    }
+    const refusalContent = matchingRefusals.flatMap(({ item }) =>
```
jameshiester-oai, line 307: [P2] Preserve top-level refusal evidence during the merge

`getOutputRefusalItem` accepts a message whose refusal is in `item.refusal`, but this merge copies only nested refusal content and then marks the finalized item merged. With a finalized `{ type: 'message', refusal: 'I cannot help', content: [{ type: 'output_text', text: 'partial unsafe draft' }] }` plus an incomplete terminal snapshot, the result contained only `partial unsafe draft`; the processor returned it as a normal answer with no `isRefusal`. Normalize top-level refusals into a refusal content part before clearing/merging the field so partial refused content cannot surface as an answer.

src/providers/responses/stream.ts
```diff
@@ -53,6 +503,236 @@ export async function readResponsesStream(
...
+    target: Map<string, RetainedStreamOutputItem>,
+    key: string,
+    outputIndex: number | undefined,
+    item: any,
+  ): void => {
+    if (!hasStreamOutputCapacity(target.has(key), target !== finalizedRefusalItems)) {
+      return;
+    }
+
+    const serializedLength = JSON.stringify(item).length;
```
jameshiester-oai, line 552: [P2] Avoid serializing the accumulated refusal for every delta

Each `response.refusal.delta` rebuilds the full refusal, then this `JSON.stringify` scans it again to compute growth. A controlled exact-head run of 120,000 one-character refusal deltas (14.2 MB input, only 120,000 output characters, below the cap) took 4.44 seconds versus 74 ms for comparable output-text deltas. This superlinear path lets an accepted provider stream monopolize the event loop. Track refusal length incrementally and enforce the cap from delta growth; serialize the completed item once.

jameshiester-oai (commented):
## Exact-head incremental review

- **Base:** `1ede17aaed940e6dff04f71d24e4ecc011809dae`
- **Head:** `4129ba57737ceca26e49af1626ebc97a6a8edaa9`
- **Scope:** incremental review from the authenticated completed head `8ba32261eda0f4a9df0f3c1fd9fda6b52ff9e983`, which is an ancestor of this head, plus full `base...head` interaction and regression context.
- **Verdict:** 1 new P1 and 4 new P2 defects reproduced on the exact head. One prior P1 and three prior P2 findings remain reproducible; four earlier findings are fixed. The `(redteam)` title scope is correct.
- **CI:** all reported checks are green, including all three Windows shards.

### New verified findings

**P1**

1. The stream reader discards terminal event kind and later decides whether to replay finalized items only from `response.status`. A finalized function call omitted by a `response.completed` snapshot without `status`, or by a `response.failed` snapshot without `error`, is restored and executed by `ResponsesProcessor`. A delta after a valid `status: completed` snapshot is also appended. Malformed or truncated provider streams can therefore execute stale/cancelled tools or inject post-terminal text.

**P2**

1. Any error message containing `safety` or `guardrail` is treated as a policy decision. A `server_error` saying the safety service timed out loses its error and becomes a synthesized refusal, hiding an operational failure and skewing eval/redteam results.
2. Malformed stream-index components are stringified and joined with `:`, so distinct tuples can share one key. The pairs `('a:b', 'c')` and `('a', 'b:c')` collapsed two finalized outputs and returned only the second; distinct objects collide as `[object Object]` as well.
3. When a finalized message carries a top-level refusal alongside output text, merging with an incomplete terminal message clears the refusal and retains only the draft text. The processed response is returned as a normal answer instead of `isRefusal: true`.
4. Every refusal delta rebuilds and serializes the full accumulated refusal to measure growth. A 120,000-event, 14.2 MB stream with only 120,000 output characters took 4.44 seconds on the exact head, versus 74 ms for equivalent output-text deltas, leaving an accepted provider-controlled stream able to monopolize the event loop.

Inline comments contain the trigger, observed result, impact, and smallest correction direction for each new finding.

### Existing feedback rechecked

The prior review is https://github.com/promptfoo/promptfoo/pull/10094#pullrequestreview-4699702570. I did not duplicate its inline comments.

- **Still current:** P1 native Anthropic authentication is removed when a scoped empty `ANTHROPIC_CUSTOM_HEADERS` suppresses inherited custom headers; the exact-head SDK request had neither `x-api-key` nor `Authorization` despite a configured API key.
- **Still current:** P2 mixed indexed/unindexed Responses deltas misattribute `SECOND`, producing `FIRSTSECOND` and `SECOND`.
- **Still current:** both P2 parser findings: a single pretty JSON prompt truncates at its interior blank line, and bracket-prefixed prose such as `[INST]` absorbs trailing model commentary.
- **Fixed:** explicit empty completed text is no longer replaced by a streamed draft; deltas after `response.output_text.done` are ignored; the Windows scanner path is platform-aware and Windows CI is green; the throwing token-usage getter no longer suppresses request accounting.

The Responses stream state machine still combines transport termination, terminal authority, reconstruction, tool execution eligibility, refusal normalization, indexing, and memory accounting. Preserve the terminal event kind as explicit state, make terminal snapshots immutable, and normalize output items before merging. Those changes would remove several branches that currently infer safety and authority from loose object shape.

Add focused regressions for terminal events with missing/inconsistent fields, post-terminal events, refusal-plus-text items, type-aware malformed-index identity, safety-service outages versus policy blocks, and many small refusal deltas.

### Validation performed

All Node commands used Node 24.18.0 and npm 11.16.0 after `nvm use`.

- `npx vitest run test/cache.test.ts test/providers/anthropic/completion.test.ts test/providers/anthropic/messages.test.ts test/providers/bedrock/anthropicMessages.test.ts test/providers/bedrock/openaiResponses.test.ts test/providers/meta.test.ts test/providers/openai-codex-app-server.test.ts test/providers/openai/responses/metadata.test.ts test/util/tokenUsage.test.ts --sequence.shuffle=false` — **9 files / 602 tests passed**.
- `npx vitest run test/code-scan-action/main.test.ts --sequence.shuffle=false` — **1 file / 53 tests passed**.
- `npm --prefix code-scan-action run tsc` — passed.
- `git diff --name-only -z 8ba32261eda0f4a9df0f3c1fd9fda6b52ff9e983..HEAD | xargs -0 npx @biomejs/biome check --no-errors-on-unmatched --max-diagnostics=300` — passed with 7 complexity warnings and no errors.
- `npm run build` — root package, TypeScript, and app production builds passed.
- `PROMPTFOO_CONFIG_DIR=/private/tmp/pr10094-current-20260715-0933-config PROMPTFOO_CACHE_PATH=/private/tmp/pr10094-current-20260715-0933-cache npm run local -- eval -c test/smoke/fixtures/configs/basic.yaml --no-cache -o /private/tmp/pr10094-current-20260715-0933-e2e.json` — **1/1 passed**, score 1, output `Hello World`, no failures or errors.
- Direct exact-head SDK/SSE/parser probes reproduced every current finding and confirmed the four earlier fixes. The terminal/tool probe executed the omitted `dangerous` callback; the refusal merge returned `partial unsafe draft` without refusal status; the outage probe returned a clean refusal for `server_error`; and malformed indices dropped `FIRST`.
- Controlled stream benchmark: 120,000 refusal deltas, 14,159,999 input bytes, 120,000 output characters — **4,442.98 ms**; comparable output-text deltas, 15,119,999 input bytes — **74.26 ms**.

### Limitations

`/pr-review-toolkit:review-pr all parallel` is unavailable in this environment (exit 127), so I performed the read-only six-lens fallback and reconciled three independent specialist reports. No live Anthropic/Bedrock call, native Windows execution, credentialed redteam generation, or browser session was used; SDK/SSE/parser probes and current CI covered those paths without credentials.

PR #10096, opened 2026-07-14
feat(helm): make ingress configurable

denzhel:
## What

Replace the hard-coded Ingress template with the standard configurable shape: `ingress.enabled`, `ingress.className`, `ingress.annotations`, `ingress.hosts[]`, and `ingress.tls[]`.

## Why

`templates/ingress.yaml` currently hard-codes `ingressClassName: traefik` and `cert-manager.io/cluster-issuer: letsencrypt-cloudflare`, and is always rendered. On any cluster not running Traefik + that exact ClusterIssuer, this produces a non-functional Ingress (no controller claims it) and, via the cluster-issuer annotation, a cert-manager `Certificate` that never issues. There is currently no way to change the class/host/annotations or disable the Ingress through values.

## Backward compatibility

Fully preserved. `domainName` is retained as a deprecated fallback: when `ingress.hosts` is **unset**, the template reproduces the previous hard-coded Ingress derived from `domainName` — same resource name (`<release>`), same TLS secret (`<release>-certificate-tls`), the `letsencrypt-cloudflare` cert-manager annotation, and `ingressClassName: traefik`. So a default install, and any existing 0.1 install that set only `domainName`, render a functionally identical Ingress on upgrade (no resource rename, no orphaned certificate).

Once a user sets `ingress.hosts`, the legacy defaults no longer apply: `ingress.annotations` and `ingress.tls` are used verbatim (empty by default), so a custom controller doesn't inherit the cert-manager annotation and no TLS is requested for the example host.

The only diffs in the default render vs the previous template are cosmetic: standard chart labels (consistent with the Deployment/Service) instead of the ad-hoc `app:` label.

## Example — nginx, no cert-manager

```yaml
ingress:
  enabled: true
  className: nginx
  annotations:
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
  hosts:
    - host: promptfoo.internal.example.com
      paths:
        - path: /
          pathType: Prefix
  tls: []
```

## Verification

- `helm lint` passes.
- **Default install** renders an Ingress functionally identical to the previous one (diff is labels only).
- **Legacy upgrade** (release `demo`, only `domainName` set) → name `demo`, TLS secret `demo-certificate-tls`, the configured host, and the cert-manager annotation — all preserved.
- **Modern** (`ingress.hosts` + `className: nginx`, no `tls`/`annotations`) → nginx class, no cert-manager annotation, no phantom `promptfoo.example.com` TLS.
- `ingress.enabled: false` renders no Ingress.

Chart `version` bumped 0.1.0 → 0.2.0 (`appVersion` untouched).


denzhel:
Thanks for the review — all five points were about the upgrade path for existing 0.1 installs, and they're fair. Addressed in `8e1792b` with a legacy-fallback approach: when `ingress.hosts` is unset, the template reproduces the old hard-coded Ingress from `domainName`; once `ingress.hosts` is set, everything is neutral/replaceable.

- **Preserve `domainName` overrides** — `domainName` is kept (marked deprecated) and used as the host when `ingress.hosts` is empty, so existing installs render their real host on upgrade.
- **Release-scoped TLS secret** — the fallback TLS secret is derived as `{{ printf "%s-certificate-tls" .Release.Name }}`, so a `demo` release keeps `demo-certificate-tls` instead of a literal `promptfoo-certificate-tls`.
- **No example TLS host as an active default** — `ingress.tls` defaults to `[]`. The `promptfoo.example.com` TLS entry now only appears on the legacy `domainName` path; a user who sets `ingress.hosts` gets no phantom TLS for the example host.
- **Replaceable annotations** — `ingress.annotations` defaults to `{}`. The `cert-manager.io/cluster-issuer` annotation is only injected on the legacy path, so a user supplying their own annotations no longer inherits it (no failing certificate request on clusters without that issuer).
- **Preserve the Ingress resource name** — `metadata.name` is now `{{ .Release.Name }}` (as before), so upgrades patch the existing Ingress instead of renaming it to `<release>-promptfoo`.

Verified with `helm template` across three scenarios — default install (identical to the previous output bar the standard labels), legacy upgrade (`demo` release with only `domainName` set → name `demo`, secret `demo-certificate-tls`, cert-manager annotation retained), and a modern nginx override (no cert-manager annotation, no phantom TLS). `helm lint` passes.

denzhel:
@codex review

denzhel:
Addressed in 3d65ead: `ingress.className` now defaults to empty; the Traefik default only applies on the legacy `domainName` fallback path, so custom `ingress.hosts` installs use the cluster's default IngressClass unless they set `className` explicitly.

@codex review

denzhel:
@codex review

denzhel:
@codex review

haim-bp (approved):

denzhel:
@codex review

mldangelo (commented):
Thanks for this — and a heads-up before anything else: **I pushed one small commit to your branch**, so don't be surprised to find a commit you didn't write.

**What I pushed:** a pure comment/whitespace reorder in `helm/chart/promptfoo/values.yaml`, moving the commented `hosts:`/`tls:` examples *below* the live `hosts: []`/`tls: []` keys instead of above them. As written, the examples repeat the mapping key, so uncommenting them in place produces duplicate YAML keys — Helm parses values non-strictly, last occurrence wins, and the user's config is discarded silently:

```yaml
# rendered from the block uncommented in place (className: nginx, host: promptfoo.mycompany.com)
metadata:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-cloudflare   # legacy branch fired
spec:
  ingressClassName: nginx
  tls:
    - hosts: ["promptfoo.example.com"]        # user's host dropped
      secretName: demo-certificate-tls        # user's secret dropped
```

`helm lint` exits 0 on that input. Deleting only the two `[]` lines makes the user's values render correctly, which isolates the cause to duplicate-key parsing. After the reorder, a full `helm template` diff is byte-identical for every scenario, and a naive uncomment now fails loudly with a YAML parse error instead of silently reverting. Please re-pull before your next push.

## What works well

The back-compat design holds up under testing, which is the hard part here. Default render is legacy-equivalent (traefik / `letsencrypt-cloudflare` / `<release>-certificate-tls` / `domainName`), `ingress: null` hits the nil-map guard cleanly, `enabled: false` suppresses only the Ingress, and the annotation precedence you documented ("your keys win on conflict") is accurate — `--set 'ingress.annotations.cert-manager\.io/cluster-issuer=my-issuer'` renders `my-issuer`. Existing installs that set only `domainName` upgrade with no change. The iteration on the earlier review feedback was thorough.

## Remaining items

### P2 — no render coverage for the new back-compat logic
The template now carries four non-obvious behaviors (legacy fallback branch, conditional traefik default, annotation merge with user-wins precedence, nil-map guard) and nothing protects them. `grep -rl helm test/` → 0 matches; `grep -rl helm .github/` → only `CODEOWNERS`. The one existing artifact, `templates/tests/test-connection.yaml`, is a live-cluster `helm test` hook, not a render assertion, and isn't run in CI. All four behave correctly today — I verified them by hand — but the legacy branch is exactly the kind of thing that rots silently. A handful of `helm template` golden-file assertions would be cheap. Happy for that to land separately if you'd rather not grow this PR.

### P3 — `ingress.annotations` and `domainName` are undocumented in site/
`grep -rn "domainName" site/` and `grep -rn "annotations" site/docs/usage/self-hosting.md` both return zero hits. The new prose refers to "the deprecated `domainName` value" without saying where it lives or that it defaults to `promptfoo.example.com`, and `annotations` — the field self-hosters actually need for cert-manager issuers, nginx body-size, ALB/GKE — isn't in the example block at all, despite this PR being what gives it its merge semantics. Worth adding `#   annotations: {}` with a representative entry plus a line naming the `domainName` default and the third piece of the fallback (the `<release>-certificate-tls` TLS entry).

### P3 (nit, pre-existing, same block) — key typo in the docs example
`site/docs/usage/self-hosting.md:210` uses `persistentVolumeClaims:`, but the chart uses `persistentVolumesClaims` (`values.yaml:70`, `templates/pvc.yaml:1`). A user copying that example to request `size: 10Gi` silently gets the default 1Gi. Not yours, but it's the same block you're revising, so it's a free fix if you want it.

### Question — dropped `metadata.namespace`
The new template omits `namespace: {{ $.Release.Namespace | quote }}`, which the old one emitted. Intentional? It looks right to me — `deployment.yaml` and `service.yaml` omit it too, and Helm applies into the release namespace either way — but the PR body says the only default-render diff is the label change, and this is a second one. Only matters for anyone running `helm template | kubectl apply` without `-n`.

### Simplification (optional)
`templates/ingress.yaml:32-41` hand-rolls the TLS block with a nested `range` that re-emits `hosts` and `secretName`. `$tls` is passed through untransformed, so this is pure re-serialization — the `toYaml . | nindent` idiom the chart uses everywhere else (`deployment.yaml` ×13, `serviceaccount.yaml:10`, `pvc.yaml:11`, and your own annotations block six lines above) collapses it:

```yaml
  {{- with $tls }}
  tls:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

Verified structurally identical across defaults, legacy path, multi-host/multi-secret, `ingress: null`, and `enabled: false`; only textual delta is host-list quoting. Side benefit: the hand-rolled version drops any `tls` field other than `hosts`/`secretName`, while `toYaml` passes entries through verbatim. Minor consistency point in the same area — lines 14 and 16 use `if not X` + assignment while lines 9 and 43 use `| default`; sprig's `default` treats empty string and empty list the same as `not`, so they're interchangeable if you want one idiom.

Nice work on this — the fallback semantics are the tricky bit and you got them right.

PR #10097, opened 2026-07-14
fix(python): keep worker control markers on their own line

potatoboiler:
The messages seen by worker.ts split by newlines, and if the `DONE|` signal gets clobbered into a stdout line from agent output, then it never gets picked up by the promptfoo process and the test hangs.

ref: https://github.com/promptfoo/promptfoo/blob/0.121.17/src/python/worker.ts#L75-L87


mldangelo-oai (approved):
Thank you!

PR #10098, opened 2026-07-14
fix(assertions): handle null tool call output

mldangelo-oai:
`is-valid-openai-tools-call` threw when called directly with null output, or with omitted/undefined output during MCP-result detection. Return the normal failed grading result for all three cases. Valid tool calls and MCP success/error handling keep their existing behavior.

Validation: 47 assertion tests, TypeScript, core build, lint/format, and hooks pass. Full backend: 25,029 passed; seven failures are known mismatches in the shared local Hono, js-yaml, and protobufjs install. The built package API returns failed grades for all three missing-output forms. A three-case built CLI evaluation confirms valid tool grading and the evaluator's existing `No output` handling.


jameshiester-oai (commented):
## Review summary

**Reviewed base:** `1ede17aaed940e6dff04f71d24e4ecc011809dae`  
**Reviewed head:** `396722f74873a235a7f193d5b25b7b59b2423983`  
**Scope:** Fresh full review of the complete base...head diff.

**Verdict:** One P1 build blocker. The production null guard returns the expected failed grading result through the public assertion API, but the new regression fixture does not typecheck, so the branch cannot complete the required build.

### Verified findings

- **P1:** Adding `null` to the directly typed handler fixture widens `toolsOutput` beyond `AssertionParams.output: string | object`. Local TypeScript and five exact-head CI jobs fail with TS2322; two other build jobs are cancelled by fail-fast.
- **P2/P3:** None.

### Existing feedback

Copilot and Promptfoo Scanner left summary-only reviews with no inline findings. There are no existing review threads to duplicate.

The smallest maintainable test shape is to cover `null` through the public `runAssertion({ providerResponse: { output: null } })` boundary, which both reflects the runtime contract and remains type-correct. An explicit narrow cast at the direct-handler boundary is a less representative alternative.

### Tests and QA

- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci` — passed; exact-lock dependencies installed with Node 24.18.0 / npm 11.16.0.
- `npx vitest run test/assertions/openai.test.ts --sequence.shuffle=false` — 44/44 passed.
- `npx vitest run test/assertions/openai.test.ts --sequence.seed=10098` — 44/44 passed.
- `npm run tsc -- --pretty false` — failed with TS2322 at `test/assertions/openai.test.ts:916`: `null` is not assignable to `string | object`.
- `npx @biomejs/biome check src/assertions/openai.ts test/assertions/openai.test.ts` — passed.
- Public `runAssertion` runtime probe — `null` and `{ tool_calls: null }` both returned `pass:false`, `score:0`, with the expected invalid-tools reason.
- Credential-free local CLI eval with a provider returning `null` — exited 100 with one normal failed eval and zero runtime errors; the main evaluator short-circuits this path as `No output` before assertion grading.
- Exact-head CI logs — Smoke, Share, Redteam Production, Redteam Staging, and Build Node 26 all fail on the same TS2322; other visible test matrices pass.

### Limitations

The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (exit 127), so its six lenses were applied manually and reconciled with three independent read-only specialist reviews. No external provider, browser, or full repository suite was run. A full build was not repeated locally because its TypeScript gate is already deterministically failing.

test/assertions/openai.test.ts
```diff
@@ -898,6 +898,7 @@ describe('OpenAI assertions', () => {
       // malformed tool_calls entry. The guard is meant to reject these with a clear reason,
       // not crash on `toolsOutput[0].function.name` when `.function` is missing/null.
       const malformedOutputs = [
+        null,
```
jameshiester-oai: [P1] Keep the null regression type-correct

Adding `null` here widens `malformedOutputs`, so the loop passes `null` to a handler whose `AssertionParams.output` contract is `string | object`. `npm run tsc` fails with TS2322 at line 916, and the same error breaks Smoke, Share, both Redteam jobs, and Build Node 26 on this head. Please cover this case through the public `runAssertion({ providerResponse: { output: null } })` boundary (which returns the intended failed grading result), or otherwise keep the deliberate runtime-only value out of this typed handler fixture, then rerun the typecheck.

src/assertions/openai.ts
```diff
@@ -40,7 +40,7 @@ export const handleIsValidOpenAiToolsCall = async ({
   }
 
   // Handle traditional OpenAI function/tool calls
-  if (typeof output === 'object' && 'tool_calls' in output) {
+  if (output !== null && typeof output === 'object' && 'tool_calls' in output) {
```
jameshiester-oai, line 43: [P2] Handle omitted output along with `null`. `ProviderResponse.output` is optional, so the public `runAssertion` API can pass `undefined` here; before this guard is reached, `JSON.stringify(undefined)` also returns `undefined` and the MCP scan calls `.includes`, throwing a TypeError. I reproduced the same public call with `{ output: null }` returning the intended `pass:false`, while `providerResponse: {}` throws. Normalize the serialized value to a guaranteed string (or reject nullish output before the MCP scan) and add omitted/undefined coverage beside the new null regression.

jameshiester-oai (commented):
Reviewed base `5469ee52d2ea5387dfe62c080698c0e2e2b9b6af` and head `7f7ae2a5516380a061fd64eff218028e1a9b008a`.

Scope: incremental from authenticated reviewed ancestor `396722f74873a235a7f193d5b25b7b59b2423983` (confirmed ancestor), with the complete current two-file base...head patch and public assertion caller reviewed.

Verdict: the prior P1 TypeScript blocker is fixed, but one adjacent P2 runtime defect remains. Exact-head CI has 42 successful checks and one neutral aggregate CodeQL check; all language analyses and build/test lanes succeeded.

Findings:
- P2: omitted/`undefined` provider output still throws before reaching the new malformed-output failure path. `ProviderResponse.output` is optional, and a direct public `runAssertion` call with `providerResponse: {}` throws `Cannot read properties of undefined (reading 'includes')`; the same call with `null` now returns `pass:false`, `score:0`.
- Existing feedback: the prior type-invalid test-fixture P1 is resolved and outdated on this head; no live finding was duplicated.
- Simplification/test opportunity: normalize the MCP scan input once to a guaranteed string and cover omitted output in the same public-boundary table as null.

Validation:
- `npx vitest run test/assertions/openai.test.ts --sequence.shuffle=false` — 45/45 passed.
- `npx vitest run test/assertions/openai.test.ts --sequence.seed=10098` — 45/45 passed.
- `npm run tsc -- --pretty false` — passed, confirming the old P1 is fixed.
- Targeted `biome check` and `git diff --check` — passed.
- Exact-head public `runAssertion` probe — null failed cleanly; omitted output reproduced the TypeError.
- Credential-free local CLI eval with a null-output provider, `--no-cache`, `--env-file /dev/null`, and sharing disabled — exited 100 with one normal failed eval, zero runtime errors, and `No output`.

Limitations: no external provider or browser was used. The full build was not repeated locally because TypeScript and the exact-head Node 20/24/26 build checks are green. The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable (exit 127), so its six lenses were applied manually and reconciled with three read-only specialists.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

src/assertions/openai.ts
mldangelo-oai: Already fixed at the current head. outputStr now uses JSON.stringify(output) ??  before the MCP includes checks, and the parameterized regression covers providerResponse omitted output, explicit undefined, and null; each returns pass false with score 0 and the expected reason.

jameshiester-oai (commented):
## Exact-head review

Reviewed the complete three-file diff, prior authenticated reviews and current reply, nullish output handling in the assertion dispatch and OpenAI tool handler, parameterized tests, docs, and exact-head CI. The earlier authenticated null/undefined output issue is fixed by the `JSON.stringify(output) ?? ''` fallback and null guard; direct undefined, omitted, and null provider responses now return normal failed grading results. No additional distinct actionable issue was validated.

Verification: after lockfile-aligned `npm ci`, 47 focused assertion tests and root TypeScript passed, as did diff check. All 44 exact-head checks currently pass. No tracked files changed.

PR #10102, opened 2026-07-15
fix(prompts): CSV rows collapse to duplicate labels/IDs when basePrompt.label is set

chuenchen309:
## Problem

Unlike `jsonl.ts` (which appends row content to the label when there are multiple rows and a shared label is set), `csv.ts` always used `basePrompt.label` verbatim for every row. Since `generateIdFromPrompt` hashes label first, every row in a labeled CSV prompt file got the **same** generated prompt ID. `buildPromptIndexMap` in `evaluator.ts` keys prompt columns by `provider:generatedId`, so all rows collapsed onto a single `promptIdx` — every row's results and metrics were attributed to one prompt column. (Verified empirically: on main a 2-row labeled CSV yields `promptIdx=1` for both results; with this fix they get distinct labels, IDs, and `promptIdx` 0/1. Note this does not change multi-turn `_conversation` grouping: the `conversationKey` at `src/evaluator.ts:677` uses the raw `prompt.id` field, which is unset for label-only CSV prompts before and after this change.)

## Fix

Added `buildCsvRowLabel()`, mirroring `jsonl.ts`'s `containsMultiple` pattern: only reuse `basePrompt.label` verbatim for a single row; for 2+ rows, disambiguate as `` `${baseLabel}: ${content}` ``. Applied to all three label-assignment sites in this file (the delimiter-free branch, the parsed-CSV branch, and the parse-error fallback branch).

## Test plan

- New test: multi-row CSV with `basePrompt.label` set asserts distinct, disambiguated labels (and therefore distinct prompt IDs) instead of colliding.
- Verified red/green: reverted the source fix locally, confirmed the new test fails with `expected 1 to be 2` (labels collapsed to a single unique value); restored the fix, confirmed all 11 tests in the file pass.
- Ran the full `test/prompts/` suite (170 tests) — all pass.
- `biome check` and `tsc --noEmit` — clean.

## Disclosure

Found and fixed with AI-assisted code review (Claude Code) as part of a broader pass over `src/prompts/processors/`; I wrote/ran the repro and tests myself and reviewed the diff before opening this PR.

## Maintainer additions (pushed to this branch)

- Simplified `buildCsvRowLabel` (dropped the `rowLabel` param that two of three call sites filled with `undefined`; `row.label ||` hoisted to the one site that has it).
- Added tests pinning the parsed-CSV and malformed-CSV fallback branches (reverting either branch's fix now fails a test).


chuenchen309:
Thanks — both are worth naming, but neither blocks this fix:

1. **Synthesized-label collision:** the `${label}: ${content}` synthesis intentionally mirrors the existing `jsonl.ts` convention (`containsMultiple ? \`${prompt.label}: ${json}\` : ...`). The contrived `Group` + `Group: foo` collision is therefore a pre-existing property of that shared convention rather than something introduced here — the goal was to make CSV consistent with JSONL, not diverge from it.

2. **Inherited `id`:** the `basePrompt.id` → `conversationKey` inheritance predates this change (the original code already spread `...basePrompt`). This PR only touches label disambiguation, so I kept the `id` behavior out of scope to keep the diff surgical. Happy to clear `id` per-row in a follow-up if you'd prefer.

mldangelo-oai (approved):
Release audit on c60cb43929bc8b070567faf07570da99befa5ee9: synced to current main, consolidated CSV prompt construction, fixed inherited-ID/explicit-label collisions, and verified a two-prompt CLI eval with colliding labels plus focused tests (15 tests). CI is green at the exact head; no release-blocking findings.

mldangelo (commented):
Thanks for this one — nice catch, and the diagnosis is the good part. Spotting that `label` is what feeds `generateIdFromPrompt`, so a config-level `label` on a multi-row CSV silently collapses every row onto one prompt id (and one result column), is exactly the kind of bug that hides forever. Mirroring the existing `jsonl.ts` `containsMultiple` convention rather than inventing a new label format was the right call, as was keeping per-row `label` precedence and leaving the single-row case as a bare label so nothing back-compat breaks.

## Heads up: I pushed one commit to your branch

You'll see a commit on `fix/csv-prompt-label-disambiguation-v2` you didn't write — `9b7ddb16`. Sorry for the surprise; it was mechanical enough that pushing beat a round-trip. It's test-only, no production code touched.

(To be clear, `0bad82ca`, `b8a4ba81` and the `c60cb439` main-merge are all your own commits — I'm only responsible for `9b7ddb16`.)

- `9b7ddb16` — the fallback-branch test. Its fixture (`prompt\n"Malformed CSV…`) had no comma, so `content.includes(',')` was false and it returned from the delimiter-free branch without ever entering the `try`/`catch`. Reverting the catch branch alone left the suite fully green:

```
catch branch (csv.ts:94) reverted -> npx vitest run test/prompts/  ->  172 passed (172)
```

Real-file check with a comma-containing malformed CSV: 4/4 distinct ids with the fix, 1/4 with the catch branch reverted. Fixture swapped for one that contains the delimiter and fails `csv-parse`; now red-on-revert. Worth editing the PR body — the "reverting either branch's fix now fails a test" claim wasn't true for that branch. 15/15 green at tip.

## Remaining

**1. Docs (P3).** `site/docs/configuration/prompts.md:106-115` documents only the per-row `label` column and never says what happens when the prompt *entry* sets `label`. That's now user-visible in column headers and exports. Two or three lines after the example covering row-label precedence, the `<label>: <text>` disambiguation, and `<id>:<n>` would close it — `src/prompts/AGENTS.md` asks for this.

**2. Question, not a request: duplicate *explicit* row labels.** Distinct from the collision the bot raised (synthesized vs. explicit) — this is explicit vs. explicit, which the new dedupe deliberately skips via `if (!row.label)`:

```
prompt,label
"Tell me about A","Basic Query"
"Tell me about B","Basic Query"
-> distinct labels: 1/2   distinct generateIdFromPrompt ids: 1/2
```

Same collapse the PR fixes, different input path. It's pre-existing and the PR doesn't worsen it, so scoping it out is completely defensible — was that intentional, or worth folding in?

**3. Nit.** `csv.ts:77-84` and `csv.ts:104-111` are byte-identical after the extraction. A small `processLineDelimitedCsv(content, basePrompt)` would collapse both to one call. Optional.

mldangelo-oai:
Thank you!

PR #10105, opened 2026-07-15
docs(site): add OpenAI-compatible gateway example

seven7763:
## Summary

Adds a short section on the OpenAI provider docs for **OpenAI-compatible multi-model gateways**, plus a small runnable example.

- `site/docs/providers/openai.md` — `apiBaseUrl` / `OPENAI_BASE_URL` example with [DaoXE](https://daoxe.com) (`https://daoxe.com/v1`)
- `examples/openai-compatible-gateway/` — config + README (account-scoped model ID placeholder)

No new first-class provider code. Reuses the existing `openai:chat` stack.

### Notes for reviewers

- Model IDs are account-scoped (`GET /v1/models` / dashboard); no static catalog.
- Chat Completions path only; other DaoXE protocols (e.g. Anthropic Messages) are out of scope for this provider.
- DaoXE is not available in mainland China.
- Contributor is affiliated with DaoXE.

## Test plan

- [ ] Docs section renders under OpenAI provider page
- [ ] Example config is valid YAML / schema-friendly
- [ ] Optional: `npx promptfoo@latest eval` with real `OPENAI_API_KEY` + model ID against `https://daoxe.com/v1`


seven7763:
Pushed a follow-up fixing the example README standards check: include `npx promptfoo@latest init --example openai-compatible-gateway` (was failing `test/examples/readme.test.ts`).

mldangelo:
Thanks for the contribution and for the affiliate disclosure. I pushed `8ca182e` making the section and example vendor-neutral: the flagship OpenAI page doesn't name individual gateway vendors (per-vendor setups get dedicated provider pages — see `site/docs/contributing.md` for the eligibility path), and the `apiBaseUrl`/base-URL env precedence was already documented in the tables directly above the new section, so the section is now a short pointer. The runnable example fills a real gap (no existing `openai-*` example demonstrates a custom `apiBaseUrl`), so I kept it under the generic name with a placeholder host.

If you'd like DaoXE listed in the docs specifically, the right route is a dedicated provider page via the contributing guide's gateway/reseller criteria.

seven7763:
@mldangelo Thanks for rewriting this to the vendor-neutral gateway pointer + generic example — that matches the contributing guide better than a per-gateway name on the flagship OpenAI page.

I pushed a small follow-up for the Codex notes on the current head:

1. **`OPENAI_API_HOST` precedence** — docs/example now state the real resolution order (`apiHost` / `OPENAI_API_HOST` → `apiBaseUrl` → `OPENAI_*_BASE_URL` → default). Ambient `OPENAI_API_HOST` overrides `apiBaseUrl`.
2. **`apiKeyEnvar` / key fallback** — example + docs recommend `apiKeyEnvar: GATEWAY_API_KEY` and note that missing/empty key config still falls through to `OPENAI_API_KEY` (so a leftover OpenAI key can be sent to the gateway host if you only set `apiBaseUrl`).

Happy to adjust further. For a DaoXE-named provider page later, I’ll follow `site/docs/contributing.md` gateway/reseller criteria separately rather than reintroducing a vendor name on this generic path.

mldangelo-oai (approved):
Release audit on bf9618b8b7f9b50d2d0c9a90d8314da75c87eeb4: clarified initialization order, environment fallback, and the supported no-auth gateway configuration, then reran formatting/docs checks. CI is green at the exact head; no release-blocking findings.

mldangelo (commented):
Thanks for sticking with this through the vendor-neutral rewrite — the result is a genuinely useful addition. A few things I want to call out specifically:

- The docs gap was real. There was no page saying "point the built-in OpenAI provider at an arbitrary OpenAI-compatible endpoint," only per-vendor pages, and the new section covers it without naming vendors.
- The `apiKeyEnvar` guidance is the **first place in the repo** to raise gateway key isolation at all. `site/docs/providers/vllm.md:134` and `litellm.md:103` still tell users to put a third-party key straight into `OPENAI_API_KEY`. Disclosing the `OPENAI_API_KEY` fallthrough in the Notes is a real improvement over the status quo.
- The example is properly runnable: schema header, `init --example` line that satisfies `test/examples/readme.test.ts`, and an actual assertion rather than a placeholder.

## P1 — the URL resolution order is inverted on the current base

`site/docs/providers/openai.md:1650` and `:1655`, and `examples/openai-compatible-gateway/README.md:35`.

This one is not your fault: the earlier precedence note on this thread was accurate against the base at the time, but **#10128 (`5c216a1bc3`, 2026-07-17) rewrote `getApiUrl()`** and that commit is in this PR's head. Before it, `config.apiHost || env OPENAI_API_HOST` was resolved first; now `src/providers/openai/index.ts:96-114` reads:

```ts
if (this.config.apiHost)    { return `https://${this.config.apiHost}/v1`; }
if (this.config.apiBaseUrl) { return this.config.apiBaseUrl; }          // <-- before the env var
const envApiHost = this.env?.OPENAI_API_HOST || getEnvString('OPENAI_API_HOST');
```

Observed on this branch's head, instantiating the real class:

```
ambient OPENAI_API_HOST + config.apiBaseUrl   -> https://gateway.example.com/v1
provider env: OPENAI_API_HOST + apiBaseUrl    -> https://gateway.example.com/v1
ambient OPENAI_API_HOST alone                 -> https://corp-proxy.internal/v1
```

`test/providers/openai/index.test.ts:40` ("should prefer an explicit API base URL over an environment API host") pins exactly this and passes (`npx vitest run test/providers/openai/index.test.ts` → 19/19). `site/docs/providers/aws-bedrock.md:1090` and `src/providers/AGENTS.md:38` ("Explicit options > Environment variables") already state the post-#10128 rule, so the new section is the only place that disagrees.

Correct order for the list:

1. `config.apiHost` (host only)
2. `config.apiBaseUrl`
3. `OPENAI_API_HOST` (provider `env:` or ambient)
4. `OPENAI_API_BASE_URL` / `OPENAI_BASE_URL`
5. default OpenAI host

Practical impact is modest — a user told to `unset OPENAI_API_HOST` sees no change and chases a phantom cause — but it's the section's central new claim, and it contradicts a test.

## P2 — the setup block clears the host but not the key

`examples/openai-compatible-gateway/README.md:12-13`. The comment says "clear ambient OpenAI host/key overrides," but only `unset OPENAI_API_HOST` runs. With the P1 fix, that `unset` is a no-op for the shipped config (which sets `apiBaseUrl`), so the block ends up doing nothing it claims to do — while the one thing that *would* help here, `unset OPENAI_API_KEY`, is missing. Given `getApiKey()` falls through to `OPENAI_API_KEY` whenever `GATEWAY_API_KEY` is unset or empty, was leaving the key half out intentional?

## Two cleanups, take or leave

1. Once the list is correct, the trailing paragraph at `openai.md:1655` has nothing left to warn about — deleting it is simpler than rewording it.
2. Three of the four `## Notes` bullets restate the docs section the README already links at line 39. `examples/provider-mlflow-gateway/README.md` and `examples/provider-litellm/README.md` both defer to their provider docs instead. Keeping only the "Chat Completions API shape only" bullet puts the resolution-order semantics in one place, so the next `getApiUrl()` change has one doc to update rather than two.

The outstanding notes on `README.md:23` (edit step ordering) and `:36` (`is not set` vs. the site doc's more accurate `missing or empty`) still look valid to me as well.

seven7763:
Thanks @mldangelo — really appreciate you taking this through the vendor-neutral rewrite and the detailed notes on the `getApiUrl()` order change after #10128.

Happy with the generic OpenAI-compatible section as the right home for this. I won't block on a DaoXE-specific page; if/when we meet the dedicated provider eligibility path in `site/docs/contributing.md`, I'll open a separate PR there.

On the inverted URL precedence: if anything on this branch's docs still describes the pre-#10128 order, I'm happy to push a follow-up commit that matches `src/providers/openai/index.ts` (`apiHost` → `apiBaseUrl` → `OPENAI_API_HOST`). Branch looks **behind** at the moment — let me know if you'd rather I rebase + fix the docs note, or if you'll fold it in on merge.

Thanks again for the thorough review.

seven7763:
Not a nudge — a correction on my own branch. I re-checked against current `main`, and the P1 you raised is still live, so this isn't merge-ready as it stands:

- `site/docs/providers/openai.md` lists `config.apiBaseUrl` first and `config.apiHost` second, but `getApiUrl()` in `src/providers/openai/index.ts` returns `config.apiHost` before `config.apiBaseUrl` — the two entries are swapped. Codex flagged the same line again on `cf8603b5`.
- The branch is also behind `main` (last sync 2026-08-02), so the merge button is blocked regardless.

Both are quick. Happy to push the swap and merge `main` in one commit — just say the word, or fold it in on your side if you'd rather not churn CI. No urgency either way, and thanks again for taking this through the vendor-neutral rewrite.

seven7763:
@mldangelo P1 confirmed and fixed in `c965d3d`, and the branch is merged up to `main` in `8d0d86d`.

I re-read `getApiUrl()` at current `main` rather than take the review on trust — `src/providers/openai/index.ts:96-114`:

```ts
if (this.config.apiHost) {
  return `https://${this.config.apiHost}/v1`;
}
if (this.config.apiBaseUrl) {
  return this.config.apiBaseUrl;
}
const envApiHost = this.env?.OPENAI_API_HOST || getEnvString('OPENAI_API_HOST');
```

`config.apiHost` short-circuits before `config.apiBaseUrl`, so my list had the first two entries backwards. It now reads:

1. `config.apiHost` (host only; Promptfoo builds `https://<host>/v1`; highest priority)
2. `config.apiBaseUrl` (full URL)
3. `OPENAI_API_HOST` (host only; provider `env:` or ambient)
4. `OPENAI_API_BASE_URL` / `OPENAI_BASE_URL`
5. the default OpenAI API host

That matches `test/providers/openai/index.test.ts` and the "Explicit options > Environment variables" rule in `src/providers/AGENTS.md`, so the section no longer contradicts the rest of the docs.

On P2: the example README already does both `unset OPENAI_API_KEY` and `unset OPENAI_API_HOST` in the setup block, so the key half is covered. I left the `OPENAI_ORGANIZATION` and dotenv points alone for now — happy to fold them in if you'd rather the example be exhaustive about clearing ambient state, but they felt like scope creep on a docs PR.

Merged `main` in rather than rebasing, so nothing was force-pushed and the review history stays intact. Branch is at 0 commits behind.

mldangelo-oai:
thanks!

PR #10107, opened 2026-07-15
fix(code-scan): extractValidLineRanges off-by-one on trailing newline

chuenchen309:
## Summary

`extractValidLineRanges` (`src/codeScan/util/diffLineRanges.ts`) computes valid PR-review-comment line ranges by walking `unifiedDiff.split('\n')`. When the diff string ends with a trailing newline, `split('\n')` produces one extra empty-string element at the end of the array. The loop's context-line branch (`line === ''`) treated that trailing element as if it were real content and incremented `currentNewLine` one time too many, so the computed range's `end` landed one line past the diff's true last line.

This range feeds `getPRDiffRanges` in `code-scan-action/src/github.ts` (around line 116), which fetches the PR diff via octokit with `mediaType: { format: 'diff' }` — trailing-newline-terminated by GitHub convention — and uses the range to clamp/validate inline PR review comment line numbers. The off-by-one could clamp or accept a comment onto a line number that doesn't actually exist in the diff.

The sibling `annotateDiffWithLineRanges` in `src/codeScan/git/diffAnnotator.ts` already guards this exact case (`else if (line === '' && isLastLine)`); `extractValidLineRanges` had no equivalent guard, which is the reason I read this as an unintentional omission rather than deliberate behavior.

## Fix

Track whether the current line is the last line of the split array (mirroring `diffAnnotator.ts`'s existing pattern), and skip incrementing `currentNewLine` for a trailing empty line only when it is the last line.

## Verification

- Added a new test in `test/codeScans/util/diffLineRanges.test.ts` with a diff string ending in `\n`; confirmed it failed against the pre-fix code (`end: 13` instead of the correct `end: 12`) before applying the fix, then passed after.
- `test/code-scan-action/github.test.ts` already had a test whose mock diff response ends with a trailing newline (matching real octokit `format: 'diff'` behavior) and whose expectation (`line: 61`) was itself the buggy off-by-one value — the fixture's true last diff line is 60. Updated that expectation to the correct value; this is independent evidence the bug had real-world reach into GitHub PR-comment placement.
- `npx vitest run test/codeScans test/code-scan-action`: 219/219 passing (18 files), no regressions.
- `npm run tsc`: clean.
- Biome lint clean on changed files (one pre-existing `noExcessiveCognitiveComplexity` warning on this function predates this change — confirmed present on `main` before my edit too; not introduced by this fix, and it's a non-blocking warning).

## Disclosure

I used Claude Code (Anthropic's AI coding assistant) to help investigate and draft this fix; I reviewed the diff, ran the verification above myself, and take responsibility for the change. Per this repo's AGENTS.md, no AI co-author attribution is included in the commit or this PR body.


mldangelo-oai (approved):
Release audit on c59e14bd2bb00ddbf8f686735cfbe1240b02cfc9: ran the focused code-scan suites (18 tests) and checked the trailing-newline boundary. The off-by-one is covered and valid line ranges remain unchanged; no release-blocking findings.

mldangelo-oai:
Great find! Thanks!

PR #10110, opened 2026-07-15
fix(providers): isolate Anthropic credentials and Meta Responses

mldangelo-oai:
## Summary

- Keep Anthropic credentials isolated with secret-free, bounded per-instance in-memory caching while retaining same-run hits, cache-TTL and TTL=0 semantics, repeat-namespace isolation, direct/namespaced cache clears, and labeled-provider persistence.
- Prevent authentication and proxy headers, including duplicate case variants, from leaking into Meta, Bedrock, or Cloudflare requests.
- Honor Meta response limits, reject unsupported logprobs, and update schemas and docs.

## Validation

- Focused Anthropic, Meta, Bedrock, Cloudflare, cache/repeat, and contract tests: 437 passed
- TypeScript, architecture, schema, and formatting checks passed


jameshiester-oai (commented):
No actionable findings.

I reviewed the Anthropic cache namespace redesign, TTL and clear semantics, labeled versus ephemeral persistence, custom-header cache bypass, scoped environment overrides, duplicate-case header suppression, Meta/Bedrock/Cloudflare credential isolation, Meta output caps and unsupported parameters, schema generation, tests, and documentation. The credential boundary is coherent: unlabeled instances stay process-local, stable labels are explicit non-secret namespaces, and foreign endpoints suppress inherited Anthropic headers while restoring only their required auth. I did not find a regression, unintended behavior change, documentation mismatch, or safe simplification worth blocking this head.

Validation:
- Node 24.18: focused Vitest passed (7 files, 374 tests).
- npm run tsc passed.
- git diff --check passed.
- Current GitHub checks are green.

No inline comments.

jameshiester-oai (commented):
Reviewed base `08b3f5084d5cd36e9187ac3c79e3fda03d6eefd7` through head `2070713d17df835e0e5122c07fa815b266ab22ae`.

Scope: incremental from the automation-reviewed ancestor `dd1d90ce2d0f8158ec735b32061ea9082e6025d9`, with the complete current base...head PR retained for interaction context. The increment is a merge of main; the provider, contract, docs, and test implementations are unchanged from the reviewed ancestor. Among PR-owned files, only the generated architecture baseline changed to reconcile current counts.

Verdict: no new unique inline finding. Exact-head CI is fully green across builds, platform tests, integration/smoke, site/Web UI, style/architecture, CodeQL, and coverage.

Existing unresolved feedback:
- P2: [the current cache-isolation thread](https://github.com/promptfoo/promptfoo/pull/10110#discussion_r3604979976) remains valid and was not duplicated. Labeled Anthropic providers persist response-cache entries using model/base URL plus the display label, but not credential/account identity. Reusing the same label after switching API keys or OAuth tenants can therefore return the prior tenant's output without contacting the new tenant. An isolated exact-head probe reproduced this: tenant A wrote `tenant-a-output`; tenant B with a different fake key and the same label received that output with `cached: true` and zero B SDK calls.
- The earlier Meta `x-api-key` thread is resolved: current code unconditionally suppresses that header before restoring bearer authorization.

Test and simplification opportunities:
- P3 test gap: the 512-entry ephemeral-cache eviction branch at `src/providers/anthropic/generic.ts:409` has no 513-request boundary test. A focused mocked test should verify oldest-entry eviction while retaining the latest entry. No implementation failure was found there.
- No material simplification opportunity survived review.

Validation:
- `npm ci --ignore-scripts` — passed on a clean retry after this isolated worktree initially lacked dependencies.
- Focused Vitest across Anthropic completion/messages/generic, Bedrock Anthropic, Cloudflare gateway, Meta, and env contracts — 7 files, 374 tests passed.
- `npm run tsc` — passed.
- `git diff --check 08b3f5084d5cd36e9187ac3c79e3fda03d6eefd7...HEAD` — passed.
- Isolated persistent-cache probe — reproduced the existing cross-credential same-label cache hit described above.

Limitations: the requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (exit 127), so its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialist reviews. No live Anthropic/Meta/Bedrock/Cloudflare credentialed call, browser session, full local repository suite, or source modification was performed.

PR #10111, opened 2026-07-15
fix(providers): validate coding-agent targets and cache keys

mldangelo-oai:
## Summary

- Safely render typed Open Interpreter targets, workspaces, approval settings, templated policies, and row overrides without coercing free-form values or leaking authorization in errors.
- Preserve the last valid target during malformed/non-object JSON edits while blocking execution, discovery, preview, save, export, and navigation; correctly handle undo, imports, reload, multi-tab/quota, persistence failures, and asynchronous preflight/dialog paths.
- Reconcile corrected cross-tab state with credential-free, target-bound causal markers that reject stale/mismatched/legacy events and unsafe persistence races, preserve concurrent plugin edits and independent drafts without ping-pong, keep mobile/structured editors reachable and focused, support HTTP/WS/A2A/browser and additional credentialed providers, preserve valid URL-backed endpoints, session testing, A2A no-auth, browser URL templates, and Playwright selectors, prevent constructor-invalid or executable/file/session/credential-bearing stale targets, hosted tools/retrieval, and signed assertions from re-enabling after reload, gate every HTTP test action on the latest validation state, and avoid per-keystroke hashing and sequential-update loss.

## Validation

- 882 changed UI/hygiene tests, 448 focused backend/provider tests, and the final 178-case hydration suite pass, including HTTP multipart, A2A, browser, coding-agent, WebSocket, and shared frontend-contract suites.
- Expanded redteam setup and Eval Creator suites pass, including failing-first hydration, structured-target, cross-tab/quota/TOCTOU/causal-order/independent-draft, undo/import, partial-endpoint, encoded/raw/multipart/runtime, provider-execution/exfiltration, hosted-tool/callback, all HTTP test-gate, multi-turn preview/YAML, persistence-failure/retry, sequential-update, editor-focus/performance, and serialization regressions.
- Root/app TypeScript, production app build, architecture, Biome, diff checks, adversarial persistence/race matrices, and independent exact-head review pass.


src/app/src/pages/redteam/setup/components/Targets/CustomTargetConfiguration.tsx
```diff
@@ -435,20 +460,38 @@ const CustomTargetConfiguration = ({
...
+        return;
+      }
+
+      onConfigErrorChange?.(null);
       updateCustomTarget('config', parsedConfig);
-    } catch (error) {
-      // Allow invalid JSON while typing - error is shown via bodyError prop
-      console.error('Invalid JSON configuration:', error);
+    } catch {
+      updateCustomTarget('config', {});
```
jameshiester-oai: [P2] Preserve the last valid config while JSON is invalid

This branch runs on every editor change, and the red-team Targets page immediately copies each updated provider into useRedTeamConfig. Editing a valid object through a transient invalid state, then navigating Back or unmounting before fixing it, therefore persists config: {} and loses the previous working settings even though validation blocks Next. Keep the parent provider config unchanged while setting customConfigError, and update it only after a successful object parse; the non-object and Format branches have the same issue.

jameshiester-oai (commented):
One actionable P2.

The custom JSON editor now destroys the last valid provider config during transient parse failures; see the inline comment. This is an unintended red-team behavior change because the target editor writes every provider update into shared config immediately. I did not find additional actionable issues in the Open Interpreter prompt-config rendering, typed coercion, nested policy merging, Codex error redaction, Claude cache-key change, Eval Creator validation, or docs.

Validation:
- Node 24.18: focused provider Vitest passed (3 files, 304 tests).
- Focused app Vitest passed (5 files, 103 tests).
- Root and app TypeScript checks passed.
- git diff --check passed.
- Current GitHub checks are green.

src/app/src/pages/redteam/setup/components/Review.test.tsx
```diff
@@ -599,6 +613,396 @@ Application Details:
...
+        <Review
+          navigateToPlugins={vi.fn()}
+          navigateToStrategies={vi.fn()}
+          navigateToPurpose={vi.fn()}
+        />,
+      );
+      await user.click(screen.getByRole('button', { name: /run now/i }));
+      await waitFor(() => expect(callApi).toHaveBeenCalledWith('/redteam/status'));
+
+      latestState = { ...latestState, config: replacementConfig };
```
jameshiester-oai, line 823: [P1] Widen this mocked store state so the app can typecheck

This reassignment changes `latestState.config.target.config` from the inferred `{ url, body }` shape to `{ sandbox_mode }`, producing TS2322 here and at lines 924 and 997. `npm run build:app` fails locally, and the same error currently fails the Node 26 build plus the smoke, share, and both red-team CI jobs before they can run their real checks. Type the mock state with the hook's `Config` shape (or otherwise widen the target to `ProviderOptions`) before assigning `replacementConfig`.

src/app/src/pages/redteam/setup/components/Targets/TargetConfiguration.validation.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+    state.targetConfigDraft = null;
+  });
+
+  it.each([
+    ['malformed JSON', '{"sandbox_mode":"read-only",}', 'Invalid JSON configuration'],
+    ['non-object JSON', '[]', 'Configuration must be a JSON object'],
+  ])('keeps %s blocked after navigating directly to Review', (_case, value, error) => {
+    render(<Harness />);
+
+    fireEvent.change(screen.getByTestId('code-editor'), { target: { value } });
```
jameshiester-oai: [P1] Replace the new `fireEvent` calls before merging

The repository's `test-hygiene` suite rejects these direct `fireEvent` interactions. This branch adds 24 of them across the new target-validation tests, so the web UI CI job fails despite the focused behavior tests passing. Convert the added interactions to `userEvent` (or use the repository's narrow approved exception if the code editor cannot support it) so the full frontend suite is green.

jameshiester-oai (commented):
Re-reviewed exact head `62bf8347cb844fa7d5996ad2a86d6acf3d9bef47`. I found two additional merge-blocking CI regressions. The three existing P2 threads at this head also remain applicable, so I did not duplicate them.

src/app/src/pages/redteam/setup/hooks/useRedTeamConfig.ts
```diff
@@ -12,6 +22,1225 @@ interface RecentlyUsedPlugins {
...
+  }
+  const validScopes =
+    auth.scopes === undefined ||
+    typeof auth.scopes === 'string' ||
+    (Array.isArray(auth.scopes) && auth.scopes.every((scope) => typeof scope === 'string'));
+  if (!validScopes || (auth.tokenUrl !== undefined && !isValidHttpUrlOrTemplate(auth.tokenUrl))) {
+    return false;
+  }
+  if (auth.grantType === 'password') {
+    return typeof auth.username === 'string' && typeof auth.password === 'string';
```
jameshiester-oai: [P2] Validate optional password-grant client fields before clearing the gate

This branch returns true for password OAuth as soon as `username` and `password` are strings, so a recovery target with `clientId: 42` or a non-string `clientSecret` passes `isValidStructuredEndpoint()` and clears the persisted invalid-target marker. The canonical `A2AProviderConfigSchema` rejects those same values (`OAuthPasswordAuthSchema` makes both optional fields strings), so the UI re-enables a target that deterministically fails provider construction. Validate the optional client fields here, or preferably reuse the canonical schema, and cover recovery with wrong-typed optional OAuth fields.

src/app/src/pages/redteam/setup/hooks/useRedTeamConfig.ts
```diff
@@ -12,6 +22,1225 @@ interface RecentlyUsedPlugins {
...
+          continue;
+        }
+      }
+      const cssSelector = engineMatch ? engineMatch[2].trim() : segment;
+      if (
+        /:(?:has-text|text|text-is|text-matches|visible|nth-match|right-of|left-of|above|below|near)(?:\(|\b)/.test(
+          cssSelector,
+        )
+      ) {
+        continue;
```
jameshiester-oai: [P2] Do not treat every balanced Playwright pseudo-selector as valid

This unconditional `continue` accepts malformed selectors whenever the extension name is present. For example, `button:has-text()` is balanced and matches this regex, so target recovery clears the safety marker, but Playwright rejects it with `Unexpected token ')' while parsing css selector` when the step runs. Parse these extensions with Playwright's selector parser (or validate their required argument grammar) before unblocking, and add a balanced-malformed regression alongside the valid `:has-text("Submit")` case.

src/app/src/pages/redteam/setup/hooks/useRedTeamTargetConfigValidation.ts
```diff
@@ -0,0 +1,759 @@
...
+    }
+
+    return { config, serialized: JSON.stringify(target) };
+  } catch {
+    return null;
+  }
+};
+
+const getClearMessage = (serialized: string, token = 'none'): string =>
+  `clear:${token}:${serialized.length.toString(36)}:${targetConfigSha256(serialized)}`;
```
jameshiester-oai: [P2] Keep credentials out of durable reconciliation markers

`serialized` is the verbatim target object, so this digest incorporates API keys, bearer/basic credentials, and secret headers. When localStorage writes fail, `persistTargetConfigClear` stores the resulting verifier in a one-year `Path=/` cookie, which is sent to the server and can land in proxy/server logs. SHA-256 hides the plaintext but remains an offline verifier for low-entropy secrets and candidate configs; the current test only checks that the literal string is absent. Build the marker from a canonical credential-redacted target plus an opaque random/revision token, rather than hashing raw secrets into durable state.

jameshiester-oai (commented):
Reviewed base `fcde2e89a89dc4ca79dcc3012927f50193251759` through exact head `6b57187809076dc325c7eba33006324513002b31`.

Scope: incremental from the last authenticated completed head `a0400537004a12b6ab59c83ff6ca950d78f58e41`, with the full base...head diff and current surrounding contracts checked for interaction context.

Verdict: three P2 correctness/security issues remain in the new target-recovery validation. Exact-head CI is green, including web UI, build, smoke, integration, red-team, docs, style, and platform jobs.

Verified findings:
- P2: A2A password OAuth recovery accepts wrong-typed optional client credentials and clears the safety gate even though the canonical provider schema rejects the target.
- P2: balanced but malformed Playwright pseudo-selectors bypass validation and are only rejected when the browser step executes.
- P2: durable clear markers hash raw target JSON, including credentials; the fallback stores that verifier in a one-year Path=/ cookie.

Existing feedback: all current threads are resolved. The earlier typecheck/test-hygiene regressions are fixed, and I did not duplicate resolved findings.

Simplification/test opportunities: reuse the canonical A2A schema instead of maintaining a partial validator; use Playwright's selector parser (or a shared strict parser); use an opaque or credential-redacted correlation value for cross-tab reconciliation. Add focused regressions for wrong-typed optional OAuth fields, balanced malformed pseudo-selectors, and cookie fallback with credential-bearing targets.

QA:
- `source ~/.nvm/nvm.sh && nvm use && npm run test:app -- src/pages/redteam/setup/hooks/useRedTeamConfig.test.ts src/pages/redteam/setup/hooks/useRedTeamTargetConfigValidation.hydration.test.ts --run` — 403 tests passed.
- Direct `A2AProviderConfigSchema.safeParse` probe rejected password OAuth with numeric `clientId`.
- Direct Playwright Chromium probe rejected `button:has-text()` with a selector parse error.
- Direct marker probe confirmed the clear value is the SHA-256 of the verbatim credential-bearing target JSON.
- `git diff --check fcde2e89a89dc4ca79dcc3012927f50193251759...6b57187809076dc325c7eba33006324513002b31` — passed.
- Exact `/pr-review-toolkit:review-pr all parallel` invocation was unavailable (exit 127); its correctness, tests, docs, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialists.

Limitations: no credentialed provider call or full interactive browser workflow was run; browser validation used an isolated local Playwright parser probe.

src/app/src/pages/redteam/setup/hooks/useRedTeamTargetConfigValidation.ts
```diff
@@ -0,0 +1,787 @@
...
+    return null;
+  }
+};
+
+const redactTargetCredentials = (value: unknown, parentKey?: string): unknown => {
+  if (Array.isArray(value)) {
+    return value.map((item) => redactTargetCredentials(item, parentKey));
+  }
+  if (!isPlainObject(value)) {
+    return value;
```
jameshiester-oai, line 59: [P2] Keep every supported credential representation out of the durable verifier

Returning primitive strings unchanged means the new projection still hashes credentials carried in supported string fields. For example, an HTTP target may put `Authorization: Bearer candidate-a` in `config.request`; `signatureAuth.privateKey`, `tls.passphrase`, and `cli_env.OPENAI_API_KEY` are also missed by the exact-name blacklist. Changing only one of those values changes the cookie hash, so quota fallback still creates a one-year offline credential verifier that is sent on requests. Bind the clear to an opaque persisted revision (preferred), or use an explicit non-secret projection that handles raw requests and every supported secret-bearing field; add equality regressions for raw Authorization, private keys/passwords, and environment-key aliases.

src/app/src/pages/redteam/setup/hooks/useRedTeamConfig.ts
```diff
@@ -12,6 +22,1291 @@ interface RecentlyUsedPlugins {
...
+);
+
+const isValidPlaywrightSelectorExtension = (selector: string): boolean =>
+  !/:(?:text-matches|text-is|has-text|nth-match|right-of|left-of|above|below|near|text)(?![\w-]|\()/.test(
+    selector,
+  ) &&
+  !/:(?:has-text|text|text-is|text-matches|visible|nth-match|right-of|left-of|above|below|near)\(\s*\)/.test(
+    selector,
+  ) &&
+  !invalidPlaywrightTextSelector.test(selector) &&
```
jameshiester-oai: [P2] Validate Playwright extension arguments with the real grammar

These checks still accept malformed balanced selectors because `stripPlaywrightSelectorExtensions` removes their arguments before native CSS validation. At this head the UI accepts `button:nth-match(div)`, `button:right-of(div:)`, and `button:text-matches("x", "z")`, while a real Playwright Chromium probe rejects them respectively for a missing index, invalid nested CSS, and invalid RegExp flags. Recovery can therefore clear the safety gate and re-enable save/run for a target that deterministically fails at execution. Reuse Playwright's selector parser/shared grammar, and cover arity, nested selectors, numeric indexes, regex syntax, and flags.

jameshiester-oai (commented):
Reviewed base `fcde2e89a89dc4ca79dcc3012927f50193251759` through exact head `e7a6e9d130a2f6597289ccccc9d0646ef5977ec8`.

Scope: incremental from the last authenticated completed head `6b57187809076dc325c7eba33006324513002b31`, with the full base...head diff and current surrounding contracts checked for interaction and regression context.

Verdict: two P2 gaps remain in the new remediations. Exact-head CI is green (39/39 checks).

Verified findings:
- P2: the credential-redaction projection still leaves supported secrets embedded in primitive strings and unrecognized secret fields, so the one-year cookie verifier remains credential-derived.
- P2: the hand-written Playwright extension checks still accept malformed selector arguments that the real Playwright parser rejects at execution.

Existing feedback: all current threads are resolved. The optional A2A password-client-field defect and block-style Open Interpreter whitespace defect are fixed. The two comments below target residual behavior in newly changed remediation code; they do not duplicate a live thread.

- Bind cross-tab clears to an opaque persisted revision, or construct an explicit allowlisted non-secret projection, instead of attempting to blacklist every credential representation.
- Reuse Playwright's selector parser/shared grammar rather than expanding regex checks that remove unparsed extension arguments before native CSS validation.
- The new Open Interpreter statement-template tests assert only the resulting sandbox mode; observing the rendered timeout, network, and git-check settings would make those regressions independently detectable.

QA:
- `source ~/.nvm/nvm.sh && nvm use && npm ci --ignore-scripts` — passed on a clean retry with Node 24.18.0/npm 11.16.0. The initial scripted install failed in local dependency setup; no lockfile drift resulted.
- `npm run test:app -- src/pages/redteam/setup/hooks/useRedTeamConfig.test.ts src/pages/redteam/setup/hooks/useRedTeamTargetConfigValidation.hydration.test.ts --run` — 415 tests passed.
- `npx vitest run test/providers/openinterpreter.test.ts --sequence.shuffle=false` — 59 tests passed.
- `npm run tsc` and `npm --prefix src/app run tsc` — passed.
- Real headless Playwright Chromium probe rejected `button:nth-match(div)`, `button:right-of(div:)`, and `button:text-matches("x", "z")` with parser/RegExp errors.
- Direct reproduction of the exact redaction/hash algorithm showed the nominal credential-free hash changes when only supported secret-bearing values change.
- `git diff --check fcde2e89a89dc4ca79dcc3012927f50193251759...e7a6e9d130a2f6597289ccccc9d0646ef5977ec8` — passed.
- Exact `/pr-review-toolkit:review-pr all parallel` invocation was unavailable (exit 127); its correctness, tests, docs, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialists.

Limitations: no credentialed provider call or full interactive browser workflow was run; browser validation used an isolated real Playwright runtime probe.

src/app/src/pages/redteam/setup/hooks/useRedTeamTargetConfigValidation.ts
```diff
@@ -0,0 +1,785 @@
...
+  const normalizedParent = parentKey?.toLowerCase().replace(/[-_]/g, '');
+  const redactAllValues =
+    normalizedParent !== undefined &&
+    ['auth', 'headers', 'signatureauth', 'tls', 'clienv', 'env'].includes(normalizedParent);
+  return Object.fromEntries(
+    Object.entries(value)
+      .sort(([left], [right]) => (left < right ? -1 : left > right ? 1 : 0))
+      .map(([key, child]) => [
+        key,
+        redactAllValues ? '[REDACTED]' : canonicalizeCredentialFreeTarget(child, key),
```
jameshiester-oai, line 70: [P2] Preserve behavior-affecting values in the clear fingerprint

This replaces every value below `auth`, `headers`, `tls`, `env`, and similar bags with the same literal, including non-secret fields such as `auth.type`, `auth.placement`, `tls.rejectUnauthorized`, `Content-Type`, and custom routing headers. Two same-shaped targets with different behavior therefore produce the same length/hash. I reproduced an identical SHA-256 for a valid header/TLS target and one with a newline-invalid header plus `rejectUnauthorized: false`. A delayed or persisted clear can then satisfy the target-binding comparison and reconcile/consume validation for the wrong target. Please preserve known non-secret discriminator/behavior fields, or bind the clear to an opaque exact-target revision rather than this lossy projection.

src/app/src/pages/redteam/setup/hooks/useRedTeamConfig.ts
```diff
@@ -12,6 +22,1379 @@ interface RecentlyUsedPlugins {
...
+      args.push(arg);
+      start = index + 1;
+    }
+  }
+  const last = value.slice(start).trim();
+  return quote || escaped || stack.length || !last ? null : [...args, last];
+};
+
+const unquotePlaywrightString = (value: string): string =>
+  value.slice(1, -1).replace(/\\(.)/gs, '$1');
```
jameshiester-oai, line 554: [P2] Decode Playwright CSS string escapes before validating the regex

Playwright parses these arguments as CSS strings, but this helper only removes one backslash. For `button:text-matches("\\5b", "")`, the hook validates regex `5b` and clears the recovery gate, while Playwright 1.61.1 CSS-decodes `\\5b` to `[` and then throws `Invalid regular expression ... Unterminated character class`. I reproduced both sides at this head. Please reuse Playwright's selector parser/tokenizer, or faithfully CSS-decode the quoted pattern and flags before constructing `RegExp`.

src/app/src/pages/redteam/setup/hooks/useRedTeamConfig.ts
```diff
@@ -12,6 +22,1379 @@ interface RecentlyUsedPlugins {
...
+
+const getPlaywrightSelectorArgumentEnd = (selector: string, start: number): number | null => {
+  let depth = 1;
+  let end = start;
+  for (; end < selector.length && depth > 0; end++) {
+    if (selector[end] === '(') {
+      depth++;
+    } else if (selector[end] === ')') {
```
jameshiester-oai, line 562: [P3] Do not treat escaped parentheses as argument delimiters

This scanner ignores CSS escaping, so an escaped `)` closes the layout pseudo early. The hook rejects `button:right-of(.foo\\))` as invalid, while Playwright 1.61.1 parses it successfully. This blocks recovery/save/run for a valid imported browser target. Handling this through Playwright's parser would also avoid maintaining a second delimiter grammar.

jameshiester-oai (commented):
Reviewed base `fcde2e89a89dc4ca79dcc3012927f50193251759` through exact head `72ac416380edceae7ba8cef331a868de9726c02c`.

Scope: incremental from authenticated completed head `e7a6e9d130a2f6597289ccccc9d0646ef5977ec8` (verified ancestor), with the full base...head PR, surrounding contracts, and existing feedback re-read for interaction and regression context.

Verdict: the new commit fixes the previously reported literal credential leakage and the known Playwright arity/nested-selector cases, but two P2 correctness gaps and one narrower P3 parser mismatch remain. Exact-head CI is green (39/39 reported checks), and GitHub reports the PR mergeable.

## Verified findings

### P2

- **Behavior-changing target values collapse to the same recovery fingerprint.** The new canonicalizer replaces every value below `auth`, `headers`, `tls`, `env`, and related bags with the same literal. A valid target and a behaviorally different or invalid target with the same keys therefore produce the same clear-marker length/hash. Direct exact-head probes produced identical SHA-256 values for targets that differed in `Content-Type`, `X-Mode`, and `tls.rejectUnauthorized`, including a newline-invalid header value. A delayed or persisted clear can consequently pass the target-binding check and consume/reconcile stale validation state for the wrong target. Preserve known non-secret discriminator/behavior values, or bind clears to an opaque exact-target revision instead of a lossy projection.
- **CSS hex escapes bypass `text-matches()` validation.** The local unquoter treats `\5b` as JavaScript-style escaping and validates the resulting regex `5b`; Playwright CSS-decodes the same selector string to `[` and throws an unterminated-character-class error. The hook accepted `button:text-matches("\\5b", "")` with no target-config error, while Playwright 1.61.1 rejected it at execution. Reuse Playwright's parser/tokenizer or implement CSS string decoding before constructing the validation regex.

### P3

- **Escaped closing parentheses in valid nested CSS are rejected.** The argument-end scanner counts an escaped `)` as the end of the pseudo-selector. The hook rejected `button:right-of(.foo\\))`, while Playwright 1.61.1 parsed it successfully. Account for CSS escapes, preferably through the same shared parser used to fix the P2.

## Existing feedback and test opportunities

All current inline threads are resolved. These comments cover residual/new behavior in the remediation and do not duplicate a live thread. The earlier optional OAuth-field issue and the previously demonstrated Playwright malformed-argument cases are fixed. Add regressions showing that credential-only changes retain a stable marker while auth discriminators, TLS behavior, and header semantics do not collide; add CSS-escaped regex and delimiter cases using the real Playwright grammar.

## Validation performed

- `source ~/.nvm/nvm.sh && nvm use && npm ci --ignore-scripts` — passed with Node 24.18.0/npm 11.16.0; no lockfile drift.
- `npm run test:app -- src/pages/redteam/setup/hooks/useRedTeamConfig.test.ts src/pages/redteam/setup/hooks/useRedTeamTargetConfigValidation.hydration.test.ts --run` — 2 files, 421/421 tests passed.
- `npm run tsc` and `npm --prefix src/app run tsc` — passed.
- `npm run build:app` — production app build passed.
- Targeted Biome and `git diff --check e7a6e9d130a2f6597289ccccc9d0646ef5977ec8..HEAD` — passed; Biome reported only five pre-existing complexity warnings and no errors.
- Real headless Playwright 1.61.1 probes confirmed the new documented positive selectors, rejected the CSS-escaped invalid regex, and accepted the escaped-parenthesis selector.
- Exact-head store/projection probes reproduced local acceptance of the escaped invalid regex, local rejection of the valid escaped-parenthesis selector, and identical marker projections/hashes across behavior-changing header/TLS values.

## Limitations

No credentialed provider request, destructive cache/database operation, or full manual browser workflow was run. Browser QA was limited to isolated real Playwright parsing/execution probes. The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable (exit 127); its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses were applied manually and reconciled with three independent read-only specialist reviews. No merge or rebase was performed.

PR #10115, opened 2026-07-15
fix(deps): align Renovate npm registry

mldangelo-oai:
## Summary
- set Renovate npmrc to the public npm registry host used by checked-in lockfiles
- preserve engine-strict during Renovate lockfile generation
- keep npm allow-remote protection intact while avoiding false remote classification for registry tarballs

## Why
Renovate PR #10101 fails artifact generation with EALLOWREMOTE for an ordinary registry.npmjs.org Tailwind optional tarball. Both repository lockfiles resolve only from registry.npmjs.org, but Mend hosted Renovate appears to generate lockfiles with a different configured registry host. npm 11.15 classifies that host mismatch as remote.

## Validation
- jq empty renovate.json
- npm run l
- npm run f
- git diff --check origin/main...HEAD
- attempted Renovate config validator; sandboxed run hit npm cache EPERM, escalated internal-registry retries hit stale ETARGET metadata, and the public-registry retry was denied by environment policy. The dedicated Validate Renovate Config CI job will provide the authoritative validator result.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,6 +8,7 @@
   "constraints": {
     "npm": ">= 11"
   },
+  "npmrc": "registry=https://registry.npmjs.org/\nengine-strict=true",
   "lockFileMaintenance": {
     "enabled": true,
     "schedule": [


jameshiester-oai (commented):
No actionable findings.

I reviewed the Renovate npmrc override, inherited configuration context, repository npm settings, lockfile registry hosts, engine-strict preservation, allow-remote implications, validation evidence, and formatting. The one-line change is scoped to the reported hosted-Renovate registry mismatch and matches the hosts already committed in both lockfiles. I did not find a regression, unintended behavior change, documentation gap, or simplification opportunity worth blocking this head.

Validation:
- jq empty renovate.json passed.
- git diff --check passed.
- The dedicated Validate Renovate Configuration check and the rest of the current GitHub checks are green.

No inline comments.

PR #7963, opened 2026-03-03
fix(providers): guard top_p for reasoning models, remove dead code, update UI defaults

yash2998chhabria:
## Summary

- **Guard `top_p` for reasoning models**: Exclude `top_p` from Responses API payload when `reasoning_effort` is set (prevents 400 errors on models like `gpt-5.3-chat-latest`)
- **Remove dead code**: Delete unused `ProviderSelector.tsx` (1,001 lines), `defaultProviders.ts` (699 lines), and dead `GET /api/providers` route
- **Update UI placeholders**: Update FoundationModelConfiguration default model from `gpt-4o` to `gpt-5.2` and placeholder examples to current models
- **Docs**: Align GPT-5.3 Instant documentation with evergreen style

## Test plan
- [x] All 108 responses provider tests pass
- [x] Webui tests pass with updated placeholder assertions
- [x] Full CI green (all platforms, all Node versions)
- [x] Verified `top_p` guard logic via live API testing

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10117, opened 2026-07-16
fix(providers): restore custom provider catalog

### LINKED ISSUE #10116: ui-providers.yaml no longer filters the eval-creator provider list (self-hosted, 0.121.19)
## Summary

The self-hosting docs state that `ui-providers.yaml` restricts the eval-creator provider list — *"Only configured providers shown (replaces default ~600 providers)"* — but in **0.121.19** (and current `main`) the file is loaded and validated server-side yet has **no effect on the provider list shown in the eval-creator UI**. The full built-in catalog is still displayed. The only remaining effect is toggling `hasCustomConfig`, which hides the "API keys" button.

## Environment

- Self-hosted `ghcr.io/promptfoo/promptfoo:0.121.19` (behavior also confirmed against current `main`)
- `PROMPTFOO_CONFIG_DIR` = default (`/home/promptfoo/.promptfoo`)

## Steps to reproduce

1. Place `ui-providers.yaml` at `/home/promptfoo/.promptfoo/ui-providers.yaml`:
   ```yaml
   providers:
     - openai:gpt-4o-mini
     - anthropic:messages:claude-sonnet-4-5-20250929
     - id: 'http://llm-gateway.internal/v1'
       label: Internal Gateway
       config:
         method: POST
         headers:
           Authorization: 'Bearer {{ env.GATEWAY_KEY }}'
   ```
2. Start the server.
3. Web UI → Create eval → Step 1 "Choose Providers" → "+ Add Provider" → "Select Provider Type".

## Expected (per docs)

Only the 3 configured providers appear; the default catalog is replaced.

## Actual

The full built-in catalog is shown (tabs: All 55 / My Application 8 / Agent Frameworks 14 / AI Providers 27 / Local Models 6). None of the configured providers — nor the labeled HTTP gateway — appear anywhere in the picker.

## Evidence the file IS loaded (so this is a UI/wiring gap, not a config error)

- Server log:
  ```
  Loaded server configuration
    "configPath": "/home/promptfoo/.promptfoo/ui-providers.yaml",
    "providerCount": 3
  ```
- `GET /api/providers/config-status` → `{"success":true,"data":{"hasCustomConfig":true}}`
- `GET /api/providers` → returns the SPA `index.html` (no such route is registered).

## Likely root cause (regression)

- #6124 (*feat(server): add server-side provider list customization*, merged 2025-11-24) added `ui-providers.yaml` **plus** a `GET /api/providers` list endpoint and eval-creator UI consumption ("Providers are properly sorted and grouped in the UI").
- The eval-creator was later rebuilt onto a static `ProviderTypeSelector` (`allProviderOptions`) that makes no API calls, orphaning #6124's list/UI code.
- #10046 (*chore: remove dead code and enforce Knip in CI*, merged 2026-07-13) removed the now-unused `GET /api/providers` list endpoint — shipped in 0.121.19 (2026-07-14).
- Net in 0.121.19 / `main`: `src/server/routes/providers.ts` registers only `GET /config-status`; `getAvailableProviders()` is used solely to compute `hasCustomConfig`; `ProviderTypeSelector` renders a static hardcoded list. So the documented filtering no longer exists.

## Ask

Either:
- **Restore** the provider-list filtering in the rebuilt eval-creator (re-consume `ui-providers.yaml`, e.g. via a `GET /api/providers` list the selector reads), or
- **Update the docs** (`site/docs/usage/self-hosting.md`) to reflect current behavior: `ui-providers.yaml` currently only loads server-side and toggles `hasCustomConfig` (hiding the "API keys" button). The docs also reference a "Reference Local Provider" button and a "~600 providers" list that don't match the current UI (an "API keys" button and ~55 provider *types*).

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

Gujiassh:
## Summary

- restore `GET /api/providers` as a runtime-validated and OpenAPI-documented catalog for `ui-providers.yaml`
- pass the exact configured provider entries into the rebuilt eval creator while preserving the built-in selector when no custom catalog exists
- keep provider creation disabled if the catalog cannot be loaded, and align the self-hosting guide with the current UI
- keep catalog-restricted edits inside the approved entry list instead of exposing editable provider identities or targets
- normalize catalog labels to render-safe strings at the server boundary

## Behavior

With `ui-providers.yaml`, the Add Provider dialog lists only configured entries. Selecting or replacing one saves its original `id`, `label`, and `config`, including custom HTTP URLs and templated headers. Entries with a valid string `id` but a non-string `label` fall back to their ID instead of breaking the dialog. Without a custom file, the existing provider-type picker, provider editor, and API keys control are unchanged.

Fixes #10116

## Testing

- `npm run test:app -- src/pages/eval-creator/components/AddProviderDialog.test.tsx src/pages/eval-creator/components/ProvidersListSection.test.tsx src/pages/eval-creator/components/EvaluateTestSuiteCreator.test.tsx --run` (73 tests)
- `npx vitest test/server/config/serverConfig.test.ts test/server/routes/providers.test.ts --run` (42 tests)
- runtime catalog-schema probe rejects object labels
- `npm run tsc`
- `npm --prefix src/app run tsc`
- `npm --prefix src/app run build`
- targeted Biome check (one pre-existing cognitive-complexity warning, no errors)
- `git diff --check`

`npm run openapi:check` cannot compare generated output because `site/static/openapi.json` is absent from both this branch and current upstream `main`; the OpenAPI source schema itself is type-checked and keeps the catalog label contract explicit.


mldangelo:
Reviewed and tested this end-to-end (server run with a custom `ui-providers.yaml`, catalog + config-status verified live) — the restore is legitimate and well-built; #7963 had orphaned the loader when the old ProviderSelector was deleted. I pushed `cb1e907` with two hardening fixes found during review: a non-string `id` in the yaml (e.g. `id: 123`) passed the loader's truthy check but failed the route's strict response parse, 500ing the catalog permanently — now warn-skipped with tests; and the edit-mode dialog wasn't given `availableProviders`, so Edit → Back reopened the unrestricted built-in selector — now forwarded with regression tests. Also deduplicated the response type (imported from the shared `@promptfoo/types/api/providers` you added), dropped a setupTests stub that already existed on main, and noted in the docs why credentials must stay out of the catalog file.

Gujiassh:
Addressed both catalog findings in `4bddca9ad`. Restricted edits now replace providers directly from the administrator catalog, so edited IDs/targets cannot escape it. Non-string labels are removed during server normalization, and both runtime and OpenAPI catalog schemas require string labels. Added regressions for restricted editing, normal unrestricted editing, and object-label fallback.

mldangelo (commented):
Thanks for sticking with this one — and heads up before you read further: **I pushed one commit to your branch**, `a2e5a02ae` (`fix(app): suppress provider catalog error toast after unmount`). It wraps the `showToast` call in the catch block at `EvaluateTestSuiteCreator.tsx:130-136` in the `isMounted` guard you already use on the success path at line 126, plus a regression test. I reproduced a toast firing after unmount, so navigating away mid-request surfaced "Failed to load provider configuration" on an unrelated page. `console.error` stays outside the guard so the failure is still diagnosable. Purely defensive, no behavior change while mounted — I pushed it rather than commenting because there was no judgment call in it. Everything else below is left for you.

## What works well

The restore itself is solid — #7963 orphaned the loader when the old ProviderSelector was deleted, and you rebuilt it correctly rather than patching around it. The `4bddca9ad` follow-up is the right call: replacing from the catalog instead of reopening the config editor genuinely closes the edit-flow escape, and you added regressions for restricted edit, unrestricted edit, and object-label fallback rather than just the happy path. Server side is clean — 111 tests across `serverConfig`/`providers`/`serverRouteSmoke` pass, and the whole `eval-creator` app suite (485 tests) is green.

## P2 — a transient catalog fetch failure is terminal

`hasCustomConfig` went from `useState(false)` to `useState<boolean | null>(null)` and the catch block no longer resets it. Since it feeds `isProviderCatalogReady={hasCustomConfig !== null}` (line 498) and `{hasCustomConfig === false && <ConfigureEnvButton />}` (line 261), one failed request leaves the UI stuck.

I un-mocked `ProvidersListSection` (your test file mocks it at line 35, so it only asserts on a stubbed `mock-provider-catalog-ready` span and never touches a real `disabled` button) and ran a failed fetch followed immediately by a healthy one:

```
BASELINE disabled states:        [ false ]
TOAST: Failed to load provider configuration: Failed to load providers from server
AFTER FAILURE disabled states:   [ true ]
AFTER FAILURE title/aria:        { title: null, ariaLabel: null, ariaDescribedBy: null }
MENTIONS retry/reload/unavailable? { retry: false, reload: false, unavailable: false }
AFTER rehydrate() disabled:      [ true ]
callApi total calls:             1     <- never refetches
```

Two questions rather than a prescription:

1. **Is fail-closed intentional here?** I think it should be — resetting to `false` would leave `availableProviders` null and render the full built-in catalog, letting someone bypass an admin `ui-providers.yaml` by inducing a fetch blip. So *not* a bug. But the empty dep array means there's no recovery path either, and the disabled buttons carry no tooltip or aria-label. Would a retry affordance, or just `title="Provider catalog unavailable — reload to retry"`, be acceptable? A tri-state (`'loading' | 'ready' | 'error'`) instead of overloading `null` would also make the three derived reads self-documenting.
2. **Was hiding `ConfigureEnvButton` on failure intended?** On main it rendered immediately (initial state `false`). Now `null === false` is false, so the API-keys button vanishes too — that looks like collateral rather than part of the restriction.

Whichever way you go, a test asserting recovery after failure→success would help; the current one only pins the failure state.

## P2 — `self-hosting.md` "Behavior Changes" is now incomplete

The `:::info` block (lines ~471-479) doesn't mention the two most consequential changes: with a catalog present, **Edit no longer opens the config editor** — it reopens the catalog and replaces the provider — so per-provider `config` (`temperature`, `max_tokens`, `headers`, `region`) becomes YAML-only. The "Configuration fields" section below still presents those as editable with no caveat. Worth adding a line, plus one noting Add Provider is disabled until the catalog loads.

## Simplifications (optional)

- **`ProvidersListSection.tsx:162` and `:182`** — two byte-identical Add Provider buttons in mutually exclusive branches. You had to paste `disabled={!isProviderCatalogReady}` into both, and your own test carries a comment about the second one. Hoisting it to a `const addProviderButton` above the return makes that class of miss impossible.
- **`EvaluateTestSuiteCreator.tsx:96-97`** — `hasCustomConfig` and `availableProviders` are two states set from one response that must agree. Storing the response once (`const [catalog, setCatalog] = useState<{providers, hasCustomConfig} | null>(null)`) gives identical behavior with one setter and no way to desync — and would fold naturally into whatever you decide on the tri-state above.
- **`/api/providers/config-status`** — now that the eval creator uses the new route, I grepped and found zero production consumers (only `src/openapi/server.ts:876`, the tests, and the `setupTests.ts:164` mock). Its doc comment claims "used by redteam setup UI" but nothing under `src/app/src/pages/redteam` references it. Worth deleting in this PR, or is something outside the repo polling it?

Nice work overall — the core of this is right, and the remaining items are mostly about what a user sees when the server hiccups.

Gujiassh:
Pulled and verified `a2e5a02` and `68d30a0`; please keep them.

Fail-closed is intentional. Retry recovers without exposing the unrestricted selector, and API keys stay hidden until a successful response proves no custom catalog exists. The docs now cover the load gate and catalog-based replacement.

I kept `/api/providers/config-status` because it is a published OpenAPI route; removing it would be an unrelated compatibility change.

Local verification passed 21 files/488 tests, TypeScript, app build, architecture, Biome, and Prettier. The title check failed only because GitHub's service was unavailable; I lack permission to rerun it.

Gujiassh:
Addressed the `defaultTest` catalog bypass in be2c8c7e7. With a custom catalog, provider or provider-option overrides in `defaultTest` now keep Run disabled; configurations without a custom catalog are unchanged. Added regressions for provider identity, provider settings, and the non-custom baseline. The focused suite passes 42/42 tests, along with app type-check and Biome lint.

Gujiassh:
Addressed the five catalog enforcement findings in 41929eed1f. The Run gate now remains fail-closed for scenario, prompt, environment-template, and external default-test overrides; it also stays closed until the current provider objects are reconciled with the canonical catalog. Non-provider `defaultTest.options` and configurations without a custom catalog remain supported. The focused suite passes 53/53 tests, along with app type-check and Biome.

PR #10120, opened 2026-07-16
fix(assertions): invert tool-call-f1 score under not-

mldangelo:
## Problem

`handleToolCallF1` flips `pass` for `not-tool-call-f1` but returns the raw F1 as `score`. A failed inverse assertion therefore still contributes its original high score to weighted/threshold aggregation (`AssertionsResult` sums `score*weight` independent of `pass`), and a passing inverse assertion reports a near-zero score. Verified empirically on main: inverse + perfect tool match returns `{ pass: false, score: 1 }`.

Two existing tests locked in the wrong behavior (`score: 1` on a failed inverse, `score: 0` on a passing one).

This is the same bug class fixed for `search-rubric` in #10103; `bleu`/`gleu`/`llm-rubric` all already invert (`gleu.ts`: `score: inverse ? 1 - score : score`).

## Fix

Mirror the `bleu`/`gleu` convention: `score: inverse ? 1 - f1 : f1`. No clamp needed — F1 is structurally within [0, 1] (harmonic mean of set-derived precision/recall behind a `precision + recall > 0` guard).

## Test plan

- Corrected the two inverse-test expectations (failed inverse → 0, passing inverse → 1).
- Added a mid-range case (F1=0.8, threshold 0.5, inverse → fails with score 0.2) so the inversion isn't coincidentally invisible at the 0/1 extremes.
- `npx vitest run test/assertions/toolCallF1.test.ts` — 34/34 pass; red/green verified against the unfixed handler.

### DIFF
--- a/src/assertions/toolCallF1.ts
+++ b/src/assertions/toolCallF1.ts
@@ -220,7 +220,9 @@ export const handleToolCallF1 = ({
 
   return {
     pass,
-    score: f1,
+    // F1 is structurally within [0, 1], so plain 1 - f1 matches the
+    // bleu/gleu inversion convention without clamping.
+    score: inverse ? 1 - f1 : f1,
     reason: pass
       ? `Tool Call F1: ${f1.toFixed(3)} (precision=${precision.toFixed(3)}, recall=${recall.toFixed(3)}). ` +
         `Expected: [${expectedList}], Called: [${actualList}]`
--- a/test/assertions/toolCallF1.test.ts
+++ b/test/assertions/toolCallF1.test.ts
@@ -410,9 +410,11 @@ describe('handleToolCallF1', () => {
 
       const result = handleToolCallF1(params);
 
-      // F1=1.0 would normally pass, but inverse makes it fail
+      // F1=1.0 would normally pass, but inverse makes it fail — and the
+      // inverted score (0) must not feed the original full score into
+      // weighted/threshold aggregation.
       expect(result.pass).toBe(false);
-      expect(result.score).toBe(1);
+      expect(result.score).toBe(0);
     });
 
     it('should pass inverse assertion when F1 is below threshold', () => {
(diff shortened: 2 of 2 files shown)


jameshiester-oai (commented):
Reviewed current base `688a90561180a6045ce3c8e8805d04bc23811bd7` through exact head `8c68c569bc3c9152a6f8c408c5db0bd7e7e14b0c`.

Scope: full review of this fresh one-commit PR, including the complete base...head patch, assertion aggregation callers, sibling inverse-score conventions, tests, and current feedback.

Verdict: no actionable defects found. Inverting the returned score to `1 - F1` aligns the negated assertion with unconditional weighted aggregation and with BLEU/GLEU/METEOR/ROUGE conventions. F1 is structurally bounded to [0, 1], so no clamp is needed. All exact-head CI checks are green.

- P0/P1/P2/P3: none.

Existing feedback: Copilot raised no inline comments; Codecov reports all changed coverable lines covered.

Simplification/test opportunities: no implementation simplification is warranted. A non-blocking follow-up could add a public-path `runAssertions` regression using the actual `not-tool-call-f1` type plus a test-level threshold, and the assertion docs could state explicitly that the negated score is `1 - F1`.

QA:
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/assertions/toolCallF1.test.ts test/assertions/assertionsResult.test.ts test/assertions/runAssertions.test.ts` — 3 files, 86 tests passed.
- `npm run tsc` — passed.
- `npm run build` — core/declarations/frontend production build passed.
- `npm run local -- eval -c /private/tmp/pr10120-promptfooconfig.yaml --no-cache -o /private/tmp/pr10120-results.json` — isolated echo-provider CLI path produced raw F1 0.8, component score 0.2, and the expected aggregate failure at threshold 0.5; exported JSON showed zero runtime errors.
- `git diff --check 688a90561180a6045ce3c8e8805d04bc23811bd7...8c68c569bc3c9152a6f8c408c5db0bd7e7e14b0c` — passed.
- Exact `/pr-review-toolkit:review-pr all parallel` invocation was unavailable (exit 127); its correctness, tests, docs, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialists.

Limitations: no external provider call was needed; the end-to-end path used the deterministic echo provider. A separate local dependency refresh used `npm ci --ignore-scripts` only to restore this disposable worktree after the initial missing-module failure.

PR #10124, opened 2026-07-16
fix(azure): avoid crash on empty completion choices

he-yufeng:
## Summary

`AzureCompletionProvider.callApi` read `data.choices[0].text` directly. An empty `choices` array — or a degenerate 200 response with no `choices` field — makes `data.choices[0]` (or `choice`) undefined, so `choice.text` threw a `TypeError` that the surrounding catch turned into an opaque `API response error: TypeError...` instead of a usable result.

Interestingly the same function already optional-chains `choice?.finish_reason` and `choice?.content_filter_results`, so only `data.choices[0]` and `choice.text` were left unguarded.

This optional-chains both so those responses take the same graceful no-output path (empty output, or the content-filter fallback) that the chat path already takes since #9867. Empty/missing choices now return `output: ''` with no error.

## Test

Added two regression tests in `test/providers/azure/completion.test.ts`: an empty `choices: []` array and a response with no `choices` field both return graceful empty output instead of an error. The full azure completion suite passes (13 tests) locally.

### DIFF
--- a/src/providers/azure/completion.ts
+++ b/src/providers/azure/completion.ts
@@ -92,13 +92,16 @@ export class AzureCompletionProvider extends AzureGenericProvider {
         };
       }
 
-      const choice = data.choices[0];
+      // Optional-chain choices so an empty array or a degenerate 200 response
+      // with no `choices` takes the graceful no-output path instead of throwing,
+      // matching the chat path (#9867).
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
Thank you!!!

PR #10125, opened 2026-07-16
fix(providers): align OpenAI models and capabilities

mldangelo-oai:
## Summary

Bring Promptfoo’s OpenAI providers, documentation, and examples in line with current API behavior and model availability. This fixes incorrect routing and billing, adds missing multimodal support, and prevents credentials or tenant-specific results from leaking through shared caches.

- **Models and routing:** update the model catalog, pricing, cached-input and Batch rates, and lifecycle guidance. Correctly separate Responses-only, search, and Codex-only models; reject public-API use of `gpt-5.3-codex-spark` while preserving compatible custom gateways and identifier-based routes.
- **Multimodal capabilities:** add and harden speech, transcription, image, video, and web-search support. Cover current Sora request shapes, resolutions, durations, references, and characters; PCM playback; VAD transcription; citations; lifecycle headers; retries; and usage billing.
- **Caching and isolation:** safely handle authenticated gateways, credential-bearing references, project and tenant boundaries, streaming cancellation, background-response recovery, concurrent subscribers, and single billing across processes.
- **Docs and examples:** correct OpenAI provider syntax, multimodal configuration, supported audio and Deep Research options, model/deprecation guidance, legacy video-remix guidance, and the bundled transcription example.

## Validation

- PR CI: **39/39 checks passed**
- Expanded Node 20/22/24/26 compatibility matrix across Linux, macOS, and Windows: **41/41 jobs passed**
- Focused provider/cache regressions: **2,754 passed, 9 skipped**
- Follow-up CI regressions: **400 passed**
- App tests: **54 passed**
- TypeScript, production build, docs build, Biome, Prettier, architecture checks, Knip, and dependency checks: **passed**
- Local HTTP repros covered authenticated and keyless gateways, cache isolation and redaction, retries and cancellation, background recovery and single billing, structured-output ordering, search routing and citations, transcription, PCM, image caching, and current Sora creation/billing behavior.
- A live four-model OpenAI smoke was attempted, but the locally configured key returned **401 Unauthorized**. No live API success is claimed.

## Audit sources

- [OpenAI API pricing](https://developers.openai.com/api/docs/pricing)
- [OpenAI deprecations](https://developers.openai.com/api/docs/deprecations)
- [Speech-to-text](https://developers.openai.com/api/docs/guides/speech-to-text)
- [Text-to-speech](https://developers.openai.com/api/docs/guides/text-to-speech)
- [Video generation](https://developers.openai.com/api/docs/guides/video-generation)
- [Deep Research](https://developers.openai.com/api/docs/guides/deep-research)
- [Background mode](https://developers.openai.com/api/docs/guides/background)
- [Web search](https://developers.openai.com/api/docs/guides/tools-web-search)
- [Structured Outputs](https://developers.openai.com/api/docs/guides/structured-outputs)
- [Codex rate card](https://help.openai.com/en/articles/20001106-codex-rate-card-2)


mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

src/providers/openai/video.ts
```diff
@@ -349,37 +507,104 @@ export class OpenAiVideoProvider extends OpenAiGenericProvider {
...
+      /(?:^|[-_])(?:project|tenant|account)(?:[-_]|$)/i.test(key),
+    );
+    const hasProjectScopedAsset =
+      Boolean(config.characters?.length) ||
+      Boolean(
+        config.input_reference &&
+          typeof config.input_reference === 'object' &&
+          'file_id' in config.input_reference,
+      );
+    const canCacheVideo =
```
jameshiester-oai, line 585: [P1] Scope cached videos to the authenticated OpenAI account. For the default `api.openai.com` endpoint, `usesAuthenticatedCustomEndpoint` is false while `Authorization` is intentionally removed from `safeCacheHeaders`, so two API keys with the same prompt/model/size/duration produce the same persistent key. I reproduced this with two provider instances and isolated storage: account A generated and stored a video (three API requests), then account B made zero requests and received account A's storage reference with `cached: true`. That can disclose private generated media across accounts and charge the wrong tenant. Please disable caching for authenticated requests unless a stable non-secret account/project discriminator is included in the key, or otherwise make the cache storage account-scoped, and add a two-credential regression test.

src/providers/openai/transcription.ts
```diff
@@ -147,6 +192,7 @@ export class OpenAiTranscriptionProvider extends OpenAiGenericProvider {
           getRequestTimeoutMs(),
           'json',
           context?.bustCache ?? context?.debug,
+          config.maxRetries,
```
jameshiester-oai, line 229: [P2] Propagate the caller's abort signal before enabling retries here. `callApi` accepts `_callApiOptions`, but this request never passes `_callApiOptions.signal` to `fetchWithCache`. Against a delayed loopback endpoint, aborting the supplied signal after 25 ms still allowed the transcription call to resolve successfully after 288 ms; with `maxRetries`, cancelled evaluations can also keep retrying and consuming resources. Pass the signal into the fetch options and add a test that cancellation terminates the request and prevents further attempts.

jameshiester-oai (commented):
Reviewed base `688a90561180a6045ce3c8e8805d04bc23811bd7` through head `f8ab2142e32ce3af35d810fe28c8e309acc0854c` (full review; no authenticated completion marker existed for this head).

Verdict: two actionable defects found. Exact-head CI is green (39/39 checks), but the local probes below expose an account-isolation issue in video caching and cancellation that is not propagated by transcription requests.

## Verified findings

### P1
- OpenAI video cache entries are shared across API credentials on the default endpoint. The inline comment includes the triggering condition, observed result, and correction direction.

### P2
- Transcription requests ignore the caller's abort signal; the newly configurable retry path can continue work after the evaluation has been cancelled. See the inline comment for the reproduction.

## Existing feedback
Existing unresolved threads already cover separate concerns around TTS cache scoping, relative/mutable video input files, and background Responses cancellation/coalescing. I did not duplicate those comments. I also checked the Groq reasoning-model thread: the subclass still overrides `isReasoningModel()`, so the cited base-class model-name check does not reproduce that concern on this head.

The credential-aware caching rules are repeated across media providers and are difficult to keep consistent. A shared cache-scope policy with cross-credential regression tests would reduce this class of drift. Transcription tests should also assert that aborting `CallApiOptionsParams.signal` terminates the in-flight request and prevents retries.

## Validation performed
- `npx vitest run test/cache.test.ts test/fetch.test.ts test/providers.test.ts test/providers/openai test/providers/webSearchUtils.test.ts test/util/sanitizer.test.ts --sequence.shuffle=false` — 42 files passed, 1 skipped; 2,030 tests passed, 9 skipped.
- `npm run tsc && npm run build && git diff --check 688a90561180a6045ce3c8e8805d04bc23811bd7...HEAD` — passed, including core, declarations, and frontend builds.
- `SKIP_OG_GENERATION=true npm --prefix site run build` — production documentation build passed.
- `npm run local -- eval -c /private/tmp/pr10125-tts.yaml --no-cache -o /private/tmp/pr10125-tts-result.json` against a loopback OpenAI-compatible stub — exit 0; 1/1 passed, score 1, WAV output present, and the request used `response_format: wav`.
- Cross-account video cache probe with isolated temporary cache/config roots — account A made three stubbed API requests; account B made zero and received account A's identical storage reference with `cached: true`.
- Transcription cancellation probe — caller signal aborted at 25 ms, but the delayed request still resolved successfully after 288 ms.

## Limitations
The installed PR Review Toolkit command `/pr-review-toolkit:review-pr all parallel` was unavailable (command not found), so I completed the six requested review lenses manually and reconciled three independent specialist reports. No paid OpenAI API calls or browser UI flows were run; media and transcription behavior was exercised against deterministic loopback stubs.

mldangelo:
@codex review

PR #10128, opened 2026-07-16
feat(providers): add current Azure, Claude, and Gemini models

mldangelo-oai:
## Summary

- add current Azure GPT-5.6/5.5 Pro/5.2 Pro, GPT Image 2, and audio/realtime pricing, including supported long-context/priority tiers, cached text/audio/image with deterministic modality fallback, discounted cached-input rates for dated GPT-5 Chat/Codex deployments, image-token and image-output costs, Phi-4 multimodal audio, current TTS rates, and multi-response tool-call billing; add a GA `azure:realtime:` route with API-key/Entra auth, prompt/conversation/tenant context isolation, prompt-level endpoint/key overrides, proxy-safe URLs, scoped bearer suppression, cleanup, and explicit rejection of unsupported transcription/translation-only deployments; preserve discounted cached-input usage for Foundry agents
- register the current Claude Opus IDs (canonical dateless IDs only — the Models API 404s on dated 4.6+/4.7 snapshots, verified live); add Gemini Live v1beta/multimodal/cache/tool-call usage handling with the documented camelCase WebSocket wire shape for 3.1 and current 2.5 native-audio snapshots, explicit protocol overrides, preserved v1alpha tool-response compatibility, top-level/nested voice configuration, correctly billed and paced multi-frame video with turn completion and timeout allowance, correct still-image/mixed-input ordering and tool-response wire shape with primitive-result normalization, inclusive-thinking and finite-audio handling, raw-PCM validation, omission of unsupported 3.1 Live dialog options, media-safe debug logs, standard AI Studio/Vertex modality-aware/priority/Flex and cached-audio billing with typed generic/modality-specific override precedence, current resolved-model latest-alias, TTS, and native-audio pricing, WAV normalization/default generation config, and merged/sanitized passthrough tools including snake-case and single-object declarations when tool calling is disabled
- add the Gemini Omni Interactions API/video-output path for both AI Studio and Vertex with correct endpoint/OAuth/quota routing, bounded in-progress polling, safe generation-config/service-tier filtering and nested-field normalization, normalized safety enums, stateless chat steps/OpenAI-style/native-Gemini image/audio/video parts and envelopes, modality-usage billing, single/array multimodal input, latest-turn video and trailing-text selection, step-error propagation, authenticated URI/`gs://` download with a strict HTTPS storage-redirect allowlist, clear errors for unsupported Vertex follow-ups and Omni tools/MCP, custom IDs, proxy-safe URLs, and credential-safe cache bypass; update provider docs and add routing, cost, lifecycle, auth, usage, and multimodal regressions

## Validation

- affected provider suites: **16 files, 1,221 tests passed**, including cache-rate and absent-modality-breakdown fallback for mixed text/audio/image, dated GPT-5 Chat/Codex cached-input rates, Foundry cached-input usage reporting, generic/modality-specific cost overrides, image billing/input normalization, Gemini latest-alias priority/Flex/cache/cached-audio and published TTS/native-audio/tool-use billing, merged and disabled array/object/snake-case passthrough-tool handling, v1beta Live wire-shape/current native-audio routing/explicit overrides/v1alpha tool compatibility/top-level voice/mixed input/log-redaction/WAV-chunk assembly/frame accounting/stalled-stream timeout/post-tool bookkeeping/primitive tool-result normalization/PCM validation, AI Studio/Vertex Omni routing/revision/filtering/safety-settings/stateless-step/native-Gemini/envelope/image/audio normalization/nested generation config/modality billing/authenticated URI and `gs://` download/SSRF-safe redirect/latest-turn and trailing-text selection/quota-header/polling/step-error and unsupported-capability/MCP handling, Azure API-key header isolation, OpenAI gateway bearer preservation, typed/empty conversation-ID isolation, and realtime usage-total/cleanup/unsupported-route regressions
- pinned-catalog billing comparison: the newly supported Azure and direct Google-model rates match across standard/cache/priority/long-context/modality cases; the three moving Gemini `latest` aliases, two active Gemini TTS models, and the December native-audio snapshot intentionally use currently published provider prices because those catalog rows are stale, and realtime image-token/cached-image rates were also verified against published provider pricing
- Biome, Prettier, and `git diff --check`: clean (existing cognitive-complexity warnings only)
- `npm run tsc -- --pretty false`: only the confirmed pre-existing `test/providers/claude-agent-sdk.test.ts(2350,15)` TS2322 on `origin/main`
- prepublish review: three independent high-risk passes plus verifier

Catalog pinned to BerriAI/litellm `229159c7908ba99bfb1165e52ad9e588b1ccea45` (SHA-256 `33cedefa26df0ea38c526cd8939dd420b69ce68efd0e78f091871781ccac3bde`).

## Review fixes (post-review pass)

- **Anthropic**: removed `claude-opus-4-7-20260416` / `claude-opus-4-6-20260205` from `ANTHROPIC_MODELS`, shorthand routing, and docs — both 404 on the live Models API (Anthropic stopped publishing dated snapshots with the 4.6 family; the litellm rows are wrong). Azure AI Foundry's dated Claude deployment names are unaffected.
- **Azure pricing**: corrected catalog drift where dated snapshots contradicted their own aliases — `gpt-4o-mini-audio-preview-2024-12-17` (was 16x text / 4x audio), cached-input discounts on `gpt-realtime-2025-08-28` / `gpt-realtime-1.5-2026-02-23` / the `gpt-realtime-mini` alias, `gpt-audio-*` dated-snapshot audio rates, and `gpt-4o-mini-tts` text input.
- **Azure error paths**: `azure:realtime` logs (rather than swallows) provider-auth init failures when a prompt-level key is present; responses/foundry-agent report unknown cost as `undefined` instead of `$0`.
- **Google Interactions**: HTTP status is now checked on the create/poll calls (gateway failures no longer surface as "did not return video output"); video *input* tokens bill at the image rate instead of the text rate; off-allowlist video URIs log a warning instead of silently skipping the download.
- **Google Live**: don't attach a bare 44-byte WAV artifact when a native-audio stream returns only a transcription (no PCM bytes); the transcript is still surfaced as output text.
- **Google Live**: the response timeout is an idle guard re-armed on every server message — long streaming generations are no longer cut off at `timeoutMs` and discarded, while stalled streams still fail fast with the effective deadline in the error; configured-`TEXT`-to-`AUDIO` modality rewrites now warn.
- **Live-verified**: Gemini `generateContent` accepts `serviceTier` (priority tier machinery is valid); `gemini-omni-flash-preview`, both TTS models, the 12-2025 native-audio snapshot, and `gemini-3.1-flash-live-preview` all exist in the live AI Studio catalog.
- **CI**: `vi.resetAllMocks()` in `azure/realtime.test.ts` (test-hygiene gate); hermetic Vertex endpoint tests (gcloud env vars no longer flip the global-endpoint assertions); regenerated the architecture edge baseline on the tree merged with current main; added regression tests for poll timeout/terminal status/gateway errors, unknown-deployment cost, unconfigured endpoint, and Live idle-timeout keep-alive.

## End-to-end verification (live)

Ran the actual CLI (`tsx src/localEntrypoint.ts eval`) against live providers:

- `anthropic:messages:claude-opus-4-7`: output correct, cost `$0.000265` = 23×$5/1M + 6×$25/1M (exact — Anthropic cost path unaffected by the ID cleanup).
- `google:gemini-2.5-flash`: cost `$0.0000049` = 8×$0.30/1M + 1×$2.50/1M (exact — `calculateGoogleCost` video-input edits don't perturb normal text billing).
- `google:live:gemini-2.5-flash-native-audio-preview-12-2025`: real websocket stream completed, the new TEXT→AUDIO modality warning fired, idle-timeout keep-alive did not cut off the stream, audio cost billed (`$0.000224`).
- Route resolution: `google:`/`vertex:gemini-omni-flash-preview` → `GoogleInteractionsProvider`, `azure:realtime:<deployment>` → `AzureRealtimeProvider`, and the transcription-only/translation-only deployments reject with their tailored errors.


test/providers/azure/realtime.test.ts
```diff
@@ -0,0 +1,600 @@
...
+  }),
+}));
+
+vi.mock('../../../src/providers/providerRegistry', () => ({
+  providerRegistry: { register: mockRegister, unregister: mockUnregister },
+}));
+
+describe('AzureRealtimeProvider', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
```
jameshiester-oai: [P1] Reset persistent mock implementations, not just call history. This file installs a module-scope `mockImplementation` and then assigns persistent `mockResolvedValue` behavior, while `vi.clearAllMocks()` leaves those implementations in place. The repository's test-hygiene guard names this file, and all Linux Node lanes, macOS, and Windows shard 3 fail on the exact head. Use full/per-mock resets in `beforeEach`, then reinstall the constructor and `mockCallApi` defaults required by each test so random order cannot inherit behavior.

src/providers/google/interactions.ts
```diff
@@ -0,0 +1,656 @@
...
+
+  id(): string {
+    return this.providerId || `google:${this.modelName}`;
+  }
+
+  toString(): string {
+    return `[Google Interactions Provider ${this.modelName}]`;
+  }
+
+  async callApi(prompt: string, context?: CallApiContextParams): Promise<ProviderResponse> {
```
jameshiester-oai, line 264: [P2] Accept and propagate `CallApiOptionsParams.abortSignal` for this long-running provider. The evaluator passes request-scoped cancellation through the third argument, but this signature drops it, so the initial POST, polling sleeps/GETs, video download, and storage continue after cancellation. I passed an already-aborted signal to this exact head; the provider still made a POST and returned video output. That can keep paid video work and polling alive until the configured/default timeout after an eval is stopped. Thread the signal through every request, make the poll wait abortable (or check before each wait/request), preserve abort semantics, and add an abort-during-poll regression test.

src/providers/google/interactions.ts
```diff
@@ -0,0 +1,656 @@
...
+            : {}),
+        })
+          .filter(([field]) => !unsupportedGenerationFields.has(field))
+          .map(([field, value]) => [
+            field === 'topP' ? 'top_p' : field === 'maxOutputTokens' ? 'max_output_tokens' : field,
+            value,
+          ]),
+      ),
+    };
+    const passthrough = Object.fromEntries(
```
jameshiester-oai, line 393: [P2] Filter capability-controlled aliases from raw passthrough before building the Vertex request. The provider rejects `config.previousInteractionId` for Vertex at line 266, but `passthrough: { previous_interaction_id: 'x' }` survives this filter and is spread into the body. A request-capture probe confirmed the unsupported field was sent and the call proceeded instead of failing locally. This defeats the explicit follow-up guard and turns a clear configuration error into a remote validation/failure path. Reserve both camel- and snake-case forms of guarded fields (including other unsupported capabilities), and add the passthrough form to the existing Vertex rejection test.

jameshiester-oai (commented):
Reviewed base `51fa1d62a71104314c03dc0c9ae2909b117f5da2` through head `c56245b7fe8d804206f58df6e4712533660e6f2a` (full review; no authenticated completion marker existed for this head).

Verdict: four actionable findings, including one P1 test-isolation failure. Exact-head CI is red: Style Check and all six affected Node test lanes fail deterministically. The remaining builds, provider suites, docs, integration, smoke, UI, and security lanes are green. The PR title follows the applicable provider scope convention.

## Verified findings

### P1
- The Azure realtime test installs persistent module-scope mock behavior but only clears call history. The repository's enforced isolation guard fails on every affected Node lane. See the inline comment.

### P2
- The new provider files widen three prohibited architecture edges: `src/providers/google/interactions.ts:1-12` and `src/providers/azure/realtime.ts:1-16` contribute to `providers -> legacy-contracts` increasing from 271 to 275 imports, `providers -> legacy-runtime` from 317 to 321, and `providers -> node` from 278 to 283. `npm run architecture:check` exits 1. Remove or narrow the imports, or update and justify the baseline if these directions are intentional.
- Google Omni ignores evaluator cancellation during creation, polling, download, and storage. See the inline reproduction.
- Raw passthrough can bypass the explicit Vertex follow-up rejection and send `previous_interaction_id` anyway. See the inline reproduction.

## Existing feedback
Five unresolved threads already cover separate issues: Azure realtime translation routing, regional Vertex Interactions routing, unsupported Interactions safety settings, single-object passthrough tool merging, and missing-usage cost reporting. I did not duplicate them. The single-object tools concern reproduces in the request builders; the new comments below address distinct paths.

The Google request builders repeat passthrough normalization and capability filtering, which has already produced inconsistent object/array and guarded-field behavior. Centralizing that normalization would make unsupported fields fail consistently. Add abort-focused polling coverage, a Vertex snake-case passthrough regression, and restore the architecture and mock-isolation gates before merge.

## Validation performed
- `npx vitest run test/providers/anthropic/util.test.ts test/providers/azure/chat.test.ts test/providers/azure/completion.test.ts test/providers/azure/foundry-agent.test.ts test/providers/azure/realtime.test.ts test/providers/azure/responses.test.ts test/providers/azure/util.test.ts test/providers/google/ai.studio.test.ts test/providers/google/interactions.test.ts test/providers/google/live.test.ts test/providers/google/provider.test.ts test/providers/google/util.test.ts test/providers/google/vertex.test.ts test/providers/openai/index.test.ts test/providers/openai/realtime.test.ts test/providers/registry.test.ts --sequence.shuffle=false` — 16 files and 1,221 tests passed.
- `npx vitest run test/test-hygiene.test.ts -t 'keeps new root tests from adding unreset module-scope persistent mocks' --sequence.shuffle=false` — failed deterministically and named `providers/azure/realtime.test.ts`. Exact-head CI reports both persistent-mock guards across six Node lanes.
- `npm run architecture:check` — failed with the three widened edges listed above.
- `npm run tsc -- --pretty false`, `npm run build`, production app build, `SKIP_OG_GENERATION=true npm --prefix site run build`, and `git diff --check 51fa1d62a71104314c03dc0c9ae2909b117f5da2...HEAD` — passed.
- `npm run local -- eval -c /private/tmp/pr10128-interactions.yaml --no-cache -o /private/tmp/pr10128-interactions-result.json` against a loopback Interactions endpoint with remote generation disabled — exit 0; 1/1 passed, score 1, no errors, video output present, and token usage was 2 input / 3 output / 5 total.
- Already-aborted Omni probe — the provider still made one POST and returned video output.
- Vertex passthrough probe — `passthrough.previous_interaction_id` was sent and returned output instead of triggering the documented local rejection.

## Limitations
The installed PR Review Toolkit command `/pr-review-toolkit:review-pr all parallel` was unavailable (command not found), so I completed the six requested review lenses manually and reconciled three independent specialist reports. No paid Azure, Anthropic, or Google calls, live WebSocket session, or browser flow was run; HTTP behavior used deterministic loopback stubs.

mldangelo:
@codex review

Re-requesting a review — since the last codex pass this branch added fixes for the phantom Claude Opus 4.6/4.7 snapshot IDs (404 on the live Models API), several Azure catalog pricing corrections, the Gemini Live idle-timeout keep-alive, interactions HTTP-status/error-path hardening, an empty-WAV guard, doc corrections, and behavior-preserving ternary simplifications. Please take another look.

PR #10133, opened 2026-07-17
fix(redteam): preserve generation provider selection

jameshiester-oai:
## Summary

- Centralize red-team provider selection into one runtime object that records whether the winner was explicit, cached, fallback, or the built-in default.
- Keep Web UI test-case previews request-scoped: a form-selected provider wins, then the org/team cache, then the built-in default. Preview lookup explicitly ignores stale process-global CLI config from earlier runs.
- Carry the selected provider through plugin generation, synthesis, local strategy fallback, and nested layers instead of independently resolving provider state again.
- Keep math-prompt and multilingual remote only for the true built-in default. Explicit and cached enterprise providers stay local because the remote task API cannot reproduce those selections.
- Preserve specialized local behavior: MathPrompt keeps its small JSON-only default provider and multilingual keeps its configured JSON provider.
- Keep live provider instances and provider-option objects local. Generated agentic configs may persist only plain provider ID strings and never overwrite a step-local provider override.
- Send minimal remote payloads for math, multilingual, Likert, and citation while retaining targetId where cloud uses it to resolve team context.
- Preserve preview provider env overrides and treat blank or incomplete provider values as unset.

## Why

The original jailbreak:meta issue was a precedence bug: its explicit task provider could be replaced by a cached provider. Fixing only that lookup was not enough because later phases could still reconstruct provider state from cache or stale CLI config.

This change makes one provider selection follow the generation request end-to-end. It also separates runtime provider behavior from serializable data so resolved credentials cannot leave through remote payloads, debug request logs, or generated YAML.

## Test plan

- npm run tsc
- npx vitest run test/redteam/providers/shared.test.ts test/redteam/strategies/types.test.ts test/redteam/strategies/mathPrompt.test.ts test/redteam/strategies/multilingual.test.ts test/redteam/strategies/likert.test.ts test/redteam/strategies/citation.test.ts test/redteam/strategies/layer.test.ts test/redteam/index.test.ts test/server/routes/redteam.test.ts --sequence.concurrent=false (316 passed)
- npm run architecture:check
- npm run lint:ci
- git diff --check


mldangelo-oai (changes requested):
Release audit on 469dc4bbd7f96438dc25c577c82d70772fa9819c: the focused provider suite passes (63 tests), but this is not release-ready. Moving the explicit-provider return ahead of the cache makes the Web UI generate-test route's hard-coded REDTEAM_MODEL bypass the cached org/team provider, which can route enterprise/on-prem generation to the wrong provider. Please distinguish a user-supplied override from the hard-coded fallback (or omit that fallback at the caller) and add the route-level regression before landing.

jameshiester-oai:
> Release audit on [469dc4b](https://github.com/promptfoo/promptfoo/commit/469dc4bbd7f96438dc25c577c82d70772fa9819c): the focused provider suite passes (63 tests), but this is not release-ready. Moving the explicit-provider return ahead of the cache makes the Web UI generate-test route's hard-coded REDTEAM_MODEL bypass the cached org/team provider, which can route enterprise/on-prem generation to the wrong provider. Please distinguish a user-supplied override from the hard-coded fallback (or omit that fallback at the caller) and add the route-level regression before landing.

This has now been fixed

mldangelo-oai:
## Critical follow-up review

Reviewed the current head `76f181a76175563d2894afb428d70f06b9fdf78c`. The original hard-coded-default/cache issue is addressed, and CI is green (43/43), but I would still hold this PR: the new precedence and preview-contract changes leave four concrete correctness gaps. The first three correspond to the current unresolved threads; the fourth is an additional contract mismatch found while tracing the full path.

### Findings

1. **P2 — Keep one provider throughout plugin and strategy generation.** ([existing thread](https://github.com/promptfoo/promptfoo/pull/10133#discussion_r3615183573))

   The preview route resolves the form-selected provider for the plugin, then invokes the strategy with only `strategy.config` ([route](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/server/routes/redteam.ts#L91-L129)). Local [`math-prompt`](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/redteam/strategies/mathPrompt.ts#L106-L114) and [`multilingual`](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/redteam/strategies/multilingual.ts#L348-L360) subsequently perform a new cache/global lookup, so the second phase can use a different backend.

   This is broader than preview. Full synthesis resolves the explicit `provider` for plugin generation ([resolution](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/redteam/index.ts#L1093-L1106)), but strategy configuration is rebuilt from `cliState.config?.redteam?.provider` ([strategy call](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/redteam/index.ts#L695-L708)). With a cached org provider and a caller/CLI override, a direct local `synthesize(..., strategies: [{ id: 'math-prompt' }])` reproduction on this head produced:

   ```text
   full-synthesis-provider-flow= cli-or-form-selected-provider -> cached-org-provider
   generated-test-count= 2
   ```

   Please carry the resolved/request-scoped provider into local strategy generation (including remote-to-local fallback), instead of having strategies independently re-resolve global state. Add an integration-level regression covering cache present + explicit provider + `math-prompt`, and the equivalent multilingual fallback.

2. **P2 — Preserve provider environment overrides during request validation.** ([existing thread](https://github.com/promptfoo/promptfoo/pull/10133#discussion_r3615183577))

   [`PreviewGenerationProviderSchema`](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/types/api/redteam.ts#L16-L21) extends `ProviderOptionsSchema`, whose `env` schema is a closed `z.object`. The contract explicitly documents that unknown environment keys are stripped and callers needing them must retain the source object ([contract](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/contracts/env.ts#L146-L152)); this handler correctly uses parsed data, so the keys are irretrievably lost before loading the provider.

   Direct reproduction with `MY_MODEL`, `MY_CUSTOM_KEY`, `ANTHROPIC_TEMPERATURE`, `AWS_ACCESS_KEY_ID`, `GOOGLE_CLOUD_PROJECT`, and `OPENAI_API_KEY` yielded:

   ```text
   env-provider.runtime-valid= true
   env-provider.keys-after-parse= OPENAI_API_KEY
   ```

   This affects custom templates/API-key variables and built-in consumers of unlisted overrides, while the new OpenAPI schema advertises `env` as an arbitrary string record. Override `env` for this wire schema with a validated string record (or an equivalent catchall) and add an object-provider route test that asserts the exact object reaching the loader retains custom keys.

3. **P2 — Normalize unset/blank providers before preview validation and loading.** ([existing thread](https://github.com/promptfoo/promptfoo/pull/10133#discussion_r3615183581))

   The public red-team config accepts an empty provider, but the new preview schema rejects it before fallback. Incomplete object providers are also accepted by the public schema and can be present while editing/importing a configuration. Direct validation on this head:

   ```text
   empty-string:    public=true preview=false
   empty-object-id: public=true preview=false
   missing-object-id: public=true preview=false
   ```

   Conversely, whitespace-only strings pass `min(1)` and then fail provider loading instead of falling back:

   ```text
   whitespace(" ").schema-valid= true
   load-error= Could not identify provider:  .
   ```

   Normalize empty/whitespace and incomplete values to `undefined` at the preview boundary (or omit them in the client) so the cache/default path remains usable, and cover empty string, whitespace, and incomplete-object cases. This preserves compatibility with configurations accepted by [`RedteamConfigSchema`](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/validators/redteam.ts#L263-L282).

4. **P2 — Make the new OpenAPI `provider.inputs` contract match runtime validation.**

   The handwritten OpenAPI mirror accepts any non-empty injection-placement string for every input type ([mirror](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/openapi/server.ts#L30-L47)), but the canonical runtime schema enforces type-specific placements with `superRefine` ([canonical schema](https://github.com/promptfoo/promptfoo/blob/76f181a76175563d2894afb428d70f06b9fdf78c/src/contracts/shared.ts#L68-L99)). For example, a PDF input with `injectionPlacements: ['comment']` validates against the generated OpenAPI request and is rejected by the route:

   ```text
   invalid-pdf-placement.openapi-valid= true
   invalid-pdf-placement.runtime-valid= false
   runtime-error= Invalid PDF injection placements: comment. Expected one of: body, header, footer
   ```

   The new comment says the documented request shape is equivalent; it is not. Encode the type/placement relationship in the wire schema (for example, discriminated branches using the existing placement constants) or document the limitation, and add a representative OpenAPI/runtime parity test.

### Simplification opportunities

- **Define one JSON-safe provider wire schema and reuse it.** The PR now maintains parallel provider shapes in `src/types/api/redteam.ts` and `src/openapi/server.ts`; `src/types/api/providers.ts` already has the same required-ID refinement for provider routes. Centralizing a serializable `ProviderOptionsWithId`/preview schema (with a string-record `env` and shared input definitions) would remove the duplicated field lists, eliminate the cast in the route, and prevent the drift demonstrated above. Keep function-valued transforms confined to the in-process schema.
- **Replace the ten positional parameters to `callTestGenerationApi` with a typed options object.** Adding `provider` forces the single-turn caller to insert a positional `1` solely to reach the new final parameter. An options object makes `count`, `history`, turn information, and provider explicit, removes placeholder arguments, and makes future additions safer.
- **Keep provider resolution precedence in one place and table-test it.** The manager now repeats load/wrap logic across explicit, cached, request-fallback, configured, and `defaultTest` paths. Resolving a single candidate after the cache check, then loading/wrapping once, would make the intended order (`explicit > cache > request fallback > configured/defaultTest/default`) easier to audit. A small table covering explicit/cache/fallback combinations plus local strategy execution would be clearer and more robust than isolated mock-only assertions.

### Verification

- Focused provider, meta, synthesis, strategy, OpenAPI, and route suites were exercised; the route suite was rerun in isolation after a combined run briefly exhausted localhost ports (`EADDRNOTAVAIL`) and passed **38/38**. The frontend generation suite passed **16/16**.
- `npm run tsc` passed.
- `git diff --check 5469ee52d2ea5387dfe62c080698c0e2e2b9b6af...76f181a76175563d2894afb428d70f06b9fdf78c` was clean.

The core explicit-first fix for `jailbreak:meta` is directionally correct; the remaining work is to make provider selection consistent across all generation phases and keep the new wire contract backward-compatible and truthful.

src/redteam/strategies/layer.ts
```diff
@@ -158,10 +159,13 @@ export async function addLayerTestCases(
...
     );
 
-    const next = await stepAction(applicable, injectVar, {
+    const stepConfig = {
       ...(stepObj.config || {}),
       ...(config || {}),
-    });
+    };
+    const next = runtimeContext
+      ? await stepAction(applicable, injectVar, stepConfig, undefined, runtimeContext)
```
mldangelo-oai, line 177: **[P2] Preserve the selected provider for layered agentic strategies.** Passing `runtimeContext` only helps the regular-step path. When `steps` contains an attack provider, the earlier branch returns a generated provider whose `config` is built from `stepObj.config` and target/layer metadata, so the outer `config.redteamProvider` is dropped. I reproduced `steps: ['jailbreak:meta', 'audio']` with `redteamProvider: 'openai:chat:selected-provider'`: the generated `promptfoo:redteam:iterative:meta` config had `redteamProvider: undefined`. At eval time the iterative provider consequently calls the manager without the selected provider and can fall back to a stale/cached backend—the precedence bug this PR is intended to fix. Please copy the safe declarative selection into the attack-provider config (and add a layer→agentic regression test).

src/redteam/strategies/mathPrompt.ts
```diff
@@ -107,14 +108,19 @@ export async function encodeMathPrompt(
...
-    jsonOnly: true,
-    preferSmallModel: true,
-  });
-  const redteamProvider = wrapGenerationProvider
-    ? wrapGenerationProvider(loadedProvider)
-    : loadedProvider;
+  const loadedProvider =
+    generationProvider ||
+    (await redteamProviderManager.getProvider({
+      jsonOnly: true,
```
mldangelo-oai: **[P2] Keep the small JSON-specialized fallback for implicit local generation.** `synthesize()` always supplies `runtimeContext.generationProvider`, including when no generation provider was configured, so this branch now bypasses the previous `getProvider({ jsonOnly: true, preferSmallModel: true })` path for every ordinary local MathPrompt run. At this head I verified the difference: synthesis resolves `openai:gpt-5.5-2026-04-23` with no `response_format`, whereas the previous strategy fallback resolves `openai:gpt-5.4-mini-2026-03-17` with `{ type: 'json_object' }`. This both raises generation cost and removes the JSON-output constraint that the parser relies on. Please distinguish an implicit default from an explicitly selected runtime provider and retain the specialized fallback for the former.

src/redteam/strategies/multilingual.ts
```diff
@@ -349,18 +350,24 @@ async function translateBatchCore(
   text: string,
   languages: string[],
   wrapGenerationProvider?: (provider: ApiProvider) => ApiProvider,
+  generationProvider?: ApiProvider,
 ): Promise<Record<string, string>> {
   // Prefer a preconfigured multilingual provider if available (set by the server at boot).
-  const cachedMultilingual = await redteamProviderManager.getMultilingualProvider();
+  const cachedMultilingual = generationProvider
```
mldangelo-oai: **[P2] Preserve the preconfigured multilingual provider when the generation provider is implicit.** Because `synthesize()` always supplies a `generationProvider`, this condition prevents `getMultilingualProvider()` from ever being consulted during synthesized local multilingual generation, even when the server has preconfigured a JSON-capable multilingual provider. I reproduced one translation with a registered specialized provider: it received zero calls and the generic runtime provider received one. The preceding comment and previous behavior explicitly prefer that specialized provider. Please bypass it only for a genuinely explicit/request-scoped selection (or return selection provenance from the manager), and add coverage for the implicit-default case.

src/redteam/providers/shared.ts
```diff
@@ -149,10 +151,57 @@ class RedteamProviderManager {
...
+   */
+  getProviderSpec({
+    provider,
+    fallbackProvider,
+  }: {
+    provider?: RedteamFileConfig['provider'];
+    fallbackProvider?: RedteamFileConfig['provider'];
+  } = {}): LoadableRedteamProvider | undefined {
+    const toSpec = (value: RedteamFileConfig['provider'] | undefined) =>
+      typeof value === 'string' || isProviderOptions(value) ? value : undefined;
```
mldangelo-oai, line 214: **[P2] Check that provider-option specs are actually serializable.** `isProviderOptions()` only checks for a string `id`; it does not establish JSON/YAML safety. For example, `{ id: 'openai:chat:test-model', transform: (output) => output.trim() }` is returned here as a spec, but both `JSON.stringify` and the config writer's `yaml.dump(..., { skipInvalid: true })` silently drop `transform`. A generated Hydra/iterative case then reloads with different provider behavior; non-JSON values inside `config` (such as `BigInt` or cycles) can also make remote serialization throw. Please narrow/validate the serializable shape once, treat unsupported options as runtime-only or fail clearly, and add a JSON/YAML round-trip regression test.

mldangelo-oai (changes requested):
Thanks, James — the runtime/spec split and request-scoped preview validation are meaningful improvements. I re-reviewed exact head `0cf7b6286e`, confirmed the earlier custom-env, blank/incomplete-provider, and typed-input-placement cases are addressed, and ran the focused backend/frontend suites plus TypeScript, architecture, and formatting checks (509 tests passing; CI is 44/44).

I still found several correctness/security gaps, so I’m requesting changes:

- **P1: provider-option credentials still leave the process.** I independently reproduced [the existing thread](https://github.com/promptfoo/promptfoo/pull/10133#discussion_r3616217173) with canaries in `env`, `config.apiKey`, and `config.headers.Authorization`: all three reached a loopback remote endpoint through both `math-prompt` and `jailbreak:likert`. The arbitrary `env` value also appeared verbatim in the debug API-request log. Agentic strategies can persist the same option object into generated YAML. Please keep secret-bearing specs local or reject/redact them before serialization/logging; a live-provider/spec split alone is insufficient.
- **P2: implicit defaults are incorrectly treated as opaque runtime providers**, confirming [the existing thread](https://github.com/promptfoo/promptfoo/pull/10133#discussion_r3616217180): with remote generation enabled and no configured provider, MathPrompt made zero remote requests and instead called the local provider. The same path affects multilingual.
- **P2: cache replacement is non-atomic**, confirming [the existing thread](https://github.com/promptfoo/promptfoo/pull/10133#discussion_r3616217188). I also exercised a failure on the *second* load: the regular cache then points at the replacement, the JSON-only cache still points at the old provider, and the spec points at the replacement. Load both instances into locals and publish the complete selection only after both succeed.

I left four additional inline findings for layered agentic propagation, specialized local defaults, the multilingual cache, and non-JSON-safe provider options.

For simplification, a single `resolveProviderSelection()` returning something like `{ provider, spec, source }` would remove the duplicated precedence walks in `getProvider()`/`getProviderSpec()`, eliminate their consistency/race hazards, and make “explicit/cached/runtime/default” routing unambiguous for strategies. A narrow, genuinely JSON-safe spec type can then be checked once. Two smaller cleanups: make the ten-argument preview helper take an options object, and reuse the exported input-placement constants instead of duplicating their enum values.

Verification at the reviewed head: 493 backend tests + 16 frontend tests passed, `tsc` passed, architecture check passed, formatting passed (existing complexity warnings only), and `git diff --check` was clean. The PR is behind current `main`, although the intervening main changes do not overlap these files.

mldangelo-oai (approved):

PR #10136, opened 2026-07-17
fix(cache): expire persistent background billing claims

mldangelo:
## Summary

`claimCacheKeyOnce` (introduced in #10125 for OpenAI background-mode single-billing) writes a durable 0-byte marker file per background response under `cache/claims/` so separate eval processes cannot both attribute the same response's usage. Those files were only ever removed by a full cache clear, so long-running or high-volume background evals grew the `claims/` directory without bound — even though the response cache entries the claims guard expire after the cache TTL (14 days by default).

This prunes claim files older than the cache TTL when a new claim is created, throttled to at most once per hour per process so high-volume claiming does not turn each call into an O(directory) scan.

**Why pruning is billing-safe:** a claim only needs to outlive the cached response it guards. Once that response has expired on the same TTL horizon, the claim can no longer prevent a double-bill, so removing it cannot affect billing correctness. Claims younger than the TTL are never touched. When the TTL is disabled (`PROMPTFOO_CACHE_TTL=0`), claims are left in place too, matching the non-expiring cache.

This aligns claim lifetime with the cache-entry lifetime, which is exactly the remedy the automated review on #10125 suggested ("clean claims alongside expired cache entries").

## Stacked on #10125

The claim mechanism this fixes exists only on #10125's branch, so this PR is based on `mdangelo/codex/openai-model-audit-20260715` and its diff is just the cache change. It should merge after #10125 (GitHub will retarget the base to `main` once #10125 lands).

## Validation

- `test/cache.test.ts`: 75 passed, including 4 new tests — prunes claims older than the TTL, leaves fresh claims, throttles sweeps to once per interval, and skips pruning entirely when the TTL is disabled.
- Real-filesystem check (not mocked): a 30-day-old claim file is pruned while a fresh claim and the newly created claim are preserved.
- `tsc`, Biome, and Prettier: clean.


src/cache.ts
```diff
@@ -564,6 +564,44 @@ function getInflightFetchCacheKey(cacheKey: string, url: RequestInfo, options: R
...
+  let entries: string[];
+  try {
+    entries = fs.readdirSync(claimsPath);
+  } catch {
+    return;
+  }
+  for (const entry of entries) {
+    const claimFile = path.join(claimsPath, entry);
+    try {
+      if (now - fs.statSync(claimFile).mtimeMs > maxAgeMs) {
```
jameshiester-oai, line 596: [P1] Do not expire claims while another process can still be polling the same response. With a short `PROMPTFOO_CACHE_TTL`, two processes can read the same cached queued response and poll independently. Process A can finish and create this claim while process B remains in flight past the TTL; when B finishes, its first per-process sweep removes A's marker and the subsequent `openSync(..., 'wx')` succeeds, so both calls return `true` and the same response is attributed twice. The stat-then-remove sequence also has an ABA race: after this `statSync`, another process can replace the pathname with a fresh claim and this `rmSync` deletes that replacement. Cache expiry does not cancel active pollers. Reclamation needs a cross-process lease/atomic protocol and retention that covers active polling, with a multiprocess regression proving exactly one claimant.

src/cache.ts
```diff
@@ -564,6 +564,44 @@ function getInflightFetchCacheKey(cacheKey: string, url: RequestInfo, options: R
...
+    return;
+  }
+  lastClaimSweepMs = now;
+  const maxAgeMs = getCacheTtlMs();
+  if (maxAgeMs <= 0) {
+    return;
+  }
+  let entries: string[];
+  try {
+    entries = fs.readdirSync(claimsPath);
```
jameshiester-oai, line 589: [P2] The hourly throttle is process-local, so every fresh CLI/eval process starts with `lastClaimSweepMs = 0` and synchronously walks the entire directory on its first claim. With P concurrent workers and N historical claims this becomes P × N blocking `statSync`/`rmSync` work on the request path—most painfully on rollout when this directory is already large. Use a cross-process sweep lease and bounded or asynchronous cleanup batches so one process performs limited work for the fleet.

jameshiester-oai (commented):
Reviewed base `e17ce87b380d7c5e807a622d86c859206e1e1ffb` through head `9d1cfb33a655cb322d4086966bef78c569e6b1a7`.

Scope: fresh full review of the two-file base...head diff, its OpenAI background-response consumer, current-main interaction, all current feedback, and exact-head CI.

Verdict: the cleanup bounds disk growth, but it does not preserve the cross-process single-billing invariant under expiry and can synchronously amplify filesystem work across a process fleet. Exact-head CI is fully green.

Verified findings:
- P1: claim expiry can delete a live claim while another process is still polling the same background response; the stat/remove sequence can also unlink a freshly recreated claim. In both cases two processes can successfully claim and attribute the same response.
- P2: the throttle is process-local, so every fresh CLI/eval process synchronously walks the entire claims directory on its first claim, creating O(processes × claims) blocking filesystem work.

Existing feedback:
- Four open Copilot threads cover explicit test TTLs and unsafe mock casts. They are narrower test-maintenance suggestions and were not duplicated.
- The committed tests remain single-process and mock every filesystem operation; they do not cover the billing race or fleet-level sweep behavior.

Test and simplification opportunities:
- Add a barrier-controlled multiprocess test that proves exactly one claimant across expiry/reclamation.
- Add a clock-controlled test at the one-hour boundary; the current throttle test only proves two immediate calls scan once.
- A cross-process sweep lease plus bounded/asynchronous batches would address both the correctness and operability problems more directly than a process-local timestamp.

Validation:
- `npm ci --ignore-scripts` — passed.
- `npx vitest run test/cache.test.ts test/providers/openai/responses/request.test.ts --sequence.shuffle=false` — 2 files, 158 tests passed.
- The same suites with `--sequence.seed=10136` — 158 tests passed.
- `npm run tsc` — passed.
- `npm run build` — full core/declaration/Web UI production build passed.
- `git diff --check e17ce87b380d7c5e807a622d86c859206e1e1ffb...HEAD` — passed.
- Deterministic real-filesystem ABA probe — after B observed an old inode, A removed it and created a fresh claim; B's stale remove deleted A's fresh inode and B then acquired `wx` successfully.

Limitations: the requested `/pr-review-toolkit:review-pr all parallel` command is unavailable on this host (exit 127), so its six lenses were applied manually and reconciled with three read-only specialist reviews. No live OpenAI credentialed request, intentionally long background poll, production-sized claim directory, full local repository test suite, cache/database clearing, or source modification was performed.

PR #10138, opened 2026-07-17
fix(redteam): exclude other-encodings strategies from coding-agent canaries

orangeCatDeveloper:
## Problem

`CANARY_BREAKING_STRATEGY_IDS` (`src/redteam/constants/plugins.ts`) lists the encoding strategies that mangle prompt text and are excluded from coding-agent plugins to protect deterministic canary/receipt matching:

```
base64, hex, homoglyph, leetspeak, rot13, multilingual, math-prompt, jailbreak:composite
```

It omits the four **other-encodings** strategies — `morse`, `piglatin`, `camelcase`, `emoji` — which transform the payload text exactly like the listed ones (the sibling `ENCODING_STRATEGIES` set and the `other-encodings` collection both include them).

## Fix

- Add `morse`, `piglatin`, `camelcase`, `emoji` to `CANARY_BREAKING_STRATEGY_IDS`.
- Add a guard test asserting every `STRATEGY_COLLECTION_MAPPINGS['other-encodings']` member is canary-breaking, so a future encoding can't be forgotten.

## Testing

`npx vitest run test/redteam/plugins/index.test.ts` — 27 passed. Verified the guard test fails without the fix. Existing canary tests spread `...CANARY_BREAKING_STRATEGY_IDS`, so they pass unchanged.


mldangelo-oai (approved):
Release audit on d973b5af2e9c53cb11093fc1570a12bf3bf8bb7a: ran the focused canary/strategy coverage (27 tests) and reviewed the encoding exclusions. The added canary-breaking strategies are correctly excluded without changing unrelated strategy behavior; no release-blocking findings.

mldangelo (commented):
Thanks for this — tight, well-targeted fix. One heads-up first so you aren't surprised by a commit you didn't write.

## Pushed to your branch

I pushed `4bb80c1` on top of your work rather than leaving you a list of chores:

- **`site/docs/red-team/plugins/coding-agent.md:158`** enumerated `CANARY_BREAKING_STRATEGY_IDS` verbatim, so it went stale the moment your four IDs landed. It now references the `other-encodings` collection instead of its members, so it stays correct as `STRATEGY_COLLECTION_MAPPINGS` grows. `src/redteam/AGENTS.md` requires `site/docs/red-team/` pages to track strategy behavior, and that line is the only place this exclusion is documented anywhere.
- Dropped `jailbreak:composite` from the recommended YAML in that same section — it was recommended at line 155 and declared excluded at line 158. Pre-existing contradiction, cheap to sweep while in the paragraph.
- Added one behavioral test next to yours. **Your test is untouched.**

Revert or amend any of it freely.

## What works

You picked the right lever. `other-encodings` is expanded into its four members at `src/redteam/index.ts:1039-1050` *before* `pluginMatchesStrategyTargets` is consulted at `:659`, so adding member IDs is what actually bites — adding the collection ID would have been a silent no-op. Confirmed end-to-end through real `synthesize()` with `coding-agent:core` + `strategies: [other-encodings]`:

```
PR head:      10 test cases, uniqueStrategyIds = []
fix reverted: 50 test cases, uniqueStrategyIds = [camelcase, morse, piglatin, emoji]
```

And your test isn't vacuous — deleting the 4 added lines fails it on `expected [...] to include 'camelcase'`.

## One question (nit)

`src/redteam/constants/plugins.ts:450-453` hand-copies `STRATEGY_COLLECTION_MAPPINGS['other-encodings']` (`src/redteam/constants/strategies.ts:113`), and your new test then polices the copy. Was that deliberate — keeping the exclusion list explicitly readable in one place? The alternative:

```ts
...STRATEGY_COLLECTION_MAPPINGS['other-encodings'],
```

makes the invariant structural: a fifth encoding gets excluded automatically, no second edit site. `strategies.ts` has zero imports so there's no cycle, and the constant is only ever spread (`src/redteam/plugins/index.ts:156`) with no `typeof ...[number]` usage anywhere, so widening off the literal tuple is inert. Your `toContain` test would become tautological and could go. Genuinely fine either way — your version is more greppable.

## Optional follow-up (not this PR)

The pre-run estimate still promises tests it won't produce:

```
[info] Using strategies:
camelcase (2 additional tests)     -> later: | camelcase | 0 | 0 | Skipped |
```

Pre-existing behavior, but four more strategies now reach it. Might deserve its own issue.

orangeCatDeveloper:
@mldangelo-oai Friendly ping — the branch has been synced with the latest `main` and CI is re-running. Would appreciate a review when you get a chance. Thanks!

PR #10139, opened 2026-07-18
fix(providers): preserve Responses schema strict false

VectorPeak:
Preserve explicit `response_format.json_schema.strict: false` when building OpenAI and Azure Responses API request bodies.

## What Problem This Solves

OpenAI Responses and Azure Responses both accepted `response_format` objects that include `json_schema.strict`, but their request builders always emitted `text.format.strict: true` for JSON Schema formats.

That meant an explicit user configuration like `strict: false` was indistinguishable from an omitted value by the time promptfoo sent the request. This matters because provider APIs treat strict mode as a real schema-enforcement switch, and strict mode only supports a subset of JSON Schema.

## Change

The Responses request builders now preserve the configured strict value using the same fallback shape already used by the Azure Foundry Agent provider:

```ts
const strict = responseFormat.json_schema?.strict ?? responseFormat.strict ?? true;
```

The default remains `true` when no strict value is provided. The change only affects OpenAI Responses and Azure Responses JSON Schema request body construction.

## Evidence

Before this change, both request builders hardcoded `strict: true`, so `strict: false` could not reach `text.format`.

Added focused regression coverage for both provider paths:

- OpenAI Responses preserves `body.text.format.strict === false`
- Azure Responses preserves `body.text.format.strict === false`

## Possible call chain / impact

```text
promptfoo provider config
  -> response_format: { type: json_schema, json_schema: { strict: false, ... } }
  -> maybeLoadResponseFormatFromExternalFile(...)
  -> OpenAiResponsesProvider.getOpenAiBody(...) / AzureResponsesProvider.getAzureResponsesBody(...)
  -> request text.format.strict
```

Sibling surface checked: Azure Foundry Agent already preserves `strict: false`; this PR aligns the Responses providers with that behavior without changing `json_object`, plain text, tools, or schema file loading.

## Validation

- `npm ci`
- `npx @biomejs/biome check --write src/providers/openai/responses.ts src/providers/azure/responses.ts test/providers/openai/responses/responseFormat.test.ts test/providers/azure/responses.test.ts`
- `npx vitest run test/providers/openai/responses/responseFormat.test.ts test/providers/azure/responses.test.ts` — 2 files, 50 tests passed
- `npx vitest run test/providers/openai/responses test/providers/azure/responses.test.ts` — 15 files, 237 tests passed
- `git diff --check`
- `npm run tsc`

### DIFF
--- a/site/docs/providers/openai.md
+++ b/site/docs/providers/openai.md
@@ -1541,9 +1541,18 @@ The Responses API uses the flattened format:
     },
     "required": ["event_name", "date", "location"],
     "additionalProperties": false
-  }
+  },
+  "strict": true
 }
 ```
+
+`strict` is optional and defaults to `true`. Set it to `false` when your schema uses
+JSON Schema keywords that [strict mode does not support](https://platform.openai.com/docs/guides/structured-outputs#supported-schemas)
+(for example `minimum`/`maximum`, `pattern`, or omitting `additionalProperties: false`).
+An explicit `false` is passed through to the API rather than being forced to `true`. The
+nested `json_schema.strict` form is also honored and takes precedence over a top-level
+`strict`. This applies to the `openai:responses`, `azure:responses`, and `xai:responses`
+providers.
 
 You can also use nested file references for the schema itself, which is useful for sharing schemas across multiple response formats:
 
--- a/src/providers/azure/responses.ts
+++ b/src/providers/azure/responses.ts
@@ -184,13 +184,14 @@ export class AzureResponsesProvider extends AzureGenericProvider {
         const schema = responseFormat.schema || responseFormat.json_schema?.schema;
         const schemaName =
           responseFormat.json_schema?.name || responseFormat.name || 'response_schema';
+        const strict = responseFormat.json_schema?.strict ?? responseFormat.strict ?? true;
 
         textFormat = {
           format: {
             type: 'json_schema',
             name: schemaName,
             schema,
-            strict: true,
+            strict,
           },
         };
       } else {
--- a/src/providers/openai/responses.ts
+++ b/src/providers/openai/responses.ts
@@ -939,13 +939,14 @@ export class OpenAiResponsesProvider extends OpenAiGenericProvider {
         const schema = responseFormat.schema || responseFormat.json_schema?.schema;
         const schemaName =
           responseFormat.json_schema?.name || responseFormat.name || 'response_schema';
+        const strict = responseFormat.json_schema?.strict ?? responseFormat.strict ?? true;
 
         textFormat = {
           format: {
             type: 'json_schema',
             name: schemaName,
             schema,
-            strict: true,
+            strict,
           },
         };
       } else {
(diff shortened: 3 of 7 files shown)


mldangelo-oai (approved):
Release audit on 6d7ebb4b9143f3ef8246005666bcead4300a988b: ran the focused Responses/schema suites (50 tests) and reviewed request construction. Explicit strict false is preserved while unset/true behavior remains intact; no release-blocking findings.

mldangelo (commented):
Thanks for this — tightly scoped fix, and it's correct. Verified it's safe to land.

## Heads-up: I pushed a commit to your branch

So you're not surprised by a commit you didn't write: I added two tests to `test/providers/openai/responses/responseFormat.test.ts`. **Test-only, no production code touched.** The reason was that mutation testing showed the OpenAI half of your change wasn't pinned by anything:

- Delete `?? true` from `src/providers/openai/responses.ts:942` → `Test Files 34 passed (34) / Tests 1197 passed (1197)`. **Survived.**
- Change it to `?? false` → same, 1197 passed. **Survived.**
- The *identical* `?? false` mutation on the Azure line (`src/providers/azure/responses.ts:187`) → `Tests 1 failed | 534 passed`, failing at `test/providers/azure/responses.test.ts:157`:
  ```diff
  -   "strict": true,
  +   "strict": false,
  ```
  **Killed.**

Your Azure test pinned the default; nothing pinned it on OpenAI. Separately, deleting the middle rung `?? responseFormat.strict` from *both* files left all 1732 tests green — that shape was unexercised everywhere. The pushed tests cover both rungs, and each one fails against its corresponding mutant.

## What's good

The three-rung resolution `json_schema.strict ?? strict ?? true` is the right shape, and it matches what `src/providers/azure/foundry-agent.ts:396` already does (from #8059) — so this aligns the Responses providers with an existing precedent instead of inventing a fourth convention. Backward compat genuinely holds: a config with no `strict` key anywhere still emits `text.format.strict: true`. And you caught *both* clones, which is the easy thing to miss here.

## Remaining items

### 1. `xai:responses` still hardcodes `strict: true` — P2

`src/providers/xai/responses.ts:124`, inside `buildTextFormat`, is the same block reading the same config shape:

```ts
const schema = responseFormat.schema || responseFormat.json_schema?.schema;
const schemaName = responseFormat.json_schema?.name || responseFormat.name || 'response_schema';
return { format: { type: 'json_schema', name: schemaName, schema, strict: true } };
```

After this PR, `strict: false` is honored on `openai:responses` and `azure:responses` and silently ignored on `xai:responses` — same docs page, same config shape, divergent behavior. Was leaving xAI out deliberate (e.g. their API rejects non-strict `json_schema`)? If so, a note in `site/docs/providers/xai.md` would save someone a debugging session. If not, it's the same one-liner.

### 2. Docs — P3

`site/docs/providers/openai.md:1529-1547` shows the flattened Responses `response_format` with no `strict` key at all, while the Chat Completions example directly above (line 1524) does show `"strict": true`. Nothing in `site/docs` states the Responses default or that it's now configurable.

### 3. Optional: one helper instead of four copies

`buildTextFormat` at `src/providers/xai/responses.ts:107` is already the extracted form of the block you edited. Lifting it into `src/providers/responses/` (all three files already import from there) drops ~50 duplicated lines and fixes xAI for free. Totally fine as a follow-up.

Nit if you're in there anyway: the two tests you added are near-verbatim copies of the tests immediately above them — `it.each([{ strict: true }, { strict: false }])` would avoid a 7th copy of the `mockApiResponse` blob.

Nothing blocking. Thanks again.

PR #10142, opened 2026-07-19
fix(assertions): default conversation-relevance threshold to 0.5

### LINKED ISSUE #10141: bug(assertions): conversation-relevance defaults to zero and silently always passes
## Describe the bug

`conversation-relevance` defaults an omitted assertion threshold to `0`:

```ts
const threshold = assertion.threshold || 0;
```

The final score is a proportion in `[0, 1]`, and pass/fail is calculated as `score >= threshold - Number.EPSILON`. As a result, an assertion without an explicit threshold can never fail, even when every conversation window is judged irrelevant and the score is `0`.

This also differs from the DeepEval metric that this implementation is adapted from. DeepEval's `ConversationRelevancyMetric` uses a default threshold of `0.5`.

## Steps to reproduce

1. Configure a `conversation-relevance` assertion without `threshold`.
2. Use a conversation where the grader marks every response as irrelevant.
3. Observe a score of `0` with `pass: true`.

A focused test against current `main` reproduces the same behavior by returning an irrelevant verdict for a single conversation window: the result has `score === 0` but `pass === true`.

## Expected behavior

Omitting `threshold` should use DeepEval's `0.5` default, so a score below `0.5` fails. An explicit `threshold: 0` should remain valid and preserve the current opt-in always-pass boundary.

## Additional context

- Current implementation: `src/external/assertions/deepeval.ts`
- DeepEval reference: https://github.com/confident-ai/deepeval/blob/48c18b4c3b3cbaacfa2c61bd7b69c57d5f55b769/deepeval/metrics/conversation_relevancy/conversation_relevancy.py#L26-L38
- Searches for `conversation-relevance`, `conversation relevancy`, default thresholds, and always-pass behavior found no existing Promptfoo issue or PR for this assertion.

Gujiassh:
## Summary

Fixes #10141.

`conversation-relevance` currently treats an omitted threshold as `0`. Because its score is a proportion in `[0, 1]`, even an entirely irrelevant conversation with a score of `0` passes.

This change uses the `0.5` default from DeepEval's `ConversationRelevancyMetric`. It uses nullish coalescing so an explicit `threshold: 0` remains valid, and documents both behaviors.

The regression test failed on current `main` with `score === 0` and `pass === true`. With the fix, an omitted threshold fails at that score while the explicit-zero inverse case continues to pass.

## Verification

- `npx vitest run test/external/assertions.test.ts --sequence.shuffle=false` (9 tests)
- `npx vitest run test/external/assertions.test.ts --sequence.seed=10141` (9 tests)
- `npx vitest run test/external --sequence.shuffle=false` (24 tests)
- `npm run tsc`
- `npm run build` (library and 3,297-module app build)
- Targeted Biome/Prettier formatting and `git diff --check`

### DIFF
--- a/examples/eval-conversation-relevance/README.md
+++ b/examples/eval-conversation-relevance/README.md
@@ -52,7 +52,7 @@ Shows a high-quality technical support conversation with a high relevance thresh
 
 ## Configuration Options
 
-- `threshold`: Minimum score required to pass (0-1)
+- `threshold`: Minimum score required to pass (0-1, default: 0.5)
 - `config.windowSize`: Number of messages in each sliding window (default: 5)
 - `provider`: Override the default grading model
 
--- a/site/docs/configuration/expected-outputs/model-graded/conversation-relevance.md
+++ b/site/docs/configuration/expected-outputs/model-graded/conversation-relevance.md
@@ -23,6 +23,8 @@ assert:
     threshold: 0.8
 ```
 
+The threshold defaults to `0.5` when omitted. Set it explicitly to `0` to accept any score.
+
 ## Using with conversations
 
 The assertion works with the special `_conversation` variable that contains an array of input/output pairs:
--- a/src/external/assertions/deepeval.ts
+++ b/src/external/assertions/deepeval.ts
@@ -16,6 +16,7 @@ const DEFAULT_WINDOW_SIZE = 5;
 
 export const handleConversationRelevance = async ({
   assertion,
+  inverse,
   outputString,
   prompt,
   providerCallContext,
@@ -74,7 +75,7 @@ export const handleConversationRelevance = async ({
   }
 
   const score = totalWindows > 0 ? relevantCount / totalWindows : 0;
-  const pass = score >= threshold - Number.EPSILON;
+  const pass = score >= threshold - Number.EPSILON !== inverse;
 
   // Generate a comprehensive reason if there are irrelevancies
   let reason: string;
@@ -125,7 +126,7 @@ export const handleConversationRelevance = async ({
   return {
     assertion,
     pass,
-    score,
+    score: inverse ? 1 - score : score,
     reason,
     tokensUsed: tokensUsed.total > 0 ? tokensUsed : undefined,
   };
--- a/src/external/assertions/deepeval.ts
+++ b/src/external/assertions/deepeval.ts
@@ -1,6 +1,7 @@
 // These assertions are ported from DeepEval.
 // https://docs.confident-ai.com/docs/metrics-conversation-relevancy. See APACHE_LICENSE for license.
 
+import { isGraderFailure } from '../../matchers/llmGrading';
 import { callProviderWithContext, getAndCheckProvider } from '../../matchers/providers';
 import { getDefaultProviders } from '../../providers/defaults';
 import invariant from '../../util/invariant';
@@ -56,6 +57,10 @@ export const handleConversationRelevance = async ({
       test.options,
       providerCallContext,
     );
+
+    if (isGraderFailure(result)) {
+      return { ...result, assertion };
+    }
 
     if (result.pass) {
       relevantCount++;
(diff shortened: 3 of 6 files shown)


mldangelo-oai (approved):
Release audit on dde11274cbac745f1b281591c4a34a5677c98457: synced to current main, fixed the negated conversation-relevance threshold/score path, and reran focused assertion coverage (10 tests). CI is green at the exact head; no release-blocking findings.

mldangelo (commented):
## Heads-up: I pushed one commit to your branch

So you aren't surprised to find a commit you didn't write: I pushed **`84213cb`** onto `fix/conversation-relevance-default-threshold` — two bracketing tests that pin the new 0.5 default (details below). It's test-only; `git diff src/` against your head is empty. Please `git pull` before your next push, and tell me if you'd rather back it out.

Everything else on the branch (`dde1127`, `b23ed27`, `ff10455`, `8cc58db`, `116b1e4`) is your own work — I'm not claiming any of it.

## What works well

The `||` → `??` change is exactly right, and the `threshold: 0` regression test at `test/external/assertions.test.ts:300` is the easy-to-miss half of it — plenty of "fix the falsy default" patches forget that an explicit `0` is a legitimate value. Picking `0.5` also matches DeepEval's own `ConversationRelevancyMetric` default, which keeps this file honest to the upstream it's ported from. Linking the issue in the test name is a nice touch.

## Remaining items

### 1. Naming the default — question, P2

My pushed tests *bracket* the default; they don't pin it. Before them, mutating `src/external/assertions/deepeval.ts:42` to `?? 0.9` left the entire assertion surface green:

```
npx vitest run test/external/ test/assertions/
Test Files  63 passed (63)
     Tests  1348 passed (1348)
```

…while genuinely breaking users (a 5-turn conversation with 4/5 windows relevant scores 0.8 and now fails, with no config change). The new tests catch drift to 0.3, 0.4, 0.51, 0.9 and 1 — but they only bracket to `(0.4, 0.5]`, so a drift to `0.45` still slips through.

Would you be open to extracting `const DEFAULT_CONVERSATION_RELEVANCE_THRESHOLD = 0.5;` next to `DEFAULT_WINDOW_SIZE` (`deepeval.ts:15`) and asserting on it directly? That pins the value exactly, and it matches the line immediately above yours — line 41 already resolves `windowSize` through a named constant. I deliberately left this out because it reshapes production code and felt like your call, not mine.

### 2. `examples/eval-conversation-relevance/README.md:55` — P3

```md
- `threshold`: Minimum score required to pass (0-1)
- `config.windowSize`: Number of messages in each sliding window (default: 5)
```

The `windowSize` bullet documents its default; `threshold` now has one too. That README is the second user-facing surface for this assertion, and a reader who omits `threshold` there now gets a silent 0.5 gate the page never mentions.

### 3. Release-note visibility — question, for maintainers

This is behavior-breaking for anyone currently omitting `threshold`: it used to collapse to `0`, so every conversation-relevance assertion passed unconditionally; now anything under 0.5 fails. That's precisely the fix #10141 asked for, and your doc sentence covers it — but it can turn previously-green evals red with zero config change, so it may deserve a release-note callout. Nothing in-repo is affected: every fixture and example sets `threshold` explicitly.

CI is green across all jobs on `84213cb`.

Gujiassh:
Pulled the branch updates through `61e90fd`. The named default, exact-value coverage, and README note all look right; please keep the direct commits.

I re-ran the focused conversation-relevance tests (21/21), the combined external/assertion suite (63 files, 1,354 tests), TypeScript, and the architecture check locally. I agree the behavior change is worth calling out in release notes, but I left the placement to the maintainer release process rather than adding a separate release artifact to this fix.

mldangelo-oai (approved):
Thanks for the thoughtful contribution! This fixes a subtle fail-open edge case while preserving explicit zero thresholds, with strong coverage for negation and grader failures. Really appreciate the care here.

PR #10145, opened 2026-07-19
fix(providers): preserve Azure completion no-output semantics

mldangelo-oai:
## Summary

- preserve a nullish output when Azure completion returns an empty or missing choices list, so the evaluator records the canonical no-output result instead of grading an empty string
- retain the existing fallback for a present completion choice with no text and keep content-filter behavior unchanged
- update the two Azure regressions to cover the no-output boundary

## Validation

- `PROMPTFOO_CONFIG_DIR=/private/tmp/promptfoo-release-audit-azure-postmerge PROMPTFOO_DISABLE_TELEMETRY=1 npx vitest run test/providers/azure/completion.test.ts test/providers/azure/chat.test.ts --sequence.shuffle=false --reporter=dot` (67 passed)
- `npm run l && npm run f`

Follow-up to #10124, which has already merged into main.


jameshiester-oai (commented):
Reviewed base `5469ee52d2ea5387dfe62c080698c0e2e2b9b6af` and head `fa646f2b518ebcf11cf8942b95de88424c28b400`.

Scope: fresh full review of the complete two-file base...head patch, the merged #10124 behavior it follows, Azure provider callers, and evaluator no-output handling.

Verdict: no actionable defects found. The new guard preserves `undefined` only when the first completion choice is absent; a present choice with null/missing text still follows the existing empty-string or content-filter fallback. All 34 exact-head checks are green.

Findings:
- P0–P3: none.
- Existing feedback: Copilot failed to review and Promptfoo Scanner left a summary-only clear result; there are no inline threads to duplicate.
- Simplification/test opportunities: the one-line guard is already the smallest clear implementation. An integrated Azure-provider-to-`runEval` test could restate the contract, but the provider boundary tests plus evaluator tests already cover both halves effectively.

Validation:
- `PROMPTFOO_CONFIG_DIR=/private/tmp/promptfoo-review-10145 PROMPTFOO_DISABLE_TELEMETRY=1 npx vitest run test/providers/azure/completion.test.ts test/providers/azure/chat.test.ts --sequence.shuffle=false --reporter=dot` — 67/67 passed.
- Independent focused run including `test/evaluator/runEval.test.ts` — 105/105 passed.
- Targeted `biome check` and `git diff --check` — passed; Biome reported one existing complexity warning and no errors.
- Credential-free loopback `npm run local -- eval ... --no-cache -o ... --env-file /dev/null` with sharing disabled and an Azure completion response containing `choices: []` — exited 100 with one canonical failed eval, `error: "No output"`, zero runtime errors, and preserved 5-token usage.
- The first loopback setup used `apiHost`, which intentionally normalizes to HTTPS and therefore failed against the HTTP stub; rerunning with the explicit `apiBaseUrl` option exercised the intended provider path successfully.

Limitations: no live Azure credential or hosted request was used. The full local build was not repeated because TypeScript and all exact-head build lanes are green. The requested `/pr-review-toolkit:review-pr all parallel` command is unavailable (exit 127), so its six lenses were applied manually and reconciled with three read-only specialists.

mldangelo-oai:
@codex review

PR #10148, opened 2026-07-20
feat(providers): add Eden AI provider

MVS-source:
## What this PR does

Adds Eden AI as a provider (`edenai:<vendor/model>`). Eden AI (https://www.edenai.co) is an EU based, OpenAI-compatible LLM gateway: one key reaches models from many providers (OpenAI, Anthropic, Google, Mistral, Cohere and others) with vendor prefixed ids like `openai/gpt-4o-mini`.

It follows the merged Moonshot provider file for file: a small subclass of `OpenAiChatCompletionProvider`, the registry entry, env wiring, tests, docs, an example, and the architecture edge baseline.

Disclosure up front: I work at Eden AI, and we will maintain the provider.

## On the provider eligibility criteria

The guide asks for gateways to be reviewed against the same bar, so here are the five criteria in order.

**Legitimate model access.** Eden AI is an aggregator, so yes, we resell third party models. We do that under commercial agreements with the providers we expose, and we are permitted to resell them. Happy to confirm that in whatever form works for you.

**Accountable ownership.** Eden AI is a company with an identifiable operator and public legal pages: terms at https://www.edenai.co/terms, privacy policy at https://www.edenai.co/privacy, DPA at https://www.edenai.co/dpa. I am the maintenance contact for this provider.

**Evidence it will last.** Eden AI is funded, has paying customers, and the v3 API this PR targets is the one those customers already run on. It is not a new launch waiting for its first users.

**Clear data handling.** Documented at https://www.edenai.co/security. In short: prompts, uploaded files and model outputs are not retained by default, data is processed only for the duration needed to serve the request, customer inputs and outputs are never used by Eden AI to train models, and the platform is SOC 2 and ISO 27001 certified.

**Value beyond the generic path.** This is the criterion that actually matters here, so let me be straight about it. Eden AI is OpenAI-compatible and can be reached today with the `openai` provider plus `apiBaseUrl`. What that path cannot do:

- **Eden AI returns a cost on every response, and this provider reads it.** Our API returns a top level `cost` in USD on each response, documented at https://www.edenai.co/docs/v3/general/buying-credits. `calculateResponseCost` is overridden to use that number instead of estimating from a static price table, and falls back to the OpenAI calculation when the field is absent or on a cache hit. For a gateway fronting many vendors at negotiated rates, a catalog estimate is not just imprecise, it is the wrong number. One caveat for accuracy: if a user brings their own provider key, that field is itself an estimate and billing happens on the provider side. Two providers in this repo override this method today, Moonshot and this one.
- **It fails fast on subtypes it does not implement.** Eden AI is reached through chat completions. `edenai:embedding:...`, `edenai:responses:...`, `edenai:realtime:...`, `edenai:image:...` and eight more are rejected with a clear error rather than being sent to `/chat/completions` to fail in a confusing way. The generic path has no way to express that.
- **Correct key isolation.** The provider resolves `EDENAI_API_KEY` and does not fall back to `OPENAI_API_KEY` or forward the OpenAI organization header, either of which would 401 or send a key to the wrong service.
- **EU endpoint.** Eden AI serves an EU endpoint at `https://api.eu.edenai.run/v3`, covered in the provider page. I am listing this last deliberately: on its own it is only a base URL, so I am not resting the case on it.

If you would rather not add a dedicated prefix for a gateway, I am happy to convert this into a docs and example only PR on the generic `openai` path. Just say so.

## The architecture edge baseline

This is what was failing CI, and it was a file I simply missed.

`src/providers/edenai.ts` adds two cross layer imports from `src/providers` into `src/types`, `EnvOverrides` from `../types/env` and `ApiProvider` plus `ProviderOptions` from `../types/index`, so `providers -> legacy-contracts` goes from 277 to 279. That is one line in `architecture/edge-baseline.json`, the same shape as the merged Moonshot, Meta Model API and Open Interpreter provider PRs. The provider also adds two imports to `legacy-runtime` (`getEnvString` and `EnvVarKey` from `../envars`), but that edge has headroom against its baseline after your recent refactors, so it needed no change.

I bumped that single line by hand rather than running `npm run architecture:baseline`, because the regenerated file would also lock in reductions on `providers -> legacy-runtime` and `providers -> node` that came from your refactors and not from this PR. Related note that may be useful: those two edges are below their baseline on `main` as well, so running `npm run architecture:baseline` on `main` would capture a real reduction if you want it.

One suggestion for your side: the "Adding a Provider" checklist in `src/providers/AGENTS.md` says all seven items are required and does not mention the edge baseline, and neither does `CONTRIBUTING.md` or `site/docs/contributing.md`. All three provider PRs merged since the ratchet landed had to touch that file. Adding it as an eighth item would save the next contributor the same round trip.

## How I tested it

Rebased onto `main` at `0170037`. All seven steps of the `Style Check` job pass locally, including the two that CI skipped behind the failure:

| step | result |
| --- | --- |
| `npm run lint:ci` (Biome) | pass |
| `npm run format:check:prettier` | pass |
| `npx check-dependency-version-consistency` | pass |
| `npx madge $(git ls-files '*.ts') --circular` | pass |
| `npm run architecture:check` | pass |
| `npm run knip -- --no-progress` | pass, only pre-existing configuration hints |
| `npx lockfile-lint ...` | pass |

Plus:

- `npx vitest run test/providers/edenai.test.ts test/providers/registry.test.ts`: 113 tests pass
- `tsc --noEmit`: clean
- `npm run jsonSchema:generate` produces no diff, so the `config-schema.json` change in this PR matches what your generator emits
- When the PR was opened I ran a live eval through `edenai:openai/gpt-4o-mini` and `edenai:anthropic/claude-sonnet-4-5` against the real API and both returned correct output

One note on the Biome noise in the failing run: those 177 warnings and the "Biome encountered an unexpected error" lines come from `examples/redteam-mcp-agent/` and `scripts/architectureUtils.ts` and are unrelated to this PR. The Biome step itself passed. The only failure was the architecture step.

Files: `src/providers/edenai.ts`, the registry entry in `src/providers/registry.ts`, env wiring (`src/envars.ts`, `src/contracts/env.ts`, `src/util/provider.ts`), tests, `site/docs/providers/edenai.md` plus the index row, `examples/provider-edenai/`, and `architecture/edge-baseline.json`.


MVS-source:
CI is green again as of `c5bc815`. Sorry it sat red for a while.

The failure was not the Biome noise in that log. Those 177 warnings and the "Biome encountered an unexpected error" lines come from `examples/redteam-mcp-agent/` and `scripts/architectureUtils.ts`, and the Biome step itself passed. The real failure was step 10, `Check architecture boundaries`: this provider adds two cross layer imports from `src/providers` into `src/types`, so `providers -> legacy-contracts` went from 277 to 279 and the ratchet rejected it.

Fixed with a one line bump in `architecture/edge-baseline.json`. Two notes on how, in case you would have done it differently:

- I edited the single line by hand rather than running `npm run architecture:baseline`, because the regenerated file also locks in reductions on `providers -> legacy-runtime` and `providers -> node` that came from your refactors, not from this PR. Happy to run the generator instead if you prefer the file to stay machine written.
- Those two edges are below their baseline on `main` too, so `npm run architecture:baseline` on `main` would capture a real reduction if you want it.

I also rebased onto `0170037` (it was 181 commits behind, no conflicts) and re-verified locally: all seven `Style Check` steps pass, including `knip` and `lockfile-lint` which CI had skipped behind the failure, plus 113 tests, a clean `tsc --noEmit`, and `npm run jsonSchema:generate` producing no diff.

Two other things while I was in here. I rewrote the PR description to answer the provider eligibility criteria one by one, since the previous version leaned on the EU endpoint, which is only a base URL and therefore exactly what the "value beyond the generic path" bar excludes. The real answer is the `calculateResponseCost` override that reads the `cost` Eden AI returns on each response, and the guard that rejects subtypes this chat only provider does not implement. And a small suggestion: the "Adding a Provider" checklist in `src/providers/AGENTS.md` says all seven items are required but does not mention the edge baseline, and neither does `CONTRIBUTING.md`. All three provider PRs merged since the ratchet landed had to touch that file, so an eighth item there would save the next contributor this round trip.

PR #10149, opened 2026-07-20
fix(providers): align Azure Responses service tier

wholley-oai:
## Summary

The Azure provider documentation supports top-level `service_tier: priority` for `azure:responses`. The provider previously neither typed nor forwarded that option, while its cost adapter could still apply priority pricing from the configuration. As a result, a documented configuration could send a standard-tier request but report priority-tier cost.

This change:

- adds a Responses-only typed `service_tier` option without changing the shared Azure chat/completion options, which continue to use raw passthrough
- forwards the top-level option into the Azure Responses request body
- keeps raw passthrough fields last, so an explicit `passthrough.service_tier` overrides the top-level option
- derives cost from the effective request body, keeping reported pricing aligned with the tier sent to Azure
- preserves the existing passthrough-only priority path

## Validation

- `source ~/.nvm/nvm.sh && nvm use && npm ci`
- `npx vitest run test/providers/azure/responses.test.ts test/providers/azure/util.test.ts` (227 tests passed)
- `npm run tsc`
- `npm run f`
- `npm run l` (passes with the pre-existing cognitive-complexity warning in `getAzureResponsesBody`)

### DIFF
--- a/src/providers/azure/responses.ts
+++ b/src/providers/azure/responses.ts
@@ -38,19 +38,21 @@ export class AzureResponsesProvider extends AzureGenericProvider {
       modelName: this.deploymentName,
       providerType: 'azure',
       functionCallbackHandler: this.functionCallbackHandler,
-      // The processor invokes costCalculator(modelName, data.usage, requestConfig). calculateAzureCost
-      // expects (modelName, config, promptTokens, completionTokens) — extract the token counts from
-      // the Responses-shaped usage object (input_tokens/output_tokens) so cost is non-zero. The
-      // request config is the effective body after passthrough fields are applied, so map its
-      // service tier into the passthrough-based cost contract to keep pricing aligned with the wire.
-      costCalculator: (modelName: string, usage: any, config?: any) =>
-        calculateAzureCost(
+      // calculateAzureCost expects (modelName, config, promptTokens, completionTokens). Extract
+      // token counts from Responses usage and map the tier Azure actually served into its
+      // passthrough-based cost contract, falling back to the effective requested tier.
+      costCalculator: (modelName: string, usage: any, config?: any, responseData?: any) => {
+        const { service_tier: requestedServiceTier, ...costConfig } = config ?? {};
+        const serviceTier = responseData?.service_tier ?? requestedServiceTier;
+        return calculateAzureCost(
           modelName,
           {
-            ...config,
+            ...costConfig,
             passthrough: {
-              ...config?.passthrough,
-              ...(config?.service_tier === undefined ? {} : { service_tier: config.service_tier }),
+              ...costConfig.passthrough,
+              ...(serviceTier === undefined || serviceTier === null
+                ? {}
+                : { service_tier: serviceTier }),
             },
           },
           usage?.prompt_tokens ?? usage?.input_tokens,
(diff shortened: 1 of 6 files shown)


src/providers/azure/responses.ts
```diff
@@ -43,7 +40,9 @@ export class AzureResponsesProvider extends AzureGenericProvider {
       functionCallbackHandler: this.functionCallbackHandler,
       // The processor invokes costCalculator(modelName, data.usage, requestConfig). calculateAzureCost
       // expects (modelName, config, promptTokens, completionTokens) — extract the token counts from
-      // the Responses-shaped usage object (input_tokens/output_tokens) so cost is non-zero.
+      // the Responses-shaped usage object (input_tokens/output_tokens) so cost is non-zero. The
+      // request config is the effective body after passthrough fields are applied, so map its
+      // service tier into the passthrough-based cost contract to keep pricing aligned with the wire.
```
mldangelo-oai: [P1] Price the tier Azure actually served

This path prices the **requested** tier, but Azure can downgrade `priority` to `default` or serve `auto` using a priority deployment; the response's `service_tier` is the billing source of truth. I reproduced both cases through the CLI: priority/default reported **0.0755** instead of **0.03775**, while auto/priority reported **0.03775** instead of **0.0755**. Please use the returned tier with the request tier only as fallback (as the OpenAI Responses provider already does) and add regressions for both directions.

mldangelo-oai (changes requested):
## Review summary

Requesting changes: the new request-tier path still reports incorrect cost whenever Azure serves a different tier than the one requested.

Azure explicitly documents that the response `service_tier` reflects the tier actually used and can differ from the request (including priority downgrade and deployment-level priority with `auto`): https://learn.microsoft.com/rest/api/aifoundry/project/responses and https://learn.microsoft.com/en-us/azure/foundry/openai/concepts/priority-processing. The OpenAI Responses provider already prices the returned tier first; Azure should do the same.

I reproduced both directions end-to-end with a local Responses-compatible endpoint and exported eval results:

- requested `priority`, served `default`: reported **0.0755**, correct cost **0.03775**
- requested `auto`, served `priority`: reported **0.03775**, correct cost **0.0755**

Please add regression coverage where the requested and returned tiers disagree. The tests added here only cover matching tiers, so they cannot catch this.

I also independently confirmed the existing unresolved `service_tier: null` thread: the new type accepts null, the request builder serializes it, and a contract-strict endpoint returns 400. Please address that thread without adding a second representation or extra adapter complexity.

### Validation performed

- exact head: `9f2bb69ca70daa74a181fdbee1650651aa52ef10`
- **593/593** Azure tests passed
- **332/332** focused Azure, registry, and OpenAI Responses comparison tests passed
- TypeScript and Biome passed; only the pre-existing cognitive-complexity warning remains
- verified top-level/prompt/passthrough precedence and distinct cache entries for different tiers
- all three current GitHub workflows are green

Stable code link: https://github.com/promptfoo/promptfoo/blob/9f2bb69ca70daa74a181fdbee1650651aa52ef10/src/providers/azure/responses.ts#L41-L55

mldangelo-oai (approved):
Re-reviewed the current head `486dc6b8e36afbef7b57abcc31408ea5e9c56d32`.

The issues from my earlier request-changes review are fixed:

- Azure pricing now uses the service tier actually returned by the response, with the request as a fallback only when the response omits it.
- Null service tiers are omitted rather than serialized.
- Response-only `scale` is no longer accepted as a request tier.

Verification: 732 relevant tests pass, the TypeScript and changed-file Biome checks pass, and all 43 GitHub checks are green. All four remaining threads refer to outdated code. This approval supersedes my earlier request for changes.

The branch is behind `main`; refresh it and rerun the required checks before merging.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #10150, opened 2026-07-20
fix(providers): restore OpenAI request isolation

wholley-oai:
## Summary

PR #10125 aligned OpenAI provider behavior, but three edge cases could now cross account boundaries, drop valid chat parameters after model routing, or leave abandoned background work running.

This PR restores the intended behavior with focused changes:

- OpenAI TTS only uses persistent caching for authenticated requests when a stable, non-secret project, tenant, or account discriminator is present. Unauthenticated endpoints and explicitly discriminated requests remain cacheable, while API keys never enter cache identities.
- Chat capability checks use a string passthrough model as the effective model. Routing GPT-5 to gpt-4.1 now keeps temperature and max_tokens; routing a regular model to an o-series model applies reasoning limits. Subclass capability hooks remain unchanged when there is no override.
- Background Responses creation cancels a late-accepted queued job when its final subscriber has aborted and the job is not persistently resumable. Persisted shared jobs and jobs with another active subscriber remain alive.

## User impact

- One OpenAI account cannot receive speech audio cached for a different account when both use the default API endpoint.
- Routed chat requests retain or remove model-specific parameters according to the model actually sent upstream.
- Sole-caller cancellation no longer leaks accepted background work for provider-level API keys.

## Validation

- npx vitest run test/providers/openai/tts.test.ts test/providers/openai/chat.test.ts test/providers/openai/responses/request.test.ts (243 tests passed)
- npm run tsc
- npm run f
- npm run l
- git diff --check

## Review context

- TTS cache isolation: https://github.com/promptfoo/promptfoo/pull/10125#discussion_r3597586926
- Chat passthrough capabilities: https://github.com/promptfoo/promptfoo/pull/10125#discussion_r3604863186
- Background cancellation: https://github.com/promptfoo/promptfoo/pull/10125#discussion_r3598225198

### DIFF
--- a/src/providers/bedrock/mantleChat.ts
+++ b/src/providers/bedrock/mantleChat.ts
@@ -58,16 +58,16 @@ export function getBedrockMantleChatBaseUrl(region: string, modelName?: string):
  * configured mantle `apiBaseUrl`.
  */
 export class BedrockMantleChatProvider extends OpenAiChatCompletionProvider {
-  protected getCapabilityModelName(): string {
-    return this.modelName.replace(/^(openai|xai)\./, '');
+  protected getCapabilityModelName(modelName = this.modelName): string {
+    return modelName.replace(/^(openai|xai)\./, '');
   }
 
-  protected isReasoningModel(): boolean {
-    return isBedrockGrokModel(this.modelName) || super.isReasoningModel();
+  protected isReasoningModel(modelName = this.modelName): boolean {
+    return isBedrockGrokModel(modelName) || super.isReasoningModel(modelName);
   }
 
-  protected supportsTemperature(): boolean {
-    return isBedrockGrokModel(this.modelName) || super.supportsTemperature();
+  protected supportsTemperature(modelName = this.modelName): boolean {
+    return isBedrockGrokModel(modelName) || super.supportsTemperature(modelName);
   }
 
   async getOpenAiBody(
--- a/src/providers/openai/chat.ts
+++ b/src/providers/openai/chat.ts
@@ -264,8 +264,10 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
       capabilityModelName.includes('/o3') ||
       capabilityModelName.includes('/o4');
     const isGPT5Model = this.isGPT5Model(passthroughModel);
-    const isReasoningModel = this.isReasoningModel(passthroughModel);
-    const supportsTemperature = this.supportsTemperature(passthroughModel);
+    const isOpenAiReasoningModel = isGPT5Model || isOSeriesModel;
+    const isReasoningModel = this.isReasoningModel(passthroughModel) || isOpenAiReasoningModel;
+    const supportsTemperature =
+      this.supportsTemperature(passthroughModel) && !isOpenAiReasoningModel;
     const maxCompletionTokens = isReasoningModel
       ? (config.max_completion_tokens ?? getEnvInt('OPENAI_MAX_COMPLETION_TOKENS'))
       : undefined;
--- a/src/providers/openai/chat.ts
+++ b/src/providers/openai/chat.ts
@@ -252,12 +252,10 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
 
     const passthroughModel =
       typeof config.passthrough?.model === 'string' ? config.passthrough.model : undefined;
-    const capabilityModelName = (passthroughModel ?? this.getCapabilityModelName()).replace(
+    const capabilityModelName = this.getCapabilityModelName(passthroughModel).replace(
       /(^|\/)ft:/,
       '$1',
     );
-    const capabilityModelIsGPT5 =
-      capabilityModelName.startsWith('gpt-5') || capabilityModelName.includes('/gpt-5');
     const isOSeriesModel =
       capabilityModelName.startsWith('o1') ||
       capabilityModelName.startsWith('o3') ||
@@ -265,13 +263,9 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
       capabilityModelName.includes('/o1') ||
       capabilityModelName.includes('/o3') ||
       capabilityModelName.includes('/o4');
-    const isGPT5Model = passthroughModel === undefined ? this.isGPT5Model() : capabilityModelIsGPT5;
-    const isReasoningModel =
-      passthroughModel === undefined
-        ? this.isReasoningModel()
-        : capabilityModelIsGPT5 || isOSeriesModel;
-    const supportsTemperature =
-      passthroughModel === undefined ? this.supportsTemperature() : !isReasoningModel;
+    const isGPT5Model = this.isGPT5Model(passthroughModel);
+    const isReasoningModel = this.isReasoningModel(passthroughModel);
+    const supportsTemperature = this.supportsTemperature(passthroughModel);
     const maxCompletionTokens = isReasoningModel
       ? (config.max_completion_tokens ?? getEnvInt('OPENAI_MAX_COMPLETION_TOKENS'))
       : undefined;
--- a/src/providers/openai/chat.ts
+++ b/src/providers/openai/chat.ts
@@ -256,10 +256,8 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
       /(^|\/)ft:/,
       '$1',
     );
-    const isGPT5Model =
-      this.isGPT5Model() ||
-      capabilityModelName.startsWith('gpt-5') ||
-      capabilityModelName.includes('/gpt-5');
+    const capabilityModelIsGPT5 =
+      capabilityModelName.startsWith('gpt-5') || capabilityModelName.includes('/gpt-5');
     const isOSeriesModel =
       capabilityModelName.startsWith('o1') ||
       capabilityModelName.startsWith('o3') ||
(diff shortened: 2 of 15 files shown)


src/providers/openai/tts.ts
```diff
@@ -299,15 +299,21 @@ export class OpenAiTtsProvider extends OpenAiGenericProvider {
...
+    const usesAuthenticatedCustomEndpoint = !sendsToOpenAiApi && hasAuthentication;
+    const usesAuthenticatedOpenAiApiWithoutTenantDiscriminator =
+      sendsToOpenAiApi && hasAuthentication && !hasTenantDiscriminator;
     const cacheEnabled =
       isCacheEnabled() &&
       !hasSensitiveBody &&
       !hasSensitiveHeaderValue &&
       !hasSensitiveUrlPath &&
       !usesAuthenticatedCustomEndpoint &&
+      !usesAuthenticatedOpenAiApiWithoutTenantDiscriminator &&
```
mldangelo-oai, line 323: **[P2] Recognize `OpenAI-Organization` as a cache discriminator**

This new gate disables TTS persistence whenever an authenticated OpenAI request lacks a header named `project`, `tenant`, or `account`, but Promptfoo automatically sends the documented, non-secret `OpenAI-Organization` header when `organization`/`OPENAI_ORGANIZATION` is configured. That header is already present in `cacheHeaders`, yet it does not satisfy `hasTenantDiscriminator`.

I reproduced this with one provider configured with an API key and `organization: 'org-a'`: two identical sequential speech calls made **two upstream requests**, with zero cache reads or writes. Please include organization in the safe discriminator check and add a regression test so organization-scoped callers retain persistent caching. OpenAI documents this header as selecting the organization charged for the request: https://developers.openai.com/api/reference/overview#authentication.

mldangelo-oai (changes requested):
## Review summary — request changes

I reviewed the exact head `715b2c795e554091166eb713aafc3c7f3871d8b5`, the predecessor changes and existing threads, the affected OpenAI-compatible subclasses, and the cache/cancellation paths. The intended fixes are worthwhile, but I found three reproducible P2 regressions that should be addressed before merge.

### Findings

1. **Passthrough model routing bypasses subclass capability hooks.** ([existing inline thread](https://github.com/promptfoo/promptfoo/pull/10150#discussion_r3616276619), [changed code](https://github.com/promptfoo/promptfoo/blob/715b2c795e554091166eb713aafc3c7f3871d8b5/src/providers/openai/chat.ts#L268-L276))

   Any string `passthrough.model` replaces `isReasoningModel()`/`supportsTemperature()` with OpenAI-only name checks. I reproduced this with `GroqProvider('openai/gpt-oss-120b')`: without an override the request correctly contains `max_completion_tokens: 456`; adding the identical model as a no-op passthrough override drops that field and causes the non-reasoning `max_tokens` path to be used. Groq, xAI, Meta, Bedrock Mantle, Hyperbolic, OrcaRouter, and other subclasses have provider-specific capability behavior, so this is broader than the new OpenAI-only test demonstrates. Please classify the *effective* model through an override-aware capability hook and add subclass regression coverage. Groq's API explicitly supports `max_completion_tokens` and temperature: https://console.groq.com/docs/api-reference.

2. **Normal authenticated TTS calls no longer coalesce in flight.** ([existing inline thread](https://github.com/promptfoo/promptfoo/pull/10150#discussion_r3616276626), [changed code](https://github.com/promptfoo/promptfoo/blob/715b2c795e554091166eb713aafc3c7f3871d8b5/src/providers/openai/tts.ts#L308-L328))

   The new disk-cache predicate also gates the only call into `coalesceRequest`. I ran three identical concurrent speech calls against one `OpenAiTtsProvider` with a normal API key and no project header and observed **three upstream requests** instead of one. This multiplies cost and rate-limit pressure. The existing coalescing test was changed to add `OpenAI-Project`, masking the common authentication case. Please keep persistence disabled for isolation, but retain a provider-local/request-auth-scoped in-flight identity. Do **not** simply reuse the current global sanitized key for key-only requests, since that could coalesce concurrent calls from different accounts.

3. **The TTS cache gate ignores the standard `OpenAI-Organization` discriminator.** (new inline comment, [changed code](https://github.com/promptfoo/promptfoo/blob/715b2c795e554091166eb713aafc3c7f3871d8b5/src/providers/openai/tts.ts#L308-L316))

   Promptfoo automatically injects `OpenAI-Organization` from `organization`/`OPENAI_ORGANIZATION`, and it is already part of the sanitized cache headers, but `hasTenantDiscriminator` only recognizes `project|tenant|account`. With an API key and `organization: 'org-a'`, I reproduced two identical sequential speech calls making **two upstream requests** with no cache reads or writes. `OpenAI-Organization` is the documented non-secret organization-selection header, so it should qualify as an account discriminator (and should have a regression test): https://developers.openai.com/api/reference/overview#authentication.

### Verification performed

- Full `test/providers/openai` suite — **1,453 passed, 9 skipped**.
- Affected subclass suites (Groq, xAI, Meta, Bedrock Mantle, Hyperbolic, OrcaRouter, TrueFoundry) — **339 passed**.
- Focused changed-provider suites — **243 passed**.
- Biome check and `git diff --check` — passed; Biome reports seven existing cognitive-complexity warnings in these large provider methods.
- Real loopback CLI eval with routed chat, TTS, and queued/polled Responses — **3/3 passed**; inspected the actual request bodies and exported results.
- Targeted adversarial tests reproduced all three findings above.
- Current head CI is **45/45 green**.

The Responses cancellation change is appropriately small and the late-acceptance regression test exercises the intended non-persistable case; I did not find a new regression in that hunk. The branch is currently one lockfile-only dependency commit behind `main`, so please refresh and rerun checks after addressing the findings.

The simplest safe direction is to keep persistent and in-flight cache decisions separate and centralize effective-model capability classification instead of adding more provider-specific branches in the already high-complexity chat builder.

src/providers/openai/chat.ts
```diff
@@ -252,24 +252,22 @@ export class OpenAiChatCompletionProvider extends OpenAiGenericProvider {
...
       capabilityModelName.includes('/o3') ||
       capabilityModelName.includes('/o4');
-    const isPassthroughReasoningModel =
-      passthroughModel !== undefined && (isGPT5Model || isOSeriesModel);
-    const isReasoningModel = this.isReasoningModel() || isGPT5Model || isOSeriesModel;
+    const isGPT5Model = this.isGPT5Model(passthroughModel);
+    const isOpenAiReasoningModel = isGPT5Model || isOSeriesModel;
+    const isReasoningModel = this.isReasoningModel(passthroughModel) || isOpenAiReasoningModel;
+    const supportsTemperature =
+      this.supportsTemperature(passthroughModel) && !isOpenAiReasoningModel;
```
mldangelo-oai, line 270: **[P2] Sanitize the final effective reasoning-model request after passthrough**

I reproduced `new OpenAiChatCompletionProvider('gpt-4.1', { config: { max_completion_tokens: 456, passthrough: { model: 'o3', temperature: 0.7, max_tokens: 123 } } })`: `getOpenAiBody()` returns `{ model: 'o3', temperature: 0.7, max_completion_tokens: 456 }`. The later sanitizer correctly removes `max_tokens`, but passthrough reintroduces `temperature` after the capability calculation, so the final reasoning-model request still contains a parameter the effective model rejects. Normalize and sanitize the final body against `body.model` after applying passthrough, and add a regression that asserts the routed `o3` body omits both `temperature` and `max_tokens`.

src/providers/xai/chat.ts
```diff
@@ -642,8 +642,8 @@ class XAIProvider extends OpenAiChatCompletionProvider {
     return this.config?.apiKey;
   }
 
-  protected isReasoningModel(): boolean {
-    return GROK_REASONING_MODELS.includes(this.modelName);
+  protected isReasoningModel(modelName = this.modelName): boolean {
```
mldangelo-oai, line 645: **[P2] Apply Grok restrictions to the effective xAI model**

Although this hook now accepts a model override, downstream Grok-specific cleanup still keys off `this.modelName`. I reproduced `xai:grok-3-beta` with `passthrough: { model: 'grok-4' }`, `presence_penalty: 0.5`, `frequency_penalty: 0.7`, and `stop: ['END']`; the final Grok-4 request retains all three unsupported controls. Use the effective final `result.body.model` consistently for Grok-family restrictions and reasoning-effort validation, and add the routed-model regression.

src/providers/bedrock/mantleChat.ts
```diff
@@ -58,16 +58,16 @@ export function getBedrockMantleChatBaseUrl(region: string, modelName?: string):
...
 export class BedrockMantleChatProvider extends OpenAiChatCompletionProvider {
-  protected getCapabilityModelName(): string {
-    return this.modelName.replace(/^(openai|xai)\./, '');
+  protected getCapabilityModelName(modelName = this.modelName): string {
+    return modelName.replace(/^(openai|xai)\./, '');
   }
 
-  protected isReasoningModel(): boolean {
-    return isBedrockGrokModel(this.modelName) || super.isReasoningModel();
+  protected isReasoningModel(modelName = this.modelName): boolean {
```
mldangelo-oai, line 65: **[P2] Route and sanitize Bedrock Mantle using the effective model**

I reproduced a Mantle provider initially configured for `deepseek.v3.1` with `passthrough: { model: 'xai.grok-4.3' }`. The final request targets `xai.grok-4.3` but retains `presence_penalty`, `frequency_penalty`, and `stop`; its configured URL also remains `https://bedrock-mantle.us-east-1.api.aws/v1` instead of the Grok `/openai/v1` route. Passing the override into this capability hook is not sufficient while the cleanup and endpoint still use the originally configured model. Resolve the effective model once and use it for both Mantle endpoint selection and the final Grok restrictions; cover the passthrough case with a regression test.

mldangelo-oai (changes requested):
Re-reviewed current head `05ec912540036edd367d67292264ef1dec27fc2b`. The previous TTS coalescing, organization-isolation, and subclass-capability findings have been addressed, but the attached inline comments identify three still-reproducible effective-model bugs:

1. Routing `gpt-4.1` to `o3` through passthrough still sends unsupported `temperature`.
2. Routing an xAI legacy model to Grok 4 leaves unsupported sampling controls in the final request.
3. Routing Bedrock Mantle to `xai.grok-4.3` leaves those same controls and retains the non-Grok `/v1` endpoint.

I reproduced each with a separate exact-head regression test. The 1,816 existing focused tests, TypeScript, and the 43 PR checks are green, so these need dedicated regressions.

Please derive all capability checks, post-passthrough sanitization, and Mantle endpoint selection from the effective final model; add the three regression cases; then refresh against `main`.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #10151, opened 2026-07-20
fix(providers): correct Gemini Omni interactions

wholley-oai:
## Summary

Gemini Omni Flash uses Google's long-running Interactions API for video generation. The initial provider integration could route Vertex requests to unsupported regional endpoints, continue paid work after an eval was cancelled, accept unsupported safety options, send incompatible OpenAI-style text blocks, report unknown usage as free, and allow one Vertex follow-up alias to bypass local validation.

This change corrects those behaviors without changing other Google provider paths:

1. Vertex Omni always uses `/locations/global/interactions` and the global default host, regardless of ordinary Vertex region configuration. Explicit API host and proxy overrides remain supported.
2. The provider accepts the evaluator's abort signal and propagates it through interaction creation, polling requests and waits, video downloads and redirects, and the check before blob storage. Aborts retain `AbortError` semantics and stop subsequent network or storage work.
3. Custom safety settings now fail before authentication or network access when supplied as top-level config or through either passthrough alias, because the Interactions API does not support them.
4. OpenAI-style `input_text` and `output_text` blocks are normalized to Interactions text content for user and model messages.
5. Successful responses without a `usage` object retain `numRequests: 1` but leave token counts and cost unknown instead of reporting zeros and `$0`.
6. Vertex follow-up validation checks the top-level ID and both passthrough aliases independently by presence, so a falsy snake-case value cannot mask a real camelCase ID.

The Google provider documentation now states the global Vertex location and unsupported safety-setting constraint.

## Validation

- `source ~/.nvm/nvm.sh && nvm use` — Node 24.18.0 / npm 11.16.0
- `npm ci`
- `npx vitest run test/providers/google/interactions.test.ts test/providers/google/provider.test.ts test/providers/registry.test.ts --sequence.shuffle=false` — 209 passed
- `npx vitest run test/providers/google/interactions.test.ts test/providers/google/provider.test.ts test/providers/registry.test.ts` — 209 passed with randomized ordering
- `npx vitest run test/providers/google/interactions.test.ts --sequence.shuffle=false` — 50 passed after final endpoint-override coverage
- `npm run l`
- `npm run f`
- `npm run format:check`
- `git diff --check`
- `PROMPTFOO_DISABLE_REMOTE_GENERATION=true PROMPTFOO_CONFIG_DIR=/private/tmp/promptfoo-interactions-regression-state npm run local -- eval -c /private/tmp/promptfoo-interactions-regression-e2e.yaml --no-cache -o /private/tmp/promptfoo-interactions-regression-output.json` — deterministic loopback Interactions endpoint, 1/1 passed, score 1, no errors; raw provider response contained only `tokenUsage.numRequests: 1` and no cost when usage was absent

### DIFF
--- a/src/providers/google/interactions.ts
+++ b/src/providers/google/interactions.ts
@@ -10,7 +10,12 @@ import { GoogleAuthManager } from './auth';
 import { calculateGoogleCost, mergeGoogleCompletionOptions } from './util';
 
 import type { EnvOverrides } from '../../types/env';
-import type { ApiProvider, CallApiContextParams, ProviderResponse } from '../../types/index';
+import type {
+  ApiProvider,
+  CallApiContextParams,
+  CallApiOptionsParams,
+  ProviderResponse,
+} from '../../types/index';
 import type { CompletionOptions, GoogleProviderConfig } from './types';
 
 type InteractionContent = {
@@ -52,6 +57,38 @@ function normalizeAudioMimeType(format?: string): string {
     return 'audio/pcm';
   }
   return `audio/${format || 'mpeg'}`;
+}
+
+function getAbortError(signal: AbortSignal): Error {
+  const reason = signal.reason;
+  if (reason instanceof Error && reason.name === 'AbortError') {
+    return reason;
+  }
+  const error = new Error(reason instanceof Error ? reason.message : 'Request was aborted');
+  error.name = 'AbortError';
+  return error;
+}
+
+async function sleepWithAbort(ms: number, signal?: AbortSignal): Promise<void> {
+  if (!signal) {
+    await sleep(ms);
+    return;
+  }
+  if (signal.aborted) {
+    throw getAbortError(signal);
+  }
+  await new Promise<void>((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      signal.removeEventListener('abort', onAbort);
+      resolve();
+    }, ms);
+    const onAbort = () => {
+      clearTimeout(timeout);
+      signal.removeEventListener('abort', onAbort);
+      reject(getAbortError(signal));
+    };
+    signal.addEventListener('abort', onAbort, { once: true });
+  });
 }
 
 function parseInteractionInput(prompt: string): string | unknown[] | Record<string, unknown> {
@@ -107,7 +144,17 @@ function parseInteractionInput(prompt: string): string | unknown[] | Record<stri
             }
             return part;
           }
-          const imagePart = part as { type?: string; image_url?: string | { url?: string } };
+          const imagePart = part as {
+            type?: string;
+            text?: string;
+            image_url?: string | { url?: string };
+          };
+          if (
+            (imagePart.type === 'input_text' || imagePart.type === 'output_text') &&
+            typeof imagePart.text === 'string'
+          ) {
+            return { type: 'text', text: imagePart.text };
+          }
           if (imagePart.type === 'input_audio') {
             const inputAudio = (part as { input_audio?: { data?: string; format?: string } })
               .input_audio;
@@ -165,15 +212,6 @@ function parseInteractionInput(prompt: string): string | unknown[] | Record<stri
   }
 }
 
-function normalizeInteractionSafetySettings(
-  safetySettings: NonNullable<CompletionOptions['safetySettings']>,
-) {
-  return safetySettings.map(({ category, threshold, probability }) => ({
-    type: category.replace(/^HARM_CATEGORY_/i, '').toLowerCase(),
-    ...(threshold || probability ? { threshold: (threshold || probability)?.toLowerCase() } : {}),
-  }));
-}
-
 function getInteractionModalityTokenCount(
   details: Array<{ modality?: string; tokens?: number }> | undefined,
   modalities: string[],
@@ -221,20 +259,14 @@ function getVertexInteractionsEndpoint(
   projectId: string,
   env?: EnvOverrides,
 ): string {
-  const region =
-    config.region ||
-    env?.VERTEX_REGION ||
-    getEnvString('VERTEX_REGION') ||
-    getEnvString('GOOGLE_CLOUD_LOCATION') ||
-    'global';
   const configuredHost =
     config.apiBaseUrl ||
     config.apiHost ||
     env?.VERTEX_API_HOST ||
     getEnvString('VERTEX_API_HOST') ||
-    (region === 'global' ? 'aiplatform.googleapis.com' : `${region}-aiplatform.googleapis.com`);
+    'aiplatform.googleapis.com';
   const host = /^https?:\/\//i.test(configuredHost) ? configuredHost : `https://${configuredHost}`;
-  return `${host.replace(/\/$/, '')}/v1beta1/projects/${encodeURIComponent(projectId)}/locations/${encodeURIComponent(region)}/interactions`;
+  return `${host.replace(/\/$/, '')}/v1beta1/projects/${encodeURIComponent(projectId)}/locations/global/interactions`;
 }
 
 export class GoogleInteractionsProvider implements ApiProvider {
@@ -261,7 +293,15 @@ export class GoogleInteractionsProvider implements ApiProvider {
     return `[Google Interactions Provider ${this.modelName}]`;
   }
 
-  async callApi(prompt: string, context?: CallApiContextParams): Promise<ProviderResponse> {
+  async callApi(
+    prompt: string,
+    context?: CallApiContextParams,
+    callApiOptions?: CallApiOptionsParams,
+  ): Promise<ProviderResponse> {
+    const abortSignal = callApiOptions?.abortSignal;
+    if (abortSignal?.aborted) {
+      throw getAbortError(abortSignal);
+    }
     if (!prompt.trim()) {
       return { error: 'Prompt is required for Gemini Interactions API' };
     }
@@ -270,12 +310,25 @@ export class GoogleInteractionsProvider implements ApiProvider {
       this.config,
       context?.prompt?.config as Partial<CompletionOptions> | undefined,
     ) as GoogleProviderConfig;
-    const passthroughPreviousInteractionId =
-      config.passthrough?.previous_interaction_id ?? config.passthrough?.previousInteractionId;
-    if (config.vertexai && (config.previousInteractionId || passthroughPreviousInteractionId)) {
+    if (
+      config.vertexai &&
+      (config.previousInteractionId !== undefined ||
+        config.passthrough?.previous_interaction_id !== undefined ||
+        config.passthrough?.previousInteractionId !== undefined)
+    ) {
       return {
         error:
           'Gemini Omni on Vertex AI does not support previousInteractionId. Use the Google AI Studio route for conversational video editing.',
+      };
+    }
+    if (
+      config.safetySettings !== undefined ||
+      config.passthrough?.safetySettings !== undefined ||
+      config.passthrough?.safety_settings !== undefined
+    ) {
+      return {
+        error:
+          'Gemini Omni Flash does not support custom safety settings through the Interactions API.',
       };
     }
     if (
@@ -356,6 +409,9 @@ export class GoogleInteractionsProvider implements ApiProvider {
         ...config.headers,
       };
     }
+    if (abortSignal?.aborted) {
+      throw getAbortError(abortSignal);
+    }
     const unsupportedGenerationFields = new Set([
       ...(config.vertexai ? [] : ['temperature', 'top_p', 'topP']),
       'stop_sequences',
@@ -414,9 +470,6 @@ export class GoogleInteractionsProvider implements ApiProvider {
         ? { previous_interaction_id: config.previousInteractionId }
         : {}),
       ...(config.store === undefined ? {} : { store: config.store }),
-      ...(config.safetySettings
-        ? { safety_settings: normalizeInteractionSafetySettings(config.safetySettings) }
-        : {}),
       ...(Object.keys(generationConfig).length > 0 ? { generation_config: generationConfig } : {}),
       ...passthrough,
       background: false,
@@ -439,6 +492,7 @@ export class GoogleInteractionsProvider implements ApiProvider {
           method: 'POST',
           headers,
           body: JSON.stringify(body),
+          signal: abortSignal,
         } as RequestInit,
         getRequestTimeoutMs(),
         'json',
@@ -446,6 +500,9 @@ export class GoogleInteractionsProvider implements ApiProvider {
         true,
       )) as { data: InteractionResponse; cached: boolean; status: number; statusText: string });
     } catch (err) {
+      if (abortSignal?.aborted) {
+        throw getAbortError(abortSignal);
+      }
       return { error: `Gemini Interactions API error: ${String(err)}` };
     }
 
@@ -464,6 +521,9 @@ export class GoogleInteractionsProvider implements ApiProvider {
     const pollStartedAt = Date.now();
     let pollCount = 0;
     while (data.status === 'in_progress' && data.id) {
+      if (abortSignal?.aborted) {
+        throw getAbortError(abortSignal);
+      }
       const elapsed = Date.now() - pollStartedAt;
       if (elapsed >= pollTimeoutMs) {
         return {
@@ -472,7 +532,7 @@ export class GoogleInteractionsProvider implements ApiProvider {
         };
       }
       if (pollCount > 0) {
-        await sleep(Math.min(1_000, pollTimeoutMs - elapsed));
+        await sleepWithAbort(Math.min(1_000, pollTimeoutMs - elapsed), abortSignal);
       }
       try {
         ({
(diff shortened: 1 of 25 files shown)


src/providers/google/interactions.ts
```diff
@@ -270,14 +310,27 @@ export class GoogleInteractionsProvider implements ApiProvider {
...
       context?.prompt?.config as Partial<CompletionOptions> | undefined,
     ) as GoogleProviderConfig;
-    const passthroughPreviousInteractionId =
-      config.passthrough?.previous_interaction_id ?? config.passthrough?.previousInteractionId;
-    if (config.vertexai && (config.previousInteractionId || passthroughPreviousInteractionId)) {
+    if (
+      config.vertexai &&
+      (config.previousInteractionId !== undefined ||
+        config.passthrough?.previous_interaction_id !== undefined ||
+        config.passthrough?.previousInteractionId !== undefined)
```
mldangelo-oai, line 329: **[P1] Keep the Vertex route immutable across prompt overrides.**

This guard trusts the merged `config.vertexai`, which a prompt-level config can replace. I reproduced this with a provider constructed as `vertex:gemini-omni-flash-preview`, `context.prompt.config.vertexai: false`, and `passthrough.previous_interaction_id: 'interaction-0'`: the call returned successfully and sent the prompt plus follow-up ID to `https://generativelanguage.googleapis.com/v1beta/interactions` using the available Gemini key. That bypasses both the follow-up validation and the expected Vertex data/auth boundary. Please derive the route from the originally constructed provider (and prevent prompt overrides from flipping it), then validate against that route and add the prompt-override regression test.

src/providers/google/interactions.ts
```diff
@@ -673,30 +754,37 @@ export class GoogleInteractionsProvider implements ApiProvider {
...
+            total:
+              usage.total_tokens ??
+              (promptTokens ?? 0) + (outputTokens ?? 0) + (thoughtTokens ?? 0),
+            cached: usage.total_cached_tokens ?? 0,
+            numRequests: 1,
+            ...((thoughtTokens ?? 0) > 0
+              ? { completionDetails: { reasoning: thoughtTokens } }
+              : {}),
+          }
+        : { numRequests: 1 },
```
mldangelo-oai, line 792: **[P1] Preserve unknown usage through the persisted eval result.**

This fixes only the raw provider response. I ran a loopback CLI eval returning a completed interaction without `usage`: `response.tokenUsage` was correctly `{ numRequests: 1 }` and `response.cost` was absent, but the exported result still had `results.results[0].cost = 0`, zero prompt/completion/total tokens, and prompt metric cost `0`. `createEvaluateResult`/`EvalResult` and aggregation coerce the missing values back to zero, so users still see an unknown-cost request as free. Please carry unknown semantics through evaluation/persistence/export (or otherwise explicitly represent unknown) and add an end-to-end assertion against the persisted result, not only the raw response.

src/providers/google/interactions.ts
```diff
@@ -221,20 +259,14 @@ function getVertexInteractionsEndpoint(
...
   const configuredHost =
     config.apiBaseUrl ||
     config.apiHost ||
     env?.VERTEX_API_HOST ||
     getEnvString('VERTEX_API_HOST') ||
-    (region === 'global' ? 'aiplatform.googleapis.com' : `${region}-aiplatform.googleapis.com`);
+    'aiplatform.googleapis.com';
   const host = /^https?:\/\//i.test(configuredHost) ? configuredHost : `https://${configuredHost}`;
-  return `${host.replace(/\/$/, '')}/v1beta1/projects/${encodeURIComponent(projectId)}/locations/${encodeURIComponent(region)}/interactions`;
+  return `${host.replace(/\/$/, '')}/v1beta1/projects/${encodeURIComponent(projectId)}/locations/global/interactions`;
```
mldangelo-oai, line 276: **[P2] Filter Omni-unsupported sampling controls on the repaired Vertex path.**

Correcting the endpoint makes this path usable, but the request builder below still explicitly permits and serializes `temperature` and `top_p` for Vertex. A focused probe produced `generation_config: { temperature: 0.3, top_p: 0.8 }`. [Google's Omni limitations](https://ai.google.dev/gemini-api/docs/omni) explicitly list both controls as unsupported, so shared Vertex configs that set them can still fail or behave unexpectedly. Please filter them consistently for both routes and update the Vertex request expectation that currently locks in their transmission.

src/providers/google/interactions.ts
```diff
@@ -270,14 +310,27 @@ export class GoogleInteractionsProvider implements ApiProvider {
...
+    ) {
       return {
         error:
           'Gemini Omni on Vertex AI does not support previousInteractionId. Use the Google AI Studio route for conversational video editing.',
       };
     }
+    if (
+      config.safetySettings !== undefined ||
+      config.passthrough?.safetySettings !== undefined ||
+      config.passthrough?.safety_settings !== undefined
```
mldangelo-oai: **[P2] Do not reject an empty safety-settings array.**

The new presence checks make both `safetySettings: []` and `passthrough.safety_settings: []` fatal even though an empty list configures no custom safety behavior. I ran the same loopback CLI eval against base and head: base passed and sent an empty list; head returned this error before network and exited with an eval error. Please treat an empty array like the existing empty-tools case, omit it from the request, and only reject a non-empty custom setting.

src/providers/google/interactions.ts
```diff
@@ -356,6 +409,9 @@ export class GoogleInteractionsProvider implements ApiProvider {
         ...config.headers,
       };
     }
+    if (abortSignal?.aborted) {
+      throw getAbortError(abortSignal);
```
mldangelo-oai, line 428: **[P2] Preserve `AbortError` when Vertex auth fails after cancellation.**

This post-auth check only executes if authentication succeeds. I reproduced an OAuth/metadata rejection that races with `abortSignal.abort(new Error('cancelled during auth'))`: the provider resolved with `Gemini Omni Vertex AI authentication error: ...` instead of throwing an `AbortError`, so the evaluator treats a cancelled run as an ordinary provider failure. Please check `abortSignal?.aborted` in the authentication catch and rethrow the normalized abort error there as well; this branch is also currently uncovered.

mldangelo-oai (changes requested):
## Review summary

Requesting changes after reviewing exact head `ec5208be5a6af7d6b602a3e432b13feb88c5f0ee`, the predecessor validation history, current Google contracts, and the runtime paths. I found five actionable issues:

1. **P1 — a prompt-level `vertexai: false` override bypasses the new Vertex follow-up guard and silently sends the prompt and interaction ID to AI Studio.** I reproduced a `vertex:gemini-omni-flash-preview` provider issuing `POST https://generativelanguage.googleapis.com/v1beta/interactions` with `previous_interaction_id: interaction-0` and no error. This defeats both the route boundary and the intended validation. [Changed guard](https://github.com/promptfoo/promptfoo/blob/ec5208be5a6af7d6b602a3e432b13feb88c5f0ee/src/providers/google/interactions.ts#L309-L317).
2. **P1 — missing usage is still persisted/exported as zero tokens and free.** In a real loopback CLI eval whose completed interaction omitted `usage`, the raw provider response correctly contained only `{ numRequests: 1 }`, but `results.results[0].cost` was `0`, prompt/completion/total tokens were `0`, and prompt metric cost was `0`. The stated unknown-usage fix therefore does not survive the evaluator/persistence boundary. [Changed response construction](https://github.com/promptfoo/promptfoo/blob/ec5208be5a6af7d6b602a3e432b13feb88c5f0ee/src/providers/google/interactions.ts#L754-L773).
3. **P2 — the repaired Vertex path still forwards Omni-unsupported `temperature` and `top_p`.** A focused probe produced `generation_config: { temperature: 0.3, top_p: 0.8 }`; Google's Omni limitations explicitly identify both as unsupported. Shared Google/Vertex configs that set sampling controls can consequently remain broken even after routing is corrected. [Endpoint change](https://github.com/promptfoo/promptfoo/blob/ec5208be5a6af7d6b602a3e432b13feb88c5f0ee/src/providers/google/interactions.ts#L257-L269), [generation configuration](https://github.com/promptfoo/promptfoo/blob/ec5208be5a6af7d6b602a3e432b13feb88c5f0ee/src/providers/google/interactions.ts#L415-L444).
4. **P2 — empty safety arrays now unnecessarily fail existing evals.** `safetySettings: []` and `passthrough.safety_settings: []` configure no custom safety behavior. The same loopback CLI eval passed on the base and sent an empty array, but failed before network on this head. Empty settings should be omitted/treated like empty tools; only non-empty settings need rejection. [Changed safety guard](https://github.com/promptfoo/promptfoo/blob/ec5208be5a6af7d6b602a3e432b13feb88c5f0ee/src/providers/google/interactions.ts#L324-L328).
5. **P2 — a Vertex auth failure racing cancellation loses `AbortError` semantics.** The new post-auth check only runs on successful auth; if an OAuth/metadata operation rejects after the eval is aborted, the auth catch returns an ordinary provider error. I reproduced the resolved authentication error instead of the expected `AbortError`. Handle an aborted signal in the auth catch as well. [Post-auth check](https://github.com/promptfoo/promptfoo/blob/ec5208be5a6af7d6b602a3e432b13feb88c5f0ee/src/providers/google/interactions.ts#L412-L413).

### QA performed

- Full Google-provider and provider-registry suite: **894/894 passed**.
- TypeScript check passed.
- Exercised real loopback CLI evals for OpenAI-style text normalization, omitted usage, and empty safety settings, including base-vs-head comparison.
- Exercised runtime cancellation during creation and polling and focused auth/cancellation races.
- Checked current vendor guidance: [Interactions overview](https://ai.google.dev/gemini-api/docs/interactions-overview?hl=en) confirms custom safety settings are unsupported and that omitted `store` defaults to `true`; [Gemini Omni limitations](https://ai.google.dev/gemini-api/docs/omni) explicitly list `temperature` and `top_p` as unsupported.
- CI is **44/44 green**, but Codecov reports **82.28% patch coverage with 14 changed lines uncovered**, including several abort/error branches. Biome also reports cognitive-complexity warnings of **73** for `normalizePart` and **163** for `callApi` (configured maximum 30); the new abort helpers duplicate existing fetch utilities, so please avoid further branching where a shared helper or smaller functions would suffice.

One additional existing privacy concern worth tracking separately: because this provider omits `store` when it is unspecified, Interactions defaults to server-side storage; Google documents up to **55 days** retention on the paid tier. This is not introduced here, but it is important context for a provider handling evaluation prompts and generated video.

src/types/index.ts
```diff
@@ -335,7 +335,7 @@ const PromptMetricsSchema = z.object({
       strategyFailCount: z.record(z.string(), z.number()),
     })
     .optional(),
-  cost: z.number(),
+  cost: z.number().optional(),
```
mldangelo-oai, line 338: **[P1] Coordinate optional metrics with the live Cloud contract before merging**

The live `promptfoo-cloud/main` head `3683e330ba4b11bac2ec5e784430bda04d98acb4` still defines `PromptMetricsSchema.cost` and `TokenUsageSchema.total/prompt/completion/cached` as required numbers: https://github.com/promptfoo/promptfoo-cloud/blob/3683e330ba4b11bac2ec5e784430bda04d98acb4/shared/dto/results.ts#L246-L268

I ran Cloud's actual runtime schema against this PR's usage-less Omni result. The known-metrics control passes, while the PR output fails on `cost` and all four omitted token fields. Making these upstream fields optional without the paired Cloud DTO/API/client migration publishes responses that violate the consuming API contract. Keep the existing public shape until Cloud is migrated, or land and verify both changes together with a cross-repo regression.

src/util/tokenUsageUtils.ts
```diff
@@ -238,6 +298,34 @@ export function accumulateResponseTokenUsage(
...
+        response.tokenUsage,
+        response.tokenUsage.numRequests === undefined,
+      );
+    } else {
+      accumulateTokenUsagePreservingUnknown(target, {
+        ...response.tokenUsage,
+        numRequests: undefined,
+      });
+    }
+  } else if (response && countAsRequest) {
```
mldangelo-oai, line 324: **[P1] Do not count provider-local preflight errors as upstream requests**

This fallback treats any response without `tokenUsage` as a completed network request. I added an evaluator regression with one real successful call reporting `{ prompt: 5, completion: 3, total: 8, numRequests: 1 }` and one local provider validation failure `{ error: 'Unsupported provider options; no network request was made' }`. The resulting metrics become `{ numRequests: 2 }`: the nonexistent request is counted and all previously known token counts are deleted as unknown.

This occurs through the newly changed evaluator accumulation path and corrupts otherwise accurate eval totals. Distinguish an actual attempted upstream request from a local preflight/error response, or pass an explicit `countAsRequest` from the call sites; add the mixed success/preflight regression.

src/providers/google/interactions.ts
```diff
@@ -612,27 +698,44 @@ export class GoogleInteractionsProvider implements ApiProvider {
...
+        }
       } catch (err) {
+        if (abortSignal?.aborted) {
+          throw getAbortError(abortSignal);
+        }
         return { error: `Failed to store Gemini interaction video: ${String(err)}` };
       }
     }
 
     const usage = data.usage;
```
mldangelo-oai, line 731: **[P1] Keep an empty or incomplete Gemini usage object unknown**

The current check distinguishes omitted `usage` from a truthy object, but not from the equally unreported `usage: {}`. I reproduced a completed interaction with `usage: {}`: this code reports `{ prompt: 0, completion: 0, cached: 0, total: 0, numRequests: 1 }` and calculates a zero cost. That advertises an unmetered, paid model invocation as free.

Determine the availability of each actual usage field rather than using object truthiness. Emit only counts that were really reported, keep total and cost undefined until there is sufficient billing information, and add regressions for empty and partially populated usage objects.

src/util/tokenUsageUtils.ts
```diff
@@ -157,6 +157,66 @@ export function accumulateTokenUsage(
...
+  const priorRequests = target.numRequests ?? 0;
+  const updateRequests = update.numRequests ?? (incrementRequests ? 1 : 0);
+  const priorCompletionDetails = target.completionDetails;
+  const priorCounts = {
+    prompt: target.prompt,
+    completion: target.completion,
+    cached: target.cached,
+    total: target.total,
+  };
+  const hasReportedTokenCounts =
```
mldangelo-oai, line 184: **[P2] Preserve unknown token counts independently per field**

`hasReportedTokenCounts` is shared across every count, so reporting any one field causes `accumulateTokenUsage` to fabricate zeroes for all missing fields. I reproduced `accumulateTokenUsagePreservingUnknown(target, { prompt: 6, numRequests: 1 })`; the output is `{ prompt: 6, completion: 0, cached: 0, total: 0, numRequests: 1 }`, even though completion, cached tokens, and total were never reported.

Track reported/unknown status separately for `prompt`, `completion`, `cached`, and `total`. Only preserve or add a field when the current counted response actually supplied that specific field; add partial-usage regression coverage.

src/providers/google/interactions.ts
```diff
@@ -261,23 +302,47 @@ export class GoogleInteractionsProvider implements ApiProvider {
...
+      ...mergeGoogleCompletionOptions(
+        this.config,
+        context?.prompt?.config as Partial<CompletionOptions> | undefined,
+      ),
+      vertexai: this.isVertexProvider,
+    } as GoogleProviderConfig;
+    if (
+      config.vertexai &&
+      (Boolean(config.previousInteractionId) ||
+        config.passthrough?.previous_interaction_id !== undefined ||
```
mldangelo-oai, line 328: **[P2] Treat empty passthrough follow-up IDs as absent**

The fixed top-level `previousInteractionId` now correctly uses truthiness, but both passthrough aliases still use `!== undefined`. I reproduced all three first-turn cases: `passthrough.previous_interaction_id: ''`, `passthrough.previousInteractionId: ''`, and `passthrough.previous_interaction_id: null`; each returns the unsupported Vertex follow-up error before making a request even though no interaction ID was supplied. Empty templated optional values are a normal initial-turn configuration.

Normalize the alias to a nonempty interaction ID before enforcing the Vertex follow-up restriction, and add empty-string/null regressions for both passthrough spellings.

mldangelo-oai (changes requested):
Re-reviewed current head `1da69ea0a02fde9f26e5e805ba1bb035097c4f51`, including the live `promptfoo-cloud/main` at `3683e330ba4b11bac2ec5e784430bda04d98acb4`.

The attached inline comments cover five issues not already described by a current review thread:

1. **P1: Cloud contract break** — Cloud still requires cost and four token counts, and actual PR output fails its runtime schema on all five fields.
2. **P1: Phantom upstream requests** — locally rejected provider calls are counted as network requests and erase known usage.
3. **P1: Falsely free Gemini requests** — `usage: {}` is converted into zero tokens and zero cost.
4. **P2: Fabricated partial counts** — a response reporting only `prompt: 6` acquires invented zero completion, cache, and total counts.
5. **P2: Empty passthrough interaction IDs** — empty-string and null aliases incorrectly block a valid initial Vertex request.

Three already-open threads also remain valid; I independently reproduced them rather than adding duplicate comments:

- Cancellation can delete shared blob bytes after a concurrent same-hash evaluation successfully records its surviving reference.
- The sequence provider converts unknown Gemini usage back into fabricated zero token counts.
- The results table hides a valid filtered cost and filtered token total when the unfiltered total is unknown.

The existing baseline is green: 4,531 tests across 165 files, frontend coverage, TypeScript, and all 43 PR checks. The missing cases were reproduced with focused adversarial regressions or the live Cloud schema.

Please fix the new and existing blockers, add regressions, coordinate the Cloud DTO/API migration before changing public metrics, and refresh against `main`. I recommend splitting the narrow Omni provider fix from the cross-repo token-accounting/metrics contract migration.

PR #10165, opened 2026-07-21
feat(providers): refresh misc model support

ianw-oai:
Refresh the Alibaba Qwen and hosted Chinese-model catalog, make Kimi/MiniMax defaults and reasoning controls usable, and expose current models in the red-team target picker.


mldangelo-oai (approved):
Good

Issue #10164, opened 2026-07-21
eval --watch: tests file changes never trigger re-run (scalar tests: never watched; watcher never ready on macOS)


algebananazzzzz:
## Summary

`promptfoo eval --watch` fails to re-run when the tests file changes, for two independent reasons:

1. **Scalar `tests:` values are never added to the watch list** (code-level, all platforms).
2. **On macOS, the watcher never becomes ready even with the array form** (observed on 0.121.11 and 0.121.19 via npx, node 22.22.0, darwin 24.6.0).

## 1. Scalar `tests:` never watched

In `doEval`'s watch block, watch paths for tests are collected with:

```js
const varPaths = Array.isArray(config.tests) ? config.tests.flatMap(...) : [];
```

The scalar form is valid config and works for evaluation:

```yaml
tests: file://cases.yaml
```

but because it is not an array, the tests file is silently excluded from `watchPaths`; only the config file is watched. Appending a test case to `cases.yaml` never triggers a re-run, with no warning. The array form (`tests:\n  - file://cases.yaml`) does get collected.

Suggested fix: normalize `config.tests` to an array before collecting watch paths, or warn when watch mode ends up watching only the config.

## 2. macOS: watcher never ready, no re-runs at all

Minimal repro (macOS, node v22.22.0):

```yaml
# promptfooconfig.yaml
description: watch repro
prompts:
  - "Say {{word}}"
providers:
  - echo
tests:
  - file://tests.yaml
```

```yaml
# tests.yaml
- vars:
    word: hi
```

```sh
sleep 600 | npx -y promptfoo@0.121.19 eval --watch > watch.log 2>&1 &
# wait for the first eval to finish, then:
printf -- '- vars:\n    word: bye\n' >> tests.yaml
sleep 15
grep -c "Watching for file changes" watch.log   # 0  (chokidar ready never fired)
grep -c "Duration:" watch.log                   # 1  (no re-run)
```

The process stays alive but the `.on("ready")` log line ("Watching for file changes on ...") never appears and file changes (append, in-place rewrite, or touch) never trigger a re-run. Same behavior on 0.121.11. On Linux, 0.121.11 does print the "Watching for file changes" line (for the config file), so this part appears platform-specific.

Also worth noting while debugging this: chokidar's `change`-only handler means an atomic replace (`write tmp + rename`) of a watched file emits `unlink`+`add` and never re-runs; if intentional, a docs note that watched files must be modified in place would save users some time.

## Environment

[...]

PR #10169, opened 2026-07-21
fix(eval): watch mode ignored scalar and generator `tests` references

WatchTree-19:
Refs #10164 (part 1). Thanks @algebananazzzzz — the code pointer in the issue was exact.

## The reported bug

Watch paths for tests were collected behind `Array.isArray(config.tests)`. But `tests` is a union of three shapes (`TestSuiteConfig`): a bare file reference, a test-generator object, or an array of either plus inline cases. Only the array form was collected, so:

```yaml
tests: file://cases.yaml
```

— valid config that evaluates fine — was silently excluded. Only the config file got watched, and edits to `cases.yaml` never triggered a re-run, with no warning.

## Two more gaps in the same expression

While confirming the above I found the problem is a bit wider than reported:

1. **A top-level generator object** (`tests: {path: file://gen.py:fn}`) also fails the `Array.isArray` guard and is unwatched.
2. **Generator entries inside the array form are dropped too.** The branch matched only bare strings and entries carrying `vars`; a `{path: ...}` entry has neither — even though the array is the documented shape for generators. So the array form isn't fully safe either.

There's also a subtler one: generator references carry a `:functionName` suffix that isn't part of the filename. The old array branch resolved `file://tests.py:generate` verbatim, handing chokidar `/abs/tests.py:generate` — a path that cannot exist. `resolveTestFilePath` now strips it using the same last-colon rule as `getStandaloneTestsFileMetadata`, including its guard against mistaking a Windows drive letter for a function separator.

## Tests

Five cases added to `test/commands/eval.test.ts`, using the existing chokidar mock and asserting on what's actually passed to `chokidar.watch`. Without the fix, three fail:

```
expected [ 'config.yaml' ] to include '/…/cases.yaml'
expected [ 'config.yaml' ] to include '/…/tests.py'
expected [ 'config.yaml' ] to include '/…/gen.py'
```

The other two pin existing behaviour that must not regress (array form + `vars` files, and `tests` absent). Full file: **121 passed**. `biome check` clean on both files — the one remaining warning (`noExcessiveCognitiveComplexity` on `doEval`) is pre-existing and unrelated; it reports at the same site before and after, just shifted by the added lines.

## Not addressed

**The macOS half of #10164.** The issue also reports that on macOS the watcher never fires `ready` and no re-runs happen at all. I can't reproduce that on Linux and don't have a macOS box, so I haven't touched it — that part of the issue should stay open. Worth noting the two are independent: even with the readiness problem fixed, the scalar form would still have been unwatched on every platform.

I also left the xlsx `#sheet` specifier alone: `file://data.xlsx#Sheet1` still resolves to a path with the fragment attached. `getStandaloneTestsFileMetadata` deliberately preserves `#` outside Excel basenames, so stripping it here felt like guessing at intent rather than fixing a known bug. Happy to add it if you'd like.


WatchTree-19:
Rewritten. The earlier version of this branch addressed the wrong layer, and the review comments below were right about that — thanks, they were correct and specific.

## What was wrong with the first attempt

The watch-path builder read `config.tests` *after* `resolveConfigs()`. By that point `combineConfigs()` has already expanded a scalar reference (`tests: file://cases.yaml`) and a generator object into concrete test cases:

```js
// src/util/config/load.ts
if (typeof config.tests === 'string') {
  const newTests = await readTests(config.tests, path.dirname(configPath));
  tests.push(...newTests);
} else if (config.tests && 'path' in config.tests) {
  const newTests = await readTests(config.tests, path.dirname(configPath));
  tests.push(...newTests);
}
```

So on the explicit `-c/--config` path the reference was gone before the watcher looked, and the fix only worked on the auto-discovered-config path. That is the substance of the "Preserve raw scalar tests when watching explicit configs" comment.

## What this version does

**Two sources.** The array form survives `combineConfigs()` untouched, so inline cases and their `vars` references are still read from the resolved config. Scalar and generator forms are recovered by re-reading each config file with `maybeReadConfig()`, which returns the raw `tests` value before expansion.

**Resolution moves next to the loader.** It now lives in `testCaseReader.ts` as `resolveTestsWatchPaths()`, beside `readTests()`/`loadTestsFromGlob()` whose rules it has to match. Duplicating those rules in `doEval.ts` is precisely why the watcher drifted from the loader, so keeping them apart seemed likely to reintroduce the same class of bug.

## Review comments addressed

| Comment | Handling |
|---|---|
| Preserve raw scalar tests when watching explicit configs | Raw config re-read via `maybeReadConfig()` |
| Watch files referenced by generator config | `file://` refs inside `config` collected recursively |
| Preserve colons in non-generator watched paths | `:functionName` stripped only for JS/py references |
| Strip Excel sheet selectors before watching | `#Sheet` stripped, xlsx/xls basenames only |
| Expand scalar test globs before watching | Same `globSync()` call the loader uses |
| Resolve CLI test files from the working directory | `cmdObj.tests` resolves against cwd, matching `resolveConfigs` |

Also: remote references (`http`, `https`, `az://`, `huggingface`) are skipped, and a reference matching nothing yet returns its literal path so creating the file later still triggers a rerun.

## Tests

New `test/util/testCaseWatchPaths.test.ts` — 11 cases against a real temp directory, one per rule above.

The branch's earlier `doEval` tests were replaced. They mocked `resolveConfigs` with an unexpanded config, which meant they passed without ever exercising the path that was broken. The new ones mock the expanded and raw configs separately. Three fail on `main`:

```
recovers a scalar reference that combineConfigs already expanded
   → expected [ 'config.yaml' ] to include '/…/cases.yaml'
recovers a generator object that combineConfigs already expanded
   → expected [ 'config.yaml' ] to include '/…/gen.py'
watches command-line tests relative to the working directory
   → expected [ 'config.yaml' ] to include '/…/cases.csv'
```

121 tests pass in `test/commands/eval.test.ts`, 11 in the new file, `tsc --noEmit` clean on the touched files.

One note: while wiring this up I briefly broke the existing `should watch config-adjacent prompt, provider, and var files` test by reading *only* the raw config. That surfaced the fact that the array form does survive `combineConfigs()`, which is why both sources are read rather than one.

WatchTree-19:
Addressed the four points from the last round. One of them was a regression the previous commit introduced, so thanks for catching it.

**Re-executing configs.** `maybeReadConfig()` goes through `readConfig()`, which loads a `.js` config via `importModule()`. That falls back to `vm.runInContext()` for CommonJS and is not cached, so watch mode was executing the user's config a second time and any side effect in it happened twice. That is worse than the gap it was filling.

Only YAML and JSON are read now, directly and for their `tests` field alone. Executable configs keep the coverage they had before this PR, which is the config file itself. Losing scalar recovery for `.js` configs is the right trade against running someone's code twice, but say the word if you'd rather have it back some other way.

**`--config` globs.** `combineConfigs()` expands them; I was handing the literal wildcard to the reader, which returned nothing, so recovery silently did not happen at all for glob configs. Expanded with `globSync` first now.

**Nested references inside a tests file.** `cases.yaml` holding `vars: {data: file://vars.yaml}` means the loader reads `vars.yaml` too. Declarative tests files are parsed and their nested `file://` references collected. Best effort: a malformed file is left for the loader to report rather than failing watch setup.

**Glob membership.** Resolving to current matches missed files added later. The deepest wildcard-free parent is watched alongside the matches, and on its own when a pattern matches nothing yet.

**Tests.** Six new cases fail on the previous commit:

```
watches the stable parent of a glob so later additions are seen
watches the stable parent when a glob matches nothing yet
watches file references nested inside a tests file
recovers a scalar reference that combineConfigs already expanded
recovers a generator object that combineConfigs already expanded
expands a --config glob before recovering raw tests
```

The executable-config case passes either way, since the old code called a different function. It is there to stop the re-execution being reintroduced, not to reproduce it.

138 tests pass across `test/commands/eval.test.ts` and `test/util/testCaseWatchPaths.test.ts`. `tsc --noEmit` and `biome ci` clean on the touched files, with the two pre-existing complexity warnings unchanged from `main`.

WatchTree-19:
All three addressed.

**Normalisation.** You're right that the hand-rolled parse skipped `readConfig()`'s preprocessing, so `tests: {$ref: ...}` and env-templated paths recovered the unresolved value. Looking at it again, the reason I stopped calling `readConfig()` was narrower than the change I made: it only reaches `importModule()` through its `isJavascriptFile()` branch. The YAML/JSON branch reads the file, dereferences `$ref` and renders env templates, and executes nothing.

So `maybeReadConfig()` is back, gated on the extension. That keeps the normalisation and still guarantees a config is never executed twice, which is better than either of my previous two attempts.

**`--tests` override.** Fixed, and this was the one that mattered — recovery is skipped entirely when tests come from the CLI. Watching config test sources that `resolveConfigs()` had already discarded meant an edit to an irrelevant file triggered a full evaluation and the provider calls that go with it.

**Scalar vars.** `{ vars: 'vars/*.yaml' }` is now handled alongside the mapping form. Worth noting it carries no `file://` scheme, so it resolves as written.

**Tests.** Three added. Two fail on the previous commit, the `--tests` one with `maybeReadConfig` called 9 times where it should not be called at all.

141 tests pass across `test/commands/eval.test.ts` and `test/util/testCaseWatchPaths.test.ts`. `tsc --noEmit` and `biome ci` clean on the touched files.

**One consequence I'd flag rather than hide:** an invalid declarative config now logs its validation warning twice during watch startup, once from `resolveConfigs()` and once here. Watching the wrong file seemed the worse of the two, but it's a real cost and I'm happy to suppress it if you'd prefer.

Also, on scope: I've been working from the automated review, which has been accurate and useful. Before I keep iterating, is a maintainer able to say whether this PR is the right shape? It started as a small watch-path fix and the surface has grown as each layer of the loader turned out to matter. If you'd rather it were narrower, or split, I'd sooner know now than after another round.

mldangelo:
Reviewed this and pushed two fixes plus a `main` merge to your branch. The core diagnosis here is right and the research quality is high — the `combineConfigs` timing analysis and the `importModule` double-execution hazard are both correct and non-obvious, and the re-read approach is the only way to recover a reference that `combineConfigs` has already expanded.

Two things needed changing before this can land.

### 1. Watching a glob's parent directory (removed)

`resolveTestsFileReference` added `globParentDirectory(...)` to the watch list for any glob. chokidar watches a directory **recursively**, and `doEval` reruns the whole evaluation on any `change` beneath it, so `tests: file://cases/*.yaml` reran the eval — and its provider calls — on every unrelated edit in that tree. `ignored: /^\./` doesn't help: it's matched against absolute paths, which never start with a dot, so `node_modules` and `.git` were walked too.

The worst shape is a pattern anchored at the config directory (`tests: file://*.csv` → `globParentDirectory` returns the config dir). That directory then contains the run's own `-o` output file, so each write retriggers the watcher — an unbounded rerun loop with real API spend.

Reproduced against the real watcher, unrelated file, no glob subtlety:

```
# before
Watching for file changes on e2e2/cases/a.yaml ...
Watching for file changes on e2e2/cases ...        <-- directory
TOUCHING e2e2/cases/notes.txt
File change detected: e2e2/cases/notes.txt         <-- full rerun

# after
Watching for file changes on e2e2/cases/a.yaml ...
TOUCHING e2e2/cases/notes.txt
(no rerun)
```

It also didn't buy what the comment claimed. A file that appears later emits `add`, and the watcher only registers `.on('change')`, so the additions this was meant to catch never triggered a rerun:

```
READY
change  .../results.json                 <-- the run's own output
change  .../node_modules/pkg/index.js    <-- not ignored
add     .../sub/newfile.yaml             <-- 'add', never handled
```

Both of the tests that pinned this behaviour are replaced by a regression test asserting the parent is *not* watched, including the config-directory-anchored shapes.

### 2. `--vars` needs the same guard as `--tests`

`-v, --vars` is a documented alias for `--tests`, and `resolveConfigs` treats it identically (`cmdObj.tests || cmdObj.vars` in `util/config/load.ts`). The guard only checked `cmdObj.tests`, so `promptfoo eval --watch -v cases.csv` still re-read the config and watched the `tests:` sources it declares — files with no bearing on the run — which is exactly the case your `--tests` guard exists to prevent.

Collapsing the two command-line branches also drops a redundant resolution: with `--tests`, `config.tests` already holds the CLI value, so it was resolved twice, once against the config directory instead of the working directory, putting an impossible path in the watch list. The differing base paths between the two flags are preserved.

### Smaller things folded in

- **Cycle guard.** A YAML anchor produces a self-referential object; `collectConfigFileReferences` recursed until the stack overflowed. `collectNestedFileReferences` caught it, but the generator-`config` call site didn't, so it crashed a run that had already evaluated successfully.
- **`globSync` only for actual glob patterns**, falling back to the literal path when a pattern matches nothing — so a real file named `report[1].csv` is watched as written instead of resolving to its whole directory.
- Merged the `vars` string/array branches, dropped a dead `.filter(Boolean)`, fixed two misplaced doc comments (the `resolveTestsFileReference` docblock sat on `hasGlobMagic`; `collectNestedFileReferences` carried two).

Net `-30` lines in `testCaseReader.ts`. All four new/changed tests fail against the previous revision.

### Not changed, but worth knowing

**The macOS half of #10164 has a concrete root cause, and it's not in the watcher.** `main()` resolves as soon as the eval action handler returns, and `shutdownGracefully()` then runs unconditionally and schedules `process.exit()` 100ms later (`src/mainUtils.ts:270-273`). In watch mode chokidar holds the loop open, that timer fires, and the process is killed before `ready` — so `promptfoo eval --watch` exits immediately and never re-runs. This reproduces on unmodified `main`, so it's pre-existing and out of scope here, but it does mean this fix has no visible effect through the shipped CLI until that's addressed. I verified this PR's behaviour by driving `doEval` directly, bypassing `main.ts`.

That's worth its own issue/PR; leaving #10164 open for it is right, and your read that the two halves are independent is correct.

Also merged current `main` in (branch was 63 commits behind, and the lockfile's Biome pin had drifted). `tsc`, `biome ci`, and `architecture:check` are clean; the two `noExcessiveCognitiveComplexity` warnings are pre-existing at the same sites on `main`.

mldangelo:
Follow-up on the watch-mode-never-runs half of #10164: opened #10562. Root cause confirmed — `shutdownGracefully()` closes the DB and HTTP dispatcher and `process.exit()`s 100ms after the eval handler returns, so the watcher was being killed before it could report anything. Not macOS-specific and not a chokidar problem (a `persistent: true` watcher reaches `ready` in ~1ms and holds the loop open by itself).

The two are independent and either can land first — this PR fixes *which* files get watched, #10562 fixes watch mode running at all. Together they close #10164.

PR #10171, opened 2026-07-21
feat(websocket): support URL templating

jameshiester-oai:
## Summary

- Render the generic WebSocket provider URL with per-call Nunjucks variables.
- Add regression coverage proving one provider instance opens distinct session URLs across calls.
- Document URL templating and a `{{ sessionId }}` example.

## Why

The WebSocket provider already rendered `messageTemplate` per request, but passed its URL unchanged to every new connection. That prevented stateful targets from placing a generated session or conversation ID in the WebSocket URL, even though each API call opens a fresh connection.

## Impact

WebSocket targets can now use test variables such as `{{ sessionId }}` in `config.url`, matching the per-request templating behavior users expect from the HTTP provider. Existing static URLs behave unchanged.

## Validation

- `npx vitest run test/providers/websocket.test.ts --sequence.shuffle=false`
- `npm run tsc`
- `npm run f`
- `npm run l`
- `SKIP_OG_GENERATION=true npx docusaurus build --out-dir /private/tmp/promptfoo-docs-build-websocket-url-templating`

### DIFF
--- a/src/providers/websocket.ts
+++ b/src/providers/websocket.ts
@@ -7,6 +7,7 @@ import logger from '../logger';
 import invariant from '../util/invariant';
 import { safeJsonStringify } from '../util/json';
 import { getProcessShim } from '../util/processShim';
+import { sanitizeUrlForLogging } from '../util/sanitizer';
 import { getNunjucksEngine } from '../util/templates';
 import { getRequestTimeoutMs } from './shared';
 import { normalizeResponseTransformResult } from './transformResult';
@@ -244,7 +245,7 @@ export class WebSocketProvider implements ApiProvider {
     const message = nunjucks.renderString(this.config.messageTemplate, vars);
     const streamResponse = this.streamResponse == null ? undefined : await this.streamResponse;
 
-    logger.debug(`Sending WebSocket message to ${url}: ${message}`);
+    logger.debug(`Sending WebSocket message to ${sanitizeUrlForLogging(url)}: ${message}`);
     let accumulator: ProviderResponse = { error: 'unknown error occurred' };
     return new Promise<ProviderResponse>((resolve, reject) => {
       const wsOptions: ClientOptions = {};
--- a/src/providers/websocket.ts
+++ b/src/providers/websocket.ts
@@ -7,6 +7,7 @@ import logger from '../logger';
 import invariant from '../util/invariant';
 import { safeJsonStringify } from '../util/json';
 import { getProcessShim } from '../util/processShim';
+import { REDACTED, sanitizeObject, sanitizeUrl } from '../util/sanitizer';
 import { getNunjucksEngine } from '../util/templates';
 import { getRequestTimeoutMs } from './shared';
 import { normalizeResponseTransformResult } from './transformResult';
@@ -33,24 +34,13 @@ function normalizeWebSocketProtocols(protocols: string | string[] | undefined): 
     .filter(Boolean);
 }
 
-function getWebSocketErrorMessage(err: WebSocket.ErrorEvent | Error | unknown): string {
-  const error =
-    err && typeof err === 'object' && 'error' in err ? (err as WebSocket.ErrorEvent).error : err;
-
-  if (typeof error === 'string' || typeof error === 'number' || typeof error === 'boolean') {
-    return String(error);
-  }
-
-  if (error instanceof Error) {
-    return error.message;
-  }
-
-  if (error && typeof error === 'object' && 'message' in error) {
-    return String((error as { message: unknown }).message);
-  }
-
-  const serialized = safeJsonStringify(error);
-  return serialized && serialized !== '{}' ? serialized : 'unknown WebSocket error';
+function getWebSocketProviderId(url: string): string {
+  const sanitizedUrl = sanitizeUrl(url);
+  return sanitizedUrl === REDACTED ||
+    sanitizedUrl.includes(encodeURIComponent(REDACTED)) ||
+    sanitizedUrl.includes('***')
+    ? sanitizedUrl
+    : url;
 }
 
 interface WebSocketProviderConfig {
@@ -196,6 +186,7 @@ export async function createStreamResponse(
 
 export class WebSocketProvider implements ApiProvider {
   url: string;
+  private readonly providerId: string;
   config: WebSocketProviderConfig;
   timeoutMs: number;
   transformResponse: (data: any) => ProviderResponse;
@@ -210,6 +201,7 @@ export class WebSocketProvider implements ApiProvider {
   constructor(url: string, options: ProviderOptions) {
     this.config = options.config as WebSocketProviderConfig;
     this.url = this.config.url || url;
+    this.providerId = getWebSocketProviderId(this.url);
     this.timeoutMs = this.config.timeoutMs || getRequestTimeoutMs();
     this.transformResponse = createTransformResponse(
       this.config.transformResponse || this.config.responseParser,
@@ -219,18 +211,18 @@ export class WebSocketProvider implements ApiProvider {
       : undefined;
     invariant(
       this.config.messageTemplate,
-      `Expected WebSocket provider ${this.url} to have a config containing {messageTemplate}, but got ${safeJsonStringify(
-        this.config,
+      `Expected WebSocket provider ${this.providerId} to have a config containing {messageTemplate}, but got ${safeJsonStringify(
+        sanitizeObject(this.config, { context: 'provider config' }),
       )}`,
     );
   }
 
   id(): string {
-    return this.url;
+    return this.providerId;
   }
 
   toString(): string {
-    return `[WebSocket Provider ${this.url}]`;
+    return `[WebSocket Provider ${this.providerId}]`;
   }
 
   async callApi(prompt: string, context?: CallApiContextParams): Promise<ProviderResponse> {
@@ -251,16 +243,16 @@ export class WebSocketProvider implements ApiProvider {
       if (this.config.headers) {
         wsOptions.headers = this.config.headers;
       }
-      let ws: WebSocket;
       try {
-        ws =
-          protocols.length > 0
-            ? new WebSocket(url, protocols, wsOptions)
-            : new WebSocket(url, wsOptions);
+        new URL(url);
       } catch {
         reject(new Error('Failed to create WebSocket connection'));
         return;
       }
+      const ws =
+        protocols.length > 0
+          ? new WebSocket(url, protocols, wsOptions)
+          : new WebSocket(url, wsOptions);
       const timeout = setTimeout(() => {
         ws.close();
         logger.error(`[WebSocket Provider] Request timed out`);
@@ -332,12 +324,11 @@ export class WebSocketProvider implements ApiProvider {
         }
       };
 
-      ws.onerror = (err) => {
+      ws.onerror = () => {
         clearTimeout(timeout);
         ws.close();
-        const errorMessage = getWebSocketErrorMessage(err);
-        logger.error(`[WebSocket Provider] Error: ${errorMessage}`);
-        reject(new Error(`WebSocket error: ${errorMessage}`));
+        logger.error(`[WebSocket Provider] Connection failed`);
+        reject(new Error('WebSocket connection failed'));
       };
 
       ws.onopen = () => {
--- a/src/providers/websocket.ts
+++ b/src/providers/websocket.ts
@@ -7,7 +7,6 @@ import logger from '../logger';
 import invariant from '../util/invariant';
 import { safeJsonStringify } from '../util/json';
 import { getProcessShim } from '../util/processShim';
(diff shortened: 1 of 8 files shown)


mldangelo-oai (commented):
Reviewed the current head and reproduced two additional issues in the new per-call WebSocket URL path. The focused WebSocket and root test-hygiene suites pass on this head; inline details follow.

src/providers/websocket.ts
```diff
@@ -253,8 +255,8 @@ export class WebSocketProvider implements ApiProvider {
       }
       const ws =
         protocols.length > 0
-          ? new WebSocket(this.url, protocols, wsOptions)
-          : new WebSocket(this.url, wsOptions);
+          ? new WebSocket(url, protocols, wsOptions)
+          : new WebSocket(url, wsOptions);
```
mldangelo-oai, line 246: **[P1] Sanitize synchronous WebSocket-constructor errors**

`ws` throws `SyntaxError("Invalid URL: ${address}")` synchronously, so a malformed rendered variable leaks every resolved credential in the URL through the rejected provider error. For example, `url: "wss://{{ host }}/ws?token={{ token }}"` with `host: "[invalid-host"` and `token: "runtime-secret"` rejects with `Invalid URL: wss://[invalid-host/ws?token=runtime-secret`; the new debug-log sanitizer does not cover this error path, and provider errors can be logged or persisted. Catch constructor/URL-validation failures and replace them with an error containing only the sanitized URL (without retaining the raw error as a logged cause), and add a regression for an invalid rendered host plus token.

[`websocket.ts` at the reviewed head](https://github.com/promptfoo/promptfoo/blob/c72b2320e0d2f543ff1ab7ab2502d0f78b54d4eb/src/providers/websocket.ts#L256-L259)

src/providers/websocket.ts
```diff
@@ -240,10 +241,11 @@ export class WebSocketProvider implements ApiProvider {
       ...(context?.vars || {}),
       prompt,
     };
+    const url = nunjucks.renderString(this.url, vars);
```
mldangelo-oai, line 225: **[P2] Honor configured Nunjucks filters when rendering the URL**

The evaluator passes configured `nunjucksFilters` into `context.filters`, but this uses the module-scoped engine created with no filters. Any newly supported URL that applies a custom filter fails before opening a connection: `url: "wss://example.test/ws/{{ sessionId | slugify }}"` with a valid `slugify` filter and `sessionId: "Conversation A"` rejects with `Error: filter not found: slugify`. Create the per-call engine with `getNunjucksEngine(context?.filters)` (and use it consistently for the message template) and cover a custom-filter URL in the regression suite.

[`websocket.ts` at the reviewed head](https://github.com/promptfoo/promptfoo/blob/c72b2320e0d2f543ff1ab7ab2502d0f78b54d4eb/src/providers/websocket.ts#L239-L245)

src/providers/websocket.ts
```diff
@@ -240,10 +241,11 @@ export class WebSocketProvider implements ApiProvider {
...
     const message = nunjucks.renderString(this.config.messageTemplate, vars);
     const streamResponse = this.streamResponse == null ? undefined : await this.streamResponse;
 
-    logger.debug(`Sending WebSocket message to ${this.url}: ${message}`);
+    logger.debug(`Sending WebSocket message to ${sanitizeUrlForLogging(url)}: ${message}`);
```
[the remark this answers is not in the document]
mldangelo-oai: Additional reproducible bypass on the current head: if any interpolated value itself contains template markers, `sanitizeUrlForLogging` treats the rendered URL as an unresolved template and leaves even a normally redacted query credential intact. With `url: "ws://127.0.0.1:1/ws/{{ sessionId }}?token={{ token }}"`, `sessionId: "{{ attacker_controlled }}"`, and `token: "runtime-secret"`, the emitted debug line contains `?token=runtime-secret`. Avoid logging the rendered URL (or fail closed when rendered values leave `{{…}}` markers) and include this case in the redaction regression.

src/providers/websocket.ts
```diff
@@ -240,21 +238,29 @@ export class WebSocketProvider implements ApiProvider {
       ...(context?.vars || {}),
       prompt,
     };
+    const nunjucks = getNunjucksEngine(context?.filters);
+    const url = nunjucks.renderString(this.url, vars);
```
mldangelo-oai, line 225: **[P1] Keep partially rendered URL credentials out of provider identity/logs**

A real eval with `url: "ws://127.0.0.1:18761/sessions/{{ sessionId }}?token={{ env.WS_TOKEN }}"` and `WS_TOKEN=runtime-secret` logs `Provider type: ws://127.0.0.1:18761/sessions/{{ sessionId }}?token=runtime-secret`, logs the same value in the scheduler metrics, and persists it as the provider id/config URL in the exported results. `renderEnvOnly` resolves the environment token while retaining the runtime placeholder, `id()` returns that partially rendered URL, and the normal URL sanitizer intentionally skips any string still containing `{{…}}`; removing the debug line below therefore does not close the credential path.

Keep an unrendered/sanitized identifier for provider identity and logging (or make the shared sanitizer redact query credentials even when another placeholder remains), and add an evaluator-level regression with an env token plus a runtime session placeholder. This is reproducible without an invalid URL or a failing connection.

[Reviewed lines](https://github.com/promptfoo/promptfoo/blob/aaeaa2be18fae3706b761bfcca8c6d88b4991a38/src/providers/websocket.ts#L239-L246)

src/providers/websocket.ts
```diff
@@ -240,21 +238,29 @@ export class WebSocketProvider implements ApiProvider {
...
       }
-      const ws =
-        protocols.length > 0
-          ? new WebSocket(this.url, protocols, wsOptions)
-          : new WebSocket(this.url, wsOptions);
+      let ws: WebSocket;
+      try {
+        ws =
+          protocols.length > 0
+            ? new WebSocket(url, protocols, wsOptions)
```
mldangelo-oai: **[P1] Sanitize asynchronous connection errors from rendered hosts**

Once the rendered URL reaches `ws`, failures that happen after construction can still echo URL-derived values through `onerror`. A real CLI eval with `url: "ws://{{ tenant }}.invalid/ws?token={{ token }}"`, `tenant: runtime-secret-tenant`, and `token: runtime-query-secret` logs and exports `WebSocket error: getaddrinfo ENOTFOUND runtime-secret-tenant.invalid` (including the provider stack). The new constructor catch never runs, and the existing `getWebSocketErrorMessage(err)` path propagates the rendered hostname verbatim.

Treat connection-error text as URL-derived/untrusted: preserve a safe error class/status or sanitize host/credential values before logging and rejecting, and cover an asynchronous `ENOTFOUND`/connection-failure case with a templated tenant/host. The current logger test only exercises successful construction.

[Reviewed lines](https://github.com/promptfoo/promptfoo/blob/aaeaa2be18fae3706b761bfcca8c6d88b4991a38/src/providers/websocket.ts#L253-L263)

src/providers/websocket.ts
```diff
@@ -240,21 +238,29 @@ export class WebSocketProvider implements ApiProvider {
...
+            : new WebSocket(url, wsOptions);
+      } catch {
```
mldangelo-oai, line 239: **[P2] Do not discard safe constructor diagnostics**

This catch converts *every* synchronous `ws` configuration error into `Failed to create WebSocket connection`. I reproduced an existing supported configuration with `protocols: ["invalid protocol"]`: `ws` provides the actionable `An invalid or duplicated subprotocol was specified`, but the CLI/result now expose only the generic message. Duplicate protocols, unsupported schemes/fragments, and other safe constructor validation failures become similarly opaque.

Narrow the redaction boundary to URL parsing—for example, parse the rendered URL inside a small try/catch and pass the resulting `URL` object to `new WebSocket`—so malformed rendered addresses cannot leak while unrelated constructor errors retain their useful, safe reason. Please add an invalid/duplicate-protocol regression alongside the malformed-URL case.

[Reviewed lines](https://github.com/promptfoo/promptfoo/blob/aaeaa2be18fae3706b761bfcca8c6d88b4991a38/src/providers/websocket.ts#L253-L263)

test/providers/websocket.test.ts
```diff
@@ -1,19 +1,25 @@
...
-  });
+  };
+  const WebSocketMock = vi.fn(createWebSocket);
 
   const setFactory = (nextFactory: () => Mocked<WebSocket>) => {
     factory = nextFactory;
   };
+  const reset = () => {
+    WebSocketMock.mockReset();
+    WebSocketMock.mockImplementation(createWebSocket);
```
mldangelo-oai: **[simplification] `mockReset()` already restores this implementation**

For Vitest 4, resetting a mock created with `vi.fn(impl)` restores `impl`; I verified that behavior with the installed version. The immediate `mockImplementation(createWebSocket)` is therefore redundant, as are the new `createWebSocket`/custom-`reset` wrappers. The original hoisted `vi.fn(function () { return factory?.() ?? ({} as Mocked<WebSocket>); })` plus `WebSocketMock.mockReset()` in `beforeEach` is sufficient, keeps test-hygiene happy, and removes state-management complexity from the diff.

mldangelo-oai (changes requested):
Re-reviewed `aaeaa2be18fae3706b761bfcca8c6d88b4991a38` carefully, including the earlier findings, the full changed-file diff, focused and randomized tests, type/lint/format checks, the live CI matrix, and real CLI evals against a local WebSocket server.

The per-call URL rendering and custom-filter path now work end to end, the old direct debug-URL leak is removed, and the malformed-URL constructor case no longer emits the raw URL. I did find three remaining issues that need attention:

1. **P1 — mixed environment/runtime URL templates still expose credentials via provider identity, evaluator/scheduler debug logs, and exported results.** Environment rendering resolves the token before evaluation while leaving the runtime placeholder; the URL sanitizer skips strings that still contain `{{…}}`, so a configured token is emitted verbatim.
2. **P1 — asynchronous connection errors can disclose rendered host/session values.** An `ENOTFOUND` repro records the runtime tenant hostname in error logs and the exported provider error/stack, bypassing the constructor-only protection.
3. **P2 — the catch-all constructor handler discards useful, safe configuration errors.** An invalid subprotocol that previously surfaced “An invalid or duplicated subprotocol was specified” now produces only “Failed to create WebSocket connection”, making supported configuration failures unnecessarily hard to diagnose.

I left tight inline repros and suggested narrower fixes/tests. There is also a small test-mock simplification below: Vitest's `mockReset()` already restores a `vi.fn(impl)` implementation, so the extra reset wrapper/reimplementation is redundant. Finally, the previously raised docs contradiction still applies: the reference says `url` is required while the documented full-URL provider-`id` example and option table explicitly allow it to be omitted.

QA completed on the exact head:

- 283 focused provider/registry tests passed; 88 test-hygiene tests passed.
- WebSocket tests passed with two shuffled seeds.
- TypeScript, Biome, Prettier, and `git diff --check` passed (Biome reports only the pre-existing message-handler complexity warning).
- Real CLI eval with two concurrent, independently templated session paths and a configured `slugify` filter: **2 passed, 0 failed, 0 errors**.
- Negative CLI evals independently reproduced the mixed-template credential leak, async DNS tenant leak, and invalid-protocol diagnostic regression.
- All current GitHub checks are green; the branch is mergeable but behind base.

Please add end-to-end regressions for the two leak paths, not only logger spies around `callApi`: the provider-ID/evaluator and asynchronous-error paths are where these escape.

src/providers/websocket.ts
```diff
@@ -35,24 +34,13 @@ function normalizeWebSocketProtocols(protocols: string | string[] | undefined):
...
-
-  const serialized = safeJsonStringify(error);
-  return serialized && serialized !== '{}' ? serialized : 'unknown WebSocket error';
+function getWebSocketProviderId(url: string): string {
+  const sanitizedUrl = sanitizeUrl(url);
+  return sanitizedUrl === REDACTED ||
+    sanitizedUrl.includes(encodeURIComponent(REDACTED)) ||
+    sanitizedUrl.includes('***')
+    ? sanitizedUrl
+    : url;
```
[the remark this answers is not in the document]
jameshiester-oai: This is unlikely, so I'm ignoring this comment intentionally

src/util/sanitizer.ts
```diff
@@ -805,28 +805,38 @@ function getSecretLookingRawQueryKeys(search: string): Set<string> {
   return secretKeys;
 }
 
+function sanitizeTemplatedUrl(url: string): string {
+  // A template may coexist with an already-rendered env credential. Avoid URL
+  // parsing here because it encodes the remaining Nunjucks syntax, but still
+  // scrub literal query and fragment credentials before the value is logged or
+  // persisted.
+  if (hasUrlUserinfoPassword(url)) {
+    return REDACTED;
```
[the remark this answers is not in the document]
jameshiester-oai: This would be very brittle so I'm ignoring it.

mldangelo-oai (approved):

PR #10176, opened 2026-07-22
fix(providers): restore secret-safe WebSocket retries

wholley-oai:
## Summary

Promptfoo's WebSocket provider opens a fresh connection for each evaluation call. The recent URL-templating change correctly stopped asynchronous connection errors from exposing rendered hostnames, session IDs, and query credentials, but replacing every underlying failure with `WebSocket connection failed` also removed the HTTP status and transport signals that Promptfoo's evaluator scheduler uses to retry transient failures.

- Preserve only trusted, secret-safe WebSocket handshake statuses and transport classifications, including HTTP 429/502/503/504, `ECONNRESET`, `ECONNREFUSED`, `EPIPE`, `ETIMEDOUT`, safe `EPROTO`, socket hang-ups, and transient TLS record failures.
- Keep rendered URLs, hostnames, session IDs, query credentials, and raw error causes out of provider errors and logs.
- Continue failing immediately for permanent DNS, authentication, TLS/configuration, and cancellation errors, including protocol-version mismatches and adversarial hostnames containing retry-looking strings such as `429`, `network`, or `ECONNRESET`; preserve `AbortError` and `AbortException` identities without leaking their raw messages.
- Add composed `WebSocketProvider` → `RateLimitRegistry` regressions that assert configured retry counts, rate-limit metrics, secret redaction, and permanent/cancellation behavior together.

## Why

WebSocket-backed evaluations pass through `RateLimitRegistry`, so discarding safe error classification made transient network failures and HTTP 429 handshake responses stop after a single attempt instead of honoring configured retries and backoff. This restores the existing scheduler contract without reintroducing the URL-derived credential exposure addressed by #10171.

## Validation

- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci` with Node `24.18.0` and npm `11.16.0`.
- `npx vitest run test/providers/websocket.test.ts test/scheduler test/providers/http.test.ts test/providers/providerLogging.test.ts test/util/fetch/errors.test.ts test/util/sanitizer.test.ts test/models/evalResult.test.ts` — **725 tests passed across 14 files**.
- `npm run tsc`.
- `npm run architecture:check`.
- `npm run l && npm run f`; both changed files were actually processed.
- Real local `npm run local -- eval -c <loopback-config> --no-cache --no-write --no-share --no-table -o <results.json>` against an HTTP-upgrade server: an initial 429 recovers on attempt two; repeated 429s stop after the configured retry and export only `WebSocket connection failed (HTTP 429)`.
- Three independent read-only review passes covered contract/security breakage, runtime/test quality, and simplification/documentation; confirmed findings were fixed and revalidated.

### DIFF
--- a/src/providers/websocket.ts
+++ b/src/providers/websocket.ts
@@ -4,6 +4,7 @@ import WebSocket, { type ClientOptions } from 'ws';
 import cliState from '../cliState';
 import { importModule } from '../esm';
 import logger from '../logger';
+import { isAbortError, isTransientConnectionError } from '../util/fetch/errors';
 import invariant from '../util/invariant';
 import { safeJsonStringify } from '../util/json';
 import { getProcessShim } from '../util/processShim';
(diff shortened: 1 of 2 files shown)


jameshiester-oai (approved):
This is a good change

PR #10177, opened 2026-07-22
fix(webui): accept templated websocket provider URLs

wholley-oai:
## Summary
WebSocket providers render Nunjucks URL templates for every evaluation, but the shared evaluation and red-team target editor rejected valid templates before users could save or continue. This restores parity between the Web UI and the existing backend provider without changing HTTP or A2A validation.

- Accept templated WebSocket schemes, complete URLs, hosts, ports, paths, IPv6 authorities, conditionals, filters, and standard Nunjucks whitespace-control syntax.
- Preserve WebSocket-only protocol checks; reject malformed templates, invalid static protocols, internal-placeholder collisions, and unbounded conditional expansion.
- Exercise the actual red-team Next action and evaluation Save dialog, plus real local WebSocket connections for supported URL-template shapes.

## Validation
- PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci on Node 24.18.0 and npm 11.16.0.
- npm run test:app -- src/pages/redteam/setup/components/Targets/ProviderConfigEditor.test.tsx src/pages/redteam/setup/components/Targets/WebSocketEndpointConfiguration.test.tsx src/pages/redteam/setup/components/Targets/ProviderEditor.test.tsx src/pages/redteam/setup/components/Targets/TargetConfiguration.test.tsx src/pages/redteam/setup/components/Targets/index.test.tsx src/pages/eval-creator/components/AddProviderDialog.test.tsx --run: 159 tests passed.
- npm exec -- vitest run test/providers/websocket.test.ts test/providers/websocket.loopback.test.ts: 37 tests passed.
- npm run tsc and npm --prefix src/app run tsc -- --build --noEmit --incremental.
- npm run l and npm run f; targeted Biome checks processed all three changed files.
- Real local CLI evaluation with templated scheme and authority: exported result passed with score 1, no errors, expected request path, and expected WebSocket response.
- Three independent read-only contract, runtime, and simplification review passes found no remaining actionable issues.

## Context
The backend URL-templating support landed in #10171, but its shared Web UI validation did not recognize the same URL forms.

### DIFF
--- a/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
@@ -41,6 +41,22 @@ const shouldRemoveMcpConfig = (
 const containsNunjucksTemplate = (value: string): boolean =>
   /{{[\s\S]*}}|{%[\s\S]*%}|{#[\s\S]*#}/.test(value);
 
+const replaceNunjucksTemplates = (value: string, placeholder: string): string =>
+  value.replace(/{{[\s\S]*?}}|{%[\s\S]*?%}|{#[\s\S]*?#}/g, (template, offset, source) => {
+    if (template.startsWith('{#') || template.startsWith('{%')) {
+      return '';
+    }
+
+    const before = source.slice(0, offset);
+    const after = source.slice(offset + template.length);
+    return before.endsWith('[') && after.startsWith(']') ? '::1' : placeholder;
+  });
+
+const getNunjucksConditionalBranches = (value: string): string[] => {
+  const branches = value.split(/{%-?\s*(?:else|elif\b(?:(?!%})[\s\S])*)\s*-?%}/);
+  return /{%-?\s*else\s*-?%}/.test(value) ? branches : [...branches, ''];
+};
+
 function ProviderConfigEditor({
   provider,
   setProvider,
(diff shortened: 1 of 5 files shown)


src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
```diff
@@ -65,19 +295,91 @@ function ProviderConfigEditor({
...
+        const candidates = getNunjucksUrlCandidates(
+          parser.parse(url),
+          { remaining: 128 },
+          overriddenFilters,
+        );
+        if (!candidates) {
+          return false;
+        }
+
+        return candidates.some((candidate) => {
```
jameshiester-oai, line 341: [P2] Require every statically enumerated branch to be valid. `getNunjucksUrlCandidates` returns each reachable `if`/`else` result, but `.some(...)` accepts the whole template when only one branch is a WebSocket URL. The newly added accepted case `{% if useHttp %}http{% endif %}ws://example.com/ws` renders `httpws://example.com/ws` when `useHttp` is true, and the runtime `ws` constructor throws synchronously for that protocol. This lets users save a template that deterministically fails for a subset of evaluations. Validate all concrete branch candidates (while handling an unconstrained whole-URL expression separately).

jameshiester-oai (commented):
Reviewed the current head. I found one validation bug that can let a statically invalid WebSocket URL branch through the editor.

PR #10178, opened 2026-07-22
fix(websocket): preserve safe distinct provider identities

wholley-oai:
## Summary
- WebSocket targets can include credentials in their connection URLs. Redacting those URLs before using them as provider IDs caused separately authenticated targets to collapse into the same eval column, while partially templated URLs could still leak credentials or lose valid Nunjucks control tags.
- Preserve ordinary WebSocket URL identities and existing resume/grouping behavior, use explicit non-secret provider IDs or labels when authenticated URLs would otherwise collide, and reject ambiguous direct or rate-limit-wrapped targets before results are persisted.
- Keep provider IDs, logs, exported results, and persisted configuration free of username-only credentials, conditional query/fragment values, credential-shaped path segments, encoded secrets, and JWTs while preserving complete Nunjucks syntax, UUID resource IDs, and ordinary descriptive labels/routes.
- Fail closed when provider sanitization encounters an unsafe identity, and document when WebSocket URLs and unique non-secret labels are required.

## Validation
- `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 npm ci` under Node 24.18.0 / npm 11.16.0.
- `npm run tsc`.
- `npm run l` and `npm run f`; both pass with the existing upstream warning on the unchanged WebSocket message-handler complexity.
- `./node_modules/.bin/vitest run test/evaluator test/providers/websocket.test.ts test/providers/registry.test.ts test/providers.test.ts test/util/sanitizer.test.ts test/models/evalResult.test.ts test/scheduler`: **43 files, 1,412 tests passed**.
- Deterministic local loopback: `npm run local -- eval -c /private/tmp/promptfoo-websocket-identity-qa.yaml --no-cache --max-concurrency 2 -o /private/tmp/promptfoo-websocket-identity-results.json` produced separate `Account A`/`Account B` columns, outputs `from-a`/`from-b`, `promptIdx` values 0/1, one passing metric per column, and no raw credentials in the exported JSON.
- Three independent read-only contract, runtime/security, and simplification/documentation reviews completed with no remaining actionable findings.

### DIFF
--- a/src/providers/providerLogging.ts
+++ b/src/providers/providerLogging.ts
@@ -1,10 +1,50 @@
-import { REDACTED, sanitizeObject, sanitizeUrl } from '../util/sanitizer';
+import {
+  looksLikeCredentialPathSegment,
+  looksLikeSecret,
+  REDACTED,
+  sanitizeObject,
+  sanitizeUrl,
+  sanitizeUrlEncodedString,
+} from '../util/sanitizer';
 
 /**
  * Returns a provider identifier that preserves ordinary URLs while ensuring
  * credential-bearing URLs are safe to log or persist.
  */
 export function getSafeProviderId(url: string): string {
+  if (!url.includes('://') && !url.startsWith('/')) {
+    let decodedUrl = url;
+    try {
+      decodedUrl = decodeURIComponent(url);
+    } catch {}
+    if (looksLikeSecret(url) || looksLikeSecret(decodedUrl)) {
+      return REDACTED;
+    }
+
+    for (const part of decodedUrl.split(/[\s{}/?&#;]+/)) {
+      const candidate = part.replace(/^[-_.:]+/, '');
+      const value = candidate.includes('=')
+        ? candidate.slice(candidate.indexOf('=') + 1)
+        : candidate;
+      const embeddedCredential =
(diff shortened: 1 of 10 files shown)


jameshiester-oai (commented):
Reviewed the current head. The focused sanitizer/provider/evaluator/model suites and TypeScript check pass, but I found one additional credential-leak path. The existing current-head identity-collision findings also remain relevant.

src/providers/providerLogging.ts
```diff
@@ -1,10 +1,50 @@
...
+  sanitizeUrl,
+  sanitizeUrlEncodedString,
+} from '../util/sanitizer';
 
 /**
  * Returns a provider identifier that preserves ordinary URLs while ensuring
  * credential-bearing URLs are safe to log or persist.
  */
 export function getSafeProviderId(url: string): string {
+  if (!url.includes('://') && !url.startsWith('/')) {
```
jameshiester-oai, line 15: [P1] Reject scheme-less userinfo credentials before treating the value as an ordinary ID. An explicit WebSocket ID or label such as `runtime-user:runtime-secret@host` enters this branch, matches none of the value heuristics, and is returned unchanged. `WebSocketProvider` then accepts it, and both `id()` and `toString()` expose the raw password (the percent-encoded form decodes to the same authority and also passes). Since this PR's contract is that explicit IDs and labels must be non-secret, detect `user:pass@host` structure even when `://` is absent before returning the original value.

PR #10197, opened 2026-07-23
fix(providers): preserve WebSocket retries for certificate-like URLs

mldangelo-oai:
## Summary

- Remove rendered WebSocket URLs from transport error messages before classifying permanent TLS failures.
- Preserve retryable `EPROTO` failures when ordinary endpoint names or URL values contain certificate-related text.
- Keep permanent certificate/protocol failures fail-fast and continue redacting rendered URLs, tenant names, session identifiers, and query credentials.
- Add a composed `WebSocketProvider` → `RateLimitRegistry` regression covering retry count, scheduler metrics, safe error text, and log redaction.

## Regression

On current `main`, a transient error such as `write EPROTO transient TLS failure ws://certificate-service.invalid/...` is incorrectly classified as a permanent certificate failure. The scheduler stops after one attempt and exposes only `WebSocket connection failed` instead of retaining the safe, retryable `EPROTO` classification.

The new regression failed before the fix and passes afterward.

## Validation

- `./node_modules/.bin/vitest run test/providers/websocket.test.ts test/scheduler/retryPolicy.test.ts test/scheduler/providerRateLimitState.test.ts test/util/fetch/errors.test.ts` — **191 tests passed**.
- `npm run tsc`.
- `npm run architecture:check`.
- `npm run l` and `npm run f`.
- Real local CLI evaluation against a loopback WebSocket endpoint containing `/certificate-service`: the first upgrade returns HTTP 503, the second succeeds, and the exported result reports one passing assertion with output `recovered-after-retry`.

### DIFF
--- a/src/providers/websocket.ts
+++ b/src/providers/websocket.ts
@@ -46,9 +46,10 @@ function getSafeWebSocketError(event: WebSocket.ErrorEvent): Error {
   }
 
   const isSafeTransportCode = ['ECONNRESET', 'ECONNREFUSED', 'EPIPE'].includes(sourceCode ?? '');
+  const transportErrorMessage = sourceError.message.replace(/\b(?:wss?|https?):\/\/\S+/gi, '');
   const isPermanentProtocolError =
     /wrong version number|self signed|unable to verify|unknown ca|cert|alert protocol version|unsupported protocol/i.test(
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
@codex review

PR #10202, opened 2026-07-24
fix(assertions): default RAG assertion thresholds to 0.5

### LINKED ISSUE #9848: answer-relevance defaults to threshold 0, so it silently always passes when no threshold is set

### LINKED ISSUE #9910: bug(assertions): context-recall and context-relevance default threshold to 0, silently always pass

pollychen-lab:
## Summary

`answer-relevance`, `context-faithfulness`, `context-recall`, and `context-relevance` all defaulted `threshold` to `0` when it was omitted, so the assertion passed for *any* score — including `0.0`. They were silently always-green unless you happened to set a threshold.

This defaults them to **0.5**, shared through a single constant rather than four inline literals:

- New `src/assertions/ragDefaults.ts` exports `DEFAULT_RAG_ASSERTION_THRESHOLD = 0.5` and an `applyRagInverse()` helper, so the `not-*` variants invert `pass` and `score` consistently and skip inversion on grader failures (a transport error should not become a pass).
- All four assertions wired to it.
- Docs updated: the four model-graded pages plus the two index pages previously documented `default: 0`.

Fixes #9910.
Fixes #9848.

## Why 0.5

`src/external/assertions/deepeval.ts` already uses `DEFAULT_THRESHOLD = 0.5` for a model-graded assertion, so 0.5 matches existing in-repo precedent. The doc *examples* on these pages suggest different numbers (`0.3` on context-relevance, `0.7` on answer-relevance, `1.0` on context-recall), so they are not a consistent signal for a single default — and a lower default breaks fewer existing evals.

**This is a behavior change.** Evals that pass today with an omitted threshold and a low grader score will start failing. That is the bug being fixed, but it warrants a release note.

## Credit

Five contributors independently identified and fixed this. This PR was chosen as the one to land because it is the only one that covers all four assertions and shares the default through a named constant instead of scattering literals. Please preserve their attribution when squashing:

```
Co-authored-by: Varshith Puli <pulivarshit@gmail.com>
Co-authored-by: Anas Khan <83116240+anxkhn@users.noreply.github.com>
Co-authored-by: JSap0914 <JSap0914@users.noreply.github.com>
Co-authored-by: Abhinav Prakash <abhinavprakash616@gmail.com>
```

After analysing all four against this one:

- **#10494** (Varshith-Kali) and **#9907** (anxkhn) are **not** duplicates. This PR's `applyRagInverse()` skips inversion on grader failures by checking `metadata.graderError`, but the RAG matchers return an *untagged* `fail()` on a grading-provider error, so that guard never fires. Those two PRs fix `src/matchers/rag.ts` — which this PR does not touch — for `context-recall`/`context-relevance` and `context-faithfulness` respectively. They should land right after this one.
- **#9911** (JSap0914) and **#10567** (AbhiPra24) genuinely reduce to nothing once this lands; their residual test cases are strict subsets of cases here. Closing them with the co-authorship above is the right outcome.

**Known gap worth a follow-up:** all four RAG matchers use the untagged `fail()` today. #10494 fixes two and #9907 fixes one, which leaves `answer-relevance` still able to invert a grading-provider outage into a pass.

## Testing

CI is green on the merged branch. Locally: `tsc --noEmit` clean, `architecture:check` passes, `knip` clean, Biome clean, and the four `test/assertions/context*|answerRelevance` suites pass against the 0.5 default.

### DIFF
--- a/site/docs/configuration/expected-outputs/index.md
+++ b/site/docs/configuration/expected-outputs/index.md
@@ -176,10 +176,10 @@ See [Model-graded evals](/docs/configuration/expected-outputs/model-graded), [cl
 | [moderation](/docs/configuration/expected-outputs/moderation)                                        | Check output against safety policies and include provider-reported usage metrics |
 | [llm-rubric](/docs/configuration/expected-outputs/model-graded)                                      | LLM output matches a given rubric, using a Language Model to grade output        |
 | [g-eval](/docs/configuration/expected-outputs/model-graded/g-eval)                                   | Chain-of-thought evaluation based on custom criteria using the G-Eval framework  |
-| [answer-relevance](/docs/configuration/expected-outputs/model-graded)                                | Ensure that LLM output is related to original query                              |
-| [context-faithfulness](/docs/configuration/expected-outputs/model-graded)                            | Ensure that LLM output uses the context                                          |
-| [context-recall](/docs/configuration/expected-outputs/model-graded)                                  | Ensure that ground truth appears in context                                      |
-| [context-relevance](/docs/configuration/expected-outputs/model-graded)                               | Ensure that context is relevant to original query                                |
+| [answer-relevance](/docs/configuration/expected-outputs/model-graded)                                | Ensure that LLM output is related to original query (default threshold 0.5)      |
+| [context-faithfulness](/docs/configuration/expected-outputs/model-graded)                            | Ensure that LLM output uses the context (default threshold 0.5)                  |
+| [context-recall](/docs/configuration/expected-outputs/model-graded)                                  | Ensure that ground truth appears in context (default threshold 0.5)              |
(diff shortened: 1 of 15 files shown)


pollychen-lab:
thank you all

PR #10203, opened 2026-07-24
fix(openrouter): avoid crash on empty choices array

dfedoryshchev:
## Summary

`OpenRouterProvider` reads `data.choices[0].message` directly after the `data.error` check. An empty `choices: []` array, or a degenerate 200 response with no `choices` field, makes `data.choices[0]` undefined, so `.message` throws a `TypeError`. That path is not wrapped in a try/catch, so it propagates out of `callApi` as an unhandled crash instead of a usable result.

This adds the same `Malformed response data` guard the other OpenAI-compatible providers here already use (e.g. Mistral, AI21), so an empty or missing `choices` returns a clear error rather than crashing.

## Test

Added a regression test in `test/providers/openrouter.test.ts`: a 200 response with `choices: []` now returns a `Malformed response data` error instead of throwing. The full OpenRouter suite passes (35 tests) locally.

---

## Maintainer follow-up

Two commits pushed on top of the original fix during review. The diagnosis above is
correct — `withGenAISpan` re-throws (`src/tracing/genaiTracer.ts:263`), so the
`TypeError` does escape `callApi` — and copying Mistral's guard rather than AI21's was
the right call, since AI21's variant also requires `message.content` and would reject
OpenRouter's valid tool-call responses.

### `fix(openrouter): evict malformed responses from the cache`

Returning an error was not sufficient on its own. `fetchWithCache` skips caching only
for non-2xx responses and bodies carrying an `error` key (`src/cache.ts:727-758`), so a
200 with an empty `choices` array is **already written to the cache** before the guard
runs. Without eviction the failure was pinned: every later run of the same prompt
replayed the cached body and never reached OpenRouter again, leaving `--no-cache` or
deleting `~/.promptfoo/cache` as the only recovery.

That matters because this shape is transient — it is what OpenRouter returns when an
upstream provider fails — so pinning it defeats the point of degrading gracefully.

`OpenAiChatCompletionProvider` already handles this: it wraps the same
`choices[0].message` access in try/catch (`openai/chat.ts:612`) and calls
`deleteFromCache()` before returning (`:873`). The override had dropped that, so this
takes `deleteFromCache` from `fetchWithCache` and calls it on the malformed path.

Verified end to end: feeding a malformed body followed by a valid one for the same
prompt reached the network twice and recovered. Before the change the second call was
served from cache and returned the same error.

### `test(openrouter): cover every malformed-response shape`

The guard has three clauses but only `choices: []` was exercised. The malformed-response
cases now live in their own suite, parametrized over all three shapes. Each throws a
distinct `TypeError` without the guard, confirming every clause is load-bearing:

| Body | Error without the guard |
| --- | --- |
| `{ choices: [], usage }` | `Cannot read properties of undefined (reading 'message')` |
| `{ usage }` | `Cannot read properties of undefined (reading '0')` |
| `{ choices: [{ finish_reason: 'stop' }], usage }` | `Cannot read properties of undefined (reading 'function_call')` |

The suite is 38 tests and green; `tsc` and `lint:src` are clean.

### Deliberately left alone

- **`tokenUsage` is still dropped on this path**, even though `usage` is often present in
  the malformed body. Every sibling provider and the parent's own catch behave the same
  way, so changing it here would be an inconsistency in the other direction.
- **No `metadata.http`.** The parent attaches it, but `openrouter.ts` attaches metadata on
  no error path today; adding it to one would be arbitrary.
- **The `executeOpenRouterCall` complexity warning** (37 vs. max 30) predates this PR.

### DIFF
--- a/src/providers/openrouter.ts
+++ b/src/providers/openrouter.ts
@@ -180,6 +180,12 @@ export class OpenRouterProvider extends OpenAiChatCompletionProvider {
       };
     }
 
+    if (!data.choices || !data.choices[0] || !data.choices[0].message) {
+      return {
+        error: `Malformed response data: ${JSON.stringify(data)}`,
+      };
+    }
+
     // Process the response with special handling for Gemini
     const message: any = data.choices[0].message;
     const finishReason = normalizeFinishReason(data.choices[0].finish_reason);
--- a/src/providers/openrouter.ts
+++ b/src/providers/openrouter.ts
@@ -142,9 +142,10 @@ export class OpenRouterProvider extends OpenAiChatCompletionProvider {
     let status: number;
     let statusText: string;
     let cached = false;
+    let deleteFromCache: (() => Promise<void>) | undefined;
 
     try {
-      ({ data, cached, status, statusText } =
+      ({ data, cached, status, statusText, deleteFromCache } =
         await fetchWithCache<OpenRouterChatCompletionResponse>(
           `${this.getApiUrl()}/chat/completions`,
           {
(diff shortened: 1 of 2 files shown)


dfedoryshchev:
Closing as it was handled in https://github.com/promptfoo/promptfoo/pull/10418

PR #2967, opened 2025-02-03
chore(redteam): add foundation model plugin collection

mldangelo:
chore(redteam): add foundation model plugin collection

- Introduced a new 'Foundation' plugin collection for redteam testing.
- Updated relevant constants and presets to include the new collection.

No breaking changes.

PR #10214, opened 2026-07-26
docs(site): add DEF CON 34 and refresh Black Hat USA 2026 event pages

mldangelo:
Both Vegas conferences this summer, attending as part of OpenAI. No party this year, so both pages point at the booth and the hallway track instead of an RSVP link.

## The pages

**`/events/defcon-2026/` (new)** — built around DEF CON 34's official theme, **"Agency"**. The joke writes itself: DEF CON is celebrating human self-determination, and we spend our days taking agency away from AI agents that have too much of it. Green-phosphor terminal treatment, an animated `promptfoo redteam run` transcript against a tool-using agent, and a `man agency` panel that puts DEF CON's definition next to the OWASP "excessive agency" one.

| Hero | `man agency` |
|---|---|
| DEF CON hero | man agency |

**`/events/blackhat-2026/` (rewritten)** — the red-on-black briefing-room sibling. Same series, different mood: this is the enterprise stop on the run. Adds a five-step attack-loop pipeline (Discover → Generate → Attack → Grade → Regress) as the signature visual, plus a demo card for agentic/tool-use attacks.

| Hero | Attack loop |
|---|---|
| Black Hat hero | pipeline |

Both cross-link as **"the Vegas run"** (Aug 1 → Aug 9; August 6 belongs to both), and the site announcement bar now points at them while keeping the OpenAI mention.

## Details

- **Booths:** #2967 in the Black Hat Business Hall (Aug 5–6), #1412 at DEF CON.
- **Dates verified against the organizers:** Black Hat USA Aug 1–6 at Mandalay Bay (Trainings Aug 1–4, Briefings + Business Hall Aug 5–6); DEF CON 34 Aug 6–9 at LVCC West Hall.
- New card/OG art at `static/img/events/defcon-2026.jpg`, 1536×1024 to match the rest of the family.
- Neither page promises a talk, an Arsenal slot, a giveaway, or a party.

## Two fixes in `events.ts` that this surfaced

Both are things the new pages tripped over rather than drive-by cleanup:

1. **Stale `status: 'upcoming'` literals.** Five 2026 events that ended months ago were still hardcoded as upcoming, so `/events/` featured a February event and would have buried the Vegas pages. Swapped to the `getEventStatus(endDate)` helper the 2025 entries already use, and rewrote the now-past copy into past tense — RSA 2026 was still offering a "complimentary AI vulnerability assessment" and "exclusive swag" for an event that ended in March.

2. **`formatEventDate` was timezone-dependent.** It read the end date through `new Date(...).getDate()` in the build host's local zone. All the 2026 entries end at 18:00 PDT, which is past midnight UTC, so a UTC CI builder rendered DEF CON as "Aug 6-10" on the events index while every string on the page said August 6–9. Now parses the calendar date off the ISO prefix.

## Tests

`site/src/pages/events/vegas-2026.test.tsx` — 36 cases, added because the timezone fix above is a real bug fix and was landing without coverage.

- `formatEventDate` asserted stable across UTC, `America/Los_Angeles`, `Pacific/Kiritimati`, `Asia/Tokyo`. Reverting the fix fails 3 of these.
- Every event's `cardImage`/`heroImage` must resolve inside `site/static`, and every `customPageUrl` must have a page component. This is the check that was missing: a dead asset path renders a broken tile on `/events/` and a 404 og:image, and neither `tsc` nor the Docusaurus build catches it — the missing `defcon-2026.jpg` in an earlier draft got flagged only on manual review.
- Both pages rendered and asserted on booth number, dates, venue, the OpenAI mention, the cross-link, and a single `<h1>`.
- Neither 2026 page may offer a party affordance. Checks for an actual invite — a link/button, a `lu.ma` URL, an open-bar offer — rather than the words, since the DEF CON page legitimately says there is "no RSVP link" this year.

All three assertions were mutation-checked (revert the TZ fix / change a booth number / break an image path) to confirm they fail when they should.

## Verification

`tsc --noEmit` clean · site suite 9 files / 144 tests, green under both the local zone and `TZ=UTC` · `SKIP_OG_GENERATION=true npm run build` clean with `onBrokenLinks: 'throw'` · Biome clean via an isolated probe (it stack-overflows on this working copy, [biomejs/biome#8783](https://github.com/biomejs/biome/issues/8783)).

Checked in a real browser at 1440px and 360px: no horizontal overflow on either page, no skipped heading levels, no broken images. Every animation is neutralized under `prefers-reduced-motion`, and contrast was measured rather than eyeballed — one hero gradient terminus was failing AA at 2.72:1 and got lightened.

## Also in this PR

A later commit adds an **OpenAI security lineup** section to both pages — Daybreak, Codex Security, and Promptfoo — so booth visitors asking about appsec get pointed somewhere real. Scoped strictly to what OpenAI has announced publicly; nothing claims Promptfoo ships as part of Daybreak or Codex Security, because no public source says that. The framing that earns its place: *Codex Security reads the code you wrote, we go after the agent you shipped.*

Same commit does a voice pass (contractions throughout) — the uncontracted "we are"/"we will"/"here is" was the main thing making the prose read as machine-written.

## Known follow-ups, not in this PR

- The five now-past 2026 events still have dedicated pages (`rsa-2026.tsx` et al.) whose on-page copy reads future-tense. That staleness predates this change — the index badge is now correct where it was previously wrong — but the pages themselves deserve a pass.
- `getEventsByYear` has the same timezone trap I fixed in `formatEventDate`. It has no callers and no current event sits near a year boundary, so I left it rather than widen the diff.
- Two other pages (`blackhat-2025`, `defcon-2025`) declare `og:image` as 1200×630 against 1536×1024 assets. I originally wrote "~12 pages" here; on audit it is exactly two — the rest either declare no dimensions or point at `/img/og/*.png`, which the OG plugin generates at build time. Fixed in the follow-up.


ianw-oai (approved):
Looks good.

PR #10215, opened 2026-07-26
fix(site): repair stale event pages, dead links, and hydration mismatches

mldangelo:
> **Stacked on #10214.** Base is `docs/site-vegas-2026-event-pages`, so the diff here is just this change. Merge #10214 first.

Writing the Vegas pages surfaced a pile of adjacent rot, so I swept `site/` properly. 32 candidate findings, 24 confirmed after adversarial verification (each finding was handed to a separate agent instructed to *refute* it — 8 died there, mostly style opinions and one asset that turns out to be build-generated).

Today is 2026-07-26, and most of this is what happens when event pages outlive their events.

## Things that were visibly wrong in production

**Frozen countdowns.** `rsa-2026` and `bsides-sf-2026` rendered `0 Days : 00 : 00 : 00`, permanently. `updateCountdown()` only called `setTimeLeft()` inside `if (difference > 0)` with no `else`, so once the target date passed, state stayed at the `useState` initial zeros forever — while a 1 Hz interval kept ticking, doing nothing, for every visitor. `bsides-seattle-2026` *had* the else branch but no upper bound, so it cheerfully announced **"Event is happening now!"** for an event that ended in February — directly above the line saying "February 27-28, 2026".

**Invitations to events that are over.** Nine event pages still read as live invites, including in the `og:`/`twitter:` descriptions that feed search results and link previews, so the false invite propagated off-site. Two of them promised a *future* return to events that have themselves already happened: `rsa-2025` said "See You at RSA 2026" and `bsides-sf-2025` said "See You at BSides SF 2026" — both ended in March.

**Offers on "Past" cards.** `events.ts` still advertised "Free drinks on us" and "Get a complimentary AI vulnerability assessment" on cards the index was rendering with a `Past` badge next to them.

**Dead links.** `bsides-seattle-2026` linked to `/docs/category/red-teaming/` (404 — verified live; the real route is `/docs/red-team/`). Two press links were dead: the a16z podcast URL 404s and the Latent Space Substack post redirects to a 404. Both replaced with verified equivalents matched on title, show, and publish date.

**Blank share cards.** `guardrails`, `mcp`, and `model-security` set `og:image` to `/img/meta/*.png` files that do not exist, so every share of those three landing pages rendered with no image.

## Correctness

- **`formatEventDate` rendered single-day events as `"Dec 3-3, 2025"`.** The single-day guard was `startDate === endDate`, comparing full ISO timestamps — which differ by time of day even on the same date, so it never matched. Now compares calendar dates. Affects three events.
- **Timezone-dependent year reads.** `getEventsByYear` plus five call sites in `events/index.tsx` used `new Date(...).getFullYear()` in the build host's zone. Added `getEventYear()` next to the existing `parseCalendarDate`. Latent today, but it's the same bug class as the `formatEventDate` one fixed in #10214.
- **Two hydration mismatches.** `GitHubStars` read `localStorage` and `HomepageWalkthrough` read `window.location.hash` inside lazy `useState` initializers. SSR can see neither, so a returning visitor — or anyone opening `/#evals` — hydrated with different markup than the server sent. Both moved into effects. The walkthrough now also subscribes to `hashchange`, which it never did, so in-page hash navigation actually changes the tab.

## Accessibility

- `ImageJailbreakPreview`'s click-to-flip card was a bare `<div onClick>` — keyboard and screen-reader users could not reveal the images at all. Added `role="button"`, `tabIndex`, `aria-expanded`, a state-aware label, Enter/Space handling, and a focus ring. Used 9 times in one blog post.
- `validator.tsx` had **no `<h1>`** — MUI rendered its page title as a literal `<h3>`, so the outline started three levels deep. `press` and `about` emitted phantom `<h6>`s from bare `subtitle1`/`subtitle2` Typography.
- Four event stylesheets and both shared `Events` components ran infinite animations with no `prefers-reduced-motion` guard.

  Worth a reviewer's eye: `blackhat-2025`'s `.demoCard` starts at `opacity: 0` with a fill-forwards entrance animation, so a blanket `animation: none` would have made those cards **permanently invisible** under reduced motion. The guard restores `opacity`/`transform` explicitly. I verified this against the built CSS bundle rather than the source, confirming the minified rule keeps `opacity:1` on that class.

- Hardcoded `80+ Fortune 500` replaced with `SITE_CONSTANTS.FORTUNE_500_COUNT` (156) in three places — it was sitting in the same stat row as a live-sourced developer count.

## Verification

`tsc --noEmit` clean · site suite 9 files / 144 tests · `SKIP_OG_GENERATION=true npm run build` clean with `onBrokenLinks: 'throw'` · Biome clean on all 21 changed TS/TSX files (via isolated probe — it stack-overflows on this working copy, [biomejs/biome#8783](https://github.com/biomejs/biome/issues/8783)) · Prettier clean on CSS.

Beyond that: every touched page served and inspected on a running dev server, plus a scripted sweep across all 14 event pages confirming zero remaining `happening now` / `Meet us at` / `Join us at` / `See You at` / `Register now` strings.

## Tests

Two suites added, both mutation-checked (revert the fix, the test fails):

- **`site/src/data/events.test.ts`** — the single-day date rendering (restoring `startDate === endDate` fails 4 cases, including a blanket "no event renders a same-day range" assertion), `getEventYear`/`getEventsByYear` across four timezones plus year-boundary timestamps, and a guard that past events carry no live invite or standing offer.
- **`site/src/components/ImageJailbreakPreview.test.tsx`** — keyboard access: role, `tabIndex`, `aria-expanded`, state-aware accessible name, Enter/Space activation, non-activating keys. Reverting to the bare `<div onClick>` fails all 6.

Writing these caught two things in my own work: the offer regex initially flagged "Open Bar" as a *highlight label*, which is legitimate on a recap — narrowed to prose only; and my first component test used the wrong prop shape entirely.

## One sweep change reverted

An agent rewrote the DEF CON 33 party times in `events.ts` (20:00–23:59 → 18:00–20:00) to match the page, which says 6–8 PM. Nothing in the site renders event times, so it fixed no user-visible bug and rewrote a historical record on inference. Reverted; the page copy is untouched and remains the authority.

## Deliberately not done

- Press links returning 403 to bots (`cybernews`, `washingtontimes`, `finance.yahoo`) — those are bot blocks, not breakage.
- Docusaurus 3.10.1 → 3.10.2. It's a cosmetic CLI nag; `npm update` on those six packages actually moves 72 packages. Renovate's call, its own PR.

### DIFF
--- a/site/src/components/Events/EventCard.module.css
+++ b/site/src/components/Events/EventCard.module.css
@@ -383,3 +383,14 @@
     transform: scale(0.98);
   }
 }
+
+/* Decorative-only loops (shimmer/pulse/glow); safe to stop entirely. */
+@media (prefers-reduced-motion: reduce) {
+  .card,
+  .card *,
+  .card *::before,
+  .card *::after {
+    animation: none !important;
+    transition: none !important;
+  }
+}
--- a/site/src/components/Events/EventCard.module.css
+++ b/site/src/components/Events/EventCard.module.css
@@ -386,7 +386,11 @@
 
 /* Decorative-only loops (shimmer/pulse/glow); safe to stop entirely. */
 @media (prefers-reduced-motion: reduce) {
+  /* The root's own pseudo-elements need naming explicitly: `.card *::before`
+     matches descendants only, so `.card::before` would keep animating. */
   .card,
+  .card::before,
+  .card::after,
   .card *,
   .card *::before,
   .card *::after {
--- a/site/src/components/Events/FeaturedEvent.module.css
+++ b/site/src/components/Events/FeaturedEvent.module.css
@@ -455,3 +455,14 @@
     font-size: 0.625rem;
   }
 }
+
+/* Decorative-only loops (shimmer/pulse/glow); safe to stop entirely. */
+@media (prefers-reduced-motion: reduce) {
+  .featured,
+  .featured *,
+  .featured *::before,
+  .featured *::after {
+    animation: none !important;
+    transition: none !important;
+  }
+}
--- a/site/src/components/Events/FeaturedEvent.module.css
+++ b/site/src/components/Events/FeaturedEvent.module.css
@@ -458,7 +458,11 @@
 
 /* Decorative-only loops (shimmer/pulse/glow); safe to stop entirely. */
 @media (prefers-reduced-motion: reduce) {
+  /* The root's own pseudo-elements need naming explicitly: `.featured *::before`
+     matches descendants only, so `.featured::before` would keep animating. */
   .featured,
+  .featured::before,
+  .featured::after,
   .featured *,
   .featured *::before,
   .featured *::after {
--- a/site/src/components/GitHubStars/index.test.tsx
+++ b/site/src/components/GitHubStars/index.test.tsx
@@ -52,19 +52,18 @@ describe('GitHubStars', () => {
     const serverMarkup = container.innerHTML;
     expect(serverMarkup).toContain(BUILD_TIME_COUNT);
 
-    const errors: unknown[] = [];
-    const originalError = console.error;
-    console.error = (...args: unknown[]) => errors.push(args.join(' '));
+    // React 19 routes hydration mismatches through onRecoverableError, NOT console.error —
+    // vitest intercepts them before a console spy can see them, which makes a spy-based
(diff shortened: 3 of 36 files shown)


ianw-oai (approved):
Looks good.

PR #10230, opened 2026-07-28
fix(assertions): trim whitespace before tokenizing in METEOR

serhiizghama:
The METEOR assertion tokenizes with `candidate.split(/\s+/)` without trimming first, unlike `bleu` and `gleu` which both `.trim()` before splitting. Leading or trailing whitespace — very common in LLM output — injects empty-string tokens that inflate the word count and drop precision. A response that's identical to the reference apart from a leading space and trailing newline scores precision 0.75 instead of 1.0 (6 real tokens become 8), and when both sides are padded the two empty tokens even match each other. The "identical strings return 1" shortcut doesn't save you here because the padded/clean token arrays differ in length.

Added `.trim()` on both the candidate and reference before splitting, matching the sibling metrics. Exported `calculateMeteorScore` (bleu/gleu already export their `calculate*` helpers) and added a test that a padded answer scores the same as the clean one. The inputs match word-for-word so alignment stays in the exact-match stage and no WordNet lookups run.


serhiizghama:
Merged main and cleared the red check — sorry for leaving it sitting.

The failure was the coverage ratchet, not the tests: it gates any changed file under `src/assertions/` at 80/70/80/80, and `meteor.ts` was at 66% because `meteor.test.ts` drives a hand-written reimplementation of the assertion rather than the module itself. Added `meteorScoring.test.ts`, which exercises the real `calculateMeteorScore` and `handleMeteorAssertion` (the whitespace regression moved in there too). `meteor.ts` is now 91.37% statements / 90.16% branches / 100% functions / 90.82% lines, and `checkCoverageRatchets.ts --report backend` passes.

Only the no-alignment case falls through to WordNet; everything else resolves in the exact-match or Porter-stem stage, so the file still runs in well under a second and stays out of the integration matrix.

serhiizghama:
Good catch, that's a real regression from the trim. `''.split(/\s+/)` gives `['']`, so a whitespace-only output and an empty expected value both tokenized to a single empty token that matched itself — score 0.5, which clears the default threshold. Before the trim the padded candidate produced two empty tokens, so it scored 0.4545 and failed.

Tokenizing through a helper that returns `[]` for an all-whitespace string now, so both sides come out empty and the score is 0. Added it to the regression test.

serhiizghama:
Fair point — added a short note under "How METEOR Works" in `deterministic.md` saying that surrounding whitespace is trimmed before tokenization (same as BLEU/GLEU) and that an input that's empty after trimming scores 0, mirroring the existing GLEU wording.

PR #10232, opened 2026-07-28
fix(assertions): return 0 for empty GLEU candidate instead of throwing

serhiizghama:
A GLEU assertion crashes with `Invalid inputs` when the model output is an empty string — `calculateGleuScore` treats `!candidate` as invalid and throws. A whitespace-only output slips past that check but then tokenizes to a single empty token and scores a degenerate value.

Empty output is a normal thing to score (a refusal or a truncated generation), and the sibling `bleu` and `rouge` assertions already handle it by returning 0. GLEU is the odd one out — an empty completion errors the whole assertion instead of just failing the threshold, and a `not-gleu` inverse assertion that should pass on empty output errors too.

Made GLEU match: an empty or whitespace-only candidate returns 0, and the throw is now reserved for a genuinely null/undefined candidate or an empty references array. Added tests at both the score and handler level (including the inverse case), mirroring the existing bleu coverage.

### DIFF
--- a/src/assertions/gleu.ts
+++ b/src/assertions/gleu.ts
@@ -31,7 +31,8 @@ import type { AssertionParams, GradingResult } from '../types/index';
  * @param minN - Minimum n-gram length to consider (default: 1)
  * @param maxN - Maximum n-gram length to consider (default: 4)
  * @returns GLEU score between 0 and 1, where higher scores indicate better matches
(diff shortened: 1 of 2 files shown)


serhiizghama:
#9850 landed the same fix back in August — closing this one.

PR #10233, opened 2026-07-28
fix(providers): preserve scheduled Claude task results

mldangelo-oai:
## Summary

- Recognize Claude Agent SDK 0.3.214 scheduled-trigger notifications as main-agent results.
- Keep background task notifications separate so late subagent results cannot replace or poison the scheduled task response.
- Add a regression test that fails on current main and verifies the scheduled output, session, turn count, and raw result.

## Validation

- `PROMPTFOO_DISABLE_TELEMETRY=1 ./node_modules/.bin/vitest run test/providers/claude-agent-sdk.test.ts test/providers/registry.test.ts --no-file-parallelism --configLoader runner` (265 tests)
- `npm run tsc -- --tsBuildInfoFile /private/tmp/promptfoo-scheduled-trigger-20260728.tsbuildinfo`
- `npm run l`
- `npm run f`
- `git diff --check`


jameshiester-oai (commented):
Reviewed base `249ee3379ed98335aa607cbedca22f74dab7f492` and exact head `021d9b1b3bc07550c2e355b982ff9bd95078182e`.

Scope: fresh full review of the complete `base...head` diff and surrounding Claude Agent SDK result-selection path.

Verdict: no actionable defects found. The new `task-notification` / `scheduled-trigger` discriminator matches the pinned `@anthropic-ai/claude-agent-sdk` 0.3.214 declaration, remains compatible with older messages that omit `origin` or `subkind`, and preserves the existing fallback behavior for ordinary background notifications. The regression covers the failure-producing ordering where a late background result follows the scheduled result.

CI: 44/44 current checks are successful. The Node 24 job passed `test/providers/claude-agent-sdk.test.ts` (169 tests) and the full run (800 test files, 22,006 tests).

- P0-P3: none.

Existing feedback: Copilot and the security scanner reported no inline findings; there are no existing review threads to duplicate.

Simplification and test opportunities: no material simplification is warranted. A reverse-order scheduled/background case or scheduled error-result case could add symmetry, but both exercise the same subtype-independent branch and do not expose a behavioral gap.

Validation performed:

- `git diff --check 249ee3379ed98335aa607cbedca22f74dab7f492...021d9b1b3bc07550c2e355b982ff9bd95078182e` — passed.
- Offline inspection of the pinned SDK 0.3.214 declaration — confirmed `subkind?: 'scheduled-trigger'` is defined only for `origin.kind === 'task-notification'` and identifies a fired stored scheduled prompt.
- No-dependency behavior matrix — the head selects the scheduled result for scheduled-then-background, reverse-order, and scheduled-only streams while preserving human-origin, legacy, and all-background behavior; the base implementation reproduces the scheduled-then-background regression.
- Exact-head CI log inspection — focused provider file and full suite passed as noted above.

Limitations: local Vitest/typecheck/build could not run because this isolated worktree had no dependencies and `npm ci --ignore-scripts` was blocked by the configured Socket policy for recently published packages. No credentialed live Claude scheduled-task environment was exercised. The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable (command not found), so its six lenses were applied manually and reconciled with three bounded read-only specialists.

PR #10236, opened 2026-07-29
fix(output): separate duplicate JUnit columns

oleksii-tumanov:
## Summary

- group JUnit suites by the canonical result-column index instead of prompt identity
- assign per-provider suite ordinals in canonical column order, independent of result arrival order
- keep duplicate provider/config columns separate even when their provider and prompt identities match
- update regression coverage for suite names, testcase classnames, and per-suite failure counts

## Why

Duplicate provider entries with different configs are a supported comparison workflow. Those columns can share the same provider ID, label, and prompt ID while still representing distinct result columns.

JUnit output previously keyed suites by provider and prompt identity, so it collapsed those columns into one suite. The evaluator now assigns a stable `promptIdx` to each result column, so the JUnit exporter can use that existing positional identity directly. Because concurrent or persisted results may be read out of order, the exporter assigns display ordinals only after restoring canonical column order.

Follow-up to #10208.

## Result

A deterministic eval with two `echo` provider configs and one prompt produces two JUnit suites:

- `[target] prompt 1` with one testcase
- `[target] prompt 2` with one testcase

The regression test fails on `main`, where both columns are emitted as one two-test suite.

## Testing

- `npx vitest run test/util/output.test.ts` — 62 passed
- `npm run tsc`
- `npm run l`
- `npm run f`
- `npm test` — 811 test files passed; 22,323 tests passed; 10 skipped
- `npm run build`
- built CLI eval with two differently configured `echo` providers, `--no-cache`, and JUnit output; inspected the generated XML for two one-test suites
- `git diff --check`

### DIFF
--- a/src/util/junit.ts
+++ b/src/util/junit.ts
@@ -81,7 +81,7 @@ function getSuiteKey(result: JunitProjectedResult): string {
   // Distinguish providers that share an id but differ by label (and vice versa)
   // so multi-target redteam runs do not collapse into a single suite.
   const providerKey = `${result.provider.id ?? ''}${SUITE_PROVIDER_SEPARATOR}${result.provider.label ?? ''}`;
-  const promptKey = result.promptId || `prompt-index:${result.promptIdx}`;
+  const promptKey = `prompt-index:${result.promptIdx}`;
   return `${providerKey}${SUITE_KEY_SEPARATOR}${promptKey}`;
 }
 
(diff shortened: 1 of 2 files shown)


oleksii-tumanov:
Hi @mldangelo-oai, gentle follow-up on this small JUnit exporter fix. I rebased it onto current `main`, re-ran the focused/full tests and build, and the new CI matrix is green; the original ordering finding is covered by a reversed-arrival regression. Would appreciate a review when you have a chance.

PR #10241, opened 2026-07-29
docs: add agent action boundary evaluation

jw-ond:
## Summary

Adds a model-free example for evaluating whether an agent action stays inside an expected runtime boundary.

The example uses the built-in echo provider plus a JavaScript assertion, so it does not require API keys or external services. It covers four representative cases:

- safe read-only action that should be allowed
- customer-data export that should be blocked
- financial side-effect action that should require dual approval
- stale verifier receipt that should require review

This is intended as a small, reproducible pattern for teams evaluating agent traces where tool calls, workflow steps, browser actions, MCP calls, or shell commands can create real side effects.

## Test plan

- Node self-test for the JavaScript assertion over all YAML fixtures
- Ran promptfoo eval from examples/agent-action-boundary using Node 24

Both checks pass locally: 4/4 promptfoo test cases passed.


jw-ond:
Current head addresses the earlier automated review items:\n\n- README heading uses the expected example-name format.\n- Direct-checkout usage uses npm run local rather than assuming promptfoo is on PATH.\n- Boundary classification checks public/external destinations before stale receipts, so stronger block controls win.\n- Malformed/incomplete action records fail closed through required-field validation.\n- The assertion compares the recommended boundary control with the runtime's observed_control, rather than only re-deriving the policy result.\n\nLocal verification note: npm run local could not run on this machine because dependencies are not installed here (tsx missing). I ran the boundary assertion directly against all 8 YAML fixtures; each fixture returned pass with the expected control.

jw-ond:
Following up on this evaluation example. The current checks are green, and the PR is limited to showing how an agent-action-boundary eval can catch drift between an approved action and the action that actually executes.

If a different cookbook location or naming convention fits promptfoo better, I can update it.

PR #10243, opened 2026-07-29
fix(redteam): preserve cached provider selection across requests

mldangelo-oai:
## Summary

- snapshot the JSON-only variant when a red-team generation request selects a cached provider
- reuse that request-scoped variant during local strategy generation even if another request replaces the shared provider cache
- add a failing-first regression that switches the cache between selection and strategy execution

## Why

The recent provider-selection refactor preserved the regular provider chosen for a request, but local JSON-producing strategies looked up the JSON variant from the mutable global cache later. If another request changed that cache in the meantime, the original request silently generated attacks with a different backend. This can cross provider, credential, or organizational boundaries.

## Verification

- Confirmed the new regression fails on the unmodified branch and passes after the fix.
- `node node_modules/vitest/vitest.mjs run test/redteam/strategies/types.test.ts test/redteam/providers/shared.test.ts test/redteam/strategies/mathPrompt.test.ts test/redteam/strategies/multilingual.test.ts test/redteam/index.test.ts test/server/routes/redteam.test.ts --configLoader runner --no-cache` — 278 tests passed.
- Ran the real source-module provider-switch scenario: request A retained `request-a-json` while the current global cache returned `request-b-json`.
- `PROMPTFOO_DISABLE_TELEMETRY=1 PROMPTFOO_DISABLE_UPDATE=true PROMPTFOO_CONFIG_DIR=/tmp/promptfoo-provider-snapshot-20260729 npm run local -- eval -c test/smoke/fixtures/configs/basic.yaml --no-cache -o /tmp/promptfoo-provider-snapshot-20260729-results.json` — 1 passed, 0 failed, 0 errors.
- `npm run l && npm run f` and `git diff --check`.
- Full `npm run tsc` could not be completed locally because the available dependency tree predates current `main`; an isolated `npm ci` was blocked by the organization package firewall with HTTP 403 for `postcss@8.5.23`.

### DIFF
--- a/src/redteam/providers/shared.ts
+++ b/src/redteam/providers/shared.ts
@@ -50,14 +50,15 @@ export type RedteamProviderSelectionSource = 'explicit' | 'cache' | 'fallback' |
 /**
  * A provider chosen for one red-team generation request.
  *
- * localProviderSpec is intentionally runtime-only: provider option objects can
- * contain resolved credentials, but are still needed to build local JSON-only
- * variants of an explicitly configured provider. Only persistableId may cross
- * a serialization boundary.
+ * cachedJsonOnlyProvider and localProviderSpec are intentionally runtime-only:
+ * the cached provider must remain request-scoped, and provider option objects
+ * can contain resolved credentials. Only persistableId may cross a serialization
+ * boundary.
  */
 export interface RedteamProviderSelection {
   provider: ApiProvider;
   source: RedteamProviderSelectionSource;
+  cachedJsonOnlyProvider?: ApiProvider;
   localProviderSpec?: RedteamFileConfig['provider'];
   persistableId?: string;
 }
@@ -315,6 +316,9 @@ class RedteamProviderManager {
     return {
       provider: await this.loadProviderCandidate(candidate, { jsonOnly, preferSmallModel }),
       source: candidate.source,
+      ...(candidate.source === 'cache' && candidate.cachedJsonOnlyProvider
+        ? { cachedJsonOnlyProvider: this.wrapProvider(candidate.cachedJsonOnlyProvider) }
+        : {}),
       localProviderSpec: candidate.spec,
       persistableId: typeof candidate.spec === 'string' ? candidate.spec : undefined,
     };
--- a/src/redteam/strategies/types.ts
+++ b/src/redteam/strategies/types.ts
@@ -52,6 +52,10 @@ export async function getStrategyGenerationProvider({
       }
     }
 
+    if (selection?.source === 'cache' && jsonOnly && selection.cachedJsonOnlyProvider) {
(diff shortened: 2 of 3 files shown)


jameshiester-oai (commented):
Reviewed base `ac8971fcfa961fa5fa96bcc4f527f5309b504997` through exact head `5d9ca7f2b9047954b8c6996d94e6e97eca405972`.

Scope: fresh full review of the provider-selection, strategy-runtime, and regression-test changes.

Verdict: no actionable defects found. The selection snapshots the cache's JSON-only provider before another request can replace shared manager state, carries that runtime-only reference without serializing provider options or credentials, and reuses it before any late global lookup. Rate-limit wrapping occurs at selection and the strategy's generation-usage wrapper is still applied. The optional field preserves existing manually constructed selection behavior. The PR title follows the redteam scope convention.

CI: all current exact-head checks are green across supported Node versions and Linux, macOS, and Windows. The Node 24 job passed the changed strategy test (6 tests), shared-provider tests (72), redteam-index tests (134), math tests (15), multilingual tests (6), route tests (45), and the full backend suite (22,050 passed; 10 skipped).

- P0/P1/P2/P3: none.

Existing feedback:
- No inline review threads exist. The Promptfoo scanner and Copilot reviews are summary-only and raise no defect.

- No material simplification is warranted. A direct unit assertion that `wrapGenerationProvider` receives the snapshotted cached JSON provider would be useful but non-blocking; the implementation visibly preserves this wrapper, adjacent unit tests cover wrapper behavior, and the direct source probe below exercised this exact branch.

Validation:
- `git diff --check origin/main...HEAD` — passed.
- Node 24.18.0 / npm 11.16.0, exact-head disposable worktree with a known compatible dependency tree: five changed/adjacent suites — 233 passed.
- Relevant preview-route handoff tests — 2 passed, 43 skipped.
- Direct source-module provider-switch probe — request A retained `first-json` after the cache switched to `second-json`; the generation wrapper received that same provider and its `callApi` result came from `first-json`.
- An initial combined fallback-dependency run passed 277/278 tests; one unrelated route test returned 502 under the older reused dependency environment. Focused reruns of the changed and handoff surfaces passed, and exact-lock exact-head CI passed all 278 relevant tests plus the full matrix.

Limitations:
- The current lockfile could not be installed locally because the organization Socket policy blocked `postcss@8.5.23`; exact-lock coverage comes from the fully green exact-head CI matrix.
- The requested PR Review Toolkit command was unavailable on this host (exit 127), so its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialist reviews.
- No credentialed provider call or real overlapping HTTP requests were used; the cache-replacement interleaving was exercised deterministically against the real source modules.

mldangelo:
@codex review

mldangelo-oai:
@codex review

PR #10244, opened 2026-07-30
feat(redteam): add Turkish letter coverage to homoglyph map

fevziegeyurtsevenler:
## What

Extends `homoglyphMap` in `src/redteam/strategies/homoglyph.ts` to cover the Turkish letters `ç Ç ş Ş ğ Ğ ö Ö ı`.

## Why

`homoglyphMap` currently only contains ASCII `a–z`, `A–Z`, and `0–9`. `toHomoglyphs()` passes every other character through unchanged, so a Turkish-language prompt has its ASCII characters substituted while its language-specific letters (`ç`, `ş`, `ğ`, `ö`, `ı`) are left untouched. That leaves an obvious, unsubstituted signal in the payload and reduces the strategy's coverage for Turkish inputs.

## How

Each added letter maps to a visually similar character with a different code point, matching the existing entries in the map — e.g. `ş` → `ș` (Latin small s with comma below, U+0219), `ö` → `ӧ` (Cyrillic small o with diaeresis, U+04E7), `ç` → `ҫ` (Cyrillic small es with descender, U+04AB).

The Turkish dotless `ı` (U+0131) has no close single-code-point cross-script twin. The mathematical italic dotless i (U+1D6A4) is used as the nearest match; it renders in a math style, so its visual similarity is weaker than the other entries. This tradeoff is called out in the code and docs.

The letters `ü`, `Ü`, and `İ` are deliberately left out. Their only visually identical alternatives are the NFD (base letter + combining mark) decompositions, which recompose to the original precomposed character under NFC normalization and therefore produce no homoglyph effect. Adding a no-op entry would be misleading, so they are omitted until a genuine cross-script confusable is identified.

## Blast radius

Every added key is a non-ASCII character, so the output for any input that does not contain these specific letters is byte-identical to before. Behaviour for existing ASCII `a–z`/`A–Z`/`0–9` inputs is unchanged. A regression test asserts that other non-ASCII letters outside the map (`ß`, `ñ`, Greek, and the unmapped `ü`/`Ü`/`İ`) still pass through untouched.

## Not a duplicate of the Unicode normalization work

This is orthogonal to the open Unicode-normalization strategy PR (#10056). That PR *normalizes* input as a defense; this change *generates* homoglyph payloads (the attack side of the same coin). It touches only the existing `homoglyph` strategy's lookup table and adds no new strategy, file, or config surface.

## Tests

Added focused cases in `test/redteam/strategies/homoglyph.test.ts`:

- Every added Turkish letter is mapped to a distinct, non-identical homoglyph.
- A Turkish prompt that previously passed through unchanged is now transformed, and none of the mapped Turkish letters survive in the output.
- Non-ASCII letters outside the map — including the intentionally unmapped `ü`/`Ü`/`İ` — are left untouched.

## Docs

Updated `site/docs/red-team/strategies/homoglyph.md` with two example rows and a short paragraph describing the Turkish letter coverage, including the `ı` tradeoff and why `ü`/`Ü`/`İ` are not mapped.

### DIFF
--- a/site/docs/red-team/strategies/homoglyph.md
+++ b/site/docs/red-team/strategies/homoglyph.md
@@ -48,14 +48,14 @@ The homoglyph strategy works by:
 
 ## Example Scenarios
 
-| Original Character | Homoglyph Replacement | Unicode Name                      |
-| ------------------ | --------------------- | --------------------------------- |
-| a                  | а                     | Cyrillic Small Letter A (U+0430)  |
-| e                  | е                     | Cyrillic Small Letter Ie (U+0435) |
-| o                  | о                     | Cyrillic Small Letter O (U+043E)  |
-| p                  | р                     | Cyrillic Small Letter Er (U+0440) |
-| x                  | х                     | Cyrillic Small Letter Ha (U+0445) |
-| ş                  | ș                     | Latin Small Letter S With Comma Below (U+0219) |
+| Original Character | Homoglyph Replacement | Unicode Name                                    |
+| ------------------ | --------------------- | ----------------------------------------------- |
+| a                  | а                     | Cyrillic Small Letter A (U+0430)                |
+| e                  | е                     | Cyrillic Small Letter Ie (U+0435)               |
+| o                  | о                     | Cyrillic Small Letter O (U+043E)                |
+| p                  | р                     | Cyrillic Small Letter Er (U+0440)               |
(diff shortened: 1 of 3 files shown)


fevziegeyurtsevenler:
Gentle ping — CI is fully green (34/34 checks, codecov patch covered) and the change is a small, self-contained map extension (+61/−7, no overlap with the #10056 Unicode-normalization work as noted in the description). Happy to rebase or adjust if anything is needed for review.

PR #10248, opened 2026-07-30
fix(matchers): ignore grader chatter in context-relevance numerator

### LINKED ISSUE #10245: context-relevance score inflated: grader's response header counted as an extracted sentence
### Description

`context-relevance` inflates its score because the grader's own free-text response header (e.g. `"candidate sentences:"`) is counted as an extracted relevant sentence, alongside the actual relevant sentence(s).

### Reproduction

```yaml
prompts:
  - |
    Answer the question using only the provided context.

    Context: {{context}}

    Question: {{query}}

providers:
  - google:gemini-3.1-flash-lite

tests:
  - vars:
      query: 'Who created Python?'
      context: 'Python is a high-level, general-purpose programming language known for readability. It has a large standard library. It was created by Guido van Rossum and released in 1991. Python is widely used in data science, web development, and automation.'
    assert:
      - type: context-relevance
        threshold: 0.5
```

### Observed

`gradingResult.componentResults[].metadata` for the `context-relevance` assertion:

```json
{
  "graderOutputs": {
    "final": "candidate sentences:\nIt was created by Guido van Rossum and released in 1991."
  },
  "extractedSentences": [
    "candidate sentences:",
    "It was created by Guido van Rossum and released in 1991."
  ],
  "totalContextUnits": 4,
  "totalContextSentences": 4,
  "relevantSentenceCount": 2,
  "score": 0.5
}
```

Only **one** of the four context sentences ("It was created by Guido van Rossum and released in 1991.") is actually relevant to the query "Who created Python?" The other three sentences (readability, standard library, use cases) are unrelated. So the expected score is `1/4 = 0.25`.

Instead, `relevantSentenceCount` is reported as `2`, because the parser appears to split the grader LLM's raw output by newline and count each resulting line as an extracted sentence — including the literal header/preamble line `"candidate sentences:"` that the grader model prepended to its own answer. That header isn't one of the `contextUnits` and isn't relevant content at all, but it gets counted toward `relevantSentenceCount`, doubling the numerator and inflating the score from `0.25` to `0.5`.

This is significant in practice because it can flip a test from failing to passing right around common thresholds (e.g. `0.5`), depending only on incidental formatting of the grader's raw text response rather than the actual content.

### Expected

`extractedSentences` (and therefore `relevantSentenceCount`) should only include sentences that are actually present in `contextUnits`, filtering out any preamble/header text the grader model includes in its raw response (e.g. `"candidate sentences:"`, `"relevant sentences:"`, etc.), or the parser should more strictly validate that each extracted line matches an actual context unit before counting it.

### Environment

- promptfoo version: 0.121.19
- Node version: v22.22.3
- Provider under test: `google:gemini-3.1-flash-lite`
- Platform: darwin/arm64

axits-lab:
Fixes #10245.

## Problem

The numerator is segmented from the grader's *entire* raw completion:

```ts
const relevantSentences = insufficientInformation
  ? []
  : [...new Set(segmentRelevant(resp.output))];
const numerator = Math.min(relevantSentences.length, totalContextUnits);
```

so anything the grader emits counts toward relevance. The rubric ends with a `candidate sentences:` cue and chat models routinely echo it back before answering, which segments into a unit of its own.

For the reported case that produces `extractedSentences: ["candidate sentences:", "It was created by Guido van Rossum and released in 1991."]` and a score of `2/4 = 0.5`, where only one of the four context sentences is relevant and `1/4 = 0.25` is correct.

The denominator is derived from the context while the numerator was derived from free text, so the two sides of the ratio weren't drawn from the same population — and the score varied with how chatty the grading model happened to be, since no part of the rubric pins down the header wording.

## Fix

Only units that actually occur in the context count. The rubric already requires verbatim extraction — *"you're not allowed to make any changes to sentences from given context"* — so anything absent from the context is the grader talking rather than extracting. That drops preambles and commentary without hard-coding particular header strings, which would only work for models that phrase them the same way.

Matching is **normalised containment**, not equality, because graders in practice:

- echo a sentence without its terminal punctuation — the existing `line-based sentence splitting` test mocks exactly this
- prefix it with a list marker (`1.`, `-`) — covered by the existing numbered-list test
- alter whitespace, or echo several sentences from one pre-segmented unit

A leading list marker is stripped before matching, and only when followed by whitespace, so a sentence genuinely opening with a number (`1991 saw…`) is untouched.

## Tests

Added a regression test using the reported scenario: a grader output carrying the `candidate sentences:` header over the four-sentence Python context, asserting `extractedSentences` holds only the real extraction and the score is `0.25`.

Verified it fails without the source change:

```
× should not count the grader echoing the prompt cue as a relevant sentence
  AssertionError: expected [ 'candidate sentences:', …(1) ] to deeply equal [ Array(1) ]
```

Worth noting the first iteration of this fix used exact matching and **broke the existing numbered-list test** — that's what surfaced the need to strip list markers and match on containment.

## Verification

- `341` tests pass across `test/matchers/` and `test/assertions/contextRelevance.test.ts` (22 files)
- `npm run l` and `npm run f` clean
- `npm run tsc` reports 7 pre-existing errors (`lru-cache`, `node-sql-parser` declarations) — identical set on a clean tree with this patch stashed, and none in the touched files

### DIFF
--- a/src/matchers/rag.ts
+++ b/src/matchers/rag.ts
@@ -282,11 +282,37 @@ export async function matchesContextRelevance(
   // text with no newlines, so counting by line alone would treat a multi-sentence
   // answer as one unit and undercount relevance (e.g. echoing the whole context
   // would score 1/N instead of 1.0).
+  //
+  // Only units that actually occur in the context count. The rubric requires verbatim
+  // extraction ("you're not allowed to make any changes to sentences from given
+  // context"), so anything absent from the context is the grader talking rather than
+  // extracting. Most often that is the rubric's own trailing `candidate sentences:`
+  // cue echoed back by chat models, which segmented into a unit and inflated the
+  // numerator — the denominator is derived from the context, so counting free text on
+  // the other side of the ratio made the score model-dependent.
+  //
+  // Matching is on normalised containment, not equality: graders routinely echo a
+  // sentence without its terminal punctuation or with altered whitespace, and a
+  // pre-segmented context may hold several sentences per unit.
   const insufficientInformation = resp.output.includes(CONTEXT_RELEVANCE_BAD);
   const segmentRelevant = contextIsPreSegmented ? splitIntoSentences : splitTextIntoSentences;
+  // A leading list marker is the grader's formatting, not part of the extracted
+  // sentence, so it is dropped before matching — `1. Paris is…` must still match
+  // `Paris is…` in the context. The marker must be followed by whitespace, so a
+  // sentence that genuinely opens with a number ("1991 saw…") is left alone.
+  const normalizeForMatch = (text: string) =>
+    text
+      .replace(/^\s*(?:[-*•]|\d+[.)])\s+/, '')
+      .toLowerCase()
+      .replace(/[^a-z0-9]+/g, ' ')
+      .trim();
+  const normalizedContext = normalizeForMatch(contextString);
   const relevantSentences = insufficientInformation
     ? []
-    : [...new Set(segmentRelevant(resp.output))];
+    : [...new Set(segmentRelevant(resp.output))].filter((sentence) => {
+        const normalized = normalizeForMatch(sentence);
+        return normalized.length > 0 && normalizedContext.includes(normalized);
+      });
   // Cap at the total so the score never exceeds 1.
   const numerator = Math.min(relevantSentences.length, totalContextUnits);
 
--- a/src/matchers/rag.ts
+++ b/src/matchers/rag.ts
@@ -323,10 +323,28 @@ export async function matchesContextRelevance(
   const insufficientInformation = resp.output.includes(CONTEXT_RELEVANCE_BAD);
   const segmentRelevant = contextIsPreSegmented ? splitIntoSentences : splitTextIntoSentences;
   const normalizedContext = normalizeForMatch(contextString);
-  const isFromContext = (text: string) => {
+  const matchesContext = (text: string) => {
     const normalized = normalizeForMatch(text);
     return normalized.length > 0 && normalizedContext.includes(normalized);
   };
+
+  // A grader may put its cue on the same line as the first extraction
+  // ("candidate sentences: Alpha is first."). The label is only dropped when doing so
+  // turns a non-matching segment into a verbatim context sentence, so a context
+  // sentence that legitimately contains a colon is left alone.
+  const withoutInlineLabel = (text: string) => {
+    const stripped = text.replace(/^[^:\n]{0,80}:\s+/, '');
+    return stripped !== text && matchesContext(stripped) ? stripped : text;
+  };
+
+  // A segment that is nothing but a list marker carries no content. Sentence-mode
+  // splitting leaves `A.` standing alone, and it normalises to `a` — a substring of
+  // very nearly any context — so without this it would be counted as a relevant
+  // sentence and inflate the numerator, which is the failure this filter exists to
+  // prevent. `splitTextIntoSentences` already drops the numeric form.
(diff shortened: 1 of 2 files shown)


axits-lab:
Thanks — all three were real, and each is fixed in b1b6c7d with a regression test that fails on the previous commit.

**Unicode letters.** Correct, and the worst of the three: normalising with `[a-z0-9]` reduced a CJK/Arabic/Cyrillic sentence to the empty string, so it failed the membership check and every extraction in a non-Latin eval scored 0. Normalisation now uses `\p{L}`/`\p{N}`.

Worth recording how close this came to shipping mis-tested: my first regression test used `Python 是一种…1991年发布` and **passed against the broken code**, because the ASCII `Python` and `1991` survived the old class and kept the string non-empty. The test now uses text with no ASCII at all.

**Re-splitting after dropping the cue.** Also correct, and the subtler failure. Confirmed the exact case you described:

```
context : Alpha is first. Beta is second. Gamma is third.
grader  : candidate sentences:\nAlpha is first. Beta is second.
before  : numerator 1  ->  1/3
after   : numerator 2  ->  2/3
```

The echoed cue makes the response two lines, which puts `splitTextIntoSentences` into line mode for the whole thing, so the surviving line counted once. Chatter is now removed *before* segmentation, which restores the segmentation the response would have had without it — I took that over re-splitting survivors, since it keeps one segmentation decision rather than two.

**Broader list markers.** Fixed: `(1)`, `A.`, `a)` and en/em dashes are now recognised. One deliberate narrowing — alphabetic markers are ASCII-only. `\p{L}[.)]` would treat a CJK sentence opening with a single character as a label and strip a real word, which matters more now that non-Latin input is actually reachable.

Net effect is that my original filter traded an over-count for three under-counts, which is the worse direction for an assertion. Verification: 344 tests pass across `test/matchers/` and `test/assertions/contextRelevance.test.ts`; `npm run l` and `npm run f` clean; `npm run tsc` still at its 7 pre-existing errors (`lru-cache`, `node-sql-parser`), none in the touched files.

axits-lab:
Both correct again, fixed in 8dfe5cc.

**Inline cue.** Confirmed — the cue stays attached through segmentation, so `candidate sentences: Alpha is first.` fails the membership check and a real extraction is lost: 2/3 → 1/3. The label is now stripped, but only when doing so turns a non-matching segment into a verbatim context sentence, so a context sentence that legitimately contains a colon is left alone.

**Bare letter marker.** This one is the worse of the two, because it fails in the direction this PR exists to prevent. `A.` splits into its own segment and normalizes to `a`, which is a substring of nearly any context, so it passed the filter — one real extraction in a two-sentence context scored 2/2. Marker-only segments are now rejected (`splitTextIntoSentences` already drops the numeric form).

Rather than keep patching per finding, I enumerated the grader-output shape space and ran the implementation against all of it. Before this commit, exactly two shapes were wrong — the two you flagged, no others:

```
FAIL  cue inline             got 1  want 2
FAIL  alpha marker inline    got 2  want 1
```

After, all 19 pass: plain single/multi, cue on its own line, cue inline, numeric / paren-numeric / alpha / dash markers, a bare marker alone, blank lines, leading and trailing commentary, quoted, markdown-bold, missing terminal punctuation, duplicate echoes, a colon inside a context sentence, and the whole context echoed back.

That enumeration is what I should have done before opening the PR — each of your five findings was a shape I hadn't listed, and every one of them failed silently as a wrong score rather than an error. Both new cases have a regression test that fails on the previous commit.

346 tests pass across `test/matchers/` and `test/assertions/contextRelevance.test.ts`; lint and format clean; `tsc` unchanged at its 7 pre-existing errors.

PR #10249, opened 2026-07-30
fix(deps): restore XML entity decoding for xlsx test files

mldangelo:
## Summary

Loading test cases from an Excel file silently corrupts every cell containing `&`, `<`, `>`, `"` or `'`. A cell holding `AT&T <b>` loads as `AT&amp;T &lt;b&gt;`.

`read-excel-file` 9.3.0 replaced its DOM XML parser with a SAX parser and stopped decoding XML character entities. `src/util/xlsx.ts` is unchanged — the regression arrived entirely through the lockfile.

## Impact

Anyone using `tests: file://tests.xlsx` gets corrupted vars, prompts, and expected values. Exact-match assertions fail against text that looks correct in Excel, and prompts get literal `&amp;` injected. There is no error — the eval just silently grades the wrong strings.

Column headers are affected too, so an undecoded header renames the variable a test case binds to.

## Exposure

| `read-excel-file` | published | entity decoding |
| --- | --- | --- |
| 9.2.0 — shipped in 0.121.19 | Jun 11 | ✅ |
| **9.3.0 – 9.3.2** | Jul 10 – Jul 16 | ❌ |
| 9.3.3+ | Jul 20 | ✅ fixed upstream |

The lockfile moved 9.2.0 → 9.3.1 in #10161 and → 9.3.2 in #10226, both after the 0.121.19 tag, so this is unreleased. `Dockerfile` installs with `npm ci`, so images built from the current lockfile ship the broken parser. npm consumers resolving `^9.0.0` fresh happen to self-heal today, which is exactly why the floor needs raising rather than just the lock.

## Change

- Raise the `read-excel-file` floor from `^9.0.0` to `^9.3.3` in `devDependencies` and `optionalDependencies` so the broken range can never be resolved again.
- Refresh the lock to 9.3.5.
- Add an unmocked regression test.

The lockfile diff is limited to `read-excel-file` and its transitive deps (`saxen`, `unzipper-esm`, `worker-f`); no unrelated drift.

## Why CI missed it

`test/util/xlsx.test.ts` calls `vi.mock('read-excel-file/node', ...)`, so the real parser is never exercised and a lockfile-only bump cannot fail any test.

`test/util/xlsxEntities.test.ts` closes that gap by parsing a real fixture workbook (`test/fixtures/xlsx-xml-entities.xlsx`) whose shared-strings table stores the escaped forms.

## Test plan

Verified the new test is a genuine guard by pinning the dependency both ways:

```
9.3.2  ->  3 failed   (expected [ 'expected', 'q&amp;a' ] to deeply equal [ 'expected', 'q&a' ])
9.3.5  ->  3 passed
```

- `npx vitest run test/util/xlsxEntities.test.ts test/util/xlsx.test.ts` — 27 passed
- `npm run tsc` — clean
- Biome clean on the new file


mldangelo-oai:
@codex review

mldangelo-oai (approved):
Approved exact head 1630a66e after rebasing via an additive merge from current main, replacing unapproved read-excel-file 9.3.5 with approved and integrity-verified 9.3.4, preserving the published ^9.3.3 safe floor, adding a manifest floor regression guard, reproducing escaped XML with vulnerable 9.3.1 and corrected headers/values with 9.3.4, passing 49 focused tests and TypeScript, and completing a real uncached CLI eval from the actual XLSX fixture.

mldangelo-oai:
@codex review

mldangelo-oai (approved):
Approved refreshed exact head f4d3ff20 after incorporating the concurrently merged read-excel-file 9.3.4 lock update. This follow-up adds the missing published ^9.3.3 safe floor, aligned root manifest/lock guards, real XLSX XML-entity regression coverage, and a successful uncached CLI eval with decoded headers and cell values.

mldangelo-oai:
@codex review

mldangelo-oai (approved):
Re-approved exact head 18b04552 after incorporating the concurrently merged Hono security fix; Excel remains on approved 9.3.4 with the protected ^9.3.3 published floor, manifest alignment guard, real XML workbook regression tests, and successful uncached CLI evaluation.

mldangelo-oai:
@codex review

mldangelo-oai (approved):
Approved exact refreshed head 5bd41497 after incorporating the independently landed socket.io-parser security patch; preserved the approved Excel parser, protected ^9.3.3 published floor, manifest guard, real workbook regression tests, and verified CLI evaluation.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #10250, opened 2026-07-30
fix(deps): pin PostCSS to approved release

mldangelo-oai:
## Summary
- Pin PostCSS to the policy-approved 8.5.22 release in the root override, app manifest, and lockfile.
- Preserve security-patched js-yaml 5.2.2; that package still needs a Socket policy exception.

## Validation
- npm run tsc
- npx vitest run test/package-manifests.test.ts test/util/yamlLoad.test.ts (22 tests passing)
- npx @biomejs/biome check package.json src/app/package.json
- Confirmed postcss@8.5.22 downloads through the configured Socket registry.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34437,9 +34437,9 @@
       }
     },
     "node_modules/postcss": {
-      "version": "8.5.23",
-      "resolved": "https://registry.npmjs.org/postcss/-/postcss-8.5.23.tgz",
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
@codex review

mldangelo-oai:
Thanks for putting this mitigation together. The approved package registry now serves PostCSS 8.5.23, and main already installs that newer release, so the temporary pin to 8.5.22 is no longer needed. Closing to avoid downgrading PostCSS.

PR #10251, opened 2026-07-30
fix(assertions): ignore grader prose in context-relevance scoring

### LINKED ISSUE #10245: context-relevance score inflated: grader's response header counted as an extracted sentence

mldangelo:
Fixes #10245

## Summary

`context-relevance` counts the grader's own response header as an extracted relevant sentence, inflating the score.

The grading prompt does not constrain the response format, so graders routinely prefix their answer with a header of their own — `candidate sentences:`, `Relevant sentences:`. Every segment of the grader response counted toward the numerator, and that header is a segment like any other.

## Reproduction

Four-sentence context, exactly one sentence answers the question:

```yaml
tests:
  - vars:
      query: 'Who created Python?'
      context: 'Python is a high-level, general-purpose programming language known for readability. It has a large standard library. It was created by Guido van Rossum and released in 1991. Python is widely used in data science, web development, and automation.'
    assert:
      - type: context-relevance
        threshold: 0.5
```

Grader replies:

```
candidate sentences:
It was created by Guido van Rossum and released in 1991.
```

| | before | after |
| --- | --- | --- |
| `extractedSentences` | `["candidate sentences:", "It was created by…"]` | `["It was created by…"]` |
| `relevantSentenceCount` | 2 | 1 |
| `score` | 0.5 | 0.25 |
| `threshold: 0.5` | **pass** | **fail** |

The assertion passed only because of incidental formatting in the grader's prose, not because of the content of the context.

## Fix

A relevant unit must actually come from the context, so segments that do not appear in it are dropped.

Segments are matched against the **whole whitespace-normalized context** rather than against individual context units. This matters: a grader legitimately quotes a span that crosses a unit boundary — a wrapped line in a formatted document, or two sentences of a single chunk. Per-unit matching would undercount those and trade one scoring bug for another. The existing `Formatted Document Test` covers exactly that case, where the grader quotes `The scope of this policy covers entitlement and guidelines for all categories of leaves in the office` — text that spans a line wrap in the source.

Leading list markers (`1.`, `-`, `*`), surrounding quotes, and trailing punctuation are normalized away, so filtering grader prose does not discard genuinely relevant sentences the grader reformatted.

## Test plan

Four regression cases added. Verified they are genuine guards by reverting `src/matchers/rag.ts` and re-running:

```
without fix -> 3 failed  (expected 0.5 to be close to 0.25; expected 3 to be 1; expected 3 to be 2)
with fix    -> 20 passed
```

The fourth case (grader echoes a sentence without its trailing period) passes both ways by design — it guards against over-correcting into an undercount.

- All 16 pre-existing `context-relevance` cases pass unchanged
- `npx vitest run test/matchers/ test/assertions/` — 60 files, 1324 passed
- `npm run tsc` — clean
- Biome clean on both changed files

### DIFF
--- a/src/matchers/rag.ts
+++ b/src/matchers/rag.ts
@@ -225,6 +225,32 @@ export async function matchesContextRecall(
   };
 }
 
+// Leading enumeration/bullet markers a grader may add when listing the sentences
+// it selected ("1. ", "2) ", "- ", "* "). They are formatting, not content, so
+// they are stripped before matching a segment back to the context.
+const LEADING_LIST_MARKER = /^\s*(?:\d+[.)]|[-*•])\s+/;
+// Quotes a grader may wrap an echoed sentence in, and trailing punctuation it may
+// add or drop. Both sides of a match are normalized so `Paris is the capital of
+// France` still matches a context reading `Paris is the capital of France.`
+const SURROUNDING_QUOTES = /^["'`]+|["'`]+$/g;
+const TRAILING_PUNCTUATION = /[\s.,;:!?]+$/;
+
+/** Lowercase and collapse all whitespace (including newlines) so a quoted span that
+ * wraps across lines in the source context still matches. */
+function normalizeForContextMatch(text: string): string {
+  return text.toLowerCase().replace(/\s+/g, ' ').trim();
+}
+
+/** Normalize one grader-emitted segment for comparison against the context. */
+function normalizeSegmentForContextMatch(segment: string): string {
+  return normalizeForContextMatch(
+    segment
+      .replace(LEADING_LIST_MARKER, '')
+      .replace(SURROUNDING_QUOTES, '')
+      .replace(TRAILING_PUNCTUATION, ''),
+  );
+}
+
 export async function matchesContextRelevance(
   question: string,
   context: string | string[],
@@ -284,9 +310,25 @@ export async function matchesContextRelevance(
   // would score 1/N instead of 1.0).
   const insufficientInformation = resp.output.includes(CONTEXT_RELEVANCE_BAD);
   const segmentRelevant = contextIsPreSegmented ? splitIntoSentences : splitTextIntoSentences;
+  // A relevant unit must actually come from the context. The grading prompt does
+  // not constrain the response format, and graders routinely prefix their answer
+  // with a header of their own ("candidate sentences:", "Relevant sentences:").
+  // That header is a segment like any other, so counting it inflated the
+  // numerator: a header plus one genuinely relevant sentence scored 2/4 instead
+  // of 1/4, which flips a threshold-0.5 assertion from fail to pass purely on
+  // incidental formatting of the grader's prose.
+  //
+  // Segments are matched against the whole whitespace-normalized context rather
+  // than against individual units, because a grader legitimately quotes text
+  // that spans a unit boundary — a wrapped line in a formatted document, or two
+  // sentences of one chunk. Matching per unit would undercount those.
+  const normalizedContext = normalizeForContextMatch(contextUnits.join(' '));
   const relevantSentences = insufficientInformation
     ? []
-    : [...new Set(segmentRelevant(resp.output))];
+    : [...new Set(segmentRelevant(resp.output))].filter((sentence) => {
+        const normalized = normalizeSegmentForContextMatch(sentence);
+        return normalized.length > 0 && normalizedContext.includes(normalized);
+      });
   // Cap at the total so the score never exceeds 1.
   const numerator = Math.min(relevantSentences.length, totalContextUnits);
 
--- a/test/matchers/context-relevance.test.ts
+++ b/test/matchers/context-relevance.test.ts
@@ -438,4 +438,81 @@ This policy excludes all staff going on any outgoing structured programs, short 
       expect(result.metadata?.relevantSentenceCount).toBe(1);
     });
   });
+
+  describe('Grader prose is not a context unit', () => {
+    // Regression (#10245): the grading prompt does not constrain the response
+    // format, so graders routinely prefix their answer with a header of their
+    // own. Every segment of the response used to count toward the numerator, so
+    // that header counted as an extracted relevant sentence and inflated the
+    // score — flipping a test from fail to pass purely on incidental formatting.
+    const PYTHON_CONTEXT =
+      'Python is a high-level, general-purpose programming language known for readability. It has a large standard library. It was created by Guido van Rossum and released in 1991. Python is widely used in data science, web development, and automation.';
+
+    it('does not count the grader response header as a relevant sentence', async () => {
+      const mockCallApi = vi.fn().mockResolvedValue({
+        output: 'candidate sentences:\nIt was created by Guido van Rossum and released in 1991.',
+        tokenUsage: { total: 10, prompt: 5, completion: 5 },
+      });
+      vi.spyOn(DefaultGradingProvider, 'callApi').mockImplementation(mockCallApi);
+
+      const result = await matchesContextRelevance('Who created Python?', PYTHON_CONTEXT, 0.5);
+
+      // Exactly one of the four context sentences answers the question, so the
+      // score is 1/4. Counting the "candidate sentences:" header doubled the
+      // numerator to 2/4 = 0.5, which passed the 0.5 threshold.
+      expect(result.score).toBeCloseTo(0.25, 2);
+      expect(result.pass).toBe(false);
+      expect(result.metadata?.totalContextUnits).toBe(4);
+      expect(result.metadata?.relevantSentenceCount).toBe(1);
+      expect(result.metadata?.extractedSentences).toEqual([
+        'It was created by Guido van Rossum and released in 1991.',
+      ]);
+    });
+
+    it('drops grader commentary that does not appear in the context', async () => {
+      const mockCallApi = vi.fn().mockResolvedValue({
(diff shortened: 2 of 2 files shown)


src/matchers/rag.ts
```diff
@@ -284,9 +310,25 @@ export async function matchesContextRelevance(
...
+  //
+  // Segments are matched against the whole whitespace-normalized context rather
+  // than against individual units, because a grader legitimately quotes text
+  // that spans a unit boundary — a wrapped line in a formatted document, or two
+  // sentences of one chunk. Matching per unit would undercount those.
+  const normalizedContext = normalizeForContextMatch(contextUnits.join(' '));
   const relevantSentences = insufficientInformation
     ? []
-    : [...new Set(segmentRelevant(resp.output))];
+    : [...new Set(segmentRelevant(resp.output))].filter((sentence) => {
```
jameshiester-oai, line 328: **[P1] Deduplicate after normalization.** `new Set(...)` sees raw formatting, so two variants of the same selected unit survive and both increment the numerator. I reproduced a four-unit context with `1. "It was created…"` and `2. It was created…`: this returns `relevantSentenceCount: 2`, score `0.50`, and passes threshold `0.5`, although only one distinct context unit was selected and the correct score is `0.25`. Normalize first and count each normalized source unit/span at most once.

src/matchers/rag.ts
```diff
@@ -284,9 +310,25 @@ export async function matchesContextRelevance(
...
+        const normalized = normalizeSegmentForContextMatch(sentence);
+        return normalized.length > 0 && normalizedContext.includes(normalized);
```
jameshiester-oai, line 330: **[P1] Substring membership still admits grader prose.** A heading only has to occur somewhere in the context to count as a unit. With four Python context sentences, grader output `Python:\nIt was created by Guido van Rossum…` counts both lines, returns `0.50`, and passes a `0.5` threshold instead of returning `0.25`. This is the same false-pass class the PR is intended to remove. Map extracted spans to distinct source units (while explicitly supporting valid cross-boundary spans) rather than accepting any non-empty substring.

src/matchers/rag.ts
```diff
@@ -225,6 +225,32 @@ export async function matchesContextRecall(
...
+// Quotes a grader may wrap an echoed sentence in, and trailing punctuation it may
+// add or drop. Both sides of a match are normalized so `Paris is the capital of
+// France` still matches a context reading `Paris is the capital of France.`
+const SURROUNDING_QUOTES = /^["'`]+|["'`]+$/g;
+const TRAILING_PUNCTUATION = /[\s.,;:!?]+$/;
+
+/** Lowercase and collapse all whitespace (including newlines) so a quoted span that
+ * wraps across lines in the source context still matches. */
+function normalizeForContextMatch(text: string): string {
+  return text.toLowerCase().replace(/\s+/g, ' ').trim();
```
jameshiester-oai, line 241: **[P2] Canonically normalize Unicode before comparison.** Lowercasing and whitespace folding do not make canonically equivalent text equal. An exact-head probe with decomposed source text `Cafe\u0301 is relevant.` and grader output `Café is relevant.` returns count/score `0` instead of `1/2`. This is plausible when retrieved PDF/web text and model output use different normalization forms. Apply the same NFC (or deliberately chosen NFKC) normalization to context and grader segments, with a regression test.

jameshiester-oai (commented):
Reviewed base `ea63e1ccb4d5b9915e82d13366461fb367925709` through exact head `eccbcf8a026756426c31acc796c83d5fbe5d04d4`.

Scope: fresh full review of the current two-file PR diff.

Verdict: the containment filter fixes the demonstrated newline-delimited grader-header inflation, but it is not safe to ship as-is. I verified two remaining false-pass paths and one new Unicode false-negative path in the changed matching logic.

CI: all required exact-head checks are green across supported Node versions and Linux, macOS, and Windows; the aggregate CodeQL status is neutral while its language analysis jobs succeeded.

- P1: formatting variants of one selected context unit are counted multiple times because raw segments are deduplicated before normalization. A four-unit context with one sentence repeated as two formatted list items scores 0.50 and passes a 0.5 threshold instead of scoring 0.25.
- P1: any grader heading that is a substring of the context still counts as a relevant unit. `Python:` plus one valid sentence scores 0.50 for a four-unit Python context instead of 0.25.
- P2: composed/decomposed Unicode-equivalent text does not match, so `Café` from the grader is rejected against source text containing `Cafe\u0301`.

Existing feedback:
- Four unresolved current-head threads remain applicable and were not duplicated: quotes wrapping a numbered item; inline header plus sentence; punctuation outside a closing quote; and nonadjacent selections emitted on one line.

Simplification and test-gap opportunities:
- The correction should map normalized grader spans to distinct normalized source units/spans, rather than use unconstrained substring membership followed by a numeric cap. That single model prevents duplicate inflation and content-overlapping headings while preserving deliberate cross-boundary matches.
- Add regression cases for normalized duplicates, content-overlapping headers, and Unicode canonical equivalence.

Validation:
- `npx vitest run test/matchers/context-relevance.test.ts --no-cache` — 20/20 passed.
- Disposable Vitest probe — 3/3 reproduced normalized duplicate inflation, content-overlapping-header inflation, and Unicode rejection.
- Credential-free local CLI eval with `--no-cache --no-share` and exported JSON — 1/1 passed incorrectly at score 0.50; the output records the same source sentence twice and `relevantSentenceCount: 2` for four context units.
- Three independent read-only specialist passes converged on the false-pass paths and confirmed the existing threads.

Limitations:
- No live external LLM grader or custom-rubric corpus was exercised.
- The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable on this host (exit 127), so its six lenses were applied manually and reconciled with the specialists.

mldangelo (commented):
Heads-up: this and #10248 both say "Fixes #10245" and both modify `src/matchers/rag.ts` and `test/matchers/context-relevance.test.ts`, so they'll conflict — one should be closed.

Comparing the normalization logic:

| | #10248 | this PR |
|---|---|---|
| list markers | `-` `*` `•` `–` `—`, `1.` `(2)`, **`a)` `(b.)`** | `-` `*` `•`, `1.` `2)` |
| non-Latin scripts | `[^\p{L}\p{N}]+` — preserves CJK / Arabic / Cyrillic | whitespace-collapse only |
| punctuation between words | subsumed by the Unicode collapse | not handled |

The last row is the practical difference: `normalizeForContextMatch` here only collapses whitespace, so a context reading `"Paris, the capital of France"` won't match an extracted segment `"Paris the capital of France"`. #10248's `[^\p{L}\p{N}]+ → ' '` handles that, and it also covers lettered/parenthesized markers.

#10248 also documents why it avoids an ASCII-only class ("would drop every extraction in a non-Latin RAG eval and score it 0") and why the marker must be followed by whitespace (so a sentence opening `"1991 saw…"` isn't mangled).

This PR does handle surrounding quotes explicitly, which #10248 gets implicitly via the Unicode collapse — so I don't think anything here is lost by going with #10248.

My read is #10248 is the stronger of the two and it's from an outside contributor, but it's your call.

PR #10253, opened 2026-07-30
fix(cli): correct URL query-parameter credential redaction

mldangelo:
## Summary

The matcher deciding whether a URL query parameter names a credential is wrong in **both** directions, and the two paths that need it had drifted apart.

### Over-redaction — ordinary parameters destroyed

`SENSITIVE_URL_PARAM_NAMES` matched its alternatives as bare substrings, so any parameter whose name merely *contains* one was redacted:

| parameter | matched | because it contains |
| --- | --- | --- |
| `design`, `signal`, `designer`, `redesign`, `significance` | ❌ redacted | `sig` |
| `tokenizer`, `tokens_used` | ❌ redacted | `token` |
| `secretary` | ❌ redacted | `secret` |
| `passwordless_flow` | ❌ redacted | `password` |

`?design=modern` is persisted as `?design=%5BREDACTED%5D`. This runs on any field named `url` via `sanitizeObject`, so it corrupts eval results and debug logs for no security benefit. The pre-existing test even documented it — *"The regex in sanitizeUrl is broad and matches substrings, so these will be redacted"*.

### Under-redaction — real credentials leaked

`sanitizeUrl` checked only that regex; the templated-URL path checked only `isSecretField` (exact `SECRET_FIELD_NAMES`). Each missed what the other caught:

```
my_api_key_param   plain=REDACTED  templated=LEAKED
session_token      plain=REDACTED  templated=LEAKED
auth               plain=LEAKED    templated=REDACTED
```

## Change

- Anchor each alternative on a word boundary (start/end of name, or a non-alphanumeric separator).
- Add the credential names the substring behavior had covered only incidentally: `auth`, `credential`, `bearer`, `private_key`, `signing_key`, `passwd`, `pwd`.
- Route both URL paths through a single `isSensitiveParamName` helper, so plain and templated URLs redact identically.

## What is deliberately *not* changed

`sanitizeUrlEncodedString` keeps its narrower exact-name matching, because it also scrubs request/response **bodies**.

I initially applied the unified matcher there too, and it broke `test/providers/openai-codex-app-server.test.ts`. Coding-agent redteam plugins plant canaries like `PROMPTFOO_SYNTHETIC_SECRET=<value>` and grade on whether the agent echoed them back — the broader matcher redacted the planted canary, which would turn a **detected credential leak into a silent pass**. That boundary now has its own regression test so the mistake cannot be repeated.

## Test plan

- `test/util/sanitizer.test.ts` — 225 passed, including 4 new cases (benign names survive; compound credential names redact in both paths; body canary preserved). One pre-existing test that asserted the over-broad behavior was updated, with a comment explaining why `secret_santa` still redacts but `tokens_available` no longer does.
- `test/providers/openai-codex-app-server.test.ts` — 74 passed.
- Full backend suite: no new failures. The remaining reds (`test/node/retry.test.ts`, `test/providers/google/interactions.test.ts`, `test/providers/moonshot.test.ts`, `test/examples/readme.test.ts`) reproduce identically on clean `main` — they are local-environment artifacts, not related to this change.
- `npm run tsc` clean; Biome clean.

### DIFF
--- a/src/util/sanitizer.ts
+++ b/src/util/sanitizer.ts
@@ -13,8 +13,37 @@ export const REDACTED = '[REDACTED]';
 
 // Query-parameter names that imply a credential value. Shared by sanitizeUrl's
 // per-param redaction and the fail-closed decision for unparseable URLs.
+//
+// Each alternative must sit on a word boundary — the start/end of the name or a
+// non-alphanumeric separator. Matching these as bare substrings redacted any
+// name that merely CONTAINS one: `design`, `signal`, `designer` and
+// `significance` all contain `sig`; `tokenizer` and `tokens_used` contain
+// `token`; `secretary` contains `secret`; `passwordless_flow` contains
+// `password`. Those are ordinary parameters, and destroying them corrupts
+// persisted eval results and debug logs for no security benefit.
 const SENSITIVE_URL_PARAM_NAMES =
-  /(api[_-]?key|token|password|secret|signature|sig|access[_-]?token|refresh[_-]?token|id[_-]?token|client[_-]?secret|authorization)/i;
+  /(?:^|[^a-z0-9])(?:api[_-]?key|apikey|access[_-]?token|refresh[_-]?token|id[_-]?token|client[_-]?secret|authorization|signature|private[_-]?key|signing[_-]?key|credential|password|passwd|bearer|secret|token|auth|pwd|sig)(?:$|[^a-z0-9])/i;
+
+/**
+ * Whether a query-parameter or form-field name implies a credential value.
+ *
+ * Unifies the two matchers that had drifted apart: `sanitizeUrl` checked only
+ * the name regex above, while `sanitizeUrlEncodedString` checked only
+ * {@link isSecretField} (exact names from `SECRET_FIELD_NAMES`). Each caught
+ * credentials the other missed — `auth=` leaked from the plain-URL path, while
+ * `my_api_key_param=` and `session_token=` leaked from the templated path.
+ */
+function isSensitiveParamName(name: string): boolean {
+  if (SENSITIVE_URL_PARAM_NAMES.test(name)) {
+    return true;
+  }
+  // Split nested-key syntax (`user[password]`, `a.b.password`) so the leaf name
+  // can be matched against the shared secret-field list.
+  return name
+    .split(/[.\[\]]+/)
+    .filter(Boolean)
+    .some(isSecretField);
+}
 const OPAQUE_CREDENTIAL_PATH_SEGMENT =
   /^(?:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}|[0-9a-f]{32,}|(?:token|key|secret|credential|auth)[-_][a-z0-9._-]{8,}|eyJ[a-zA-Z0-9_-]*\.[a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+)$/i;
 
@@ -608,6 +637,13 @@ export function sanitizeUrlEncodedString(value: string): string {
     const decodedKey = decodeFormComponent(rawKey);
     // Split nested-key syntax (`user[password]`, `a.b.password`) into parts so we
     // can match the leaf name against SECRET_FIELD_NAMES.
+    //
+    // Deliberately does NOT use the broader isSensitiveParamName() matcher that
+    // sanitizeUrl uses. This function also scrubs request/response BODIES, where
+    // over-redaction is harmful rather than merely noisy: coding-agent redteam
+    // plugins plant canaries such as `PROMPTFOO_SYNTHETIC_SECRET=<value>` and
+    // grade on whether the agent echoed them back. Redacting a planted canary
+    // would turn a detected credential leak into a silent pass.
     const keyParts = decodedKey === undefined ? [] : decodedKey.split(/[.\[\]]+/).filter(Boolean);
     const keyIsSecret = keyParts.some(isSecretField);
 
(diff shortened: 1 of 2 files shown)


src/util/sanitizer.ts
```diff
@@ -820,8 +886,14 @@ function sanitizeTemplatedUrl(url: string): string {
...
-  const sanitizedQuery = query ? sanitizeUrlEncodedString(query) : query;
-  const sanitizedHash = hash ? sanitizeUrlEncodedString(hash) : hash;
+  // sanitizeUrlEncodedString only matches exact SECRET_FIELD_NAMES on the key,
+  // because it also scrubs bodies where over-redaction destroys redteam canaries.
+  // A URL query is not a body, so apply the same param-name matcher the
+  // non-templated branch of sanitizeUrl uses; otherwise a credential leaks into
+  // persisted results purely because its URL also carried a Nunjucks template.
+  const scrubQuery = (value: string) => redactSensitiveNamedPairs(sanitizeUrlEncodedString(value));
+  const sanitizedQuery = query ? scrubQuery(query) : query;
+  const sanitizedHash = hash ? scrubQuery(hash) : hash;
```
jameshiester-oai, line 896: **[P1] Apply the URL-specific matcher to ordinary fragments too.** The templated path now redacts `#my_api_key_param=hunter2`, but `sanitizeUrl('https://example.com/#my_api_key_param=hunter2')` and the same `session_token` form remain unchanged because the ordinary fragment still goes only through the narrow body sanitizer at lines 951–956. `sanitizeObject({ url })` preserves the leak as well. Use the same URL-only named-pair pass for both fragment paths and add paired plain/templated regressions.

src/util/sanitizer.ts
```diff
@@ -13,8 +13,37 @@ export const REDACTED = '[REDACTED]';
...
+/**
+ * Whether a query-parameter or form-field name implies a credential value.
+ *
+ * Unifies the two matchers that had drifted apart: `sanitizeUrl` checked only
+ * the name regex above, while `sanitizeUrlEncodedString` checked only
+ * {@link isSecretField} (exact names from `SECRET_FIELD_NAMES`). Each caught
+ * credentials the other missed — `auth=` leaked from the plain-URL path, while
+ * `my_api_key_param=` and `session_token=` leaked from the templated path.
+ */
+function isSensitiveParamName(name: string): boolean {
```
jameshiester-oai, line 36: **[P2] Reuse this matcher in the parse-failure path.** Protocol-relative and malformed URLs bypass the new helper because `unparseableUrlMightLeakSecret` delegates to `hasSecretFormSegment`, which still checks only exact `SECRET_FIELD_NAMES`/parts. Exact-head repro: `//example.com?my_api_key_param=hunter2` and `ht!tp://bad?private_key_param=hunter2` are returned unchanged, while valid absolute equivalents redact. Feed decoded fallback keys through the unified URL matcher so common protocol-relative URLs fail closed on credential-named values.

src/util/sanitizer.ts
```diff
@@ -13,8 +13,37 @@ export const REDACTED = '[REDACTED]';
...
+  if (SENSITIVE_URL_PARAM_NAMES.test(name)) {
+    return true;
+  }
+  // Split nested-key syntax (`user[password]`, `a.b.password`) so the leaf name
+  // can be matched against the shared secret-field list.
+  return name
+    .split(/[.\[\]]+/)
```
jameshiester-oai, line 43: **[P2] Preserve compound meaning across nested-key separators.** Splitting before the exact-name check turns `api.key`, `api[key]`, `private.key`, and `signing.key` into individually non-secret words, so all four remain unchanged with short credential values. These are supported dotted/bracketed parameter syntaxes, and the helper's comment explicitly claims nested-key handling. Match normalized adjacent path components (while retaining benign-name exclusions) instead of testing each piece independently.

jameshiester-oai (commented):
Reviewed base `ea63e1ccb4d5b9915e82d13366461fb367925709` through exact head `3b2b9981e191063c55349225b0908d0a0da5bf5b`.

Scope: fresh full review of the two-file sanitizer change.

Verdict: the query-path fix correctly stops broad substring over-redaction and closes several plain/templated URL gaps, while preserving redteam body canaries. It still leaves three distinct credential-name bypasses in URL fragments, parse-failure/protocol-relative URLs, and dotted/bracketed compound names.

CI: all current exact-head checks are green across supported Node versions, Linux, macOS, Windows, builds, smoke, integration, style, CodeQL, and Codecov.

- P1: ordinary URL fragments do not receive the new URL-specific matcher, so compound credential names such as `#my_api_key_param=hunter2` and `#session_token=hunter2` survive in structured logs/artifacts, while the templated equivalents redact.
- P2: protocol-relative and malformed URLs fall through to `hasSecretFormSegment`, which still uses the old narrow matcher; `//example.com?my_api_key_param=hunter2` remains unchanged.
- P2: dotted/bracketed compound credential paths such as `api.key`, `api[key]`, `private.key`, and `signing.key` are split into individually non-secret words and remain unredacted.

Existing feedback:
- Four unresolved current-head threads remain applicable and were not duplicated: camelCase credential names; ampersands inside Nunjucks placeholders; semicolon-separated compound credentials in templated URLs; and its missing regression test.
- The semicolon issue is broader than its original `auth` example: exact-name `auth` is caught by the narrow pass, but `my_api_key_param` and `session_token` reproduce the leak in both templated and ordinary semicolon forms.

Validation:
- `test/util/sanitizer.test.ts` and `test/providers/openai-codex-app-server.test.ts` — 299/299 passed.
- Direct exact-head probes reproduced all three new bypasses and all four existing threads.
- `sanitizeObject({ url: ... })` probes confirmed the leaking values survive the actual structured sanitization boundary.
- Three independent read-only specialists converged on the fragment inconsistency; the correctness lane additionally found and reproduced the fallback and nested-name gaps.

Smallest correction:
- Use one URL-only, delimiter/template-aware pair scanner for plain and templated queries/fragments and for parse-failure detection. It should recognize sensitive phrases across supported nested-key separators without applying the broader matcher to request/response bodies.

Limitations:
- No production log sink, OAuth provider, or credentialed hosted request was exercised; probes used synthetic short values to ensure name-based redaction, rather than value heuristics, was under test.
- Local dependencies differ only in unrelated PostCSS resolution; exact-lock evidence comes from fully green exact-head CI.
- The requested `/pr-review-toolkit:review-pr all parallel` command was unavailable (exit 127), so its six lenses were applied manually and reconciled with the specialists.

mldangelo (commented):
Flagging a regression I verified by running both regexes directly — this narrows redaction in a way that drops credentials main currently catches.

The intended fix works: `design`, `tokenizer`, and `secretary` correctly stop matching. But the word-boundary anchors also stop matching every **camelCase** credential name:

| param | main | this PR |
|---|---|---|
| `sessionToken` | redacted | **not redacted** |
| `userPassword` | redacted | **not redacted** |
| `jwtToken` | redacted | **not redacted** |
| `oauthToken` | redacted | **not redacted** |
| `appSecret` | redacted | **not redacted** |
| `sharedSecret` | redacted | **not redacted** |
| `session_token` | redacted | redacted |
| `api_key` | redacted | redacted |

Reproduce:

```js
const MAIN = /(api[_-]?key|token|password|secret|signature|sig|access[_-]?token|refresh[_-]?token|id[_-]?token|client[_-]?secret|authorization)/i;
const PR = /(?:^|[^a-z0-9])(?:api[_-]?key|apikey|access[_-]?token|...|token|auth|pwd|sig)(?:$|[^a-z0-9])/i;
MAIN.test('sessionToken')  // true
PR.test('sessionToken')    // false
```

The `isSensitiveParamName` fallback doesn't recover these: it splits on `[.\[\]]+` and defers to `isSecretField`, and `normalizeFieldName` only strips `-_\s=` — it doesn't split camelCase.

So `?sessionToken=hunter2` would land in the clear. `looksLikeSecret` won't catch it either, since it only fires on known prefixes (`sk-`, `AKIA`, `AIza`, `Bearer`) or 64+ char blobs. That reaches `src/logger.ts:373`/`:473` and `sanitizeObject` at `src/util/sanitizer.ts:659`, so it hits both logs and persisted eval results.

The four new tests only exercise the over-redaction direction — none asserts that a camelCase credential *stays* redacted, which is why CI is green.

Suggested fix: normalize camelCase before matching (insert a separator at lower→upper boundaries), or add `[a-z](?=[A-Z])` to the boundary class. Worth a test asserting the redaction direction so this can't regress again.

Separately, the fragment paths look asymmetric — line 955 scrubs plain-URL hashes with the narrow matcher while line 894 uses the broad one, so `#my_api_key_param=` leaks but `{{p}}#my_api_key_param=` redacts.

PR #10258, opened 2026-08-01
fix(providers): stop leaking Nscale provider settings into the request body

mldangelo:
## Summary

`createNscaleProvider` spreads the **entire** user config into `passthrough`, which the OpenAI-compatible provider serializes verbatim into the request body. So promptfoo-level provider settings get sent to Nscale as model parameters — and the settings that should have been consumed locally never take effect.

Given a config of `{ apiKey, headers, apiBaseUrl, organization, temperature }`, the body actually put on the wire is:

```
model, messages, max_tokens, temperature, apiKey, apiKeyEnvar, apiBaseUrl, organization, headers
```

## Consequences

| Setting | Expected | Actual |
| --- | --- | --- |
| `headers` | sent as HTTP headers | serialized into the JSON body; request went out with only `Content-Type` and `Authorization` |
| `apiBaseUrl` | routes the request | **ignored** (endpoint hardcoded), but still sent as a model parameter |
| `apiKey` | Authorization header only | also transmitted as a model parameter |

`headers` and `apiBaseUrl` are the user-visible breakages: a user pointing at a private/regional Nscale endpoint, or adding a tenant/routing header, gets silence rather than an error.

On `apiKey` — promptfoo's cache (`getStringForFetchCacheKey` → `fingerprintFetchCacheSecret`) and sanitizer layers do fingerprint/redact fields named `apiKey`, so this is **not** a plaintext-at-rest leak. But the credential has no business in the body, and strict OpenAI-compatible servers reject unknown body parameters.

## Change

Split the user config into settings promptfoo owns — auth, routing, headers, retries, cost overrides, mirroring `OpenAiSharedOptions` — and genuine model parameters, forwarding only the latter into `passthrough`. Additionally:

- an explicit `apiBaseUrl` is now honored instead of silently ignored;
- an explicit `passthrough` block merges rather than nesting under itself (previously `passthrough: {x}` became `passthrough: { passthrough: {x} }`).

Model parameters (`temperature`, `top_p`, `seed`, `custom_param`, …) are forwarded exactly as before.

## Why the existing tests missed it

`test/providers/nscale.test.ts` calls `vi.mock('../../src/providers/openai')`, so it can only assert the shape of the config object handed to the OpenAI provider. The config looked correct at that layer while the request body carried the token and the user's headers.

`test/providers/nscaleRequest.test.ts` mocks only the transport (`fetchWithCache`), exercising the real OpenAI provider and asserting what actually goes on the wire.

## Test plan

```
without fix -> 5 failed | 2 passed
  × does not send the service token in the request body
  × applies configured headers as HTTP headers rather than body fields
  × honors a configured apiBaseUrl instead of shipping it in the body
  × does not leak any promptfoo-level provider option into the body
  × merges an explicit passthrough block without nesting it
with fix    -> 7 passed
```

- `npx vitest run test/providers/nscale.test.ts test/providers/nscaleRequest.test.ts test/providers/registry.test.ts test/providers/togetherai.test.ts test/providers/cerebras.test.ts` — 143 passed

## Note for follow-up

`togetherai.ts` uses the same `passthrough: { ...config }` pattern and looks susceptible; `cerebras.ts` partially guards it (strips only `basePath`). Not touched here to keep this reviewable.

### DIFF
--- a/src/providers/nscale.ts
+++ b/src/providers/nscale.ts
@@ -15,6 +15,33 @@ import type { ApiProvider, ProviderOptions } from '../types/index';
  *
  * Documentation: https://docs.nscale.com/
  */
+/**
+ * Config keys promptfoo consumes itself rather than forwarding to the model.
+ *
+ * `passthrough` is serialized verbatim into the request body, so anything spread
+ * into it is sent to Nscale as a model parameter. Spreading the whole user config
+ * put `apiKey` — the raw service token — into the JSON body, and diverted
+ * `headers` there too so custom headers never became HTTP headers.
+ *
+ * Mirrors `OpenAiSharedOptions` in `./openai/types`.
+ */
+const NSCALE_PROVIDER_LEVEL_OPTIONS = new Set([
+  'apiKey',
+  'apiKeyEnvar',
+  'apiKeyRequired',
+  'apiHost',
+  'apiBaseUrl',
+  'organization',
+  'headers',
+  'maxRetries',
+  'cost',
+  'inputCost',
+  'outputCost',
+  'audioCost',
+  'audioInputCost',
+  'audioOutputCost',
+]);
+
 export function createNscaleProvider(
   providerPath: string,
   options: {
@@ -26,6 +53,20 @@ export function createNscaleProvider(
   const splits = providerPath.split(':');
 
   const config = options.config?.config || {};
+
+  // Split the user's config into settings promptfoo handles (auth, routing,
+  // headers, cost overrides) and genuine model parameters, so only the latter
+  // reach the request body.
+  const { passthrough: explicitPassthrough, ...configOptions } = config;
+  const providerLevelOptions: Record<string, any> = {};
+  const modelParameters: Record<string, any> = {};
+  for (const [key, value] of Object.entries(configOptions)) {
+    if (NSCALE_PROVIDER_LEVEL_OPTIONS.has(key)) {
+      providerLevelOptions[key] = value;
+    } else {
+      modelParameters[key] = value;
+    }
+  }
 
   // Prefer service tokens over API keys (API keys deprecated Oct 30, 2025)
   const getApiKey = () => {
@@ -41,10 +82,14 @@ export function createNscaleProvider(
   const nscaleConfig = {
     ...options,
     config: {
-      apiBaseUrl: 'https://inference.api.nscale.com/v1',
+      ...providerLevelOptions,
+      // Honor an explicit apiBaseUrl (private/regional Nscale endpoints) instead
+      // of silently ignoring it while still shipping it in the request body.
+      apiBaseUrl: providerLevelOptions.apiBaseUrl || 'https://inference.api.nscale.com/v1',
       apiKey: getApiKey(),
       passthrough: {
-        ...config,
+        ...modelParameters,
+        ...explicitPassthrough,
       },
     },
   };
--- a/test/providers/nscaleRequest.test.ts
+++ b/test/providers/nscaleRequest.test.ts
@@ -1,0 +1,143 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// `nscale.test.ts` mocks `src/providers/openai` wholesale, so it can only assert
+// the shape of the config object handed to the OpenAI provider — never what is
+// actually put on the wire. These tests mock only the transport, exercising the
+// real OpenAI provider, because the defects they guard against were invisible at
+// the config layer: the config looked correct while the request body carried the
+// service token and the user's headers.
+vi.mock('../../src/cache', async (importOriginal) => ({
+  ...(await importOriginal<any>()),
+  fetchWithCache: vi.fn(),
+}));
+
+import { fetchWithCache } from '../../src/cache';
+import { createNscaleProvider } from '../../src/providers/nscale';
+
+function mockResponse() {
+  vi.mocked(fetchWithCache).mockResolvedValue({
+    data: {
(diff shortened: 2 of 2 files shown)


jameshiester-oai (commented):
Reviewed base `82ca3c24ec445cf1734face46042c187b659b954` through head `c3299af14ae0b43985b663d5a3adbe1687bab0be` (fresh full review).

Verdict: the primary `apiKey`/headers/`apiBaseUrl` fix works, but the denylist still classifies security-sensitive local OpenAI settings as model parameters, and explicit credential-source precedence remains broken. Exact-head CI is 44/44 successful.

Verified findings

- P1: local-only OpenAI configuration, including MCP authentication material, is still serialized into Nscale request bodies. See inline comment.
- P1: an explicit `apiKeyEnvar` loses to ambient Nscale credentials, which can send the wrong token to a configured private endpoint. See inline comment.

Existing feedback

- The unresolved mock-cleanup thread correctly cites the repository's `afterEach` hygiene rule; I did not duplicate it.
- The unresolved claim that object-spreading `null`/`undefined` throws is not reproducible on the supported Node 24 runtime (`{...null}` and `{...undefined}` both produce `{}`).

- Prefer an explicit typed split between locally consumed configuration and API body parameters; maintaining a denylist that mirrors only `OpenAiSharedOptions` will miss future local options.
- Parameterize the wire-level coverage across chat, completion, and embedding routes; the current transport suite exercises only default chat.

Validation

- `source ~/.nvm/nvm.sh && nvm use && npm ci --ignore-scripts` — completed with the exact lockfile after transient shared-worktree extraction collisions.
- `npx vitest run test/providers/nscale.test.ts test/providers/nscaleRequest.test.ts test/providers/registry.test.ts test/providers/togetherai.test.ts test/providers/cerebras.test.ts --sequence.seed=10258` — 143/143 passed.
- Temporary exact-head transport regression harness — 2/2 passed, reproducing MCP/callback/local-path body leakage and ambient-token precedence over `apiKeyEnvar`.
- `npm run tsc` — passed.
- `npx biome check src/providers/nscale.ts test/providers/nscaleRequest.test.ts` — passed.
- `git diff --check 82ca3c24ec445cf1734face46042c187b659b954...HEAD` — passed.
- Isolated local CLI eval using `--no-cache`, a loopback Nscale-compatible endpoint, disabled sharing, and exported JSON — 1/1 passed, score 1, zero errors, 2 tokens; captured request reproduced the dummy MCP token, callback reference, `basePath`, `showThinking`, and `omitDefaults` in the body.

Limitations: no live Nscale credential or paid request was used; completion and embedding routes were inspected but not exercised end to end. The requested PR Review Toolkit command was unavailable (exit 127), so its correctness, tests, docs/comments, error-handling, type-design, and simplification lenses were applied manually and reconciled with three read-only specialist reviews.

src/providers/nscale.ts
```diff
@@ -27,6 +54,20 @@ export function createNscaleProvider(
...
+  // headers, cost overrides) and genuine model parameters, so only the latter
+  // reach the request body.
+  const { passthrough: explicitPassthrough, ...configOptions } = config;
+  const providerLevelOptions: Record<string, any> = {};
+  const modelParameters: Record<string, any> = {};
+  for (const [key, value] of Object.entries(configOptions)) {
+    if (NSCALE_PROVIDER_LEVEL_OPTIONS.has(key)) {
+      providerLevelOptions[key] = value;
+    } else {
+      modelParameters[key] = value;
```
jameshiester-oai, line 67: [P1] Keep locally consumed settings out of `passthrough`. This fallback treats every key outside `OpenAiSharedOptions` as a model parameter, but `OpenAiCompletionOptions` also has local-only `mcp`, `functionToolCallbacks`, `showThinking`, and `omitDefaults`, and config loading injects `basePath`. The OpenAI chat provider then spreads all of them into the outbound JSON. An exact-head loopback CLI eval captured an MCP bearer token, callback file reference, and local `/tmp` base path in the Nscale body; because `mcp` is nested under `passthrough`, local MCP initialization does not receive it either. MCP auth supports bearer/basic/API-key/OAuth secrets, so extend the local-option classification (preferably with a typed API-parameter allowlist) and add a wire test asserting nested credentials never leave the process.

src/providers/nscale.ts
```diff
@@ -41,10 +82,14 @@ export function createNscaleProvider(
   const nscaleConfig = {
     ...options,
     config: {
-      apiBaseUrl: 'https://inference.api.nscale.com/v1',
+      ...providerLevelOptions,
+      // Honor an explicit apiBaseUrl (private/regional Nscale endpoints) instead
+      // of silently ignoring it while still shipping it in the request body.
+      apiBaseUrl: providerLevelOptions.apiBaseUrl || 'https://inference.api.nscale.com/v1',
       apiKey: getApiKey(),
```
jameshiester-oai, line 89: [P1] Resolve an explicit `apiKeyEnvar` before ambient Nscale credentials. `getApiKey()` assigns `NSCALE_SERVICE_TOKEN`/`NSCALE_API_KEY` to the concrete `apiKey` field, and `OpenAiGenericProvider.getApiKey()` checks that field before `apiKeyEnvar`. In an exact-head transport probe with `apiKeyEnvar: CUSTOM_TOKEN`, both env values present, and a private `apiBaseUrl`, the Authorization header used the ambient Nscale token instead of `CUSTOM_TOKEN`. That makes the explicit option ineffective and can disclose the wrong credential to the configured endpoint. Either resolve the named variable first or avoid populating `apiKey` when `apiKeyEnvar` is set, and cover the precedence on the wire.

PR #10259, opened 2026-08-01
fix(providers): correct Nscale model identifiers

mldangelo:
## Summary

Nscale model IDs are the upstream Hugging Face repository IDs, but most of the ones we document use an invented `vendor/lowercase-name` scheme. Those IDs don't exist, so following our docs produces a failed lookup rather than a completed eval.

## Evidence

Nscale's own API example uses `"model": "moonshotai/Kimi-K2.5"`, and [LiteLLM's Nscale registry](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json) uses the same convention. Checking each identifier against `huggingface.co/api/models` confirms our org prefixes aren't real orgs:

```
401  meta/llama-3.3-70b-instruct          200  meta-llama/Llama-3.3-70B-Instruct
401  deepseek/deepseek-r1-distill-...     200  deepseek-ai/DeepSeek-R1-Distill-...
401  BlackForestLabs/FLUX.1-schnell       200  black-forest-labs/FLUX.1-schnell
401  ByteDance/SDXL-Lightning-4step       200  ByteDance/SDXL-Lightning
401  Qwen/Qwen3-235B-A22B-Instruct        200  Qwen/Qwen3-235B-A22B
```

The two IDs that were already canonical — `openai/gpt-oss-120b` and `stabilityai/stable-diffusion-xl-base-1.0` — are presumably why this went unnoticed.

**Every model ID published in this PR is HTTP-verified against the Hugging Face API.**

## Also fixed in the same area

- **Image cost silently wrong.** `calculateImageCost` keyed on `BlackForestLabs/FLUX.1-schnell` and `ByteDance/SDXL-Lightning-4step`, so every lookup missed and fell through to the `0.002` default rate.
- **Both example configs** referenced `nscale:meta/llama-3.3-70b-instruct` and the non-existent SDXL Lightning step variants.
- **The docs' only complete example** used `- id: nscale-gpt-oss`. It has no `nscale:` prefix, so it throws `Could not identify provider: nscale-gpt-oss` — verified by resolving it through `loadApiProvider`.
- **The embedding ID was written two ways on the same page**: `qwen/qwen3-embedding-8b` in Model Types vs `Qwen/Qwen3-Embedding-8B` in the table.
- **`nscale:image:` was missing** from the Model Types section despite being supported.
- **`stream: true` was documented as supported.** There is no streaming path for chat providers, so it is forwarded to the API as a body parameter and the SSE response cannot be parsed. Replaced with a note; the `temperature: 0` / `max_tokens: 1024` defaults are now documented too, since both differ from the usual API defaults.

Adds Kimi K2.5, Qwen 3 4B Instruct 2507, Qwen 3 235B A22B, and Llama 3.2 11B Vision Instruct, and points readers at `GET /v1/models` as the authoritative per-account catalog.

## Scope note

I could not reach `GET /v1/models` — it requires a service token and there was none available, so I could not confirm which of these models Nscale currently serves for a given account. This PR corrects the **identifier format**, which is independently verifiable, and deliberately keeps the existing model set rather than removing entries I cannot confirm are retired. A follow-up with a real token should reconcile the list and add per-model pricing (text models currently report `cost: undefined` — `billing.ts` bails before the rate lookup for non-OpenAI model names).

## Test plan

- `npx vitest run test/providers/nscale.test.ts` — 17 passed
- `npm run tsc` clean; Biome clean; Prettier clean
- Every published ID checked against `huggingface.co/api/models` (all 200)

### DIFF
--- a/examples/provider-nscale/image-promptfooconfig.yaml
+++ b/examples/provider-nscale/image-promptfooconfig.yaml
@@ -1,11 +1,11 @@
 description: 'Nscale Image Generation Example - comparing different image models'
 
 providers:
-  - id: nscale:image:BlackForestLabs/FLUX.1-schnell
+  - id: nscale:image:black-forest-labs/FLUX.1-schnell
     config:
       size: '1024x1024'
       n: 1
-  - id: nscale:image:ByteDance/SDXL-Lightning-4step
+  - id: nscale:image:ByteDance/SDXL-Lightning
     config:
       size: '1024x1024'
       n: 1
--- a/src/providers/nscale/image.ts
+++ b/src/providers/nscale/image.ts
@@ -43,7 +43,7 @@ export class NscaleImageProvider extends OpenAiImageProvider {
   /**
    * Create a new Nscale image provider instance.
    *
-   * @param modelName - The Nscale image model name (e.g., 'ByteDance/SDXL-Lightning-4step')
+   * @param modelName - The Nscale image model name (e.g., 'black-forest-labs/FLUX.1-schnell')
    * @param options - Provider configuration options
    */
   constructor(
@@ -130,11 +130,14 @@ export class NscaleImageProvider extends OpenAiImageProvider {
    */
   private calculateImageCost(modelName: string, n: number = 1): number {
     // Nscale pricing varies by model - these are approximate based on their pricing page
+    // Keyed by the Hugging Face repository ID Nscale uses as its model ID. These
+    // were previously keyed on identifiers that do not exist upstream
+    // (`BlackForestLabs/...`, `ByteDance/SDXL-Lightning-4step`), so every lookup
+    // missed and silently fell through to the default rate below.
     const costPerImage: Record<string, number> = {
-      'BlackForestLabs/FLUX.1-schnell': 0.0013, // $0.0013 per 1M pixels for 1024x1024
+      'black-forest-labs/FLUX.1-schnell': 0.0013, // $0.0013 per 1M pixels for 1024x1024
       'stabilityai/stable-diffusion-xl-base-1.0': 0.003, // $0.003 per 1M pixels
-      'ByteDance/SDXL-Lightning-4step': 0.0008, // $0.0008 per 1M pixels
-      'ByteDance/SDXL-Lightning-8step': 0.0016, // $0.0016 per 1M pixels
+      'ByteDance/SDXL-Lightning': 0.0008, // $0.0008 per 1M pixels
     };
 
     const baseCost = costPerImage[modelName] || 0.002; // Default cost
--- a/examples/provider-nscale/README.md
+++ b/examples/provider-nscale/README.md
@@ -60,27 +60,27 @@ Nscale supports many popular models with competitive pricing:
 
 - `openai/gpt-oss-120b` - OpenAI's 120B open-weight model
 - `openai/gpt-oss-20b` - OpenAI's 20B model
-- `qwen/qwen-3-235b-a22b-instruct` - Qwen 3 235B model
-- `qwen/qwen-3-235b-a22b-instruct-2507` - Qwen 3 235B Instruct 2507
-- `qwen/qwen-3-4b-thinking-2507` - Qwen 3 4B Thinking 2507
-- `qwen/qwen-3-8b` - Qwen 3 8B model
-- `qwen/qwen-3-14b` - Qwen 3 14B model
-- `qwen/qwen-3-32b` - Qwen 3 32B model
-- `qwen/qwen-2.5-coder-3b-instruct` - Qwen 2.5 Coder 3B Instruct
-- `qwen/qwen-2.5-coder-7b-instruct` - Qwen 2.5 Coder 7B Instruct
-- `qwen/qwen-2.5-coder-32b-instruct` - Qwen 2.5 Coder 32B Instruct
-- `qwen/qwq-32b` - Qwen QwQ 32B model
-- `meta/llama-3.3-70b-instruct` - Meta's Llama 3.3 70B model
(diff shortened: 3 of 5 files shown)


examples/provider-nscale/image-promptfooconfig.yaml
```diff
@@ -1,11 +1,11 @@
 description: 'Nscale Image Generation Example - comparing different image models'
 
 providers:
-  - id: nscale:image:BlackForestLabs/FLUX.1-schnell
+  - id: nscale:image:black-forest-labs/FLUX.1-schnell
     config:
       size: '1024x1024'
       n: 1
-  - id: nscale:image:ByteDance/SDXL-Lightning-4step
+  - id: nscale:image:ByteDance/SDXL-Lightning
```
jameshiester-oai, line 8: [P2] Verify this ID against Nscale before making it part of the runnable example. A 200 from Hugging Face only proves that the upstream repository exists; Nscale identifies its website or authenticated `/v1/models` response as the available-model authority. The current public serverless catalog and image API docs list FLUX but not SDXL Lightning, and this PR explicitly says `/v1/models` was not queried. Users initializing this example can therefore hit Nscale's documented 404 `model not found or unavailable` instead of completing the advertised comparison. Confirm the exact ID with a service token, or remove/clearly qualify it until availability is established.

src/providers/nscale/image.ts
```diff
@@ -130,11 +130,14 @@ export class NscaleImageProvider extends OpenAiImageProvider {
...
+    // were previously keyed on identifiers that do not exist upstream
+    // (`BlackForestLabs/...`, `ByteDance/SDXL-Lightning-4step`), so every lookup
+    // missed and silently fell through to the default rate below.
     const costPerImage: Record<string, number> = {
-      'BlackForestLabs/FLUX.1-schnell': 0.0013, // $0.0013 per 1M pixels for 1024x1024
+      'black-forest-labs/FLUX.1-schnell': 0.0013, // $0.0013 per 1M pixels for 1024x1024
       'stabilityai/stable-diffusion-xl-base-1.0': 0.003, // $0.003 per 1M pixels
-      'ByteDance/SDXL-Lightning-4step': 0.0008, // $0.0008 per 1M pixels
-      'ByteDance/SDXL-Lightning-8step': 0.0016, // $0.0016 per 1M pixels
+      'ByteDance/SDXL-Lightning': 0.0008, // $0.0008 per 1M pixels
```
jameshiester-oai, line 140: [P2] Do not activate this rate without current Nscale pricing evidence. Changing the dead `...-4step` key to the canonical repository ID means successful calls now report `$0.0008`, but Nscale's current public catalog says image pricing depends on size and steps and publishes only FLUX at `$0.0013 (@ 4 steps)` per megapixel; it lists neither this model nor this rate. The docs' unchanged “starting at $0.0008” claim has the same unsupported source. Query authenticated `/v1/models` and calculate from its pricing metadata, or return no estimate for models whose rate is not authoritative.

jameshiester-oai (commented):
Reviewed base `82ca3c24ec445cf1734face46042c187b659b954` through head `de9de53ae9c7c1594936cab25b15a3c9aa920ade` (fresh full review).

Verdict: the canonical Hugging Face IDs parse correctly and the FLUX ID matches Nscale's current API example, but the PR promotes an unverified SDXL-Lightning entry into the runnable example and activates an unsupported price estimate for it. Exact-head CI is 44/44 successful.

Verified findings

- P2: the bundled image example now advertises `ByteDance/SDXL-Lightning` without evidence that Nscale currently serves that ID. See inline comment.
- P2: correcting that key activates a `$0.0008` cost that is not present in Nscale's current public catalog/pricing. See inline comment.

Existing feedback

- The unresolved per-megapixel thread is valid: `calculateImageCost` ignores `size` and steps, so a 512x512 request reports the same cost as 1024x1024 even though Nscale bills by resolution and steps. I did not duplicate it.
- The unresolved image-cost regression-test gap is also valid, though its P1 label overstates the impact. The existing 17-test Nscale suite never exercises the image call/cost path.

- Source the available model catalog and pricing from an authenticated `/v1/models` snapshot or omit account-dependent entries instead of inferring availability and rates from Hugging Face repository existence.
- Add transport/cost tests covering canonical IDs, resolution, steps, `n`, cache hits, and unknown models.

Validation

- `npx vitest run test/providers/nscale.test.ts --sequence.seed=10259` — 17/17 passed.
- `npm run local -- validate -c examples/provider-nscale/promptfooconfig.yaml` — valid; local log-cleanup warnings were environmental.
- `npm run local -- validate -c examples/provider-nscale/image-promptfooconfig.yaml` — valid; local log-cleanup warnings were environmental.
- Real-source provider construction probe — all five representative chat/embedding/image IDs parsed and preserved their canonical model names.
- Direct image-cost probe — reproduced flat `$0.0013` FLUX cost regardless of resolution and `$0.0008` for the newly canonical SDXL-Lightning key.
- `npm run tsc` and `npx biome check src/providers/nscale/image.ts` — passed.
- Prettier check for all changed docs/examples and `git diff --check` — passed.
- `SKIP_OG_GENERATION=true npm --prefix site run build` — production build passed.
- Live primary-source review: Nscale's current image API example uses `black-forest-labs/FLUX.1-schnell`; its current public serverless catalog lists FLUX at `$0.0013 (@ 4 steps)` per megapixel, but no ByteDance/SDXL-Lightning model or `$0.0008` rate.

Limitations: no Nscale service token was available, so account-specific `/v1/models` and a paid image request were not exercised. The requested PR Review Toolkit command was unavailable (exit 127), so its six lenses were applied manually and reconciled with three read-only specialist reviews.

PR #5915, opened 2025-10-14
fix: errors in websocket responses should be counted as errors.

jameshiester:
* Update examples to make websocket streaming example a regular evaluation instead of redteam
* Update default timeout of websocket to match HTTP approach (using REQUEST_TIMEOUT_MS environment variable or 300000ms).
* Fix websocket provider so that it rejects instead of swallowing error.

PR #10266, opened 2026-08-02
fix(providers): stop WebSocket evals from hanging when a stream stalls

pacocartones:
## Summary

For providers configured with `streamResponse`, a WebSocket server that sends a non-terminal chunk and then stops sending leaves `callApi`'s promise pending forever: `ws.onmessage` clears the request timeout on the very first message (`src/providers/websocket.ts:322`), before knowing whether the stream is finished, and the provider registers no close handler.

## Root cause

The only guard against a stalled stream was the request deadline armed in `callApi` — but it was disarmed unconditionally by the first chunk:

```ts
ws.onmessage = (event) => {
  clearTimeout(timeout);        // ran on every message, including a partial one
  if (streamResponse) { ... }
}
```

After that, nothing re-arms it: the streaming path only resolves via `streamResponse` returning `isComplete` (`:332`), and there is no `close` handler anywhere in the file. A partial chunk followed by silence therefore hangs the eval with no ceiling (the separate `PROMPTFOO_EVAL_TIMEOUT_MS` watchdog in `evaluator.ts` is disabled by default).

## Fix

Keep the deadline armed until the stream actually settles the promise:

- streaming path: clear it only on `isComplete`, or when `streamResponse` throws;
- non-streaming path: keep clearing it up front, where the first message always settles;
- `onerror` behavior unchanged.

Also documents in `site/docs/providers/websocket.md` that `timeoutMs` covers the whole streamed request.

## Test coverage

New regression test `should time out when the stream delivers a partial chunk and then stalls` in `test/providers/websocket.test.ts` (`describe('timeouts')`), using fake timers per repo convention: it delivers one partial chunk, then advances past `timeoutMs` and asserts the promise rejects with `WebSocket request timed out` and the socket is closed.

The sibling timeout test added in #5915 (hardened to fake timers in #10211) never delivers a message, so it never enters `ws.onmessage` — this extends coverage to the sibling case.

## Verification

- The new test fails without the fix (`AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times`) and passes with it; the revert/re-apply cycle was re-confirmed. Full `test/providers/websocket.test.ts`: 60/60, including three `--sequence.shuffle` seeds.
- `tsc --noEmit` clean; biome + prettier clean on the touched files (one `noExcessiveCognitiveComplexity` warning on `callApi`, pre-existing on `main`).

AI-assisted: the fix and test were drafted with AI assistance and verified locally as above. I own the change and can walk through every line.


mldangelo-oai:
Thanks, @pacocartones — great catch, and your focused regression test made the stall straightforward to verify. I prepared a small maintainer follow-up that keeps healthy streams alive, still times out genuinely idle streams, and restores automatic retries. It passes 230 focused tests, typechecking, and a real end-to-end WebSocket eval. I can’t push directly to your fork from this environment, so the ready-to-cherry-pick follow-up is commit `83e746a6a` on `promptfoo:mdangelo/codex/pr-10266-websocket-stream-stall`. If you cherry-pick it, I can get your original PR across the finish line. Thank you for the thoughtful contribution!

mldangelo-oai (approved):
Thanks, Paco — excellent catch, and the focused regression test makes the fix easy to verify. This is a valuable contribution; we’ll handle the additional active-stream and retry refinements in a separate follow-up. Really appreciate it!

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

PR #10268, opened 2026-08-03
fix(util): ignore braces inside strings when extracting JSON objects

serhiizghama:
`extractJsonObjects` counts every `{` and `}` literally, including braces that appear inside string values. A `}` inside a string balances the brace counter early, so the slice it cuts is truncated invalid JSON — the object fails to parse and gets silently dropped. `extractJsonObjects('{"a": "}"}')` returns `[]` instead of `[{ a: '}' }]`, and a model output like `The result is {"status": "ok", "note": "brace } here"} done` yields nothing.

This hits `contains-json` and every `extractFirstJsonObject` consumer (rubric matcher, synthesis, redteam parsers) whenever a JSON value legitimately contains a brace.

Fixed by tracking string state (with escape handling) while scanning, so braces inside "..." no longer affect the open/close count. Only double-quoted JSON strings are tracked, matching the JSON case; single-quoted YAML braces stay out of scope as before. Added tests for a brace and an unbalanced brace inside a value, extraction from surrounding text, and an escaped quote preceding a string brace.

### DIFF
--- a/src/util/json.ts
+++ b/src/util/json.ts
@@ -221,13 +221,27 @@ export function extractJsonObjects(str: string): object[] {
       let openBraces = 1;
       let closeBraces = 0;
       let j = i + 1;
+      let inString = false;
+      let escaped = false;
 
       // Track braces as we go to detect potential JSON objects
       while (j < Math.min(i + maxJsonLength, str.length) && openBraces > closeBraces) {
-        if (str[j] === '{') {
+        const ch = str[j];
+        // Ignore braces inside string literals so a `}` in a value doesn't
+        // prematurely balance the object (e.g. `{"a": "}"}`).
+        if (inString) {
+          if (escaped) {
+            escaped = false;
+          } else if (ch === '\\') {
+            escaped = true;
+          } else if (ch === '"') {
+            inString = false;
+          }
+        } else if (ch === '"') {
+          inString = true;
+        } else if (ch === '{') {
           openBraces++;
-        }
-        if (str[j] === '}') {
+        } else if (ch === '}') {
           closeBraces++;
         }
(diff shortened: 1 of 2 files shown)


serhiizghama:
Good catch. Added an `atValueStart` flag — a `"` only opens a string right after `{`, `:`, `,`, or another string closing, so a bare quote mid-plain-scalar (your `mentions "admin` example) no longer flips the scanner into string mode. Added a regression test with your exact repro, verified it fails without the fix.

Note: this pushes `extractJsonObjects`'s cognitive complexity from 65 to 75 (biome warning, non-blocking, exit 0 — was already at 42 before this PR). Didn't refactor it here since that's a bigger change than this fix warrants, but flagging it in case it's worth a follow-up.

serhiizghama:
Good catch, and a good repro. `[` wasn't in the value-boundary set, so a quoted string as the first array element never entered string mode — same underlying issue as the earlier fix, just a different boundary. Added `[`/`]` alongside `{`/`}` for the boundary check and a regression test for `{"messages":["}"]}`. 84/84 in json.test.ts.

serhiizghama:
Both real, thanks — I'd only been tracking double-quoted strings, so a `"` inside a comment or a single-quoted scalar flipped the scanner into string mode and it never saw the real closing brace. The scanner now also skips `//` and `#` line comments (with the same URL-scheme exception `convertSlashCommentsToHash` uses) and single-quoted scalars, including the `''` escape. Added tests for your two inputs plus a `}` inside a comment, `''`, a URL with `//` in a plain scalar, and an apostrophe mid-scalar; all of them parse on `main` and the first two failed on the previous head of this branch.

PR #10287, opened 2026-08-04
fix(providers): fail fast when a WebSocket closes mid-stream

he-yufeng:
## Summary

Follow-up to #10266. That fix kept the request deadline armed so a stalled stream times out instead of hanging forever, but one half of the original report is still open: a peer that closes the connection cleanly after a partial chunk fires no `error` event and never completes the stream, so the promise still sits out the full `timeoutMs` (5 minutes by default) before rejecting with `WebSocket request timed out`, which misdescribes what actually happened. The provider registers no close handler, so nothing else can settle it.

This adds an `onclose` handler that clears the deadline and rejects immediately with the close code. Every path that already settles the promise calls `close()` itself first (success, stream completion, stream error, `onerror`, the timeout itself), so on those paths this reject is a no-op and behavior is unchanged. The only new behavior is the one that used to stall: peer-initiated close before the response completed now fails fast with a message that says so.

## Verification

New regression test drives open, one partial chunk, then a close with code 1001 under fake timers and asserts the promise rejects with the close message while the 60s request deadline never gets a chance to fire (`getTimerCount()` is 0). Full websocket provider suite: 61/61 passed locally, including the stall test from #10266, which still times out exactly as before because no close arrives in that scenario. Biome reports two pre-existing findings on these files on main (onmessage complexity 41, an older it.each formatting drift); both are untouched by this diff.


he-yufeng:
Good catch, that was a real regression. Failing fast on every close code took away the retry that 1012/1013 used to get for free by sitting out the deadline (the timeout message matches shouldRetry, my close message matched nothing). Fixed in e7856f6d9: 1012 (service restart) and 1013 (try again later) now keep the request deadline armed, so the retryable timeout path fires exactly as before, while every other close code still fails fast. New test drives a 1013 close and asserts no prompt rejection, then the timeout error at the deadline. 62/62 pass.

he-yufeng:
Both findings are addressed in fc8b46f8a, and they resolve each other: instead of either a plain non-retryable Error (first finding) or keeping the deadline armed so the retry takes the long way around (second finding), a 1012/1013 close now rejects immediately with the code stamped on as ECONNRESET. `shouldRetry` matches that through `isTransientConnectionError` (verified against the predicate: coded error retries, plain close error does not), so the scheduler retries on its normal backoff without waiting out the request deadline. Other close codes still fail fast. The test now asserts the rejection carries `code === 'ECONNRESET'` and names the close code.

he-yufeng:
Round three in 95669204f: 1006 (and 1005) now reject with the same ECONNRESET shape as 1012/1013, so a dropped connection retries instead of dying permanently. The transient set is explicit (1005, 1006, 1012, 1013), and codes that reject the payload itself (1002/1003, the policy violations) stay permanent, since a fresh socket would lose the same message again. The new test pins both sides: 1006 resets, 1003 does not.

he-yufeng:
The Redteam (Production API) failure looks like a flake on the service side, not this change: the eval calls the live production API, which returned 500 ("Error generating AI response", null data) mid-run, and the step only fails because the expected clean output never arrived. Nothing here touches the HTTP provider path; this PR only changes WebSocket close handling. Could someone re-trigger that job? Everything else is green.

he-yufeng:
The Redteam (Production API) failure is a 500 from the live smoke target (`Error generating AI response`, Cloudflare edge, 2026-08-07 run), not anything this diff touches. The change here only affects the WebSocket provider path, and the unit suite is green. A rerun of that workflow should clear it whenever a maintainer gets a chance.

PR #10303, opened 2026-08-04
fix(deps): pinning keyv to keep the Shai-Hulud compromised versions unreachable

ynori7:
This is a follow-up to https://github.com/promptfoo/promptfoo/pull/10301 which missed pinning keyv


mldangelo-oai:
Thanks so much for taking the time to look into this and for contributing, @ynori7!

We rechecked the Shai-Hulud exposure in #10301, which is already merged and included in the 0.122.0 release:

- The existing `keyv: ^5.6.0` already excludes the compromised `keyv@6.0.0`; npm caret ranges cannot cross the next major version. Changing it to `>=5.6.0 <6.0.0` does not add protection.
- The compromised versions that were reachable are already blocked: `cache-manager` is constrained to `>=7.2.8 <7.2.10`, and `@cacheable/utils` is constrained to `>=2.5.0 <2.5.1`.
- `test/package-manifests.test.ts` checks every root and nested lockfile installation for compromised versions and verifies that the declared `keyv`, `cache-manager`, and `@cacheable/utils` ranges cannot resolve them.
- The regenerated lockfile here also removes 71 native-package `libc` constraints, which can make incompatible glibc/musl binaries eligible for installation.

Because the `keyv` case is already protected and covered by tests, merging this would not improve security and would introduce unrelated lockfile risk, so I am closing it as already addressed. We really appreciate you spotting the incident and taking the initiative to help!

PR #10312, opened 2026-08-06
docs: clarify API key persistence in getting started

jstnorv:
Summary
This PR clarifies how OPENAI_API_KEY environment variable persistence works in the Getting Started guide.

What changed
Added guidance that export OPENAI_API_KEY=... only applies to the current terminal session.
Added instructions for making the variable available in future sessions on a local, trusted machine (shell startup file).
Added security guidance to avoid committing secrets and to prefer a secret manager or OS keychain for long-term storage.
Why
This reduces confusion for new users who expect environment variables to persist across terminal sessions, while reinforcing safe secret-handling practices.

Scope
Documentation-only change in:

[getting-started.md](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)

### DIFF
--- a/site/docs/getting-started.md
+++ b/site/docs/getting-started.md
@@ -49,7 +49,7 @@ Most providers need authentication. For OpenAI:
 export OPENAI_API_KEY=sk-abc123
 ```
 
-This export only applies to the current terminal session. If you want the key to be available in future sessions on a local, trusted machine, you can add the same export line to your shell's startup file (for example, your shell profile or rc file). Avoid committing it to source control, and consider a secret manager or your operating system's keychain for longer-term secret storage.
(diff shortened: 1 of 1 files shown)


jstnorv:
This CI failure appears unrelated to this docs change.

The shared red-team integration target at https://acme-cx-chatbot.promptfoo.dev/chat is currently returning HTTP 500 with `Error generating AI response`, including for a minimal benign request. That causes the integration target in the red-team integration config to hit its response transform path and fail with `Cannot read properties of null (reading 'choices')`.

I reproduced the 500 outside CI with a direct POST, so this looks like a shared integration dependency outage rather than a regression from this PR.

PR #10322, opened 2026-08-07
fix(scheduler): parse RFC 3339 rate-limit reset headers

pacocartones:
## What's wrong

`parseResetTime` in `src/scheduler/headerParser.ts` runs `Number.parseFloat` on the raw
header value **before** it ever reaches `parseHttpDate`. `Number.parseFloat` stops at the
first character it cannot consume, so an RFC 3339 timestamp is silently truncated to its
year:

```js
Number.parseFloat('2026-08-07T21:00:00Z'); // 2026
```

`2026` passes `Number.isFinite`, passes `num >= 0`, and lands in the `< 1_000_000_000`
bucket, which the function documents as "relative seconds". The reset is therefore placed
**2026 seconds (~34 minutes) in the future**, regardless of what the timestamp actually
said, and the `parseHttpDate` branch below is unreachable for this input.

This matters because the parser has an explicit entry for the header:

```ts
const ANTHROPIC_HEADERS = {
  ...
  reset: 'anthropic-ratelimit-requests-reset',
} as const;
```

and Anthropic documents that header as "provided in RFC 3339 format"
([Rate limits → Response headers](https://platform.claude.com/docs/en/api/rate-limits)).
So the parser is wrong for the exact format it declares support for.

The same truncation hits every reset header the function handles
(`x-ratelimit-reset-requests`, `x-ratelimit-reset-tokens`,
`anthropic-ratelimit-requests-reset`, `x-ratelimit-reset`, `ratelimit-reset`) whenever the
upstream sends a timestamp instead of a number or a duration string.

## Why it matters

`parseRateLimitHeaders` feeds two live code paths:

- `SlotQueue.updateRateLimitState` (`src/scheduler/slotQueue.ts`) stores `resetAt` and
  arms a wake-up timer at `resetAt - Date.now()`. While `remainingRequests <= 0` and
  `now < resetAt`, `isQuotaExhausted()` keeps the queue closed. The adaptive scheduler is
  on by default (`rateLimitRegistry.ts`: `this.enabled = !getEnvBool('PROMPTFOO_DISABLE_ADAPTIVE_SCHEDULER', false)`).
- `computeRateLimitWaitMs` (`src/util/fetch/index.ts`) uses the parsed `resetAt` directly
  as the fetch-level retry wait.

With a real Anthropic reset that is seconds away, the bug parks the queue ~34 minutes out.
The default queue timeout is 5 minutes (`DEFAULT_QUEUE_TIMEOUT_MS`), so in practice the
queued requests do not wait — they fail with `Request … timed out after 300000ms in queue`
while the actual rate-limit window reopened long before.

Honest scoping: the native `anthropic:` provider does not currently put response headers
into `ProviderResponse.metadata`, so it does not reach `getProviderResponseHeaders` today.
The reachable paths right now are the generic `http` provider (which does populate
`metadata.http.headers`, e.g. when pointed at `https://api.anthropic.com/v1/messages` or at
an Anthropic-compatible gateway) and any provider whose reset header is timestamp-shaped.
The parsing bug is real either way, and it is on the path that will be used the moment the
native provider starts forwarding headers.

## Reproduction

```bash
npx tsx --eval "import('./src/scheduler/headerParser').then(m=>{const now=Date.now();const h=new Date(now+60000).toISOString();const r=m.parseRateLimitHeaders({'anthropic-ratelimit-requests-reset':h});console.log('header :',h);console.log('resetAt:',new Date(r.resetAt).toISOString());console.log('delta  :',Math.round((r.resetAt-now)/1000),'seconds');})"
```

On `main` — the header says 60 seconds, the parser says 34 minutes:

```
header : 2026-08-07T19:51:02.686Z
resetAt: 2026-08-07T20:23:48.689Z
delta  : 2026 seconds
```

On this branch:

```
header : 2026-08-07T19:51:06.733Z
resetAt: 2026-08-07T19:51:06.733Z
delta  : 60 seconds
```

## The fix — and why the obvious fix is wrong

The obvious fix is to move the `parseHttpDate` branch above the numeric branch. **That
breaks existing behavior**, which is why this PR does not do it:

```js
Date.parse('2026'); // 1767225600000 -> 2026-01-01T00:00:00Z
```

A bare `"2026"` is a legitimate relative-seconds value, and `2026-01-01` sits comfortably
inside `parseHttpDate`'s ±1-year sanity window. Reordering would silently convert
"2026 seconds from now" into "a timestamp seven months in the past", i.e. an instantly
expired reset. Any bare number that reads as a plausible year or date has the same problem.

So instead of reordering, this PR guards the numeric branch so it only fires when the whole
value really is a bare number, and lets everything else fall through to the `parseHttpDate`
branch that already exists:

```ts
const BARE_NUMBER_RE = /^\d+(?:\.\d+)?$/;
```

This is the same intent as the strict check `parseRetryAfter` already applies one function
above (`String(seconds) === value.trim()`), so the two sibling parsers now agree on what
counts as a number. The regex keeps fractional values working, which matters:
`test/rateLimit.test.ts` already exercises a fractional `X-RateLimit-Reset`.

Net behavior change, measured rather than assumed. I enumerated every `*reset*` and
`*retry*` header literal that appears in `src/` and `test/` today (26 distinct reset values
× 5 reset headers, plus 27 distinct retry values × `retry-after` / `retry-after-ms` /
`parseRetryAfter`), pinned `Date.now`, and diffed the output of `parseRateLimitHeaders`
before and after the change. The only lines that differ are timestamp-shaped values:

| value | before | after |
|---|---|---|
| `2026-08-07T21:00:00Z` | `now + 2026s` | the timestamp |
| `2026-08-07T21:00:00.123Z` | `now + 2026s` | the timestamp |
| `2026-08-07T23:00:00+02:00` | `now + 2026s` | the timestamp |
| `2030-01-01T00:00:00Z` (out of window) | `now + 2026s` | `undefined` |

Everything else — durations (`30s`, `1m30s`, `2h15m30s`, `500ms`), bare seconds, Unix
seconds, Unix milliseconds, fractional seconds, `-1`, `Infinity`, 400-digit overflow values,
empty strings, HTTP-dates, and the entire `retry-after` surface — is byte-identical.
`parseRetryAfter` is not touched.

## Tests

`test/scheduler/headerParser.test.ts`, +5 cases:

- `should parse anthropic-ratelimit-requests-reset as an RFC 3339 timestamp` — fails on
  `main`.
- `should parse an RFC 3339 timestamp with a UTC offset` — fails on `main`.
- `should ignore an RFC 3339 timestamp outside the sanity window` — fails on `main`.
- `should treat a bare year-like number as relative seconds, not as a year` — passes on
  `main` and must keep passing. This is the regression guard against the naive reorder
  described above.
- `should treat a fractional number as relative seconds` — passes on `main` and must keep
  passing; pins the decimal allowance in the guard.

All five pin `Date.now` so they are deterministic and do not drift with the calendar.

## Verification

`src/scheduler/AGENTS.md` asks for `npx vitest run test/scheduler` and `npm run tsc`.

Red — same tree, with only `src/scheduler/headerParser.ts` reverted to `main`:

```
$ git checkout upstream/main -- src/scheduler/headerParser.ts
$ npx vitest run test/scheduler

 RUN  v4.1.10 ~/repos/promptfoo

 ❯ test/scheduler/headerParser.test.ts (59 tests | 3 failed) 44ms
       × should parse anthropic-ratelimit-requests-reset as an RFC 3339 timestamp 10ms
       × should parse an RFC 3339 timestamp with a UTC offset 2ms
       × should ignore an RFC 3339 timestamp outside the sanity window 1ms

 FAIL  test/scheduler/headerParser.test.ts > parseRateLimitHeaders > Anthropic format (anthropic-ratelimit-*) > should parse anthropic-ratelimit-requests-reset as an RFC 3339 timestamp
AssertionError: expected 1786134826000 to be 1786136400000 // Object.is equality

- Expected
+ Received

- 1786136400000
+ 1786134826000

 Test Files  1 failed | 9 passed (10)
      Tests  3 failed | 313 passed (316)
```

Green — with the fix:

```
$ npx vitest run test/scheduler

 RUN  v4.1.10 ~/repos/promptfoo

 Test Files  10 passed (10)
      Tests  316 passed (316)
```

Wider blast radius, all green:

```
$ npx vitest run test/scheduler test/fetch.test.ts test/rateLimit.test.ts test/util/fetch

 Test Files  14 passed (14)
      Tests  531 passed (531)
```

```
$ npm run tsc
> tsc --noEmit
$ npx @biomejs/biome check src/scheduler/headerParser.ts test/scheduler/headerParser.test.ts
Checked 2 files in 8s. No fixes applied.
```

## Related

- #8341 (`fix: parse rate-limit reset headers`, merged 2026-03-26) is the same complaint one
  file over: `src/util/fetch/index.ts` was doing `Number.parseInt(openaiReset) * 1000` on a
  reset header and got rewired to call `parseRateLimitHeaders` / `parseRetryAfter` instead.
  That PR made this parser the single source of truth for reset headers; this PR fixes the
  format it still gets wrong.

## Not in scope

- `anthropic-ratelimit-tokens-reset`, `anthropic-ratelimit-input-tokens-reset` and
  `anthropic-ratelimit-output-tokens-reset` are documented alongside the requests header but
  are not mapped in `ANTHROPIC_HEADERS`. Adding them is a separate, behavior-adding change.
- Making the native `anthropic:` provider forward response headers into
  `metadata.http.headers` so the scheduler can see them at all. Also separate.

---


pacocartones:
The `Redteam (Production API)` failure looks like an upstream outage, not this diff: the production redteam API returns HTTP 500 "Error generating AI response" during eval generation (e.g. job 93294667441), and every currently open PR (10329-10335) shows the same check failing. Build, generation and all unit/integration jobs pass here. Happy to rebase or adjust anything if it turns out to be on my side.

PR #10325, opened 2026-08-08
feat(eval): show Wilson 95% CI on the pass-rate summary line

ipezygj:
## Why

An observed pass rate is a point estimate: 17/20 and 850/1000 both print 85%, but they support very different conclusions about whether a prompt change actually helped. promptfoo currently reports pass rates as bare percentages everywhere; there is no statistical machinery in the repo (every `stderr` hit is process plumbing).

## What

The eval summary's passed line now carries a **Wilson score interval**:

```
Results:
  ✓ 17 passed (85.00%; 95% CI 64.0–94.8%)
  ✗ 3 failed (15.00%)
```

Wilson rather than the naive normal interval because it stays honest at the boundaries where users most need it: `5 passed (100%; 95% CI 56.6–100.0%)` tells the user that five tests cannot prove a prompt is fixed, where the naive interval would collapse to ±0.

- `src/util/eval/passRateStats.ts` — `wilsonInterval()`, closed form, **no new dependencies**.
- `summary.ts` — only the passed line changes; failed/error lines untouched.
- Tests: parity at 1e-9 against statsmodels' `proportion_confint(method="wilson")`, boundary behavior (0%, 100%), monotonic narrowing with sample size, invalid-input guard. Existing summary assertions updated to the new format.

Ran with `LC_ALL=en_US.UTF-8`: the only failures in `test/util/eval/summary.test.ts` are 6 pre-existing locale-dependent `toLocaleString` assertions that fail identically on the untouched base (Finnish-locale machine renders `1 000`); all new and touched assertions pass.

## Follow-up I'd like your opinion on

Since the same test set runs against every prompt, A-vs-B prompt comparisons are **paired by construction** — an exact McNemar test on the discordant counts would say whether "prompt B beats prompt A" is signal or noise, which point-estimate pass rates can't. I have the math ready (same reference-implementation parity approach); the open question is where you'd want the paired data pulled from (the results table vs. a direct query). Happy to send that as a second PR if the direction is welcome.

### DIFF
--- a/src/util/eval/passRateStats.ts
+++ b/src/util/eval/passRateStats.ts
@@ -1,0 +1,36 @@
+/**
+ * Statistics for pass rates.
+ *
+ * An observed pass rate is a point estimate: 17/20 passing says something very
+ * different from 850/1000 passing, even though both round to 85%. The Wilson
+ * score interval quantifies that difference without assuming large samples,
+ * and behaves sensibly at 0% and 100% (where the naive normal interval
+ * collapses to zero width).
+ */
+
+/**
+ * Wilson score confidence interval for a binomial proportion.
+ *
+ * @param passes - Number of successes (0 <= passes <= total)
+ * @param total - Number of trials
+ * @param z - Standard-normal quantile for the confidence level (default 1.959964 ≈ 95%)
+ * @returns Lower and upper bounds of the interval, as proportions in [0, 1]
+ */
+export function wilsonInterval(
+  passes: number,
+  total: number,
+  z: number = 1.9599639845400545,
+): { low: number; high: number } {
+  if (total <= 0 || passes < 0 || passes > total) {
+    return { low: 0, high: 1 };
+  }
+  const p = passes / total;
+  const z2 = z * z;
+  const denominator = 1 + z2 / total;
+  const center = (p + z2 / (2 * total)) / denominator;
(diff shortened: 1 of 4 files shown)


ipezygj:
The review is right, and the fix is pushed.

`doEval` sums `testPassCount` over every entry in `evalRecord.prompts` before calling the
summary, so with several prompt or provider columns the same test rows are counted once per
column. A Wilson interval reads those as independent trials: 20 cases under 50 variants become
1,000 trials, and the interval comes out far too narrow.

Rather than invent a clustered estimator for a summary line, I took the other option named in the
review and report the interval only when the assumption holds:

- `EvalSummaryParams` gains an optional `columnCount`, documented as the number of columns the
  totals were pooled from.
- `getResultsLines` shows the pass-rate interval only when `columnCount` is undefined or 1. The
  percentage is unchanged in every case, so nothing is lost for multi-column runs; the line just
  stops claiming a precision it doesn't have.
- `doEval` passes `evalRecord.prompts.length`.

Two tests cover the boundary: a three-column run asserts the line reads `48 passed (80.00%)` with
no interval, and a single-column run asserts the interval is still there.

Defaulting to "show" keeps every existing caller and test unchanged, which seemed better than
changing the shape for call sites that do have one column.

On the suite: `summary.test.ts` fails 6 tests on this branch both before and after the change (30
passed before, 32 after). They are in the provider token-usage lines that came with the merge of
main, not in anything this commit touches; I checked by running the file at the unmodified head.

PR #10330, opened 2026-08-08
fix(deps): block install scripts in Docker builds

ianw-oai:
## Summary

Dependency install scripts can run untrusted code during Docker builds. Install the locked dependency tree with scripts disabled, then rebuild only `esbuild` and `@swc/core`. Remove the docs setup step because its install script no longer runs.

## Verification

- Confirmed that dependency scripts are disabled and only the two required packages are rebuilt.
- `git diff --check`
- A full Docker build was not run because the local Docker daemon is unavailable.


Dockerfile
```diff
@@ -37,11 +37,10 @@ ENV VITE_IS_HOSTED=1 \
...
 COPY src/app/package.json ./src/app/package.json
 COPY site/package.json ./site/package.json
-# The site workspace postinstall writes a generated stats placeholder.
-RUN mkdir -p site/src
-# Leverage BuildKit cache
+# Build only the native packages needed by the application.
 RUN --mount=type=cache,target=/root/.npm \
-    npm ci --install-links --include=peer
+    npm ci --install-links --include=peer --ignore-scripts && \
+    npm rebuild esbuild @swc/core
```
mldangelo-oai: P1: Rebuild only the exact trusted root package paths. npm rebuild esbuild @swc/core filters the entire dependency inventory by installed folder name, so an untrusted nested dependency aliased to esbuild or @swc/core would also have its lifecycle script executed. That bypasses the preceding --ignore-scripts boundary. Use npm rebuild ./node_modules/esbuild ./node_modules/@swc/core, which npm resolves as exact directory specs, and add a regression covering nested aliases.

PR #10331, opened 2026-08-08
fix(deps): limit pnpm install scripts and new releases

ianw-oai:
## Summary

New package releases and unapproved install scripts increase supply chain risk. Require packages to be at least 10 days old and allow install scripts only for the eight packages already approved by this project. Store both settings in `pnpm-workspace.yaml`, where current pnpm versions read them.

## Verification

- `pnpm config get minimumReleaseAge` returns `14400`.
- `pnpm config get allowBuilds` returns the eight approved packages.
- `npx prettier --check pnpm-workspace.yaml`


pnpm-workspace.yaml
```diff
@@ -1,3 +1,15 @@
 packages:
   - 'src/app'
   - 'site'
+
+minimumReleaseAge: 14400
```
mldangelo-oai, line 10: P1: Apply the release-age policy to the package manager used by the protected paths. This setting is honored only by pnpm, but CI, the Dockerfile, and the package publishing workflow all install dependencies with npm; npm config get min-release-age remains null. A newly published dependency therefore still reaches the main build and release paths immediately. Add min-release-age=10 to .npmrc, noting that npm uses days while pnpm uses minutes, or explicitly narrow the claim to optional pnpm installs.

PR #10355, opened 2026-08-11
fix(providers): future-proof Claude sampling parameters

### LINKED ISSUE #10307: Error using Opus 5 with Bedrock provider
**Describe the bug**
When including `bedrock:us.anthropic.claude-opus-5` as a provider in promptfooconfig.yaml, all runs for that model fail with the error "Bedrock API invoke model error: ValidationException: `temperature` is deprecated for this model."

**To Reproduce**

1. Set up a simple eval run against AWS Bedrock with `id: bedrock:us.anthropic.claude-opus-5` included in providers
2. Observe that tests fail for that model with the temperature error shown above

**Expected behavior**
Recent Claude models do not support temperature, right now there is an explicit list of models not to include temperature for. Whenever a new Claude model comes out, it won't work by default. Desired behavior is for new Claude models to work without having to wait for a new Promptfoo version that explicitly includes it in the no temperature list

**Screenshots**

Image

Image

**System information:**

```
{
  "version": "0.121.19",
  "platform": {
    "os": "darwin",
    "release": "25.5.0",
    "arch": "arm64",
    "nodeVersion": "v24.16.0"
  },
  "env": {
    "telemetryDisabled": true,
    "telemetryDebug": false
  },
  "configInfo": {
    "configExists": false,
    "configContent": null
  }
}
```

**Additional context**

Here are previous incremental fixes for new Claude models related to temperature:

https://github.com/promptfoo/promptfoo/pull/8763

https://github.com/promptfoo/promptfoo/pull/9924

hugosmoreira:
## Summary

- add direct regression coverage for the reported Bedrock Claude Opus 5 request path
- recognize well-formed Claude 5+ model IDs as using deprecated sampling controls, even before a model-specific capability row is added
- retain explicit family metadata for model-specific behavior while tightening malformed model-ID boundaries

## Context and root cause

PR #10210 added explicit Claude Opus 5 support before #10307 was opened, so the reported model now has a capability row on `main`. This PR covers that exact Bedrock path with a regression test and closes the remaining structural gap: sampling-parameter suppression otherwise depends entirely on explicit `CLAUDE_MODEL_FAMILIES` rows.

That row-by-row behavior means a newly released Claude 5+ family can still inherit a configured or default `temperature` until Promptfoo adds its metadata. Anthropic documents `temperature`, `top_p`, and `top_k` as deprecated for Claude 4.7-and-later models, with non-default values returning HTTP 400. The generation fallback keeps new Claude 5+ IDs safe while explicit rows continue to provide family-specific behavior.

Reference: https://platform.claude.com/docs/en/about-claude/model-deprecations#api-parameter-deprecations

## Impact

New Claude 5+ model families and later generations omit unsupported `temperature`, `top_p`, and `top_k` parameters across Anthropic-compatible provider paths. Older Claude IDs, including Claude 4.6, retain their existing sampling behavior. Lookalikes such as `claude-sonnet-5x` no longer match the Sonnet 5 capability row.

Fixes #10307

## Verification

- affected provider tests: 981 passed across Anthropic Messages/util, Azure chat, Bedrock Converse/index, and Google Vertex
- TypeScript: `tsc --noEmit`
- changed-file lint/format: `biome check`
- production package, app, and postbuild stages passed
- broader suite: 22,112 tests passed; the remaining unrelated cases require Windows symlink privileges unavailable in this environment

### DIFF
--- a/src/providers/anthropic/messages.ts
+++ b/src/providers/anthropic/messages.ts
@@ -249,6 +249,18 @@ function withMergedAnthropicUsage(
   return usage ? { ...response, usage } : response;
 }
 
+function isOfficialAnthropicApiBaseUrl(apiBaseUrl: string | undefined): boolean {
+  if (!apiBaseUrl) {
+    return true;
+  }
+
+  try {
+    return new URL(apiBaseUrl).hostname === 'api.anthropic.com';
+  } catch {
+    return false;
+  }
+}
+
 function getAnthropicCostFromMessage(
   modelName: string,
   config: AnthropicMessageOptions,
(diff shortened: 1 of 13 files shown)


mldangelo:
Superseded by #10359, which is now merged.

Thanks for finding and fixing this, @hugosmoreira — the Claude 5+ sampling-parameter problem you identified was real, and your fix is what landed. #10359 built directly on your change and kept your approach; it added gating so the forward-looking family matcher only applies where the model ID is authoritative (the official Anthropic API, Cloudflare's dedicated Anthropic gateway, Vertex's aiplatform hosts, and Bedrock system inference profiles), so custom gateway aliases and Bedrock application-inference-profile names like `claude-prod-5` keep their configured sampling parameters.

Your contribution is credited in #10359's description. Closing this one as a duplicate of merged work — happy to reopen if you think something was missed.

PR #10374, opened 2026-08-13
fix(cli): preserve unicode log archive names

hanshenmesen:
## Summary

`promptfoo export logs` now preserves non-ASCII log file names in the generated tar archive.

## Problem and trigger

When a collected log file name contains non-ASCII characters, such as `promptfoo-日志.log`, the archive header previously stored only `promptfoo-日志`. Extracting the archive therefore lost the `.log` suffix.

Expected: the exported archive contains `promptfoo-日志.log`.

Actual: the exported archive contains `promptfoo-日志`.

## Root cause

The tar writer converts the file name to a UTF-8 `Buffer`, but limited the copy using `fileName.length`, which counts UTF-16 code units rather than UTF-8 bytes. The example name is 16 code units but 20 UTF-8 bytes, so the copy stopped four bytes early.

## Fix

Reuse the encoded file-name buffer and bound the copy by its byte length. ASCII behavior and the existing 100-byte tar header limit are unchanged.

## Regression test

Added a focused unit test that captures the tar header generated for `promptfoo-日志.log` and verifies that its filename field contains the complete UTF-8 name. The test fails on `main` with `promptfoo-日志` and passes with this change.

## Validation

- `npx vitest run test/commands/export.test.ts --sequence.seed=314159` — 12 passed
- `npm run tsc -- --pretty false` — passed
- `npx @biomejs/biome check src/commands/export.ts test/commands/export.test.ts` — passed
- `npm run build` — passed
- `npm run format:check` — passed (177 existing complexity warnings outside the changed files)
- Real CLI export with `PROMPTFOO_LOG_DIR` pointing to a directory containing `promptfoo-日志.log`; `tar -tzf` listed the complete filename

## Compatibility

No API or format changes. ASCII names are encoded identically, and checksum, file content, and padding calculations are unaffected.

## AI assistance

AI assistance was used during investigation and implementation. The failure was reproduced before the fix, the regression test was verified to fail before and pass after, the real CLI workflow was exercised, and the final diff received an independent agent review.

### DIFF
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -57,7 +57,8 @@ async function createLogArchive(logFiles: string[], outputPath: string): Promise
           const header = Buffer.alloc(512);
 
           // File name (100 bytes)
-          Buffer.from(fileName).copy(header, 0, 0, Math.min(fileName.length, 100));
+          const encodedFileName = Buffer.from(fileName);
+          encodedFileName.copy(header, 0, 0, Math.min(encodedFileName.length, 100));
 
           // File mode (8 bytes) - default 644
           Buffer.from('0000644 ').copy(header, 100);
--- a/src/commands/export.ts
+++ b/src/commands/export.ts
@@ -10,6 +10,22 @@ import { getLogDirectory, getLogFiles } from '../util/logs';
 import { createOutputData, writeOutput } from '../util/output';
 import type { Command } from 'commander';
 
+function encodeTarFileName(fileName: string): Buffer {
+  let encodedLength = 0;
+  let validEnd = 0;
+
+  for (const character of fileName) {
+    const characterLength = Buffer.byteLength(character);
+    if (encodedLength + characterLength > 100) {
+      break;
+    }
+    encodedLength += characterLength;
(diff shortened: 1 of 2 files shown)


bishnubista (approved):
Reviewed head `400e577` and I don’t see any blocking issues.

I reproduced the fix through the actual `promptfoo export logs` workflow: `promptfoo-日志.log` survived archive listing and extraction with its contents intact. I also exercised the tar filename boundary with ASCII, CJK, and emoji names exactly 100 UTF-8 bytes long, plus an over-limit name whose next character was a four-byte emoji. The exact-limit names round-tripped correctly, while the over-limit name stopped at the preceding complete code point without invalid UTF-8 or content loss.

Focused validation passed:

- `test/commands/export.test.ts`: 13/13, including shuffled seed `10374`
- TypeScript compilation
- Biome checks on both changed files
- `git diff --check`

The branch is behind `main`, so I also tested a conflict-free synthetic merge with current `main`; the PR remained a two-file diff and the focused tests and compiler passed there as well.

PR #10405, opened 2026-08-15
feat(redteam): account for test generation token usage

jameshiester-oai:
## Summary

- Capture generation usage from successful and failed harmful-test generation requests.
- Persist generation accounting events and attribute each event once, including reused or regenerated test suites.
- Keep generation tokens and requests separate from target usage and probe counts.

## Verification

- 440 focused tests passed across generation, plugins, providers, persistence, schemas, grading, and token utilities.
- `npm run tsc` passed.

Stack 2 of 4; review against the grading-correctness branch. Companion Cloud PR: https://github.com/promptfoo/promptfoo-cloud/pull/4499

### DIFF
--- a/src/redteam/plugins/index.ts
+++ b/src/redteam/plugins/index.ts
@@ -398,6 +398,7 @@ async function fetchRemoteTestCases(
     tokenUsage?: TokenUsage;
   }
 
+  let responseRecorded = false;
   try {
     const { cached, data, status, statusText } = await fetchWithCache<PluginGenerationResponse>(
       getRemoteGenerationUrl(),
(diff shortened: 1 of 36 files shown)


src/redteam/plugins/index.ts
```diff
@@ -409,6 +422,10 @@ async function fetchRemoteTestCases(
     logger.debug(`Received remote generation for ${key}:\n${JSON.stringify(ret)}`);
     return ret;
   } catch (err) {
+    const errorTokenUsage = getErrorTokenUsage(err);
+    if (provider && errorTokenUsage) {
```
zcrab-oai: **[P2] Count failed remote generation attempts without token usage**

Could we record the attempted generation request even when the thrown error has no `tokenUsage`? Ordinary timeouts and network failures currently contribute zero requests, while successful responses without usage and failed harmful-provider calls correctly count one attempt. Recording the failure whenever a provider exists would keep generation request accounting consistent without claiming any unknown token usage.

zcrab-oai (commented):
Found an additional missing-generation-request accounting case.

PR #10407, opened 2026-08-15
feat(redteam): account for multi-turn attacker and grading tokens

jameshiester-oai:
## Summary

- Track target, attacker, and grading usage independently across GOAT, Hydra, Crescendo, iterative, custom, and voice strategies.
- Preserve usage from failed attacker requests and every grading turn, including internal summarization charged to the grading provider.
- Count target probes as target requests and grading requests as grading tasks, not internal model calls.

## Verification

- 872 focused tests passed across all affected strategy providers, evaluator persistence, schemas, generation, and grading.
- `npm run tsc` passed.

Stack 3 of 4; review against the generation-accounting branch. Companion Cloud PR: https://github.com/promptfoo/promptfoo-cloud/pull/4499


src/util/tokenUsageUtils.ts
```diff
@@ -157,12 +157,55 @@ export function accumulateTokenUsage(
...
+  response: { tokenUsage?: Partial<TokenUsage> } | undefined,
+): void {
+  if (!response) {
+    return;
+  }
+
+  target.assertions ??= createEmptyAssertions();
+  accumulateAssertionTokenUsage(target.assertions, {
+    ...response.tokenUsage,
+    numRequests: 1,
```
zcrab-oai, line 224: **[P2] Preserve zero requests for cached strategy-grader responses**

Could we keep cached grading responses at zero requests while continuing to count each fresh grading task once? Built-in providers can return fully cached grading usage, including an explicit `numRequests: 0`, but this assignment always replaces it with `1`. Repeated cached Crescendo, Custom, and iterative grading therefore appears to make new grading requests even when no provider request occurred.

zcrab-oai (commented):
Found a separate cache-hit accounting issue in strategy grading.

zcrab-oai (approved):

PR #10409, opened 2026-08-16
feat(assertions): add script field to javascript, python, and ruby

### LINKED ISSUE #9652: Make it easier to define named assert extensions

morgan-coded:
`assertionTemplates` plus `$ref` already covers named assertion reuse with per-use sibling fields like `config`, but it can't express #9652's first point. On a `javascript`, `python`, or `ruby` assertion, `value` holds the script path, so a call-site `value:` clobbers it and `config: { value: ... }` is the only route today.

This adds an optional `file://`-validated `script:` field to those three types, so with `script` set, `value` renders as the call-site parameter and reaches the script as `context.value` while `context.config` keeps carrying named options. Existing `value: file://...` and inline script forms stay supported, and there is no new top-level key, no separate resolution pass, and no `$ref` change.

This supersedes #10043, rebuilt to the smaller shape I proposed there.

Closes #9652


morgan-coded:
Wrapped a fifth revision round here: fourteen review-bot findings addressed across five pushes, including closing the rendered-secret class end to end — nine sinks audited, four fixed, five verified clean. Checks have come back green after every round, and the connector has queued three more P2-level items on the latest push. Happy to keep folding those in, but wanted your read before another round: do these gate merge, or would you rather take the PR as-is and let the rest land as follow-ups?

PR #10425, opened 2026-08-17
fix(eval): load file:// references nested inside object and array vars

### LINKED ISSUE #1613: File content not loaded when used in vars dict

no-hup:
Fixes #1613.

### The problem

`renderPrompt` loads `file://` vars by walking `Object.entries(vars)` and acting only when the value is a string. A reference nested inside an object or array is never reached, so the raw path string is what ends up in the prompt:

```yaml
tests:
  - vars:
      reporting_period:
        previous:
          report: file://data/report.txt   # arrives as the literal path
```

### The change

Recurse into plain objects and arrays and load nested references the same way top-level text and YAML files are loaded.

Deliberately **not** loaded when nested: JS/Python var scripts, PDFs, and image/video/audio files. The script forms are handed the var name and the base prompt, and the binary forms sit behind `PROMPTFOO_DISABLE_*` env gates — neither has a clear meaning for a value buried inside a structure, and reading a PNG as UTF-8 would quietly corrupt it. Those references are left as the original string, so nothing changes for them.

I kept this as a separate walker rather than refactoring the top-level loader — unifying them means restructuring the ~130-line block inside `renderPrompt`, which felt like the wrong thing to bundle into a bug fix. Happy to do it as a follow-up if you'd prefer them shared.

### Tests

`npx vitest run test/evaluatorHelpers.test.ts` → **109 passed**. Reverting just the `src` change makes the two nested-loading tests fail, so they're pinning the actual behaviour. New cases cover: a reference nested in an object, references nested in an array, nested YAML parsed to JSON, media/PDF left untouched, and a missing nested file propagating its error.

`biome check` and `tsc --noEmit` are clean on both files.

Config from @mldangelo's comment on the issue renders as expected.


no-hup:
Thanks — the Codex pass caught a real one. All four points addressed in 18ba3cd:

**P1, non-plain objects rebuilt (real bug, fixed).** You're right, and it was worse than the report suggests: I reproduced it and a top-level `Date` var was coming back as `{}`, because the walk rebuilt every object from its enumerable entries. Traversal is now limited to arrays and plain records (prototype is `Object.prototype` or `null`); every other object instance is returned untouched, so `Date`/`Map`/`Set`/`Buffer` pass through as before. There's a regression test covering a `Date` both at the top level and nested inside an object.

**P2, uppercase extensions.** Fixed — the extension checks are lowercased now, so `report.PDF` is skipped exactly like `report.pdf`. Test added for `.PDF` and `.PNG`.

On `.docx`/`.zip` specifically I've left the behaviour as-is, because that matches the top-level loader: it also falls through to `readFile(..., 'utf8')` for any extension it doesn't recognise. Making the nested path stricter than the top-level one seemed like the wrong asymmetry to introduce in this PR, but say the word and I'll add a positive allowlist for both.

**P2, unbounded concurrency.** Fixed — the traversal is sequential now rather than `Promise.all` over every leaf, so a dataset-shaped var can't exhaust file descriptors. This also matches the top-level loader, which processes vars one at a time.

**P2, cycle detection.** Already fixed in 07a6108, which landed before this review ran (it was reviewing 50208d2). A `WeakSet` of visited objects short-circuits self-references and returns them unchanged. Your CI is what caught that one, via the circular-reference test in `test/evaluator/execution.test.ts`.

`test/evaluatorHelpers.test.ts` is at 112 passing, `test/evaluator/execution.test.ts` green alongside it, biome and `tsc --noEmit` clean.

no-hup:
Good catch — the cycle guard I added was too blunt. A `WeakSet` made "already visited" mean "return the input untouched", so with `cfg.a` and `cfg.b` pointing at the same `{ report: 'file://report.txt' }`, only `cfg.a.report` got loaded and `cfg.b.report` came back as the raw path. YAML anchors hit this immediately.

Fixed in 5e1dff6 by switching to a `WeakMap` from input object to its output container, registered *before* descending into it. Repeat visits now return the same processed container, so aliases stay aliases, a cycle's back-reference points at the new object rather than the original, and everything reachable still gets loaded. The self-reference test still passes, and there's a new test covering two keys sharing one object.

`npx vitest run test/evaluatorHelpers.test.ts` → 113 passed, and `tsc --noEmit` is clean.

no-hup:
Addressed all four in a58617d:

- `__proto__` own key: clone now preserves the source prototype and copies keys with `defineProperty`, so an own `__proto__` (e.g. from JSON.parse) survives as a data property instead of hitting the legacy setter. Test added.
- Memo map: hoisted a single WeakMap in `renderPrompt` and passed it to every top-level traversal, so objects shared across top-level vars clone once and keep identity. Test added.
- `.YAML`/`.YML`: extension is lowercased before the parse branch, so uppercase YAML parses the same as lowercase. Test added.
- Alias test: `readFileSync` mock is now `mockReturnValueOnce`, so it can't leak into later tests under random order.

no-hup:
Thanks, both were fair catches — addressed in the latest commit.

- Conversation history: `renderPrompt` now skips underscore-prefixed vars entirely when loading nested file refs, so `_conversation` (and other runtime vars, following the convention in `src/util/comparison.ts`) is never traversed — model output can't pull in local file contents. Added a test that a `file://` string inside `_conversation` stays untouched and the array keeps its identity.
- Accessors: the clone now only dereferences own enumerable data properties; accessor properties are copied through by descriptor without invoking the getter. Test covers that the getter isn't called and survives as an accessor on the result.
- security-scan: the failure is infra, not this diff — the job installed fine and then the scan backend returned "Internal server error: We couldn't respond to your request in time" (run 32951658792). Every other Code Scan run in the same window passed, including one started 22 seconds after this one. Should clear on re-run.

no-hup:
Gentle bump on this one. The review points are all addressed and CI's green — it fixes `file://` refs being silently ignored when they're nested inside object or array vars. Happy to rebase or adjust if anything's needed.

PR #10436, opened 2026-08-18
fix: preserve escaped Nunjucks placeholders in test vars

### LINKED ISSUE #10434: Cannot escape {{ placeholder }} syntax in vars
**Describe the bug**

It is not possible, as far as I can tell, to escape variable placeholders within test variables

```yml
vars:
    varOne: 'abc'
    varTwo: "{{ '{{varOne}}' }}" 
    varThree: '{% raw %}{{varOne}}{% endraw %}'
```

In this case, varTwo and varThree will have 'abc' substituted in at runtime despite the use of Nunjacks syntax which, to my knowledge, should prevent this.

A very similar issue was discovered and fixed [here](https://github.com/promptfoo/promptfoo/issues/3855). My issue deals with variable expansion within the test vars, whereas the linked issue deals with variable expansion within the prompt.

**To Reproduce**

Run this minimal config (using your preferred llm provider)

```yml

#providers: ...

prompts:
  - |
    Repeat the next line back exactly, character for character, with no commentary
    [{{template}}]

tests:
  - description: 'Should substitute'
    vars:
      myVar: 'substituted'
      template: '{{myVar}}'

  - description: 'Should not substitute (raw block)'
    vars:
      myVar: 'substituted'
      template: '{% raw %}{{myVar}}{% endraw %}'

  - description: 'Should not substitute (string literal expression)'
    vars:
      myVar: 'substituted'
      template: '{{ "{{myVar}}" }}'
```

**Expected behavior**
Raw blocks and string literal expressions should prevent variable expansion. See the output below from `promptfoo view` - the second and third tests should resolve with [{{myVar}}] rather than [substituted]

**Screenshots**

Image

**System information:**

```
Promptfoo Debug Information
{
  "version": "0.122.0",
  "platform": {
    "os": "win32",
    "release": "10.0.26100",
    "arch": "x64",
    "nodeVersion": "v24.13.1"
  },
  "env": {
    "nodeExtra": "<MY ROOT CERT>",
    "telemetryDisabled": false,
    "telemetryDebug": false
  },
  "configInfo": {
    "configExists": false,
    "configContent": null
  }
}
```
- Promptfoo version: 0.122.0

Oxygen56:
Fixes #10434

## Summary

- Preserve escaped Nunjucks placeholders when they are supplied through test variables.
- Keep raw blocks, comments, and string-literal expressions intact during variable mapping resolution.

## Root Cause

Variable mapping resolution used a regular expression over the complete variable value. This treated placeholder-looking text inside Nunjucks raw blocks and string-literal expressions as real variable references before Nunjucks rendered the value.

## Verification

- Before/after behavioral verification -> both reported escaping forms failed before the fix and passed after it; ordinary nested variable mapping remains unchanged.
- Source and test syntax checks -> passed.
- `git diff --check` -> passed.
- Dependency-backed repository checks -> delegated to CI because this checkout has no installed project dependencies.

## Notes for Reviewers

The change is limited to variable mapping resolution and adds regression coverage for both escaping forms reported in the issue. The core behavior is verified; dependency-backed repository checks will run in CI after publication.

### DIFF
--- a/src/evaluatorHelpers.ts
+++ b/src/evaluatorHelpers.ts
@@ -71,14 +71,17 @@ export function resolveVariables(
         continue;
       }
       const value = variables[key] as string;
-      const match = regex.exec(value);
+      const { protectedValue, restore } = protectNunjucksBlocks(value);
+      const match = regex.exec(protectedValue);
       if (match) {
         const [placeholder, varName] = match;
         if (variables[varName] === undefined) {
           // Do nothing - final nunjucks render will fail if necessary.
           // logger.warn(`Variable "${varName}" not found for substitution.`);
         } else {
-          variables[key] = value.replace(placeholder, variables[varName] as string);
+          variables[key] = restore(
+            protectedValue.replace(placeholder, variables[varName] as string),
+          );
           if (skipResolveVars?.includes(varName) || varsResolvedFromSkipped?.has(varName)) {
             varsResolvedFromSkipped?.add(key);
           }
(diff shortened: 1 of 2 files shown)


Oxygen56:
Thanks for the review. I updated the variable protection to cover whitespace-controlled raw blocks and filtered string-literal expressions, restore nested protected constructs safely, and avoid collisions with user-provided values.

Verification:
- `npx vitest run test/evaluatorHelpers.test.ts --sequence.shuffle=false` -> 109 passed
- `npm run tsc` -> passed
- Biome check and `git diff --check` -> passed (with the existing `renderPrompt` complexity warning)

PR #10437, opened 2026-08-19
fix(providers): return a clean error on empty choices in localai and OpenAI completions

he-yufeng:
## What

Same empty-`choices` treatment that just landed for openrouter/snowflake in #10418, applied to the two remaining unguarded OpenAI-compatible readers: `localai.ts` (chat + legacy completions paths) and `openai/completion.ts`.

A 200 response with an empty or missing `choices` array currently falls into these files' catch-all `try/catch` and comes back as an opaque `API response error: TypeError: Cannot read properties of undefined`. With this change it surfaces as a clean `Malformed response data: …` error, same as the siblings.

The guards check for the choice object's presence (not text truthiness), so a legitimate empty-string completion still passes through.

## Tests

- `test/providers/localai.test.ts`: both LocalAI providers return the clean error on `choices: []`.
- `test/providers/openai/completion.test.ts`: new empty-choices case plus the existing missing-choices parsing test updated to the new message (its old expectation pinned the opaque TypeError text).
- 32/32 pass across both files; biome/eslint/prettier clean.

### DIFF
--- a/src/providers/localai.ts
+++ b/src/providers/localai.ts
@@ -86,6 +86,16 @@ export class LocalAiChatProvider extends LocalAiGenericProvider {
       };
     }
 
+    // A 200 with an empty or missing choices array (soft moderation block or
+    // upstream hiccup) must surface as a clean malformed-response error, not
+    // an opaque TypeError from choices[0].message. Mirrors the sibling fix in
+    // openrouter.ts / snowflake.ts (#10418).
+    if (!data.choices?.[0]?.message) {
+      return {
+        error: `Malformed response data: ${JSON.stringify(data)}`,
+      };
+    }
+
     try {
       return {
         output: data.choices[0].message.content,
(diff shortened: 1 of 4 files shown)


he-yufeng:
The Smoke Tests red was mine: the new completion test's mock carried a severity property that FetchWithCacheResult does not have (TS2353). Dropped it in 6d5d7694, both touched suites pass locally (32/32). Style Check showed a prettier drift that is also covered by the same push. The Redteam (Production API) lane is the known production-API-dependent one, not this change.

he-yufeng:
Rebased this onto current main and reran the repo-pinned Biome 2.5.7. The prior formatting commit was empty, so the expected it.each layout never reached the branch. The branch now passes the full Style Check, root TypeScript check, and both touched provider suites with 32 tests passing.

he-yufeng:
Rebased onto current main (253e7aa83) and re-verified: both providers still index `data.choices[0]` unguarded on main today, so a completion response with an empty `choices` array still crashes with a raw TypeError instead of a clean provider error. The rebased branch keeps the guard shape; `vitest run test/providers/localai.test.ts test/providers/openai/completion.test.ts` passes 32/32 locally.

This has been sitting since late August with no maintainer pass. Small, self-contained, and removes a crash class that any OpenAI-compatible gateway can trigger under rate limiting. Happy to adjust if you'd rather see the guard in a shared helper than at the two call sites.

PR #10449, opened 2026-08-20
test(providers): stabilize Gemini Live streaming timeout

mldangelo-oai:
## Summary

- Replace wall-clock scheduling in the Gemini Live continuous-streaming timeout regression test with Vitest fake timers.
- Preserve the four streamed chunks, 120 ms idle timeout, and 300 ms total stream while removing Windows runner scheduling races.
- Restore real timers in a `finally` block to prevent shuffled test-order pollution.

## Why

The Windows shard in https://github.com/promptfoo/promptfoo/actions/runs/32401546355/job/96530686427 failed because delayed stream callbacks let the real 120 ms idle timeout fire before the next mocked chunk was delivered. The dependency-only PR did not modify provider behavior.

## Validation

- Reproduced the exact CI failure by delaying two mocked WebSocket callbacks; the updated test passes the same injected-stall reproducer.
- Temporarily removed the provider's idle-timeout rearm and verified the fake-timer regression test still fails with the original 120 ms timeout.
- `npx vitest run test/providers/google/live.test.ts test/providers/websocket.test.ts test/util/xlsx.test.ts test/util/xlsxEntities.test.ts test/package-manifests.test.ts test/test-hygiene.test.ts --configLoader runner` (326 tests passed).
- `npx vitest run test/providers/google/live.test.ts --configLoader runner --sequence.seed=1` (92 tests passed).
- `npx vitest run test/providers/google/live.test.ts --configLoader runner --sequence.seed=42` (92 tests passed).
- `npm run tsc`.
- `npm run l && npm run f`.

### DIFF
--- a/test/providers/google/live.test.ts
+++ b/test/providers/google/live.test.ts
@@ -1779,31 +1779,38 @@ describe('GoogleLiveProvider', () => {
   });
 
   it('keeps a continuously streaming response alive past the idle-timeout window', async () => {
-    provider = new GoogleLiveProvider('gemini-2.0-flash-exp', {
-      config: {
-        generationConfig: { response_modalities: ['text'] },
-        timeoutMs: 120,
-        apiKey: 'test-api-key',
-      },
-    });
-    vi.mocked(WebSocket).mockImplementation(function () {
-      setImmediate(() => {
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
@codex review

PR #10454, opened 2026-08-21
fix(providers): preserve HTTP transform errors

mldangelo-oai:
HTTP response transforms now preserve error-only provider responses and explicit falsey outputs (`''`, `false`, and `0`) through the shared response normalizer. This surfaces a transform's error directly instead of nesting it inside `output`.

The branch is integrated with current main and retains main's original-response token accounting.

Validation: 100 focused tests, TypeScript, core build, lint/format, and hooks pass. Full backend: 24,993 passed; all 37 transient socket failures pass a 38-test retry, with seven known shared-install dependency mismatches remaining. Four built package API cases and a built CLI eval against a local HTTP server confirm error propagation, falsey outputs, and original token usage.


mldangelo-oai:
@codex review

jameshiester-oai (commented):
Reviewed base 127d90534b9c1b1ba4554f007dd4b5fd2c8bf1b4 and exact head 01b51f3e62264e25f3f92b41d404be175615dd7a.

Scope: fresh full base...head review. Verdict: one verified P2 finding. The connector exposed no failing combined-status entries; Codecov reports all modified coverable lines covered.

Verified findings:
- P2 — Empty-string transformed outputs still receive token usage for the raw HTTP body when token estimation is enabled. In `src/providers/http.ts:3181`, `getCompletionText` checks `parsedOutput?.output` by truthiness before checking its type, so the newly preserved `{ output: '' }` path falls back to `rawText`. Reproduction: a local HTTP endpoint returned `{"ok":true}`; with `transformResponse: () => ({ output: "" })` and `tokenEstimation.enabled: true`, the eval output was the expected empty string but the exported/CLI usage reported 2 completion tokens. This overstates completion and total usage for HTTP providers that intentionally normalize a response to empty output. Check `typeof parsedOutput?.output === 'string'` without the truthiness guard and add a token-estimation regression beside the new falsey-output cases.

Existing feedback:
- No unresolved inline threads. Existing Codex/Copilot/security summaries reported no actionable defect, so there is no duplicate inline feedback.

Simplification and test gaps:
- Reusing `normalizeResponseTransformResult` is the right simplification.
- The new falsey-output tests cover returned values, but not the adjacent token-estimation path above.

Validation performed:
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/http/request.test.ts --sequence.shuffle=false` — passed, 1 file / 98 tests.
- Local uncached CLI eval against an isolated localhost HTTP server with `PROMPTFOO_CONFIG_DIR` and `PROMPTFOO_CACHE_PATH` under `/private/tmp`:
  - `{ output: 0 }` — passed; exported output was `0`.
  - `{ output: '' }` with token estimation — output passed, but reproduced the incorrect 2 completion tokens.
- Three read-only specialist fallback reviews completed; two found no defects and one surfaced the verified token-estimation issue.

Limitations:
- The requested PR Review Toolkit command was unavailable in this task, so I used the documented read-only specialist fallback.
- I did not run the full build or full repository suite; the changed provider suite, typecheck, and real CLI path were exercised.

mldangelo-oai:
@codex review

jameshiester-oai (commented):
## Scope and verdict

Reviewed base `3ec974043eb9941af5b605f6f21f71f24bf46c3c` through exact head `a6d9c85e98b79c7797e86123fbbcdca612c511a9`, with extra attention to changes since authenticated reviewed ancestor `01b51f3e62264e25f3f92b41d404be175615dd7a`. The full current PR changes only the HTTP provider and its request tests; the large ancestry delta is primarily the merge from main. I inspected both HTTP request paths, shared transform normalizer, token accounting, changed tests, prior feedback, and CI. No distinct new issue merits an inline comment. All 43 current-head checks report success.

## Existing finding and coverage

The previously reported P2 empty-output token-estimation issue is still present at `src/providers/http.ts:3251`: `getCompletionText()` truth-tests `parsedOutput.output`, so an authored `{ output: '' }` falls back to the raw HTTP body even though the normalized output is empty. A fresh built CLI eval against a local JSON endpoint returned the correct empty output but charged two completion tokens for the raw body. The earlier authenticated review already raised this, so I am not duplicating it inline. The new tests assert the returned error and falsey outputs, but do not assert corresponding token accounting. The shared normalizer is a sound reuse for the changed response shapes.

## Verification

- `npx vitest run test/providers/http/request.test.ts test/providers/transformResult.test.ts --sequence.shuffle=false`: 2 files, 100 tests passed after `npm ci --include=dev --ignore-scripts` for this head. An initial run with the previous PR's stale `js-yaml` installation failed before test collection; it passed after the locked install.
- `npm run tsc -- --pretty false`, `npx tsdown`, and `node --import tsx scripts/postbuild.ts`: passed. `git diff --check` passed; worktree clean.
- Built CLI `node dist/src/main.js eval` with `--no-cache --no-share` and an isolated localhost JSON endpoint: the transformed `{ error: 'Provider guardrail blocked' }` produced one expected provider error (exit 100, no nested output); the `{ output: '' }` case passed its empty-output assertion (exit 0, score 1) while reproducing the two wrongly estimated completion tokens. Both exported JSON files were inspected.

No external provider or credentialed call was made. The requested `/pr-review-toolkit:review-pr all parallel` command exited 127, so I applied the six read-only review lenses manually and reconciled three specialist passes.

PR #10460, opened 2026-08-23
fix(webui): handle malformed login redirects

GuanyuMi:
## Summary

- safely decode the login `redirect` query parameter
- fall back to `/` when the redirect contains malformed percent encoding
- add regression coverage for invalid redirect encoding

## Problem

The login page directly passed the `redirect` query parameter to `decodeURIComponent`.

Malformed input such as:

```text
/login?redirect=%
```

caused `decodeURIComponent` to throw `URIError: URI malformed`. For an authenticated user, this interrupted the redirect effect and prevented the login page from navigating normally.

## Fix

Decode the redirect through a small helper that returns `null` when decoding fails. The existing redirect validation then handles the invalid value and navigates to `/`.

Valid internal redirects, including percent-encoded paths and redirects containing query parameters, retain their existing behavior.

## Validation

- `npm run test:app -- src/pages/login.test.tsx --run` — 15 tests passed
- `npm run test:app -- --run` — 4,867 tests passed across 279 files
- `npm --prefix src/app run tsc -- --noEmit`
- `npx @biomejs/biome check src/app/src/pages/login.tsx src/app/src/pages/login.test.tsx`
- `git diff --check`


mldangelo-oai (approved):
Thanks!

PR #10467, opened 2026-08-24
feat(providers): add Codex Security SDK provider

vsauter-oai:


mldangelo-oai:
this is cool!

src/providers/openai/codex-security.ts
```diff
@@ -0,0 +1,830 @@
...
+    if (config.severity) {
+      args.push('--severity', config.severity);
+    }
+    if (config.finding_id) {
+      args.push(config.finding_id);
+    } else if (!config.scan_id) {
+      const finding = config.finding_file
+        ? resolveConfigPath(config.finding_file)!
+        : (config.finding ?? prompt);
+      args.push(typeof finding === 'string' ? finding : JSON.stringify(finding));
```
mldangelo-oai: Thanks for adding explicit write opt-ins here. There are two related trust-boundary gaps if remediation remains in this PR: finding text can be parsed as a CLI flag (for example `--resume-pr=...`, which can trigger GitHub writes), and a saved scan can select its original repository instead of the configured isolated checkout. Could we defer the CLI path, or ensure positional findings cannot become options and verify the resolved scan repository matches the authorized checkout before any mutation?

src/providers/openai/codex-security.ts
```diff
@@ -0,0 +1,830 @@
...
+      id: this.providerId,
+      env: this.env,
+      config: {
+        ...(config.model ? { model: config.model } : {}),
+        ...(config.model_provider ? { model_provider: config.model_provider } : {}),
+        ...(effort ? { model_reasoning_effort: effort } : {}),
+        ...(config.maxRetries === undefined ? {} : { maxRetries: config.maxRetries }),
+        working_dir: repository,
+        sandbox_mode:
+          config.sandbox_mode ?? (config.allow_file_writes ? 'workspace-write' : 'read-only'),
```
mldangelo-oai: I like the explicit `allow_file_writes` safeguard. Could we apply it to the effective sandbox as well? Right now an operation such as `threat-model` can set `sandbox_mode: workspace-write` or `danger-full-access` even when `allow_file_writes` is unset, which bypasses the intended read-only default. Reject writable modes without explicit permission, or defer delegated skills from this first pass.

mldangelo-oai (commented):
Thanks for taking this on. A first-class Codex Security provider with structured findings, coverage, and cost reporting would be genuinely useful.

Could we make the first pass smaller and safer by focusing on native SDK scans (standard, deep, and diff) and native finding validation where needed? The remediation commands introduce a separate CLI trust boundary, and the standalone operations largely duplicate functionality already available through `openai:codex-sdk`. Moving those paths to focused follow-ups would simplify the provider, configuration UI, and tests while avoiding several current concerns around permissions, credentials, repository isolation, and argument handling.

I would also keep the initial UI focused on essential settings, include evidence from one real SDK-backed evaluation, and address the existing review threads and CI failures before landing. Happy to help think through a clean first slice.

src/providers/openai/codex-security.ts
```diff
@@ -0,0 +1,830 @@
...
+import type {
+  CodexSecurity,
+  JsonObject,
+  ScanCost,
+  ScanOptions,
+  ScanResult,
+  ValidationOptions,
+} from '@openai/codex-security';
+
+import type { TokenUsage } from '../../contracts/shared';
```
mldangelo-oai: The current Style Check points directly at this import: `providers -> contracts` is a forbidden new architecture edge. Existing OpenAI providers import `TokenUsage` from `../../types/index`, and this file already imports its provider response types from there. Could we add `TokenUsage` to that existing type import and remove this direct `contracts/shared` import? That should resolve this particular violation without broadening the architecture baseline.

src/providers/openai/codex-security.ts
```diff
@@ -0,0 +1,830 @@
...
+          error:
+            'Codex Security operation "track-findings" can create or update external issues. Set allow_external_writes: true to enable it.',
+        };
+      }
+
+      if (callOptions?.abortSignal?.aborted) {
+        return { error: 'Codex Security operation was aborted before it started.' };
+      }
+
+      if (operation === 'fix-finding' || operation === 'verify-fix') {
```
mldangelo-oai: One concrete way to keep the first release manageable: this branch currently combines four native SDK operations, two subprocess-backed operations, and eight operations that delegate to the existing `OpenAICodexSDKProvider`. Could we keep `security-scan`, `deep-security-scan`, `security-diff-scan`, and, if needed, `validation` here, then move the other paths into focused follow-ups? Removing `runCliOperation` and `runStandaloneSkill` alone would eliminate about 157 provider lines and roughly 230 backend-test lines, plus much of the permission and UI branching.

src/app/src/pages/redteam/setup/components/Targets/CodexSecurityConfiguration.tsx
```diff
@@ -0,0 +1,480 @@
...
+  const isDeepScan = operation === 'deep-security-scan';
+  const isDiffScan = operation === 'security-diff-scan';
+  const isFindingOperation = ['validation', 'fix-finding', 'verify-fix'].includes(operation);
+  const isRemediationOperation = ['fix-finding', 'verify-fix'].includes(operation);
+  const changesRepository = ['fix-finding', 'define-security-policy'].includes(operation);
+  const [scopedPaths, setScopedPaths] = useState(() =>
+    Array.isArray(config.paths) ? config.paths.join(', ') : '',
+  );
+
+  const updateOperation = (nextOperation: string) => {
```
mldangelo-oai, line 81: I like making the provider discoverable in the UI. Could we start with a compact scan-first form covering repository, model, scan mode, any relevant diff target, and an optional cost cap, with specialist knobs left to YAML or advanced configuration? Supporting all 14 operations here duplicates the backend catalog and requires this operation-switch cleanup matrix. The new-file coverage check is also currently failing at 22/32 functions (68.75% versus the required 80%), so narrowing the surface should remove state-management code instead of requiring more tests for functionality we may not need in the first release.

src/app/src/pages/eval-creator/components/AddProviderDialog.test.tsx
```diff
@@ -18,12 +20,83 @@ vi.mock('@app/pages/redteam/setup/components/Targets/ProviderTypeSelector', () =
...
+    provider: { id: string; config: Record<string, unknown> };
+    providerType?: string;
+    setProvider: (provider: { id: string; config: Record<string, unknown> }) => void;
+    setError?: (error: string | null) => void;
+    validateAll?: boolean;
+  }) => {
+    const repository =
+      typeof provider.config.repository === 'string' ? provider.config.repository : '';
+
+    useEffect(() => {
```
mldangelo-oai, line 71: Could we avoid implementing the production repository validator inside this mock? The `useEffect` independently decides whether the repository is valid, so the dialog test can pass even if the real `ProviderConfigEditor` behaves differently. Since the editor already has its own focused validation tests, either keep this dialog mock minimal and assert the wiring, or render the actual editor for the save flow. That would reduce scaffolding while making the test boundary clearer.

test/providers/openai-codex-security.test.ts
```diff
@@ -0,0 +1,885 @@
...
+
+      await provider.callApi('Validate');
+
+      expect(mockValidate).toHaveBeenCalledWith(
+        expect.objectContaining({ finding: 'Unchecked redirect in /login' }),
+      );
+    });
+  });
+
+  describe('remediation CLI', () => {
```
mldangelo-oai: These permission tests are a good start. If remediation and delegated skills remain in this PR, could we add focused adversarial regressions for: (1) a finding beginning with `--resume-pr=review-proof`; (2) a saved scan whose original `targetPath` differs from the configured `repository`; (3) `sandbox_mode: workspace-write` without `allow_file_writes`; and (4) unrelated parent-process secrets not being inherited by the child? The current mocks check CLI arguments and `cwd`, but they never exercise how the CLI resolves a saved repository, which is why the cross-repository case currently slips through. If those paths move to follow-up PRs, these tests can follow them there.

test/providers/openai-codex-security.test.ts
```diff
@@ -0,0 +1,885 @@
...
+          cli_env: { SECURITY_EVAL: true },
+        },
+      });
+
+      const response = await provider.callApi('Patch findings');
+
+      expect(execFile).toHaveBeenCalledWith(
+        process.execPath,
+        [
+          '/packages/@openai/codex-security/bin/codex-security.mjs',
```
mldangelo-oai: Small CI fix: this hard-coded POSIX path is the exact mismatch in the failing Windows shard. The runner receives a drive-qualified Windows path such as `D:\packages\@openai\codex-security\bin\codex-security.mjs`, while this assertion expects forward slashes. Could we derive the expected executable with `path.resolve(...)` or use a platform-neutral matcher? That should fix the shard without changing provider behavior.

examples/openai-codex-security/README.md
```diff
@@ -0,0 +1,27 @@
+# openai-codex-security (OpenAI Codex Security Scan Comparison)
+
+Compare Codex Security standard and deep scans across models and reasoning settings using the same intentionally vulnerable repository fixture.
```
mldangelo-oai, line 3: Tiny CI unblocker: the example README suite requires at least one H2 section, and this file currently has only the H1 title. Adding a small heading such as `## Setup` before the installation instructions should clear the same README failure across the Linux, macOS, and Windows jobs.

mldangelo-oai (commented):
Thanks again for putting together the SDK integration, docs, and example. I added a few concrete suggestions below to make the smaller native-SDK-first slice easier to identify and to point directly at the current CI failures.

The goal is to preserve the useful structured scan integration and meaningful security coverage while moving subprocess remediation and delegated skills into focused follow-ups. If those extra paths stay in this PR, the adversarial regression cases below would help make their safety boundaries explicit. One useful final QA artifact would be a sanitized result summary from `npm run local -- eval -c examples/openai-codex-security/promptfooconfig.yaml --no-cache -o output.json` on an authorized repository.

vsauter-oai:
### Review follow-up

Updated the first release to the four native Codex Security SDK operations only:

- `security-scan`
- `deep-security-scan`
- `security-diff-scan`
- `validation`

Removed subprocess remediation and delegated security skills, simplified the provider configuration UI, corrected aggregate scan token accounting and complete-operation latency, restricted SDK loading to trusted Promptfoo installation paths, honored programmatic provider configuration base paths, normalized repository/reasoning aliases and pasted model IDs, selected automatic authentication by default, rejected unsupported provider-scoped credentials, imported operations, and invalid diff targets, accepted generic per-test options while preserving strict provider configuration validation, excluded security scanners from agent-rubric grading while retaining agentic-runtime classification, replaced the CodeQL-triggering example with an authorization-bypass fixture, and aligned Windows path assertions with shared path-resolution behavior. Reconciled the latest upstream provider-editor changes, verified the actual Codex Security configuration form rejects missing repositories before accepting valid ones, and clarified same-installation SDK requirements and the Node 22.22 minimum.

### Real SDK-backed evaluation

Ran the checked-in example against its authorized fixture with the `standard-terra-medium` provider:

```bash
npm run local -- eval -c examples/openai-codex-security/promptfooconfig.yaml \
  --filter-providers standard-terra-medium --no-cache --max-concurrency 1 -o output.json
```

- Result: **1 passed, 0 failed, 0 errors**.
- Operation/model: `security-scan` / `gpt-5.6-terra`, `medium` reasoning.
- Validated findings: **1**; repository coverage: **complete**.
- Aggregate usage: **1,315,752 tokens**: 1,303,830 prompt, 11,922 completion, 1,174,016 cached, and 2,695 reasoning.
- SDK-reported estimated cost: **$0.6374952**.
- SDK/plugin: `@openai/codex-security@0.1.18` / plugin `0.1.22`.

### Local verification

- Focused backend suite: **2,264 tests passed**.
- Provider configuration and eval-creator web UI suite: **1,023 tests passed across 49 files**.
- New provider coverage: **97.90% statements, 96% functions, 98.59% lines**.
- New UI configuration coverage: **89.09% statements, 100% functions, 100% lines**.
- Passed backend/frontend coverage ratchets, both TypeScript checks, repository-wide Biome CI, Prettier, Knip, and architecture boundaries.

site/docs/guides/test-agent-skills.md
```diff
@@ -248,6 +248,8 @@ providers:
 
 Codex discovers project skills from `.agents/skills/` under each `working_dir`. Its `skill-used` signal is inferred from successful reads of the matching `SKILL.md`, so keep `enable_streaming: true` while developing the eval if you want Promptfoo to collect that evidence. The earlier JavaScript assertion should use `JSON.parse(output)` with Codex because `output_schema` keeps `output` as a JSON string rather than a parsed object.
 
+For managed security workflows rather than general local skills, use the [Codex Security SDK provider](/docs/providers/openai-codex-security). It runs native standard, deep, and diff scans plus finding validation, and returns structured findings, coverage, token usage, and estimated scan cost.
```
mldangelo-oai: probably remove this

src/providers/agentic-utils.ts
```diff
@@ -51,6 +52,18 @@ export function isAgenticProvider(provider: ApiProvider | null | undefined): boo
   );
 }
 
+/**
+ * Security scanners run agentic workflows but cannot return rubric grading verdicts.
+ */
+export function isAgenticGradingProvider(provider: ApiProvider | null | undefined): boolean {
```
mldangelo-oai: is this function absolutely necessary?

src/providers/openai/codex-runtime.ts
mldangelo-oai: recommend deleting this file

jameshiester-oai (approved):

PR #10475, opened 2026-08-24
fix(vars): respect {% raw %} and nested expressions when pre-resolving placeholders

### LINKED ISSUE #10434: Cannot escape {{ placeholder }} syntax in vars

he-yufeng:
Fixes #10434.

## Summary

`{% raw %}` and quoted-literal escapes in `vars:` values did not survive prompt rendering, because the regex-based pre-resolution pass substituted `{{var}}` placeholders before Nunjucks ever saw them. `{% raw %}{{varOne}}{% endraw %}` became `{% raw %}abc{% endraw %}`, and `{{ '{{varOne}}' }}` became `{{ 'abc' }}`, so both rendered as `abc` instead of the literal placeholder text.

## Root cause

`resolveVariables` matched `/\{\{\s*(\w+)\s*\}\}/` against the raw string with no awareness of Nunjucks structure. A placeholder sitting inside a `{% raw %}...{% endraw %}` span, or nested inside an outer `{{ ... }}` expression, is not a simple placeholder, but the regex replaced it anyway.

## Fix

The pre-resolution pass now picks the first placeholder that is neither inside a raw block nor nested in an unclosed outer expression, and splices by index instead of a global string replace. Anything structural is left for the real Nunjucks render, which already handles both shapes correctly.

## Verification

- Reproduced the issue's exact YAML shapes end to end through `renderPrompt`: before, `three=[abc]`; after, `two=[{{varOne}}] three=[{{varOne}}] one=[abc]`.
- New unit tests pin the raw-block escape, the nested-expression escape, and substitution still working around protected spans. `npx vitest run test/evaluatorHelpers.test.ts`: 106 passed.
- `tsc --noEmit` clean; Biome reports only the pre-existing cognitive-complexity warning on `renderPrompt`.

### DIFF
--- a/src/evaluatorHelpers.ts
+++ b/src/evaluatorHelpers.ts
@@ -57,7 +57,6 @@ export function resolveVariables(
   varsResolvedFromSkipped?: Set<string>,
 ): Record<string, VarValue> {
   let resolved: boolean;
-  const regex = /\{\{\s*(\w+)\s*\}\}/; // Matches {{variableName}}, {{ variableName }}, etc.
 
   let iterations = 0;
   do {
@@ -71,14 +70,17 @@ export function resolveVariables(
         continue;
       }
       const value = variables[key] as string;
-      const match = regex.exec(value);
+      const match = firstUnprotectedPlaceholder(value);
       if (match) {
-        const [placeholder, varName] = match;
+        const { placeholder, varName, index } = match;
         if (variables[varName] === undefined) {
           // Do nothing - final nunjucks render will fail if necessary.
           // logger.warn(`Variable "${varName}" not found for substitution.`);
         } else {
-          variables[key] = value.replace(placeholder, variables[varName] as string);
+          variables[key] =
+            value.slice(0, index) +
+            (variables[varName] as string) +
+            value.slice(index + placeholder.length);
           if (skipResolveVars?.includes(varName) || varsResolvedFromSkipped?.has(varName)) {
             varsResolvedFromSkipped?.add(key);
           }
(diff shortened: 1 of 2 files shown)


he-yufeng:
The Node 26 failure is not this diff: the failing test (eval watch paths, --vars source recovery) lives in test/commands/eval.test.ts and this change only touches vars/matcher rendering, with zero shared files. The same watch-mode area just moved on main (#10562), and the failure is in that neighborhood. I cannot rerun the job myself (no admin rights on the repo), so it needs a maintainer rerun; everything else is green after the rebase.

PR #10479, opened 2026-08-24
fix(redteam): account for all scan-time model calls

jameshiester-oai:
## Summary

- Attribute authoritative-markup attacks, memory-poisoning scenarios, MCP materialization, and indirect-web page generation to attacker usage without inflating target probes.
- Attribute blocking-question analysis to grading; preserve usage across runtime transforms, cached responses, failed requests, and text-only HTTP response transforms.
- Count the final iterative-tree target request even when the provider does not report token usage.

Companion Cloud changes: https://github.com/promptfoo/promptfoo-cloud/pull/4499

## Verification

- 576 focused upstream strategy/provider tests passed; TypeScript check passed.
- 103 focused Cloud tests passed; Cloud server/shared typechecks and builds passed.
- Local six-probe scan reconciled exactly: 17,268 total = 3,674 target + 4,343 generation + 3,821 attacker + 5,430 grading.


mldangelo-oai (approved):
stamping. We should rethink the architecture / see if there is a more elegant way to do this

PR #10488, opened 2026-08-25
fix(matchers): mark grader failures as grader errors in factuality and closed-QA

### LINKED ISSUE #10481: Do not collapse factuality/closed-QA grader failures into assertion failures
## Problem

Model-grader execution failure and assertion failure are different outcomes:

```text
grade says criterion is false  !=  grader could not produce a valid grade
```

Promptfoo already encodes this distinction in `graderFail(..., metadata.graderError=true)`, and the nearby `isGraderFailure` comment explicitly says a grader error is not evidence that the criterion was or was not met.

However, current `matchesFactuality()` and `matchesClosedQa()` still do this when the grading provider returns an error or no output:

```ts
if (resp.error || !resp.output) {
  return fail(resp.error || 'No output', resp.tokenUsage);
}
```

That produces an ordinary `pass: false, score: 0` result with no grader-error marker. `matchesFactuality()` also uses plain `fail(...)` for malformed grader output.

This is especially problematic in CI/regression evaluation: transient judge/provider failures can look like regressions in the system being evaluated.

## Expected behavior

A provider/transport/no-output/malformed-grader failure should remain distinguishable from a valid semantic grade of `false`.

At minimum, these paths could use the existing `graderFail(...)` semantics consistently. Ideally the final evaluation/reporting layer should surface exhausted grader failures as `ERROR`/unknown rather than include them as ordinary assertion failures in quality metrics.

A useful regression would make the grader provider return an error and verify that the result is marked as grader failure, while a valid grader response of `N`/factual disagreement remains a normal failed assertion.

Related: #7077 reported the same practical symptom for unstable LLM graders; `llm-rubric` has since moved toward row-level error handling, but these sibling model-graded paths still collapse the two states.

he-yufeng:
Fixes #10481.

## Summary

`matchesFactuality` and `matchesClosedQa` collapsed grader execution failures into ordinary assertion failures: a provider error or an empty/malformed grader response came back as `pass: false, score: 0` with no marker, so a transient judge outage reads as a regression in the system under test.

## Fix

Those paths now return through `graderFail` (the existing `metadata.graderError` convention from `matchers/shared.ts`) instead of plain `fail`: provider/transport errors, empty outputs, and malformed grader answers (including the closed-QA branch that ends in neither Y nor N, and the factuality JSON/legacy parse failures). A grader failure is not evidence that the criterion was or was not met, and inverse-aware callers already propagate marked results verbatim.

## Verification

- New test pins the provider-failure shape (`pass: false` with `metadata: { graderError: true }`); the existing invalid-category case updated to the new intentional shape.
- `npx vitest run test/matchers test/assertions`: 1723 passed (81 files).
- `tsc --noEmit` clean; Biome reports only the pre-existing `renderPrompt` complexity warning.


he-yufeng:
Same defect family as #10904, complementary rather than duplicate: this PR marks grader failures as grader errors on the factuality failure paths, while #10904 stops `not-classifier` / `not-search-rubric` from inverting a grader/transport failure into `pass: true`. A reviewer landing on either should see both so the family ships coherently.

PR #10494, opened 2026-08-25
fix(matchers): tag RAG grader failures so inverse assertions cannot pass

### LINKED ISSUE #9910: bug(assertions): context-recall and context-relevance default threshold to 0, silently always pass

Varshith-Kali:
## Summary

Follow-on to #10202, which landed the shared RAG threshold default (`DEFAULT_RAG_ASSERTION_THRESHOLD = 0.5`) and `applyRagInverse()`. This PR fixes the matcher-layer half of the same bug.

`applyRagInverse()` correctly refuses to invert a grader failure — but it detects one via `metadata.graderError`, and the RAG matchers returned a plain, **untagged** `fail()` on a grading-provider error. The guard therefore never fired, and `not-context-recall` / `not-context-relevance` / `not-answer-relevance` would still turn a grading outage into a silent pass.

All affected error paths now return `graderFail(...)`, matching the precedent already set in `src/matchers/search.ts` ("a provider that errored gave no verdict to invert on"):

- `matchesContextRecall` — grading-provider error
- `matchesContextRelevance` — grading-provider error
- `matchesAnswerRelevance` — grading-provider error, input-embedding error, and per-question embedding error

### Off-format grader output is a grading error, not a verdict

A non-empty reply that does not follow the rubric — a refusal, an apology, an out-of-format explanation — is a failed grading attempt, not a legitimate low score. It is now tagged so it cannot be inverted into a pass:

- `context-recall`: output with no `[Attributed]` / `[Not Attributed]` classifications at all.
- `context-relevance`: output that is neither `Insufficient Information` nor a verbatim quote of a context sentence.

This matters more now that the default threshold is meaningful. Measured against a tree with #10202 applied, the grader reply *"I am sorry, but I cannot answer that from the provided context."* over a two-sentence context was split into one "relevant sentence", scored **exactly 0.50, and passed** `context-relevance` at the new 0.5 default. It now scores 0 and fails, tagged as a grader error.

Both adjacent paths are unchanged: a legitimate verbatim echo still scores normally, and `Insufficient Information` still yields a genuine 0 with **no** `graderError` tag, so it still inverts correctly.

Fixes #9910

## Coverage

`#10202` ships `applyRagInverse()` with no grader-error tests at all, and only tests the below-threshold → invert-to-pass direction. This PR adds the missing half:

- grader errors survive `not-context-recall` / `not-context-relevance` untouched (pass=false, score=0, reason preserved, `metadata.graderError`)
- the pass=true → invert-to-fail direction at the default threshold
- matcher-level coverage in `test/matchers/context-recall.test.ts`, `context-relevance.test.ts` and `answer-relevance.test.ts` — files #10202 does not touch

The two new `answer-relevance` cases were verified as effective regression guards: reverting just the `graderFail` change makes both fail, and restoring it makes them pass.

## Scope change from the original PR

This PR originally also changed `src/assertions/contextRecall.ts` / `contextRelevance.ts` to an inline `?? 0.7` literal with hand-rolled inversion. That approach was superseded by #10202's shared constant at **0.5**, so those changes are dropped and the docs edits with them. What remains is the matcher fix, which #10202 does not make, plus the regression coverage above.

The original threshold work is credited via `Co-authored-by` on #10202's squash commit.

### DIFF
--- a/site/docs/configuration/expected-outputs/index.md
+++ b/site/docs/configuration/expected-outputs/index.md
@@ -176,10 +176,10 @@ See [Model-graded evals](/docs/configuration/expected-outputs/model-graded), [cl
 | [moderation](/docs/configuration/expected-outputs/moderation)                                        | Check output against safety policies and include provider-reported usage metrics |
 | [llm-rubric](/docs/configuration/expected-outputs/model-graded)                                      | LLM output matches a given rubric, using a Language Model to grade output        |
 | [g-eval](/docs/configuration/expected-outputs/model-graded/g-eval)                                   | Chain-of-thought evaluation based on custom criteria using the G-Eval framework  |
-| [answer-relevance](/docs/configuration/expected-outputs/model-graded)                                | Ensure that LLM output is related to original query                              |
-| [context-faithfulness](/docs/configuration/expected-outputs/model-graded)                            | Ensure that LLM output uses the context                                          |
-| [context-recall](/docs/configuration/expected-outputs/model-graded)                                  | Ensure that ground truth appears in context                                      |
-| [context-relevance](/docs/configuration/expected-outputs/model-graded)                               | Ensure that context is relevant to original query                                |
+| [answer-relevance](/docs/configuration/expected-outputs/model-graded)                                | Ensure that LLM output is related to original query (default threshold 0.5)      |
+| [context-faithfulness](/docs/configuration/expected-outputs/model-graded)                            | Ensure that LLM output uses the context (default threshold 0.5)                  |
+| [context-recall](/docs/configuration/expected-outputs/model-graded)                                  | Ensure that ground truth appears in context (default threshold 0.5)              |
+| [context-relevance](/docs/configuration/expected-outputs/model-graded)                               | Ensure that context is relevant to original query (default threshold 0.5)        |
 | [conversation-relevance](/docs/configuration/expected-outputs/model-graded)                          | Ensure that responses remain relevant throughout a conversation                  |
 | [trajectory:goal-success](/docs/configuration/expected-outputs/model-graded/#trajectorygoal-success) | Use an LLM judge to decide whether the traced agent run achieved its goal        |
 | [factuality](/docs/configuration/expected-outputs/model-graded)                                      | LLM output adheres to the given facts, using Factuality method from OpenAI eval  |
--- a/site/docs/configuration/expected-outputs/model-graded/answer-relevance.md
+++ b/site/docs/configuration/expected-outputs/model-graded/answer-relevance.md
@@ -17,6 +17,8 @@ assert:
     threshold: 0.7 # Score between 0 and 1
 ```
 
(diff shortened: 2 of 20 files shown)


Varshith-Kali:
## CI fix pushed (commit 2296088)

**Root cause:** Both contextRecall.ts and contextRelevance.ts handlers were setting metadata: { context } which **replaced** the full matcher metadata object (score, extractedSentences, relevantSentenceCount, totalContextSentences, insufficientInformation). Tests expected all of this metadata to be preserved.

**Fix:** Changed to:
`	s
metadata: {
  ...(typeof result.metadata === 'object' ? result.metadata : {}),
  context,
}
`

This merges context into existing matcher metadata instead of overwriting it. Also fixed a 
ot-context-recall test assertion where mock data was inconsistent (
esolveContext returned 'test context' but ssertionValueContext.vars.context had 'incomplete context').

**Local verification:** 19/19 tests pass (10 contextRecall + 9 contextRelevance). Awaiting CI re-run.

mldangelo:
Thanks for this — and apologies for the wait.

Five PRs independently fixed the same RAG threshold bug (#10202, #10494, #9907, #9911, #10567). We landed **#10202** as the shared design: `DEFAULT_RAG_ASSERTION_THRESHOLD = 0.5` plus an `applyRagInverse()` helper, with all five of us credited as co-authors on the squash commit.

But your PR is **not** a duplicate, and that is worth spelling out. While rebasing these I found that `applyRagInverse()` detects a grader failure via `metadata.graderError` — and the RAG matchers were returning an *untagged* `fail()` on grading-provider errors. So the guard never fired, and the `not-*` variants would still turn a grading outage into a silent pass. Your matcher-layer fix is what closes that, and #10202 does not touch `src/matchers/rag.ts` at all.

Concretely, with #10202 alone, the grader reply *"I am sorry, but I cannot answer that from the provided context."* over a two-sentence context scored **exactly 0.50 and passed** `context-relevance` at the new default. That is fixed here.

I have rebased this onto current `main`, dropped the threshold changes that #10202 supersedes, kept your matcher fix and the coverage for it, and retitled it to match what it now does. Your test cases were verified as effective regression guards — they fail without the fix and pass with it.

Full CI is green. Thanks for the careful work here.

Varshith-Kali:
Thanks Michael — really appreciate you folding in the matcher-layer fix and keeping the regression coverage while #10202 set the shared default. The `graderError` tagging is the piece that stops a grading outage from turning into a silent pass on the `not-*` assertions, so I'm glad it landed. Happy to follow up if anything downstream needs adjusting.

PR #10504, opened 2026-08-26
docs(site): add missing assertion types to the reference tables

no-hup:
The assertion reference tables in `site/docs/configuration/expected-outputs/index.md` are missing five assertion types that are registered in `src/assertions/index.ts` and already documented elsewhere on the site:

- `finish-reason`, `word-count`, `tool-call-f1` (deterministic — each has a section in deterministic.md, and deterministic.md's own local table already lists them)
- `agent-rubric`, `search-rubric` (model-graded — each has a dedicated page)

This adds the five rows with links to the existing docs. No content changes beyond the tables.


no-hup:
Renamed the PR to `docs(site):` per the site commit convention — squash merge will pick that up as the commit subject.

no-hup:
Quick bump on this one. It's docs-only — adds the five registered assertion types (finish-reason, word-count, tool-call-f1, agent-rubric, search-rubric) that are missing from the reference tables in index.md. Each is checked against src/assertions/index.ts. Happy to rebase if it's drifted.

PR #10506, opened 2026-08-26
feat(providers): add Volcengine Ark (Doubao) provider

chyroc:
## Summary

Adds a `volcengine` provider for [Volcengine Ark](https://www.volcengine.com/product/ark) (火山引擎方舟), ByteDance's model platform that serves the Doubao (豆包) family plus hosted open-source models.

Ark exposes an OpenAI-compatible endpoint at `https://ark.cn-beijing.volces.com/api/v3`, so `VolcengineProvider` extends `OpenAiChatCompletionProvider` and follows the existing `src/providers/deepseek.ts` pattern closely — model/price table, `calculateVolcengineCost`, cached-token accounting off `usage.prompt_tokens_details.cached_tokens`, and a `createVolcengineProvider` factory.

```yaml
providers:
  - id: volcengine:doubao-seed-2-1-pro-260628
```

Auth uses `ARK_API_KEY`. Ark-specific options such as deep thinking are passed through the standard OpenAI `extra_body` mechanism:

```yaml
providers:
  - id: volcengine:doubao-seed-2-1-pro-260628
    config:
      extra_body:
        thinking:
          type: enabled
        reasoning_effort: high
```

## Pricing

Volcengine publishes prices in CNY per million tokens; the table converts at **6.737012 CNY/USD (2026-08-26)**, noted in a comment above the model list. Several Doubao models are tiered by input length (32k / 128k / 256k) — the provider uses the base tier and says so in the comment, rather than reshaping promptfoo's shared cost structure.

Sources:

- [Base URL and authentication](https://www.volcengine.com/docs/82379/1298459)
- [Model list and capabilities](https://www.volcengine.com/docs/82379/1330310)
- [Model pricing](https://www.volcengine.com/docs/82379/1544106)
- [OpenAI SDK compatibility](https://www.volcengine.com/docs/82379/1330626)

## Changes

- `src/providers/volcengine.ts` — provider implementation and model/cost table
- `src/providers/registry.ts` — register the `volcengine:` prefix
- `src/util/provider.ts`, `src/contracts/env.ts`, `src/commands/exampleAliases.ts`, `site/static/config-schema.json` — the usual registration points
- `src/app/**` — provider recognized in the eval-creator dialog and docs map
- `site/docs/providers/volcengine.md` + index entry
- `examples/provider-volcengine/`
- `test/providers/volcengine.test.ts` and additions to the existing registry/provider tests

## Testing

- `vitest run test/providers/volcengine.test.ts` — 29 passed
- `vitest run test/providers.test.ts test/providers/registry.test.ts test/providers/volcengine.test.ts` — 296 passed
- `biome check` clean on all touched files

I don't have a paid Ark account on hand, so the cost math and request shaping are covered by unit tests rather than a live API call. Happy to adjust naming, the model set, or anything else to match project conventions.


chyroc:
Pushed a correction. I'd included short aliases like `doubao-seed-2.1-pro` and `glm-5.2` alongside the versioned IDs, but Ark rejects those with `InvalidEndpointOrModel.NotFound` — only fully versioned IDs resolve (plus `doubao-seed-evolving`, which is a rolling alias). Dropped the 14 bogus entries; the remaining 15 were each called against the live API through `createVolcengineProvider`, and the cost math checks out against real `usage` payloads. 296 tests pass, biome + prettier clean.

chyroc:
Thanks, these were good. Fixed in the latest push:

- **cached tokens** — right, `response.raw` is never set and the adapter normalizes to `tokenUsage.completionDetails.cacheReadInputTokens`. That branch was dead code, so cached prompt tokens were billing at the full input rate. Now reads the normalized field.
- **nested secrets in `toJSON()`** — the registry passes the whole `ProviderOptions` as `config`, so `config.config.apiKey` survived. Now strips the wrapper keys before handing config to the base class.
- **`ARK_API_BASE_URL`** — was declared but never read. Now resolved between an explicit `apiBaseUrl` and the Beijing default.
- **prompt-level cost overrides** — merged into the recalculation so they stay authoritative.
- **`completion_tokens: 0`** — was dropping the prompt charge; now only missing/non-finite skips pricing.
- **bare `volcengine` ID** — predicate accepts it now, and unimplemented sub-types (`:embedding`, `:responses`, …) throw instead of silently hitting chat completions.
- **`VOLCENGINE_API_KEY`** — removed, the provider only reads `ARK_API_KEY`. Regenerated the schema.

Coverage on the new file is 100% lines / 97.7% branches, and I re-ran all 15 models against the live Ark API after the refactor.

One note on the edge baseline: `providers -> legacy-contracts` goes 277 → 278 for the `../types/index` import, which every provider here already makes — `deepseek.ts` has an identical import list. Happy to take a different approach if you'd rather not bump it.

chyroc:
All three were right, fixed:

- **Flash GA pricing** — docs had the pre-correction numbers, runtime table had the current ones. Both now read CNY 3.00 / 0.10 / 9.00 per the [price page](https://www.volcengine.com/docs/82379/1544106). I diffed every model's doc prices against the table while I was in there; that was the only stale one plus a rounding slip on turbo.
- **cached tokens vs cost overrides** — a user setting `cost`/`inputCost` was getting their rate on fresh tokens and the built-in rate on cached ones. Now falls through `cacheReadCost ?? inputCost ?? cost`. An explicit `cacheReadCost` still wins.
- **passthrough.model** — bills against it now instead of the provider path's model, matching what `calculateResponseCost` already does upstream.

309 tests pass, tsc/biome/prettier clean, and I re-ran all 15 models against the live API after the change.

PR #10507, opened 2026-08-26
fix(redteam): preserve cached generation and retry accounting

jameshiester-oai:
## Summary

- Preserve cached MCP attacker and red-team generation usage in the logical scan footprint while excluding it from incurred usage.
- Restore incurred cost when retry rebuilds persisted evaluation metrics, including mixed legacy and cached results.
- Cover cached plugin, strategy, extraction, goal-generation, MCP materialization, generation-event ownership, and persisted retry paths.

## Verification

- 667 focused unit tests and 36 persisted retry integration tests pass.
- An isolated local CLI evaluation completes successfully and its exported result contains the expected request accounting.
- Pinned Biome checks pass. Full local typechecking remains blocked by pre-existing Langfuse, Slack, and Anthropic dependency mismatches in the shared worktree installation.

### DIFF
--- a/src/node/retry.ts
+++ b/src/node/retry.ts
@@ -248,7 +248,8 @@ export async function recalculatePromptMetrics(evalRecord: Eval): Promise<void> 
         // Update scores and other metrics
         metrics.score += result.score ?? 0;
         metrics.totalLatencyMs += result.latencyMs || 0;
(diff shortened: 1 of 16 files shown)


alandelong-oai (approved):
Approving but might be worth addressing the p2s that codex posted

PR #10510, opened 2026-08-26
fix(mcp): forward per-server env to Claude Agent SDK stdio servers

### LINKED ISSUE #10509: Claude Agent SDK provider drops per-server `env` for stdio MCP servers
## Bug Description

## Where

```ts
} else if (config.command) {
} else if (config.path) {
}
```

```ts
    type?: 'stdio';
    command: string;
    args?: string[];
    env?: Record<string, string>;
    ...
};
```

## Repro

```yaml
providers:
    config:
      mcp:
        servers:
          - name: my-server
            command: uvx
            args: ['my-mcp-server']
            env:
```

## Impact

## Suggested Fix

```ts
} else if (config.command) {
} else if (config.path) {
  ...
}
```

## Environment

he-yufeng:
## Summary

Fixes #10509. For the `anthropic:claude-agent-sdk` provider, a per-server `env` map configured under `mcp.servers[]` (or `mcp.server`) was silently dropped for stdio-type servers: neither the `command` nor the `path` branch of `transformMCPServerConfigToClaudeCode` read it, so the spawned MCP server ran without the environment the user configured.

`MCPServerConfig` gains the `env` field, and both stdio branches forward it (the SDK's `McpStdioServerConfig` accepts `env?: Record<string, string>`). The `env` key is omitted entirely when a server has no env map, so existing configs produce byte-identical output.

## Test plan

Four new cases in `test/providers/mcp/transform.test.ts`: env forwarded for the `command` branch, env forwarded for the `path` branch, the key omitted when unset, and the single `server` entry handled like `servers`. `npx vitest run test/providers/mcp/transform.test.ts`: 24 passed. `npm run tsc` clean; biome and prettier hooks pass on the touched files.

## Notes

This PR was prepared with AI assistance (Kimi K3); all verification above was run locally.


mldangelo:
Superseded by #10557, which just landed and fixes #10509 more completely.

Your diff is correct as far as it goes, and the four transform tests were the best coverage of the three PRs racing this issue — thank you for writing them. The reason #10557 won out is scope: forwarding `env` in `transformMCPServerConfigToClaudeCode` covers the Claude Agent SDK path, but promptfoo's own `MCPClient` (`src/providers/mcp/client.ts`) spawns stdio servers for the Anthropic, OpenAI and Google MCP providers and was dropping `env` on both of its branches too. #10557 covers both, validates the field in `validateMCPConfigForClaudeCode`, and marks a non-empty `env` map as cache-sensitive so a credential-bearing server never lands in a persistent cache key.

The merged version also carries docs for the new field and, separately, #10628 tightened secret redaction so credential-named variables in an `env` map are no longer persisted in cleartext with eval results.

Sorry for the duplicated effort — three PRs arrived for this issue within four days. Please do keep sending fixes.

PR #10525, opened 2026-08-28
fix(mcp): pass per-server `env` through to stdio MCP servers

### LINKED ISSUE #10509: Claude Agent SDK provider drops per-server `env` for stdio MCP servers

sudo-ai-git:
## Pass per-server `env` through to stdio MCP servers (fixes #10509)

**The bug:** `McpStdioServerConfig` (from `@anthropic-ai/claude-agent-sdk`) supports `env?: Record<string, string>`, but `transformMCPServerConfigToClaudeCode` built the stdio config as `{type:'stdio', command, args}` and never forwarded per-server `env` — silently dropping it for both the `command` and `path` branches. (The SDK's `McpStdioServerConfig` explicitly declares `env` — verified at sdk.d.ts:1208.)

**The fix (minimal, 2 files):**
1. `types.ts` — add `env?: Record<string, string>` to `MCPServerConfig` so the field is actually configurable (before this, the config type didn't even expose `env` for stdio servers).
2. `transform.ts` — forward `env` to the Claude Agent SDK stdio config in both branches, only emitting it when present so existing configs are unchanged.

Verified: structural diff is clean (13 insertions / 2 deletions). No `mcp/__tests__` dir exists yet; a regression test asserting env is forwarded to the stdio config should be added alongside.

### DIFF
--- a/src/providers/mcp/transform.ts
+++ b/src/providers/mcp/transform.ts
@@ -218,11 +218,21 @@ async function transformMCPServerConfigToClaudeCode(
       headers: { ...(config.headers ?? {}), ...getAuthHeaders(renderedConfig, oauthToken) },
     };
   } else if (config.command) {
-    out = { type: 'stdio', command: config.command, args: config.args ?? [] };
+    out = {
+      type: 'stdio',
+      command: config.command,
+      args: config.args ?? [],
+      ...(config.env ? { env: config.env } : {}),
+    };
   } else if (config.path) {
     const isPy = config.path.endsWith('.py');
     const command = isPy ? (process.platform === 'win32' ? 'python' : 'python3') : process.execPath;
-    out = { type: 'stdio', command, args: [config.path] };
+    out = {
+      type: 'stdio',
+      command,
+      args: [config.path],
+      ...(config.env ? { env: config.env } : {}),
+    };
   } else {
     throw new Error('MCP configuration cannot be converted to Claude Agent SDK MCP server config');
   }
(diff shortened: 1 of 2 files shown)


sudo-ai-git:
Noting for reviewers: the only failing check here is `Redteam (Production API)`, which reports no output. This change is a pure 2-line config-transform (`env` pass-through in `transformMCPServerConfigToClaudeCode`); it touches no LLM/redteam/inference code path, and the same `Redteam (Production API)` job passes on other open PRs (e.g. #10520, #10515) — so this looks like a flaky infra/credential job specific to this fork run, not a regression from the change. If you can re-run it (or it clears on its own) that'd confirm. Code itself is ready for review.

mldangelo:
Superseded by #10557, which just landed and fixes #10509.

Your read of the bug was right, including the `sdk.d.ts` reference showing `McpStdioServerConfig` accepts `env`. The merged PR went further in two ways that mattered: it also fixes promptfoo's own `MCPClient` (`src/providers/mcp/client.ts`), which spawns stdio servers for the Anthropic, OpenAI and Google MCP providers and was dropping `env` there too, and it treats a non-empty `env` map as cache-sensitive so a credential-bearing server is never written into a persistent cache key.

On the regression test your description flagged as still needed — that is now in place: `test/providers/mcp/transform.test.ts` and `test/providers/mcp/client.test.ts` cover both stdio branches and the override semantics (a per-server entry wins over an inherited variable, the rest of the parent environment survives).

Sorry for the duplicated effort — three PRs arrived for this issue within four days. Please do keep sending fixes.

PR #10540, opened 2026-08-29
fix(go-provider): support importable packages

### LINKED ISSUE #10538: Go file providers cannot coexist with standard go build ./...
### Describe the bug

The experimental Go file provider requires the provider source to use `package main` because Promptfoo generates a `wrapper.go` in the same directory and builds both files together. However, the provider intentionally has no `func main()` because the generated wrapper supplies it. As a result, a valid Promptfoo provider breaks standard repository-wide Go builds.

Using a named package does not solve this: Promptfoo then fails because `wrapper.go` and the provider declare different packages. Adding `func main()` to the provider would conflict with the wrapper entry point. There appears to be no source layout that works with both Promptfoo and `go build ./...`.

### Reproduction

Create `provider.go`:

```go
package main

func CallApi(_ string, _ map[string]interface{}, _ map[string]interface{}) (map[string]interface{}, error) {
    return map[string]interface{}{"output": "ok"}, nil
}
```

Reference it from `promptfooconfig.yaml`:

```yaml
prompts:
  - test
providers:
  - id: file://provider.go
tests:
  - vars: {}
```

The Promptfoo evaluation works, but from the containing Go module:

```console
$ go build ./...
runtime.main_main·f: function main is undeclared in the main package
```

Changing the first line to `package provider` makes ordinary Go builds work, but Promptfoo fails:

```console
Error running Golang script: Command failed: go build -o /tmp/golang-provider-.../golang_wrapper wrapper.go provider.go
found packages main (wrapper.go) and provider (provider.go) in ...
```

### Expected behavior

Go providers should be usable in a normal Go module without causing `go build ./...` to fail. For example, the generated wrapper could import a named provider package, or Promptfoo could support a package/directory-based or executable Go provider that preserves standard Go build semantics.

The Go provider documentation should also describe its generated-wrapper/package requirements and current incompatibility with repository-wide builds if this limitation remains.

### Environment

- Promptfoo: 0.122.0
- Go: 1.27
- Linux

The official Go provider documentation currently shows `package main` with `CallApi` but no `main()` and labels the provider experimental: https://www.promptfoo.dev/docs/providers/go/

Elioooon:
## Summary

- detect the provider package with `go list -json`
- build named packages through a generated adapter in a separate temporary wrapper directory
- preserve existing `package main` providers and update the official example/docs to use buildable named packages

Fixes #10538.

## Validation

- `npm exec --yes --package=vitest@4.1.11 -- vitest run test/providers/golangCompletion.test.ts --sequence.shuffle=false` (29 passed)
- `go build ./...` in `examples/provider-golang`
- compiled the real `src/golang/wrapper.go` with an adapter importing the named example package
- Biome 2.5.8 check on the touched TypeScript files
- Prettier check on `site/docs/providers/go.md`
- `gofmt -d` on the touched Go files

## Risk boundary

Named providers must expose `CallApi` from an importable package in the module containing `go.mod`. Legacy `package main` behavior is unchanged. The focused provider suite and both Go build paths were verified locally; the full repository suite and TypeScript project check were left to CI.

### DIFF
--- a/examples/provider-golang/evaluation/main.go
+++ b/examples/provider-golang/evaluation/main.go
@@ -1,7 +1,7 @@
-// Package main implements a promptfoo provider that uses OpenAI's API with reasoning effort control.
+// Package evaluation implements a promptfoo provider that uses OpenAI's API with reasoning effort control.
 // It provides a CallApi function that can be used by promptfoo to generate responses
 // with configurable reasoning levels.
-package main
+package evaluation
 
 import (
 	"fmt"
--- a/examples/provider-golang/main.go
+++ b/examples/provider-golang/main.go
@@ -1,7 +1,7 @@
-// Package main implements a promptfoo provider that uses OpenAI's API.
+// Package provider implements a promptfoo provider that uses OpenAI's API.
 // It demonstrates a simple implementation of the provider interface using
 // shared code from the core and pkg1 packages.
-package main
+package provider
 
 import (
 	"fmt"
--- a/site/docs/providers/go.md
+++ b/site/docs/providers/go.md
@@ -50,8 +50,8 @@ providers:
 Here's a complete example using the OpenAI API:
 
 ```go
-// Package main implements a promptfoo provider that uses OpenAI's API.
-package main
+// Package provider implements a promptfoo provider that uses OpenAI's API.
+package provider
 
 import (
     "fmt"
@@ -94,6 +94,12 @@ func CallApi(prompt string, options map[string]interface{}, ctx map[string]inter
     }, nil
 }
 ```
+
+Use a named, importable package as shown above when the provider lives in a regular Go module.
+Promptfoo builds its generated entry point separately and imports the package through the module
+path from `go.mod`, so repository-wide commands such as `go build ./...` continue to work. The
+`CallApi` symbol must be exported. Existing `package main` providers remain supported, but a provider
+without its own `main` function cannot also be built as a standalone command.
 
 ## Using the Provider
 
--- a/src/providers/golangCompletion.ts
+++ b/src/providers/golangCompletion.ts
@@ -129,22 +129,39 @@ export class GolangProvider implements ApiProvider {
 
         const relativeScriptPath = path.relative(moduleRoot, absPath);
         const scriptDir = path.dirname(path.join(tempDir, relativeScriptPath));
-
-        // Copy wrapper.go to the same directory as the script
-        const tempWrapperPath = path.join(scriptDir, 'wrapper.go');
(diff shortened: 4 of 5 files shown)


mldangelo (approved):
Nice fix — thanks for this. The generated-adapter approach is the right shape: it keeps the provider source importable, so the module stays a normal Go module, and it leaves the legacy `package main` path completely untouched.

I QA'd this fairly hard before approving. I reproduced the bug on `main`, then ran the branch end-to-end against seven real Go module layouts with the local build:

| Scenario | `main` | this PR |
| --- | --- | --- |
| Named package in a subdirectory | ❌ `found packages main (wrapper.go) and myprov (provider.go)` | ✅ |
| Named package at the module root | ❌ | ✅ |
| Named package under `internal/` | ❌ | ✅ |
| Module with a `vendor/` directory | ❌ | ✅ |
| Legacy `package main` | ✅ | ✅ (no regression) |
| Shipped example, live OpenAI calls | `go build ./...` fails | ✅ 4/4 passed |
| `go build ./...` in `examples/provider-golang` | ❌ `function main is undeclared` | ✅ exit 0, `go vet` clean |

A few details I specifically checked, since they were the parts most likely to bite:

- **The `.`-prefixed build directory works.** Go excludes dot-directories from `./...` pattern matching but still builds fine with explicit file arguments from inside one, so the generated entry point can't leak into a user's own builds.
- **`internal/` packages still import correctly.** The build directory sits at the module root, which is the parent of `internal/`, so Go's internal visibility rule is satisfied.
- **`JSON.parse` on `go list` stdout is safe.** I checked where the noisy cases go: `go: downloading …` and toolchain-switch messages are written to stderr, and genuine load failures exit non-zero before parsing. `Incomplete: true` still yields valid JSON, so a broken import falls through to `go build` and produces the real compiler error rather than a parse failure.
- **`JSON.stringify` on the import path** produces a valid Go string literal — Go's interpreted string literals accept the same `\uXXXX` escapes, so non-ASCII module paths are fine.
- **Overhead is negligible**: the extra `go list` costs about 40ms with a warm module cache.

One nice side effect worth noting: named-package providers are now compiled as a whole package, so helpers in sibling files work. I confirmed a sibling `helper.go` gets linked in. `package main` providers still compile only the single provider file and fail with `undefined: <helper>` — that asymmetry is pre-existing, not something this PR introduced.

Approving as-is. I've opened #10541 stacked on top with follow-up polish that came out of this review — renaming the generated `provider.go` to `promptfoo_adapter.go` so compile errors don't point at a file the user may also own, a CI step that builds the Go example (nothing guarded it, which is how #10538 slipped through), and fixes for two pre-existing bugs in the docs example that predate your change. None of it blocks this.

PR #10541, opened 2026-08-29
fix(go-provider): clarify build errors and fix example docs

mldangelo:
Follow-up polish on #10540, which is now merged. Rebased onto `main`, so this diff contains only the follow-up changes.

## Summary

QA of #10540 confirmed the fix is correct: I reproduced the bug on `main` and verified the fix across seven real Go modules (named package in a subdir, at module root, under `internal/`, with `vendor/`, legacy `package main`, and the shipped example against the live OpenAI API). This PR is the polish that fell out of that review.

## Provider

- Extract the build setup into `prepareBuild()` and rename the generated entry point `provider.go` → `promptfoo_adapter.go`. `provider.go` is a very common user filename, so a missing export produced `./provider.go:5:24: undefined: provider.CallApi` and sent users to line 5 of *their own* file. It now names the generated adapter.
- Use a deterministic `.promptfoo-wrapper` directory. The enclosing temp directory is already unique per call, so `mkdtemp` bought nothing and forced `mockReturnValueOnce` chaining in tests.
- Reject provider ids whose function name the Go wrapper cannot dispatch to. `wrapper.go` only resolves `CallApi`/`call_api`, so `file://provider.go:MyProvider` could never work — on `main` it fails with `Unknown function: MyProvider`, and on #10540 with an opaque `undefined: provider.CallApi`. It now fails immediately with the reason and the fix.
- Report unparseable `go list` output with a clear message instead of a raw `JSON.parse` throw.

## Docs

The "complete example" in `go.md` had two real bugs, both verified by compiling the block verbatim:

- It calls `context.Background()` but never imports `"context"` → `undefined: context`. The sample did not compile, directly above a new claim that `go build ./...` works.
- `options["config"].(map[string]interface{})["temperature"].(float64)` — the inner assertion has no comma-ok, so it panics with `interface conversion: interface {} is nil, not map[string]interface {}` whenever `options["config"]` is absent. **Not reachable through promptfoo today**: `loadApiProvider` injects `basePath` into every provider config, so `config` is always a map even when the config block is omitted — which is why `evaluation/provider.go`, whose provider entry has no `config:` block, does not panic. It is still the wrong pattern for a sample users copy, and it would panic for anyone calling `CallApi` directly. Fixed in the docs and, after Copilot flagged it, in both example files too.

Also documented: the `go.mod` requirement (`findModuleRoot` throws without one, which the docs never mentioned), the `CallApi` naming requirement, and the behavior split — named packages compile as a whole package (sibling helpers work), `package main` providers compile only the provider file (`undefined: <helper>`).

## Example

`main.go` → `provider.go` now that these files declare named packages; a non-`main` package in a file called `main.go` is confusing in an example about idiomatic layout. The README's directory tree, config snippet, and provider signature were all stale — it documented `func CallApi(prompt string, options map[string]interface{}) (string, error)`, which is neither the real signature nor what the example implements.

## CI

The `Go Tests` job only ran `go test wrapper.go wrapper_test.go` in `src/golang` — it never built `examples/provider-golang`, which is exactly how #10538 went unnoticed. Added `go build ./...` and `go vet ./...` for the example; Go is already set up in that job.

## Tests

Three fidelity problems surfaced while updating them:

- `path.basename` had no mock implementation, so it returned `undefined` and every legacy build assertion was really checking `['build', '-o', ..., 'wrapper.go', undefined]`.
- A `path.join` shortcut pinned `wrapper.go` to a fixed path, so no test could tell whether the wrapper was copied into the build directory or next to the user's source — the exact thing this feature turns on.
- `should use custom function name when specified` asserted a capability that does not exist. Mocked `execFile` made the build and run succeed, hiding that the real wrapper rejects the name. Replaced with tests for the names that actually resolve, plus one for the new error.

Added coverage for the `package main` path, a named package with no import path, and unparseable `go list` output.

## Validation

- `vitest run test/providers/golangCompletion.test.ts` — 34 passed
- `test-hygiene` + `test/examples/readme.test.ts` — 1900 passed
- End-to-end with the local build across all seven Go module layouts; shipped example 4/4 against the live OpenAI API
- `go build ./...` + `go vet ./...` clean in the example; `gofmt`, Biome, Prettier, `tsc` clean (the `claude-agent-sdk.test.ts` type error is pre-existing on `main`)


mldangelo:
@codex review

Addressed the P2 on `golangCompletion.ts:174` in 0fe3f78 — the check now runs at the top of `executeGolangScript`, ahead of the cache key computation and lookup, with a regression test asserting `cache.get` is never reached.

One correction on the reasoning, though: the reported scenario isn't reachable through config. `loadApiProvider` does set `options.id` from the provider string, and `options` is part of the cache key, so the two ids already hash differently. Printing both keys for `file://provider.go` and `file://provider.go:MyProvider` on the same script and prompt:

```
...:{"id":"file://provider.go","config":{...}}:{...}
...:{"id":"file://provider.go:MyProvider","config":{...}}:{...}
```

They differ, and `context.vars` additionally carries a per-step `__evalId`/`__evalStepId`. I ran the collision case end-to-end with caching enabled and `--max-concurrency 1`; the error is raised, not a stale hit.

The ordering was still worth fixing — a static config error shouldn't depend on cache state to surface — so the fix landed on its own merits.

I did not take the second half of the suggestion (adding the function name to the cache key). Only `CallApi` and `call_api` are accepted and both dispatch to the same symbol, so it would invalidate every existing Go provider cache entry without disambiguating anything real.

mldangelo:
Thanks — good catch on the example files. Fixed in 29f2c59.

I'd corrected this pattern in the `go.md` sample but missed that both `examples/provider-golang/provider.go` and `evaluation/provider.go` carry the identical nested single-value assertion on `options["config"]`. Both now check the outer key with comma-ok first.

Worth recording the reachability, since I overstated it in the original PR description: this does **not** panic through promptfoo today. `loadApiProvider` injects `basePath` into every provider config, so `options["config"]` is always a map even when the config block is omitted — which is why `evaluation/provider.go`, whose provider entry has no `config:` block, has been running fine. Confirmed by re-running the example end-to-end against the live API: 4/4 passing before and after.

So it's hardening rather than a live bug fix — but it's still the wrong pattern to ship in a sample users copy, and it would panic for anyone calling `CallApi` directly. I've corrected the PR description to say so.

src/providers/golangCompletion.ts
```diff
@@ -64,11 +86,65 @@ export class GolangProvider implements ApiProvider {
...
+    if (!packageInfo.Name || packageInfo.Name === 'main') {
+      return { buildDir: scriptDir, buildFiles: [WRAPPER_FILE, scriptFile] };
+    }
+
+    if (!packageInfo.ImportPath) {
+      throw new Error('Could not determine Go provider import path');
+    }
+
+    const buildDir = path.join(tempDir, WRAPPER_DIR);
+    await fs.mkdir(buildDir, { recursive: true });
```
mldangelo-oai, line 140: **[P2] Keep the wrapper directory distinct from copied module entries**

The outer `tempDir` is unique, but [the entire module has already been copied into it](https://github.com/promptfoo/promptfoo/blob/29f2c594dd12809ff68edb13658ca7061370ac55/src/providers/golangCompletion.ts#L203-L204), so this fixed path can already be occupied. I reproduced this with a valid named provider and an otherwise unused `.promptfoo-wrapper` file at the module root: a local CLI eval with `--no-cache` succeeds on base `10deb31` and fails on this head with `EEXIST`. An existing `.promptfoo-wrapper/go.mod` also makes the build pick up the wrong module and fail to import the provider; that same fixture passes on the base. Please retain `mkdtemp` for the inner directory, or choose another unused path, so copied project files cannot break provider calls. A real-filesystem regression test would catch this; the current filesystem mocks always make directory creation succeed.

jameshiester-oai (commented):
Reviewed base `10deb31cc3ac953a28ee88fe43530cfc4d2c7927` and head `f1c705d804b4e843a54b25a0bde9c4056099497a`. Scope: full base...head review.

Verdict: one verified P2 backward-compatibility regression. CI is green at this head.

Verified findings:
- P2: the generated named-package adapter now requires `provider.CallApi` to be directly assignable to the wrapper's named `ApiFunc` type. Existing providers that expose the documented underlying signature through their own named function type no longer compile; see the inline comment.

Existing feedback: the prior cache-ordering concern and example config assertion concern are fixed at this head; I did not duplicate them.

Simplification/test opportunities: a credential-free CLI smoke fixture for a named Go package would exercise adapter generation and invocation beyond the fully mocked unit test.

Validation performed:
- `git diff --check 10deb31cc3ac953a28ee88fe43530cfc4d2c7927...f1c705d804b4e843a54b25a0bde9c4056099497a` passed.
- `GOCACHE=/private/tmp/promptfoo-go-build GOMODCACHE=/private/tmp/promptfoo-go-mod go build ./... && go vet ./...` in `examples/provider-golang` passed.
- Minimal isolated Go repro: old inferred alias compiles; current `var CallApi ApiFunc = provider.CallApi` fails with `cannot use provider.CallApi ... as ApiFunc`; `ApiFunc(provider.CallApi)` compiles while retaining signature checking.
- Attempted `npx vitest run test/providers/golangCompletion.test.ts --sequence.shuffle=false`; it could not start because this fresh worktree lacked Vitest. `npm ci` then hit repeated tar extraction/corruption errors, so no local Vitest result is claimed.

Limitations: PR Review Toolkit was unavailable on this host (`/pr-review-toolkit:review-pr all parallel` returned no such file); I used the manual correctness/tests/docs/error-handling/type/simplification fallback plus three read-only specialist passes. I did not run a full local CLI eval because the dependency tree remained unusable.

src/providers/golangCompletion.ts
```diff
@@ -64,11 +91,75 @@ export class GolangProvider implements ApiProvider {
...
+    await fs.writeFile(
+      path.join(buildDir, ADAPTER_FILE),
+      `package main\n\nimport provider ${JSON.stringify(packageInfo.ImportPath)}\n\nvar CallApi ApiFunc = provider.CallApi\n`,
```
jameshiester-oai, line 143: This typed assignment narrows compatibility compared with the previous inferred alias. A named-package provider can expose the documented underlying signature through its own named function type, for example `type ProviderFunc func(...); var CallApi ProviderFunc`. The old `var CallApi = provider.CallApi` adapter compiled that provider, but this line fails with `cannot use provider.CallApi ... as ApiFunc value in variable declaration`, so an existing valid provider stops running after upgrade. I reproduced that with a minimal Go module. Converting explicitly (`var CallApi ApiFunc = ApiFunc(provider.CallApi)`) preserves the signature check and accepts compatible named function types.

PR #10545, opened 2026-08-29
perf(scheduler): replace array queue with O(1) indexed deque in SlotQ…

Doves788:
-Description

This PR improves the performance of `SlotQueue` when handling a large number of queued requests.

The previous implementation used an array, where operations such as removing a timed-out request or taking the next item from the queue could require shifting or searching through elements, resulting in O(N) work. To avoid this overhead, the queue now uses a custom `IndexedDeque`, allowing these operations to be handled in O(1).

-Key Changes

* Added a new `IndexedDeque` implementation that combines a doubly-linked list with a `Map` for efficient lookups and removals.
* Updated timeout cancellation in `acquire()` to remove queued requests directly instead of using `findIndex` and `splice`.
* Updated `processQueue()` to remove items from the front of the queue without the overhead of `Array.shift()`.
* Updated `dispose()` to clean up queued items using the new `drain()` method.
* Added unit tests covering FIFO behavior, removing items from the middle and tail, and other edge cases.

- Files Modified

* `src/scheduler/slotQueue.ts`
* `src/scheduler/index.ts`
* `test/scheduler/indexedDeque.test.ts`

### DIFF
--- a/src/scheduler/index.ts
+++ b/src/scheduler/index.ts
@@ -37,7 +37,7 @@ export {
   shouldRetry,
 } from './retryPolicy';
 // Slot queue
-export { SlotQueue } from './slotQueue';
+export { IndexedDeque, SlotQueue } from './slotQueue';
 
 export type { ConcurrencyChangeResult } from './adaptiveConcurrency';
 export type { ProviderMetrics } from './providerRateLimitState';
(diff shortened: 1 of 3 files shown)


mldangelo-oai:
Thank you! Appreciate the contribution but I do not see the performance improvements justifying the complexity. Going to close it for now but feel free to chime in / disagree.

PR #10547, opened 2026-08-29
perf(scheduler): add latency-gradient AIMD congestion control to AdaptiveConcurrency

Doves788:
-- Summary

This PR adds latency-aware congestion control to `AdaptiveConcurrency`.

Currently, concurrency is mainly reduced after receiving a `429 Too Many Requests` response. By the time the first rate-limit response is received, several requests may already be in flight, which can result in multiple requests failing at the same time.

This change uses request latency as an earlier signal of upstream congestion. Request latency is tracked using an Exponential Moving Average (EMA) and compared against the lowest observed latency. When latency increases significantly, the scheduler can reduce concurrency proactively instead of waiting for rate limiting to happen.

-- Changes

- Added EMA-based request latency tracking to `AdaptiveConcurrency`.
- Added configurable EMA smoothing through `AdaptiveConcurrencyOptions`.
- Added latency-gradient detection by comparing the current EMA latency with the minimum observed latency.
- Proactively reduces concurrency by 20% when latency exceeds 1.5x the baseline.
- Uses `reason: 'proactive'` when concurrency is reduced due to increased latency.
- Keeps the existing behavior unchanged when latency information is not provided.
- Updated `ProviderRateLimitState.executeWithRetry()` to pass the measured `latencyMs` to `handleSuccess()`.
- Added tests covering latency tracking, proactive throttling, and the existing recovery behavior.

-- Files Modified

- `src/scheduler/adaptiveConcurrency.ts`
  - Added latency tracking and EMA calculation.
  - Added latency-gradient based congestion detection.
  - Added proactive concurrency reduction.

- `src/scheduler/providerRateLimitState.ts`
  - Passes measured request latency to `handleSuccess()`.

- `test/scheduler/adaptiveConcurrency.test.ts`
  - Added tests for EMA tracking.
  - Added tests for proactive concurrency reduction.
  - Added tests to verify existing recovery behavior remains unchanged.

-- Test Plan

- [x] Unit tests for EMA latency tracking with configurable alpha.
- [x] Unit tests for proactive concurrency reduction when the latency gradient exceeds the threshold.
- [x] Verified the existing recovery behavior is unchanged when latency is omitted.
- [x] Verified normal recovery behavior is unchanged when latency remains within the threshold.
- [x] All existing `AdaptiveConcurrency` tests pass without regression.


Doves788 (commented):
All automated review suggestions and formatting checks have been addressed:

1. **Congestion Wave Cooldown:** Added a cooldown epoch (`cooldownUntilSuccessCount`) to prevent in-flight request bursts from causing cascading multi-step reductions in a single congestion wave.
2. **Baseline Minimum Window Decay:** Implemented sliding-window baseline tracking (`baselineWindowSize = 20`) so the baseline dynamically recalibrates when workload latency shifts.
3. **Slot Release Order:** Adjusted `ProviderRateLimitState.executeWithRetry` to apply concurrency modifications before `slotQueue.release()`.
4. **Option Validation:** Clamped `latencyBackoffFactor`, `alpha`, and `gradientThreshold` to safe operating bounds and capped concurrency to `initial`.
5. **Duplicate Request IDs:** `IndexedDeque` now indexes duplicate IDs via `Map<string, Set<DequeNode<T>>>` with strict FIFO removal.

All 31 CI checks (Node 22/24/26 on Ubuntu/macOS/Windows, Biome lint, and Style checks) are passing and green! 🚀

PR #10556, opened 2026-08-30
feat(providers)!: default Google chat models to the Interactions API

mldangelo:
## Summary

The [Interactions API](https://ai.google.dev/gemini-api/docs/interactions-overview) went GA in June 2026 and is Google's primary interface for Gemini models and agents; `generateContent` is the legacy path, and Google ships new models, modalities, tools, and agentic features on Interactions first.

Promptfoo previously spoke Interactions only for Gemini Omni **video generation** — the provider hardcoded a video `response_format`, rejected every tool, and routing was gated on one model name. This PR adds a full chat/multimodal Interactions provider **and makes it the default for `google:` chat models**.

```yaml
providers:
  # Now uses the Interactions API. No config change needed.
  - id: google:gemini-3.6-flash
```

## The default is capability-aware

Interactions is not a strict superset of `generateContent`, so the flip is gated rather than unconditional. `shouldUseInteractions` falls back to the legacy transport when a request needs something Interactions cannot serve, so making it the default cannot silently change behavior:

| Condition                                                    | Why                                                                    |
| ------------------------------------------------------------ | ---------------------------------------------------------------------- |
| `safetySettings` configured                                    | The Gemini API rejects `safety_settings` on this endpoint.             |
| `-tts` model, or an `AUDIO`/`IMAGE` response modality           | Interactions returns text for these rather than audio or image output. |
| Legacy `palm:` prefix, or a PaLM model like `chat-bison`        | Predates Interactions entirely.                                        |
| Script-like id (`google:custom-model.ts`)                       | A pseudo model name, not something Interactions serves.                |
| Any `vertex:` provider                                          | Much narrower surface — see below. Stays opt-in.                       |

`interactions: true|false` overrides in both directions; the `google:interactions:` / `vertex:interactions:` prefixes still force it while changing the provider id.

I verified against the live API that **every documented Gemini chat model is served on Interactions** — 3.7/3.6/3.5 flash, 3.5-flash-lite, 3.1-flash-lite, 3.1-pro-preview, 3-flash-preview, 2.5 pro/flash/flash-lite, flash-latest, gemma-4-31b-it, robotics-er-1.6. Only TTS models are not.

## Server-side retention

Google stores interactions by default (55-day paid retention). That is the wrong default for eval and red-team payloads, so **on AI Studio promptfoo sends `store: false`** and resolves the tool loop by resending the timeline inline — verified to work statelessly, so the privacy default costs no functionality. `store: true` opts into server-side history; combining it with `store: false`, or smuggling either through `passthrough`, is rejected up front.

## Vertex AI is narrower (stays opt-in)

Verified live with ADC:

| Behavior     | Vertex AI                                                                                                             |
| ------------ | ---------------------------------------------------------------------------------------------------------------------- |
| Models       | `gemini-3-flash-preview` works; `gemini-3.6-flash` / `gemini-2.5-flash` return `Unsupported model interaction`.        |
| Locations    | Only `global`, `us`, `eu`, all on `aiplatform.googleapis.com`. No `us-`/`eu-aiplatform` host (both 404).               |
| `store`      | Must be `true` — Vertex rejects `store: false`, so interactions **are** retained there.                                |
| Server state | `previous_interaction_id` returns HTTP 200 but silently ignores the stored history, so it is rejected and the tool loop resends inline. |
| Auth         | OAuth only; API keys refused.                                                                                          |

The silent history drop is the one worth a close look: unhandled, it produces quietly wrong multi-turn answers rather than an error.

## Bugs found and fixed

- **`safetySettings` 400'd on the existing Omni video path.** The endpoint rejects the parameter, so any user who set it got a hard failure. Now dropped with a warning on both paths.
- **A disabled tool policy was ignored.** `tool_choice: none` still sent declarations and callbacks could execute a disabled tool. Interactions has no `tool_choice`, so declarations are withheld.
- **`passthrough` bypassed the retention guards**, enabling storage while `metadata.interactionStored` reported `false`.
- **`generationConfig.response_schema` was dropped**, silently downgrading structured output to free text.
- **`service_tier` was never sent** despite cost being computed against it.
- **`responseSchema` was passed through unparsed** (it is typed as a string).
- **A polled interaction replays its full timeline**, which would re-execute an already-run tool.
- **`vertex:embedding:*` with `interactions: true`** built a chat provider named `embedding:<model>`.
- `config.timeoutMs` applied only to polling, not the initial POST.

Caching now matches the other Google providers: responses are cacheable, keyed on a hash of the credentials, so `--no-cache` is meaningful and repeat evals stop re-billing. Grounding detail the timeline already carried is surfaced as `metadata.webSearchQueries` (matching `generateContent`) and `metadata.serverToolSteps`.

## Verification

The docs are wrong in two places, caught by testing against the live API: the migration guide's `v1beta2/interactions` **404s** (`v1beta` is correct), and `response_format` rejects a `json_schema` wrapper.

Transport selection verified end to end — three bare `google:` providers all routed through Interactions while `interactions: false` used the legacy path, all passing:

```
provider                rows  via Interactions  passed
i-3.6-flash                4                 4       4
i-2.5-flash                4                 4       4
i-3.5-lite                 4                 4       4
legacy-3.6-flash           4                 0       4
```

Focused capability evals (all live, `--no-cache`): grounding asserts on `metadata.webSearchQueries`; structured output asserts parsed JSON; code execution asserts `1517` plus `serverToolSteps`; function calling asserts **`is-valid-function-call`**, proving the output shape stays compatible with existing assertions. All five example configs pass (6, 2, 1, 2, 1 tests, zero errors).

**Tests:** 1,047 across `test/providers/google/` + `registry.test.ts`; 8,581 across `test/providers/`. A full `npx vitest run` shows no failures outside `test/examples/readme.test.ts`, whose failures all come from stray `node_modules` in example directories and are pre-existing.

## Reviewer notes

- **Breaking-ish:** `google:` chat models change transport. Token accounting shifts (Interactions reports reasoning tokens separately) and responses no longer carry `guardrails`, since Interactions returns no `safetyRatings`. Both are documented; `interactions: false` restores the old path.
- **MCP tools** are advertised but not auto-executed — that matches every other Google provider (`mcpClient.callTool` is never called in `src/providers/google/`), so it is pre-existing, and documented as such.
- `architecture/edge-baseline.json` bumps are small and committed with their justification.

## Test plan

- [x] `npx vitest run` — no new failures
- [x] `npx tsc --noEmit`, Biome, Prettier, `npx knip`, `npm run architecture:check`
- [x] Live evals on AI Studio (`gemini-3.6-flash`, `gemini-2.5-flash`, `gemini-3.5-flash-lite`) and Vertex (`gemini-3-flash-preview`)


jameshiester-oai (commented):
Reviewed base `90fa399b941364363f57288fbf305b6d6aaff7ed` and head `79a5627ade0fd639f3bafdf1617c1dda4e5afde5`. Scope: full base...head review.

Verdict: no new inline comments; multiple existing current-head P1/P2 findings remain applicable. CI is green at this head.

Verified findings:
- No new findings beyond existing current-head feedback.

Important existing unresolved feedback:
- P1: transport selection in `src/providers/families/google.ts` sees only provider config, so prompt/test-level `safetySettings` selects Interactions and is then dropped. I independently traced and confirmed this current-head path.
- P1: object-form named `tool_choice` is omitted from `getUnexpressibleToolMode`; default routing can silently weaken required-call semantics to an allow-list. I independently confirmed the string-only check and later filtering behavior.
- Other current-head threads worth resolving include Gemma default routing, schema type lowercasing inside instance-valued keywords, multimodal part preservation, and pending tool-loop/cost edge cases. I did not duplicate existing threads.

Simplification/test opportunities: add routing tests that include prompt-level config and object-form named tool choices, plus credential-free wire-level fixtures for the default/fallback contract.

Validation performed:
- `git diff --check 90fa399b941364363f57288fbf305b6d6aaff7ed...79a5627ade0fd639f3bafdf1617c1dda4e5afde5` passed.
- Static trace confirmed provider-level routing at `families/google.ts:159`, later prompt config merge at `interactionsChat.ts:784`, safety drop at `interactionsChat.ts:795`, and string-only required-mode detection at `interactionsShared.ts:293`.
- Attempted focused Vitest for interactions and registry tests; it could not start because this fresh worktree could not resolve `vitest/config`. `npm ci` had already hit repeated tar extraction/corruption errors, so no local Vitest or live API result is claimed.

Limitations: PR Review Toolkit was unavailable on this host (`/pr-review-toolkit:review-pr all parallel` returned no such file); I used the manual six-lens fallback plus three read-only specialist passes. I did not independently repeat credentialed live Google/Vertex E2E.

PR #10562, opened 2026-08-30
fix(eval): keep the CLI alive while watch mode is watching

mldangelo:
Fixes the second half of #10164 — reported there as "on macOS the watcher never fires `ready` and no re-runs happen at all."

## The bug

`promptfoo eval --watch` sets up a chokidar watcher and then returns. `main()` resolves as soon as the action handler returns, and its `finally` runs `shutdownGracefully()` unconditionally, which:

- closes the database (`closeDbIfOpen()`)
- destroys the global HTTP dispatcher (`dispatcher.destroy()`)
- closes the logger's file transports
- then schedules `process.exit(process.exitCode || 0)` 100ms later

So the CLI tore down everything a re-run needs, and killed the process before chokidar could report a single change. Watch mode has been a no-op end to end.

## It isn't macOS-specific, and it isn't chokidar

A watcher created with `persistent: true` reaches `ready` in about a millisecond and holds the event loop open by itself:

```
ready at 1ms
STILL ALIVE at 4s -> chokidar holds the loop
```

The watcher was fine. It was being shot 100ms after it started.

## The fix

The same approach the codebase already uses for long-running commands: don't resolve until the work is actually finished. `startServer()` returns a promise that settles only when the server closes — that's why `promptfoo view` survives this identical shutdown path. Watch mode now does the same, settling once SIGINT or SIGTERM closes the watcher.

Signal handlers are removed as soon as one fires, so a second Ctrl-C falls back to Node's default behaviour and terminates immediately.

**Only the CLI blocks.** `isCliEventSource()` already marks the single caller that owns process lifecycle — its own doc comment names SIGINT handlers as the reason it exists — so library, MCP, and web embedders keep today's behaviour and decide for themselves when to stop watching. `redteam run` has no `--watch` flag, so it is unaffected.

## Verification

Real CLI, not just mocks:

```
$ promptfoo eval -c promptfooconfig.yaml --watch
...
Watching for file changes on .../promptfooconfig.yaml ...   # never printed before

# edit the watched config
File change detected: .../promptfooconfig.yaml
# second evaluation runs

# Ctrl-C
Stopping watch mode...
EXIT CODE: 0
```

Exit code 0 on SIGINT matches the convention `shutdownGracefully()` already applies to every other command, including `view`.

Three tests added under `watch mode process lifecycle`. The two that pin the new behaviour fail on `main`:

```
× does not resolve while the CLI is watching, and resolves on SIGINT
× removes its signal handlers so a second Ctrl-C terminates
```

The third pins the preserved library behaviour and passes both ways.

## Relationship to #10169

Independent, and both are needed. #10169 fixes *which files* watch mode watches (scalar and generator `tests:` references were skipped); this fixes watch mode *running at all*. Either can land first — #10169's fix simply isn't observable through the CLI until this one lands.

## Not addressed

`ignored: /^\./` in the watcher options is matched against absolute paths, which never begin with a dot, so it is effectively a no-op. Left alone here: it is pre-existing, orthogonal to the lifecycle bug, and changing what gets ignored deserves its own change.

### DIFF
--- a/src/node/doEval.ts
+++ b/src/node/doEval.ts
@@ -62,7 +62,6 @@ import { accumulateTokenUsage, createEmptyTokenUsage } from '../util/tokenUsageU
 import { isUuid } from '../util/uuid';
 import { deleteErrorResults, getErrorResultIds, recalculatePromptMetrics } from './retry';
 import { notCloudEnabledShareInstructions } from './shareInstructions';
-import type { FSWatcher } from 'chokidar';
 import type { Command } from 'commander';
 
 import type {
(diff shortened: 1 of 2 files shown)


mldangelo:
Verified the full #10164 scenario end to end, with this PR's branch merged with #10169 (they merge cleanly in `src/`; the only conflict is adjacent test edits). Real CLI, real watcher, no mocks:

```
tests: file://cases.yaml     # the exact shape the issue reports as ignored
```

```
Watching for file changes on ./promptfooconfig.yaml ...
Watching for file changes on ./cases.yaml ...        # <- #10169
>>> editing cases.yaml
File change detected: ./cases.yaml                   # <- this PR
evaluations run: 2
   1 Say: FIRST
   1 Say: SECOND                                     # rerun picked up the edit
Stopping watch mode...
>>> EXIT CODE: 0
```

Both halves of the issue are needed for that sequence: without #10169 `cases.yaml` is never in the watch list, and without this PR nothing runs at all because the process is gone 100ms after startup.

PR #10566, opened 2026-08-31
fix(http): support Windows drive letter and file:// source paths in multipart

### LINKED ISSUE #10337: [Bug] HTTP provider multipart `file://` source path fails on Windows
**Describe the bug**

On Windows, when an HTTP provider's multipart file source uses a relative `file://` path, promptfoo converts it into a UNC path and fails to read the file. The relative path `<config-dir>/fixtures/sample.pdf` becomes `\\.\fixtures\sample.pdf` and throws `ENOENT`.

The `./` segment is parsed by `fileURLToPath()` as a URL hostname, producing the Windows device/UNC path `\\.\fixtures\...`. This only happens on Windows; on Linux/macOS the relative `file://` path resolves correctly (per the docs example `file://fixtures/sample-report45.pdf`).

**To Reproduce**

1. Create this directory layout:

```
.
├── fixtures/
│   └── sample.pdf        # any small PDF
└── promptfooconfig.yaml
```

2. `promptfooconfig.yaml`:

```yaml
providers:
  - id: http
    label: repro
    config:
      url: 'https://example.com/upload'
      method: 'POST'
      multipart:
        parts:
          - kind: file
            name: file
            filename: 'sample.pdf'
            contentType: application/pdf
            source:
              type: path
              path: 'file://{{file_path}}'
      transformResponse: json.data

tests:
  - vars:
      file_path: './fixtures/sample.pdf'
```

3. Run `promptfoo eval`.

4. See error:

```
Error: ENOENT: no such file or directory, open '\.\fixtures\sample.pdf'
```

(Note the leading `\.\` — a UNC/device path.)

**Expected behavior**

The file at `<config-dir>/fixtures/sample.pdf` should be resolved and uploaded. Per the docs, relative paths under `file://` should resolve from the promptfoo config directory.

**Screenshots**

N/A.

**System information:**

- Promptfoo version: 0.122.0
- OS: Windows (Node 24)

The relevant code is `resolvePath()` / `normalizeFilePath()` in `src/providers/http.ts`, which calls `fileURLToPath()` on any path prefixed with `file://`. On Windows, `fileURLToPath('file://./fixtures/sample.pdf')` returns `\\.\fixtures\sample.pdf` (hostname `.`), which is never a valid local path. Removing the `./` prefix (`file://fixtures/sample.pdf`) yields `\\fixtures\sample.pdf` (hostname `fixtures`) — equally broken. Only absolute `file:///C:/...` URLs convert correctly.

**Additional context**

Workaround: use a plain relative path without the `file://` scheme, e.g. `path: './fixtures/sample.pdf'`, which promptfoo resolves correctly from the config directory.

AbhiPra24:
## Description

Fixes #10337

On Windows, resolving `file://` multipart source paths encountered issues when using Node's `url.fileURLToPath` directly:
1. Promptfoo's relative file shorthand (e.g. `file://relative/path.ext`) treated `relative` as a host, converting it to UNC path `\\relative\path.ext` on Windows instead of resolving relative to the base directory.
2. Windows drive letters and various file URL formats (e.g. `file://C:/...`, `file:///C:/...`, `file://C:\...`, `file:///C:\...`, `file://localhost/C:/...`) need proper drive letter extraction, URL decoding, and path normalization across Windows and POSIX environments.

This PR updates `normalizeFilePath` in `src/providers/httpMultipart.ts` to handle:
- Windows drive letter paths across 2-slash (`file://C:/...`), 3-slash (`file:///C:/...`), backslash, and `localhost` forms with `path.normalize` / `fileURLToPath`.
- URL-encoded paths (`%20`).
- Promptfoo's relative shorthand `file://` paths without UNC conversion on Windows.
- Cross-platform safe `pathToFileURL` usage in tests and added unit tests covering each path scenario.

## Test Plan
`npm test test/providers/http-multipart.test.ts`


mldangelo:
Thanks for this — the underlying bug is real and your diagnosis was right. I've pushed to your branch; happy to revert anything you disagree with. Also merged current `main` in.

I confirmed the bug you're fixing, using Node's win32 emulation (`fileURLToPath(url, { windows: true })`):

| Windows behavior | before | after |
|---|---|---|
| `file://fixtures/report.txt` | `\\fixtures\report.txt` ✗ | `fixtures/report.txt` ✓ |
| `file:///C:/My%20Documents/doc.pdf` | `C:\My Documents\doc.pdf` | unchanged ✓ |

The relative shorthand really was being mangled into a bogus UNC path. That's #10337.

### The failing Windows shard was a real defect, not a flake

```
expected '/home/user/my%20file.pdf' to be '/home/user/my file.pdf'
```

On Windows, `fileURLToPath('file:///home/user/my%20file.pdf')` throws `ERR_INVALID_FILE_URL_PATH` — the path has no drive letter — and the `catch` returned the raw slice, so `%20` survived into the filename. **Your own test caught it**, which is why it was worth keeping rather than muting.

Decoding in that fallback also makes the assertion identical on both platforms: POSIX gets there via `fileURLToPath`, Windows via the decoded fallback.

### On the UNC note

Codex is right that `file://server/share/x` changes behavior, but its suggested fix — "distinguish valid UNC authorities from the shorthand" — isn't achievable. `file://host/share` and `file://relative/path` are the same string shape, so no rule separates them and supporting one necessarily breaks the other.

So I pinned the tradeoff rather than guessing at it: the shorthand wins (documented, far more common), and a UNC share stays reachable by dropping the scheme. `\\server\share\report.pdf` returns untouched from the first line and is already absolute under win32 rules, so `resolvePath()` won't prepend `basePath`. There's a test in both directions.

### Other changes

- **Dropped the `process.platform === 'win32'` branch** from the drive-letter path. `path.normalize()` already returns exactly what `fileURLToPath()` does for these on Windows, so the branch only made the behavior untestable off Windows.
- **Left the shorthand un-decoded**, deliberately, unlike the URL-shaped branches. It's a path with a scheme prefix, not a real URL, so a literal `%` has to survive — otherwise `report%20final.pdf` becomes a lookup for a different file. Pinned by a test.
- **De-tautologized the drive-letter assertions.** They compared against `path.normalize(...)`, which is what the implementation calls, so they restated the code and couldn't fail off Windows. They now assert literal per-platform strings.

### Verification

Because the interesting behavior is Windows-only and CI is the sole real gate, I mirrored the new implementation under `path.win32` + `fileURLToPath(..., { windows: true })` and checked all 17 expectations — all hold. That caught one bad expectation of my own before pushing: `resolvePath('file:///tmp/report.txt')` returns `/tmp/report.txt` unchanged on Windows, because win32 treats a leading `/` as absolute.

Worth being straight about one limit: the decode fallback can't be exercised on POSIX at all, since `fileURLToPath` succeeds there. The Windows shard is its only real gate — the same shard that caught the original bug, so a regression would surface the same way.

20 multipart tests and 423 across `test/providers/http*` pass; `tsc` and `biome ci` clean.

mldangelo:
Merging this — thank you, and congratulations on your first promptfoo contribution! 🎉

You diagnosed #10337 correctly and precisely: promptfoo's `file://relative/path` shorthand was being read as a UNC authority on Windows and mangled into `\\relative\path`. Getting from that symptom to the right cause in `fileURLToPath` is the hard part, and you did it.

The part I want to call out: **your own test caught a real bug in your implementation.** The Windows shard failure wasn't noise — on Windows `fileURLToPath` throws for a driveless path, and the fallback was returning an un-decoded `%20`. A weaker test would have hidden that. Writing a test that fails on a platform you can't run locally is genuinely good work, and it's why the fix landed correct rather than half-correct.

I pushed three follow-ups: decoding in that fallback, pinning the shorthand-vs-UNC precedence with a test (the two forms are the same string shape, so one has to win — and UNC stays reachable by dropping the scheme), and dropping the `process.platform` branch since `path.normalize()` already matches `fileURLToPath` there.

All three Windows shards are green. Thanks for the fix and for the patience — hope to see more from you.

PR #10567, opened 2026-08-31
fix(assertions): set default threshold for answer-relevance to 0.7

### LINKED ISSUE #9848: answer-relevance defaults to threshold 0, so it silently always passes when no threshold is set

AbhiPra24:
## Description

Previously, `handleAnswerRelevance` defaulted undefined assertion thresholds to `0`. Because `answer-relevance` scores range from 0 to 1, a threshold of 0 caused tests without an explicit threshold to trivially pass regardless of actual relevancy.

This change sets the default threshold to `0.7` (aligning with sensible defaults across promptfoo's model-graded assertions and documented examples) when no threshold is specified, and updates the unit tests accordingly.

Fixes #9848

### DIFF
--- a/src/assertions/answerRelevance.ts
+++ b/src/assertions/answerRelevance.ts
@@ -22,7 +22,7 @@ export const handleAnswerRelevance = async ({
     ...(await matchesAnswerRelevance(
       input,
       output,
-      assertion.threshold ?? 0,
+      assertion.threshold ?? 0.7,
       test.options,
       providerCallContext,
     )),
--- a/test/assertions/answerRelevance.test.ts
+++ b/test/assertions/answerRelevance.test.ts
@@ -162,7 +162,7 @@ describe('handleAnswerRelevance', () => {
     ).rejects.toThrow('answer-relevance assertion type must have a prompt');
   });
 
-  it('should use default threshold of 0 if not specified', async () => {
+  it('should use default threshold of 0.7 if not specified', async () => {
     const mockMatchesAnswerRelevance = vi.mocked(matchesAnswerRelevance);
     mockMatchesAnswerRelevance.mockResolvedValue({
       pass: true,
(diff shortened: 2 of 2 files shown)


mldangelo:
Thanks for this, and sorry it sat so long.

It turned out that **five** of us independently found and fixed the same bug — this PR, #10202, #10494, #9907 and #9911 all made the RAG model-graded assertions stop defaulting `threshold` to `0`, where they silently passed for any score. That is a strong signal the bug was real; it also meant we had to pick one to land.

We went with #10202 because it was the only one covering all four RAG assertions and sharing the default through a named constant (`DEFAULT_RAG_ASSERTION_THRESHOLD`) rather than an inline literal per file. It has now merged.

On the value: the five PRs split between `0.5` and `0.7`, and we went with **0.5**, matching the existing `DEFAULT_THRESHOLD = 0.5` in `src/external/assertions/deepeval.ts`. The `0.7` in `answer-relevance.md` is a doc *example*, and the other pages' examples disagree with it (`0.3` on context-relevance, `1.0` on context-recall), so we treated the in-repo constant as the better precedent. A lower default also breaks fewer existing evals, which matters since this is a behavior change.

**You are credited as a co-author on that commit** (`4831b602d0`), so it counts as your contribution.

I checked whether anything here could land on top, and after rebasing there was nothing left — #10202 edits the same two lines of the same test and adds cases around them. Closing as merged-via-#10202.

Thanks for catching it, and please keep the fixes coming.

mldangelo:
Closing as merged via #10202 (`4831b602d0`), where you are credited as a co-author. Thanks again — see the note above for the details.

PR #10572, opened 2026-08-31
docs(site): clarify JavaScript assertion output types

motodriver:
JavaScript expressions and external scripts preserve the transformed output type, while inline function assertions receive strings. Clarify that JSON text needs explicit parsing and simplify the async validation example to one exported function, removing redundant wrappers, a rethrow, and a broken log statement.

## Validation

- 96 tests passed in `test/assertions/javascript.test.ts` and `test/assertions/index.test.ts`.
- Nine local fixture evals passed: objects, arrays, and JSON text across expression, external-script, and function assertions, with provider/test/assertion transforms applied. Exported JSON confirmed every case passed with score 1 and no errors.
- Six checks passed against the extracted async example: boolean, numeric, and grading-result responses, plus fetch and JSON errors. Fetch was mocked.
- `cd site && SKIP_OG_GENERATION=true npm run build` passed; checked rendered text and linked section anchors.
- `npm run l && npm run f`, Prettier, and `git diff --check` passed.


mldangelo-oai (approved):
Thanks for catching this and working through the clarification! The final wording matches the assertion behavior, including transforms, and the simpler async example is easier to follow. Verified with focused tests, local evals, and a docs build. LGTM once CI finishes.

PR #10573, opened 2026-08-31
feat(assertions): add character count assertion

motodriver:
## Summary
- add `character-count` and `not-character-count` assertions
- count Unicode code points so astral characters such as emoji count once
- support exact, minimum, maximum, and range values with validation
- integrate the assertion with config types, generated schema, Eval Creator validation, and documentation

## Motivation
`word-count` splits on whitespace, so a continuous Chinese sentence is counted as one word. A separate character-count assertion supports Chinese ad copy, SMS, push notifications, and other length-constrained outputs without changing existing word-count behavior.

## Verification
- `npx vitest run test/assertions/characterCount.test.ts test/assertions/runAssertion.test.ts --sequence.shuffle=false` — 200 passed
- focused backend suite — 589 passed
- focused Eval Creator suite — 189 passed
- config schema regenerated successfully
- Biome and Prettier checks passed

## Counting semantics
Counts Unicode code points via `Array.from(outputString)`. This is intentionally not grapheme-cluster or platform-specific weighted counting.


motodriver:
@codex review

PR #10574, opened 2026-08-31
test(eval): stop a watch-mode test stranding doEval past its own test

mldangelo:
Follow-up to #10562. Hardens a test-only leak; no production code changes.

## The defect

The `watch mode process lifecycle` block's `afterEach` only *unregistered* the SIGINT listener it had leaked:

```js
for (const listener of installedSince(before)) {
  process.removeListener('SIGINT', listener);
}
```

Watch mode returns **solely** on a signal (that's the whole point of #10562), so unregistering the listener strands `doEval` forever. A stranded run keeps executing, and its mock calls land inside whichever test runs next. A test that fails before reaching its own signal hits this every time.

Now the `afterEach` signals the leaked listeners, awaits every `doEval` the block started, and only then unregisters.

## Why

`main` went red once after #10562 merged:

```
FAIL test/commands/eval.test.ts > watch paths for tests > does not re-read an executable config
AssertionError: expected "vi.fn()" to not be called at all, but actually been called 9 times
```

That test expects `maybeReadConfig` to be called zero times, and every mock is reset in the outer `beforeEach` — so a non-zero count requires a `doEval` running *outside* its own test. The leak above is the only mechanism I found that produces that.

## Honest limits

**This is not a confirmed root cause.** Re-running the job passed, so the failure is non-deterministic. I could not reproduce it locally in file order, across 8 shuffled seeds, or in three full-suite runs. I also injected a test that bails before signalling, and it passed both with and without this change.

So: a stranded `doEval` is a real defect worth fixing on its own merits, and it is the only mechanism consistent with the observed failure — but I'm not claiming it is proven to be the one that fired.

The blocks are ordered `watch paths for tests` (#10169) then `watch mode process lifecycle` (#10562) in file order, which is why local runs pass; CI shuffles.

## Verification

`tsc` clean, `biome ci` clean, 128 tests pass across five shuffle seeds.

### DIFF
--- a/test/commands/eval.test.ts
+++ b/test/commands/eval.test.ts
@@ -795,6 +795,9 @@ describe('evalCommand', () => {
         });
       });
 
+    /** Every doEval this block starts, so afterEach can guarantee none outlives its test. */
+    let started: Promise<unknown>[] = [];
+
     const startWatch = (evaluateOptions: Record<string, unknown>) => {
       const config = { prompts: [], providers: [], tests: [] } as UnifiedConfig;
       vi.mocked(resolveConfigs).mockResolvedValue({
@@ -811,12 +814,14 @@ describe('evalCommand', () => {
       vi.mocked(evaluate)
         .mockReset()
         .mockImplementationOnce(async (_testSuite, evalRecord) => evalRecord as Eval);
-      return doEval(
+      const run = doEval(
         { watch: true, write: false },
         config,
         defaultConfigPath,
         evaluateOptions as never,
       );
+      started.push(run);
+      return run;
     };
 
     /**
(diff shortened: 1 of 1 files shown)


jameshiester-oai (commented):
Reviewed base `3f942845637918a72111e632952b16daf9fb2066` and head `35a0adfe8f716af37e4ea53899ab9c61e3e5c2af`. Scope: full base...head review.

Verdict: no actionable defects found. CI is green at this head.

Verified findings:
- None.

Existing unresolved feedback: none found on the current head.

Simplification/test opportunities: none material beyond the focused lifecycle coverage already added.

Validation performed:
- `git diff --check 3f942845637918a72111e632952b16daf9fb2066...35a0adfe8f716af37e4ea53899ab9c61e3e5c2af` passed.
- `npx vitest run test/commands/eval.test.ts --sequence.shuffle=false -t "watch mode process lifecycle"` passed: 3 tests, 125 skipped.
- Five shuffled focused runs with `--sequence.seed=1..5` passed: 3 tests each, 125 skipped each.

Limitations: PR Review Toolkit was unavailable on this host (`/pr-review-toolkit:review-pr all parallel` returned no such file); I used the manual six-lens fallback plus three read-only specialist passes. I did not run the full suite locally; the current-head CI matrix is green.

PR #10576, opened 2026-08-31
fix: harden eval and provider edge cases

mldangelo-oai:
## Summary

- keep replay action wording consistent across eval warnings
- harden Codex Security usage extraction and document finding precedence
- validate malformed CLI tags and watch nested file references in JSONL tests
- add regressions for malformed tags, multiple incompatible SDK versions, missing turn results, and JSONL watch paths

## Validation

- `npm run f` (passes; existing complexity warnings remain in touched legacy functions)
- `npm run l` (passes; same existing complexity warnings)
- `npm run tsc`
- `npx vitest run test/commands/eval.test.ts test/providers/openai-codex-security.test.ts test/util/testCaseWatchPaths.test.ts --no-cache` (194 passed)

### DIFF
--- a/src/node/doEval.ts
+++ b/src/node/doEval.ts
@@ -531,6 +531,9 @@ export async function doEval(
       } = await resolveConfigs(cmdObj, defaultConfig));
     }
 
+    const describeReplayAction = (isRetryErrors: boolean | undefined) =>
+      isRetryErrors ? 'retrying errors for' : 'resuming';
+
     const persistedProviderFilterOptions = resumeEval
       ? getPersistedProviderFilterOptions(resumeEval.runtimeOptions?.providerFilter)
       : {};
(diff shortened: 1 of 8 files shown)


mldangelo:
Reviewed and pushed fixes in 4c1a8de.

### 1. The `turnResult` guard was incomplete — and its test was vacuous (P1)

`getTokenUsage` was hardened to `result?.turnResult?.usage`, but `buildScanResponse` still dereferenced `result.turnResult.model` directly 2 lines later, so the crash just moved:

```
Codex Security operation failed: Cannot read properties of undefined (reading 'model')
```

That turns a scan that already produced findings into a total error response — findings, report paths and coverage all discarded.

The test that was supposed to cover this passed anyway, because it only asserted `tokenUsage`/`cost` were `undefined`, which is trivially true of an error response. Changing `turnResult: {}` → `turnResult: undefined` therefore swapped a meaningful assertion for a vacuous one.

Fixed by guarding `result.turnResult?.model`, restoring the original `turnResult: {}` case (usage genuinely omitted, metadata still present), and adding a separate case for a fully missing `turnResult` that asserts `response.error` is `undefined` and that findings still surface. Verified the new test fails without the production fix.

Note `turnResult` is non-optional in the SDK type (`readonly turnResult: Readonly<TurnResultMetadata>`), so this is defence at the third-party boundary — reasonable given the `^0.1.18` range, but it has to be applied consistently to buy anything.

### 2. The value whitespace rule hit `--var` and contradicted `key=` (P2)

`collectKeyValueOption` backs `--var` as well as `--tag`. Rejecting a whitespace-only value made `--var "sep= "` — a deliberate single space — a hard error, while `--var "sep="` (empty) stayed valid per the existing `accepts an explicitly empty value (key=)` test, and `--var "x=hi "` kept its trailing space. Inconsistent, and an undocumented tightening of two public flags.

Constrained the key only. That still rejects the ` =value` / `=value` cases the check was added for, and adds `key =value` coverage. Values are now taken verbatim, with tests for the space, trailing-space and newline cases.

### 3. Duplicate JSONL parser (P3)

The watch path reimplemented line splitting + `JSON.parse` inline, while `parseJsonlLines` already exists ~380 lines up in the same file and is what the loader uses. Switched to the shared helper so the watcher and loader can't drift on parsing.

### Confirmed good

The JSONL watch gap itself is real — `readTests` runs `.jsonl` through `maybeLoadConfigFromExternalFile` + `_deref`, so nested `file://` refs were loaded but never watched. The `describeReplayAction` extraction and the multi-version SDK test are fine.

### Validation

`npx tsc --noEmit` clean; `biome ci` clean on all touched files (the `readTests` complexity warning is byte-identical on `main`); 323 tests pass across `openai-codex-security`, `cliOptions`, `testCaseWatchPaths`, `testCaseReader`, `eval` and `redteam/commands/run`.

PR #9517, opened 2026-05-29
ci(release): advance last-release-sha to fix release-please history-scan timeout

mldangelo:
## Summary

The `release-please` job for the **0.121.13** release ([run 26609142822](https://github.com/promptfoo/promptfoo/actions/runs/26609142822)) failed. The release itself was created first, so **npm and docker publication were unaffected** — `promptfoo@0.121.13` is live and verified (clean install + a real eval both pass), and the multi-arch docker images published and were attested. The failure is isolated to release-please's post-release history scan, and left unfixed it will recur on every future release and could eventually fail *before* a release is created.

## Root cause

After creating the release/tag, release-please walks back through `main`'s merge commits via GitHub's GraphQL API to collect commits since the last releases. The walk is floored by `last-release-sha`. That value was pinned to the **0.121.9** commit (`195df778`, 2026-04-27) by the original "harden release-please history scan" change and **never advanced**.

Four releases later that floor sits **468 commits** behind HEAD. release-please paginated ~123 pages of merge commits and GitHub's GraphQL query errored out:

```
❯ Fetching merge commits on branch main with cursor: 4860e990... 123
##[error]release-please failed: Request failed due to following response errors:
 - Something went wrong while executing your query ...
```

Measured distances from the release commit (`4860e990`):

| Boundary | Commits behind HEAD | Date |
|---|---|---|
| `last-release-sha` = `195df778` (0.121.9) | **468** ← binding floor | Apr 27 |
| `code-scan-action-0.1.6` (`40501990`) | 143 | May 20 |
| `0.121.12` tag | 144 | May 20 |

## Fix

Advance `last-release-sha` to the **code-scan-action 0.1.6** release commit (`40501990`, 2026-05-20) — the oldest still-relevant package release boundary. This:

- Cuts the scan from **468 → 143 commits**, clearing the GraphQL timeout.
- Is old enough to **preserve the pending code-scan-action 0.1.7 release** (PR #9446) — those commits all come after `40501990`.
- Keeps root (`promptfoo`) releases correct — root bounds at its own `0.121.13` tag (0 commits since), not at this floor.
- Passes the existing `test/release-please.test.ts` guard (40-char, reachable SHA).

## Test plan

- [x] `npx vitest run test/release-please.test.ts` — 3/3 pass
- [x] Confirmed `40501990…` is reachable on `main` and is the `chore(main): release code-scan-action 0.1.6` commit
[...]

PR #10579, opened 2026-08-31
ci(release): advance release-please drift floor

mldangelo-oai:
## Summary

- advance release-please’s static history floor to the oldest current package release boundary
- restore the scheduled SHA-drift guard after it crossed its 400-commit threshold

## Why this SHA

The root package’s current `0.122.2` release is `89052308…`, while code-scan-action’s current `0.2.0` release is later at `2c45764c…`. Using `89052308…` preserves unreleased commits for both packages and leaves the floor only 17 commits behind current `main`.

## Verification

- reproduced the failing guard calculation: old floor `af7ec0b5…` was 499 commits behind `main` (> 400)
- verified the new floor is reachable and 17 commits behind `main` (<= 400)
- `jq empty release-please-config.json`
- `git diff --check`

Note: focused Vitest startup was blocked by an incomplete local dependency tree (`vitest/dist/cli.js` missing), so the workflow’s exact guard calculation was validated directly.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -5,7 +5,7 @@
   "include-v-in-release-name": false,
   "commit-batch-size": 25,
   "commit-search-depth": 500,
-  "last-release-sha": "af7ec0b5f1072685a919015faf5799cc0886d71e",
+  "last-release-sha": "89052308bce06f53645b1f189ada5ac9d1897347",
   "separate-pull-requests": true,
   "packages": {
     ".": {


mldangelo-oai:
@codex review

mldangelo:
Reviewed and synced with `main` (`4dcbb55`). **The change is correct as written** — verified independently rather than taking the description's numbers on faith.

### The SHA choice is right

The guard's own rule is "advance to the OLDEST current package release commit — never newer than any package's last release, or that package loses its unreleased commits." Both release commits landed the same morning:

| Commit | Release | Time |
|---|---|---|
| `89052308` | root `0.122.2` (#10515) | 2026-08-28 03:03 |
| `2c45764c` | code-scan-action `0.2.0` (#9833) | 2026-08-28 03:33 |

`89052308` is the older of the two, so it's the correct floor. Confirmed the invariant mechanically — the floor is an ancestor of **both** packages' last-release commits, so neither package can lose unreleased commits:

```
0.122.2 (89052308bc):            floor is ancestor ✓
code-scan-action-0.2.0 (2c45764ca1): floor is ancestor ✓
```

Also confirmed there is no third package (config + `.release-please-manifest.json` list exactly `.` and `code-scan-action`, matching the newest tags), and that no newer release commit has landed on `main` that would move the correct floor — the only release commit after the proposed floor is `2c45764c` itself.

### Guard simulation

Ran the workflow's exact logic against the synced tip:

```
last-release-sha 89052308… is 26 commits behind HEAD (threshold 400)
GUARD: PASS (26 <= 400)
```

`jq empty release-please-config.json` valid; diff vs `main` is the single line and nothing else.

### This is overdue

The scheduled guard has failed **four consecutive Mondays** — Aug 10, 17, 24, and 31 — with no floor advance. The old floor was **506** commits behind `main`, already past the configured `commit-search-depth: 500`, so release-please was operating at the edge of the failure mode this guard exists to catch (#9517). Worth landing promptly.

### The CI failure was unrelated

`Test on Node 22.22 and windows-2025-vs2026 (shard 2/3)` failed on `test/providers/openai/responses/request.test.ts > should cancel an accepted background job when creation is aborted before the response arrives` — a timer/abort race. A one-line JSON diff cannot cause that. There is already a fix in flight for that exact flake in #10537, so I have not duplicated it; the resync re-runs CI regardless.

PR #10586, opened 2026-09-01
fix(http): clear error context when signature fails due to missing private key

LuckTerence:
Fixes a confusing error message when signing fails because no private key was provided. The existing logger.error only showed keyLength=0 but didn't indicate whether the key was missing or another error occurred. This patch logs keyProvided explicitly and throws a clear, actionable error when privateKey is empty.
,


mldangelo:
Thanks for this — the instinct is right (signature-auth failures are genuinely awful to debug), and I've pushed to your branch to move the fix to the layer where the information still exists. Happy to revert any of it if you disagree.

### What I found while digging in

`crypto` can't tell you what went wrong. OpenSSL 3 collapses **five** distinct operator mistakes into one opaque code — I checked each against real key material:

| Input | What crypto reports |
|---|---|
| empty | `ERR_CRYPTO_SIGN_KEY_REQUIRED: No key provided to sign` |
| whitespace-only | `error:1E08010C:DECODER routines::unsupported` |
| garbage | `error:1E08010C:DECODER routines::unsupported` |
| truncated | `error:1E08010C:DECODER routines::unsupported` |
| **public key by mistake** | `error:1E08010C:DECODER routines::unsupported` |

`createPrivateKey()` is no more specific than `sign()` here — I tried that first. So once the key reaches crypto, nothing downstream can recover the distinction. It only survives in the PEM text.

### Two things about the original approach

The `!privateKey` branch turned out to be reachable **only** for exactly-empty material — and for that case the message misdirects. Reaching it means a source *was* configured and the file behind it was empty, but the message asks the user to "ensure one of privateKey, privateKeyPath, … is set" — something they'd already done. Genuinely-absent config never gets there: the `pem`/`jks`/`pfx`/`default` branches already throw specific messages first (I added a test pinning that, since it's easy to regress).

It also left the common cases untouched — a whitespace-only or truncated key still produced the raw `DECODER` error.

### What's there now

One classifier on the PEM text, called once before `createSign`, plus `keySource` so the error names *which* input produced the key:

```
Private key from privateKeyPath /etc/certs/api.pem cannot be used:
it is a public key, not a private key
```

Distinguishes empty, whitespace-only, certificate-instead-of-key, public key, passphrase-protected, not-PEM, and truncated. Structurally-sound material returns `undefined` so crypto still reports anything subtler rather than being masked by a guess.

I kept your `?? 0` and `keyProvided=` logging additions.

Also unwrapped the outer rethrow so it no longer reads `Failed to generate signature: Error: Error: …`, and attached the original as `cause`. Worth knowing for future changes here: `new Error(msg, { cause })` **fails the build** — `src/app` typechecks this file under `lib: ES2020`, which predates `ErrorOptions`. Root `npm run tsc` passes and only the full `npm run build` catches it, so `cause` is assigned as a property instead.

### Notes

- **The failing macOS CI job wasn't you.** It was an order-dependent flake in `eval.test.ts`, fixed in #10588. I merged current `main` into your branch, which clears it.
- Added `test/providers/http-signature-key-diagnostics.test.ts` — 22 tests. Codecov flagged 0% patch coverage before; these cover every classification plus the source attribution. Verified they actually bite by deleting the call site and the public-key check and confirming the right tests fail.
- Verified live: an eval pointed at a public key now surfaces the file path and reason instead of `DECODER routines::unsupported`.
- Messages never echo key material — only which marker was found or missing — so pointing this at the wrong file (a `.env`, say) can't leak its contents. Classification is plain string matching, no regexes.

`npm run build`, `tsc`, `biome ci` clean; 432 tests pass across `test/providers/http*`.

mldangelo:
Merging this — thank you, and congratulations on your first promptfoo contribution! 🎉

Your instinct was the valuable part: signature-auth failures really were surfacing raw OpenSSL codes with no way for an operator to act on them, and nobody had picked that up. Chasing it properly turned out to be a rabbit hole — `createPrivateKey()` is no more informative than `sign()`, and OpenSSL 3 flattens five distinct mistakes into one code — so the diagnosis had to move earlier, to the PEM text. That's a subtle thing to land on from a cold start, and the bug you pointed at was real.

Net effect: pointing promptfoo at a public key, an empty file, a certificate, an encrypted key, or a truncated key now names both the offending input and the actual reason, instead of `error:1E08010C:DECODER routines::unsupported`.

Thanks for flagging it, and for being patient while I reworked the approach on your branch. Hope to see more from you.

PR #10590, opened 2026-09-01
docs(site): replace framer-motion with motion

mldangelo-oai:
The package replacement in #10582 left six blog components importing `framer-motion`, causing Knip failures and preventing strict package managers from resolving those imports. This completes the migration to [`motion/react`](https://motion.dev/docs/react-upgrade-guide) and preserves the monthly Renovate schedule for `motion`.

The lockfile adds `motion` 12.43.0 while retaining the existing `framer-motion` and `motion-dom` 12.43.0 implementation. Replaces #10582, which Renovate autoclosed.

Validation:

- Clean dependency install, Docusaurus production build, and TypeScript check.
- 256 site tests and 257 focused manifest, CLI, and provider tests passed after syncing with main.
- Knip, changed-file lint/format, and dependency consistency passed; uncached CLI echo eval passed with score 1 and no errors.
- The broader backend run encountered local socket failures; all 113 tests in those files passed on serial rerun.


mldangelo-oai:
@codex review

Please review the current head, `68dbc2cb370299581d49381c8c5ef508c6e813fc`.

mldangelo:
Reviewed. **The migration is correct and complete** — verified independently rather than from the description. One caveat about the stated validation, and one merge blocker.

### Verified correct

- **Complete.** Zero `framer-motion` references remain anywhere under `site/` (`*.tsx`/`*.ts`/`*.json`). All six components, `site/package.json`, and the Renovate rule move together.
- **`motion/react` really provides what these files use.** Its type entry is `export * from 'framer-motion'` plus explicit `motion`/`m` re-exports, so both `AnimatePresence` and `motion` resolve.
- **The retained `framer-motion` in the lockfile is legitimate, not leftover cruft** — `motion@12.43.0` declares `framer-motion: ^12.43.0` as a direct dependency, so it stays as a transitive.
- **No packaging regression from `"dev": true`** on the new `motion` entry — `framer-motion` and `motion-dom` were already recorded `dev: true` on `main`, so this matches.
- **Scope is clean.** Against the merge-base the PR touches exactly 9 files; nothing crept in.

### Verification I ran

- **Production Docusaurus build passes** — `Server: Compiled successfully`, `Client: Compiled successfully`, static files generated. This is the signal that matters here, since it SSRs the blog pages that import `motion/react`.
- **Knip clean** (only the four non-enforced config hints).
- 256 site tests pass; clean `npm ci` from the PR lockfile.

### ⚠️ The "256 site tests" aren't evidence for this change

None of the six migrated components have any test coverage, and `site`'s vitest config includes only `src/**/*.test.{ts,tsx}` — so `site/blog/**` is *structurally* excluded from the test run. Those 256 tests pass identically with or without this diff.

So I wrote a throwaway render test to get actual runtime coverage, and it passes 5/5: the bindings resolve, a `motion.div` renders inside `AnimatePresence`, and `A2ACapabilityExplorer`, `SafetySecurityQuiz`, and `MCPArchitectureVisualizer` all mount without throwing. Not proposing to add it here — widening the vitest include pattern is out of scope for a dependency swap — but flagging that the blog components are an untested surface.

### ℹ️ Worth knowing

At `12.43.0` this is future-proofing rather than a functional or size win: `motion` is a thin re-export layer over `framer-motion`, so the site now resolves both packages. Right direction per the [upgrade guide](https://motion.dev/docs/react-upgrade-guide) since `framer-motion` is the legacy alias — just not a dependency reduction.

### Merge blocker

Branch is `BEHIND` `main` (predates #10579). All 44 checks are green; it needs a sync before merge.

PR #10591, opened 2026-09-01
test(eval): isolate watch config fixtures

mldangelo-oai:
## Summary

- Isolate the executable-config and CLI test-override fixtures from unrelated default-config discovery by using explicit config paths with no default path.
- Clear the default-config cache before and after each watch-path fixture so tests do not rely on earlier reads, and verify that the executable-config test reaches the watcher.
- Make the `--tests` fixture's config re-read reachable so its no-read assertion actually verifies the override guard.
- Keep the existing assertion that executable configs are never re-read; no production behavior changes.

## Why

Main CI run [33475240233](https://github.com/promptfoo/promptfoo/actions/runs/33475240233) failed in the executable-config fixture on Node 26 Linux and Node 24 Windows. The fixture supplied an unrelated default config path, so a cold cache legitimately triggered discovery reads before the watcher logic. It passed only when an earlier test had warmed the cache. The neighboring `--tests` / `--vars` fixtures had the same dependency, confirmed by the first PR CI run's macOS shuffle seed. These tests and the implementation are unchanged by the Rollup dependency merge.

## Validation

- Reproduced on unmodified main with shuffle seed `1788242009228`: 131 passed, one failed.
- Reproduced the remaining `--vars` fixture failure with PR CI seed `1788242880264`.
- Follow-up verification is running against all three CI seeds plus seven additional seeds, then the full backend suite.

### DIFF
--- a/test/commands/eval.test.ts
+++ b/test/commands/eval.test.ts
@@ -657,6 +657,7 @@ describe('evalCommand', () => {
       // vm.runInContext() for CommonJS and is not cached. Reading it again here to
       // recover the raw `tests` value would execute the user's config a second time,
       // so executable formats are skipped.
+      defaultConfigModule.clearConfigCache();
       const config = { prompts: [], providers: [], tests: [] } as UnifiedConfig;
       vi.mocked(resolveConfigs).mockResolvedValue({
         config,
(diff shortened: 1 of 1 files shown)


mldangelo-oai:
@codex review

mldangelo-oai:
Superseded by #10588, which landed the same watch-fixture cache isolation while this PR was being validated. Keeping one repair on main; this PR is not needed.

PR #10600, opened 2026-09-01
fix(providers): normalize typed Responses messages

mldangelo-oai:
## Summary
- normalize chat-format content inside explicit Responses `type: message` input items
- preserve function calls, tool outputs, and other non-message typed items unchanged
- add regression coverage for the typed-message boundary

## Verification
- failing-before/passing-after runtime repro with Node type stripping
- `node node_modules/typescript/bin/tsc --noEmit --pretty false`
- `node node_modules/prettier/bin/prettier.cjs --check src/providers/responses/input.ts test/providers/responses/input.test.ts`
- focused Vitest/Biome were blocked locally by corrupted native optional-dependency binaries in this worktree; CI will run from a clean install

### DIFF
--- a/src/providers/responses/input.ts
+++ b/src/providers/responses/input.ts
@@ -66,9 +66,9 @@ function normalizeContentPart(part: unknown, role: unknown): unknown {
  * 'input_image', ..."), so multimodal prompts fail on every `*:responses:*` provider. Rewriting
  * the chat parts here lets the same prompt file target both surfaces.
  *
- * Only role-bearing message objects are touched. Items that carry their own `type` are
- * non-message input items (function calls, tool outputs, reasoning items) whose shape must be
- * preserved verbatim. A non-array input (a plain string prompt) is returned unchanged.
+ * Only role-bearing message objects are touched. Explicit `type: 'message'` input items are
+ * messages too; other typed items (function calls, tool outputs, reasoning items) must keep
+ * their shape verbatim. A non-array input (a plain string prompt) is returned unchanged.
  */
 export function normalizeResponsesInput<T>(input: T): T {
   if (!Array.isArray(input)) {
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
@codex review

jameshiester-oai (commented):
Reviewed base `0eb23a06116c75cc8f147febb05ced334f0721fc` through head `9e000baf2063c1646c9ac44fb08a14d19f648725`. Scope was a full fresh review; this exact head had no authenticated completion marker.

**Verdict:** Exact-head CI is fully green. I found no additional actionable defects beyond the existing current-head P1 inline finding: typed assistant history is normalized to `output_text` even though a typed easy-input message still requires `input_text`. I am not duplicating that comment.

**Verified findings**

- No new non-duplicative findings.

**Existing unresolved feedback**

- `src/providers/responses/input.ts:84`: distinguish typed input messages from actual response-output items so assistant-role history is not rewritten into an invalid Responses input shape.

**Simplification and test gaps**

- The helper remains small and cohesive; no material simplification is warranted.
- Add an assistant-role typed-message regression case alongside the new user-role case to lock the boundary down.

**Validation performed**

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/responses/input.test.ts --sequence.shuffle=false --reporter=dot` — passed: 1 file, 10 tests.
- Exact-head GitHub checks were read back: all reported checks passed, including build, platform test shards, integration, smoke, style, CodeQL, and security scan.
- Three bounded read-only specialist passes inspected correctness/API behavior, tests/QA, and independent risk; none found a distinct issue.

**Limitations:** No live paid Responses provider call was made; the focused unit test exercises normalization directly, and the OpenAI/xAI callers were inspected. The PR Review Toolkit command definition is installed locally but is not exposed as a callable slash-command endpoint in this environment, so I applied its `all parallel` lenses through the specialist passes plus orchestrator validation.

PR #10605, opened 2026-09-02
fix(assertions): reject cached latency measurements

mldangelo:
## Summary

A `latency` assertion must measure the current provider request. Previously, a cached response could silently pass using an old measurement or the cache-lookup time.

Reject stored response replays with an actionable `--no-cache` error. Track fetch-cache hits within each provider invocation so the check also works for providers that omit `cached`, including LocalAI, Hugging Face, and Ollama. Add optional `ProviderResponse.cacheHit` provenance, keeping `cached` compatible with its existing billing behavior.

OpenAI Responses reports live coalesced calls separately from stored replays. Both live subscribers remain eligible for latency grading, while billing is still attributed once. A fresh replacement for an expired cached background job also remains eligible. Update the assertion index, detailed latency documentation, and custom-provider contract.

## Validation

- Final focused run: 3,856 tests passed across assertions, evaluator, cache, OpenAI providers, and contracts. Regression coverage uses real provider/cache/evaluator code with mocked HTTP: five provider paths, cache bypass, inverse assertions, concurrent isolation, both OpenAI background coalescing paths, and expired-job replacement.
- Local CLI with an isolated disk cache and loopback fixture server: LocalAI fresh request passed (268ms), replay errored, and `--no-cache` passed (161ms). Two live background subscribers passed (527ms/526ms); replaying them produced two errors. Server counters confirmed exactly two LocalAI requests, one background creation, and one poll.
- TypeScript, application build, docs build, Knip, lint/format, and 12 built-library export checks passed.
- The full local backend run passed 24,779 tests and failed 25. Seven dependency-version checks reproduce on unchanged `main` with the same installed dependencies; the remaining failures involved loopback/proxy transport. Fresh installation was blocked by local package policy, so local validation used the existing installation, which has older Hono, js-yaml, and protobufjs versions than the lockfile. CI validates the locked dependency set.


jameshiester-oai (commented):
## Review scope

- Base: `48a71cd0163b01ba8efb2954eb0165dd810a6c6e`
- Head: `a365fd57ff5f9979648269c93427b63da22e25ce`
- Scope: fresh full `base...head` review; no prior authenticated completion marker exists for this head.
- Verdict: the change restores the handler-level guard, but the current-head existing feedback identifies one P1 and one P2 behavior gap that remain valid. Exact-head CI is green across all reported checks.

## Verified findings

### P1

- Existing feedback remains valid: [providers that discard `fetchWithCache().cached` still bypass the new guard](https://github.com/promptfoo/promptfoo/pull/10605#discussion_r3915105767). For example, LocalAI destructures only `data` and returns only `output`, so a real cache hit reaches `handleLatency` with `providerResponse.cached === undefined` while the evaluator supplies cache-lookup wall time. The latency assertion can therefore still silently pass on cached results—the defect this PR is intended to close. I verified the current code path and did not duplicate the inline comment.

### P2

- Existing feedback remains valid: [OpenAI Responses uses `cached: true` for a live coalesced subscriber](https://github.com/promptfoo/promptfoo/pull/10605#discussion_r3915105759), not only for a stored replay. The second overlapping subscriber waits for the current provider request and has current elapsed latency, but line 15 now turns its latency assertion into a cached-result error. The existing provider concurrency test confirms one live coalesced result receives this flag. I verified the current code path and did not duplicate the inline comment.

### P3

- Existing documentation feedback remains valid: [the assertion index still omits the cache-hit error contract](https://github.com/promptfoo/promptfoo/pull/10605#discussion_r3915105752). This is a discoverability gap rather than a separate runtime defect.

No additional actionable defects survived validation, so I am not adding duplicate or contrived inline comments.

The durable fix needs a cache-replay signal whose meaning is distinct from billing/coalescing and is propagated consistently by every cache-backed provider. Reusing optional `ProviderResponse.cached` for both concepts is what creates the false-negative and false-positive paths above.

The new handler tests are focused and useful, but they only construct `cached: true/false` directly. Add evaluator/provider-level regression coverage for (1) a real cache-backed provider that currently drops the flag and (2) the live coalesced Responses path.

## Tests and end-to-end QA

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/assertions/latency.test.ts --sequence.shuffle=false` — exit 0; 1 file / 11 tests passed.
- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/openai/responses/request.test.ts -t 'coalesced background' --sequence.shuffle=false` — exit 0; 1 selected test passed, 82 skipped.
- `source ~/.nvm/nvm.sh && nvm use && npm run local -- eval -c /tmp/pr10605-review.yaml --no-cache --no-progress-bar -o /tmp/pr10605-review-results.json` — expected exit 100 for one grading error; exported JSON recorded 0 passes, 0 failures, 1 error and the cache-specific latency message for a synthetic local provider response with `cached: true`.
- `git diff --check 48a71cd0163b01ba8efb2954eb0165dd810a6c6e...HEAD` — exit 0.

## Limitations

The CLI exercise used a deterministic local file provider to verify the user-visible assertion error path; I did not run a live LocalAI/Hugging Face cache replay or a remote OpenAI background request. I did not rerun the full backend/frontend suites because exact-head CI already covers those lanes and is green.

The requested PR Review Toolkit command was not available in this environment. I used the prescribed three bounded read-only specialist lenses plus orchestrator validation as the closest fallback; no claim is made that the native toolkit ran.

mldangelo-oai:
@codex review

Please review current head `e23c280f7243e6345e6d65949095ed137ab6fb06`, including cache-hit tracking for providers that omit `cached`, live background coalescing, and expired-job replacement. The three earlier findings are addressed, with provider/evaluator regressions and local CLI evidence in the updated description.

jameshiester-oai (commented):
Reviewed exact head e23c280f7243e6345e6d65949095ed137ab6fb06 against current base e5767f29, the full 12-file latency/cache diff, existing feedback, and 43 passing PR checks. Three independent read-only correctness, QA, and risk passes found one new distinct P2 in a mixed cached/live provider path (inline). The earlier P1 providers that discard cached, P2 live OpenAI Responses coalescing, and P3 assertion-index guidance are addressed by request-local tracking, explicit cacheHit:false, evaluator/provider tests, and docs. A separate P2 tracing mismatch for the coalesced Responses span is already commented on this head, so I did not duplicate it.

Validation: 106 focused latency, evaluator, and Responses tests across three files passed; root TypeScript typecheck and diff check passed. A network-free Replicate probe used a cached processing creation response followed by a fresh successful poll: the provider returned live output with one request, while the call-wide tracker returned cacheHit:true, reproducing the inline grading mismatch. The PR also reports a loopback CLI exercise for fresh, replayed, and coalesced requests; no live vendor call was made here. The optional /pr-review-toolkit:review-pr launcher was unavailable (exit 127); direct source review and QA completed.

src/evaluator.ts
```diff
@@ -1099,9 +1099,11 @@ async function callActiveProvider({
...
           )
         : activeProvider.callApi(renderedPrompt, callApiContext, callApiOptions);
-    return testSuite?.tracing
-      ? cliState.withRequestTracingConfig(testSuite.tracing, invoke)
-      : invoke();
+    const { result, cacheHit } = await withCacheHitTracking(() =>
+      testSuite?.tracing ? cliState.withRequestTracingConfig(testSuite.tracing, invoke) : invoke(),
+    );
+    // Providers can report fresh output after discarding an expired cached response.
+    return cacheHit && result.cacheHit === undefined ? { ...result, cacheHit: true } : result;
```
jameshiester-oai, line 1106: [P2] Do not turn every intermediate fetch replay into a cached final result. Replicate's `createPrediction` can replay a cached `processing` response, then poll the live status endpoint with cache busting and return a newly completed output. This call-wide marker remains true, so `handleLatency` errors with `--no-cache` even though the successful result includes a current live request. In a network-free probe, the second call replayed processing creation (one creation fetch across calls), performed a fresh poll, returned `numRequests: 1` and live output, yet tracking reported `cacheHit: true`. Use final-result provenance, or let the provider explicitly clear the intermediate marker after live completion, and cover the processing-to-live-poll path.

PR #10606, opened 2026-09-02
fix(providers): thread per-call traceparent to the OpenCode server

### LINKED ISSUE #10518: opencode:sdk provider never threads per-call traceparent to the OpenCode server, breaking trajectory:* assertions
### Summary
Every other agentic provider in promptfoo (`anthropic:claude-agent-sdk`, `openai:codex-sdk`, etc.) forwards the per-call `context.traceparent` into the subprocess/session it spawns, so OpenTelemetry spans from that call land under the right trace and `trajectory:*` assertions (`tool-used`, `tool-args-match`, `step-count`) can read them. `opencode:sdk` (`src/providers/opencode-sdk.ts`) does not — there is no `traceparent` handling anywhere in the file. As a result, `trajectory:*` assertions are silently unusable against this provider: spans either have no trace context or all share one stale traceparent from server boot, so nothing correlates back to the test case that produced it.

### Background: OpenCode has no built-in OTel, only a plugin
OpenCode itself ships with zero tracing. Real per-tool-call spans require installing a third-party plugin, `@devtheops/opencode-plugin-otel` (referenced in the user's `opencode.json`), pointed at promptfoo's OTLP receiver via `OPENCODE_ENABLE_TELEMETRY`/`OPENCODE_OTLP_*`. That plugin reads its trace-correlation context from the `OPENCODE_TRACEPARENT` env var **exactly once, at process boot**. That's the crux of why this needs a provider-side fix, not just plugin configuration: even with the plugin installed correctly, there's no mechanism today for `opencode:sdk` to give the server a new traceparent per call.

### Where the bug is (with line numbers, `promptfoo@0.122.0`)
- `src/providers/opencode-sdk.ts:1144` — `ensureClient(config)` takes no `traceparent` parameter, and short-circuits on `if (this.client) return;` — it starts one server for the whole run and never revisits it.
- `src/providers/opencode-sdk.ts:924` — `buildServerEnv(config)` builds the env object passed to the server but has no `traceparent` parameter and never sets `OPENCODE_TRACEPARENT`.
- `src/providers/opencode-sdk.ts:1456` (`callApi`) → `1498` — calls `this.ensureClient(config)` with no reference to `context.traceparent`, even though `context: CallApiContextParams` (which does carry `traceparent?: string`, see `src/types/providers.ts:96`) is right there in scope.
- Compare with the providers that get this right: `src/providers/claude-agent-sdk.ts:1464-1489` and `src/providers/openai/codex-sdk.ts:769-862` both read `context?.traceparent` and inject it into the spawned process's env per call.

### A second wrinkle once you fix the above
Simply passing a `traceparent` into `buildServerEnv()`'s returned object is **not enough** — `@opencode-ai/sdk`'s own `createOpencodeServer()` silently ignores the `env` option it's given:

```js
// node_modules/@opencode-ai/sdk/dist/v2/server.js:12-17
const proc = launch(`opencode`, args, {
    env: {
        ...process.env,                 // <-- always the caller's raw process.env
        OPENCODE_CONFIG_CONTENT: JSON.stringify(options.config ?? {}),
    },
});
```

`createOpencode()` (`node_modules/@opencode-ai/sdk/dist/v2/index.js:6-17`) spreads `options` (including `env`) straight into `createOpencodeServer`, but that function never reads `options.env` — it always spawns with the calling process's actual `process.env`. So the env object promptfoo carefully builds in `buildServerEnv()` never reaches the child process. The only env that reliably reaches it is `process.env` itself at the moment of spawn.

This is an upstream `@opencode-ai/sdk` bug, not promptfoo's — worth a separate issue against that package — but the practical implication for this fix is: promptfoo has to set `process.env.OPENCODE_TRACEPARENT` directly, immediately before calling `createOpencode()`, rather than relying on the `env` option it already builds.

### Proposed fix
1. Add an `ensureClient(config, traceparent?)` parameter. Track the currently-active traceparent on the instance (e.g. `private activeTraceparent?: string`). If a client already exists and the traceparent hasn't changed, keep it (current fast path). If it changed, close the existing server (`this.server?.close()`), clear `this.client`/`this.server`/`this.sessions`, and start a new one with the new traceparent.
2. Right before calling `createOpencode()`, set `process.env.OPENCODE_TRACEPARENT = traceparent` (or `delete` it if none), since that's the only env the spawned `opencode serve` process actually inherits (see wrinkle above).
3. Add an opt-in config flag, e.g. `restart_server_per_call?: boolean`, gating this behavior — restarting the server per call has a real perf cost, so it shouldn't be the default for users who don't need `trajectory:*`.
4. Because promptfoo can run test cases concurrently and this introduces mutable shared state (`this.client`/`this.server`) that gets swapped per call, `callApi()` needs to serialize calls on a given provider instance when `restart_server_per_call` is set — e.g. a promise-chained lock wrapping the existing call logic.
5. Pass `context?.traceparent` from `callApi()` (`src/providers/opencode-sdk.ts:1498`) into `ensureClient()`.

### Suggested touch points for whoever picks this up
- `src/providers/opencode-sdk.ts` — the fix itself (see line numbers above).
- `test/providers/opencode-sdk.test.ts` — existing suite already mocks `createOpencode`/session lifecycle; extend with: server restarts when traceparent changes + `restart_server_per_call: true`; does *not* restart when the flag is unset; `process.env.OPENCODE_TRACEPARENT` is set/cleared correctly around the `createOpencode()` call; concurrent calls under the flag don't interleave server swaps.
- `site/docs/providers/opencode-sdk.md` — document the new flag and when it's needed (trajectory assertions / real per-call trace correlation), plus the perf tradeoff.

### Workaround today
A local `patch-package` patch to `node_modules/promptfoo/dist/*` implementing exactly the above (including the `process.env.OPENCODE_TRACEPARENT` workaround) is confirmed working — `trajectory:tool-args-match` correctly reports real `opencode` tool calls with real arguments once applied. Happy to share the patch as a reference/starting diff if useful, though it targets the built `dist` bundle rather than `src/providers/opencode-sdk.ts` directly.

mldangelo:
## Summary

Fixes #10518. The `opencode:sdk` provider never forwarded `context.traceparent` to the OpenCode server it spawns, so spans emitted by `opencode serve` carried no usable trace context and `trajectory:*` assertions (`tool-used`, `tool-args-match`, `step-count`) were silently unusable against it. Every other agentic provider already threads it — `claude-agent-sdk.ts:1872`, `codex-sdk.ts:850`.

Fixing it surfaced a second, larger bug underneath.

## `buildServerEnv()` was dead code

`OpenCodeSDKModule` (this repo's local interface for the SDK) declared an `env` option that **`@opencode-ai/sdk` has never had**. Its `ServerOptions` type has no `env` field, and `createOpencodeServer()` always spawns with a hard-coded environment:

```js
// node_modules/@opencode-ai/sdk/dist/v2/server.js
const proc = launch(`opencode`, args, {
  env: { ...process.env, OPENCODE_CONFIG_CONTENT: JSON.stringify(options.config ?? {}) },
});
```

So everything `buildServerEnv()` computed was discarded in full — not just the traceparent, but the `~/.opencode/bin` PATH entry that lets `cross-spawn` resolve the CLI, the `DEBUG` log-level sync, and provider-level `env` overrides.

I verified this against **1.18.26** (latest; this repo is on `^1.18.21`) — still unfixed, and `grep -c env dist/v2/server.d.ts` is `0`. So the upstream-first path raised in the issue thread isn't available in any released version, which is why this lands as a provider-side fix.

Verified both directions against the real SDK with a stub `opencode` binary:

| what the child process saw | `OPENCODE_TRACEPARENT` | custom env var |
| --- | --- | --- |
| before (only the `env` option) | `<unset>` | `<unset>` |
| after (env applied around the spawn) | `00-4bf92f35…-00f067aa0ba902b7-01` | `from-buildServerEnv` |

The env is applied to `process.env` across the synchronous window in which `createOpencode()` reaches `cross-spawn` — it `await`s `createOpencodeServer()`, whose body runs straight through to `launch()` before its own first `await` — and restored in a `finally` that therefore runs before any suspension point. The mutation window spans no `await`, so concurrently spawning providers can neither observe the values nor inherit them. The same script confirmed `process.env` was already restored in the parent before `createOpencode()` resolved. The `env` option is still passed through so this becomes a no-op if upstream ever honors it.

## `restart_server_per_call`

`ensureClient()` started one server for the whole run and short-circuited on `if (this.client) return;`. It now takes the call's traceparent, and the new opt-in flag restarts the server when that traceparent changes — OpenCode's OTel plugin reads `OPENCODE_TRACEPARENT` only once, at process boot, so restarting is the only way to hand it a fresh one.

It's opt-in because it costs a server start per test case. When enabled:

- **The whole `ensureClient` → session → prompt lifecycle is serialized** on the provider instance. The existing `runSerializedSessionCall()` begins *after* `ensureClient()`, so it cannot protect a traceparent-driven server swap; the call now runs under a provider-wide queue key.
- **Response caching is disabled.** A cache hit returns without running the agent, so it emits no spans and the `trajectory:*` assertion the flag exists to serve would have nothing to read.
- **`baseUrl`, `persist_sessions` and `session_id` are rejected.** promptfoo can't restart a server it doesn't own, and a restart discards the session state the other two exist to preserve. Failing beats silently emitting uncorrelated traces.

If the flag is set while tracing is off there is no traceparent to tag the server with, so the provider warns once rather than silently charging you serialized, uncached calls for nothing.

With the flag off, behavior is unchanged except that `OPENCODE_TRACEPARENT` is now seeded at server boot when the ambient environment hasn't already set it — matching `claude-agent-sdk`'s precedence, and giving one correct root trace instead of none.

## Drive-by: `ensureOpenCodeModule()` race

The new concurrency test caught a pre-existing bug. Concurrent calls each ran `loadOpenCodeSDK()` and assigned `this.opencodeModule` after the await — last writer wins. `loadOpenCodeSDK()` walks several import strategies in turn, and a partially-resolved namespace could win the race, producing `createOpencode is not a function`. The in-flight load is now memoized (dropped on failure so a later call can retry).

## Testing

17 new tests in `test/providers/opencode-sdk.test.ts` (151 pass), plus a real-SDK contract test described below. Coverage: traceparent reaches the spawn; `process.env` restored afterwards; a pre-existing ambient value is restored rather than deleted, and is not overridden unless the flag owns it; all-zero traceparent treated as absent; the rest of `buildServerEnv` reaches the spawn (regression for the dead-code bug); restart on change; no restart when unchanged or when the flag is off; concurrent calls don't interleave server swaps; caching bypassed; all three incompatible-config rejections; the no-trace-context warning fires once and not at all when a usable traceparent is present.

Verified end-to-end against a real `opencode serve` with `npm run local -- eval` (tracing enabled, two test cases), A/B'd against the same config on `main`:

| | `main` | this branch |
| --- | --- | --- |
| OpenCode servers started | 1 | 2 |

with the provider logging the swap between the two test cases' actual trace IDs:

```
[OpenCode SDK] Restarting server to re-parent spans
  (traceparent 00-d0692e09e959421016b7aa5f9abf92ca-… -> 00-8de782b9ccc1548a0dd77e9925cc55c9-…)
```

(Both runs returned empty model output in my sandbox — identical on `main`, an unrelated local opencode auth issue.)

Mutation-checked rather than assumed — neutering the env application fails 4 tests, neutering the restart logic fails 2.

### Guarding against an SDK upgrade

The unit tests mock `createOpencode`, so they cannot see the synchronous-spawn assumption break — and it can break with no change to our code, if an upstream release awaits anything before spawning.

`test/providers/opencode-sdk.spawn-env.test.ts` covers that gap. It runs against the **real** SDK with a stub `opencode` CLI on `PATH` and asserts the outcome rather than the mechanism: the child process must observe the environment we asked for, and `process.env` must already be restored in the parent before the returned promise resolves. If upstream ever starts honoring the `env` option, it still passes — nothing would actually be broken.

Verified it catches the regression by simulating the upgrade — inserting `await Promise.resolve()` before the `launch()` call in `node_modules` fails both cases (the spawn no longer even sees our `PATH`, so it resolves the real CLI instead of the stub). Runs in ~0.6s; skipped when the optional SDK is absent or on Windows, where the POSIX stub would need its own shim.

`spawnWithServerEnv` moved from a private method to an exported module function so the test can drive it against the real SDK; it never used `this`.

Also verified: `tsc`, `npm run l`, `npm run f`, `npm run knip` (clean), `npm run architecture:check` (passed), and `SKIP_OG_GENERATION=true npm run build` in `site/`.

### DIFF
--- a/site/docs/providers/opencode-sdk.md
+++ b/site/docs/providers/opencode-sdk.md
@@ -646,6 +646,13 @@ Set `restart_server_per_call: true` to give each call its own server, booted wit
 traceparent:
 
 ```yaml
+# Required: without tracing there is no per-call traceparent to hand the server.
+tracing:
+  enabled: true
+  otlp:
+    http:
+      enabled: true
+
 providers:
   - id: opencode:sdk
     config:
@@ -661,6 +668,9 @@ tests:
       - type: trajectory:tool-used
         value: grep
 ```
+
+If the flag is set while tracing is off, promptfoo warns once: the restarted server has no
+trace context to be tagged with, so you pay the cost below for no correlation.
 
 ### Trade-offs
 
--- a/site/docs/providers/opencode-sdk.md
+++ b/site/docs/providers/opencode-sdk.md
@@ -188,29 +188,30 @@ When enabling write/edit/bash tools, consider how you will reset files after eac
 
 ## Supported Parameters
 
-| Parameter           | Type    | Description                                                                | Default                                |
-| ------------------- | ------- | -------------------------------------------------------------------------- | -------------------------------------- |
-| `apiKey`            | string  | Inject API key into a spawned OpenCode server for `provider_id`            | Environment variable                   |
-| `baseUrl`           | string  | URL for an existing OpenCode server                                        | Auto-start server                      |
-| `hostname`          | string  | Server hostname when starting a new server                                 | `127.0.0.1`                            |
-| `port`              | number  | Server port when starting a new server                                     | Auto-select                            |
-| `timeout`           | number  | Server startup timeout in milliseconds                                     | `30000`                                |
-| `log_level`         | string  | OpenCode server log level (`debug`, `info`, `warn`, `error`, `off`)        | Provider default                       |
-| `working_dir`       | string  | Directory for file operations and read-only default tools                  | Temporary directory                    |
-| `workspace`         | string  | Workspace identifier for workspace-aware OpenCode requests                 | None                                   |
-| `provider_id`       | string  | LLM provider (`anthropic`, `openai`, `google`, `ollama`, etc.)             | OpenCode default                       |
-| `model`             | string  | Model to use for this request                                              | OpenCode default                       |
-| `format`            | object  | Output format, including JSON Schema structured output                     | Text                                   |
-| `variant`           | string  | Provider/model variant defined in OpenCode config                          | Default variant                        |
-| `tools`             | object  | Tool configuration                                                         | None; read-only with `working_dir`     |
-| `permission`        | object  | Permission configuration for tools                                         | No extra rules; wildcard-deny baseline |
-| `agent`             | string  | Built-in or preconfigured agent to use                                     | Default agent                          |
-| `custom_agent`      | object  | Custom agent configuration when promptfoo starts the OpenCode server       | None                                   |
-| `session_id`        | string  | Resume an existing session                                                 | Create new session                     |
-| `parent_session_id` | string  | Fork from an existing session (v2 server only); inherits compacted history | None                                   |
-| `persist_sessions`  | boolean | Reuse the same session for repeated calls with the same provider config    | `false`                                |
-| `mcp`               | object  | MCP server configuration when promptfoo starts the OpenCode server         | None                                   |
-| `cache_mcp`         | boolean | Enable caching when MCP is configured                                      | `false`                                |
+| Parameter                 | Type    | Description                                                                | Default                                |
+| ------------------------- | ------- | -------------------------------------------------------------------------- | -------------------------------------- |
+| `apiKey`                  | string  | Inject API key into a spawned OpenCode server for `provider_id`            | Environment variable                   |
+| `baseUrl`                 | string  | URL for an existing OpenCode server                                        | Auto-start server                      |
+| `hostname`                | string  | Server hostname when starting a new server                                 | `127.0.0.1`                            |
+| `port`                    | number  | Server port when starting a new server                                     | Auto-select                            |
+| `timeout`                 | number  | Server startup timeout in milliseconds                                     | `30000`                                |
+| `log_level`               | string  | OpenCode server log level (`debug`, `info`, `warn`, `error`, `off`)        | Provider default                       |
+| `working_dir`             | string  | Directory for file operations and read-only default tools                  | Temporary directory                    |
+| `workspace`               | string  | Workspace identifier for workspace-aware OpenCode requests                 | None                                   |
+| `provider_id`             | string  | LLM provider (`anthropic`, `openai`, `google`, `ollama`, etc.)             | OpenCode default                       |
+| `model`                   | string  | Model to use for this request                                              | OpenCode default                       |
+| `format`                  | object  | Output format, including JSON Schema structured output                     | Text                                   |
+| `variant`                 | string  | Provider/model variant defined in OpenCode config                          | Default variant                        |
+| `tools`                   | object  | Tool configuration                                                         | None; read-only with `working_dir`     |
+| `permission`              | object  | Permission configuration for tools                                         | No extra rules; wildcard-deny baseline |
+| `agent`                   | string  | Built-in or preconfigured agent to use                                     | Default agent                          |
+| `custom_agent`            | object  | Custom agent configuration when promptfoo starts the OpenCode server       | None                                   |
+| `session_id`              | string  | Resume an existing session                                                 | Create new session                     |
+| `parent_session_id`       | string  | Fork from an existing session (v2 server only); inherits compacted history | None                                   |
+| `persist_sessions`        | boolean | Reuse the same session for repeated calls with the same provider config    | `false`                                |
+| `mcp`                     | object  | MCP server configuration when promptfoo starts the OpenCode server         | None                                   |
+| `cache_mcp`               | boolean | Enable caching when MCP is configured                                      | `false`                                |
+| `restart_server_per_call` | boolean | Restart the server per call so spans correlate to each test case           | `false`                                |
 
 ## Supported Providers
 
@@ -608,7 +609,7 @@ MCP configurations containing environment variables, positional command argument
 OAuth, URL credentials, or query strings remain uncached so secrets are not included in persistent
 cache keys. Positional command arguments can contain opaque credentials such as database URLs.
 Response caching is also disabled for credential-bearing or signed `baseUrl` values, explicit
-permission rules, `session_id`, and `persist_sessions`. API credentials use a non-secret,
+permission rules, `session_id`, `persist_sessions`, and `restart_server_per_call`. API credentials use a non-secret,
 process-local cache scope so different provider instances cannot share responses. Repeated calls
 through one locally started server retain a stable scope without storing or hashing its credentials
 or environment into the persistent cache key.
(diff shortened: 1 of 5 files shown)


jameshiester-oai (commented):
## Review scope

- Base: `48a71cd0163b01ba8efb2954eb0165dd810a6c6e`
- Head: `4e38abe83148bd210b27f23b15a4c72944c5a969`
- Scope: fresh full `base...head` review; no prior authenticated completion marker exists for this head.
- Verdict: the traceparent plumbing and spawn-contract coverage are substantial, but four existing current-head lifecycle/platform findings remain valid. Exact-head CI is green across all reported checks.

## Verified findings

### P1

- [Fixed-port restart can race the old server shutdown](https://github.com/promptfoo/promptfoo/pull/10606#discussion_r3915221428). `closeServer()` calls the SDK's synchronous `server.close()`, while the installed SDK only sends termination without awaiting child exit; a replacement using the same configured `port` can hit `EADDRINUSE`.
- [Mixed per-prompt restart settings bypass lifecycle serialization](https://github.com/promptfoo/promptfoo/pull/10606#discussion_r3915221439). `prepareCall()` permits the flag via prompt config, but only the flagged call takes the lifecycle queue; an unflagged concurrent call can use or prompt against the shared server while the flagged call swaps it.
- [The newly-effective PATH prepend is malformed on Windows](https://github.com/promptfoo/promptfoo/pull/10606#discussion_r3915221445). Production `buildServerEnv()` still joins with a hard-coded `:`; Windows needs `path.delimiter`, and the real-SDK contract test intentionally skips Windows.
- [Ephemeral-session cleanup escapes the lifecycle queue](https://github.com/promptfoo/promptfoo/pull/10606#discussion_r3915385572). The queue releases before the outer `finally` deletes the session, so the next traced call can close/swap the server first and the previous cleanup can run through a different client.

I independently verified each current code path and did not duplicate the existing inline comments. The older fixed-delay test comment no longer applies: the current head uses deferred barriers.

No additional actionable defects survived validation.

The provider needs one instance-wide lifecycle boundary that covers server swap, prompt, abort listener removal, ephemeral-session deletion, and temp-dir cleanup whenever restarts are possible. Keeping restart mode prompt-overridable while queue selection is per call makes the concurrency contract difficult to reason about.

Add focused regressions for fixed-port delayed shutdown, mixed flagged/unflagged calls, delayed session deletion during a queued restart, and Windows PATH composition.

## Tests and end-to-end QA

- `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/opencode-sdk.test.ts test/providers/opencode-sdk.spawn-env.test.ts --sequence.shuffle=false` — exit 0; 2 files / 153 tests passed.
- The real-SDK spawn contract subset ran within that command and passed 2/2 tests against a stub `opencode` CLI.
- `git diff --check 48a71cd0163b01ba8efb2954eb0165dd810a6c6e...HEAD` — exit 0.

## Limitations

I did not run a live authenticated OpenCode model call or browser-visible trajectory assertion; the real-SDK QA used the repository's bounded stub CLI. I did not rerun the full backend/frontend suites because exact-head CI already covers those lanes and is green.

The requested PR Review Toolkit command was not available in this environment. I used the prescribed three bounded read-only specialist lenses plus orchestrator validation as the closest fallback; no claim is made that the native toolkit ran.

PR #2999, opened 2025-02-06
feat(assertions): add `threshold` support for `llm-rubric`

mshavliuk:
This PR adds the `threshold` support for `llm-rubric` assertions, while preserving backward-compatibility for rubrics that include `pass` property in its output.

ref #2998

PR #10611, opened 2026-09-02
fix(matchers): fail closed when llm-rubric grader response omits its verdict

AUTHENSOR:
## The bug

In `runJsonGradingPrompt` (src/matchers/rubric.ts), a grader JSON response missing the `pass` field defaulted to passing:

```ts
let pass = parsed.pass ?? true;
```

So a grader response like `{"score": 0, "reason": ...}` recorded a **pass** with score 0, and a reason-only response (`{"reason": ...}`) passed with score 1. The grader failing to emit a verdict was indistinguishable from the grader approving the output.

This regressed in #2999, which replaced the previous score-based derivation with the unconditional default.

## The fix

- When `pass` is omitted and `score` is present: derive `pass` from `score > 0` (restores the pre-#2999 behavior).
- When the response carries **neither** a verdict nor a score: return a grader failure (`Grader response contained neither a pass verdict nor a score`) instead of a silent pass.
- Explicit boolean `pass`, string coercion (`true/yes/pass/y`), and threshold semantics are unchanged.

## Testing

Four regression tests added to `test/matchers/llm-rubric.test.ts`:

1. neither-field response fails with the explanatory reason
2. omitted pass with positive score derives pass
3. omitted pass with zero score fails
4. explicit `pass: true` alongside `score: 0` still passes

Full file suite: 104 tests passing. Found during an external integrity audit of grader-response handling across eval tooling.

### DIFF
--- a/src/matchers/rubric.ts
+++ b/src/matchers/rubric.ts
@@ -828,9 +828,10 @@ function deriveVerdictFromGrader(
     pass = /^(true|yes|pass|y)$/i.test(String(parsed.pass));
   }
 
-  let score = parsed.score;
-  if (typeof score !== 'number') {
-    score = Number.isFinite(Number(score)) ? Number(score) : Number(pass);
+  let score: number = typeof parsed.score === 'number' ? parsed.score : 0;
+  if (typeof parsed.score !== 'number') {
+    const asNum = Number(parsed.score);
+    score = Number.isFinite(asNum) ? asNum : Number(pass);
   }
 
   if (hasThreshold) {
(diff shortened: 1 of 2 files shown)


AUTHENSOR:
Both review findings addressed in the pushed update: the score-only derivation now uses the configured threshold (score >= threshold, so an exact-at-floor result passes) and the new suite restores its grader spies after each test. File suite at 105/105.

AUTHENSOR:
Following up. CI green, security review clean, ready for review when bandwidth allows.

PR #10621, opened 2026-09-03
fix(redteam): align tool discovery generation and grading

jameshiester-oai:
Tool-discovery grading can flag public capability lists and faithful transformations of supplied text as leaks, while the shared refusal shortcut can skip responses that also disclose protected information. This change distinguishes actual tool disclosure from ordinary capability descriptions and text reuse, enforces capability nondisclosure when the application purpose explicitly requires it, and grades nonempty refusals with the plugin judge.

Attack generation now excludes echoing or transforming supplied tool details, including fictional-schema substitutions during multi-turn setup. The constraint is included in local and remote generation and retained in test modifiers so Hydra receives it on every turn. Known schemas, actual-inventory disclosure, and conditional confirmation of real access remain valid probes. Multiline Q&A completion prompts are preserved.

## Validation

- 109 focused tests passed across tool-discovery generation/grading, plugin registration, and Hydra after syncing with main.
- TypeScript and changed-file lint/formatting passed. Library and web builds passed; the sandbox-blocked tsx postbuild step was rerun separately.
- Five freshly generated Hydra cases against a GPT-5.4 Mini Responses target with an explicit nondisclosure prompt and working echo tool completed with zero execution errors. All five converged on conditional access confirmation instead of the earlier fictional-schema transformation false positives.
- Paired replay: the real function returned the conditional marker and executed; a nonexistent function returned `NO_MATCH` with no tool execution.
- Earlier grader regression eval: 38/38 fixed-response cases passed with GPT-5.5, including faithful text transformations and independent disclosure cases.


mldangelo-oai (approved):
stamped!

PR #10638, opened 2026-09-04
fix(security): match env credential names case-insensitively, and stop redacting vendor names

mldangelo:
Follow-up to #10628. Codex and Copilot both reviewed it after merge and found a hole and a false positive. Both reproduce against the merged code.

## A lowercase name bypassed redaction entirely (codex, P1)

`isSecretEnvVarName` required SCREAMING_SNAKE_CASE. Nothing requires an environment variable to be uppercase, and a lowercase one reaches the subprocess identically:

```
env: { github_token: 'ghp_...' }   ->  ghp_...       LEAKED
env: { GITHUB_TOKEN: 'ghp_...' }   ->  [REDACTED]
```

The name is uppercased before matching, so both redact now.

## Suffix-matching short words redacted real settings (codex P2, Copilot)

`PORTKEY` ends with `KEY`, so `PORTKEY_API_BASE_URL` — a documented Portkey endpoint declared in `src/envars.ts` and consumed in `providers/portkey.ts` — was being replaced with `[REDACTED]` in persisted configs. `MONKEY`, `OAUTH_SCOPE` and `WATSONX_AI_AUTH_TYPE` matched the same way.

The word list is now split by how ambiguous each word is:

| Match | Words | Why |
| --- | --- | --- |
| whole word only | `KEY`, `PWD`, `DSN` | short enough to sit inside ordinary words |
| whole word **or** suffix | `PASSWORD`, `PASSWD`, `PASSPHRASE`, `SECRETKEY`, `SECRET`, `TOKEN`, `CREDENTIAL(S)`, `APIKEY` | long enough that a suffix match is unambiguous, so fused spellings like `PGPASSWORD` and `AWS_SECRETKEY` still match |

`AUTH` is dropped: on its own it names a method or endpoint far more often than a credential, and the real cases (`AUTH_TOKEN`) match on `TOKEN` anyway.

## Verified both directions

| Redacts | Does not |
| --- | --- |
| `GITHUB_TOKEN`, `github_token`, `Database_Password` | `PORTKEY_API_BASE_URL`, `WATSONX_AI_AUTH_TYPE` |
| `PGPASSWORD`, `AWS_SECRETKEY`, `MY_API_KEY` | `OAUTH_SCOPE`, `MONKEY` |
| `AWS_SECRET_ACCESS_KEY`, `STRIPE_SECRET_KEY`, `SIGNING_PASSPHRASE` | `MAX_TOKENS`, `RETRY_KEYS`, `LOG_LEVEL`, `AWS_REGION` |

Both tables are in the test file, plus an integration case asserting a lowercase credential redacts while a vendor URL survives.

`npx vitest run test/util/ test/models/ test/providers/` — 11619 passed. `npm run tsc` and `biome ci .` clean.

https://claude.ai/code/session_01F7jDD7WZBVV2GUpknv9Aff

### DIFF
--- a/src/util/sanitizer.ts
+++ b/src/util/sanitizer.ts
@@ -224,14 +224,17 @@ const ENV_SECRET_SUFFIX_WORDS = [
   'CREDENTIAL',
   'SECRET',
   'TOKEN',
-  'APIKEY',
-  'PRIVATEKEY',
-  'SECRETKEY',
-  'ACCESSKEY',
-  'SIGNINGKEY',
   'PWD',
   'DSN',
-];
+  // Every credential name in SECRET_FIELD_NAMES that ends in `key`, so a prefixed spelling
+  // (`APP_ENCRYPTIONKEY`, `VENDOR_CERTKEY`) is still caught even though the exact-name match
+  // cannot see past the prefix. Derived rather than hand-listed so the two stay in sync; all
+  // are at least six characters, so none of them matches `PORTKEY` or `MONKEY`.
+  ...[...SECRET_FIELD_NAMES].filter((name) => name.endsWith('key') && name.length > 3),
+  // Not in SECRET_FIELD_NAMES on its own — that set carries `accesskeyid` — but the bare
+  // compound shows up in env vars.
+  'accesskey',
+].map((word) => word.toUpperCase());
 
 /**
  * Secret only as the final `_`-delimited word. `MLFLOW_BASIC_AUTH` and `NPM_CONFIG__AUTH`
--- a/src/util/sanitizer.ts
+++ b/src/util/sanitizer.ts
@@ -203,28 +203,32 @@ export function isSecretField(fieldName: string): boolean {
 }
 
 /**
- * Credential words that make an environment variable name secret-bearing.
- *
- * Singular only, on purpose: each `_`-delimited word is matched by suffix, so `MAX_TOKENS`
- * (word `TOKENS`) and `RETRY_KEYS` stay out of the match while `GITHUB_TOKEN`,
- * `STRIPE_SECRET_KEY` and `PGPASSWORD` are caught.
+ * Credential words matched as a whole `_`-delimited word only. These are short enough to
+ * appear inside ordinary words, so suffix-matching them redacts real settings: `PORTKEY`
+ * ends with `KEY` (`PORTKEY_API_BASE_URL` is an endpoint, not a secret), and so does
+ * `MONKEY`.
  */
-const ENV_SECRET_WORDS = new Set([
-  'TOKEN',
-  'SECRET',
+const ENV_SECRET_WHOLE_WORDS = new Set(['KEY', 'PWD', 'DSN']);
+
+/**
+ * Credential words matched as a whole word *or* a suffix, so the fused vendor spellings are
+ * caught: `PGPASSWORD`, `AWS_SECRETKEY`, `MYAPPTOKEN`. Each is long and specific enough that
+ * a suffix match is unambiguous. Singular on purpose — `MAX_TOKENS` ends with `TOKENS`, which
+ * does not end with `TOKEN`.
+ */
+const ENV_SECRET_SUFFIX_WORDS = [
   'PASSWORD',
   'PASSWD',
-  'PWD',
-  'KEY',
-  'APIKEY',
-  'CREDENTIAL',
-  'CREDENTIALS',
   'PASSPHRASE',
-  'AUTH',
-  'DSN',
-]);
+  'SECRETKEY',
+  'SECRET',
+  'TOKEN',
+  'CREDENTIALS',
+  'CREDENTIAL',
+  'APIKEY',
+];
 
-/** SCREAMING_SNAKE_CASE (underscores optional): `GITHUB_TOKEN`, `PGPASSWORD` — not `apiKey`. */
+/** An environment-variable-shaped name. Case is normalized before this is applied. */
 const ENV_VAR_NAME_RE = /^[A-Z][A-Z0-9_]*$/;
 
 /**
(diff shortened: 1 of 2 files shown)


mldangelo:
@codex review

mldangelo:
Addressed all four findings — thanks, these were right and the first narrowing had reopened real holes.

| Finding | Case | Now |
| --- | --- | --- |
| Copilot: leading underscore | `_GITHUB_TOKEN` | redacted (pattern accepts a leading `_`) |
| codex P1: `AUTH` dropped wholesale | `MLFLOW_BASIC_AUTH` | redacted (`AUTH` as the **final** word) |
| codex P2: terminal `_AUTH` | `NPM_CONFIG__AUTH` | redacted (same rule) |
| codex P1: fused `KEY`/`PWD`/`DSN` | `GCP_PRIVATEKEY`, `DBPWD`, `DATABASEDSN` | redacted (explicit compounds + `PWD`/`DSN` suffix) |

The false positives that prompted the narrowing stay fixed: `PORTKEY_API_BASE_URL`, `WATSONX_AI_AUTH_TYPE`, `OAUTH_SCOPE` and `MONKEY` are all left alone. 15 credential names and 13 non-credential names are pinned in the test tables.

@codex review

mldangelo:
Fixed in the latest commit — and the finding pointed at the real defect, which was hand-maintaining a list parallel to `SECRET_FIELD_NAMES`.

The key-compound suffixes are now **derived** from `SECRET_FIELD_NAMES` (every entry ending in `key`), so they cannot drift out of sync again. `APP_ENCRYPTIONKEY`, `VENDOR_CERTKEY` and `SVC_SIGNINGKEY` all redact now, and since the shortest derived compound is six characters, `PORTKEY`/`MONKEY`/`TURKEY` stay unaffected.

Test tables now cover 19 credential names and 14 non-credential ones.

src/util/sanitizer.ts
```diff
@@ -203,29 +203,52 @@ export function isSecretField(fieldName: string): boolean {
...
+].map((word) => word.toUpperCase());
 
-/** SCREAMING_SNAKE_CASE (underscores optional): `GITHUB_TOKEN`, `PGPASSWORD` — not `apiKey`. */
-const ENV_VAR_NAME_RE = /^[A-Z][A-Z0-9_]*$/;
+/**
+ * Secret only as the final `_`-delimited word. `MLFLOW_BASIC_AUTH` and `NPM_CONFIG__AUTH`
+ * hold a credential; `WATSONX_AI_AUTH_TYPE` names a method and `OAUTH_SCOPE` is not an
+ * `AUTH` word at all.
+ */
+const ENV_SECRET_TERMINAL_WORDS = new Set(['AUTH']);
```
jameshiester-oai, line 244: [P1] Preserve fused AUTH credential names in env redaction. Restricting `AUTH` to the final underscore-delimited word leaves `HTTP_BASICAUTH` and `PROXY_BEARERAUTH` unmatched, whereas the previous suffix matcher redacted them. At this head, `sanitizeObject({env:{HTTP_BASICAUTH:'synthetic-short-canary'}})` retains the value, and an actual `npm run local -- eval --no-cache -o /tmp/...` retained it under exported `config.providers[0].env`; the same export redacted `github_token` and correctly kept `PORTKEY_API_BASE_URL`. A short basic/bearer credential evades `looksLikeSecret`, so persisted/shared configs can expose it. Cover unambiguous fused credential words such as `BASICAUTH`/`BEARERAUTH` and test the export boundary without reintroducing false positives for `OAUTH_SCOPE` or `WATSONX_AI_AUTH_TYPE`. This is distinct from the already-reported `MASTERKEY`/`HMACKEY` gap.

jameshiester-oai (commented):
Reviewed base `281f8d42984771608baf6e925ca24bb090554027` → head `9b76bb4601ee7a3db96b64ccb0425f22d070729e` (full PR; first automation review of this head). CI is green, but redaction still has a verified security gap.

**P1 — new fused-AUTH bypass (inline):** `HTTP_BASICAUTH` and `PROXY_BEARERAUTH` used to match the suffix `AUTH` but no longer match the final-word-only set. Short credential values survive `sanitizeObject` and a real local eval's exported provider `env`, even while lowercase `github_token` redacts. Recognize unambiguous credential compounds such as `BASICAUTH`/`BEARERAUTH` without broadly matching `OAUTH_SCOPE` or `WATSONX_AI_AUTH_TYPE`; add an export-boundary regression test.

**Already reported, still applicable:** the current-head, unresolved [fused-key finding](https://github.com/promptfoo/promptfoo/pull/10638#discussion_r3935561245) demonstrates `APP_MASTERKEY` and `SERVICE_HMACKEY` leaking short values; my eval independently reproduced the first. I have not duplicated that inline comment. Earlier leading-underscore, terminal `_AUTH`, and known compound-key feedback is addressed at this head.

**QA:** inspected the full two-file diff, sanitizer call sites, persisted/exported config path, existing comments and exact-head checks; applied code, tests, comments/docs, error handling, types, simplification and security lenses with three independent read-only specialist passes. `source ~/.nvm/nvm.sh && nvm use && ./node_modules/.bin/vitest run test/util/sanitizer.test.ts test/models/evalResult.test.ts --reporter=dot`: 326 passed, exit 0. `source ~/.nvm/nvm.sh && nvm use && npm run tsc`: passed, exit 0. Offline source probe: `HTTP_BASICAUTH`/`PROXY_BEARERAUTH` and the already-reported fused key names remained in an env map; `MLFLOW_BASIC_AUTH` redacted and benign `PORTKEY_API_BASE_URL` remained. `PROMPTFOO_CONFIG_DIR=/tmp/p10638-state npm run local -- eval -c /tmp/p10638-eval.yaml --no-cache -o /tmp/p10638-eval.json`: echo eval passed 1/1, score 1, no error; exported `config.providers[0].env` redacted synthetic `github_token` but retained synthetic `HTTP_BASICAUTH` and `APP_MASTERKEY`, as well as benign Portkey URL. No real credentials or external provider calls were used.

**Simplification/test gaps and limits:** deriving known key compounds from `SECRET_FIELD_NAMES` reduces list drift, but a narrow positive set for fused AUTH forms plus an export-path test is still needed. This is a heuristic sanitizer; no exhaustive assurance for arbitrary environment-variable names is claimed.

PR #10639, opened 2026-09-04
fix(providers): preserve claude mcp stdio env

### LINKED ISSUE #10509: Claude Agent SDK provider drops per-server `env` for stdio MCP servers

Asthenia0412:
## Summary
- Preserve per-server `env` when converting command/args and path-based stdio MCP servers for the Claude Agent SDK provider.
- Validate stdio server `env` maps so malformed non-string values fail closed instead of being forwarded ambiguously.
- Document per-server MCP env usage for Claude Agent SDK process-based servers.

## Why
Stdio MCP servers commonly use environment variables to decide which tools they register. Dropping the configured `env` silently broadens or changes the tool surface exposed to an agent, which makes least-privilege agent evals look stricter than they actually are.

Fixes #10509

## Validation
- `npx vitest run test/providers/mcp/transform.test.ts` initially failed on the new env preservation and malformed-env tests before the implementation change.
- `npx vitest run test/providers/claude-agent-sdk.test.ts test/providers/mcp/transform.test.ts`
- `node --import tsx - <<SCRIPT` local smoke: verified real `transformMCPConfigToClaudeCode` preserves env for command/path stdio servers and rejects non-string env values.
- `npx @biomejs/biome check src/providers/mcp/transform.ts src/providers/mcp/types.ts test/providers/mcp/transform.test.ts`
- `npx prettier --check site/docs/providers/claude-agent-sdk.md`
- `git diff --check`

## Notes
- `npm run tsc` still fails locally on the existing optional `@openai/codex-security` import/type errors in `src/providers/openai/codex-security.ts`; this branch does not touch that provider.

### DIFF
--- a/site/docs/providers/claude-agent-sdk.md
+++ b/site/docs/providers/claude-agent-sdk.md
@@ -376,6 +376,8 @@ providers:
           - command: node
             args: ['mcp-server.js']
             name: local-server
+            env:
+              MY_SERVER_ENABLE_TOOLSET_A: 'true'
 
       strict_mcp_config: true # Only use configured servers (true by default)
 ```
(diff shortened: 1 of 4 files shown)


Asthenia0412:
Closing this as superseded: upstream main already includes the stdio MCP env fix via #10557. Thanks!

PR #10656, opened 2026-09-05
docs: add The Grid guide and example using the generic openai path

CastilloLuis:
## What

Adds a docs page and an example for **The Grid** using the **generic `openai` path** — no new provider prefix and no registry entry.

- `site/docs/providers/thegrid.md`
- one row in `site/docs/providers/index.md`
- `examples/provider-thegrid/` (README + config)

## Why this shape

`site/docs/contributing.md` is explicit:

> Before writing code, check whether the service already works through the `openai` provider… If that is all the service needs, contribute docs or an example that uses the generic path; do not add a new provider prefix or registry entry.

The Grid is a plain OpenAI-compatible endpoint. It needs a base URL and an API key and nothing else, so it does not clear the "value beyond the generic path" bar and I have not proposed a dedicated provider. This follows the existing `vllm` / `llamafile` / `openllm` / `text-generation-webui` pages, which document a service via `openai:chat:` + `apiBaseUrl`.

## What's actually worth documenting

Three things that aren't obvious from the OpenAI page and cost time to discover:

1. **Model names are capability tiers**, not lab model names — `text-standard`, `code-prime`, `agent-max` route to whichever model currently qualifies. Good for application evals; the page warns to use the lab-pinned `*-latest` instruments when a number has to be reproducible.
2. **Reasoning tokens count toward the output budget** without appearing in the response, so a budget sized for the visible answer truncates it. The page recommends `max_completion_tokens` with headroom, and `showThinking: false` when The Grid is the judge.
3. **Cost can't be estimated automatically** — The Grid is market-priced and returns no per-token rate, so the page points at `inputCost` / `outputCost` overrides rather than letting cost silently read as unknown.

## Verification

- The example config validates against the committed `site/static/config-schema.json` (`examples/provider-groq` used as a control — 0 errors both).
- The endpoint behaviour described was checked directly: chat completions, streaming and tool calls all work against `https://api.thegrid.ai/v1`, and `GET /v1/models` returns a spec-shaped list.
- I could not run `promptfoo eval` locally — this checkout requires Node ≥ 22.22 and I'm on 22.16 — so the example is schema-validated rather than executed. Flagging that rather than implying I ran it.

Disclosure: I work on The Grid.


CastilloLuis:
All findings addressed. Every one was correct; I verified each against the source before changing anything.

**P1, credential fallback.** Confirmed at `src/providers/openai/index.ts:131-138`: `getApiKey()` falls through `apiKeyEnvar` to `OPENAI_API_KEY`. The example now binds `apiKey: '{{env.THEGRID_API_KEY}}'` directly, and the docs carry a warning that a missing variable sends your OpenAI credential to `api.thegrid.ai`.

**Token budget.** Confirmed at `chat.ts:195-196`: `max_completion_tokens` is only forwarded when `isReasoningModel` is true, which Grid instrument names are not. Docs and example now use `max_tokens` and say why.

**Cost overrides.** Confirmed at `billing.ts:970-973`: `if (!modelRates) return undefined` fires before `calculateTextCost` reads `inputCost`/`outputCost`. I removed that advice entirely rather than documenting an option that cannot work, and pointed at The Grid's own usage endpoints instead.

**Endpoint variable precedence.** Confirmed at `index.ts:116-125`. The page now states that `OPENAI_API_HOST` and `OPENAI_API_BASE_URL` both win over `OPENAI_BASE_URL`, and recommends setting `apiBaseUrl` on the provider.

**Front matter.** Added `title` (8 chars) and expanded the description to 158 chars, inside the 150-160 contract in `site/AGENTS.md`.

**`*-latest` reproducibility.** You were right and my wording was wrong: those aliases move when the lab ships a new version. I removed the reproducibility recommendation.

**Recording the resolved model.** This one landed on advice I had just added in the previous push, and it was also correct. `chat.ts:816-838` discards the top-level `model`, so users cannot capture provenance from an eval. The page now says plainly that no Grid instrument is immutable and promptfoo cannot record which model served a request, and directs anyone needing a reproducible published number to a provider with versioned model ids.

CastilloLuis:
Addressed the P1 posted just after my last push.

The finding is right, and the previous revision only warned about it in prose. `getApiKey()` resolves `config.apiKey || apiKeyEnvar || OPENAI_API_KEY`, and `'{{env.THEGRID_API_KEY}}'` renders to `""` when the variable is unset or blank. Empty string is falsy, so it fell through to the caller's OpenAI key.

Both the example and the docs now use:

```yaml
apiKey: '{{ env.THEGRID_API_KEY | default("THEGRID_API_KEY_NOT_SET", true) }}'
```

The second argument is the part that matters: `default` substitutes on undefined only, unless you ask it to treat any falsy value as missing. Verified against nunjucks 3 with the same globals wiring `getNunjucksEngine` uses:

| THEGRID_API_KEY | `{{env.THEGRID_API_KEY}}` | with the sentinel |
| --- | --- | --- |
| unset | `""` | `"THEGRID_API_KEY_NOT_SET"` |
| empty string | `""` | `"THEGRID_API_KEY_NOT_SET"` |
| real key | `"real-key"` | `"real-key"` |

So a missing or blank key now reaches The Grid as a sentinel and returns a 401 naming the variable, instead of silently authenticating as the caller against OpenAI. The troubleshooting table row was rewritten to match.

Unrelated but while I was in the file: I dropped the line claiming `/v1/models` currently returns `"pricing": null`. It is populated again. The durable point, that a market price moves and a static cost table would be wrong either way, is kept.

PR #10665, opened 2026-09-06
fix(assertions): recognize OpenAI Responses function_call items in tool-call-f1

### LINKED ISSUE #10649: tool-call-f1 misses OpenAI Responses function_call outputs
### Describe the bug

`tool-call-f1` returns `score: 0` and `pass: false` for a correct OpenAI Responses `function_call` emitted as a JSON string. Wrapping the same call in an array returns `score: 1`.

This is reachable through the built-in Responses providers: without `functionToolCallbacks`, `ResponsesProcessor` and `FunctionCallbackHandler` serialize each call as JSON and join multiple calls with newlines. `extractToolNames` parses the strings but does not recognize a single `type: 'function_call'` object.

### To reproduce

Save this API-key-free configuration as `repro.yaml`:

```yaml
providers:
  - echo
prompts:
  - '{{payload}}'
defaultTest:
  assert:
    - type: tool-call-f1
      value: [get_weather]
tests:
  - description: Responses single call
    vars:
      payload: '{"type":"function_call","id":"fc_1","call_id":"call_1","name":"get_weather","arguments":"{\"city\":\"NYC\"}"}'
  - description: Same call in an array
    vars:
      payload: '[{"type":"function_call","id":"fc_1","call_id":"call_1","name":"get_weather","arguments":"{\"city\":\"NYC\"}"}]'
```

Run `promptfoo eval -c repro.yaml --no-cache -o results.json`.

The echo provider replays the output shape without making a model request. I also reproduced the failure by passing the output of the real `ResponsesProcessor` and `FunctionCallbackHandler` into `handleToolCallF1`, with no callbacks configured.

### Actual behavior

- Single call: `score: 0`, `pass: false`; reason includes `Expected: [get_weather], Called: [(none)]`.
- Array control: `score: 1`, `pass: true`.
- Multiple newline-separated Responses calls and text followed by a call are also missed.

### Expected behavior

Both cases should score 1 and pass. A Responses `function_call` with a string `name` should be recognized just as an Anthropic `tool_use` block is. An unrelated object that merely has a `name` should not count as a tool call.

### System information

- Promptfoo: 0.122.2, main at `bfcd0674689c538df416b7cfac4e762614149fb7`
- Node.js: 24.20.0
- OS: Windows
- Reproduction uses no model API keys.

### Additional context

The shared Responses processor is used by the OpenAI, Azure, and xAI providers. The failure is in tool-name extraction, not the F1 calculation. #10120 changes inverse scoring and does not address this parsing path.

he-yufeng:
## Summary

Fixes #10649.

`extractToolNames` recognized an OpenAI Responses `function_call` item only when it arrived inside an array, where the simple `{name}` fallback happens to catch it. A bare single object matched no branch and extracted nothing, so `tool-call-f1` scored 0 for a correct call. That bare form is exactly what the built-in Responses providers emit when calls are serialized without `functionToolCallbacks` (one JSON object per call, newline-joined).

The fix adds the single-object branch next to the Anthropic `tool_use` one: `{ type: 'function_call', name }` contributes `name`.

## Verification

- New tests: bare object, its JSON-stringified form (the repro from the issue), and a negative case where the name does not match.
- Red-green: without the fix the two positive tests fail; with it, all 36 tests in `test/assertions/toolCallF1.test.ts` pass.
- `biome check` on the touched files: clean (the pre-existing complexity warning on `extractToolNames` is unchanged from main).

The issue's echo-provider repro passes with this build.


mldangelo-oai (approved):
Thanks for the focused fix and clear regression tests! I verified the assertion tests, the actual Responses processor output, and local CLI cases for single calls, multiple calls, and mixed text/tool output. The change fixes the incorrect zero scores while preserving the existing formats and scoring behavior. Looks good to me; I'll add the documentation update in a follow-up PR.

PR #10668, opened 2026-09-07
refactor(code-scan): reuse raw diff parser

mldangelo-oai:
## Summary

- route production code-scan diff processing through the existing raw diff parser
- remove the byte-identical private parser and type so parser tests cover the implementation used at runtime

This is a behavior-preserving cleanup. Git invocation, NUL-safe path parsing, rename/copy handling, SHA normalization, filtering, and patch generation are unchanged.

## Validation

- `npx vitest run test/codeScans --sequence.shuffle=false` (156 tests passed)
- `npx vitest run test/types/codeScan.test.ts --sequence.shuffle=false` (11 tests passed)
- `npm run tsc`
- `npm run l`
- `npx @biomejs/biome check src/codeScan/git/diffProcessor.ts`

### DIFF
--- a/src/codeScan/git/diffProcessor.ts
+++ b/src/codeScan/git/diffProcessor.ts
@@ -20,17 +20,10 @@ import logger from '../../logger';
 import { DiffProcessorError } from '../../types/codeScan';
 import { isInDenylist, MAX_BLOB_SIZE_BYTES, MAX_PATCH_SIZE_BYTES } from '../constants/filtering';
 import { annotateDiffWithLineRanges } from './diffAnnotator';
+import { parseRawDiff } from './rawDiffParser';
 
 import type { FileRecord } from '../../types/codeScan';
 import type { LineRange } from '../util/diffLineRanges';
-
-interface RawDiffEntry {
-  path: string;
-  oldPath?: string;
-  status: string;
-  shaA: string | null;
-  shaB: string | null;
-}
 
 interface NumstatEntry {
   linesAdded: number;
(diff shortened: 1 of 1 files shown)


jameshiester-oai (commented):
Reviewed base `6d0395a20520e19cf8889d572b879ec9c2831a52` → head `4821a579be06521a64e775dfa544a4434a1f8d55` (full PR; first automation review of this head). All reported CI checks are green across build, backend tests, docs, smoke, integration, style, and security.

**Verdict:** No actionable defect found. The removed local raw-diff parser and the imported existing parser are byte identical after accounting for `export`; production `processDiff()` now calls the implementation covered by parser tests. NUL-delimited paths, rename/copy handling, and zero-SHA normalization are preserved. There are no existing inline threads to duplicate. The deletion of the duplicate parser is a useful simplification.

**QA:** `source ~/.nvm/nvm.sh && nvm use && ./node_modules/.bin/vitest run test/codeScans test/types/codeScan.test.ts --reporter=dot` passed 15 files / 167 tests (exit 0). A direct read-only invocation of production `processDiff()` on this PR's base...head returned one modified file, `src/codeScan/git/diffProcessor.ts`, +1/-77 (exit 0). A local Git fixture outside tracked source with a rename from `old name.txt` to `new name.txt` plus another modified file returned the expected R100 and M records (exit 0). No hosted scanner service or credential was used. Existing parser tests cover malformed input, while the local fixture checked the actual Git processing path.

**Limits:** No material documentation or type change was needed for this internal refactor. The requested PR Review Toolkit command was unavailable (launcher exit 127); three independent read-only correctness, test, and risk passes were reconciled instead.

PR #10731, opened 2026-09-08
fix(providers): isolate SageMaker completion cache

mldangelo-oai:
SageMaker completion requests with different stop sequences, response extraction paths, or environment generation settings could reuse the same cached output. Include those effective settings and the model type resolved from the provider ID in the cache key, while preserving explicit configuration precedence and cached token accounting.

Validation: 21 mocked provider tests passed, covering fresh requests, matching cache replays, changed settings, explicit zero values, and endpoint errors. TypeScript, package bundles, and the UI build passed. A built CLI evaluation passed with six cache replays and seven mocked requests, including the expected error path. Lint and formatting passed.

The build's postbuild command hit an environment restriction on the `tsx` IPC socket; the same script passed with `node --import tsx scripts/postbuild.ts`.

### DIFF
--- a/src/providers/sagemaker.ts
+++ b/src/providers/sagemaker.ts
@@ -647,12 +647,20 @@ export class SageMakerCompletionProvider extends SageMakerGenericProvider implem
     // Create a deterministic representation of the request parameters
     const configForKey = {
       endpoint: this.getEndpointName(),
-      modelType: this.config.modelType,
+      modelType: this.modelType,
       contentType: this.getContentType(),
       acceptType: this.getAcceptType(),
-      maxTokens: this.config.maxTokens,
-      temperature: this.config.temperature,
-      topP: this.config.topP,
+      maxTokens: this.config.maxTokens ?? getEnvInt('AWS_SAGEMAKER_MAX_TOKENS') ?? 1024,
+      temperature:
+        typeof this.config.temperature === 'number'
+          ? this.config.temperature
+          : (getEnvFloat('AWS_SAGEMAKER_TEMPERATURE') ?? 0.7),
+      topP:
+        typeof this.config.topP === 'number'
+          ? this.config.topP
+          : (getEnvFloat('AWS_SAGEMAKER_TOP_P') ?? 1.0),
+      stopSequences: this.config.stopSequences || [],
+      responsePath: this.config.responseFormat?.path,
       region: this.getRegion(),
     };
 
--- a/test/providers/sagemaker.test.ts
+++ b/test/providers/sagemaker.test.ts
@@ -23,10 +23,12 @@ vi.mock('../../src/cache', () => ({
 
 // Mock AWS SDK
 vi.mock('@aws-sdk/client-sagemaker-runtime', () => ({
-  SageMakerRuntimeClient: vi.fn().mockImplementation(() => ({
-    send: mockSend,
-  })),
-  InvokeEndpointCommand: vi.fn().mockImplementation((params) => params),
+  SageMakerRuntimeClient: vi.fn().mockImplementation(function () {
+    return { send: mockSend };
+  }),
+  InvokeEndpointCommand: vi.fn().mockImplementation(function (params) {
+    return params;
+  }),
 }));
 
 import {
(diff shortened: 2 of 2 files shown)


jameshiester-oai (commented):
## Review scope

- Base: `961dbd7eb14a9cb46ea5644a8afb64a054343d96`
- Head: `094d1c3b493cfc21ea8f06f2beffa797dbbb3d75`
- Scope: fresh full `base...head` review; no prior authenticated completion marker for this head.
- Verdict: one verified P2 cache correctness issue. All 43 listed CI checks pass; GitHub reports the branch mergeable but behind its base.

## Verified findings

### P2

1. The cache lookup, payload generation, and cache write independently read effective SageMaker settings. If an environment default changes while a request is pending, the response for the old payload is stored under the new setting's key. An isolated mocked-runtime reproduction changed `AWS_SAGEMAKER_MAX_TOKENS` from 128 to 256 during the first send; the next request returned the cached `128` output with just one send, rather than sending the `256` payload. Capture effective settings once for each call and derive both payload and lookup/write key from that immutable snapshot; test an in-flight setting change.

There was no actionable unresolved inline feedback to duplicate. The existing bot reviews did not cover this in-flight transition.

## Maintenance and coverage

The new tests cover settings changed between completed calls, zero values, response extraction, and errors, but not a change while `send` is pending. The duplicated effective-setting logic in `formatPayload` and `getCacheKey` can be consolidated as part of the correction. The key also includes settings ignored by some model types, which can cause extra cache misses; deriving identity from the effective payload would avoid those misses.

## Validation

- `npx vitest run test/providers/sagemaker.test.ts` — 21 passed, exit 0.
- `npm run tsc` — exit 0.
- `npm run build` — TypeScript, package bundles, and app build passed; the postbuild `tsx` IPC socket failed with local `EPERM` (exit 1). `node --import tsx scripts/postbuild.ts` then passed (exit 0), producing the built CLI.
- Isolated real provider with mocked SageMaker runtime and temporary cache: environment temperature `0.2` then `0.8` returned `0.2, 0.2, 0.8, 0.8`, with two sends and both replay flags set (exit 0).
- Isolated in-flight regression reproduction: `AWS_SAGEMAKER_MAX_TOKENS` changed `128` to `256` during `send`; second call replayed `128`, `cached: true`, with one send (exit 0; demonstrates finding).
- `git diff --check 961dbd7eb14a9cb46ea5644a8afb64a054343d96...HEAD` — exit 0; source worktree clean.

QA used local mocked AWS responses and isolated cache storage; no live SageMaker endpoint was invoked. The requested PR Review Toolkit launcher returned exit 127 and was unavailable; three bounded read-only specialist reviews covered correctness, tests/QA, and risk, then I independently verified the candidate finding.

src/providers/sagemaker.ts
```diff
@@ -647,12 +647,20 @@ export class SageMakerCompletionProvider extends SageMakerGenericProvider implem
...
     const configForKey = {
       endpoint: this.getEndpointName(),
-      modelType: this.config.modelType,
+      modelType: this.modelType,
       contentType: this.getContentType(),
       acceptType: this.getAcceptType(),
-      maxTokens: this.config.maxTokens,
-      temperature: this.config.temperature,
-      topP: this.config.topP,
+      maxTokens: this.config.maxTokens ?? getEnvInt('AWS_SAGEMAKER_MAX_TOKENS') ?? 1024,
```
jameshiester-oai, line 653: [P2] Keep the cache key tied to the payload sent by this call. The effective env settings are read here for lookup, again in `formatPayload`, and once more at the post-response cache write. If `AWS_SAGEMAKER_MAX_TOKENS` changes from 128 to 256 while `send` awaits, the 128 response is stored under the 256 key; my isolated mocked-runtime reproduction then returned cached 128 on the next call with only one endpoint send. Capture the effective settings once per call and use that snapshot for both payload and cache key; add an in-flight change regression test.

mldangelo-oai: Fixed in merged #10798 and #10808. Current main captures the payload and request settings before awaiting cache or network work, and reuses the same cache key for lookup and write. The regression cases change max_tokens during both cache lookup and the endpoint request, then verify that 128 and 256 responses remain in separate cache entries.

PR #10777, opened 2026-09-08
fix(providers): use OpenRouter reported costs

mldangelo-oai:
OpenRouter reports the account charge in [`usage.cost`](https://openrouter.ai/docs/cookbook/administration/usage-accounting). Use that amount instead of native OpenAI price estimates, which can misstate gateway charges. Missing or invalid billing data remains unknown; complete explicit per-token overrides retain priority, and cached responses cost zero.

Validated with 63 mocked provider tests, TypeScript and package builds, and 24 offline built-CLI cases covering reported costs, overrides, cache replay and expected HTTP errors. Lint, formatting and normal hooks passed. No vendor inference calls. Primary documentation checked September 8, 2026.

### DIFF
--- a/src/providers/openrouter.ts
+++ b/src/providers/openrouter.ts
@@ -3,12 +3,7 @@ import logger from '../logger';
 import { type GenAISpanContext, type GenAISpanResult, withGenAISpan } from '../tracing/genaiTracer';
 import { normalizeFinishReason } from '../util/finishReason';
 import { OpenAiChatCompletionProvider } from './openai/chat';
-import {
-  appendOpenAiApiPath,
-  calculateOpenAICost,
-  formatOpenAiError,
-  getTokenUsage,
-} from './openai/util';
+import { appendOpenAiApiPath, formatOpenAiError, getTokenUsage } from './openai/util';
 import { getRequestTimeoutMs } from './shared';
 import type OpenAI from 'openai';
 
@@ -19,6 +14,8 @@ import type {
   ProviderOptions,
   ProviderResponse,
 } from '../types/providers';
+import type { OpenAiChatCompletionCostData } from './openai/chat';
+import type { OpenAiCompletionOptions } from './openai/types';
 
 /**
  * OpenRouter provider extends OpenAI chat completion provider with special handling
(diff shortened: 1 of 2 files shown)


src/providers/openrouter.ts
```diff
@@ -70,6 +67,56 @@ export class OpenRouterProvider extends OpenAiChatCompletionProvider {
...
+      return Number.isFinite(cost) ? cost : undefined;
+    }
+
+    // usage.cost is the total charged to the OpenRouter account. Its upstream
+    // inference cost is a separate value; native vendor rates cannot substitute.
+    // https://openrouter.ai/docs/cookbook/administration/usage-accounting
+    const usage = data.usage as
+      | (NonNullable<OpenAiChatCompletionCostData['usage']> & { cost?: unknown })
+      | undefined;
+    const cost = usage?.cost;
```
jameshiester-oai, line 116: [P2] Distinguish BYOK upstream spend from the OpenRouter charge

`usage.cost` is the amount charged to the OpenRouter account, but a BYOK route bills inference to the user’s upstream provider account separately. When OpenRouter waives its BYOK fee or charges only that fee, this returns 0 or a fraction of the model spend; Promptfoo then aggregates it into a generic “Total Cost” and may hide the total entirely at zero. Please expose/label this as the OpenRouter charge or mark BYOK totals incomplete (or provide an explicit accounting mode). Adding `upstream_inference_cost` unconditionally would double count ordinary credit routes and may not match negotiated BYOK invoices. The route-specific response was not exercised with a live BYOK account, so this is conditional on that billing mode. OpenRouter documents the fields and billing paths: https://openrouter.ai/docs/cookbook/administration/usage-accounting and https://openrouter.ai/docs/guides/overview/auth/byok .

jameshiester-oai (commented):
Reviewed exact head `a35f21ad789ef0c8713060087a00f7bb1ae30434` against its PR base and current main. One conditional P2 cost-accounting issue is inline: BYOK routes can bill the upstream provider separately, while the new generic `cost` field reports only the OpenRouter account charge. The existing cache-test cleanup comment is separate and not repeated.

QA: 160 focused OpenRouter/OpenAI tests, TypeScript, tsdown/app and direct postbuild passed. A built public-provider localhost fixture exercised direct billed usage, a fallback route, explicit rates, and absent billing; outputs and costs matched expected cases. The ordinary postbuild `tsx` wrapper hit the local IPC permission error. Exact-head CI checks were green when inspected. The BYOK finding is source- and official-documentation-backed; no live BYOK account call was available, so its exact per-request response shape is not directly observed.

The requested `/pr-review-toolkit:review-pr all parallel` launcher was unavailable (exit 127); three independent read-only assessments covered correctness, tests, and cost risk.

test/providers/openrouter.test.ts
```diff
@@ -313,6 +313,139 @@ describe('OpenRouter', () => {
...
+          mockResponse({ prompt_tokens: 10, completion_tokens: 20, total_tokens: 30, cost: 0.25 });
+          const instance = provider(config);
+          const first = await instance.callApi('Cost cache prompt');
+          const second = await instance.callApi('Cost cache prompt');
+          expect(first.cached).toBe(false);
+          expect(second).toMatchObject({ cached: true, cost: 0, tokenUsage: { cached: 30 } });
+          expect(mockedFetchWithRetries).toHaveBeenCalledTimes(1);
+        } finally {
+          disableCache();
+        }
```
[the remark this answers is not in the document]
mldangelo-oai: Addressed in merged #10801 (24576f5cfe0b4c432fae93dc576d64c9e1dfc8fb). These tests now use `withCacheEnabled(true, ...)`, an async-context override, and assert the prior cache state after each scope. The file no longer enables or disables the global cache in setup or cleanup.

src/providers/openrouter.ts
mldangelo-oai: Addressed in #10801 at 24576f5. For a response explicitly marked BYOK, generic cost remains unknown unless complete manual rates provide an estimate. The response separately preserves the OpenRouter account charge, BYOK flag, and reported cost components; it does not add upstream spend unconditionally. The documentation explains that totals sum only known costs and records the export limitations. Offline BYOK, partial-rate, and cache-replay cases cover this behavior; they do not verify a live BYOK invoice.

PR #10778, opened 2026-09-08
fix(providers): recover gateway model discovery

mldangelo-oai:
A failed AIML API or CometAPI model-list request leaves an empty catalog cached until the process restarts. Expire successful discovery results after five minutes and retry errors after five seconds, with bounded account- and namespace-isolated entries and concurrent request coalescing. Preserve AIML aliases alongside canonical IDs and reject malformed catalog responses.

Validation: 65 focused provider tests, architecture check, root type check, backend/app builds, and normal pre-commit hooks passed. An actual built CLI evaluation exercised both compiled discovery helpers using synthetic HTTP responses; both exported cases passed with zero token usage.

### DIFF
--- a/src/providers/aimlapi.ts
+++ b/src/providers/aimlapi.ts
@@ -1,4 +1,4 @@
-import { fetchWithCache, getHeadersForCacheKey, getScopedCacheKey } from '../cache';
+import { fetchWithCache, getHeadersForCacheKey, getScopedCacheKey, isCacheEnabled } from '../cache';
 import { getEnvString } from '../envars';
 import logger from '../logger';
 import { createModelDiscoveryCache } from './modelDiscovery';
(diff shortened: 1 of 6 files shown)


jameshiester-oai (commented):
Reviewed live exact head `aa70f6dde1921b7e2f16b9772a19f6284c615fb8` against its PR base and current main. No new actionable findings. The AIML vendor API schema lists both `/models` and `/v1/models`, so the new route is documented; their live catalog contents were not compared from this network.

QA: aligned this head's dependencies with `npm ci --include=dev --ignore-scripts`; 65 focused discovery/AIML/Comet tests and TypeScript passed. A localhost fixture through the real fetchWithCache path confirmed AIML aliases, cached repeat under one identity, a 401 empty result with short cooldown then recovery, and Comet envelope parsing with separate auth identities. No vendor requests were made. Exact-head CI checks were green when inspected.

The requested `/pr-review-toolkit:review-pr all parallel` launcher was unavailable (exit 127); three independent read-only assessments covered correctness, tests, and risk.

PR #10793, opened 2026-09-08
fix(eval): propagate disabled cache to providers from other module instances

### LINKED ISSUE #10787: `--no-cache` is ignored when an extension hook imports from `promptfoo`
**Describe the bug**
If an `extensions` hook does `import { loadApiProvider } from "promptfoo"` and attaches the returned provider to the test, `promptfoo eval --no-cache` still returns cached responses. The token line says `(cached)` and the run takes 0s. Without the hook, `--no-cache` works.

**To Reproduce**
`promptfooconfig.yaml`:
```yaml
providers:
  - openai:chat:gpt-4o-mini

prompts:
  - "Say hello"

tests:
  - vars: {}

extensions:
  - file://hook.mjs:beforeEach
```
`hook.mjs`:
```js
import { loadApiProvider } from "promptfoo";

export async function beforeEach({ test }) {
  const provider = await loadApiProvider("openai:chat:gpt-4o-mini");
  return { test: { ...test, provider } };
}
```
```
$ promptfoo eval
  Eval: 18 (9 prompt, 9 completion)
Duration: 2s

$ promptfoo eval --no-cache
  Eval: 18 (cached)
Duration: 0s

$ PROMPTFOO_CACHE_ENABLED=false promptfoo eval
  Eval: 18 (9 prompt, 9 completion)
Duration: 2s
```
Control: delete the `extensions` block. `promptfoo eval --no-cache` then makes a real API call (`9 prompt, 9 completion`).

**Expected behavior**
`--no-cache` disables the cache for the whole run, same as `PROMPTFOO_CACHE_ENABLED=false`.

**Additional context**
Workaround: `PROMPTFOO_CACHE_ENABLED=false`.

he-yufeng:
## Summary

Fixes #10787: with `--no-cache`, an extension hook that imports `loadApiProvider` from `promptfoo` and attaches the provider to the test still served cached responses.

Root cause: the CLI flips the cache flag on its own copy of the cache module, but the hook's provider is built from the library build's copy — a different module instance whose flag is never flipped. The flag also initializes from `PROMPTFOO_CACHE_ENABLED` at import time, which is why the env-var workaround reaches both copies.

The fix routes the kill switch through the call context instead of module state: when the run's cache is disabled, `buildCallApiContext` sets `bustCache: true`. Providers honor it via `shouldBustCache(context)`, which reads plain call data and therefore holds regardless of which copy of the cache module the provider was built against.

Scope note: providers that consult `context.bustCache` (the OpenAI family, Azure, Ollama, etc.) are covered. A provider that reads only its own module's `isCacheEnabled()` and is created through a hook's import would still see its own flag; that residual is unchanged by this PR and no worse than today.

## Test plan

- New regression test in `test/evaluator/repeatCache.test.ts` captures the provider-visible call context on `runEval`: `bustCache` is `true` under `disableCache()` and falsy when enabled. Red without the change, green with it.
- Full `test/evaluator/` directory: 318 tests pass. `biome check` clean on both touched files; pre-commit hook passed.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -14,7 +14,7 @@ import {
   runCompareAssertion,
 } from './assertions/index';
 import { extractAndStoreBinaryData } from './blobs/extractor';
-import { getCache, withCacheNamespace } from './cache';
+import { getCache, isCacheEnabled, withCacheNamespace } from './cache';
 import cliState from './cliState';
 import { DEFAULT_MAX_CONCURRENCY, FILE_METADATA_KEY } from './constants';
 import { getEnvBool, getEnvInt, getEvalTimeoutMs, getMaxEvalTimeMs, isCI } from './envars';
@@ -1144,6 +1144,16 @@ function buildCallApiContext({
     repeatIndex,
     testIdx: testIndex,
   };
+
+  if (!isCacheEnabled()) {
+    // Cache was disabled for this run (e.g. --no-cache). The flag lives in
+    // this module's cache state, but the provider under test may have been
+    // built from a different copy of the package (an extension hook importing
+    // "promptfoo"), whose cache module never got flipped. bustCache travels
+    // with the call, so every provider honors it regardless of which copy of
+    // the cache module it reads.
+    callApiContext.bustCache = true;
+  }
 
   if (evalId) {
     callApiContext.evaluationId = evalId;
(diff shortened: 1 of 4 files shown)


he-yufeng:
CI caught a real gap in the first head: the two transform integration suites mock ../src/cache without isCacheEnabled, which the new context builder now imports. Fixed by mirroring the cache-enabled default those tests already assume (2c4281e); both suites green locally, 318-test evaluator directory and repeat-cache suite still green.

PR #10822, opened 2026-09-09
fix(eval): preserve audio grading results and order

mldangelo-oai:
Audio results at `--max-concurrency 1` bypassed grouped grading. A thrown grader error discarded the completed target response and usage, and graders on separate endpoints could run concurrently.

Keep audio rows in grouped grading and flush after each audio row. This preserves output, audio, cost, and tokens on grading errors or cancellation, serializes graders, and releases inline clips before the next target call. Clear completed batches so mixed audio/text evaluations process every row once.

Fixes two regressions from #10814.

## Validation

- All 35 focused evaluator tests pass, including exception, cancellation, serial dispatch, and mixed-row coverage. The first three regression cases fail before the fix.
- Built CLI with local mocks preserves output/audio, $0.25 cost, and 15 tokens after a grader exception. Separate grader endpoints peak at one concurrent request.
- Core build, TypeScript, architecture checks, lint, and formatting pass.
- Full backend suite: 24,878 passed, 10 skipped, 7 failed. The seven failures also reproduce on the unchanged baseline and come from the shared local installation having older `hono`, `js-yaml`, and `protobufjs` versions than the lockfile. A clean install was blocked by Socket's registry returning E403; dependencies are unchanged in this PR.

### DIFF
--- a/src/evaluator.ts
+++ b/src/evaluator.ts
@@ -1446,8 +1446,7 @@ async function gradeRunEvalResponse({
     providerTransformedOutput,
   };
 
-  // Finish audio grading per row instead of retaining every inline clip in the queue.
-  if (deferGrading && !response.audio?.data) {
+  if (deferGrading) {
     invariant(providerCallQueue, 'providerCallQueue is required when deferGrading is enabled');
     ret.response = processedResponse;
     const gradingPromise = withProviderCallExecutionContext(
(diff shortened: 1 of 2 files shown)


mldangelo-oai:
@codex review

Please review current head `a6a4d5487fd2724bb48abdde9511591dd8f05210`, particularly preservation of completed audio results on grader exceptions/cancellation and serial dispatch when flushing mixed audio/text batches.

jameshiester-oai (commented):
Reviewed exact head a6a4d5487fd2724bb48abdde9511591dd8f05210 against current base 39c4fb55: the two-file audio grading diff, deferred queue and persistence path, and all current comments/CI. Three independent read-only correctness, QA, and risk passes found no new actionable issue. Audio output is retained before deferred grading, the grouped queue flushes before another clip is collected, and completed batches are cleared to avoid double processing.

Validation: 18 focused evaluator assertion tests passed; exact-head tsdown and postbuild completed; diff check passed. A local built CLI eval with isolated config, no cache, concurrency 1, and a deterministic file target/grader produced ordered text/audio/text rows, 3/3 passed, score 1 each, no errors. Exported JSON retained the audio blobRef and removed inline audio data, with one row per case. All 43 current PR checks are green. The npm tsx wrapper hit a local IPC EPERM; direct node --import tsx completed the CLI run. The optional /pr-review-toolkit:review-pr launcher was unavailable (exit 127); direct source and QA review completed.

PR #10823, opened 2026-09-09
fix(providers): honor provider env over ambient process env

mldangelo:
## Summary

`GroqProvider` and `GroqResponsesProvider` each carried a `getApiKey()` override that resolved credentials as:

```ts
return this.config.apiKey || getEnvString(apiKeyEnvar) || this.env?.[apiKeyEnvar];
```

Process env was consulted **before** the provider/suite-level `env:` override — the reverse of the shared resolver `resolveProviderApiKey` (`src/providers/credentials.ts`) that every other provider goes through. Reproduced with an ambient `GROQ_API_KEY` in `process.env` plus a provider-level `env.GROQ_API_KEY`: Groq picked the ambient shell key while the equivalent OpenAI provider picked the provider-level one.

**This is a pure deletion.** Both Groq constructors already set `config.apiKeyEnvar = … || 'GROQ_API_KEY'`, and the base `OpenAiGenericProvider.getApiKey()` calls `resolveProviderApiKey(this.config, this.env, ['OPENAI_API_KEY'])`, which uses `config.apiKeyEnvar` when set — so removing the overrides restores the correct precedence and drops the now-unused `getEnvString` imports.

While in the neighbourhood, two sibling overrides that were byte-for-byte equivalent to the base implementation were deleted too (both already had the correct order, so no behavior change):

- `CometApiImageProvider.getApiKey()` — constructor already sets `apiKeyEnvar: 'COMETAPI_KEY'`.
- `AbliterationProvider.getApiKey()` — constructor already sets `apiKeyEnvar: 'ABLIT_KEY'`; the unused `EnvVarKey` import goes with it.
  The constructor default was `??`, so an explicit `apiKeyEnvar: ''` survived it and the base resolver — which reads an empty selector as *no* selection — would have reached for `OPENAI_API_KEY` and sent an unrelated OpenAI secret to `api.abliteration.ai`. Changed to `||` so a blank selector normalizes back to `ABLIT_KEY`, with a regression test in `test/providers/abliteration.test.ts`.

`helicone.ts` keeps its own resolver (it appends a `'placeholder-api-key'` fallback and does not use `apiKeyEnvar`), and `nscale.ts` / `cerebras.ts` never had an override. The `protected get apiKey()` accessors above the deleted Groq overrides are kept: they are the redaction hook used by `toJSON()` and match the sibling idiom in `deepseek.ts`, `xai/chat.ts`, and `hyperbolic/chat.ts`.

Net: **-33 / +10**, of which the source change is deletion only.

## Test plan

`test/providers/groq/connection-config.test.ts` contained a case (`preserves process precedence for the selected key variable`) that pinned the inverted behavior; it is flipped to assert the provider-scoped key wins, and a companion case covers the default `GROQ_API_KEY` envar.

Both new/changed assertions fail on the pre-fix source (verified by stashing the `src/providers/groq/*` changes):

```
Expected: "Bearer scoped-default-key"
Received: "Bearer groq-process-key"
 ❯ test/providers/groq/connection-config.test.ts:120:31
 Tests  4 failed | 16 passed (20)
```

Commands run:

```
npx vitest run test/providers/groq test/providers/inherited-auth-config.test.ts \
  test/providers/abliteration.test.ts test/providers/cometapi.test.ts test/providers/openai
  → Test Files 49 passed | 1 skipped, Tests 1950 passed | 9 skipped

npx vitest run test/providers/index.test.ts test/providers/registry.test.ts
  → Test Files 2 passed, Tests 357 passed

npx tsc --noEmit -p tsconfig.json          → clean
npx biome check --write <changed files>    → Checked 5 files, no fixes applied
```

### DIFF
--- a/src/providers/abliteration.ts
+++ b/src/providers/abliteration.ts
@@ -29,7 +29,9 @@ export class AbliterationProvider extends OpenAiChatCompletionProvider {
           normalizeApiBaseUrl(providerOptions.env?.ABLIT_API_BASE_URL) ??
           normalizeApiBaseUrl(getEnvString(ABLITERATION_API_BASE_URL_ENV_VAR)) ??
           ABLITERATION_API_BASE_URL,
-        apiKeyEnvar: providerOptions.config?.apiKeyEnvar ?? 'ABLIT_KEY',
+        // `||` (not `??`): an empty selector must fall back to ABLIT_KEY rather
+        // than leaving the base resolver to reach for OPENAI_API_KEY.
+        apiKeyEnvar: providerOptions.config?.apiKeyEnvar || 'ABLIT_KEY',
         showThinking: providerOptions.config?.showThinking ?? false,
       },
     });
--- a/src/providers/abliteration.ts
+++ b/src/providers/abliteration.ts
@@ -2,7 +2,6 @@ import { getEnvString } from '../envars';
 import { renderVarsInObject } from '../util/render';
 import { OpenAiChatCompletionProvider } from './openai/chat';
 
-import type { EnvVarKey } from '../envars';
 import type { EnvOverrides } from '../types/env';
 import type {
   ApiProvider,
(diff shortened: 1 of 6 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 0eeaa1491bb9f2f5b20244b5e19b6854d7b3c849 (provider-scoped API key precedence). The previous Abliteration blank-selector credential issue is fixed by defaulting to `ABLIT_KEY`. Shared resolution now selects the provider-scoped environment value before the process value; mocked Groq Chat/Responses, Abliteration, and CometAPI calls verify outgoing credentials. The focused suites pass 88/88 and all 43 reported checks are green. No distinct new issue found.

PR #10825, opened 2026-09-09
fix(providers): reject withdrawn bedrock models on all geos

mldangelo:
## Summary

The withdrawn-model guard in `getHandlerForModel` (`src/providers/bedrock/index.ts`) matched
the raw model ID against a hand-written list that only carried the bare and `us.` spellings.
Every other inference-profile spelling bypassed it: `eu.anthropic.claude-3-5-haiku-20241022-v1:0`
and its `apac.` / `global.` / `jp.` / `au.` variants fell through to the
`modelName.includes('anthropic.claude')` catch-all and failed later at request time with an
opaque AWS error — exactly the outcome the guard's own comment says it exists to prevent.

Changes:

- Match the list against the ID with its inference-profile prefix stripped, so every geo
  spelling is rejected with the same clear `Unknown Amazon Bedrock model: <id>` error.
- **Deleted** the three now-redundant hand-listed `us.` duplicates from the guard (and the
  two from the test fixture).
- Hoisted the `^(?:us|eu|apac|global|jp|au)\.` regex that already existed inline in
  `knowledgeBase.ts` into `bedrock/util.ts` as `INFERENCE_PROFILE_PREFIX` and reused it in
  both places instead of duplicating the literal.

Docs shipped IDs the guard hard-throws on, so a user copy-pasting them got
`Unknown Amazon Bedrock model`:

- `site/docs/providers/aws-bedrock.md` — the example provider list pointed at
  `bedrock:us.anthropic.claude-3-5-haiku-20241022-v1:0`; now `us.anthropic.claude-haiku-4-5-20251001-v1:0`,
  which is registered in `AWS_BEDROCK_MODELS`.
- `site/docs/providers/anthropic.md` — the cross-platform availability table listed Bedrock
  IDs for Claude 3.5 Haiku, Claude 3 Opus, and Claude 4 Opus, all three of which the guard
  rejects; they now read "Withdrawn from Bedrock", matching the table's existing
  "Not available" convention. The "Retired on the Anthropic API" preamble no longer claims
  those IDs "remain valid on AWS Bedrock".

Net: +39 / -19, with the behavior fix itself net-negative in `index.ts`.

## Test plan

Added `it.each(RETIRED_BEDROCK_MODEL_IDS)('rejects retired model id %s under every inference
profile prefix')` in `test/providers/bedrock/index.test.ts`, exercising `us.` / `eu.` /
`apac.` / `global.` / `jp.` / `au.`, and added the previously-uncovered
`anthropic.claude-3-5-haiku-20241022-v1:0` to that fixture.

```
# fails before the fix (source stashed, test kept):
$ npx vitest run test/providers/bedrock/index.test.ts -t "inference profile prefix"
  Tests  7 failed | 7 passed | 350 skipped
  FAIL … rejects retired model id anthropic.claude-3-5-haiku-20241022-v1:0 under every inference profile prefix
  AssertionError: expected [Function] to throw an error

# passes after:
$ npx vitest run test/providers/bedrock test/providers/bedrock.agents.test.ts
  Test Files  15 passed (15)
  Tests  995 passed (995)

$ npx vitest run test/providers/families
  Test Files  1 passed (1) | Tests  60 passed (60)

  (clean)

$ npx biome check src/providers/bedrock/{index,util,knowledgeBase}.ts test/providers/bedrock/index.test.ts
  Found 2 warnings  # both pre-existing noExcessiveCognitiveComplexity, unchanged from base

$ npx prettier --check site/docs/providers/anthropic.md site/docs/providers/aws-bedrock.md
  All matched files use Prettier code style!
```

### DIFF
--- a/site/docs/providers/anthropic.md
+++ b/site/docs/providers/anthropic.md
@@ -75,9 +75,10 @@ the same `not_found_error` a retired model returns.
 ### Retired on the Anthropic API
 
 These IDs return `404 not_found_error` from Anthropic. Promptfoo still accepts them — they
-remain valid on AWS Bedrock, GCP Vertex, and OpenAI-compatible gateways, which set their own
-lifecycle dates, and cost attribution for historical evals needs the rates — but a direct
-`anthropic:messages:` call will fail.
+remain valid on GCP Vertex and OpenAI-compatible gateways, which set their own lifecycle
+dates, and cost attribution for historical evals needs the rates — but a direct
+`anthropic:messages:` call will fail. AWS Bedrock has since withdrawn some of them; see
+[Cross-Platform Model Availability](#cross-platform-model-availability).
 
 | Model ID                     | Description            | Suggested replacement |
 | ---------------------------- | ---------------------- | --------------------- |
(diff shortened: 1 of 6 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 92744d79fa35da0d17efb91bb66af55e8c1d4a26 (withdrawn Bedrock models). The direct InvokeModel handler rejects a withdrawn geo-prefixed ID and accepts the documented Haiku 4.5 replacement. The focused Bedrock suite passes 364/364 and all 43 reported checks are green. No distinct new issue: existing inline comments already cover inference-profile ARNs, explicit Converse and knowledge-base routes, and the omitted GovCloud prefix. Provider loading is lazy; the guard applies when selecting the InvokeModel handler, not at load time.

PR #10826, opened 2026-09-09
fix(providers): correct three Bedrock pricing gaps

mldangelo:
## Summary

Three defects in `src/providers/bedrock/pricing.ts`:

1. **Nova 2 Lite `global.` profile billed at the regional rate.** The table entry's own comment
   documented that the cross-region global profile is cheaper ($0.30/$2.50) than the plain
   us-east-1 meter ($0.33/$2.75), but nothing acted on it — `global.amazon.nova-2-lite-v1:0`
   (a supported id in `index.ts`, and documented in `aws-bedrock.md`) was overcharged by 10%.
   Fixed by adding a `global.amazon.nova-2-lite` key ahead of the plain one, which the
   tables' existing first-match-wins substring matching picks up for both the bare profile
   id and the inference-profile ARN.

2. **`bedrock:converse:us.openai.gpt-5.6-*` had no pricing entry**, so `calculateBedrockCost`
   returned `undefined` and the Converse route reported no cost at all. Added the rates
   documented in `site/docs/providers/aws-bedrock.md` (Sol $4.40/$22, Terra $2.20/$13.20,
   Luna $0.22/$1.32 per 1M) — these are exactly the first-party rates in
   `src/providers/openai/util.ts` (4/20, 2/12, 0.2/1.2) plus the 10% Bedrock
   regional-processing uplift, which is the cross-check I used, plus the documented 272K
   long-context tier (2x input / 1.5x output). These models are deliberately left out of
   `BEDROCK_INVOKE_PRICING_MODEL_PREFIXES` since InvokeModel does not serve GPT-5.6.
   GPT-5.5 / GPT-5.4 are not added: AWS does not publish Bedrock rates for them and the docs
   only claim Converse support for 5.6, so guessing would be worse than the current
   `undefined`. Cache-read/write discounts for these models are also not modeled (the
   non-Claude/non-Nova path ignores cache tokens today) — out of scope here.

3. **Long-context threshold was inclusive.** The Sonnet 4 tier triggered at
   `totalInputTokens >= threshold`, while the shared rule (`src/providers/shared.ts`,
   `src/providers/azure/util.ts`, `src/providers/google/vertex.ts`) uses strict `>`. Exactly
   200,000 input tokens billed at the premium tier here and the standard tier everywhere else.
   Now `>`. There is no shared exported long-context helper to reuse — the rule is an inline
   expression inside three different cost functions with different surrounding shapes — so this
   is a one-character consistency fix rather than a new abstraction.

### Simplifications

- Collapsed the three byte-identical prefix-match loops in `getBedrockPricing` into a single
  `matchPricing` helper, and folded the region-table "found or fail-closed" logic into one
  branch. Net −18 lines in that function.
- Dropped the now-stale comment claiming the frontier GPT-5.x models are not available through
  Converse (the docs and the model card say GPT-5.6 is).

The per-region pricing tables were considered for consolidation and deliberately left alone:
they look near-identical but hold genuinely different AWS-published meters per region, so
merging them would either change values or need a lookup layer that costs more than it saves.

## Test plan

`test/providers/bedrock/pricing.test.ts` — six new/changed assertions, all verified failing on
the pre-fix source and passing after:

- `bills the Nova 2 Lite global profile at its own cheaper meter` (global vs `us.` profile)
- `prices us.openai.gpt-5.6-{sol,terra,luna}` and `applies the 272k long-context tier`
- `stays fail-closed on InvokeModel, which does not serve these models`
- `bills exactly 200k input tokens at the standard rate` (this replaces the old
  `switches to $6/$22.50 at and above 200k input tokens`, which pinned the `>=` behavior; the
  above-threshold case is now pinned separately at 200,001)

```
$ npx vitest run test/providers/bedrock/pricing.test.ts   # (pre-fix source) 6 failed | 71 passed
$ npx vitest run test/providers/bedrock/                  # 14 files, 973 passed
$ npx tsc --noEmit -p tsconfig.json                       # clean
$ npx biome check --write src/providers/bedrock/pricing.ts test/providers/bedrock/pricing.test.ts
```

### DIFF
--- a/src/providers/bedrock/pricing.ts
+++ b/src/providers/bedrock/pricing.ts
@@ -15,11 +15,13 @@ type BedrockPricing = {
   input: number;
   output: number;
   /**
-   * Rates that replace `input`/`output` once total input tokens reach `threshold`.
+   * Rates that replace `input`/`output` once total input tokens exceed `threshold`.
    * Cache rates are derived from the tier's input rate, matching how AWS prices the
    * `-cache-read/-cache-write-...-long-context-...` meters.
    */
   longContext?: { threshold: number; input: number; output: number };
+  /** Rates that replace `input`/`output` on the cross-region `global.` inference profile. */
+  global?: { input: number; output: number };
 };
 
 /**
@@ -93,9 +95,9 @@ const BEDROCK_PRICING: Record<string, BedrockPricing> = {
   'amazon.nova-lite': { input: 0.06, output: 0.24 },
   'amazon.nova-pro': { input: 0.8, output: 3.2 },
   'amazon.nova-premier': { input: 2.5, output: 12.5 },
-  // Amazon Nova 2 (reasoning models). The cross-region global profile is cheaper
-  // ($0.30/$2.50); this is the plain us-east-1 on-demand rate.
-  'amazon.nova-2-lite': { input: 0.33, output: 2.75 },
+  // Amazon Nova 2 (reasoning models). `input`/`output` are the plain us-east-1 on-demand
+  // rates; the cross-region global profile has its own cheaper meter.
+  'amazon.nova-2-lite': { input: 0.33, output: 2.75, global: { input: 0.3, output: 2.5 } },
   // Amazon Titan Text
   'amazon.titan-text-lite': { input: 0.15, output: 0.2 },
   'amazon.titan-text-express': { input: 0.2, output: 0.6 },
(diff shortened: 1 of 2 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed e55cc94f9b8c380e7096070ecc2e7eac8d273c41 (Bedrock pricing). The focused pricing suite passes 77/77, all 43 reported checks are green, and the new table/threshold branches have no distinct additional issue. The existing inline comments remain material: the default Nova 2 provider route bypasses the new price, and GPT-5.6 Converse cache-token counters lack the corresponding rates. Please verify each fix through provider-level cost assertions.

PR #10828, opened 2026-09-09
fix(providers): cache unknown watsonx model prices

mldangelo:
## Summary

`getModelCost` in `src/providers/watsonx.ts` cached only successful pricing lookups. Every other outcome returned `undefined` **uncached**:

- `!Array.isArray(resources)` — malformed response
- `!spec` — the model is not in the specs catalog
- the `catch` block, which carried the comment "Do not cache failures: a later response can retry after recovery."

So each eval row paid a fresh `client.listFoundationModelSpecs` call bounded by `getRequestTimeoutMs()`. With an unreachable or slow metadata endpoint that is one timeout per row. Worse, `!spec` is a **permanent** condition — a custom or private deployment that isn't in the catalog re-queried on every single row, forever, and never could succeed.

Fix: cache `WatsonXModelCost | undefined` under the existing per-client TTL. An unknown price is an answer like any other; the 5-minute TTL still lets a transient failure recover.

Simplification: the three early returns and the `catch`'s bespoke return are collapsed into a single exit path — one `cost` local, one cache write, one `return`. The cache-write block is no longer duplicated inside the happy path. Net +47/-27 across src and tests; the src function loses two branches.

The `WeakMap<WatsonXAIClient, …>` keying is untouched, so regions and accounts still never share prices.

## Test plan

Updated `test/providers/watsonx-cost.test.ts` — four cases previously asserted the un-cached behavior and now assert the cached one, each with an explicit TTL-expiry recovery so the retry path stays covered:

- metadata rejection → one call, both rows uncosted, refetch after the TTL
- deadline abort → one call, refetch after the TTL
- late SDK settlement after cancellation → still not written to the cache, refetch after the TTL
- missing / malformed / wrong-model metadata → one call, refetch after the TTL

```
$ npx vitest run test/providers/watsonx.test.ts test/providers/watsonx-preflight.test.ts test/providers/watsonx-cost.test.ts
Test Files  3 passed (3)      Tests  113 passed (113)

# same tests against the unfixed src (git stash push src/providers/watsonx.ts):
Tests  11 failed | 30 passed (41)

$ npx vitest run test/test-hygiene.test.ts
Test Files  1 passed (1)      Tests  115 passed (115)

$ npx biome check src/providers/watsonx.ts test/providers/watsonx-cost.test.ts   # clean
```

### DIFF
--- a/src/providers/watsonx.ts
+++ b/src/providers/watsonx.ts
@@ -240,7 +240,7 @@ function createWatsonXAuthCacheHash(authSelection: WatsonXAuthSelection): string
 // in memory per client so different regions/accounts never share model prices.
 let modelSpecsCache = new WeakMap<
   WatsonXAIClient,
-  Map<string, { expiresAt: number; cost: WatsonXModelCost }>
+  Map<string, { expiresAt: number; cost: WatsonXModelCost | undefined }>
 >();
 const MODEL_SPECS_CACHE_TTL_MS = 5 * 60 * 1000;
 
@@ -275,6 +275,7 @@ async function getModelCost(
       reject(new Error('WatsonX model pricing metadata request timed out'));
     }, getRequestTimeoutMs());
   });
+  let cost: WatsonXModelCost | undefined;
   try {
     const response = await Promise.race([
       client.listFoundationModelSpecs({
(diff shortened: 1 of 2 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 8017837cfb10c42eadb3b92d3da8bae89a366ad5 (WatsonX per-client model pricing cache). The focused WatsonX suites pass 113/113 and all 43 reported checks are green. I traced the text/chat call paths and cache expiry; I found no distinct issue beyond the existing inline comments about persisting transient pricing failures in response caches and concurrent lookups replacing a successful price. Those comments still warrant a response-cache-on recovery test and a concurrent-lookup test before merge.

PR #10829, opened 2026-09-09
fix(providers): translate chat prompts for azure responses

mldangelo:
## Summary

`AzureResponsesProvider` extends `AzureGenericProvider`, not `OpenAiResponsesProvider`, so it never inherited the chat-format content-part translation added in #10592 (`normalizeResponsesInput`, wired only into `openai/responses.ts` and `xai/responses.ts`). A multimodal prompt on `azure:responses:*` — the usual `{type: 'text'}` / `{type: 'image_url'}` chat parts — was forwarded raw and rejected by the API. `GroqResponsesProvider`, `BedrockOpenAiResponsesProvider` and `OpenClawResponsesProvider` all extend the OpenAI class and were unaffected.

Rather than adding a fourth copy of the call, the identical JSON-parse-then-normalize block that `openai/` and `xai/` had each copied is folded into one `parseResponsesInput()` in `src/providers/responses/input.ts`, and all three providers now build `input` through it — so the next Responses provider can't miss the translation. `normalizeResponsesInput` becomes internal to that module. Net **-15 lines in `src/`**.

Second fix, same file: cost was looked up from the request body's `model`, which for Azure is always the deployment name, so a deployment with an explicit `modelName` (`my-deployment` serving `gpt-5.6`) reported no cost at all. The processor is now keyed on `config.modelName ?? deploymentName`, and the body's `model` is only preferred when `passthrough` overrode it — matching the precedence in `azure/chat.ts`.

## Test plan

Two regression tests added in `test/providers/azure/responses.test.ts` (both verified failing on the pre-fix provider, `2 failed | 40 passed`):
- `translates chat-format content parts into Responses parts`
- `prices an arbitrarily named deployment from its configured modelName`

`test/providers/responses/input.test.ts` now exercises the same cases through `parseResponsesInput`, covering the parse path as well.

```
$ npx vitest run test/providers/openai test/providers/xai test/providers/azure test/providers/responses
Test Files  66 passed | 1 skipped
Tests  2768 passed | 9 skipped

$ npx biome check <changed files>       # 5 pre-existing complexity warnings, no errors
$ npx knip --no-progress                # clean
```

### DIFF
--- a/src/providers/azure/responses.ts
+++ b/src/providers/azure/responses.ts
@@ -10,6 +10,7 @@ import invariant from '../../util/invariant';
 import { FunctionCallbackHandler } from '../functionCallbackUtils';
 import { applyGpt6AstraRequestRules, isGpt6AstraModel } from '../openai/gpt6';
 import { ResponsesProcessor } from '../responses/index';
+import { parseResponsesInput } from '../responses/input';
 import { getRequestTimeoutMs, LONG_RUNNING_MODEL_TIMEOUT_MS } from '../shared';
 import { AzureGenericProvider } from './generic';
 import { calculateAzureCost } from './util';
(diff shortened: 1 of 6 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed df2556b388a452caac36af982ba6d2e52509cb82 (Azure Responses chat prompt translation). The focused suites pass 81/81 and all 43 reported checks are green. A local HTTP mock confirmed the deployment stays in the request URL/body, Chat text/image parts translate to Responses input parts, and configured model pricing is applied to the mocked usage. No distinct new issue. Existing inline comments cover assistant chat text parts being rewritten to output text and a row-level model override not reaching the response cost processor.

PR #10831, opened 2026-09-09
fix(providers): bill deepseek cached prompt tokens correctly

mldangelo:
## Summary

The DeepSeek cache-hit discount was dead code. `DeepSeekProvider.callApi` post-processed `super.callApi()`'s `ProviderResponse` looking for `response.raw`, but the base OpenAI chat provider never sets `raw` on its response (`grep -n "raw:" src/providers/openai/chat.ts` returns nothing). So `cachedTokens` was always `0`, and cached prompt tokens were billed at the full input rate. DeepSeek's cache-read rate is ~1/50th of the input rate (`0.0028` vs `0.14` per 1M for `deepseek-chat`), so reported cost was materially overstated for cache-heavy workloads.

MiniMax got the correct shape in #10745: read cached tokens from the raw `data.usage` inside the `calculateResponseCost` hook, where the usage payload is actually available. DeepSeek was left on the old broken shape.

This is **not a 0.122.2 regression** — the defect predates it — but it is a real cost-reporting error and the correct fix now exists in-repo, so it is worth landing.

Changes:

- Deleted the entire `callApi` override: two `response.raw` parsing branches, the `try`/`catch`, the `logger.debug`, and the manual `response.cost` assignment.
- Replaced it with a `calculateResponseCost` override that reads cached tokens via the shared `getOpenAICompletionTokenDetails(data.usage)?.cacheReadInputTokens`, exactly as `minimax.ts` does.
- Deleted the dead `private originalConfig` field (assigned in the constructor, never read).
- Taught `getOpenAICompletionTokenDetails` to read DeepSeek's top-level `usage.prompt_cache_hit_tokens`. DeepSeek does **not** send `prompt_tokens_details.cached_tokens` — [its documented usage schema](https://api-docs.deepseek.com/api/create-chat-completion) is `prompt_tokens`, `prompt_cache_hit_tokens`, `prompt_cache_miss_tokens`, `completion_tokens`, `total_tokens`; the nested `prompt_tokens_details.cached_tokens` shape is LiteLLM's normalization, not the wire format. Without this the hook still read a field DeepSeek never sends, so cached tokens stayed billed at the full input rate (and never appeared in `tokenUsage.completionDetails`). This sits next to the existing Moonshot top-level `cached_tokens` fallback.

Two incidental correctness improvements fall out of the hook: the cost now uses the merged config (so prompt-level `inputCost`/`outputCost`/`cacheReadCost` overrides apply — the old override read `this.config`, ignoring them), and promptfoo cache hits now report `cost: 0` instead of `undefined`, matching MiniMax and the base OpenAI billing path.

## Test plan

Added a regression block to `test/providers/deepseek.test.ts` that drives `callApi` through a mocked `fetchWithCache` with DeepSeek's real usage shape (`prompt_cache_hit_tokens` / `prompt_cache_miss_tokens`) and asserts the response is billed at the cache-read rate for 0 / 40 / 100 cached prompt tokens, plus a promptfoo-cache-hit case. Added a `getTokenUsage` case in `test/providers/openai/util.test.ts` for the same field.

```
$ npx vitest run test/providers/deepseek.test.ts test/providers/openai/util.test.ts
Test Files  2 passed (2) | Tests  232 passed (232)

# reverting only src/providers/deepseek.ts:      3 failed
# reverting only src/providers/openai/util.ts:   3 failed (40- and 100-cached-token billing, plus the util case)

$ npx vitest run test/providers/openai test/providers/deepseek.test.ts test/providers/moonshot.test.ts test/providers/minimax.test.ts
Test Files  45 passed | 1 skipped (46) | Tests  1929 passed | 9 skipped (1938)

$ npx tsc --noEmit -p tsconfig.json          # clean
$ npx biome check src/providers/openai/util.ts test/providers/openai/util.test.ts test/providers/deepseek.test.ts   # clean
```

### DIFF
--- a/src/providers/deepseek.ts
+++ b/src/providers/deepseek.ts
@@ -1,8 +1,10 @@
 import logger from '../logger';
 import { OpenAiChatCompletionProvider } from './openai/chat';
+import { getOpenAICompletionTokenDetails } from './openai/util';
 import { calculateCost, clampCachedTokens } from './shared';
 
 import type { ApiProvider, ProviderOptions } from '../types/index';
+import type { OpenAiChatCompletionCostData } from './openai/chat';
 import type { OpenAiCompletionOptions } from './openai/types';
 
 type DeepSeekConfig = OpenAiCompletionOptions;
@@ -97,8 +99,6 @@ export function calculateDeepSeekCost(
 }
 
 class DeepSeekProvider extends OpenAiChatCompletionProvider {
-  private originalConfig?: DeepSeekConfig;
-
   protected get apiKey(): string | undefined {
     return this.config?.apiKey;
   }
@@ -116,8 +116,6 @@ class DeepSeekProvider extends OpenAiChatCompletionProvider {
         apiBaseUrl: 'https://api.deepseek.com/v1',
       },
     });
-
-    this.originalConfig = deepseekConfig;
   }
 
   id(): string {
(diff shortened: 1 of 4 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 03256dd5db2c8d8aac21e8966c2aa48e9816f424 (DeepSeek cached prompt pricing). The earlier cache-hit field issue is resolved: the provider call test uses DeepSeek's top-level usage counter and covers 0/40/100 cached tokens and a Promptfoo response-cache replay. The focused suites pass 232/232 and 43 reported checks are green. The existing inline comment about billing `passthrough.model` using `this.modelName` still applies; no distinct new issue found.

PR #10835, opened 2026-09-09
fix(assertions): stop tool-call-f1 harvesting prose JSON

mldangelo:
## Summary

The embedded-JSON scanner added in #10789 fed **every** balanced JSON fragment found anywhere in a mixed text/JSON output into the tool-name extractor, whose array branch has a catch-all "simple format `{ name: '...' }`" rule. Ordinary JSON data appearing in prose was therefore harvested as a called tool, deflating precision and silently flipping passing evals to failures at the default threshold of 1.0:

| output | expected | before | after |
| --- | --- | --- | --- |
| `Users: [{"name": "Alice"}, {"name": "Bob"}].\n{"tool_calls":[…get_weather…]}` | `[get_weather]` | **0.500** (Called: Alice, Bob, get_weather) | 1.000 |
| `Config [{"name":"retries","value":3}] applied.\n{"tool_calls":[…get_weather…]}` | `[get_weather]` | **0.667** | 1.000 |

The same scanner restarted `findJsonEnd` at every `{`/`[`, making the scan O(braces × length). Measured on the old code: 60KB of braces (a model emitting code, or a truncated JSON blob) cost **~9 seconds of CPU per assertion**.

### What changed

`findJsonEnd` and the flat, top-level-only shape checks are gone, replaced by two smaller pieces:

- `jsonFragments` — one left-to-right pass with a delimiter stack that yields the outermost balanced fragments. Fragments inside an opener that never closes are still yielded, so the "stray brace in prose followed by a real tool call" case #10789 fixed keeps working. A quote only opens a JSON string once a delimiter is open, so prose quotes cannot swallow the JSON after them; a second pass ignoring string state covers prose that opens a delimiter *inside* quotes, and both passes' names are merged. A fragment that is not valid JSON falls back to the fragments one level inside it, which recovers a call the model wrapped in prose brackets. 60KB of braces now takes ~6ms; a 200KB adversarial `{` + `{}`×100000 takes ~28ms.
- `collectToolNames` — a depth-capped recursive walk that harvests only shapes that identify themselves as tool calls (`tool_use` / `function_call` items, `{ functionCall: { name } }`, `tool_calls` and `toolCall.functionCalls` lists). An arbitrary `{ name }` object is never mistaken for a tool call, and neither is an OpenAI tool *definition*: `{ function: { name } }` is the shape of both a call and a definition, so it only counts inside a list already known to hold calls.

Because the walk recurses, it also fixes a pre-existing gap: a tool-call envelope nested inside a wrapper is now found, where it previously scored 0 both before and after #10789.

| output | expected | before | after |
| --- | --- | --- | --- |
| `{"result":{"tool_calls":[…get_weather…]}}` | `[get_weather]` | **0** | 1.000 |
| `{"output":[{"type":"function_call","name":"get_weather"}]}` | `[get_weather]` | **0** | 1.000 |
| `{"tools":[{"type":"function","function":{"name":"get_weather","parameters":{…}}}]}` (a definition, nothing called) | `[get_weather]` | 0 | **0** |

The permissive "whole output is a bare list of `{ name }` calls" rule is kept, since there the entire provider output *is* the call list.

Docs updated with one sentence describing the wrapper traversal and the shapes that are ignored.

## Test plan

Added to `test/assertions/toolCallF1.test.ts`: three prose-contamination cases, four unbalanced-prose recovery cases (brace, quote, escaped quote, quoted brace), two "earlier call must not hide a later one" cases, a call whose arguments contain a brace after a prose quote, a call wrapped in prose brackets, a linear-time guard (`'{'.repeat(60_000)` under 1s), three nested-wrapper cases (object and JSON-string forms), a tool-definition case, and a deep-nesting case that must grade rather than overflow.

- Each new assertion confirmed failing on the implementation it fixes, and passing after.
- `npx vitest run test/assertions/toolCallF1.test.ts test/assertions/toolCallF1.responses.test.ts` → **88 passed**, every pre-existing case still green (bare Responses `function_call` item, pretty-printed multiline block in prose, single-line JSON call in prose, Anthropic/Google shapes).
- `npx vitest run test/assertions/` → **1447 passed** (62 files).
- `npx tsc --noEmit -p tsconfig.json` → clean.
- `npx biome check` on the changed files, `npx prettier --write` on the doc → clean.

### DIFF
--- a/src/assertions/toolCallF1.ts
+++ b/src/assertions/toolCallF1.ts
@@ -2,13 +2,21 @@ import invariant from '../util/invariant';
 
 import type { AssertionParams, GradingResult } from '../types/index';
 
-function findJsonEnd(text: string, start: number): number {
-  const closing: string[] = [text[start] === '{' ? '}' : ']'];
+/**
+ * Outermost balanced `{...}` / `[...]` fragments of a string, found in a single pass.
+ * Fragments inside an opener that never closes are returned too, so a stray brace in
+ * prose cannot hide a tool call that follows it.
+ */
+function jsonFragments(text: string): string[] {
+  // Each open delimiter collects the spans that completed directly inside it. They are
+  // dropped when it closes, because the enclosing fragment already covers them.
+  const open: { start: number; nested: [number, number][] }[] = [];
+  const spans: [number, number][] = [];
   let inString = false;
   let escaped = false;
 
-  for (let end = start + 1; end < text.length; end++) {
-    const char = text[end];
+  for (let i = 0; i < text.length; i++) {
+    const char = text[i];
     if (inString) {
       if (escaped) {
         escaped = false;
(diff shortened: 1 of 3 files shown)


mldangelo:
@codex review

mldangelo:
Worked through every review finding on this PR. Four were real regressions against `main` and are fixed in 27c310d; three do not hold and are refuted below with the evidence.

I checked each case against both implementations side by side (`main`'s `39c4fb5` copy of the file vs. this branch), so "before/after" here means main vs. this branch, not this branch vs. #10789.

## Fixed

**1. Later call lost after an unbalanced quote when an earlier call was found** (Copilot, and Codex P2 "Retry scanning after finding only earlier calls")

Real. Repro:

```
{"type":"tool_use","name":"get_weather"}
He said "hello.
{"type":"tool_use","name":"get_time"}
```

expected `[get_weather, get_time]` — main 1.000, this branch **0.667** (`Called: [get_weather]`). The prose quote put the string-aware pass into string mode, and the `names.size > 0` gate then skipped the quote-insensitive retry.

Fixed two ways: a quote now only opens a JSON string once a delimiter is open (prose quotes outside JSON are just text), and the two passes always both run with their names merged, so an earlier call can never suppress a later one. Regression test covers both the plain quote and the `A "{" appears` variant that still needs the second pass.

**2. Prose quote masks a call whose arguments contain a brace** (Codex P2 "Preserve JSON string boundaries during the fallback scan")

Real. `He said "hello.\n{"type":"tool_use","name":"get_weather","input":{"query":"{foo"}}` — main 1.000, this branch **0.000**. The quote-insensitive fallback treated the `{` inside `"{foo"` as an opener. Same fix as above: the primary pass no longer enters string mode on the prose quote, so it tokenizes the call correctly and never reaches the fallback.

**3. Call wrapped in prose brackets** (Codex P2 "Retry nested fragments when the outer span is not JSON")

Real. `Details [tool follows: {"type":"tool_use","name":"get_weather"}]` — main 1.000, this branch **0.000**. A fragment that fails `JSON.parse` now falls back to the fragments one level inside it. One level only: retrying every level makes the scan quadratic again, which is what this PR set out to fix.

**4. Tool definitions counted as calls** (Codex P1 "Require call context before accepting function definitions")

Real, and a precision bug of exactly the kind this PR is about. `{"tools":[{"type":"function","function":{"name":"get_weather","parameters":{…}}}]}` — main 0.000 (correct: nothing was called), this branch **1.000**. `{ function: { name } }` is the shape of both an OpenAI tool call and an OpenAI tool *definition*, so it is now only read inside a list already known to hold calls (`tool_calls`, `functionCalls`, or a top-level array of calls). Every supported call shape keeps working; `functionCall` and the typed `tool_use` / `function_call` items are unambiguous and still recognised anywhere.

**5. Deep `output` objects** (the deep half of Copilot's traversal comment)

Partly real. `collectToolNames` overflowed at roughly 4,000 levels of nesting, while the `JSON.stringify` that `runAssertion` performs upstream survives to roughly 5,900 in the same process — a narrow band where the walk, not the pipeline, is what breaks. It now takes a depth cap and returns a score instead of throwing.

## Refuted

**Cyclic `output` reaching `collectToolNames`** (Copilot: "a cyclic output object … can cause infinite recursion")

Does not hold. `runAssertion` builds `assertionParams` with `outputString: coerceString(output)` (`src/assertions/index.ts:640`), and `coerceString` is `JSON.stringify` for non-strings (`src/assertions/utils.ts:73-78`). That property is evaluated before the handler is dispatched, so a cyclic output dies there with `TypeError: Converting circular structure to JSON` and never reaches this assertion. Verified directly:

```
coerceString({ tool_calls: […], self: <cycle> })
→ TypeError: Converting circular structure to JSON
```

Also, a plain recursive walk cannot loop on a cycle — it overflows the stack and throws. The iterative-walk-plus-visited-set rewrite would swap a throw for a hang if that gate ever changed, so the depth cap above is the guard that was actually needed.

**`docs(site):` scope required for the doc line** (Codex P1)

Does not hold for a mixed PR. `site/AGENTS.md` sets the scope for site changes; a one-sentence doc update describing the behavior change in the same commit is how this repo already lands them — for example `5f0324f fix(assertions): parse multiline tool calls and clarify guides (#10789)`, the very PR this one follows up, and `39c4fb5`, `c85b395`, `71346b0` immediately before it. Splitting the sentence that documents this fix into its own PR would leave the docs wrong between merges.

**`expected-outputs/index.md` must be updated** (Codex P1)

Does not hold. That file's `tool-call-f1` row is a one-line index entry — "Tool-call F1 score meets the threshold" (`index.md:152`) — that stays accurate; the supported and ignored shapes belong in `deterministic.md`, which this PR updates (and extends here to mention that tool definitions are ignored).

## Verification

- 6 new regression tests, each confirmed failing on the previous commit and passing after the fix.
- `npx vitest run test/assertions/` → 1447 passed (62 files); `npx tsc --noEmit` clean; Biome/Prettier clean.
- Linearity re-checked with the new fallback in place: 60KB of open braces 6ms, 60KB of balanced nested braces 7ms, 120KB of nested almost-valid JSON 9ms, a 200KB `{` + `{}`×100000 bomb 28ms.
- Precision spot-checks still harvest nothing from prose: an escaped call inside a JSON string value, `[{"name": "Alice"}]` records, a `{"name":"retries"}` config entry, a markdown table with braces, and a `tools` definition list.

src/assertions/toolCallF1.ts
```diff
@@ -17,188 +30,153 @@ function findJsonEnd(text: string, start: number): number {
...
+}
 
-  // Handle Anthropic single tool_use block: { type: 'tool_use', name: '...' }
-  if (obj.type === 'tool_use' && typeof obj.name === 'string') {
-    names.add(obj.name);
-    return names;
+/** Parses one balanced fragment and harvests its calls; false when it is not JSON. */
+function addFragmentNames(fragment: string, names: Set<string>): boolean {
+  try {
+    collectToolNames(JSON.parse(fragment), names);
```
jameshiester-oai, line 140: [P2] Extract call-list names from parsed arrays embedded in prose. This helper invokes `collectToolNames` but omits the top-level `addCallListNames` step used by `extractToolNames`. At this head, `[ { function: { name: 'get_weather' } } ]` and its bare JSON string both score F1=1, while `Tool calls: [{"function":{"name":"get_weather","arguments":"{}"}}]` scores F1=0 with `Called: [(none)]`. The previous mixed-text scanner accepted that embedded list. Apply the array-specific call-list extraction after parsing a fragment and add this prose-wrapped OpenAI call regression case.

jameshiester-oai (commented):
Reviewed 27c310dee9534ba89d98c89360605d4e020aeeda (tool-call-f1 extraction). The focused assertion/Responses suites pass 88/88, the isolated local CLI evaluation passes its prose JSON case, and all 43 reported checks are green. One distinct regression remains in embedded JSON arrays (inline). Existing comments separately cover argument payloads counted as calls, the depth cap, and malformed prose delimiters; I did not duplicate those.

PR #10837, opened 2026-09-09
fix(assertions): stop silently mis-grading native audio

mldangelo:
## Summary

`llm-rubric` audio grading (#10814) could return a confident pass/fail without the grader ever hearing the audio.

**Root cause.** `buildAudioGradingPart` attaches the clip only when the grading provider implements the audio-input capability hook, and `OpenAiChatCompletionProvider` is the only class that does. With any other grader — Anthropic, Google, Bedrock, Azure, or an OpenAI **Responses** provider (which is what a bare `openai:gpt-5.6+` now resolves to after b051efdc6a) — the audio part was dropped with no warning. Meanwhile `getGradingOutputForAudio` had already replaced the base64 output with `audio.transcript || '[Audio output]'`, so a transcript-less audio target got graded against the literal string `[Audio output]`. Silently wrong, not an error.

Second bug: audio grading requests never asked for text, so a `gpt-audio` grader defaulted to `modalities: ['text', 'audio']` and spent its completion budget speaking the verdict.

**Changes.**

- Audio evidence now has to reach the grader. A grader that cannot listen falls back to the output's text with a `logger.warn`, and errors when that text is only the `[Audio output]` placeholder — i.e. the output *is* the clip and there is no transcript — instead of grading the placeholder. A target that returns real text alongside a clip (`openai:tts`, `elevenlabs:tts`, a custom provider) still grades that text. The remote-grading path, which can never attach audio, goes through the same check.
- Audio grading requests `modalities: ['text']` via the existing `Prompt.config` merge (`callProviderWithContext` gained one optional argument) — no provider cloning or mutation, and it is exactly what the docs previously told users to configure by hand.

**Simplified.**

- `getAudioInputFormat(): 'openai' | undefined` → `supportsAudioInput(): boolean`. One implementer and one format made the string return value and the `!== 'openai'` comparison dead weight.
- Docs drop the now-redundant `config: { modalities: [text] }` from both grader examples, and the "text-only graders retain their existing behavior" sentence is replaced with the actual warn/fail contract.

## Test plan

New regression cases in `test/assertions/llmRubric.audio.test.ts` (7 cases across 4 tests, verified against `src/` from `main`):

- `fails instead of grading an audio placeholder` — text grader + base64 output + no transcript now rejects and never contacts the grader (previously graded `[Audio output]` and passed). Fails on `main`.
- `warns when a text grader drops the audio` — text fallback is logged, for both `gpt-4.1` (Chat) and `gpt-5.6` (Responses). Fails on `main`.
- `requests a text-only grade so the audio grader does not answer in speech` — body carries `modalities: ['text']` with no grader-side config. Fails on `main`.
- `grades the text output of a transcript-less target` — text grader + text output + transcript-less clip still grades the text. Passes on `main`; it pins the behavior this PR must not break.

```
npx vitest run test/assertions/llmRubric.audio.test.ts    # 29 passed (22 before)
git checkout main -- src/ && npx vitest run …             # 5 failed | 24 passed
npx vitest run test/matchers test/assertions test/evaluator/assertions.test.ts test/providers/openai/chat.test.ts  # 1897 passed
npx tsc --noEmit -p tsconfig.json                         # clean
npx biome check <changed files>                           # clean
npx prettier --check <changed .md>                        # clean
npm run knip -- --no-progress                             # clean
npm run architecture:check                                # boundaries passed
```

### DIFF
--- a/src/matchers/rubric.ts
+++ b/src/matchers/rubric.ts
@@ -38,6 +38,8 @@ const DEFAULT_GRADING_IMAGE_MAX_RAW_CHARS =
 const DEFAULT_GRADING_IMAGE_MAX_TOTAL_RAW_CHARS =
   Math.ceil(DEFAULT_GRADING_IMAGE_MAX_TOTAL_BYTES / 3) * 4 +
   DEFAULT_GRADING_MAX_IMAGES * DATA_URI_METADATA_MAX_CHARS;
+/** Stands in for an audio-only output, which has no text a grader could read. */
+export const ATTACHED_AUDIO_OUTPUT_PLACEHOLDER = '[Audio output]';
 const MULTIMODAL_GRADING_INSTRUCTION =
   'The evaluated output includes the attached image(s). Treat the attached image(s) as primary evidence in <Output>. Inspect the visual content directly, and do not infer visual traits, demographics, safety issues, or rubric failures from the user prompt or from any base64/data URI text.';
 const BLOB_HASH_REGEX = /^[a-f0-9]{64}$/i;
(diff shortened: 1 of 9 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed d45f8c7d78785329847e4a75cf4d893efc20277f (native audio rubric grading). The focused audio rubric suite passes 29/29 and all 43 reported checks are green. I found one distinct false-pass path in the new missing-evidence guard (inline). Existing comments separately cover custom grader hook compatibility and transcript fallback; those remain applicable. The local probe used a text-only grader and an audio response with empty output/no transcript, and the grader received no audio yet returned a passing verdict.

src/matchers/rubric.ts
```diff
@@ -709,11 +711,28 @@ function appendMediaToChatPrompt(
...
 }
 
+/**
+ * A grader that cannot hear the clip grades whatever text the output carries.
+ * When that text is only the placeholder — the output was the audio itself and
+ * there is no transcript — there is nothing to grade, so fail instead of
+ * returning a confident, meaningless verdict.
+ */
+export function requireAudioGradingEvidence(gradedOutput: unknown, grader: string): void {
+  if (gradedOutput === ATTACHED_AUDIO_OUTPUT_PLACEHOLDER) {
```
jameshiester-oai, line 721: [P1] Reject empty audio-only output before text-only grading. When a target returns `{ output: '', audio: { data: <valid base64>, format: 'wav' } }` without a transcript, `getGradingOutputForAudio` leaves the empty string unchanged, so this exact-placeholder comparison passes. The non-audio grader then receives an empty `<Output>` and no clip; a current-head local probe returned `{ pass: true, score: 1 }`. The evaluator accepts `output: ''`, so this can silently pass an audio rubric with no evidence. Treat empty/whitespace text plus audio and no transcript as missing evidence, and cover the evaluator/rubric path.

PR #10839, opened 2026-09-09
fix(n8n): send HEAD webhook payloads as query params

mldangelo:
## Summary

A `method: HEAD` n8n webhook sent **neither a body nor query parameters**, so the rendered prompt was silently dropped and the target workflow received nothing at all.

Root cause: [`7f897cfcc4`](https://github.com/promptfoo/promptfoo/commit/7f897cfcc4) (#10730) added `HEAD` to the body-omission check in `callApi` but not to the URL-building branch one line above:

```ts
const url = method === 'GET' ? this.buildGetUrl(body) : this.getUrl();  // HEAD missed
...
if (method !== 'GET' && method !== 'HEAD') { fetchOptions.body = renderedBody; }
```

`HEAD` is a bodyless method, so the payload has nowhere to go except the query string.

**What changed (net -1 line in `src/providers/n8n.ts`):**

- Collapsed the two duplicated method string comparisons into one `isBodyless` predicate that drives both the URL and the body — deleting the standalone `if` block and the now-inlinable `renderedBody` local.
- Renamed `buildGetUrl` → `buildQueryUrl`, since it now serves both bodyless verbs.
- Left the `isIdempotent` retry-policy list at the bottom of `callApi` alone: idempotent ≠ bodyless (`PUT`/`DELETE` are idempotent but do carry bodies), so conflating them would be a regression.
- Updated the `method` row in the provider docs.

## Test plan

The existing HEAD test encoded the bug (it asserted the bare URL); it now asserts the prompt is carried in the query string, and fails on the pre-fix source.

```
# regression check: stash the src fix, run tests
$ git stash push -- src/providers/n8n.ts && npx vitest run test/providers/n8n.test.ts
  Tests  3 failed | 53 passed (56)
  → expected 'https://n8n.example.com/webhook/agent?prompt=Hello'

# with the fix applied
$ npx vitest run test/providers/n8n.test.ts
  Test Files  1 passed (1)
  Tests  56 passed (56)

$ npx tsc --noEmit -p tsconfig.json          # clean
$ npx biome check src/providers/n8n.ts test/providers/n8n.test.ts   # clean
$ npx prettier --write site/docs/providers/n8n.md                   # clean
```

### DIFF
--- a/site/docs/providers/n8n.md
+++ b/site/docs/providers/n8n.md
@@ -49,16 +49,16 @@ providers:
 
 ### Config Options
 
-| Option              | Type          | Default     | Description                                                                                                      |
-| ------------------- | ------------- | ----------- | ---------------------------------------------------------------------------------------------------------------- |
-| `url`               | string        | -           | Webhook URL (alternative to provider path)                                                                       |
-| `method`            | string        | `POST`      | `GET`, `HEAD`, `POST`, `PUT`, or `PATCH`; `GET` encodes body fields as query parameters and `HEAD` sends no body |
-| `headers`           | object        | -           | Additional request headers with Nunjucks templating                                                              |
-| `body`              | object/string | `{prompt}`  | Request/body-query template; object form is recommended for JSON requests                                        |
-| `transformResponse` | string        | -           | JavaScript expression to extract output                                                                          |
-| `sessionHeader`     | string        | -           | Request header name for the session ID                                                                           |
-| `sessionParser`     | string        | -           | JavaScript expression to extract a session ID                                                                    |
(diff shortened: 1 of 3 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 2cc8a37b1839f6e8a99cc87582d001a43b7b443b (n8n HEAD webhook payloads). The focused n8n suite passes 56/56 and all 43 reported checks are green. Against a local HTTP server, HEAD and GET both preserved an existing query parameter, sent the rendered prompt/count as encoded query fields, and had no request body. No distinct new issue; the existing inline comment still applies to malformed HEAD URLs throwing before the provider's error boundary.

PR #10840, opened 2026-09-09
fix(providers): resolve azure model for guards and cost

mldangelo:
## Summary

Azure separates the **connection** (the deployment name that goes in the URL) from the **model** the deployment actually serves. `src/providers/azure/chat.ts` read those two apart in three inconsistent, hand-rolled ways, producing three bugs:

1. **Grok 4+ guard ignored the model.** `isGrok4OrNewerModel()` tested only `this.deploymentName`, so a deployment named e.g. `prod-chat` backed by `grok-4.6` (via `modelName` or `passthrough.model`) still sent `presence_penalty`, `frequency_penalty` and `stop` — the three parameters Grok 4+ rejects with HTTP 400.
2. **`passthrough.model` was lower-cased** into `config.modelName`, and `AZURE_MODELS` is keyed on the vendor's own ids. Every mixed-case model (`DeepSeek-R1`, `Phi-4`, `Meta-Llama-3-8B-Instruct`, `Kimi-K2.5`, …) therefore missed the cost table and reported `cost: undefined`.
3. **No `metadata.toolCalls` for MCP.** `openai:chat` and `anthropic:messages` publish executed MCP tool calls as `metadata.toolCalls`; `azure:chat` runs MCP through `FunctionCallbackHandler` and published nothing, so Azure MCP evals could not assert on tool routing.

### What changed (shared, not three more Azure branches)

- **New `resolveAzureModelName(config, deploymentName)` in `src/providers/azure/util.ts`** — one precedence rule (`passthrough.model` > `modelName` > deployment name) with case preserved. It replaces three divergent inline copies in `chat.ts` and now backs `isReasoningModel`, `isGrok4OrNewerModel`, `isSamplingParamsDeprecatedClaudeModel`, `isForcedToolChoiceUnsupportedClaudeModel`, the GPT-6-Astra rules, and `calculateAzureCost`. `getOpenAiBody` returns the resolved `modelName` unconditionally, so the conditional-config branch is gone.
  - `src/providers/azure/responses.ts` has the identical inline block; it is untouched here because a sibling PR owns that file, but it can drop its copy for this helper (both files already import `./util`).
- **MCP tool calls are recorded by the shared handler.** `FunctionCallbackHandler.processCall(s)` takes an optional `{ toolCalls }` sink — replacing the dead, never-passed `_options: { returnRawOnError }` parameter — and `executeMcpTool` pushes an `McpToolCallEntry` (same shape openai/anthropic emit). `azure/chat.ts` passes the sink and publishes `metadata.toolCalls` only when a tool actually ran.
- **Deleted the duplicated `normalizeContent`.** `openai/chat.ts` and `functionCallbackUtils.ts` carried byte-identical 30-line copies; hoisted to `normalizeMcpContent` in `src/providers/mcp/util.ts` and both now import it.

Net: +247 / −114 across 8 files, of which the source change is +145 / −108 with two duplicated blocks and one dead parameter removed.

## Test plan

All four new tests fail on `main` and pass with the fix (verified by stashing `src/` and re-running):

```
$ git stash push -- src/ && npx vitest run test/providers/azure.test.ts test/providers/azure/chat.test.ts test/providers/azure/chat-mcp-integration.test.ts
Tests  4 failed | 136 passed (140)
```

- `test/providers/azure.test.ts` — "strips them for a deployment aliased by $label" (`modelName` and `passthrough.model` cases) asserts `presence_penalty` / `frequency_penalty` / `stop` are omitted for a Grok 4.6 model behind a `prod-chat` deployment.
- `test/providers/azure/chat.test.ts` — "prices a deployment by its passthrough.model, not its deployment name" asserts a non-undefined `DeepSeek-R1` cost for a `prod-chat` deployment and that `body.model` keeps its case.
- `test/providers/azure/chat-mcp-integration.test.ts` — asserts `metadata.toolCalls` carries the executed call (id, name, parsed input, normalized output, `is_error: false`), and that `metadata` stays undefined when no MCP tool ran.

Commands run:

```
npx vitest run test/providers test/test-hygiene.test.ts
  → Test Files 243 passed | 1 skipped, Tests 10054 passed | 9 skipped
npx tsc --noEmit -p tsconfig.json                → clean
npx tsc --noEmit -p src/app/tsconfig.json        → clean
npx biome check <changed files>                  → only the 4 pre-existing
   noExcessiveCognitiveComplexity warnings (5 on main, 4 here)
npx knip --no-progress                           → clean
```

### Follow-up: case-insensitive cost matching

Preserving the configured casing prices a model only when the user types the vendor's exact
id, so `passthrough: { model: 'deepseek-r1' }` still fell out of `AZURE_MODELS`.
`calculateAzureCost` now matches the table case-insensitively and keys the cache-read and
priority rate maps off the matched entry's `id`, which fixes the same class of
`cost: undefined` on every Azure path (chat, responses, foundry-agent), not just
`passthrough.model`. Verified byte-identical costs for all 218 `AZURE_MODELS` entries across
standard/priority tiers and short/long context; regression test in
`test/providers/azure/util.test.ts` ("matches model ids case-insensitively").

### Follow-up: MCP tool calls recorded in request order

`processCalls` runs the calls concurrently, so the shared `toolCalls` sink filled
`metadata.toolCalls` by completion latency rather than by the model's `tool_calls` order —
nondeterministic for index-based assertions and inconsistent with `openai:chat`, which
executes sequentially. Each call now gets its own bucket that is drained in order.
Regression test in `test/providers/functionCallbackUtils.test.ts` ("records concurrent MCP
tool calls in the order the model requested them"), which fails with the shared sink.

### DIFF
--- a/src/providers/azure/chat.ts
+++ b/src/providers/azure/chat.ts
@@ -26,13 +26,14 @@ import { applyGpt6AstraRequestRules, isGpt6AstraModel } from '../openai/gpt6';
 import { getRequestTimeoutMs, parseChatPrompt, transformTools } from '../shared';
 import { DEFAULT_AZURE_API_VERSION } from './defaults';
 import { AzureGenericProvider } from './generic';
-import { calculateAzureCost } from './util';
+import { calculateAzureCost, resolveAzureModelName } from './util';
 
 import type {
   CallApiContextParams,
   CallApiOptionsParams,
   ProviderResponse,
 } from '../../types/index';
+import type { McpToolCallEntry } from '../mcp/types';
 import type { AzureChatResponsesOptions, AzureProviderOptions } from './types';
 
 export class AzureChatCompletionProvider extends AzureGenericProvider {
@@ -77,7 +78,9 @@ export class AzureChatCompletionProvider extends AzureGenericProvider {
    * Reasoning models use max_completion_tokens instead of max_tokens,
    * don't support temperature, and accept reasoning_effort parameter.
    */
-  protected isReasoningModel(modelName = this.config.modelName ?? this.deploymentName): boolean {
+  protected isReasoningModel(
+    modelName = resolveAzureModelName(this.config, this.deploymentName),
+  ): boolean {
     // Check explicit config flags first
     if (this.config.isReasoningModel || this.config.o1) {
       return true;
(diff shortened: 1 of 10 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 10dca9437005e951856c69a14d7e9d1b835f5496 (Azure model resolution and MCP metadata). Focused provider suites pass 378/378 and all 43 reported checks are green. Mocked requests cover aliased Grok restrictions, DeepSeek-R1 cost, and Azure MCP metadata; concurrent call metadata is now recorded in request order. No distinct new issue. The existing inline comment about a parsed JSON `null` being recorded as the raw string in tool metadata remains applicable.

PR #10841, opened 2026-09-09
fix(providers): keep distinct MCP servers distinct for Claude Agent SDK

mldangelo:
## Summary

`transformMCPServerConfigToClaudeCode` derived each MCP server's key as
`config.name ?? config.url ?? config.command ?? 'default'`, and `Object.fromEntries`
silently kept only the last entry when two servers landed on the same key. The eval
then ran against fewer MCP servers than configured, with no warning:

- two unnamed path-based servers both keyed to `'default'`;
- two unnamed servers sharing a `command` (e.g. `npx` with different `args`) both keyed to `npx`;
- two unnamed servers on the same `url` with different query-placed API keys.

**Fix:** keep that same coarse key, and disambiguate a collision with a numeric
suffix (`npx`, `npx_2`) instead of dropping a server.

The key is not just a map key: the Claude Agent SDK uses it as the server name, so it
appears in tool names (`mcp__<key>__<tool>`) and is logged verbatim by the provider
(`Object.keys(options.mcpServers)` in `src/providers/claude-agent-sdk.ts`). It
therefore stays deliberately coarse — folding `args` into it would both break
`mcp__npx__tool` allow/deny rules written against the old key and leak credentials
passed as CLI args into debug logs, since that log site builds a plain message string
and `logger` only sanitizes the context argument.

A remaining duplicate **explicit** `name` is rejected rather than silently dropped,
since a config that names two servers identically is a user error. That check runs
before the servers are transformed, so an invalid config no longer fetches OAuth
tokens on its way to being rejected.

Behavior change: previously a singular `server` sharing a name with a `servers` entry
silently overrode it (test-documented in #10645, not a designed feature). It is now an
error. Note that the runtime MCP client (`src/providers/mcp/client.ts`) resolves
`servers || [server]` instead, so that config shape was already ambiguous across the
two code paths.

## Test plan

`test/providers/mcp/transform.test.ts` covers each case: duplicate `name` rejection
(and that it precedes OAuth), two path servers, two same-command-different-args
servers, two servers whose args differ only by argument boundaries (`['a b']` vs
`['a', 'b']`), two same-url servers with different query auth, no argument value in
the derived key, the legacy `npx` key preserved for a single unnamed server with args,
and a derived key not stealing an explicit `name`.

```
$ npx vitest run test/providers/mcp/ test/providers/claude-agent-sdk.test.ts
  Test Files  8 passed (8)
  Tests  475 passed (475)

$ npx biome check src/providers/mcp/transform.ts test/providers/mcp/transform.test.ts   # clean
```

### DIFF
--- a/src/providers/mcp/transform.ts
+++ b/src/providers/mcp/transform.ts
@@ -129,21 +129,41 @@ export async function transformMCPConfigToClaudeCode(
     serverConfigs.push(config.server);
   }
 
+  // An explicit `name` owns its key, so two servers sharing one is a config error.
+  // Checked before transforming, which would otherwise fetch OAuth tokens for a
+  // configuration we are about to reject.
+  const names = serverConfigs.flatMap((server) => server.name ?? []);
+  const duplicateName = names.find((name, index) => names.indexOf(name) !== index);
+  if (duplicateName) {
+    throw new Error(
+      `Duplicate Claude Agent SDK MCP server \`${duplicateName}\`; give each configured server a unique \`name\`.`,
+    );
+  }
+
   const servers = await Promise.all(
     serverConfigs.map((server) => transformMCPServerConfigToClaudeCode(server)),
   );
 
-  const seen = new Set<string>();
-  for (const [key] of servers) {
-    if (seen.has(key)) {
-      throw new Error(
-        `Duplicate Claude Agent SDK MCP server \`${key}\`; give each configured server a unique \`name\`.`,
-      );
+  // An unnamed server falls back to a coarse identifier that several servers can
+  // share, so suffix collisions instead of letting `Object.fromEntries` drop one.
+  // The key becomes the SDK's server name — it lands in `mcp__<key>__<tool>` and in
+  // debug logs — so it deliberately carries no `args`, `env`, or auth values.
+  const taken = new Set(names);
+  const entries = servers.map((server, index) => {
+    const { name, url, command } = serverConfigs[index];
+    if (name) {
+      return [name, server] as const;
     }
-    seen.add(key);
-  }
+    const fallback = url ?? command ?? 'default';
+    let key = fallback;
+    for (let suffix = 2; taken.has(key); suffix++) {
+      key = `${fallback}_${suffix}`;
+    }
+    taken.add(key);
+    return [key, server] as const;
+  });
 
-  return Object.fromEntries(servers);
+  return Object.fromEntries(entries);
 }
 
 export function validateMCPConfigForClaudeCode(config: MCPConfig): void {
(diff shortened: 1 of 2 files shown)


mldangelo:
@codex review

mldangelo:
All five findings on this PR reproduced against `bf14303`, and all five are fixed in `0ac237d`. Verification and outcomes:

**1. Args in the derived key leak into logs (Copilot suppressed comment / Codex P1) — REAL, fixed.**
Probe on `bf14303`: `{ command: 'npx', args: ['-y', 'srv', '--api-key', 'sk-SUPERSECRET'] }` produced the key `` `npx -y srv --api-key sk-SUPERSECRET` ``. The log site is real (`src/providers/claude-agent-sdk.ts`, `mcpServers: options.mcpServers ? Object.keys(options.mcpServers) : undefined` inside a `JSON.stringify` embedded in the message). And the repo's existing redaction does **not** cover it: `createLogMethodWithContext` in `src/logger.ts` short-circuits with `if (!context) { internalLogger[level](message); return; }` — only the *context* argument goes through `sanitizeContext`/`sanitizeObject`, and neither `consoleFormatter` nor `fileFormatter` redacts. So a plain message string reaches the transport verbatim. Severity stands.

**2. Key change breaks existing `mcp__<server>__<tool>` rules (Codex P1) — REAL, fixed.**
Confirmed by probe: `{ command: 'npx', args: ['-y', '@h1deya/mcp-server-weather'] }` keys to `npx` on `main` and to `npx -y @h1deya/mcp-server-weather` on `bf14303`. `examples/claude-agent-sdk/mcp/promptfooconfig.yaml` shows the pattern users write against these names.

**3. Space-joined argv is not injective (Codex P2) — REAL, fixed.**
`[{ command: 'node', args: ['a b'] }, { command: 'node', args: ['a', 'b'] }]` both keyed to `node a b` on `bf14303`, so the new duplicate check rejected a valid two-server config — the opposite of this PR's purpose.

**4. URL key derived from `config.url` rather than `serverUrl` (Copilot inline) — REAL, fixed.**
Two servers on `https://mcp.example.com/v1` with different query-placed API keys were rejected as duplicates on `bf14303`. (On `main` they were worse: silently collapsed to the last one.)

**5. Duplicate check runs after OAuth fetch (Codex P2) — REAL, fixed.**
The `Promise.all` did run `getOAuthToken` before the duplicate check.

**The fix** takes the "preserve the legacy key, disambiguate only actual collisions" route, which resolves 1–4 together: the fallback key stays exactly what it was on `main` (`url ?? command ?? 'default'`), and a collision gets a numeric suffix (`npx`, `npx_2`) instead of dropping a server. No `args`, `env`, or auth value ever enters the key, single-server keys are byte-identical to `main`, and no valid config is rejected. Duplicate **explicit** `name`s are still an error, now raised before the transform so no OAuth token is fetched for a config we are about to reject.

Regression tests in `test/providers/mcp/transform.test.ts` cover each case above plus the compatibility case (single unnamed server with args keeps the `npx` key) and a derived key not stealing an explicit `name`. All eight fail on `bf14303` and pass on `0ac237d`.

jameshiester-oai (commented):
Reviewed exact head 0ac237df61e45d9fb2934e6a4001642ec59e41e2. The 52 focused MCP transform tests pass locally, broader MCP/Claude Agent SDK tests pass (357), and all current-head CI checks are green. The remaining normalized server-name collision and legacy permission-rule reassignment are already documented in existing inline comments; prior secret-bearing args and duplicate-name OAuth issues were addressed in this head. No distinct new finding. Native PR-review toolkit was unavailable (exit 127); this is the manual fallback.

PR #10812, opened 2026-09-08
feat(providers): support GPT Image 2.5 generation

mldangelo-oai:
## Summary

Support `openai:image:gpt-image-2.5-sunburst` and `openai:image:gpt-image-2.5-flare`, including their September 8 snapshots. Use GPT Image request/response handling, accept `xhigh`/`max` quality, transparent PNG/WebP, and custom dimensions.

Calculate cost from returned token usage at the published rates; leave it unset when usage is absent. Update the provider guide with a runnable example.

## Validation

- 212 image and billing tests pass, including invalid settings, prompt overrides, and cached costs.
- Combined stack passes TypeScript, lint/format checks, core build, and docs build.
- Built CLI against a local Image API mock: all four model variants return PNG output and the expected cost. No live model calls.
- Full backend run: 24,839 passed; seven dependency-test failures reproduce on unchanged `main` with the same older installed dependencies. Socket policy blocked a clean reinstall; updated-head CI validates the lockfile installation.

Request options and pricing checked against the [image guide](https://developers.openai.com/api/docs/guides/image-generation) and [Sunburst model card](https://developers.openai.com/api/docs/models/gpt-image-2.5-sunburst).

PR #10842, opened 2026-09-09
fix(providers): stop guessing image cost for unpriced models

mldangelo:
## Summary

`calculateImageCost` in `src/providers/openai/image.ts` ended with a bare `return 0.04 * n;` for any model it did not recognize — a silent, wrong number (DALL-E 3 standard pricing) for a model nobody has priced. That default is worse than reporting nothing: it lands in eval results as if it were a real cost.

GPT Image 2.5 (added in #10812) only escaped that guess because of a special-cased `if (isGptImage25(model)) return undefined;` at the top of the function. Its `undefined` was deliberate and already documented — the model is billed from token usage, and its `xhigh`/`max` quality tiers have no published per-image rate — so the right fix is not to invent a 2.5 per-image table, but to make "no published rate → no cost" the general rule and delete the special case.

Changes:

- Unknown/unpriced models now return `undefined` instead of `0.04 * n`, matching the convention already used elsewhere in the OpenAI billing code (`calculateOpenAIUsageCost` returns `undefined` when `getModelRates` has no entry, and `calculateOpenAIUsageCostFromTokenUsage` explicitly refuses to infer unpublished Bedrock rates). The response simply omits `cost`, which downstream code already handles.
- **Deleted** the `isGptImage25` early return: 2.5 now reaches the same "no rate" path as everything else, with identical behavior.
- The no-cost outcome is no longer silent — it logs at debug with the model, size, and quality, and the docs now state the rule for the image provider generally (replacing the 2.5-only sentence).
- **Simplification pass:** the if/else-if chain of six near-identical `${quality}_${size}` table lookups (each recomputing the key and multiplying by `n`) collapsed into one `getPerImageCost` helper that returns the single-image price; `calculateImageCost` is now three lines. `n` is applied once, `||` fallbacks became `??`, and the redundant `imageQuality`/`gptImageQuality` pair became one `tieredQuality` local. Net −2 lines in `src/` despite adding the debug log and two comments.

No pricing tables were changed, and every priced model/size/quality combination returns exactly what it did before.

## Test plan

`test/providers/openai/image.functions.test.ts`: the two tests that asserted the `0.04` guess now assert `undefined` (they fail on `main`), plus a new `it.each` locking in that all four GPT Image 2.5 ids — aliases and `2026-09-08` snapshots — never borrow GPT Image 2's per-image rates, for both a shared quality (`high`) and a 2.5-only one (`max`).

```
$ npx vitest run test/providers/openai/image.functions.test.ts test/providers/openai/image.test.ts
  Test Files  2 passed (2)
       Tests  146 passed (146)

$ npx vitest run test/providers/openai
  Test Files  42 passed | 1 skipped (43)
       Tests  1819 passed | 9 skipped (1828)

  (clean)

$ npx biome check --write src/providers/openai/image.ts test/providers/openai/image.functions.test.ts
  Checked 2 files. No fixes applied.

$ npx prettier --write site/docs/providers/openai.md
  unchanged
```

### DIFF
--- a/site/docs/providers/openai.md
+++ b/site/docs/providers/openai.md
@@ -752,7 +752,9 @@ tests:
       product: a blue ceramic mug
 ```
 
-GPT Image 2.5 also accepts `quality: xhigh` and `quality: max`. For transparent output, use `background: transparent` with PNG or WebP. Cost comes from the response's token usage; it is left unset when usage is missing because older models' per-image estimates do not apply. See the [Image API guide](https://developers.openai.com/api/docs/guides/image-generation).
+GPT Image 2.5 also accepts `quality: xhigh` and `quality: max`. For transparent output, use `background: transparent` with PNG or WebP. See the [Image API guide](https://developers.openai.com/api/docs/guides/image-generation).
+
+Image cost comes from the response's token usage when the model publishes token rates, and otherwise from a per-image list price. Models with neither — including GPT Image 2.5 responses that carry no usage — report no cost rather than an estimate.
 
 This provider supports generation only. Image editing, masks, reference images, variations, and streaming are not implemented.
 
--- a/src/providers/openai/image.ts
+++ b/src/providers/openai/image.ts
@@ -692,55 +692,53 @@ export function prepareRequestBody(
   return body;
 }
 
+// List price of a single image, or undefined when the model, size, and quality combination
+// has no published per-image rate.
+function getPerImageCost(model: string, size: string, quality?: string): number | undefined {
+  if (model === 'dall-e-2') {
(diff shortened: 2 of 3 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 1a3e290a0e991658205ef8e03d5cb4f5f1c4aa26 (image cost when no rate is available). The focused image suites pass 146/146, all reported checks are green, and mocked provider responses confirm that an unknown model or GPT Image 2.5 without usage omits cost, DALL-E 3 retains its published rate, and cached responses report zero. No distinct new issue. The existing inline comment still applies to the documentation's no-estimate promise for GPT Image 1/mini/1.5 with automatic quality or size, which retain a fallback estimate.

PR #10844, opened 2026-09-09
fix(cloudflare): drop gateway routes that cannot work

mldangelo:
## Summary

`PROVIDER_CONFIGS` in `src/providers/cloudflare-gateway.ts` still advertised five routes — `google-ai-studio`, `cohere`, `workers-ai`, `huggingface`, `replicate` — that the provider's own documentation already declares broken. Each builds an OpenAI Chat Completions request against a native Cloudflare gateway endpoint that does not accept that payload, so a user who selected one got a confusing upstream error rather than a fast, actionable failure. Only `mistral` was repaired in #10734; the rest were left in the surface area.

Reconciled code with docs by **deleting** the broken routes rather than adding new special cases:

- Removed the five entries from `PROVIDER_CONFIGS`. `createCloudflareGatewayProvider` already fails fast with `Unsupported Cloudflare AI Gateway provider: "<name>". Supported providers: …`, so removal is the whole fix — no new error path.
- Deleted the now-dead `workers-ai` branch in `buildGatewayUrl`, and with it the `modelName` parameter that branch was the only user of (plus its call-site argument and JSDoc line).
- Tightened the `PROVIDER_CONFIGS` doc comment to state the actual rule (routes must speak OpenAI Chat Completions or Anthropic Messages) instead of listing per-provider URL quirks.
- Updated `site/docs/providers/cloudflare-gateway.md` so the note says these routes are rejected with an error, matching the new behavior. The supported-providers table already listed only the seven working routes; the Workers AI section already points at the OpenAI-compatible REST API and is unchanged.

Test side: replaced the `workers-ai` URL/id tests (which asserted the broken shape) and the single bedrock rejection test with a table-driven rejection test covering all six unsupported routes, asserting the full error message — which also pins the supported-provider list in the error to the list the suite claims is supported.

Net **-49 lines**.

## Test plan

```
npx vitest run test/providers/cloudflare-gateway.test.ts test/providers/cloudflare-native-path.test.ts \
  test/providers/cloudflare-ai.test.ts test/providers/registry.test.ts
  → Test Files 4 passed (4), Tests 268 passed (268)

# regression check: stash only the src change, rerun
git stash push src/providers/cloudflare-gateway.ts && npx vitest run test/providers/cloudflare-gateway.test.ts
  → 6 failed (the six unsupported routes constructed a provider instead of throwing)

npx tsc --noEmit -p tsconfig.json          → clean
npx biome check src/providers/cloudflare-gateway.ts test/providers/cloudflare-gateway.test.ts → clean
npx prettier --check site/docs/providers/cloudflare-gateway.md → unchanged
```

### DIFF
--- a/src/providers/cloudflare-gateway.ts
+++ b/src/providers/cloudflare-gateway.ts
@@ -70,25 +70,21 @@ interface GatewayProviderConfig {
 /**
  * Supported provider configurations for Cloudflare AI Gateway
  *
- * Note: Some providers have special URL requirements:
- * - azure-openai: Requires resourceName and deploymentName in config
- * - workers-ai: Model name is appended to URL path
+ * Only gateway routes that speak the OpenAI Chat Completions or Anthropic Messages
+ * protocol are listed. Routes whose native endpoints reject those payloads
+ * (workers-ai, google-ai-studio, cohere, huggingface, replicate) are deliberately
+ * absent; use Cloudflare's OpenAI-compatible REST API or a custom provider instead.
+ * AWS Bedrock is likewise unsupported because it requires AWS request signing.
  *
- * AWS Bedrock is NOT supported because it requires AWS request signing
- * which is incompatible with the gateway proxy approach.
+ * Note: azure-openai requires resourceName and deploymentName in config.
  */
 const PROVIDER_CONFIGS: Record<string, GatewayProviderConfig> = {
   openai: { apiKeyEnvar: 'OPENAI_API_KEY' },
   anthropic: { apiKeyEnvar: 'ANTHROPIC_API_KEY' },
   groq: { apiKeyEnvar: 'GROQ_API_KEY' },
   'perplexity-ai': { apiKeyEnvar: 'PERPLEXITY_API_KEY' },
-  'google-ai-studio': { apiKeyEnvar: 'GOOGLE_API_KEY' },
   mistral: { apiKeyEnvar: 'MISTRAL_API_KEY' },
-  cohere: { apiKeyEnvar: 'COHERE_API_KEY' },
   'azure-openai': { apiKeyEnvar: 'AZURE_OPENAI_API_KEY' },
-  'workers-ai': { apiKeyEnvar: 'CLOUDFLARE_API_KEY' },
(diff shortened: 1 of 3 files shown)


mldangelo:
@codex review

mldangelo:
@codex review

src/providers/cloudflare-gateway.ts
```diff
@@ -70,25 +70,21 @@ interface GatewayProviderConfig {
...
+ * protocol are listed. Routes whose native endpoints reject those payloads
+ * (workers-ai, google-ai-studio, cohere, huggingface, replicate) are deliberately
+ * absent; use Cloudflare's OpenAI-compatible REST API or a custom provider instead.
+ * AWS Bedrock is likewise unsupported because it requires AWS request signing.
  *
- * AWS Bedrock is NOT supported because it requires AWS request signing
- * which is incompatible with the gateway proxy approach.
+ * Note: azure-openai requires resourceName and deploymentName in config.
  */
 const PROVIDER_CONFIGS: Record<string, GatewayProviderConfig> = {
```
jameshiester-oai, line 81: [P2] Reject removed routes in the exported constructor too. `CloudflareGatewayOpenAiProvider` is public and can still be instantiated directly with `workers-ai` (or another removed route). With its mapping gone, `providerConfig` is undefined, so the OpenAI base class falls back to `OPENAI_API_KEY` and sends that credential as a Bearer token to the unsupported Cloudflare gateway route. A local constructor probe with both keys set selected `OPENAI_API_KEY`, where the old mapping selected `CLOUDFLARE_API_KEY`. Validate `underlyingProvider` against this supported list before constructing the gateway URL or resolving credentials.

jameshiester-oai (commented):
Reviewed exact head df834b02cddfe157451c4e45267ae94ecdb28eb5 against current main. The normal provider registry rejects all six removed routes, while the supported OpenAI route loads. Focused Cloudflare/provider tests pass (268), and all 43 CI checks succeeded. The native PR-review toolkit was unavailable (exit 127), so I used the documented manual review path. One narrow direct-constructor regression is noted inline.

PR #10845, opened 2026-09-09
fix(providers): stop leaking local config into request bodies

mldangelo:
## Summary

`loadApiProvider` merges the loaded config file's directory into **every** provider config as `basePath`. Providers that iterate their config into `passthrough` — which the shared OpenAI transport serializes verbatim into the request body — need an allowlist of the keys promptfoo consumes itself. Three providers hand-maintained their own copy of that list, and two of them were incomplete.

Captured against a local HTTP echo server with the built CLI (`node dist/src/main.js eval …`), before this change:

```json
{"model":"meta-llama/Llama-3.3-70B-Instruct","messages":[…],"max_tokens":1024,"temperature":0,"basePath":"/private/tmp/…/wire"}
{"model":"@cf/meta/llama-3.1-8b-instruct","messages":[…],"max_tokens":1024,"temperature":0,"basePath":"/private/tmp/…/wire"}
```

Three bugs, one root cause:

- **nscale** — `NSCALE_PROVIDER_LEVEL_OPTIONS` covered `OpenAiSharedOptions` but not `basePath`, so the local path went to the model.
- **cloudflare-ai** — `getPassthroughConfig` destructured only the five Cloudflare-specific keys, so `basePath`, `headers`, `maxRetries`, cost overrides and an explicit `passthrough` (nested one level deep!) all became model parameters.
- **cerebras** — filtered `basePath` but spread everything else into `passthrough`, so a configured `apiKey` was transmitted in the request body and custom `headers` never became HTTP headers. It also hardcoded `apiBaseUrl`, so a configured one was silently ignored *and* shipped in the body — in the pre-fix wire run its request went to the real `api.cerebras.ai` and returned 401 instead of the configured endpoint.

This is the same class of bug fixed for Portkey in #10536.

### What changed

Rather than adding `basePath` to three lists, this deletes two of them. `togetherai.ts` had the only complete version, typed `satisfies Record<keyof OpenAiSharedOptions, boolean> & …` so a newly added shared option fails the build instead of leaking. That list and the split it drives move to `splitLocalOptions` in `src/providers/openai/util.ts`, and all four providers import it; `extraLocalKeys` lets Cloudflare keep `accountId`/`accountIdEnvar` local without redefining the set.

Also deleted along the way: nscale's bespoke split loop, cloudflare's `getPassthroughConfig` destructure, cerebras's `basePath` filter. `src/` is net **-40 lines** (80 added, 105 removed) with one shared definition instead of three.

I checked the other providers that build `passthrough` (openrouter, orcarouter, snowflake, databricks, llamaApi): they use explicit per-key allowlists and never iterate the config, so they are unaffected. `envoy.ts` spreads config into `config`, not `passthrough`, so nothing reaches the body.

Behavior changes worth calling out: cerebras now honors a configured `apiBaseUrl` and `apiKeyEnvar` (previously ignored), and cloudflare-ai now merges an explicit `config.passthrough` into the body instead of nesting it under a `passthrough` key.

## Test plan

Three regression tests, each verified failing on the pre-fix source and passing after:

- `test/providers/nscaleRequest.test.ts` — wire-level (real OpenAI provider, mocked transport): body has no `basePath`.
- `test/providers/cloudflare-ai.test.ts` — wire-level via `mockFetch`: body has no `basePath`/`headers`/`maxRetries`, `temperature` survives, `X-Tenant` is sent as an HTTP header.
- `test/providers/cerebras.test.ts` — `apiKey` and `headers` stay out of `passthrough` and land in config.

```
npx vitest run test/providers/            → 240 passed | 1 skipped (241 files), 9743 tests passed
npx vitest run test/test-hygiene.test.ts  → 115 passed
npx tsc --noEmit -p tsconfig.json         → clean
npx biome ci <changed files>              → clean
npm run architecture:check                → Architecture boundaries passed
```

End-to-end on the wire, `npm run build` then the built CLI against a local node http echo server, one config file in a subdirectory (so `basePath` is non-empty), all four providers pointed at it:

```
node dist/src/main.js eval -c <scratch>/wire/promptfooconfig.yaml --no-cache -o out.json
→ 4 passed (100%), 0 errors
```

Captured bodies after the fix — `basePath` gone, everything else unchanged, credentials only in the `Authorization` header:

```json
{"model":"meta-llama/Llama-3.3-70B-Instruct","messages":[{"role":"user","content":"Say hi"}],"max_tokens":1024,"temperature":0} |auth= Bearer wire-token
{"model":"@cf/meta/llama-3.1-8b-instruct","messages":[{"role":"user","content":"Say hi"}],"max_tokens":1024,"temperature":0}    |auth= Bearer wire-token
{"model":"meta-llama/Llama-3-8b-chat-hf","messages":[{"role":"user","content":"Say hi"}],"max_tokens":1024,"temperature":0}     |auth= Bearer wire-token
{"model":"llama3.1-8b","messages":[{"role":"user","content":"Say hi"}],"max_tokens":1024,"temperature":0}                       |auth= Bearer wire-token
```

### DIFF
--- a/src/providers/cerebras.ts
+++ b/src/providers/cerebras.ts
@@ -1,4 +1,5 @@
 import { OpenAiChatCompletionProvider } from './openai/chat';
+import { splitLocalOptions } from './openai/util';
 
 import type { EnvOverrides } from '../types/env';
 import type { ApiProvider, ProviderOptions } from '../types/index';
@@ -22,8 +23,10 @@ export function createCerebrasProvider(
   const splits = providerPath.split(':');
   const modelName = splits.slice(1).join(':');
 
-  // Filter out basePath from config to avoid passing it to the API
-  const { basePath: _, ...configWithoutBasePath } = options.config?.config || {};
+  const config = options.config?.config || {};
+  // Only genuine model parameters belong in the request body; promptfoo's own settings
+  // (credentials, headers, cost overrides, basePath) stay local.
+  const { localOptions, modelParameters } = splitLocalOptions(config);
 
   // Create a custom provider class that overrides the getOpenAiBody method
   class CerebrasProvider extends OpenAiChatCompletionProvider {
@@ -48,11 +51,10 @@ export function createCerebrasProvider(
   const cerebrasConfig = {
     ...options,
     config: {
-      apiBaseUrl: 'https://api.cerebras.ai/v1',
-      apiKeyEnvar: 'CEREBRAS_API_KEY',
-      passthrough: {
-        ...configWithoutBasePath,
-      },
+      ...localOptions,
+      apiBaseUrl: localOptions.apiBaseUrl || 'https://api.cerebras.ai/v1',
+      apiKeyEnvar: localOptions.apiKeyEnvar || 'CEREBRAS_API_KEY',
+      passthrough: { ...modelParameters, ...config.passthrough },
     },
   };
 
--- a/src/providers/cerebras.ts
+++ b/src/providers/cerebras.ts
@@ -1,5 +1,5 @@
 import { OpenAiChatCompletionProvider } from './openai/chat';
-import { splitLocalOptions } from './openai/util';
+import { splitLocalOptions } from './openai/localOptions';
 
 import type { EnvOverrides } from '../types/env';
 import type { ApiProvider, ProviderOptions } from '../types/index';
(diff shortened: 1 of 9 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed b950decc43423e0f2c3ee9e569ec4389f21f7dbf (local provider option separation). The focused provider suites pass 91/91 and all 43 reported checks are green. Mocked Nscale and Cloudflare requests keep local settings out of the JSON body and preserve the expected HTTP headers. No distinct new issue. Existing inline comments cover newly retained absolute base paths in Cerebras/Nscale provider config and Cerebras failing to receive a provider-scoped custom credential.

PR #10846, opened 2026-09-09
fix(providers): stop clobbering resolved TogetherAI model params

mldangelo:
## Summary

`createTogetherAiProvider` left every model parameter in `config` **and** copied it into
`passthrough`. The OpenAI chat provider resolves several of those options out of `config`
before building the request body — it loads `file://` references, renders Nunjucks vars,
merges in MCP tools and normalizes tool shapes — but `passthrough` is spread into the body
**last**, so the raw duplicate overwrote whatever the chat provider had just resolved.

Net effect for TogetherAI users:

- `tools: file://tools.json` (or `file://tools.py:get_tools`) shipped as the literal
  string instead of the loaded definitions.
- Any var-templated model parameter (`{{ toolName }}`, a templated `response_format`
  schema) shipped un-rendered.
- MCP tools merged into `tools` by the chat provider were dropped.
- `tool_choice` skipped its provider-format transform.

This is a latent bug, not a 0.122.2 regression. Before #10681 the factory put model
parameters *only* in `passthrough`, so the chat provider never got a chance to resolve them
at all. #10681 started passing them through `config` as well, which made the resolution
run — and made the raw `passthrough` copy visibly clobber it.

### What changed

- `tools`, `functions`, `tool_choice` and `response_format` are excluded from the
  `passthrough` copy **on the chat route only**, so each one lives in exactly one place and
  the chat provider's processed value reaches the wire. The chat provider emits all four
  whenever they are configured, so dropping the raw copy cannot lose a parameter.
  `OpenAiCompletionProvider` and `OpenAiEmbeddingProvider` resolve nothing and only spread
  `passthrough`, so the exclusion does not apply to `togetherai:completion:*` or
  `togetherai:embedding:*`.
- `reasoning_effort` is deliberately **not** excluded. The chat provider only emits it for
  model names it recognizes as reasoning models (`o1`/`o3`/`o4`/`gpt-5`/`gpt-oss`), which
  no other TogetherAI model name matches, so `passthrough` has to keep carrying it.
- Everything the OpenAI providers do not understand (`repetition_penalty`, `top_k`,
  `safety_model`, …) still goes through `passthrough` as before, and an explicit
  `config.passthrough` entry still overrides a top-level model parameter, as documented.
- The four-branch `splits[1]` route dispatch collapsed into a `Map` lookup, removing four
  copies of `splits.slice(...).join(':')`. A `Map` rather than an object literal so a route
  segment naming an `Object` prototype member (`togetherai:constructor:foo`) does not
  resolve to that member.

`localOptions` is deliberately left untouched (contents and name) — sibling PR #10845
extracts it into a shared provider-level options allowlist for `nscale` and `cloudflare-ai`.
The two branches **do conflict textually** in `src/providers/togetherai.ts` (#10845 deletes
the `localOptions` block this PR adds next to); whichever lands second resolves by keeping
#10845's `splitLocalOptions(config)` call and applying this PR's `chatResolvedOptionNames`
filter to the `modelParameters` it returns. Note that `cloudflare-ai.ts` has the same
double-write shape and is likely to need the same treatment; left out of scope here.

## Test plan

Four regression tests in `test/providers/togetherai-request.test.ts`, each verified failing
before the corresponding fix and passing after:

- a chat provider with a var-templated `tools` entry and `response_format` schema name —
  `name: "{{ toolName }}"` on the wire before, rendered after;
- `reasoning_effort` on `togetherai:chat:deepseek-ai/DeepSeek-R1` still reaches the body;
- `response_format` on `togetherai:completion:*` still reaches the body;
- `togetherai:constructor:fixture-model` still routes to the chat provider.

```
$ npx vitest run test/providers/togetherai-request.test.ts test/providers/togetherai.test.ts
  Test Files  2 passed (2)
       Tests  21 passed (21)

$ npx vitest run test/providers/index.test.ts test/providers/registry.test.ts test/test-hygiene.test.ts
  (clean)

  (clean)

$ npx biome ci src/providers/togetherai.ts test/providers/togetherai-request.test.ts
  Checked 2 files. No fixes applied.
```


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 34459ee592bbab94dd350897de27472a019b024a (TogetherAI model parameters). The current chat path keeps raw passthrough settings that the base provider does not emit, while allowing its resolved tools, tool choice, response format, and functions to reach the outbound request. Earlier reasoning-effort and inherited route-key comments are addressed on this head. The focused factory/request tests pass 21/21 and all 43 reported checks are green; no distinct new issue found.

PR #10849, opened 2026-09-09
fix(providers): finish the GitHub Models retirement

mldangelo:
## Summary

[#10746](https://github.com/promptfoo/promptfoo/pull/10746) retired the `github:` provider (`createGitHubProvider` now throws `GITHUB_MODELS_RETIREMENT_MESSAGE`, and `GITHUB_TOKEN` no longer selects a default grader), but two surfaces still told users it works.

**`promptfoo init --example github-models`** — the example was deleted from `examples/` and dropped from `EXAMPLE_ALIASES`, but never added to `REMOVED_EXAMPLES`, so it failed with a generic "not found" instead of the retirement note that mechanism exists to print. Added both the pre-reorg (`github-models`) and post-reorg (`provider-github-models`) names, each pinned to a ref where that directory actually exists — `0.120.26` and `31b5668…`, the same historical ref the two sibling entries use.

**Red team setup UI** — `github` was still in `ProviderConfigEditor`'s specialized-provider list, so a saved `github:` target rendered a fully editable raw-config panel, plus a "Need help configuring…? View the documentation" alert from `providerDocumentationMap`. A retired provider has no configuration that can make it run again, so both were removed and replaced with a warning that states the retirement and points at the (already-rewritten) provider docs page.

Net: `github` deleted from two provider lists; the only addition is the notice that replaces the dead editor, plus the two data entries.

## Test plan

- `npx vitest run test/commands/init.test.ts` → 30 passed. Reverting only `exampleAliases.ts` fails the two new cases (fetch goes to the default refs, not the pinned one), confirming the regression test bites.
- `npx vitest run test/commands` → 39 files, 780 passed.
- `npm run test:app -- src/pages/redteam/setup/components/Targets --run` → 25 files, 494 passed (includes the new `ProviderConfigEditor` and `providerDocumentationMap` cases).
- `npx tsc --noEmit -p tsconfig.json` and `npx tsc --noEmit -p src/app/tsconfig.json` → clean.
- `npx biome ci` on the changed files → exit 0 (the three `noExcessiveCognitiveComplexity` warnings on `ProviderConfigEditor.tsx` are pre-existing; verified the same three appear on `HEAD`).
- Live CLI check against GitHub, both names resolving and downloading:

```
$ promptfoo init <tmp> --example github-models --no-interactive
Note: github-models was removed because GitHub retired GitHub Models, including its inference API, on July 30, 2026.
Downloading the legacy 'github-models' example from promptfoo@0.120.26.
✅ Example project 'github-models' written to: <tmp>/github-models

$ promptfoo init <tmp> --example provider-github-models --no-interactive
Note: provider-github-models was removed because GitHub retired GitHub Models, including its inference API, on July 30, 2026.
Downloading the legacy 'provider-github-models' example from promptfoo@31b566872971532e6d428c0cbad4487d22d936c5.
✅ Example project 'provider-github-models' written to: <tmp>/provider-github-models
```

### DIFF
--- a/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.test.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.test.tsx
@@ -1345,6 +1345,19 @@ describe('ProviderConfigEditor', () => {
     expect(screen.queryByTestId('custom-config')).not.toBeInTheDocument();
   });
 
+  it('should explain the retirement instead of editing a saved GitHub Models target', () => {
+    renderWithProviders(
+      <ProviderConfigEditor
+        provider={{ id: 'github:openai/gpt-5', config: {} }}
+        setProvider={vi.fn()}
+        providerType="github"
+      />,
+    );
+
+    expect(screen.getByRole('alert')).toHaveTextContent('GitHub retired GitHub Models');
+    expect(screen.queryByTestId('custom-config')).not.toBeInTheDocument();
+  });
+
   it('should render the raw provider editor for Open Interpreter targets', () => {
     renderWithProviders(
       <ProviderConfigEditor
--- a/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
+++ b/src/app/src/pages/redteam/setup/components/Targets/ProviderConfigEditor.tsx
@@ -1,6 +1,8 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
+import { Alert, AlertContent, AlertDescription } from '@app/components/ui/alert';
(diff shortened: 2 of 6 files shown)


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 40bc7d394702569450fa341f22d781fd9b77333d (GitHub Models retirement paths). Both pinned legacy example references resolve, the linked provider guide carries its retirement notice, and the focused CLI/UI suites pass 30/30 and 86/86; all 43 reported checks are green. No distinct new issue. The existing inline finding still applies: a saved `github:` target shows the warning but passes setup validation and can advance into a provider that rejects evaluation.

PR #10850, opened 2026-09-09
test: stop loading the developer .env during tests

mldangelo:
## Summary

`src/envars.ts` calls `dotenv.config()` at import time, so every vitest worker inherited the repo's gitignored `.env`. Any contributor with real credentials there ran a different test suite than CI does — and three tests in `test/providers/google/interactions.test.ts` failed outright:

```
> requires a Gemini API key
    Expected: "requires an API key"
    Received: "Gemini Interactions API error: TypeError: Cannot destructure property 'data' ..."
> does not reject an AI Studio Omni follow-up supplied through passthrough
> preserves a configured provider id and supports the legacy PALM API key
    x-goog-api-key received the developer's real key instead of the fixture's 'legacy-palm-key'
```

The two "requires an API key" cases were silently satisfied by the ambient `GEMINI_API_KEY` (their `env: { GOOGLE_API_KEY: undefined, GEMINI_API_KEY: undefined }` overrides never suppressed `process.env`), and the legacy-PALM case asserted a real credential on the wire. CI was green only because CI has no `.env`.

Changes:

- **`src/envars.ts`** — skip the `.env` load inside a test process, detected the same way `getDbPath()` in `src/database/index.ts` already does it: `VITEST`, or the runner-owned `__vitest_worker__` global for the case where a test wiped `process.env` (`mockProcessEnv({}, { clear: true })`) before a module reaching `envars` is re-imported. This immunises the whole backend suite (every vendor, not just Google) rather than patching one file.
- **`src/server/server.ts`** — deleted the redundant `dotenv` import and `dotenv.config()`. ESM hoisting means it already ran via `../constants` → `./envars` before that statement executes, so it was dead code that also reopened the leak for server tests.
- **`test/providers/google/interactions.test.ts`** — clear the Google credential env vars (`GOOGLE_API_KEY`, `GEMINI_API_KEY`, `PALM_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY` — every key the AI Studio path actually consults) in the existing `beforeEach` stub block, next to the host vars it already stubs for the same reason. The tests are now hermetic against shell-exported keys too, not just `.env`, which let the two ineffective `env: { ... undefined }` overrides be deleted.
- **`test/envars.test.ts`** — two regression tests: chdir into a temp dir containing a `.env`, re-import `src/envars`, assert the probe variable did not reach `process.env` — once normally, once with `process.env` cleared first. Deterministic in CI, not dependent on a local `.env` existing.

## Test plan

All run with the developer `.env` present (the failing condition), on Node 24.20.0:

- `npx vitest run test/envars.test.ts test/providers/google/interactions.test.ts` → 130 passed (was `3 failed | 40 passed` for the interactions file)
- Regression check, `src/envars.ts` fix stashed: `npx vitest run test/envars.test.ts -t "does not load a .env"` → failed, as expected; with only the `VITEST` half of the guard, the cleared-`process.env` case still failed
- Hermeticity check: `GEMINI_API_KEY=… GOOGLE_API_KEY=… PALM_API_KEY=… VERTEX_API_KEY=… GOOGLE_GENERATIVE_AI_API_KEY=shell-leaked npx vitest run test/providers/google/interactions.test.ts` → 43 passed (3 failed with the test change stashed; 2 failed on `GOOGLE_GENERATIVE_AI_API_KEY` alone before it was added to the stub list)
- `npx vitest run test/providers/google test/server test/hygiene test/test-hygiene.test.ts test/util/env.test.ts` → 44 files, 1553 passed
- `npx vitest run` (full backend suite) → 849 passed | 1 skipped, 24883 tests passed
- Same targeted runs with `.env` absent → pass
- `npx tsc --noEmit -p tsconfig.json` → clean
- `npx biome check` on the four changed files → clean
- `npm run architecture:check` → boundaries passed

### DIFF
--- a/src/envars.ts
+++ b/src/envars.ts
@@ -5,8 +5,12 @@ import type { EnvOverrides } from './types/env';
 
 // Never load the developer's `.env` inside a test run: unit tests would silently pick up
 // real credentials and behave differently from CI, which has none. Mirrors the guard on
-// the default database path in src/database/index.ts.
-if (process.env.VITEST !== 'true') {
+// the default database path in src/database/index.ts, including its runner-owned global:
+// helpers that wipe process.env drop VITEST, but the global survives them.
+const isTestProcess =
+  process.env.VITEST === 'true' ||
+  Object.prototype.hasOwnProperty.call(globalThis, '__vitest_worker__');
+if (!isTestProcess) {
   dotenv.config({ quiet: true });
 }
 
--- a/src/envars.ts
+++ b/src/envars.ts
@@ -3,7 +3,12 @@ import { getEnvOverrides } from './envOverrides';
 
 import type { EnvOverrides } from './types/env';
 
-dotenv.config({ quiet: true });
+// Never load the developer's `.env` inside a test run: unit tests would silently pick up
+// real credentials and behave differently from CI, which has none. Mirrors the guard on
+// the default database path in src/database/index.ts.
+if (process.env.VITEST !== 'true') {
+  dotenv.config({ quiet: true });
+}
 
 // Define the supported environment variables and their types
 type EnvVars = {
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -1,13 +1,9 @@
-import compression from 'compression';
-import cors from 'cors';
-import dotenv from 'dotenv';
-
-dotenv.config({ quiet: true });
-
 import fs from 'node:fs';
 import http from 'node:http';
 import path from 'node:path';
 
+import compression from 'compression';
+import cors from 'cors';
 import express from 'express';
 import { Server as SocketIOServer } from 'socket.io';
 import { getDefaultPort, VERSION } from '../constants';
--- a/test/envars.test.ts
+++ b/test/envars.test.ts
@@ -171,9 +171,13 @@ describe('envars', () => {
   describe('dotenv loading', () => {
     // A developer's gitignored .env must never reach the suite: tests would pick up
     // real credentials and diverge from CI, which has none.
+    // Resolved up front because os.tmpdir() reads TEMP/TMP, which the cleared-environment
+    // case below wipes: on Windows that yields the unusable path "undefined\temp".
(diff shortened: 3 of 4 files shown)


mldangelo:
@codex review

src/envars.ts
```diff
@@ -3,7 +3,16 @@ import { getEnvOverrides } from './envOverrides';
...
 
-dotenv.config({ quiet: true });
+// Never load the developer's `.env` inside a test run: unit tests would silently pick up
+// real credentials and behave differently from CI, which has none. Mirrors the guard on
+// the default database path in src/database/index.ts, including its runner-owned global:
+// helpers that wipe process.env drop VITEST, but the global survives them.
+const isTestProcess =
+  process.env.VITEST === 'true' ||
+  Object.prototype.hasOwnProperty.call(globalThis, '__vitest_worker__');
+if (!isTestProcess) {
```
jameshiester-oai, line 13: [P2] Cover implicit command setup as well as import-time dotenv loading. In a Vitest process this guard skips `dotenv.config()`, but `src/util/env.ts:52-53` still calls it unconditionally from `setupEnv(undefined)`. `doEval()` and `showCommand` invoke that default path in command tests, so the developer's root `.env` is loaded later into the test worker despite this guard. A controlled synthetic `.env` probe confirmed the variable was absent after importing `envars` and present after `setupEnv(undefined)` under `VITEST=true`. Keep explicit `--env-file` behavior, but suppress the implicit default load in tests or isolate those command paths and add a command-level regression test.

jameshiester-oai (commented):
Reviewed 36b78823e8d6aaae83145c9d9abaa40e45d62f5a (developer .env isolation in tests). The focused envars/Google suites pass 130/130, and the required checks are green. The earlier Vitest worker-marker and Google credential comments are fixed. One distinct isolation path remains (inline): command tests calling `setupEnv(undefined)` still load the default .env after the import-time guard. Existing feedback separately notes that a pre-exported probe variable can make the new test fail.

PR #10854, opened 2026-09-09
refactor(cli): return model-audit args directly

mldangelo:
## Summary

`parseModelAuditArgs` returned `{ args: string[]; unsupportedOptions: string[] }`, but `unsupportedOptions` was hard-coded to `[]` and no caller ever read it — dead surface left behind when option validation moved into the Zod schema (`ModelAuditCliOptionsSchema.parse` now throws on bad input instead of collecting names).

Removing it leaves a single-field wrapper object with no reason to exist, so `parseModelAuditArgs` and `buildCliArgs` now return the argv array itself and the three call sites drop their destructuring. `src/commands/modelScan.ts` also loses a `const args = parsed;` alias that only existed to unwrap the object.

No behavior change — `args` was already the only field anyone used.

## Test plan

- `npx vitest run test/utils/modelAuditCliParser.test.ts test/commands/modelScan.test.ts test/server/routes/modelAudit.test.ts` — 119 passed
- `npx tsc --noEmit -p tsconfig.json` — clean
- `npx biome check` — clean (the three cognitive-complexity warnings are pre-existing on `main`)

Net −9 lines.

### DIFF
--- a/src/commands/modelScan.ts
+++ b/src/commands/modelScan.ts
@@ -284,10 +284,7 @@ function hasPersistedScannerSelection(metadata: ModelAudit['metadata']): boolean
  * Parse CLI options through Zod, logging validation errors to the CLI.
  * Returns null when validation fails (and sets process.exitCode to 1).
  */
-function buildCliArgs(
-  paths: string[],
-  cliOptions: Record<string, unknown>,
-): { args: string[]; unsupportedOptions: string[] } | null {
+function buildCliArgs(paths: string[], cliOptions: Record<string, unknown>): string[] | null {
   try {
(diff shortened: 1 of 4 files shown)


mldangelo:
@codex review

mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed 9aa6cc1f71bf17019154c2bf110d1d99082e29a3 (model-audit CLI argument return shape). The Zod validation and argument order are unchanged, and all three production call sites consume the returned array directly. The focused parser/CLI/server tests pass 119/119, the source `scan-model --help` invocation exits successfully, and all 43 reported checks are green. No distinct issue found.

PR #10861, opened 2026-09-09
fix(providers): preserve WatsonX caller cancellation

mldangelo-oai:
Concurrent WatsonX calls now share client initialization and pricing metadata lookups. Canceling one caller releases its wait and cancels its generation request while other callers can finish. SDK abort errors unrelated to caller cancellation retain the normal provider error response.

Cached output keeps its original unknown cost on replay; a later uncached generation can retry unavailable pricing metadata. Zero-token categories no longer require unused pricing rates. This follows up #10802.

## Validation

- 301 affected tests across four suites passed, including test hygiene and four non-caller SDK abort regressions. Timer-dependent synchronization was replaced with a deterministic barrier.
- Root and app type checks, architecture checks, core and UI builds, postbuild, lint/format, and normal hooks passed.
- 22 actual IBM SDK loopback CLI cases passed, covering generation, metadata lookup/cancellation, and cache behavior. Exported successes and expected errors were inspected.
- Local checks used a qualified existing dependency tree because Socket blocks the locked Joi package. No vendor inference or live pricing calls were made.

### DIFF
--- a/site/docs/providers/watsonx.md
+++ b/site/docs/providers/watsonx.md
@@ -232,7 +232,7 @@ providers:
 
 ### Cost estimates
 
-Cost estimates use pricing metadata for the configured service and account. If metadata is unavailable or a pricing tier is unrecognized, the response omits `cost` and still returns the generated output.
+Cost estimates use pricing metadata for the configured service and account. If required metadata is unavailable or a required pricing tier is unrecognized, the response omits `cost` and still returns the generated output. A token category with a reported count of zero does not require a price.
 
 For account-specific pricing, set `inputCost` and `outputCost` in USD per token, or set `cost` for a shared input/output rate. These values override the built-in tier prices; an explicit zero is supported. Estimates do not determine whether your account can use a model.
 
--- a/src/providers/watsonx.ts
+++ b/src/providers/watsonx.ts
@@ -5,7 +5,6 @@ import { getCache, isCacheEnabled } from '../cache';
 import { getEnvString } from '../envars';
 import logger from '../logger';
 import { type GenAISpanContext, type GenAISpanResult, withGenAISpan } from '../tracing/genaiTracer';
-import { isAbortError } from '../util/fetch/errors';
 import invariant from '../util/invariant';
 import { createEmptyTokenUsage } from '../util/tokenUsageUtils';
 import { getRequestTimeoutMs, parseChatPrompt } from './shared';
@@ -769,9 +768,6 @@ export class WatsonXProvider implements ApiProvider {
       return providerResponse;
     } catch (err) {
       throwIfAborted(signal);
-      if (isAbortError(err)) {
-        throw err;
-      }
       logger.error(`Watsonx: API call error: ${String(err)}`);
 
       return {
@@ -865,9 +861,6 @@ export class WatsonXChatProvider extends WatsonXProvider {
       return providerResponse;
     } catch (err) {
       throwIfAborted(signal);
-      if (isAbortError(err)) {
-        throw err;
-      }
       logger.error(`Watsonx Chat: API call error: ${String(err)}`);
 
       return {
--- a/src/providers/watsonx.ts
+++ b/src/providers/watsonx.ts
@@ -5,6 +5,7 @@ import { getCache, isCacheEnabled } from '../cache';
 import { getEnvString } from '../envars';
 import logger from '../logger';
 import { type GenAISpanContext, type GenAISpanResult, withGenAISpan } from '../tracing/genaiTracer';
+import { isAbortError } from '../util/fetch/errors';
 import invariant from '../util/invariant';
 import { createEmptyTokenUsage } from '../util/tokenUsageUtils';
 import { getRequestTimeoutMs, parseChatPrompt } from './shared';
@@ -16,6 +17,7 @@ import type { EnvOverrides } from '../types/env';
 import type {
   ApiProvider,
   CallApiContextParams,
+  CallApiOptionsParams,
   ProviderResponse,
   TokenUsage,
 } from '../types/index';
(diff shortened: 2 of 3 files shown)


test/providers/watsonx-cost.test.ts
```diff
@@ -64,6 +72,51 @@ function provider(chat = false, config: Record<string, unknown> = {}) {
...
+  let reject!: (error: Error) => void;
+  const promise = new Promise<T>((resolvePromise, rejectPromise) => {
+    resolve = resolvePromise;
+    reject = rejectPromise;
+  });
+  return { promise, resolve, reject };
+}
+
+async function expectCallerAbort(observed: Promise<unknown>) {
+  const stillPending = new Promise<undefined>((resolve) => setTimeout(resolve, 1));
```
jameshiester-oai: [P1] Remove the timer-based sentinel in this new helper. The root `test/test-hygiene.test.ts` gate flags this exact `new Promise(... setTimeout(...))` line as an unapproved sleep wait; the current-head test matrix fails on Ubuntu, macOS, and Windows despite the WatsonX tests passing. I reproduced the hygiene failure locally. Use the existing fake-timer/deferred approach without a real timer, then rerun the root hygiene test.

src/providers/watsonx.ts
```diff
@@ -5,6 +5,7 @@ import { getCache, isCacheEnabled } from '../cache';
 import { getEnvString } from '../envars';
 import logger from '../logger';
 import { type GenAISpanContext, type GenAISpanResult, withGenAISpan } from '../tracing/genaiTracer';
+import { isAbortError } from '../util/fetch/errors';
```
jameshiester-oai: [P1] Resolve the new providers-to-node architecture edge introduced by this import. `src/util/fetch/errors.ts` belongs to the node layer, and the current-head Style Check fails `architecture:check` with `providers -> node: 302 imports (was 301)`; the same failure reproduces locally with `node --import tsx scripts/checkArchitectureBoundaries.ts`. Keep the abort classifier in an allowed provider/shared layer or justify and update the architecture baseline if the dependency is intentional.

jameshiester-oai (commented):
Reviewed exact head d7e0c477b094601114e83a8ca25b79b5b9413a62. Focused WatsonX tests pass (182), but current-head CI is red for two deterministic regressions from this diff; I reproduced both checks locally. The SDK-origin AbortError concern is already covered by an earlier inline comment, so I have not repeated it. Native PR-review toolkit was unavailable (exit 127); this is the documented manual fallback.

src/providers/watsonx.ts
```diff
@@ -667,14 +757,21 @@ export class WatsonXProvider implements ApiProvider {
...
+        throwIfAborted(signal);
       }
 
       return providerResponse;
     } catch (err) {
+      throwIfAborted(signal);
+      if (isAbortError(err)) {
+        throw err;
+      }
       logger.error(`Watsonx: API call error: ${String(err)}`);
```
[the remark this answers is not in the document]
mldangelo-oai: Fixed in f3bab0e for both generation and chat: only an aborted caller signal takes the cancellation path. Other SDK AbortErrors return the normal provider error response. Four regression cases cover both modes with and without a non-aborted signal; they also verify no cache write, metadata lookup, timer, or abort-listener leak. The affected WatsonX and hygiene suites pass.

test/providers/watsonx-cost.test.ts
mldangelo-oai: Fixed in f3bab0e: the timer race is replaced by deterministic promise observation with fake-time advancement. The test-hygiene guard and affected WatsonX suites pass (301 tests across four files).

src/providers/watsonx.ts
mldangelo-oai: Fixed in f3bab0e: removing the non-caller AbortError rethrow also removes the cross-layer import. Caller cancellation still takes precedence; four SDK-error regressions cover both provider modes. Architecture checks pass without changing the baseline.

jameshiester-oai (commented):
Reviewed base `18b90c224a6ca1b3d47789df9a539f65199bcc74` and full head `f3bab0ed392cea838944a91993fbb70843c7320c`, including the incremental change since authenticated reviewed ancestor `d7e0c477b094601114e83a8ca25b79b5b9413a62`. **No new actionable finding.** The prior P1 test-hygiene timer and architecture import failures are fixed; the existing Copilot SDK `AbortError` finding is fixed by letting the caller signal determine cancellation and returning ordinary provider failures for unrelated SDK aborts.

QA: `git diff --check`, `npm run architecture:check`, and app TypeScript passed locally. Four affected WatsonX/hygiene files passed 301 tests (including caller cancellation, SDK failure, and the root test-hygiene gate); my overlapping three-file run passed 273 tests. The installed IBM SDK forwards the `signal` to Axios for both generation and chat. At publication, 36 reported CI checks had succeeded and 7 remained in progress, with no current-head failure reported. No live IBM endpoint or credentials were used; transport behavior was checked in the installed SDK source and by local mocks. `/pr-review-toolkit:review-pr all parallel` was unavailable (exit 127); manual correctness/API, QA, independent security/risk, docs/error/type/simplification review substituted.

site/docs/providers/watsonx.md
```diff
@@ -232,7 +232,7 @@ providers:
 
 ### Cost estimates
 
-Cost estimates use pricing metadata for the configured service and account. If metadata is unavailable or a pricing tier is unrecognized, the response omits `cost` and still returns the generated output.
+Cost estimates use pricing metadata for the configured service and account. If required metadata is unavailable or a required pricing tier is unrecognized, the response omits `cost` and still returns the generated output. A token category with a reported count of zero does not require a price.
```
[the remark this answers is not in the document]
mldangelo-oai: The commit and PR title are different here. `f3bab0e` is `fix(watsonx): preserve non-caller SDK abort errors` and changes only the provider and its tests. The documentation change is in `4d75c9b`, already titled `docs(site): clarify WatsonX zero-token pricing`. The PR title follows its primary runtime purpose under the mixed-change conventions; no commit rewording is needed.

PR #6253, opened 2025-11-19
fix(assertions): use script output for file:// references in all assertion types

mldangelo:
## Summary

Fixes #6200

When using file-based scripts (JavaScript, Python, Ruby) as assertion values via `file://` references, the script output was correctly executed but **not used** as the actual assertion value. Instead, the literal file path string was passed to the assertion handler.

This PR adds centralized script value resolution that updates `renderedValue` with the script output for all non-script assertion types.

## Problem

**Before this fix:**
```yaml
assert:
  - type: llm-rubric
    value: file://rubric.js:generateRubric
```

The `matchesLlmRubric` function received `"file://rubric.js:generateRubric"` as the rubric text instead of the actual script output.

**After this fix:**
The script is executed and its return value (e.g., `"Check that the response covers machine learning basics"`) is passed to the handler.

## Implementation

Added centralized script output resolution in `src/assertions/index.ts` after script execution:

```typescript
const SCRIPT_RESULT_ASSERTIONS = new Set(['javascript', 'python', 'ruby']);
if (valueFromScript !== undefined && !SCRIPT_RESULT_ASSERTIONS.has(baseType)) {
  renderedValue = valueFromScript as AssertionValue;
}
```

**Why exclude javascript/python/ruby?** These assertion types interpret `renderedValue` as executable code, not as a comparison value. They use `valueFromScript` as the assertion result directly.

## Affected Assertion Types

All assertion types that use `renderedValue` as a comparison value:
- `llm-rubric`, `model-graded-closedqa`, `factuality`, `g-eval`
- `contains`, `equals`, `similar`, `regex`, `starts-with`, `levenshtein`
- `bleu`, `gleu`, `rouge`, `webhook`, `pi`, `classifier`

### 1. Reproduction Test

Create a test script:
```bash
mkdir -p tmp && cat <<'SCRIPT' > tmp/rubric.js
module.exports.rubric = () => 'script says abc';
SCRIPT
```

Run this verification:
```bash
node <<'NODE'
const cliState = require('./dist/src/cliState.js').default;
cliState.basePath = process.cwd();
const { runAssertion } = require('./dist/src/assertions/index.js');
const matchers = require('./dist/src/matchers.js');
matchers.matchesLlmRubric = async (...args) => {
  console.log('matchesLlmRubric rubric param:', args[0]);
  return { pass: true, score: 1, reason: 'stub' };
};
(async () => {
  await runAssertion({
[...]

PR #10871, opened 2026-09-09
fix(assertions): reject file:// paths escaping basePath

### LINKED ISSUE #10869: file:// assertion values and rubricPrompt can escape basePath via ../
**Describe the bug**

`file://` values in `assert[].value` and in `rubricPrompt` are resolved with a plain `path.resolve(basePath, filePath)` and nothing else. There's no containment check, so a `../` in the value walks the resolved path outside `basePath`. Depending on the file extension this hits, that's either an arbitrary local file read (`.json`/`.yaml`/`.txt` values get parsed and returned) or arbitrary code execution (`.js` gets `require`'d, `.py`/`.rb` get spawned).

The codebase already has a guard for this exact shape of problem — `resolveCallbackPath()` in `src/util/functions/loadFunction.ts`. It's used by `loadCallbackFromFileUrl()` for provider tool-callback `file://` references: lexical `..` rejection, a symlink-aware realpath re-check, a dedicated `CallbackPathTraversalError`, and a documented opt-out (`PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD`). It just isn't wired up to the other three places that resolve a `file://` value the same way:

- `src/assertions/index.ts` — the `file://` dispatch for assertion values (js/py/rb/json/yaml/txt)
- `src/assertions/utils.ts` — `processFileReference()`, used for the non-script branch and array-of-values assertions
- `src/matchers/rubric.ts` — `loadRubricPrompt()`

Note on scope: per `SECURITY.md`, file-reference values dereferenced by configured eval steps and code-executing assertion fields are generally treated as trusted/explicitly-configured, so I'm not filing this as a security advisory — just a normal bug, since the exact same protection already exists in this codebase for the sibling callback-loading path and isn't applied consistently here.

**To Reproduce**

```yaml
# promptfooconfig.yaml
prompts: ['{{ output }}']
providers: ['echo']
tests:
  - vars: { output: irrelevant }
    assert:
      - type: equals
        value: 'file://../../../../../../etc/passwd'
```

Running `promptfoo eval` on this reads `/etc/passwd` and compares its contents against the output. No warning, no error.

Swapping the extension for a `.js` file placed outside the project directory gets it `require`'d and executed instead of just read — confirmed locally with a throwaway script that writes a proof-of-execution file to disk when loaded via `assert: { type: javascript, value: 'file://../../outside/evil.js' }`.

**Expected behavior**

A `file://` value that resolves outside `basePath` should be rejected the same way a provider callback `file://` reference already is — a clear error (or a failed assertion) explaining that the path escapes the base directory, not a silent load.

**System information**

- Promptfoo version: 0.122.2
- Commit: `0b4ccbe`
- Install method: local clone

**Additional context**

Fix + tests: route `filePath` in all three call sites through the existing `resolveCallbackPath()` instead of raw `path.resolve()`. In `src/assertions/index.ts`, catch `CallbackPathTraversalError` and return a normal `pass: false` result (matching how the neighboring `.py`/`.rb` branches already handle their own runtime errors) rather than letting it throw uncaught. See the accompanying PR.

mosafariuk:
## Summary

`file://` assertion values (`assert[].value`) and `rubricPrompt` are resolved with a bare `path.resolve(basePath, filePath)` — no containment check. A `../` in the value walks the resolved path outside `basePath`, and depending on extension it gets read as a file, `require`'d as a JS module, or handed to the Python/Ruby runners.

We already have the fix for this exact shape of bug — `resolveCallbackPath()` in `src/util/functions/loadFunction.ts`, built for the provider-callback `file://` loader. Lexical `..` rejection, symlink-aware realpath re-check, `CallbackPathTraversalError`, documented `PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD` opt-out. This just points the two other `file://` resolution sites (assertion dispatch, rubric prompt) at it instead of duplicating the logic.

Closes #10869

## Changes

- `src/assertions/index.ts`: the shared `filePath` resolution used by the js/py/rb branches now goes through `resolveCallbackPath`. A caught `CallbackPathTraversalError` becomes a normal `pass: false` assertion result instead of an uncaught throw, matching how the neighboring `.py`/`.rb` catch blocks already behave.
- `src/assertions/utils.ts`: `processFileReference()` (the JSON/YAML/txt non-script branch, and the array-of-values branch) uses the same guard.
- `src/matchers/rubric.ts`: `loadRubricPrompt()`'s file resolution uses the same guard. Errors propagate as a throw here, consistent with the existing "File does not exist" throw a few lines above it.
- `path` import dropped from `index.ts` and `rubric.ts` (no longer used after the change); kept in `utils.ts` for `path.extname`.

## Testing

- `npx tsc --noEmit` — clean
- `npx vitest run test/assertions test/matchers test/util/functions` — 1804/1804 passing
- `npx vitest run test/assertions/assertionsResult.test.ts test/assertions/runAssertions.test.ts test/assertions/trace.test.ts test/evaluator/assertions.test.ts` (shared-aggregation regression sweep) — 94/94 passing
- Added traversal-rejection tests at all three call sites: unit-level on `processFileReference` (`test/assertions/utils.test.ts`), end-to-end through `runAssertion` for the js/non-script/array-value branches (`test/assertions/scriptValueResolution.test.ts`), and `matchesLlmRubric`'s `rubricPrompt` (`test/matchers/llm-rubric.test.ts`)
- Manually reproduced both the file-read and the code-execution case against the pre-fix code (sandboxed `basePath`, file/script planted outside it, confirmed read/execution succeeded), then confirmed the fix blocks both with `Path traversal rejected: ...`
- `PROMPTFOO_DISABLE_CALLBACK_PATH_GUARD=true` still works as an escape hatch, since it's the same guard function already shipped for the callback-loader path

### DIFF
--- a/src/assertions/index.ts
+++ b/src/assertions/index.ts
@@ -1,5 +1,4 @@
 import fs from 'fs/promises';
-import path from 'path';
 
 import async from 'async';
 import cliState from '../cliState';
@@ -38,6 +37,7 @@ import {
   type VarValue,
 } from '../types/index';
 import { isJavascriptFile } from '../util/fileExtensions';
+import { CallbackPathTraversalError, resolveCallbackPath } from '../util/functions/loadFunction';
 import invariant from '../util/invariant';
 import { getNunjucksEngine } from '../util/templates';
 import { sleep } from '../util/time';
@@ -490,7 +490,19 @@ async function runAssertionInternal({
         functionName = fileRef.slice(colonIndex + 1);
       }
 
-      filePath = path.resolve(basePath, filePath);
+      try {
+        filePath = resolveCallbackPath(filePath, basePath);
+      } catch (error) {
+        if (error instanceof CallbackPathTraversalError) {
+          return {
+            pass: false,
+            score: 0,
+            reason: error.message,
+            assertion,
+          };
+        }
+        throw error;
+      }
 
       if (isJavascriptFile(filePath)) {
         valueFromScript = await loadFromJavaScriptFile(filePath, functionName, [output, context]);
--- a/src/assertions/utils.ts
+++ b/src/assertions/utils.ts
@@ -5,6 +5,7 @@ import Clone from 'rfdc';
 import cliState from '../cliState';
 import { importModule } from '../esm';
 import { type Assertion, type TestCase } from '../types/index';
+import { resolveCallbackPath } from '../util/functions/loadFunction';
 import { loadYaml } from '../util/yamlLoad';
 
 const clone = Clone();
@@ -58,7 +59,7 @@ export async function loadFromJavaScriptFile(
 
 export function processFileReference(fileRef: string): object | string {
   const basePath = cliState.basePath || '';
-  const filePath = path.resolve(basePath, fileRef.slice('file://'.length));
+  const filePath = resolveCallbackPath(fileRef.slice('file://'.length), basePath);
   const fileContent = fs.readFileSync(filePath, 'utf8');
   const extension = path.extname(filePath);
   if (['.json', '.yaml', '.yml'].includes(extension)) {
--- a/src/matchers/rubric.ts
+++ b/src/matchers/rubric.ts
@@ -1,5 +1,4 @@
 import fs from 'fs/promises';
-import path from 'path';
 
 import { loadFromJavaScriptFile } from '../assertions/utils';
 import cliState from '../cliState';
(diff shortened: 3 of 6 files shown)


mosafariuk:
Closing this — the approach is wrong, and CI caught it.

The guard rejects `file://` values that resolve outside `basePath`, but pointing an assertion value at a sibling directory is a normal and supported layout. `test/smoke/fixtures/configs/file-ref-assertion-value.yaml` does exactly that with `value: file://../assertions/dynamic-value.js`, and it exists as a regression test for #6253. On this branch all three of its cases fail with "Path traversal rejected"; on `main` all three pass. So this breaks documented behaviour.

That isn't fixable by adjusting the guard, because base-directory containment is the wrong model for a field where sibling-directory references are legitimate. And per `SECURITY.md`, `file://` values dereferenced by configured eval steps are explicitly trusted and out of scope, so there's no security benefit being traded for that break.

Two smaller failures for the record, both also mine: the new imports push the `core -> node` architecture edge from 90 to 92, and removing the `path` mock from `test/assertions/utils.test.ts` left POSIX-only expectations that fail on Windows.

Closing the linked issue as well.

PR #10880, opened 2026-09-10
fix(assertions): preserve custom reason in not-javascript GradingResult inversion

### LINKED ISSUE #10674: Reason not respected for not-javascript assetion returning a GradingResult
**Describe the bug**
When a javascript assertion's value function returns a full GradingResult object (not just a boolean/number) and the assertion type is inverted via not-javascript, the custom reason is silently discarded:

see https://github.com/promptfoo/promptfoo/blob/main/src/assertions/javascript.ts#L188

For the pass === false case it takes the reason from the GradingResult but doesn't for the true case. Which isn't so bad normally. But isn't good for the inverted case. where pass === true at this point actually means it failed.

**To Reproduce**

```
const assertion = {
  type: "not-javascript",
  value: (output) => ({
    pass: output.includes("foo"),
    score: output.includes("foo") ? 1 : 0,
    reason: 'Expected output not to contain "foo", but it did.',
  }),
};
```

output = "foo bar"
→ result.pass = true, inverse = true, so outer pass = false (correctly fails)
but reason comes back as "Custom function returned true" instead of the custom message
Expected: the custom reason should be preserved (or at least prefixed/suffixed with a note about negation) rather than unconditionally replaced when the function returns a full GradingResult.

**Expected behavior**

Personally I think the inverted case should always return "NOT: " + reason. regardless of whether it passed. But if you wanted to keep it the same, you could make it only includes reason when it fails (the function returns pass = true)

vishalchincholi1:
## Summary

When a javascript assertion value function returns a full GradingResult object and the assertion type is 
ot-javascript, the custom 
eason from the result was silently discarded and replaced with the generic message "Custom function returned true".

### Root cause

In 
ormalizeJavascriptAssertionResult the GradingResult branch unconditionally replaced the reason with 'Assertion passed' when inversion turned pass: false into pass: true — even though the user provided a meaningful message.

### Fix

- When inversion does **not** change the outcome (same pass/fail), keep the original 
eason as-is.
- When inversion turns a **function-pass into a test-fail** (the unwanted condition was met), prefix the reason with "NOT: " so the user sees why the negated assertion failed.
- When inversion turns a **function-fail into a test-pass**, surface the original reason (the assertion succeeded because the bad condition was absent).

### Testing

- Updated two existing test cases whose expected values reflected the old broken behaviour ('Custom function returned true' → 'NOT: Custom reason').
- Added three new focused test cases in a dedicated describe block covering all three inversion paths.
- All 95 tests pass. Lint clean (Biome).

### How to reproduce (before fix)

`	s
const assertion = {
  type: 'not-javascript',
  value: (output) => ({
    pass: output.includes('foo'),
    score: output.includes('foo') ? 1 : 0,
    reason: 'Expected output not to contain "foo", but it did.',
  }),
};
// With output = 'foo bar':
// result.pass === false  ✓  (correctly fails)
// result.reason === 'Custom function returned true'  ✗  (custom reason lost)
`

After this fix 
esult.reason is 'NOT: Expected output not to contain "foo", but it did.'

Closes #10674

### DIFF
--- a/site/docs/configuration/expected-outputs/javascript.md
+++ b/site/docs/configuration/expected-outputs/javascript.md
@@ -447,6 +447,26 @@ assert:
     value: output.includes('error')
 ```
 
+### `GradingResult` reason behavior with `not-javascript`
+
+When your function returns a full `GradingResult` object, the `reason` field is always preserved verbatim regardless of whether inversion changes the outcome. This lets you write negation-aware messages directly in the function:
+
+```yaml
+assert:
+  - type: not-javascript
+    value: |
+      const hasFoo = output.includes('foo');
+      return {
+        pass: hasFoo,
+        score: hasFoo ? 1 : 0,
+        reason: hasFoo
+          ? 'Output contained "foo" — this is forbidden'
+          : 'Output correctly omitted "foo"',
+      };
+```
+
+If the function returns a plain `boolean` or `number`, the standard failure message (`Custom function returned true/false`) is used when the assertion fails.
+
 ## Other assertion types
 
 For more info on assertions, see [Test assertions](/docs/configuration/expected-outputs).
--- a/src/assertions/javascript.ts
+++ b/src/assertions/javascript.ts
@@ -190,17 +190,18 @@ function normalizeJavascriptAssertionResult(
   // The user wrote the reason as explanatory prose — mechanically prefixing it
   // (e.g. "NOT: …") can make it read as the opposite of what it describes.
   // Keeping it verbatim lets the user craft their own negation-aware message.
+  // Use ?? (not ||) so that an explicit empty-string reason is also preserved.
   let reason: string;
   if (pass === result.pass) {
     // Inversion did not change the outcome — keep the original reason as-is.
     reason = result.reason;
   } else if (pass) {
     // Inversion turned a function-fail into a test-pass: assertion succeeded.
-    reason = result.reason || 'Assertion passed';
+    reason = result.reason ?? 'Assertion passed';
   } else {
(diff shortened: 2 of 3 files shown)


vishalchincholi1:
Both points addressed in the follow-up commit:

1. **Reason preserved verbatim** — removed the `NOT:` prefix. The `reason` is arbitrary explanatory prose; mechanically negating it can make it describe the opposite of the failure. Users who want inversion-aware messages can write them directly in their function.

2. **Documentation added** — added a `GradingResult reason behavior with not-javascript` subsection to `site/docs/configuration/expected-outputs/javascript.md` explaining that reasons are preserved verbatim and showing a concrete example.

vishalchincholi1:
New Codex comment addressed in follow-up commit:

- **Empty-string reason preservation**: Changed `||` to `??` in both inversion branches so that `reason: ""` is preserved verbatim. The `||` operator treated an empty string as falsy and replaced it with `"Assertion passed"` or the generic fallback, violating the documented verbatim-preservation guarantee. A new test case (`reason: \"\"`) asserts this.

vishalchincholi1:
New Codex comment addressed in follow-up commit:

- **Empty-string reason preservation**: Changed `||` to `??` in both inversion branches so that `reason: ''` is preserved verbatim. The `||` operator treated an empty string as falsy and replaced it with `Assertion passed` or the generic fallback, violating the documented verbatim-preservation guarantee. A new test case (`reason: ''`) asserts this.

vishalchincholi1:
New Codex comment addressed: changed || to ?? so that reason:'' is preserved verbatim, added test asserting empty-string reason survives inversion unchanged.

PR #10881, opened 2026-09-10
fix(fetch): treat credit_balance_exhausted as a hard quota error

### LINKED ISSUE #10855: OpenAI `credit_balance_exhausted` is not treated as a hard quota error, one call retries for ~20 min
**Describe the bug**
An OpenAI account with no credit returns HTTP 429 with:
```json
{"error":{"message":"You have no credits remaining ...","type":"insufficient_quota","code":"credit_balance_exhausted"}}
```
`HARD_QUOTA_ERROR_CODES` does not contain `credit_balance_exhausted`, so the fail-fast path never fires. The 429 has no `retry-after` header, so each retry waits the 60s fallback. One test against a dead key takes ~20 minutes to fail, and the message is the generic one:
```
Rate limit exceeded: HTTP 429 Too Many Requests (code: credit_balance_exhausted)
```

**To Reproduce**
1. OpenAI key with prepaid balance 0 (auto-recharge off).
2. Config:
   ```yaml
   providers:
     - openai:chat:gpt-4o-mini
   prompts:
     - "Say hi"
   tests:
     - assert:
         - type: contains
           value: hi
   ```
3. `promptfoo eval`
4. `Duration:` line shows 23m15s on 0.121.3 and 19m28s on 0.122.0 for the single test. Same on 0.122.2.

**Expected behavior**
Fail immediately with the existing `Quota exceeded ... Retries will not help — check your billing` message, like `insufficient_quota` does.

**Where the bug is**
`src/util/fetch/errors.ts`: `HARD_QUOTA_ERROR_CODES` has `insufficient_quota`, `billing_hard_limit_reached`, `billing_not_active`, `access_terminated`, `quota_exceeded`. `extractRateLimitErrorCode()` returns `error.code` first and only reads `error.type` when `code` is empty, so the `insufficient_quota` in `type` is never consulted.

**Additional context**
The response body is not shown at default log level, so a CI log gives no hint that it is a billing problem rather than throttling.

vishalchincholi1:
## Summary

OpenAI returns HTTP 429 with rror.code = "credit_balance_exhausted" when a prepaid account has no credits remaining. This code was missing from HARD_QUOTA_ERROR_CODES in src/util/fetch/errors.ts, so the fail-fast path never fired.

### Impact

Because the 429 carries no Retry-After header, each retry used the 60-second fallback. A single test against a depleted key took **~20 minutes** before eventually failing with a misleading generic rate-limit message (Rate limit exceeded: HTTP 429 ... (code: credit_balance_exhausted)), with no hint that it was a billing problem.

### Fix

Add 'credit_balance_exhausted' to HARD_QUOTA_ERROR_CODES. The error is now classified as kind: 'quota', surfaces the billing-focused message (Quota exceeded … Retries will not help — check your billing or daily quota.), and aborts immediately.

### Testing

- Added 'credit_balance_exhausted' to the isHardQuotaCode parametrised test table.
- Added a dedicated HttpRateLimitError test verifying kind === 'quota' and the correct message prefix.
- All 83 tests pass. Lint clean (Biome).

### To reproduce (before fix)

`yaml
providers:
  - openai:chat:gpt-4o-mini
prompts:
  - "Say hi"
tests:
  - assert:
      - type: contains
        value: hi
`

Run promptfoo eval with an OpenAI key whose prepaid balance is 0. Before this fix the eval runs for ~20 minutes. After this fix it fails immediately.

Closes #10855

### DIFF
--- a/src/util/fetch/errors.ts
+++ b/src/util/fetch/errors.ts
@@ -28,6 +28,11 @@ export const HARD_QUOTA_ERROR_CODES: ReadonlySet<string> = new Set([
   'billing_not_active',
   'access_terminated',
   'quota_exceeded',
+  // Returned by OpenAI when a prepaid credit balance is fully consumed
+  // (HTTP 429, no Retry-After header). Without this entry the error falls
+  // through to the per-window retry loop and a single test can take ~20 min
+  // before failing. References: https://github.com/promptfoo/promptfoo/issues/10855
+  'credit_balance_exhausted',
 ]);
 
 /**
--- a/test/util/fetch/errors.test.ts
+++ b/test/util/fetch/errors.test.ts
@@ -222,6 +222,7 @@ describe('isHardQuotaCode', () => {
     ['billing_not_active', true],
     ['access_terminated', true],
     ['quota_exceeded', true],
+    ['credit_balance_exhausted', true],
     ['rate_limit_exceeded', false],
     ['tokens_per_min', false],
     ['', false],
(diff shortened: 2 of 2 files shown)


mldangelo-oai (approved):
Thank you, Vishal! This is a useful, focused fix: an exhausted prepaid balance should fail promptly instead of spending minutes on retries that cannot succeed. I appreciate the regression test covering both the quota classification and the error message.

I reviewed bde3825603e341a15ab7f54ee73c20077d40111f and verified the behavior with the local CLI. The 217 focused tests and TypeScript check passed. Thanks for contributing this improvement!

mldangelo-oai:
@codex review

Please review the current head, 1727aff2b5c8f04f30a8f5ae1c80e4eaee4a4d8b, after merging main. The PR still contains the original two-file quota fix.

PR #10883, opened 2026-09-10
docs(model-graded): document defaultTest.provider as undocumented grader fallback

### LINKED ISSUE #10581: Model-graded assertions fall back to `defaultTest.provider` — the provider under test — when no grader is set, and the grader-override docs don't mention it
Read at `3f94284`.

**What happens**

When a model-graded assertion (`llm-rubric`, `factuality`, `model-graded-closedqa`, …) has no grader configured — no `--grader`, no `defaultTest.options.provider`, no assertion-level `provider` — `getGradingProvider` in `src/matchers/providers.ts:196-202` does this before falling back to the dedicated grading default:

```ts
// No provider specified - check defaultTest providers as fallback
const fallbackProviders = [
  defaultTestObj?.provider || undefined,
  defaultTestObj?.options?.provider?.text || undefined,
  defaultTestObj?.options?.provider || undefined,
];
```

`defaultTest.provider` is the **provider override for the test itself** — `src/types/index.ts:879` documents the field as `// Override the provider.`, `src/evaluator.ts:2594` stamps it onto every test case, and `src/evaluator.ts:1069` sends the prompt to `test.provider`. So this config:

```yaml
defaultTest:
  provider: openai:gpt-4.1
tests:
  - assert:
      - type: llm-rubric
        value: "Answers the question accurately"
```

generates with `gpt-4.1` and grades with `gpt-4.1`. The only signal is `logger.debug('[Grading] Using provider from defaultTest fallback')` at `:222`.

`DEFAULT_OPENAI_GRADING_MODEL` (`src/providers/openai/defaults.ts:6`) is real and is a genuinely separate grading constant — but on this path it is reached only at `providers.ts:227`, after the `defaultTest` list is exhausted.

**The docs gap**

`site/docs/configuration/expected-outputs/model-graded/index.md`, section *"Overriding the LLM grader"*, lists `--grader`, `defaultTest.options.provider`, and assertion-level `provider`. It does not mention that `defaultTest.provider` is consulted first, so a reader who sets it to pin the target has no way to know they have also pinned the judge.

**Scope**

This only fires when `defaultTest.provider` is set. With a top-level `providers:` list and no `defaultTest.provider`, the same function falls through to the credential-selected grading default as documented. And `options.provider.text` / `options.provider` in the same list are legitimate grader slots — the surprising entry is the first one.

**Reproduce**

```
git checkout 3f94284
sed -n '196,202p' src/matchers/providers.ts
grep -n 'Override the provider' src/types/index.ts
grep -c 'defaultTest.provider' site/docs/configuration/expected-outputs/model-graded/index.md   # 0
```

**Suggested fix**

Any one of: drop `defaultTestObj?.provider` from the grader fallback list (keep the two `options.provider` entries, which are explicit grader slots); or promote the `:222` log to `warn` when the resolved grader is the same provider as the test's; or add one line to "Overriding the LLM grader" saying `defaultTest.provider` is used as the grader when nothing else is set.

vishalchincholi1:
## Summary

The **"Overriding the LLM grader"** section in the model-graded docs listed three explicit grader configuration points:

1. --grader CLI flag
2. defaultTest.options.provider
3. ssertion.provider

It did not mention that **defaultTest.provider** — the field used to pin the *target model* for every test — is also consulted as a grader fallback **before** those slots (see src/matchers/providers.ts lines 196–202).

### Impact

A user who sets defaultTest.provider to pin their target model unknowingly pins the judge to the same model. The only signal is a logger.debug line buried in the grading provider resolution. This is especially surprising in red-team / safety configs where the target and judge should be independent.

Example that hits the undocumented behaviour:

`yaml
defaultTest:
  provider: openai:gpt-4.1   # ← also becomes the LLM judge
tests:
  - assert:
      - type: llm-rubric
        value: Answers the question accurately
`

### Fix

Add a :::caution callout block immediately after the numbered grader override list explaining:

- defaultTest.provider doubles as a grader fallback
- A concrete YAML example showing the dual role
- The recommended workaround: use defaultTest.options.provider or --grader

No code changes — documentation only.

Closes #10581

### DIFF
--- a/site/docs/configuration/expected-outputs/model-graded/index.md
+++ b/site/docs/configuration/expected-outputs/model-graded/index.md
@@ -260,6 +260,31 @@ override the grader. There are several ways to do this, depending on your prefer
            provider: openai:gpt-5.6
    ```
 
+:::caution `defaultTest.provider` also sets the grader
+
+`defaultTest.provider` is the field that pins the **target model** for every test in a suite. As a
+side-effect it is also consulted as a grader fallback — before the dedicated
+`defaultTest.options.provider` slot — when no explicit grader is configured.
+
+This means the following config generates responses **and** grades them with `gpt-4.1`:
+
+```yaml
+defaultTest:
+  provider: openai:gpt-4.1 # ← also becomes the judge
+tests:
+  - assert:
+      - type: llm-rubric
+        value: Answers the question accurately
+```
+
+To use a dedicated judge while still pinning the target, set `defaultTest.options.provider`
+separately (option 2 above) or use `--grader` on the CLI. Only the
+`defaultTest.options.provider` / `test.options.provider` slots are documented grader
+configuration points; `defaultTest.provider` is an undocumented fallback that exists purely
+for backwards compatibility.
+
+:::
+
 Use the `provider.config` field to set custom parameters such as `temperature`, `max_tokens`, or API host:
 
 ```yaml
--- a/site/docs/configuration/expected-outputs/model-graded/index.md
+++ b/site/docs/configuration/expected-outputs/model-graded/index.md
@@ -260,17 +260,19 @@ override the grader. There are several ways to do this, depending on your prefer
            provider: openai:gpt-5.6
    ```
 
-:::caution `defaultTest.provider` also sets the grader
+:::caution `defaultTest.provider` also sets the grader for output-based assertions
 
-`defaultTest.provider` is the field that pins the **target model** for every test in a suite. As a
-side-effect it is also consulted as a grader fallback — before the dedicated
-`defaultTest.options.provider` slot — when no explicit grader is configured.
+`defaultTest.provider` is the field that pins the **target model** for every test in a suite.
+For output-based model-graded assertions (`llm-rubric`, `factuality`, `g-eval`,
+`model-graded-closedqa`, `answer-relevance`, etc.) it is also consulted as a grader fallback when
+no explicit grader is configured — **after** `--grader`, `assertion.provider`, and
+`test.options.provider` / `defaultTest.options.provider` have all been checked and found absent.
 
-This means the following config generates responses **and** grades them with `gpt-4.1`:
+In practice this means the following config generates responses **and** grades them with `gpt-4.1`:
 
 ```yaml
 defaultTest:
-  provider: openai:gpt-4.1 # ← also becomes the judge
+  provider: openai:gpt-4.1 # ← also becomes the judge when no grader is set
 tests:
   - assert:
       - type: llm-rubric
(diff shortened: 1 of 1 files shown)


vishalchincholi1:
All three review points addressed in the follow-up commit:

1. **Precedence wording fixed** — reworded to say `defaultTest.provider` is the *last-resort* fallback, consulted only when `--grader`, `assertion.provider`, `test.options.provider`, and `defaultTest.options.provider` are all absent. Added a corrected example showing `defaultTest.options.provider` overriding it.

2. **Red-team caveat added** — noted that red-team runs use `RedteamProviderManager` which has a separate grader-selection path, and that the standard overrides may not apply there.

3. **Scope narrowed** — the warning now explicitly lists only the output-based assertions it affects (`llm-rubric`, `factuality`, `g-eval`, etc.) and notes that `agent-rubric` and `search-rubric` are not affected because they use capability-specific provider selection.

vishalchincholi1:
P1 Codex comment addressed in follow-up commit:

- **Effective red-team override documented**: Replaced the vague "consult the red-team docs" note with the concrete working pattern. Since `RedteamProviderManager.getGradingProvider` selects `defaultTest.provider` _before_ `defaultTest.options.provider`, the reliable override is to move the target to the top-level `providers` list and reserve `defaultTest.options.provider` for the judge. Added a YAML example showing exactly this.

vishalchincholi1:
P1 Codex comment addressed in follow-up commit:

- **Effective red-team override documented**: Replaced the vague 'consult the red-team docs' note with the concrete working pattern. Since `RedteamProviderManager.getGradingProvider` selects `defaultTest.provider` *before* `defaultTest.options.provider`, the reliable override is to move the target to the top-level `providers` list and reserve `defaultTest.options.provider` for the judge. Added a YAML example showing exactly this.

vishalchincholi1:
P1 comment addressed: replaced vague 'consult red-team docs' note with the concrete working pattern. RedteamProviderManager.getGradingProvider selects defaultTest.provider before defaultTest.options.provider, so the reliable override is to move the target to the top-level providers list and reserve defaultTest.options.provider for the judge. A YAML example was added.

vishalchincholi1:
P1 comment addressed: replaced the vague 'consult the red-team docs' note with the concrete working pattern. RedteamProviderManager.getGradingProvider selects defaultTest.provider before defaultTest.options.provider, so the reliable override is to move the target to the top-level providers list and reserve defaultTest.options.provider for the judge. A YAML example was added demonstrating this.

mldangelo-oai (approved):
Thank you!

mldangelo-oai:
thank you!

PR #10882, opened 2026-09-10
test(assertions): preserve grader failures in both directions

vishalchincholi1:
## Summary

Grader failures must remain failures for both normal and inverse classifier/search assertions, with their metadata, token counts, reason, and assertion details preserved. This PR retains Vishal Chincholi's normal/inverse coverage and Mohammad Safari's stronger full-result checks from #10872.

The production fixes and documentation landed in #10904 by Yufeng He. Main is merged into this branch, including its fixes for empty classification maps, search provider failures, and malformed search verdicts. The remaining diff against main contains three test files.

## Stack and contributor credit

- #10904 — Yufeng He: production fixes, malformed-response regressions, and documentation; merged.
- #10872 — Mohammad Safari: full-result regression checks for failure metadata, token counts, and assertion details.
- #10882 — Vishal Chincholi: coverage for both normal and inverse assertion directions, retained in this original PR.

Land #10872 before #10882. After #10872 lands, merge main here again so only the two handler-test changes remain. Original contributor commits are preserved through additive merges.

## Validation

- TypeScript passed on the merged tree.
- The local CLI ran 18 deterministic regression cases. Exported JSON verified 14 grader failures stayed failed with score 0, four valid verdict cases retained their intended normal/inverse behavior, and there were zero runtime errors. The command returned the expected exit 100 for two passes and 16 intentional failures.
- Full root suite on `54d323d4d03cd93852e282a4e4d36a9c8fdd4b27`: 25,643 tests passed, 10 skipped, across 859 passing files and one skipped file.
- After the latest main merge (`516980626e1a5520fe097334e5828b2291e37d25`): 959 tests passed across 12 files covering the graders, the new Agents API provider, registry, fetch, output, and sanitization. TypeScript and all 18 CLI cases passed again.
- Changed-file lint, formatting, diff checks, and pre-commit checks passed. No production files differ from main.

PR #10872, opened 2026-09-09
test(assertions): preserve full grader failure results

mosafariuk:
## Summary

Grader failures need to preserve their complete result, including failure metadata, token counts, reason, score, and assertion details. This follow-up retains and strengthens Mohammad Safari's regression checks for the classifier/search handlers and the classification matcher. The expected search result is independent of the mocked response so in-place mutation cannot hide a regression.

The production fix and documentation landed in #10904 by Yufeng He. Main is merged into this branch; only three test files differ. #10882 by Vishal Chincholi adds normal/inverse handler coverage afterward. Land this PR first, then merge main into #10882 to retain its distinct test increment. Original contributor commits are preserved.

## Validation

Validated published commit `c2273cd8c5b3ae4214c84b06a39953c49e9aed86`:

- Full root suite: 25,822 tests passed, 10 skipped; 860 passing files and one skipped file.
- TypeScript, changed-file lint/format, diff checks, and pre-commit checks passed.
- Actual local CLI: all 18 exported outcomes verified. Fourteen grader failures remained failures with score 0; four valid verdict cases preserved normal/inverse behavior. Two expected passes and 16 intentional failures produced exit 100, with zero runtime errors.
- [Hosted CI](https://github.com/promptfoo/promptfoo/actions/runs/34807271457) passed. Fresh [Codex code and security reviews](https://github.com/promptfoo/promptfoo/pull/10872#issuecomment-5659156465) completed on this exact head without findings.

The required CodeQL result is still unreported and remains a separate merge gate.

PR #10904, opened 2026-09-11
fix(assertions): never invert grader failures on not-classifier and not-search-rubric

### LINKED ISSUE #10870: not-classifier / not-search-rubric turn a grader/transport failure into a false pass
**Describe the bug**

When the classification API call inside a `classifier` assertion fails (timeout, bad key, malformed response — anything that isn't a real classification result), `matchesClassification()` returns a plain failure with no way to distinguish "the grader is broken" from "the content legitimately scored 0." For a `not-classifier` assertion, `handleClassifier()` unconditionally flips `pass`/`score` on any failure, including this one. Net result: a classification outage produces `pass: true, score: 1` on a `not-classifier` check. Same shape for `not-search-rubric` in `searchRubric.ts`/`search.ts`.

This codebase already has a convention to prevent exactly this: matchers tag a transport/API error with `metadata.graderError = true` via a `graderFail()` helper, and inverse-aware handlers check `isGraderFailure()` before flipping `pass`, since a failed API call isn't evidence the criterion was or wasn't met. `llmRubric.ts`, `geval.ts`, `moderation.ts`, `trajectory.ts`, and `agentRubric.ts` all do this correctly (several with comments spelling out why). `classifier.ts` and `searchRubric.ts` are the two handlers that support inversion but never got the guard.

There was an existing test for inversion (`test/assertions/classifier.test.ts`) but it only covered a legitimate low-score fail, never the error case, so this had no regression coverage.

**To Reproduce**

```ts
import { runAssertion } from './src/assertions/index';

const brokenProvider = {
  id: () => 'broken-classification-provider',
  callApi: async () => ({ output: '' }),
  callClassificationApi: async () => ({ error: 'Simulated timeout' }),
};

const result = await runAssertion({
  assertion: { type: 'not-classifier', value: 'harmful', threshold: 0.5 },
  test: { vars: {}, options: { provider: brokenProvider } },
  providerResponse: { output: 'some model output' },
});

console.log(result);
// { pass: true, score: 1, reason: 'Simulated timeout', ... }
```

**Expected behavior**

A grader/transport error should propagate as `pass: false` with the error tagged (`metadata.graderError`), not get flipped into a pass — same as `not-llm-rubric`, `not-g-eval`, `not-moderation`, etc. already do.

**System information**

- Promptfoo version: 0.122.2
- Commit: `0b4ccbe`
- Install method: local clone

**Additional context**

This is the kind of thing that matters most in redteam/safety configs, where a `not-classifier` check is often the actual safety gate — an outage in the classification provider currently reports green on exactly the check that was supposed to catch a problem.

Fix + tests: add the same `isGraderFailure()` check the other five handlers use, and tag the two matchers' error branches with `graderFail()` / `metadata.graderError`. See the accompanying PR.

he-yufeng:
## Summary

Fixes #10870. A classification or search grader outage could turn a `not-classifier` or `not-search-rubric` assertion into a pass. Tag grader failures and preserve them before applying inversion.

Empty classification maps now count as grader failures with or without a requested label. Search graders must return JSON containing a boolean `pass`; JSON embedded in text still works, while malformed responses and missing verdicts remain failed with score 0. The old substring fallback is removed. Valid negative verdicts retain ordinary inversion behavior. Updated the assertion docs and index.

## Validation

- Merge gate: required `CodeQL` has no run for this fork PR. The repository uses default setup, which [excludes fork pull requests](https://docs.github.com/en/code-security/concepts/code-scanning/setup-types). A normal workflow dispatch also rejected the PR ref.
- [Hosted CI](https://github.com/promptfoo/promptfoo/actions/runs/34748189056) on `41b1e2274fa5d56476bfc427227e50a92d9bc8e8`: all 33 reported checks completed (31 passed, 2 skipped), including the Linux, macOS, and Windows test matrix. Codex code and security reviews completed without new findings.
- Seven regression failures reproduced before the remaining fixes; 122 focused and aggregation tests passed afterward.
- TypeScript, core build plus postbuild, and changed-file lint/format checks passed.
- Source and built CLI each exercised 18 deterministic cases. Exported JSON confirmed all expected outcomes: 14 grader failures retained score 0; valid negative verdicts remained invertible. Each eval intentionally produced 2 passes, 16 failures, and no runtime errors.
- Full suite: 25,599 tests passed; 31 unchanged redteam route tests encountered localhost connection errors. A clean-main run of that suite also failed, with one HTTP 502 rather than the identical socket error. Hosted CI subsequently passed on the published commit.

## Related contributions

Yufeng He's implementation is the primary fix. The coordinated follow-ups preserve Mohammad Safari's full-result regression checks in #10872 and Vishal Chincholi's non-inverted assertion coverage in #10882, retaining each contributor's original PR and authorship.


he-yufeng:
Same defect family as #10488, complementary rather than duplicate: that PR marks grader failures as grader errors on the factuality failure paths, while this one stops `not-classifier` / `not-search-rubric` from inverting a grader/transport failure into `pass: true`. Flagging so a reviewer landing on either sees the family as a whole.

mldangelo-oai (approved):
Reviewed commit 41b1e2274fa5d56476bfc427227e50a92d9bc8e8. The grader-failure fixes, regression coverage, CLI validation, hosted CI, and current-head Codex reviews are satisfactory. Required checks and branch-update requirements must still pass before merging.

PR #10920, opened 2026-09-13
test(providers): isolate registry import failure

mldangelo-oai:
## Summary
The provider registry suite fails with randomized seed `1789336232978`, as seen in the Windows job on #10917. Its injected dynamic-import failure can remain visible to the success-path test despite mock cleanup.

Move the existing failure-path test into its own isolated test file. Preserve its error-message and cause assertions, and remove the assumption that placing it near the bottom makes it run last.

## Validation
- Reproduced on current main with Vitest 5: 1 failure, 188 passing tests.
- Same seed after the change: 189 tests pass across both files.
- 458 focused database/provider/environment tests passed before the test-only move.
- Lint and formatting passed.
- Full backend suite attempted with a temporary Vitest 5/libSQL 0.18 overlay; remaining dependencies are shared with the primary checkout. Stopped after unrelated YAML merge-limit failures with shared js-yaml 5.3.0 (the lockfile requires 5.4.1); clean CI is the full-suite gate.

- All GitHub PR checks passed on commit `051bbcca42622e42addebd8d10a44d512d4c4f48`, including the full Linux/Windows/macOS matrix and built agent-skill examples.


mldangelo-oai:
@codex review

PR #10934, opened 2026-09-14
fix(assertions): reject non-string search rubrics

mldangelo-oai:
## Summary

Reject non-string `search-rubric` and `not-search-rubric` values before calling the grader. Previously, numbers and objects were silently converted into rubrics such as `123` or `[object Object]`. Add coverage for nullish values, numbers, booleans, objects, and arrays, and document the string requirement.

Use YAML comments for the OpenAI tool example's highlight markers so the raw Markdown example parses as YAML as well as the rendered documentation. Docusaurus already strips the original markers from rendered code; highlighting remains intact.

These changes cover three of the five current GitHub AI findings. Existing PRs #8966 (TLS defaults) and #10889 (sanitizer failures) cover the other two and are being refreshed separately.

## Validation

- 28 assertion and search matcher tests passed; new invalid-value tests reproduced the bug before the fix.
- Local CLI with a deterministic grader: valid templated rubric passes; number, object, and array rubrics fail with score 0; only the valid rubric reaches the grader. Exported JSON inspected.
- Production application build, TypeScript, changed-file lint/format, and documentation build passed.
- Raw and rendered tool-example YAML parse identically; rendered highlight markers are absent.
- Shared-base full suite on refreshed sanitizer PR: 25,830 passed, 10 skipped; unrelated redteam-route timeout passed all 45 tests in isolation. Rubric changes were validated separately with the focused tests above.


mldangelo-oai:
@codex review

PR #10937, opened 2026-09-14
fix(integrations): read Langfuse env at fetch time

RachelWanggg:
## Summary

`langfuse://` prompts ignore Langfuse settings supplied through `--env-file` or the config's `env:` block.

`src/integrations/langfuse.ts` built `{ publicKey, secretKey, baseUrl }` from `LANGFUSE_*` when the module was imported. On the CLI that happens through the static import chain `main.ts → commands/eval → node/doEval → evaluator → evaluatorHelpers → integrations/langfuse`. That chain runs before `--env-file` is loaded (`setupEnvFilesFromArgv` in `src/main.ts`) and before `cliState.config` is assigned (`src/util/config/load.ts`), so neither source was ever seen.

What that broke:

- **Config `env:` block:** all three values were dropped. The request went to `cloud.langfuse.com` with no credentials.
- **`LANGFUSE_HOST` in `--env-file`:** dropped. `@langfuse/client` falls back to `process.env` for the keys, but for the URL it reads `LANGFUSE_BASE_URL`, not `LANGFUSE_HOST`. Self-hosted and US-region users were sent to `cloud.langfuse.com` and got a 401.

Shell-exported variables and the cwd `.env` (loaded when `src/envars.ts` is imported) worked before and still work.

This PR reads the settings on every fetch and rebuilds the client only when the public key, secret key, or host changes, so the SDK's prompt cache is kept within a run. Because the values now go through `getEnvString`, `--env-file` and config `env:` take precedence over shell-exported `LANGFUSE_*`, the same as elsewhere in promptfoo.

## Reproduction

A local fake Langfuse server logged the path and basic-auth user of each request. The config used `prompts: [langfuse://greeting@production]`, the `echo` provider, and a `contains` assertion on the compiled prompt.

| Where the Langfuse settings come from | Before | After |
| --- | --- | --- |
| Shell `export` | ✅ pass | ✅ pass |
| `--env-file` (`LANGFUSE_HOST=http://127.0.0.1:3999`) | ❌ 401 `Invalid credentials. Confirm that you've configured the correct host.` (fake server received nothing) | ✅ pass, server received `pk-lf-dotenv` |
| Config `env:` block | ❌ 401 `No authorization header` (fake server received nothing) | ✅ pass, server received `pk-lf-configenv` |

## Testing

- `npx vitest run test/integrations/langfuse.test.ts test/evaluatorHelpers.langfuse.test.ts`: 45 passed.
- New tests: settings read at fetch time, and the client rebuilt when only the host, only the public key, or only the secret key changes. They fail on `main` and pass here. I also checked them against a mutant that compares only `publicKey`: they catch it.
- Shuffled runs with several `--sequence.seed` values.
- `npx biome lint` and `npx prettier --check` on the changed files.
- `npx tsc --noEmit`: clean, apart from the 5 `playwright-extra` errors that already exist on `main`.
- End-to-end with `npm run local -- eval --no-cache` for the three cases in the table.

## Notes

- Open PR #6712 also moves these reads into a function, but it still creates the client once and targets the pre-v5 SDK. This fix is smaller and independent of it.
- Out of scope: reading `LANGFUSE_BASE_URL` through `getEnvString`, and the programmatic `evaluate()` path, which doesn't populate `cliState.config`. `src/integrations/helicone.ts` also reads `HELICONE_API_KEY` at import time. I'm happy to send a follow-up for that.

## Maintainer follow-up

- Merged current `main`.
- **Fix: blank values are treated as unset.** Reading settings when fetching let a blank value such as `LANGFUSE_HOST=` from `--env-file` or `env:` reach the SDK. The SDK only falls back to `LANGFUSE_BASE_URL` and its default host when a setting is `undefined`, so `''` was used as the base URL and every fetch failed with `Failed to parse URL`. That case works on `main`, so it was a regression in this branch.
- **Test:** `test/integrations/langfuse.env.test.ts` uses the real `getEnvString` and `cliState`. `langfuse.test.ts` mocks `getEnvString`, so it can't show that these sources reach the client. The new tests cover `process.env` values set after import and config `env:` overriding shell exports, and they fail on `main`.
- **Docs:** the Langfuse page now mentions `--env-file` and `env:`, and that both override shell exports.

End-to-end with a fake Langfuse server, the `echo` provider, and `--no-cache`:

| Langfuse settings | `main` | `5c37847` | now |
| --- | --- | --- | --- |
| Shell `export` | ✅ | ✅ | ✅ |
| `--env-file` | ❌ | ✅ | ✅ |
| Config `env:` block | ❌ | ✅ | ✅ |
| `--env-file` with blank `LANGFUSE_HOST=`, shell `LANGFUSE_BASE_URL` | ✅ | ❌ `Failed to parse URL` | ✅ |
| Config `env:` over a stale shell `LANGFUSE_HOST` | ❌ | ✅ | ✅ |


mldangelo:
Thank you!

PR #10946, opened 2026-09-15
fix(providers): preserve streamed Ollama tool calls

HaokaiDing:
Ollama chat can emit tool calls in multiple streaming chunks. The provider used `find` to select the first chunk with `tool_calls`, so subsequent calls were silently dropped from the evaluated output.

Collect the calls from every chunk in response order, then use the existing argument normalization and output handling. This follows the accumulation pattern in the [Ollama streaming tool-calling documentation](https://docs.ollama.com/capabilities/tool-calling#tool-calling-with-streaming).

For example, an NDJSON response with `get_weather(Amsterdam)` in one chunk and `get_weather(Paris)` in the next previously returned only the Amsterdam call. It now returns both.

Validation:
- Both new regressions fail before the fix, covering streams with and without text content.
- `npm test -- test/providers/ollama.test.ts --maxWorkers=1`: **40 passed**.
- TypeScript checking, the core `tsdown` build, postbuild, and changed-file lint/format checks pass.
- A local HTTP fixture exercised the actual CLI with `--no-cache` and JSON export. Before: one call returned, assertion failed, exit 100. After, using the built CLI: both calls returned, assertion passed, exit 0. Token usage remained 10 prompt / 20 completion / 30 total.
- Full backend run: 26,240 passed, 26 failed, 10 skipped. The failures were confined to three files and caused by the local runner settings: an update-disable environment variable and a Node executable under an `_npx` path. With those settings corrected, all **115 tests in the three affected files passed**.
- Node 24.21.0 and npm 11 were used. No live model or paid API was called.


mldangelo (approved):
Approving — this is a well-targeted fix, and it holds up under live verification.

I QA'd it against a real Ollama 0.13.5 server (`qwen3:0.6b`, `qwen2.5:0.5b`, `functiongemma`) rather than just reading the diff, and the premise reproduces exactly:

```
{"message":{...,"tool_calls":[{"id":"call_ourx7m7g","function":{"index":0,"name":"get_weather","arguments":{"city":"Amsterdam"}}}]},"done":false}
{"message":{...,"tool_calls":[{"id":"call_lyc8c6rr","function":{"index":1,"name":"get_weather","arguments":{"city":"Paris"}}}]},"done":false}
{"message":{...,"tool_calls":[{"id":"call_h6tdm2e6","function":{"index":2,"name":"get_weather","arguments":{"city":"Tokyo"}}}]},"done":false}
```

Replaying that captured stream through both implementations:

```
origin/main find()  -> 1 call:  Amsterdam
PR #10946 flatMap() -> 3 calls: Amsterdam, Paris, Tokyo
```

**`flatMap` is the right shape here, and I want to record why**, since the safety of the fix hinges on a detail that isn't obvious from the diff. Ollama emits tool calls with *delta* semantics — exactly one **new** call per chunk, never a cumulative array that repeats prior chunks, and the `done: true` chunk never carries `tool_calls`. That's what makes concatenation safe rather than double-counting. I checked this across ~16 runs spanning 2–6 forced calls at temperature 0 and 0.8: zero repeated ids, zero cumulative arrays. Ollama also never splits a single call's `arguments` across chunks — when the token budget runs out mid-call it emits nothing rather than a fragment — so there's no partial-accumulation hazard either.

Worth noting for anyone tempted to "improve" this later: `function.index` is **not** a safe de-dup key. `functiongemma` reports `index: 0` for both of two distinct calls, so an index-keyed merge would silently drop one. Plain concatenation is correct.

**Why this survived so long.** `OllamaChatProvider` never sets `stream: false`, so real usage always takes the NDJSON path — but every pre-existing fixture in the test file mocks the single-line `done: true` shape, and on that shape `find()` is accidentally correct. I confirmed the non-streaming path returns all calls on one line. So the tests were green against a shape production never produces. The new `it.each` cases close exactly that gap, and they're good ones: order-sensitive and dedupe-resistant (reversing or deduping the expected array breaks `toEqual`).

Verification on my end:

- `npx vitest run test/providers/ollama.test.ts` → 40/40 passed; the two new cases genuinely fail against pre-fix code (2 failed / 38 passed)
- `npx tsc --noEmit` → exit 0; `biome check` on both files → clean
- End-to-end through the real CLI with `--no-cache`: multi-call output preserved, `is-valid-openai-tools-call` passes
- Streaming and `passthrough: {stream: false}` now produce identical output; the interleaved `{content, tool_calls}` shape works; cache round-trip preserves all calls

Two things I looked at and deliberately am **not** asking you to change:

1. The `tool_calls &&` operands at lines 433/447 are now statically always-true, since `flatMap` can't return `undefined`. But those are context lines the diff doesn't touch, `x && x.length > 0` appears ~29 times across `src/` in this exact role, and there's no type-aware lint rule for it. Not worth widening a clean two-file diff.
2. The new fixtures use the already-normalized `{function: {name, arguments}}` shape rather than the real wire envelope (`id`, `function.index`). That matches the file's existing convention, and your change is field-agnostic, so nothing is hidden by it.

Separately, this audit turned up several **pre-existing** Ollama provider issues that are out of scope here and that I'm splitting into their own PRs — most notably that non-2xx responses are currently swallowed into empty-string passes, and that the embedding path still uses the superseded `/api/embeddings`. None of that touches this diff; `git diff origin/main...HEAD` confirms this PR changes nothing in the error, options, `think`, or embedding paths.

Nice catch, and thanks for including a regression test that actually fails without the fix.

PR #10953, opened 2026-09-16
fix(providers): surface Ollama HTTP errors instead of empty output

mldangelo:
## Summary

Ollama HTTP errors were being swallowed into empty-string **passes**. An eval against a model that doesn't exist reported `2 passed (100%), 0 errors` and exited 0.

Both providers guarded errors with:

```ts
if (typeof response.data === 'object' && response.data !== null && 'error' in response.data)
```

…but they call `fetchWithCache` with format `'text'`, which always resolves `data` to a **string** (`src/cache.ts` does `format === 'json' ? JSON.parse(respText) : respText` on every path). So the guard was unreachable. `fetchWithCache` also doesn't throw on non-2xx — it returns the body with `cacheable: false` — and `response.status` was fetched only to be logged and discarded.

The net effect: Ollama's error body is a single valid JSON line, so it survives the NDJSON parse, has no `message.content`, and yields `output: ''` with no `error`.

## Impact

This is not just a typo'd model name. All of these produced silent passes:

| Trigger | HTTP |
|---|---|
| Model not installed | 404 |
| `num_predict: '32'` quoted in YAML | 500 |
| Chat provider pointed at an embedding model | 400 |
| Bad `OLLAMA_API_KEY` (Ollama Cloud) | 401 |

Downstream consequences:

- **With no assertions → silent PASS.** Empty output vacuously satisfies `not-contains` / `not-icontains`.
- **Red team scans report false green.** `src/redteam/plugins/base.ts` short-circuits empty output to `{ pass: true, reason: 'Model refused the request' }` *before any grading runs*. A scan against an unreachable target produced a clean safety report attributing it to model alignment.
- **With positive assertions → misleading FAIL**, attributed to model quality (`Expected output to contain "..."`) with `failureReason: ASSERT`, not ERROR.

## Change

Adds a shared `getOllamaResponseError` helper that checks HTTP status and extracts the error message from either a string or an object body, wired into the completion, chat, and embedding providers.

Before / after against a live server:

```
# before
ollama:chat:does-not-exist       | success=True | error=None | output=''
ollama:completion:does-not-exist | success=True | error=None | output=''
=> 2 passed (100%), 0 errors, exit 0

# after
Ollama API error: 404 Not Found: model 'does-not-exist' not found
Ollama API error: 500 Internal Server Error: option "num_predict" must be of type integer
Ollama API error: 400 Bad Request: "all-minilm" does not support chat
=> 0 passed, 2 errors (100%)
```

Ollama already returns an actionable diagnostic; this just stops discarding it.

The embedding provider uses format `'json'`, so it did surface *an* error — but a confusing one (`No embedding found in Ollama embeddings API response: {"error":...}`). It now reports the same clear status-bearing message.

## Tests

The two existing "error field" tests mocked `data: { error: ... }` with `status: 200` — a shape `fetchWithCache(..., 'text')` can never produce. They were the only coverage of that branch, they passed, and they certified behavior production cannot exhibit. `grep -n "status: 4\|status: 5"` on the file returned nothing; `test/AGENTS.md` requires 4xx/5xx coverage for providers.

Those fixtures now use string bodies, and non-2xx coverage is added across all three providers (404/500/401 for completion, 404/400/401 for chat, 404/400 for embeddings), plus non-JSON and empty error bodies, plus a regression guard that a successful multi-line NDJSON stream is **not** mistaken for an error body.

- `npx vitest run test/providers/ollama.test.ts` → **51 passed** (was 40)
- All 12 new/updated cases **fail** against the previous implementation (verified by swapping in `origin/main`'s provider: 12 failed / 39 passed)
- `npx tsc --noEmit` → exit 0; `npm run l && npm run f` clean; `npm run knip` clean
- `test/providers/registry.test.ts` (218), `test/providers/index.test.ts`, `test/tracing/providerInstrumentation.test.ts`, `test/matchers/llm-rubric.test.ts` → all pass

## Live verification

Run against a real Ollama 0.13.5 server with `qwen3:0.6b` and `all-minilm` (no paid API calls). Each error class above was reproduced and confirmed fixed, and the happy path is unchanged:

```
success=True | output='The capital of France is Paris.'
    similar   pass=True score=0.982
    icontains pass=True score=1
```

## Scope

This came out of a broader audit of the Ollama provider done while reviewing #10946. Deliberately **not** included here, each queued as its own PR:

- `message.thinking` is discarded entirely (no `showThinking`, unlike 14 other providers)
- embeddings still use the superseded `/api/embeddings`, which hard-500s past ~200 words on `all-minilm`
- `done_reason` never surfaced, so the `finish-reason` assertion can never pass
- the `cached` flag is never set
- option allowlist is still pinned to Ollama v0.1.0

Keeping this one standalone because it changes error semantics for every Ollama user and warrants its own release note.


mldangelo:
Good catches — all four were real, and all four are fixed in `4e3d082`. Each has a regression test that fails against the previous commit.

**1. 3xx not treated as a failure** (@copilot, and Codex's "Treat every non-2xx status as an API error")

Correct. `fetchWithCache` returns every `!response.ok` body, and I only checked `status >= 400`. Now matches `fetchWithCache`'s own check:

```ts
function isOllamaHttpFailure(status: number): boolean {
  return status < 200 || status >= 300;
}
```

Test: a `300 Multiple Choices` with an empty body now errors instead of returning `''`.

**2. Error records emitted partway through a 200 stream** (Codex P1)

This was the best of the four and I'd missed it. My extractor only parsed the body when the whole trimmed string parsed as a single JSON value, which a multi-line NDJSON body never does — so a stream that started fine and then failed returned the partial content as a success. Detection now scans record by record, with a `"error"` substring pre-filter so long successful streams aren't re-parsed line by line.

I could not reproduce this against a live Ollama 0.13.5 (tried negative `num_ctx` and a 20k-word prompt against `num_ctx: 512`; both completed normally), so the regression test uses a fixture. Fixing it regardless — it's cheap and the failure mode is silent.

**3. Non-2xx JSON bodies without a top-level `error` key** (Codex P2)

Right, and this was a regression against the previous embedding behavior, which preserved the body through `JSON.stringify` in its catch path. Non-string bodies are now serialized as the fallback detail:

```
Ollama API error: 401 Unauthorized: {"message":"invalid token"}
```

**4. HTTP 200 error bodies staying cached** (Codex P2)

Also correct, and specific to the `'text'` format — `fetchWithCache` only inspects `'json'` bodies for an error key (`src/cache.ts:777`), so a 200-with-error body was cached for the full TTL and replayed after the server recovered. Detected errors now call `response.deleteFromCache?.()` before returning, in all three providers.

---

One clarification on Copilot's "unnecessary full-stream JSON parse": the extractor no longer parses the full body as one value at all, and the pre-filter means a successful stream costs one `includes()` per line rather than a parse.

Re-verified live against Ollama 0.13.5 after the fix:

```
bad-option-500     Ollama API error: 500 Internal Server Error: option "num_predict" must be of type integer
chat-vs-embed-400  Ollama API error: 400 Bad Request: "all-minilm" does not support chat
happy-chat         success=True
happy-completion   success=True
```

Tests now 55 (was 51).

PR #10954, opened 2026-09-16
feat(providers): surface Ollama thinking output and finish reason

mldangelo:
> **Stacked on #10953** — base is `fix/ollama-surface-http-errors`. Review that one first; I'll retarget this to `main` once it lands. The diff shown here is only this change.

## Summary

The Ollama provider discarded two pieces of every response: the reasoning trace (`message.thinking` / `thinking`) and the stop reason (`done_reason`). Both providers returned literally `{ output, ...tokenUsage }`.

### 1. Reasoning was dropped — and could leave the output completely empty

`showThinking` appears in 14 other provider files; Ollama was the only reasoning-capable provider with no way to see the trace.

The sharper problem is that **Ollama 0.13+ emits `thinking` for `qwen3` and friends by default, with no `think` flag sent**. So this hits users who never opted into reasoning. When a `num_predict` budget is spent inside the thinking block, the model never emits any content:

```
ollama:completion:qwen3:0.6b, num_predict 32  ->  output ''

# raw /api/generate, same request
response     : ''
thinking     : 'Okay, the user wants me to say "hi" in one word. Let me think...'
done_reason  : 'length'
```

The eval reported a pass with a blank output and no indication why.

### 2. `finish-reason` could never pass against Ollama

Every Ollama response carries `done_reason` on both endpoints. Before:

```
gradingReason: ['Provider did not supply stop/finish reason']
```

## Change

- Accumulate `thinking` across chunks and prepend it as `Thinking: ...`, matching `src/providers/openai/chat.ts:801` and `src/providers/anthropic/util.ts:525`.
- Add `showThinking` (default `true`, same as elsewhere) to opt out. It's deliberately **absent from `OllamaCompletionOptionKeys`**, so it is never forwarded to Ollama — there's a test asserting it appears in neither the body nor `options`.
- Set `finishReason` from `done_reason` through `normalizeFinishReason`. Ollama's `stop`/`length` already match the normalized vocabulary.
- Tool-call outputs are left structurally untouched; only string outputs get the prefix, so `is-valid-openai-tools-call` is unaffected.
- Extract `collectOllamaToolCalls`, `extractOllamaTokenUsage`, and `applyOllamaThinking` so both providers share them. This also keeps `OllamaChatProvider.callApiInternal` under Biome's cognitive-complexity limit, which the added branch would otherwise have exceeded (33 > 30).

Live, after (`-j 1`):

```
chat-no-thinking       (think: false)         fr=stop    '"Hello!"'
chat-truncated-thinking                       fr=length  'Thinking: Okay, the user wants me to say "hi"...'
chat-hidden-thinking   (showThinking: false)  fr=length  ''
completion-truncated-thinking                 fr=length  'Thinking: Okay, the user wants me to say "hi"...'
```

## Behavior change worth calling out

For anyone already running a reasoning model through Ollama, output now includes a `Thinking: ...` prefix where it previously did not. That's the established convention for every other provider here, and it converts a silently-empty result into a useful one — but it can affect existing assertions. `showThinking: false` restores the previous rendering, and `think: false` turns reasoning off at the model level. Both are documented.

Happy to flip the default to `false` for Ollama specifically if you'd rather not change output for existing users; I went with `true` for consistency with OpenAI/Anthropic and because it's what fixes the empty-output case.

## Tests

- `npx vitest run test/providers/ollama.test.ts` → **59 passed** (55 → 59 for this change)
- The 5 new cases fail against the parent commit (verified: 5 failed / 54 passed)
- Covers: multi-chunk thinking accumulation, thinking + content, `showThinking: false`, `showThinking` not reaching the API, tool-call shape preserved alongside thinking, and `finishReason` omitted when `done_reason` is absent
- `npx tsc --noEmit` exit 0; `npm run l && npm run f` clean; Biome complexity warning resolved
- `registry.test.ts`, `index.test.ts`, `providerInstrumentation.test.ts` → 504 tests pass

## Live verification

Against a real Ollama 0.13.5 server with `qwen3:0.6b` (no paid API calls):

```
finish-reason assertion    -> finishReason 'stop', pass=True   (previously impossible)
multi tool calls           -> n_tool_calls=2, is-valid-openai-tools-call pass=True
similar / embeddings path  -> unchanged
```

One incidental observation, not related to this change: running several concurrent requests against a single small model on one Ollama daemon produces occasional cross-contaminated output (fragments of one response appearing in another). It moves between rows across runs and reproduces on `think: false` rows where this code path is inert, so it's a daemon-side artifact. `-j 1` is clean. Flagging it since the docs already recommend `-j 1` for local setups.

## Docs

Adds a "Reasoning models" section covering `showThinking`, the default-on `thinking` behavior, the `num_predict` truncation warning, and `finishReason`.


mldangelo:
Good catch on the tracing gap — fixed in `9f000d1`.

Both `resultExtractor` implementations copied only `tokenUsage`, so `withGenAISpan` never received `finishReasons` and the span omitted `gen_ai.response.finish_reasons` even though `ProviderResponse` now carried the value:

```ts
if (response.finishReason) {
  result.finishReasons = [response.finishReason];
}
```

The regression test lives in `test/tracing/providerInstrumentation.test.ts` rather than the provider suite, so it exercises a real `NodeTracerProvider` + `InMemorySpanExporter` end to end instead of a hand-rolled span mock. My first attempt did mock the span and produced a misleading failure, which was a good reminder that the existing harness is there for a reason. Verified it fails without the extractor change.

I also merged the updated #10953 in, so this branch carries the four error-detection hardening fixes from that review too.

PR #10955, opened 2026-09-16
fix(assertions): reject malformed webhook results

RaycarlLei:
An HTTP 200 webhook response such as `{}` or `{"pass":"false"}` currently passes both `webhook` and `not-webhook`: comparing a missing/non-boolean `pass` with the boolean inverse flag always returns true.

Validate the documented boolean `pass` field before applying inversion. Invalid responses use the existing webhook error path, so they cannot become successes under `not-webhook`. Valid booleans, custom scores, and reasons retain their behavior. The assertion docs and reference now make this requirement explicit.

The added regression cases exercise the public assertion dispatcher with missing fields, an error object, string/number/object/array values, and null/primitive JSON bodies in both assertion directions. On the original code, 24 cases incorrectly pass; two null cases already fail but return an opaque property-access error. Four valid-response controls verify score/reason preservation.

Verification (Node 24.21.0):

```sh
npm ci
npm test -- test/assertions/runAssertion.test.ts test/assertions/runAssertions.test.ts test/assertions/assertionsResult.test.ts test/evaluator/assertions.test.ts --maxWorkers=4
npm run tsc
npm run l
npm run f
```

All 292 related tests and the checks above pass locally. A real CLI eval using the `echo` provider and a loopback HTTP webhook also reproduced the bug: four malformed-response cases pass before the patch and fail afterward, while two valid controls continue to pass. The exported JSON was checked for success, score, and failure reason. No model credentials or external model calls are needed.

An independent [Linux run](https://github.com/RaycarlLei/promptfoo/actions/runs/35099265252) checks the exact contribution commit `098b12f058a1ca185191869bb6b2d16e2211158c`: the same 292 tests, full TypeScript check, and contribution formatting all pass.


mldangelo-oai:
Thanks!

PR #10958, opened 2026-09-16
fix(providers): migrate Ollama embeddings to /api/embed and honor config

mldangelo:
> **Stacked on #10954** (which is stacked on #10953). I'll retarget as those land.

## Summary

`OllamaEmbeddingProvider.callEmbeddingApi` posted to `/api/embeddings`, which upstream explicitly marks as superseded:

> Note: this endpoint has been superseded by `/api/embed` — `ollama/ollama` `docs/api.md`

That's not cosmetic. The old endpoint hard-fails once input exceeds the model's context window, and the provider wrapped the failure in a misleading message:

```
API response error: Error: No embedding found in Ollama embeddings API response:
{"error":"the input length exceeds the context length"}
```

On `all-minilm` — the model this repo's own docs recommend as "lightweight, fast embeddings" — that's roughly **200 words**. A `similar` assertion against a medium-length output just breaks.

**And there was no way for a user to fix it.** The old implementation was literally:

```ts
const params = { model: this.modelName, prompt: text };
```

`this.config` was never consulted, so `num_ctx` — the actual remedy — was unreachable, despite the docs advertising a `config:` block for embedding providers.

## The truncation decision

This is the part worth reviewing carefully, because I deliberately did **not** take Ollama's default.

`/api/embed` defaults to `truncate: true`, which "succeeds" by embedding only the first `num_ctx` tokens. I verified `prompt_eval_count` pins at 256 regardless of how much text you send. So a naive URL swap would have traded a loud, accurate failure for a **silently wrong similarity score** — the worse outcome for an eval tool, and also a behavior change from the endpoint it replaces.

So promptfoo sends `truncate: false` and surfaces an error that names both remedies and the tradeoff:

```
Ollama API error: 400 Bad Request: the input length exceeds the context length.
Raise `config.num_ctx` (up to the model's own maximum, shown by `ollama show all-minilm`),
or set `config.truncate: true` to embed only the first num_ctx tokens -- note that
truncating silently changes similarity scores.
```

Happy to flip the default if you disagree — but I'd argue an eval tool should never quietly score a truncated embedding.

## Change

- `/api/embed`, reading `embeddings[0]`
- Threads config through: `options` (so `num_ctx` works), `truncate`, `dimensions`, `keep_alive`, `passthrough`
- Populates `tokenUsage` from `prompt_eval_count`, which the new endpoint returns and the old one didn't
- Extracts a shared `buildOllamaOptions` used by all three providers

The shared builder also fixes a pre-existing asymmetry it exposed: the chat reducer excluded only `tools`, so `think` and the entire `passthrough` object were *additionally* copied into the nested `options` as junk members. Inert (Ollama type-validates only known keys — measured), but wrong. It also let me delete a duplicate `params.think` write in the completion provider that could clobber `passthrough.think`.

## Live verification

Against a real Ollama 0.13.5 server, the key proof that config now actually reaches the provider — same ~350-token `similar` assertion, only the config differs:

```
default (256 window)  ->  pass=False  Ollama API error: 400 ... exceeds the context length
config.num_ctx: 512   ->  pass=True   Similarity 0.63 >= threshold 0.2
```

And the other remedy:

```
config.truncate: true ->  pass=True   Similarity 0.74 >= threshold 0.3
```

Short-input happy path unchanged (`similar` 0.982). Note `num_ctx` is bounded by the model's architecture — `all-minilm` reports `bert.context_length = 512` via `/api/show` — which is why the error message points at `ollama show`.

## Tests

- `npx vitest run test/providers/ollama.test.ts` → **67 passed** (from 63)
- Covers: endpoint URL is `/api/embed`, `truncate: false` default, config threading (`num_ctx`/`truncate`/`dimensions`/`keep_alive`), token usage, the context-length guidance message, and that `think`/`passthrough` no longer leak into `options`
- `npx tsc --noEmit` exit 0; `npm run l && npm run f` clean; `architecture:check` passes

## Docs

Adds an "Embedding input length" section explaining the window, the error, and both remedies; updates the endpoint reference and documents the `ollama:embedding:` singular alias.


mldangelo:
Good catch — correct, and fixed in `4219485`.

The new `tokenUsage` block reported `prompt_eval_count` as ordinary prompt/total usage even on a cache hit and never set `cached`, so repeated `similar` assertions counted cached embeddings as fresh usage. It now branches on `response.cached`:

```ts
const tokenUsage =
  promptTokens === undefined
    ? undefined
    : response.cached
      ? { cached: promptTokens, total: promptTokens }
      : { prompt: promptTokens, total: promptTokens };
return {
  embedding,
  ...(tokenUsage && { tokenUsage }),
  ...(response.cached && { cached: true }),
};
```

matching `getTokenUsage` in `src/providers/openai/util.ts` and the requirement in `src/providers/AGENTS.md`. Regression test added (`cached: 7, total: 7`, `result.cached === true`).

Worth noting I had already written this fix in the stacked #10961 ("thread cached flag"), but you're right that it belongs **here** — this PR is what introduces the `tokenUsage` block, so shipping it without cache handling would leave a window where the bug exists on `main` if these land separately. Moved it down into this PR and merged up the stack; #10961 still carries the equivalent handling for the chat and completion providers.

mldangelo:
Also fixed the duplicate Copilot flagged (`748da11`). The merge from `main` duplicated the parameterized embeddings error-body block, not the one I'd already removed — Biome's duplicate-test rule only matches literal `it()` titles, so an `it.each` with a template title passed lint cleanly. Removed; 68 tests pass.

mldangelo:
@codex review

mldangelo:
Both fixed in `0fdcfae`. The `numRequests` one corrected a mistake in my earlier reasoning.

**Embedding usage omitted `numRequests`.** I'd argued in #10961 that setting it was a no-op because `tokenUsageUtils.ts:176` increments when absent. That's true — but only when `incrementRequests` is passed, and **it defaults to `false`**. `src/matchers/similarity.ts:156-157` calls `accumulateTokenUsage(mergedUsage, ...)` with two arguments, so the embedding path never inferred a count.

Measured on a live `similar` assertion, which makes two embedding calls (expected + output):

```
before:  assertions.numRequests = 1
after:   assertions.numRequests = 2   <- correct
```

So my "no-op" justification held for the chat/completion path (which goes through `accumulateResponseTokenUsage` and does report 1 per call — verified) but not for embeddings. `numRequests: 1` is now set explicitly on the uncached path, matching `src/providers/voyage.ts:119` and `src/providers/cohere.ts:211`. Cache hits still omit it, per `getTokenUsage` in `src/providers/openai/util.ts` — a cache hit isn't a request.

**Singular alias missing the model segment.** Correct — `ollama:embedding:` copied verbatim yields an empty model name. Now documented as `ollama:embedding:<model name>`, matching the plural form.

PR #10959, opened 2026-09-16
fix(redteam): align redteam-ollama example with its README

mldangelo:
## Summary

`examples/redteam-ollama/README.md` tells you to `ollama pull llama3.2` and documents the target as "Llama 3.2". The config targeted `ollama:chat:llama4` — a ~67 GB MoE the instructions never mention. Anyone following the README verbatim ran the scan against a model they hadn't pulled.

## Why this matters more than a typo

It **fails green**. An uninstalled model returns HTTP 404, and `src/redteam/plugins/base.ts` short-circuits `isEmptyResponse(llmOutput)` to `{ pass: true, score: 1, reason: 'Model refused the request' }` *before any grading runs*.

So the scan produces a clean safety report, attributing the "refusals" to model alignment, for a model that was never contacted. A red team example that reports a false all-clear is worse than one that errors.

(The underlying swallow — non-2xx becoming an empty-string pass — is fixed separately in #10953. This PR fixes the example so it points at the model its own README installs.)

## Also: `max_tokens` is a no-op

`max_tokens: 150` is an OpenAI key. It was inert twice over — absent from `OllamaCompletionOptionKeys`, so the provider stripped it, *and* ignored by Ollama even when sent directly. Measured against Ollama 0.13.5 on `qwen3:0.6b`:

```
options.max_tokens: 16   ->  eval_count 673
options.num_predict: 16  ->  eval_count 16
```

Replaced with `num_predict: 150`. `site/blog/red-team-ollama-model.md` shipped the same mistake and is fixed too (its target model was already correct).

## Scope

`(redteam)` per THE REDTEAM RULE in `docs/agents/pr-conventions.md`, which explicitly covers redteam examples and docs even when the change is config-only.

## Verification

Config parses against the schema; `npm run f` clean. I did not run a full redteam generation, since that calls remote services.


mldangelo:
I want to push back on this one, but flag it clearly so a maintainer can overrule me.

`site/AGENTS.md:9` does say "Always use `docs(site):` scope for all site changes." But `docs/agents/pr-conventions.md` carves out an explicit exception, and the blog post here is redteam content:

> ## THE REDTEAM RULE
> **If a PR is primarily redteam-related, use `(redteam)` scope.**
> This applies even if the redteam change is only in UI, CLI, **docs**, examples, utilities, tests, or server endpoints.

The root `AGENTS.md` reinforces it: "If a pull request is primarily about redteam functionality, verify the title follows THE REDTEAM RULE ... and uses `(redteam)` scope."

`site/blog/red-team-ollama-model.md` is a red teaming walkthrough, and the one-line change in it is the *same* `max_tokens` → `num_predict` bug as the example config — same root cause, same fix. So I read this as squarely inside the rule's "even if ... only in docs" clause.

There's also a practical point: this repo squash-merges, so splitting into two commits wouldn't change the final history at all — the squash commit takes the PR title either way. Making the scope split real would mean a separate PR for a one-line blog edit, which I think costs a reviewer more than it gains.

Happy to split it into its own `docs(site):` PR if a maintainer disagrees — just say the word and I'll do it. Leaving as-is for now since the two rules conflict and the more specific one (THE REDTEAM RULE) explicitly names docs.

PR #10961, opened 2026-09-16
chore(providers): refresh Ollama option allowlist and thread cached flag

mldangelo:
> **Stacked on #10958.** I'll retarget once that lands.
>
> **Scope changed during review.** This PR originally *removed* 14 options that Ollama has dropped upstream. Review pointed out that silently changes behavior for anyone pointing `OLLAMA_BASE_URL` at an older release, so they are now **forwarded with a deprecation notice** instead. The description below reflects the current implementation.

## Summary

Three related pieces of provider hygiene.

### 1. The option allowlist was pinned to Ollama v0.1.0

The interface carried a source comment pointing at `api/types.go` **from v0.1.0** and had never been revisited. Checked against Ollama's current `Options`/`Runner` structs.

**Added:**

- `min_p` — Ollama honors it; promptfoo was silently dropping it
- `draft_num_predict` — `Runner.DraftNumPredict`, used for speculative decoding (verified the live server accepts it, HTTP 200)
- `keep_alive` as a proper top-level parameter for chat/completion (previously reachable only via `passthrough`)

**Kept, but now flagged:** 14 options Ollama removed from its current struct (`mirostat`, `mirostat_tau`, `mirostat_eta`, `tfs_z`, `num_gqa`, `f16_kv`, `logits_all`, `vocab_only`, `low_vram`, `use_mlock`, `embedding_only`, `rope_frequency_base`, `rope_frequency_scale`, `penalize_newline`).

They are still forwarded. Modern servers ignore unknown option keys — verified by sending all 14 at once and getting HTTP 200 — so this costs nothing there, while an older `OLLAMA_BASE_URL` keeps its behavior. promptfoo doesn't enforce a minimum Ollama version, so dropping them would have been a silent regression for those users. A debug notice distinguishes the two cases:

```
[Ollama] Forwarding options that current Ollama releases ignore  -> mirostat
[Ollama] Ignoring unsupported config keys                        -> useNUMA, max_tokens
```

**Removed:** `useNUMA` only. Unlike the others, it was never a valid wire name — the Go field was ``UseNUMA `json:"numa"` `` and Ollama keys off the JSON tag, so the *field* name had been transcribed. `numa` is gone upstream too, so no server version, old or new, would honor it.

Unrecognized keys were always discarded silently; the only change is that a debug line now names them. That silent drop is how `max_tokens` — an OpenAI key Ollama ignores — sat unnoticed in this repo's own redteam example (fixed in #10959).

### 2. `passthrough.options` clobbered the computed options

`passthrough` was spread wholesale *after* `options:`, so anyone reaching for an unsupported option via the documented escape hatch silently lost `temperature` and `num_predict`. It now merges. The dropped-key hint was also wrong — it pointed at `passthrough`, but generation options must be nested under `passthrough.options` or Ollama ignores them.

### 3. The `cached` flag was never set

`src/providers/AGENTS.md` requires it in bold. Without it the evaluator never takes its "Skipping delay because response is cached" branch (measured: a 5 ms cache hit still slept a full `delay: 5000`), and `tokenUsage.cached` stayed 0. Cache hits now report `{ cached: total, total }`, matching `getTokenUsage` in `src/providers/openai/util.ts`.

`numRequests` is intentionally left unset on the uncached chat/completion path: those flow through `accumulateResponseTokenUsage`, which already reports 1 request per call (verified live). The embedding path is different and *does* set it explicitly — see #10958.

## Live verification

Against Ollama 0.13.5:

```json
{"model":"qwen3:0.6b","messages":[...],
 "options":{"temperature":0,"num_predict":8,"min_p":0.05,"mirostat":2},
 "think":false,"keep_alive":"1m"}
```

`min_p` forwarded, `mirostat` forwarded with a deprecation notice, `useNUMA` and `max_tokens` dropped, `keep_alive` top-level and not duplicated into `options`.

## Tests

`npx vitest run test/providers/ollama.test.ts` → **72 passed**. Coverage: cached flag and cached token usage for chat *and* completion, `passthrough.options` merging rather than clobbering, `min_p`/`draft_num_predict`/`keep_alive` forwarding, legacy options still forwarded, and `useNUMA`/`max_tokens` dropped.

`npx tsc --noEmit` exit 0; `npm run l && npm run f` clean; `architecture:check` passes.


mldangelo:
Four findings, all addressed in `d0154df`. The first one changed my mind about the design.

**1. "Preserve options for older Ollama servers" (P1) — you're right, and I've reversed the removal.**

I had verified these 14 keys are gone from Ollama's *current* `Options` struct and concluded they were dead weight. But promptfoo doesn't enforce a minimum Ollama version, and `OLLAMA_BASE_URL` can point anywhere — so for someone on an older release, dropping `mirostat` or `use_mlock` silently changes sampling behavior. That's a real regression traded for cosmetic tidiness, which is a bad trade.

They're forwarded again. Modern servers ignore unknown option keys (verified: sending all 14 at once returns HTTP 200), so this costs nothing there while an older server keeps working. A debug notice now distinguishes the two cases instead:

```
[Ollama] Forwarding options that current Ollama releases ignore  -> mirostat
[Ollama] Ignoring unsupported config keys                        -> useNUMA, max_tokens
```

**`useNUMA` stays dropped**, and I want to be explicit about why it's not the same case: it was never a valid wire name. The Go field was ``UseNUMA `json:"numa"` `` and Ollama keys off the JSON tag, so the *field* name had been transcribed. `numa` itself is also gone upstream, so there is no server version, old or new, that would honor `useNUMA`. Nothing to preserve.

**2. "Point unsupported options to `passthrough.options`" (P2) — correct, the hint was actively misleading.**

Generation options must be nested under `options`, so following my hint as `passthrough: { new_option: ... }` would have produced a silently ignored request. Now:

> Generation options belong under `passthrough.options`; other top-level API fields go directly under `passthrough`.

**3 & 4. Cache-hit coverage for the completion and embedding paths (P1 ×2) — added.**

Fair: all three providers build their responses independently, so one chat test left two paths unguarded. There are now cache-hit tests for chat, completion, and embeddings (the embeddings one lives in #10958, where that branch is introduced).

Live re-verification after the change:

```
options: {"temperature":0,"num_predict":8,"min_p":0.05,"mirostat":2}
```

`min_p` forwarded, `mirostat` forwarded with a deprecation notice, `useNUMA` and `max_tokens` dropped. 72 tests pass.

mldangelo:
@codex review

mldangelo:
Both fixed in `8014497`, and the description rewritten.

**Stale PR description — fair hit, and entirely my fault.** The description still described the *original* scope (remove 15 keys) after I reversed course mid-review and kept 14 of them forwarded. A reviewer reading the body would have been told the opposite of what the diff does. I've rewritten it to match the implementation, with an explicit note at the top that the scope changed during review and why.

**`draft_num_predict` missing.** Correct — `Runner.DraftNumPredict` maps to `draft_num_predict`, and `buildOllamaOptions` was classifying it as unsupported and dropping it, so speculative-decoding configs couldn't reach Ollama. Verified the live server accepts it (HTTP 200 on `/api/generate`), then added it to the interface, the allowlist, the documented key list, and the forwarding test.

That one is a good illustration of why the allowlist approach is fragile: a valid new option is silently dropped until someone notices. The debug line added in this PR at least makes it visible now, but it's worth considering whether the allowlist should eventually invert to a denylist.

72 tests pass; `tsc`, lint, format, and `architecture:check` all clean.

PR #10962, opened 2026-09-16
docs(site): correct and expand Ollama provider docs

mldangelo:
> **Top of the stack** (#10953 → #10954 → #10958 → #10961 → this). I'll retarget as the chain lands.

## Summary

Line-by-line audit of `site/docs/providers/ollama.md` against the code and a live Ollama 0.13.5 server. Every fix below was verified, not inferred.

### The claim that actually misleads people

> "Only one model is loaded into memory, conserving system resources."

**False.** `grep -rn keep_alive src/` returned nothing before this PR series — promptfoo never sent `keep_alive`, so Ollama holds each model for its own 5-minute window. After a serial two-provider eval, `ollama ps` showed **both** models still resident:

```
NAME          ID              SIZE      PROCESSOR   UNTIL
qwen3:0.6b    7df6b6e09427    807 MB    100% GPU    4 minutes from now
```

That's exactly backwards for the audience the section names ("Local setups with limited RAM"). Replaced with an accurate description of what `-j 1` does, plus the working remedy — `keep_alive: 0`, verified to unload immediately (`ollama ps` empty right after).

### Other corrections

| Fix | Evidence |
|---|---|
| Serial grading-grouping precondition listed 2 conjuncts | `evaluator.ts:4947` requires 3: `concurrency === 1 && !hasEvalStepTimeout && !usesConversationVar` |
| `github.com/jmorganca/ollama`, `ollama.ai/library` | both return 301 |
| `api/client.go#L19` cited for the default host | that line no longer exists; default moved to `envconfig/config.go` |
| Undocumented provider id forms | registry accepts bare `ollama:<model>` (→ completion) and singular `ollama:embedding:` |
| Silent config-key drops undocumented | now stated, with the supported `options` keys listed and a warning that `max_tokens` is an OpenAI key Ollama ignores — the exact mistake this repo's redteam example shipped |
| `tools: file://...` undocumented | `maybeLoadToolsFromExternalFile` has always supported it; also noted `tools` is chat-only, since `/api/generate` has no tool support |
| Stale model lists | every name now resolves on `ollama.com/library` (checked by HTTP status); added small models so the docs can be smoke-tested without a large download |
| 6 of ~20 config keys documented, no pointer | linked Ollama's upstream parameter reference |

## Verification

```
cd site && SKIP_OG_GENERATION=true npm run build
[SUCCESS] Generated static files in "build".
```

No broken links or anchors. All 11 model names checked against `ollama.com/library` and returning 200.


mldangelo:
Both fixed in `9bc9844`.

**Code-block titles (P1).** Correct — `site/AGENTS.md` reserves `title="filename.yaml"` for complete, runnable files, and the two blocks this PR added (external tools, `keep_alive`) are provider-only fragments. Titles dropped.

I deliberately left the pre-existing titled fragments alone. The same file says "Don't rewrite — improve incrementally," and retitling eight untouched blocks would bury the actual corrections in this PR. Happy to do it as a separate sweep if you want the file made consistent.

**`-j 1` serialization (P2).** You're right that I overstated it, and it's a bit ironic given this PR exists to fix an overstated claim in the same section. `-j 1` serializes target calls, but `src/assertions/index.ts:822` only serializes graders when the grouping queue is active:

```ts
const concurrency = getProviderCallExecutionContext()?.providerCallQueue
  ? 1
  : ASSERTIONS_MAX_CONCURRENCY;   // 3 by default
```

So with a per-eval timeout or `{{_conversation}}`, grader calls can still overlap within a test case. Reworded to distinguish the target path from the grading path and to name `PROMPTFOO_ASSERTIONS_MAX_CONCURRENCY` explicitly.

Docs site still builds clean with no broken links.

mldangelo:
@codex review

mldangelo:
Both fixed in `bc9c7bf`.

**Modelfile link 404.** Confirmed directly — `docs/modelfile.md` returns **404**, `docs/modelfile.mdx` returns **200**. Updated. Good catch; I'd verified every *model name* against `ollama.com/library` by HTTP status in this PR but didn't apply the same check to the one external link I added.

**`-j 1` ordering.** Right again, and this is the second time this sentence has been too strong. `src/evaluator.ts` partitions `options.runSerially` cases into `serialRunEvalOptions` and runs that partition ahead of the rest, so row order isn't strict test-case order even at concurrency 1. I dropped the ordering claim from the main sentence and added an explicit caveat rather than trying to qualify it inline.

Docs site still builds clean with no broken links or anchors.

jameshiester-oai (commented):
Reviewed declared base `df4a31e37aeb671ccd1dd7bf6f91bacc2adb2126` and exact head `ee077031abfa84156c2005469ddf02ffb877ec39`. Full merge-base diff review. The initial stacked diff included Ollama source/tests, which were reviewed and tested; after the main merge, the final live PR diff is documentation-only.

No new actionable defect in the final live PR diff. The typical_p acceptance claim conflicts with released Ollama v0.34.1, but final diff validation showed it is already in the current base; it is an inherited documentation caveat, not a new finding on this PR. Provider aliases, tools loading, options/passthrough placement, embedding controls, target-call serialization and grader concurrency caveats match source. Existing comments on snippet titles and global serialization are addressed in the current text; unresolved UI state was not treated as proof they remain defects. Three specialist reports reconciled; no material simplification identified.

Official [Ollama FAQ](https://docs.ollama.com/faq) and [Qwen3](https://ollama.com/library/qwen3:0.6b)/[all-minilm](https://ollama.com/library/all-minilm) pages consulted for host, residency and model facts. Verified against [Ollama v0.34.1 request validation](https://github.com/ollama/ollama/blob/v0.34.1/server/routes.go#L215-L218): server support is version-dependent and typical_p is rejected. This was validated from tagged source, not by running an Ollama server.

Validation (Node24.21.0/npm11.19.0):
- `npx vitest run test/providers/ollama.test.ts`:73 passed, exit0.
- `cd site && SKIP_OG_GENERATION=true npm run build`: exit0; rendered Ollama page exists, no broken-link build failure. Nonfatal Docusaurus update-check permission notice.
- `npm run build`: typecheck, compilation and app/declarations passed; postbuild launcher hit tsx IPC EPERM. `node --import tsx scripts/postbuild.ts`: exit0.
- `python3 /tmp/ollama-review-wire.py`: exit0. Built CLI `eval -c <fixture> --no-cache -j 1 -o <JSON>` passed two evals with zero errors. Six loopback requests covered bare completion, chat and singular embedding aliases; keep_alive:0 remained top-level, passthrough.options merged, and /api/embed sent input with truncate:false. Scores/output inspected.

No local Ollama binary/model was available; model residency itself was checked against official documentation, not claimed as live local QA. No browser visual inspection. Requested `/pr-review-toolkit:review-pr all parallel` unavailable (exit127); manual review fallback completed. CI snapshot is recorded below; review completion is not merge approval.

Restarted once after head advanced; rechecked the new Ollama diagnostic exclusion and note, separated merged-main changes, and reran provider tests, builds and actual CLI. No clean reinstall after merged dependency-only changes was performed.

CI: {'SUCCESS': 41, 'IN_PROGRESS': 2}.

PR #10964, opened 2026-09-16
feat(providers): support Ollama 0.34 features and refresh model docs

mldangelo:
## Summary

Ollama 0.34 added capabilities the provider couldn't reach. I upgraded my local server from 0.13.5 to **0.34.0** and verified each of these live.

### 1. `think` accepts a level, not just a boolean

Ollama 0.34 takes `'low' | 'medium' | 'high' | 'max'` in addition to `true`/`false`. The type was `boolean`, so levels were unreachable without `passthrough`. All six values confirmed HTTP 200 against the live server, and end to end:

```
think-level-low   -> "Thinking: Okay, the user is asking for the capital of France..."
think-level-high  -> "Thinking: Okay, the user is asking for the capital of France..."
```

### 2. `format` — structured outputs, now first-class

`format` ('json' or a JSON schema) was reachable only via `passthrough`; set at the top level it was silently dropped. Verified live:

```yaml
config:
  format:
    type: object
    properties: { capital: { type: string } }
    required: [capital]
```
```
output: '{ "capital": "Paris" }'     # not a markdown-fenced block
```

That makes `is-json` and schema assertions work reliably instead of tripping over fences.

### 3. `/api/generate`-only parameters

`suffix` (fill-in-the-middle), `system`, `template`, and `raw` are now supported on the completion provider. Confirmed `system` reaches the wire and changes the prompt:

```
{"model":"qwen2.5:0.5b","prompt":"...","options":{...},"system":"You always answer in exactly one word."}
```

(`suffix` is model-gated upstream — a model without insert support returns `does not support insert`, which proves the parameter parses.)

### 4. `prompt_eval_cached_count` → `cacheReadInputTokens`

New in 0.34: prompt tokens served from Ollama's own KV prefix cache. Confirmed it's real and grows across calls:

```
call1: prompt_eval_count=44  prompt_eval_cached_count=29
call2: prompt_eval_count=44  prompt_eval_cached_count=43
```

**Deliberate decision:** this lands in `completionDetails.cacheReadInputTokens`, **not** `tokenUsage.cached` or `response.cached`. It's a *server-side* prefix cache, not a promptfoo cache hit — conflating them would make a fresh request look like a replay and suppress the provider delay. Measured through the CLI:

```json
{"prompt": 54, "completion": 25, "total": 79, "cached": 0,
 "completionDetails": {"cacheReadInputTokens": 53}}
```

Note `cached: 0` — correctly not a promptfoo cache hit.

## Docs

- Refreshed the model lists. **Every name checked against `ollama.com/library` by HTTP status**, including the newer `qwen3.5`, `qwen3-coder`, `gemma4`, `gpt-oss`, `deepseek-r1`, and the `bge-m3` embedding model.
- Added a note that capability varies by model, pointing at the library rather than asserting a static capability matrix that will go stale.
- New "Structured outputs" and "Completion-only parameters" sections.

## Tests

`npx vitest run test/providers/ollama.test.ts` → **85 passed** (from 73). New coverage: all six `think` values forwarded top-level and absent from `options`; `format` top-level; each of `suffix`/`system`/`template`/`raw`; and `prompt_eval_cached_count` mapping to `cacheReadInputTokens` without setting `cached`.

`tsc` exit 0; lint/format clean; `architecture:check` and `knip` pass; docs site builds with no broken links.

## Note

While wiring the cache field I initially put `cacheReadInputTokens` at the top level of `TokenUsage` behind an `as Partial<TokenUsage>` cast, which compiled. It belongs inside `completionDetails` (`src/contracts/shared.ts:8`) — the cast was hiding the mistake. Removed the cast and nested it correctly.


mldangelo:
@codex review

site/docs/providers/ollama.md
```diff
@@ -114,6 +119,42 @@ Options that newer Ollama releases removed (`mirostat`, `mirostat_tau`, `mirosta
...
 
+## Structured outputs
+
+Set `format` to `json`, or to a JSON schema, to constrain the model's output:
+
+```yaml
+providers:
+  - id: ollama:chat:qwen3
+    config:
+      format:
```
jameshiester-oai, line 131: [P2] Hide reasoning in the structured-output example

The example selects Qwen3 but leaves `showThinking` at its default true. When Ollama returns both `message.thinking` and JSON content, this provider prepends `Thinking: ...` to the JSON, so the advertised `is-json` assertion fails even though `format` reached the server correctly. Reproduced with the built CLI and a loopback reasoning+JSON response; adding `showThinking: false` makes the same assertion pass. Include that option in this example (or explicitly disable thinking) so its promised output is valid JSON.

site/docs/providers/ollama.md
```diff
@@ -114,6 +119,42 @@ Options that newer Ollama releases removed (`mirostat`, `mirostat_tau`, `mirosta
...
+`ollama:completion:*` uses `/api/generate`, which accepts a few parameters the chat
+endpoint does not:
+
+```yaml
+providers:
+  - id: ollama:completion:qwen2.5-coder
+    config:
+      system: 'You are a terse assistant.' # override the model's system prompt
+      suffix: '    return result' # fill-in-the-middle, for models that support insert
+      raw: true # bypass prompt templating
```
jameshiester-oai: [P2] Separate raw mode from system and template overrides

This example combines `raw: true` with nonempty `system` and `template`, which the provider forwards together. Ollama v0.34.1 rejects this combination with HTTP 400 (`raw mode does not support template, system, or context`), before generation; see [the tagged handler](https://github.com/ollama/ollama/blob/v0.34.1/server/routes.go#L431-L433). The built CLI wire fixture confirms all fields are sent and that rejection propagates. Set raw to false here or show raw mode in a separate example without system/template, otherwise copying this config cannot produce a completion.

jameshiester-oai (commented):
Reviewed base `811092a5b2a06a93574ce8aa02cfb2575aa7de2a` and exact head `92f9b8d12f2a77b5b0da2e1d0da618f66628897e`; full three-file merge-base diff. Two validated P2 documentation findings inline. No existing review threads at inspection. Three specialist reports reconciled; source parameter forwarding and server cache token accounting otherwise match their intended behavior. No material simplification identified.

Validation (Node24.21.0/npm11.19.0):
- `npx vitest run test/providers/ollama.test.ts`: 85 passed, exit0.
- `npm run build`: typecheck, compilation, app/declarations passed; postbuild launcher hit sandbox tsx IPC EPERM. `node --import tsx scripts/postbuild.ts`: exit0.
- `cd site && SKIP_OG_GENERATION=true npm run build`: exit0. No broken-link failure; nonfatal local update-check permission notice.
- `python3 /tmp/pr10964-review/wire.py`: fixture exit0. Built CLI eval with --no-cache exported four results: two expected failures (reasoning-prefixed JSON and incompatible raw options), two passing controls (showThinking:false and generate fields with raw:false). CLI exit100 is expected for these negative controls. Verified top-level format/think/system/template/suffix/raw placement and cacheReadInputTokens:53 without promptfoo cached status on chat and completion. Initial fixture assumed result order; corrected to compare provider labels and reran.

Raw-mode rejection checked against tagged Ollama v0.34.1 server source and emulated in loopback HTTP; no real Ollama binary/model run is claimed. Qwen3.5 and Gemma4 library pages checked for new model documentation. Requested `/pr-review-toolkit:review-pr all parallel` unavailable (exit127); manual six-lens review plus specialist fallback completed. CI snapshot:23 successful,19 in progress,1 neutral. Review completion is not merge approval.

mldangelo:
All seven findings addressed in `20de5f9`. Two of them caught genuinely broken documentation I'd written, and one corrected a claim I'd made in an earlier PR.

## Docs examples that didn't work as written

**@jameshiester-oai / @Copilot — structured output + reasoning.** You're both right, and I verified it end to end: the example as written returns `Thinking: Okay, the user is asking...` and `is-json` **fails**. Embarrassing, because I documented the thinking-prefix behavior myself in #10954 and then wrote an example that trips over it.

But the suggested remedy — `showThinking: false` — is **not sufficient**, and I only found that by testing it:

```
showThinking: false  ->  is-json pass=False,  output=''
think: false         ->  is-json pass=True,   output='{ "capital": "Paris" }'
```

Hiding the trace doesn't stop the model spending its `num_predict` budget on reasoning, so the JSON content comes back empty — the same truncation failure mode #10954 documented. The example now uses `think: false`, with a warning stating both facts so nobody reaches for `showThinking` here.

**@jameshiester-oai — `raw` vs `system`/`template`.** Confirmed on 0.34.0:

```
raw + system + template  ->  HTTP 400  "raw mode does not support template, system, or context"
raw alone                ->  HTTP 200
system alone             ->  HTTP 200
```

Split into two examples with the exclusivity spelled out.

**@Copilot — grammar.** Fixed.

## Code

**`numRequests` undercounting.** This one corrects me. In #10961 I argued omitting `numRequests` was a no-op because `tokenUsageUtils` increments when absent — that holds only for the evaluator path. `accumulateTokenUsage` defaults `incrementRequests` to `false`, and `src/matchers/rag.ts` / `similarity.ts` call the two-arg form, so grader requests through those paths report 0.

I checked the double-count risk before changing it, since that was my original reason for leaving it out:

```
without numRequests:  top=1, assertions=1
with numRequests:1:   top=1, assertions=1   <- no double-count
```

Safe and strictly more correct, so it's now explicit. That did churn 8 exact-shape assertions — worth it this time, unlike in #10961 where it bought nothing.

Note `llm-rubric` specifically already reported 1 (it goes through the response-aware helper), so the undercount is confined to the matcher paths you identified.

**`prompt_eval_cached_count` truthiness.** Correct — `0` is a real observation distinct from absent. Now `!== undefined`, with a regression test.

**Completion-path `format` coverage.** Added; chat and completion build separate request objects.

87 tests pass (from 85). `tsc`, lint, format, `architecture:check`, `knip`, and the docs build are all clean, and I re-ran every corrected docs example against the live server.

mldangelo:
One fixed in `f4fe5ae`, one I'm pushing back on.

## Endpoint-incompatible keys (P1) — real, and fixed

Good catch, and a pointed one: this is the same silent-drop failure mode the rest of this series exists to remove, reintroduced by my own PR.

`suffix`/`system`/`template`/`raw` are valid Ollama keys, but only `/api/generate` accepts them. Setting one on a chat provider fell into a gap — `OllamaNonNestedOptionKeys` kept it out of `options`, and `OllamaCompletionOptionKeys` marked it recognized so it was never reported as dropped. Verified before the fix:

```
config: {suffix: 'XYZ', system: 'terse', raw: true}  on ollama:chat:*
wire:   {"model":"qwen3:0.6b","messages":[{"role":"user","content":"hi"}], ...}
log:    (nothing)
```

Gone without a trace. `buildOllamaOptions` now takes the endpoint and warns:

```
[Ollama] Ignoring config keys that the chat endpoint does not accept: suffix, system, raw
```

This also covers `tools` on a completion provider, which `/api/generate` doesn't support. Confirmed valid per-endpoint configs are unaffected — the documented chat, completion, and raw-alone examples all still run. 89 tests (from 87).

## Commit scope (P1) — respectfully disagree

`docs/agents/pr-conventions.md` addresses mixed changes directly, and it's the canonical PR-conventions doc:

> ### Type Selection for Mixed Changes
> Use the **primary change** to determine type:
>
> | PR Contains | Type | Why |
> | --- | --- | --- |
> | Feature + documentation | `feat` | **Feature is primary** |

This PR is a provider feature whose docs describe that feature; the one site file exists because the feature needed documenting. `site/AGENTS.md`'s "always use `docs(site):`" governs docs-*only* changes — `pr-conventions.md` is what covers the mixed case, and it says `feat`.

There's precedent in this same series, all merged: #10954 (`feat(providers)`), #10958 (`fix(providers)`), and #10961 (`chore(providers)`) each touched exactly one `site/` file for the same reason.

The practical argument matters too: splitting would land the feature undocumented until a second PR merged, which is worse for users than a scope label.

Happy to re-title or split if a maintainer prefers — just say so. Flagging the disagreement rather than silently ignoring it.

src/providers/ollama.ts
```diff
@@ -123,8 +136,25 @@ const OllamaNonNestedOptionKeys = new Set<string>([
...
 
+/**
+ * Which top-level (non-`options`) keys each endpoint actually accepts. Anything outside
+ * its endpoint's set is reported as dropped rather than vanishing silently: `suffix` on
+ * a chat provider, or `tools` on a completion provider, is a config mistake worth
+ * surfacing.
+ */
+const OllamaEndpointTopLevelKeys: Record<'completion' | 'chat' | 'embedding', Set<string>> = {
+  completion: new Set(['think', 'keep_alive', 'format', 'suffix', 'system', 'template', 'raw']),
+  chat: new Set(['think', 'keep_alive', 'format', 'tools']),
```
jameshiester-oai: [P2] Keep `truncate` available on generation endpoints

This endpoint matrix now classifies `truncate` as embedding-only, so `config.truncate` on both `ollama:completion:*` and `ollama:chat:*` is warned as unsupported and omitted from the wire request. Ollama v0.34.0 defines top-level `truncate` on both [`GenerateRequest`](https://github.com/ollama/ollama/blob/v0.34.0/api/types.go#L102-L104) and [`ChatRequest`](https://github.com/ollama/ollama/blob/v0.34.0/api/types.go#L147-L149), where it controls context truncation. Add `truncate` to the completion and chat sets, serialize it in both request objects, and cover both wire paths so valid long-context configs are not rejected by Promptfoo's own allowlist.

jameshiester-oai (commented):
Reviewed current base `fcd06045caab8ea338d6eab4655d07e656ff883f` and exact head `a79eba2c3283876d6736d5366bff6c04b31d9b0e`; incremental from previously reviewed `92f9b8d12f2a77b5b0da2e1d0da618f66628897e`, including the one permitted restart after the head advanced from `20de5f91323d9569c6a1749f429880b6c90a0ef6` during the publication gate.

Verdict: one new P2 finding inline. The follow-up commit fixes the previously reported documentation examples and makes endpoint-incompatible first-class settings visible via warnings. The existing current thread asking to reject those settings is only partially addressed because requests still continue after the warning; I am not duplicating that feedback. No new security issue or material simplification opportunity found. CI snapshot at publication is still in progress with no reported failures; this review is not merge approval.

Validation (Node 24.21.0/npm 11.19.0, exact lockfile dependencies):
- `npx vitest run test/providers/ollama.test.ts test/matchers/token-tracking.test.ts test/util/tokenUsage.test.ts`: 105 passed, exit 0.
- Specialist deterministic and shuffled Ollama runs: 89/89 passed in each run.
- `npm run build`: TypeScript, core bundles, declarations, and app production build passed; the sandboxed `tsx` postbuild launcher hit its known IPC `EPERM`.
- `npm exec -- node --import tsx scripts/postbuild.ts`: exit 0.
- `SKIP_OG_GENERATION=true npm run build` in `site/`: exit 0; only GitHub stats fetches hit a nonfatal rate-limit warning.
- `python3 /tmp/pr10964-review/wire.py`: built-CLI loopback matrix reproduced reasoning-prefixed JSON and raw-mode rejection, while corrected `think:false` and separated completion settings passed. Expected aggregate exit 100 because the matrix intentionally includes two negative controls.
- `git diff --check`: exit 0.

The PR Review Toolkit command was unavailable (`/pr-review-toolkit:review-pr all parallel`, exit 127); the manual six-lens fallback and three independent read-only specialist passes were reconciled. No real Ollama model/server run is claimed in this pass.

mldangelo:
@jameshiester-oai — you're right, fixed in `586d6df`. This one was a regression I introduced in the previous commit, so thanks for catching it fast.

Measured on Ollama 0.34.0, 3000-word prompt with `num_ctx: 256` on `/api/generate`:

```
truncate: true   ->  HTTP 200
truncate: false  ->  HTTP 400
```

So `truncate` genuinely gates context-overflow behavior on the generation endpoints, not just `/api/embed`. It's now accepted and **forwarded** on all three (confirmed on the wire: `"truncate":false` / `"truncate":true` in a real eval), and `format` is allowed on embed.

## A methodological point worth recording

While fixing this I checked the rest of the matrix and found that **HTTP 200 does not prove a key is meaningful for an endpoint**. Ollama silently ignores unknown top-level keys:

```
suffix     on /api/chat     -> HTTP 200   (meaningless)
dimensions on /api/generate -> HTTP 200   (meaningless)
think      on /api/embed    -> HTTP 200   (meaningless)
truncate   on /api/generate -> HTTP 200 / 400 depending on overflow  (meaningful)
```

Only a *behavioral* difference is evidence. I'd built the original matrix partly from the API docs, which list `truncate` under `/api/embed` only — that's what led me wrong. I've put this note in the commit message so the next person extending the matrix doesn't repeat it.

Worth being explicit about the blast radius, since I got this wrong once: the matrix drives **a warning only**. Each provider still forwards its own explicit key set, so a wrong entry costs a spurious warning rather than a suppressed parameter. That's deliberate after this bug.

91 tests (from 89), including per-endpoint `truncate` forwarding cases. All gates clean.

mldangelo:
Fixed in `61f808f`. This one is embarrassing in a useful way: I added `format` to the embedding set in the previous commit because `/api/embed` returns HTTP 200 for it — **the exact reasoning I had just finished documenting as insufficient** in the commit right before.

`callEmbeddingApi` never forwarded `format`, so setting it on an embedding provider was neither sent nor warned about. A silent drop, produced by the mechanism I added to eliminate silent drops.

`format` is also meaningless for `/api/embed`, which returns vectors rather than text, so the fix is to remove it from the set and let the warning fire.

## The structural fix

Rather than just correct the entry, I added a test for the invariant that was violated:

> Every key listed as accepted for an endpoint must actually reach the wire for that endpoint.

```
it.each([
  ['chat',       ['think','keep_alive','format','truncate']],
  ['completion', ['think','keep_alive','format','truncate','suffix','system','template','raw']],
  ['embedding',  ['keep_alive','truncate','dimensions']],
])('every accepted %s key reaches the wire', ...)
```

It builds a config from every key in each endpoint's set, runs the real provider, and asserts each key appears top-level in the request body and not under `options`. Had this existed, both the `truncate` regression and this `format` entry would have failed the suite instead of needing two review rounds to find.

95 tests (from 92). All gates clean.

Three rounds on this PR have each found something real, and two of the three were regressions I introduced while fixing the previous round — so I'd rather the bots take another pass before this merges than call it done.

jameshiester-oai (commented):
Exact-head review complete for `61f808f0dd2780b3b640868a66984d3e6b527653`.

No distinct current P0-P2 findings. The two issues raised on earlier heads are fixed: `truncate` is accepted and forwarded for generation endpoints, and embedding `format` is now rejected with the intended unsupported-setting warning instead of being silently dropped.

Validation performed at this exact head:

- Three independent review lenses: correctness, tests/regressions, and endpoint/API risk.
- `git diff --check` passed for the incremental changes.
- Focused Vitest suite passed: 3 files, 111 tests.
- `npm run tsc` passed.
- Core bundle and app production build passed. The `tsx` postbuild wrapper hit the known sandbox IPC restriction; running the same postbuild script through Node's `tsx` import completed successfully.
- Built-CLI loopback wire fixture passed its expected assertions for structured output, reasoning visibility, completion options, raw-mode rejection, and token accounting.
- CI snapshot had no failures; some jobs were still running when reviewed.

The requested PR Review Toolkit command was unavailable in this environment, so I used the manual six-lens fallback plus the three independent review passes.

mldangelo:
Thanks @jameshiester-oai — exact-head review on `61f808f` confirming both fixes is exactly what I wanted before merging, given the last two rounds each surfaced a regression from the previous fix.

## On the remaining thread: warn vs. reject

You noted the "reject endpoint-incompatible settings" thread is only partially addressed, since requests continue after the warning. That's a deliberate choice, and I want to record the reasoning rather than leave it looking like an oversight.

The original finding named the case that decides it: *"for example through a shared provider config."* Sharing one `config` block across a chat and a completion provider is a legitimate pattern — the completion provider needs `suffix`/`system`, the chat provider ignores them. Erroring would break that working setup, turning a harmless extra key into a hard eval failure.

So the trade is:

- **Reject:** a typo is impossible to miss, but a legitimate shared config breaks.
- **Warn + continue:** the user sees the problem, and nothing that worked before stops working.

Warning also matches how this provider now treats every other unrecognized key, which keeps the mental model consistent: promptfoo tells you what it dropped, it doesn't fail your eval over it.

Happy to switch to hard rejection if a maintainer prefers strictness over compatibility — it's a one-line change — but I'd want that to be a deliberate call, not something I slipped in.

Merging now: CI 43/43 on this exact head, scanner clear, and your exact-head review reports no current P0–P2 findings.

PR #10969, opened 2026-09-16
fix(redteam): preserve strategy grading context

jameshiester-oai:
PII subcategories generate the shared `promptfoo:redteam:pii` assertion. The final assertion handler previously failed to recognize the strategy's grade and graded the response again against the original prompt. A name introduced during an attack could therefore be incorrectly reported as leaked PII when the target repeated it.

Stored grades now require a configured attack provider, a matching plugin and full assertion configuration, and a fingerprint of the plugin, returned prompt, output, and grading conversation where used. Older unbound grades and grades for different turns receive fresh grading while retaining usage from the trusted strategy. Assertions containing runtime objects or functions that cannot be fully represented in JSON are graded afresh. Failed transforms never enter the saved target conversation; runs with no completed target request return an error, and fallback responses retain their actual attack prompt even when all judge ratings are zero or judge calls fail. All seven producers record this binding. Hydra returns the same externalized output that its grader evaluated. Backtracking preserves the conversation belonging to the returned response, and iterative/tree selection keeps the verdict for the selected attempt while retaining total grading usage.

Fresh grading uses the transformed attack input, provider-reported string or chat-array input, or saved target conversation before falling back to the original prompt. Chat arrays retain prior user/assistant context without treating system/tool messages as user input or mixing unrelated conversation sources. Provider-reported strings suppress unrelated saved history unless the strategy supplies an authoritative final prompt. Custom and GOAT graders receive the same transformed input and prior target conversation used for the binding. Hydra/Goblin retain current-turn grading without adding prior conversation history, both during strategy execution and regrading; their conversation remains available for attack generation and reports. Stored Hydra/Goblin verdicts bound to history are graded afresh. Explicit regrading continues to bypass stored grades. Selecting an earlier verdict preserves all grading usage without marking later fresh work as cached. Partial regrading failures retain the strategy grading usage already incurred. Historical strategy usage is counted once per assertion run across duplicate checks and assertion sets, and response-cache hits count the full replayed token total. A provider-reported chat without a usable user message suppresses unrelated saved history; stored verdicts are checked against the selected context source. Tree search retains the grade from the selected best attempt even when a later attempt repeats the same prompt and output.

Validation:
- Full local backend run: 26,868 tests pass across 869 files with `--maxWorkers=2 --maxConcurrency=1`; seven OTLP receiver tests encountered local loopback transport errors (`EADDRNOTAVAIL`/HTTP 502). All 50 tests in that file pass on an isolated retry. The focused assertion/provider/evaluator suite passed 345 tests before the final four context-binding regressions, which pass in the full run. Regressions cover grader errors, target-ended conversations, backtracking, selected earlier attempts, cached and fresh grading usage, changed conversation context, output transforms, missing/mismatched provenance, chat-array inputs, different assertion configurations, externalized Hydra output, and runs with no completed target request.
- 48 deterministic CLI cases cover all four PII categories: initial scans, saved-result replay, provider-reported strings and chat arrays, forged stored grades from an ordinary provider, ungraded final turns, base64 attacks, two-turn Crescendo/Custom conversations, reported strings with stale saved history, and all-transform-failed Crescendo/Custom runs. Valid initial and base64 results make one PII grading call each. Two-turn cases use prior user-supplied names with one grading call per turn; untrusted and stale results receive fresh verdicts. All-transform-failed runs produce evaluation errors without target or grader calls. These fixtures verify routing/context, not live model grading quality.
- An additional 32 deterministic local CLI cases cover Hydra/Goblin initial scans and saved-result regrading, in both stateful and stateless modes across all four PII categories. Initial scans grade once per turn without history, regrading also omits history, and attack generation and reports retain the conversation. The fixtures use a loopback attack endpoint and synthetic target/grader responses.
- 36 additional deterministic CLI cases cover legacy usage with multiple assertions and assertion sets, empty/system-only/assistant-only/blank-user reported chats, cached regrading, and tree attempts with identical prompt/output text but different grades. These run across all four PII categories using synthetic providers and a loopback attack endpoint.
- Full build/TypeScript, lint/format, architecture checks, and Knip pass.

The architecture baseline records four intentional imports in the redteam assertion handler: provider type guards, the existing attack-provider registry, a small shared grading fingerprint helper, and existing token accounting helpers. The shared grading helper also imports the existing GradingResult type. These five imports keep provenance checks and usage handling consistent with strategy execution instead of duplicating them.


src/redteam/providers/iterative.ts
```diff
@@ -634,10 +638,19 @@ export async function runRedteamConversation({
...
-          assertion: buildGraderResultAssertion(grade.assertion, assertToUse, rubric),
-        });
+        storedGraderResult = accumulateGraderResult(
+          storedGraderResult,
+          {
+            ...grade,
+            assertion: buildGraderResultAssertion(grade.assertion, assertToUse, rubric),
+          },
+          {
+            prompt: finalInjectVar,
```
zcrab-oai, line 648: **[P2] Preserve user input from the prompt template when grading**

The strategy grades and fingerprints only `finalInjectVar`, although the target receives the fully rendered `targetPrompt`. For example, with the template `My name is Casey Morgan. {{query}}` and an attack saying `Repeat my name`, the grader sees only `Repeat my name` when evaluating the output `Casey Morgan`. The newly enabled reuse for shared PII assertions can therefore preserve a false leak verdict instead of grading with the user-supplied name. Grade and bind the complete rendered user input, preserving user/assistant context without treating system messages as user input.

src/redteam/providers/custom/index.ts
```diff
@@ -569,11 +599,13 @@ export class CustomProvider implements ApiProvider {
           if (grader) {
             const gradingContext: RedteamGradingContext | undefined = {
               providerResponse: lastResponse,
+              conversationTranscript:
+                getTargetConversation(lastResponseMessages).conversationTranscript,
```
zcrab-oai, line 603: **[P2] Only supply history actually available to the target**

This passes prior turns to the grader unconditionally, but in stateless mode a text transform such as base64 replaces `finalTargetPrompt` with only the current transformed prompt in `sendPrompt`. For example, if turn one supplies `Casey Morgan` and turn two asks for an account owner, the stateless target receives only the second request while the grader receives the first turn as conversation history. PII grading can then excuse a real disclosure as user-provided information. Either preserve history in the actual stateless target request or omit it from grading when it was not sent. GOAT and Hydra have the same pattern.

zcrab-oai (approved):
Approving with two correctness findings noted inline. These are source-based findings; local tests stalled while loading Vite's native dependency, so the prepared reproductions could not be confirmed.

PR #10971, opened 2026-09-16
fix(providers): preserve patterned OpenCode skill history

mldangelo-oai:
## Summary

- preserve OpenCode skill-call tracking when a permission policy allows a wildcard set but denies a named skill
- conservatively fetch session history whenever any relevant permission rule can allow or ask for a skill
- cover the patterned-policy case with a focused regression test

## Bug

OpenCode resolves permission rules per matching pattern. The session-history optimization instead treated the final `skill` rule as a global decision. With a policy such as `* = allow` followed by `blocked-skill = deny`, Promptfoo skipped history entirely and lost allowed skill calls from intermediate turns, causing `skill-used` assertions to false-negative.

## Validation

- failing-first regression test reproduced the skipped `session.messages` call
- `npx vitest run test/providers/opencode-sdk.test.ts` (148 tests passed)
- `npm run f`
- `npm run l`
- `git diff --check`

## Local environment note

An exact clean install could not complete because the runner volume ran out of space. The shared dependency tree is older than the current Claude Agent SDK lockfile, so the repository-wide TypeScript build reports unrelated `thinkingTokens` type errors from current main. PR CI will validate the exact lockfile dependency set.


mldangelo-oai:
@codex review

jameshiester-oai (commented):
Reviewed base `fcd06045caab8ea338d6eab4655d07e656ff883f` and exact head `7f78563bc14f83c7cb88553642b8ff99e38b8904`. Full review of the ordered OpenCode permission conversion, effective provider/custom-agent overlays, history-fetch decision, bounded attribution path, and regression coverage.

No new distinct actionable defect found. Existing unresolved P2 [#discussion_r4031798948](https://github.com/promptfoo/promptfoo/pull/10971#discussion_r4031798948) is valid on this exact head: `some(non-deny)` ignores later blanket rules that shadow every earlier allow/ask, so configurations such as `tools.skill: true` followed by `permission.skill: deny` still fetch full session history although skills cannot run. This adds avoidable latency, server load, and local data retrieval, but does not enable denied skills or misattribute calls. I did not duplicate the inline comment.

The intended fix works: when a wildcard rule allows skills and a later named rule denies one skill, other names remain possible, so conservatively fetching history prevents `skill-used` false negatives. Returning false only when every potentially applicable skill/wildcard rule is a denial introduces no inverse missed-skill configuration. Three independent correctness, test and risk passes agreed.

Validation (Node 24.21.0 / npm 11.19.0):
- `npx vitest run test/providers/opencode-sdk.test.ts`: exit 0, 148 passed.
- Independent deterministic and shuffled runs of the same suite: each exit 0, 148 passed.
- `npm run tsc`: exit 0.
- `git diff --check`: clean. Current CI snapshot: 43/43 successful.

No live OpenCode service, full backend suite, production build or browser QA was run; this change is isolated to the provider decision and its mocked SDK regression suite. Requested `/pr-review-toolkit:review-pr all parallel` was unavailable (exit 127), so the manual correctness, tests, documentation, error-handling, API/type and simplification fallback was used. Review completion is not approval.

PR #10979, opened 2026-09-18
fix(providers): honor bustCache in the Ollama completion provider

mldangelo:
## Summary

`OllamaChatProvider` forwards `context.bustCache` to `fetchWithCache`; `OllamaCompletionProvider` dropped the context entirely, so cache busting was silently ignored on `/api/generate`.

This came out of the audit behind #10946/#10953 — the verifiers split on whether it was reachable in practice, so I left it open and said it was worth settling. It is reachable.

## Why it matters

Callers pass `bustCache` directly into `callApi`:

- `src/redteam/commands/discover.ts:315` — `bustCache: true`
- `src/redteam/strategies/gcg.ts:69` — `bustCache: true`

And the bare `ollama:<model>` provider id routes to the **completion** provider (`src/providers/registry.ts`). So target discovery against an Ollama target replayed cached answers instead of probing the live target — the probe believes it is seeing fresh behavior when it isn't.

## Measured

Two identical calls at `temperature: 1.5` with `bustCache: true`, against a live Ollama with an isolated `PROMPTFOO_CACHE_PATH`:

```
before   chat        cached=[undefined, undefined]  identical=false
before   completion  cached=[undefined, true]       identical=true    <- stale replay
after    completion  cached=[undefined, undefined]  identical=false
```

Ordinary caching is unaffected — without `bustCache`, the second completion call still reports `cached=true`.

## Scope

The embedding provider can't be fixed the same way: `callEmbeddingApi(input: string)` takes no context parameter (`src/types/providers.ts:126`), so there is nothing to thread. That's a framework-level limitation rather than an Ollama bug, and no caller passes `bustCache` to an embedding provider today.

## Tests

`npx vitest run test/providers/ollama.test.ts` → **98 passed** (from 96). Parameterized over `bustCache: true`, `debug: true` (which also busts, matching chat), and neither. Both new cases fail against `main`.

`tsc` clean for this file (there is a pre-existing unrelated `@aws/bedrock-token-generator` resolution error on `main`); lint, format, and `architecture:check` all pass.


mldangelo:
@codex review

mldangelo:
@codex review

mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed base `34f74d34e140b5e17d23770dfb2340057b1936b8` through head `71f56f3fc0faf4f1ba45b3bdfd0270745efea405` (fresh full review).

Verdict: no actionable defects found. Completion now forwards the caller context and uses the same `bustCache ?? debug` precedence as the Ollama chat provider; no-context callers retain ordinary caching.

CI: all listed checks passed.

QA: `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/ollama.test.ts --sequence.shuffle=false` — 98 passed; `git diff --check` passed. Reviewed route/caller reachability and the focused cache-bypass test matrix.

Limitations: no live Ollama request was made because this is a context-forwarding change with mocked transport coverage. The requested `/pr-review-toolkit:review-pr all parallel` launcher is unavailable; three independent read-only correctness, QA, and risk reviews were reconciled.

PR #10980, opened 2026-09-18
fix(providers): accept Ollama tool calls back in a multi-turn conversation

mldangelo:
## Summary

The Ollama provider normalizes tool calls in its **responses** to the OpenAI shape, where `function.arguments` is a JSON *string* — that's what makes `is-valid-openai-tools-call` work. But Ollama's own `/api/chat` rejects that shape on the way back **in**:

```
arguments as STRING -> HTTP 400
arguments as OBJECT -> HTTP 200
```

So feeding a previous turn's tool call into a multi-turn prompt fails. End to end, a prompt containing the provider's own output:

```
Ollama API error: 400 Bad Request: Value looks like object, but can't find closing '}' symbol
```

which gives no hint that the argument *shape* is the problem.

This was finding **L** from the audit behind #10946. I originally reported the dropped tool-call `id` as the multi-turn blocker; verification refuted that (Ollama correlates tool results positionally and accepts messages with no id at all) and isolated this as the actual cause. It was left unfixed at the time.

## Change

Outgoing messages convert a stringified `function.arguments` back to an object.

Non-object JSON and unparseable strings are deliberately left untouched, so Ollama reports a meaningful error rather than promptfoo silently reshaping a payload it doesn't understand.

## Verified with the real loop

Took an actual two-tool-call response from `qwen3:0.6b`:

```json
[{"function":{"name":"get_weather","arguments":"{\"city\":\"Amsterdam\"}"}},
 {"function":{"name":"get_weather","arguments":"{\"city\":\"Paris\"}"}}]
```

fed it back verbatim as an assistant message with a tool result, and got:

```
success=True
output='The weather in Amsterdam is 12°C, and in Paris it is 18°C.'
```

Before this change that request returned HTTP 400.

## Tests

`npx vitest run test/providers/ollama.test.ts` → **99 passed** (from 95). Parameterized over stringified, already-object, unparseable-string, and JSON-scalar-string arguments, asserting non-tool messages pass through untouched. The stringified case fails against `main`.

Lint, format, `architecture:check`, and `knip` all clean.


mldangelo:
@codex review

mldangelo:
@Copilot — correct, and this was a genuine regression I introduced. Fixed in `c76c935`.

`parseChatPrompt` returns whatever the prompt parsed to, not necessarily an array (`src/providers/shared.ts:147`), so a valid JSON object prompt reached the helper as an object and `messages.map` threw before the request was even built.

Worse, it made the failure *less* useful than before this PR — verified both against a live server with `'{"role":"user","content":"Say hi in one word"}'`:

```
main      Ollama API error: 400 Bad Request: json: cannot unmarshal object
          into Go struct field ChatRequest.messages
this PR   TypeError: messages.map is not a function
```

Non-array values now pass through untouched, so Ollama's own validation produces the useful message again. Confirmed the post-fix error string matches `main` exactly.

Added parameterized coverage for object, string, and number prompts asserting the body is passed through unchanged, and re-verified the actual multi-turn round trip still works (`'The weather in Amsterdam is 12°C, and in Paris it is 18°C.'`).

102 tests, from 99.

mldangelo:
@codex review

mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed base `34f74d34e140b5e17d23770dfb2340057b1936b8` through head `9e0addaa517d9bd5474456330ab4889a8221bb44` (fresh full review).

Verdict: no actionable defects found. Request normalization reverses only the provider-owned JSON-object argument encoding, retains malformed/scalar/array/null values, and guards non-array prompts. The prior non-array feedback is fixed.

CI: all listed checks passed.

QA: `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/ollama.test.ts --sequence.shuffle=false` — 117 passed; `git diff --check` passed.

Limitations: no live Ollama loop was available; transport behavior is covered by the focused mock suite. The requested toolkit launcher is unavailable; three independent read-only reviews were reconciled.

mldangelo:
On the suppressed test-coverage nit — I checked rather than adding tests, and those cases are already in the matrix on the exact head reviewed (`9e0addaa51`):

```
test/providers/ollama.test.ts:981   ['arguments JSON array string',  TOOL_CALL('[1,2]'), null]
test/providers/ollama.test.ts:982   ['arguments JSON null string',   TOOL_CALL('null'),  null]
test/providers/ollama.test.ts:983   ['arguments JSON number string', TOOL_CALL('42'),    null]
```

All 22 matrix cases execute (`vitest -t "normalizes"` → 22 passed). The `null` and array branches the comment asks about are covered, so I'm not adding duplicates.

My guess is the comment was drafted against the earlier head, before I replaced the four ad-hoc cases with the contract table — the older version genuinely only had object / unparseable / scalar.

PR #10984, opened 2026-09-18
fix(providers): degrade gracefully on malformed Ollama responses

mldangelo:
## Summary

After a review round caught an input shape I hadn't considered in #10980, I probed the **response-side** transformations the same way — systematically, across the shapes a buggy model server or an intervening proxy could emit. Three leaks, all reachable on `main` today:

```
tool_calls not an array      ->  Ollama API response error: TypeError: Cannot read
null tool call                   properties of undefined
tool call without function
null function

non-string thinking          ->  output: "Thinking: [object Object]\n\nhi"
non-string content           ->  output: "[object Object]"
```

Neither failure mode is acceptable for an eval tool: a raw `TypeError` surfacing as an eval error tells the user nothing about what went wrong, and rendering an object into the output **silently corrupts the value an assertion then grades**.

## Change

- `collectOllamaToolCalls` guards a non-array `tool_calls` and skips entries whose `function.name` isn't readable, instead of throwing.
- Content and thinking accumulation accept only strings.

The case I care most about: a response mixing readable and unreadable tool calls now **keeps the readable ones** rather than failing wholesale.

```
[null, {"function":{"name":"f","arguments":{"a":1}}}]
  ->  [{"function":{"name":"f","arguments":"{\"a\":1}"}}]
```

## Tests

Stated as a contract table rather than ad-hoc cases, matching the endpoint-key matrix test from #10964: every listed malformed input must produce no error and a sensible output.

`npx vitest run test/providers/ollama.test.ts` → **103 passed** (from 95). All 8 new cases fail against `main`.

Live regression check unaffected: a real two-tool-call response from `qwen3:0.6b` still returns 2 calls and passes `is-valid-openai-tools-call`.

## Reachability, stated honestly

Real Ollama doesn't emit these shapes — I couldn't produce them from the server directly, so the tests use fixtures. They're reachable through a proxy, gateway, or an `OLLAMA_BASE_URL` pointed at an OpenAI-compatible shim. This is robustness work (P3), not a live bug fix; I'd merge it on its own merits rather than urgency.


mldangelo:
@codex review

jameshiester-oai (commented):
Reviewed base `34f74d34e140b5e17d23770dfb2340057b1936b8` through head `00e3f5defa6937848bc1544dc828a7313f469ba5` (fresh full review).

Verdict: no distinct new defects. The chat-side hardening safely ignores malformed tool-call entries and non-string content/thinking while preserving readable calls.

Existing unresolved feedback, verified without duplication: the completion-path `thinking` guard lacks a matching completion fixture; completion `response` accumulation also still accepts non-strings and can stringify an object. This is pre-existing code but remains relevant to the stated robustness scope.

CI: all listed checks passed.

QA: `source ~/.nvm/nvm.sh && nvm use && npx vitest run test/providers/ollama.test.ts --sequence.shuffle=false` — 103 passed; `git diff --check` passed.

Limitations: no live malformed upstream was available; coverage uses fixtures. The requested toolkit launcher is unavailable; three independent read-only reviews were reconciled.

mldangelo:
@Copilot — right, and writing that coverage immediately paid for itself. Fixed in `7aa1431`.

The completion provider does have its own parsing branch, and adding the test you asked for surfaced a case the original commit missed: I had guarded chat `content` but **not** completion `response`, so a non-string response still rendered as `[object Object]`:

```
expected '[object Object]' to be ''
```

So the gap wasn't only in the tests — it was in the hardening itself, on exactly the branch you pointed at. Both completion `response` and `thinking` now accept only strings, matching the chat side.

Added three completion-side cases: non-string `thinking`, non-string `thinking` with an empty response, and non-string `response`. 106 tests, from 103.

Live check unaffected — real chat and completion calls still succeed, and the 400/500 error paths still surface their Ollama messages.

mldangelo:
Re-requesting review — the 🟡 verdict above was against head `00e3f5defa`, before the completion-endpoint coverage was added.

Current head `862f97d8a2` includes the fix that coverage surfaced: completion `response` was still rendering `[object Object]` because I had only guarded chat `content`. Both `response` and `thinking` now accept strings only, and the branch is synced with `main` (which has since picked up #10979 and #10980).

131 tests passing, CI green on the previous head.

PR #10985, opened 2026-09-18
fix(cli): validate test provider references against all defined providers when filtering

### LINKED ISSUE #10932: --filter-providers makes an exception
**Describe the bug**
When using the `--filter-providers` option, I got this error:
`ProviderReferenceValidationError: Test #1 ("Test Skill reads the config file first") references provider "gemini" which does not exist. Available providers: openrouter:deepseek/deepseek-r1, openrouter:free`

**To Reproduce**

I have a test with following providers (that are matching provider configuration):
```
  - description: Test Skill reads the config file first
     providers:
       - gemini
       - openrouter:*
```

When running:  `npx promptfoo eval --filter-providers openrouter:*`
i got this execption:
`ProviderReferenceValidationError: Test #1 ("Test Skill reads the config file first") references provider "gemini" which does not exist. Available providers: openrouter:deepseek/deepseek-r1, openrouter:free`

**Expected behavior**
The test should run on `openrouter:*` providers only, as indicated by the documentation:

> CLI --filter-providers: If you use --filter-providers to filter providers at the CLI level, validation only sees the filtered providers. Tests referencing providers excluded by --filter-providers will fail validation

- Promptfoo version: 0.122.2

siddiqueirshad:
Fixes #10932. When --filter-providers is enabled, provider reference validation now checks against all defined providers in the config rather than throwing ProviderReferenceValidationError for tests referencing valid defined providers excluded by the filter.

---

**Updated after merging `main`:** two refinements, no change to the behaviour described above.

- The unfiltered `loadApiProviders` call now runs *after* the filtered one and only when filtering actually removed a provider. Loading it first changed the order of `loadApiProviders` calls, which broke two existing tests in `test/util/config/load.test.ts` that assert on `mock.calls[0][0]`; making it conditional also avoids instantiating providers the user explicitly asked to skip on the common (unfiltered) path.
- Added a wiring-level regression test in `test/util/config/load.test.ts`. The test added here originally exercises `validateTestProviderReferences` in isolation and passes with or without the `load.ts` change — the new test drives `resolveConfigs` with `--filter-providers` and fails if the validator is handed the filtered set again.


siddiqueirshad:
Hi @he-yufeng, when you get a chance, could you please take a look at this PR? Thanks!

PR #10987, opened 2026-09-18
docs(site): fix broken provider examples and make docs evergreen

mldangelo:
## Summary

Audited all 94 pages in `site/docs/providers/` against their implementations in `src/providers/`. Fixes fall into three groups.

### Examples that could not run as written

| Page | Problem |
| --- | --- |
| `sagemaker.md` | 11 examples used `sagemaker:<endpoint>`, which throws — the model type is required. Also: `modelType` documented as defaulting to `custom` (there is no default), the `openai` model type and `basePath` missing, and a JumpStart auto-detect note that was wrong in both directions (it doesn't fire on the 2-segment form, and it *overrides* an explicit type). |
| `sequence.md` | Every example put the provider in top-level `providers:`, where it recurses into itself and dies with `RangeError: Maximum call stack size exceeded`. It's a test-level provider. |
| `aimlapi.md`, `vercel.md` | Embedding providers used as eval providers (they throw `Not implemented`), plus `is-valid-embedding` / `embedding-dimension`, which aren't real assertion types. |
| `aws-bedrock.md` | The headline `## Example` used a withdrawn model id that throws `Unknown Amazon Bedrock model`. |
| `azure.md` | Reasoning options documented for `azure:completion`, which supports none of them; a `service_tier` form never sent on the wire; the non-existent `AZURE_BASE_URL`; a legacy `openai:responses:` path that 404s. |
| 5 YAML blocks | Did not parse (`- provider` + indented `config:`, an unquoted `label:` containing `": "`). |

### Claims that didn't match the code

- **`simulated-user.md`** — `###STOP###` is read from the *simulated user*, not the agent, and the turn never reaches the agent. The page's own transcript already contradicted its prose.
- **`litellm.md`** — `OPENAI_API_KEY` is never consulted; `apiKeyEnvar` is pinned to `LITELLM_API_KEY`. Combined with `apiKeyRequired: false`, the documented setup fails auth **silently**.
- **`alibaba.md`, `mcp.md`** — options documented that are silently discarded (`vl_high_resolution_images`, `defaultArgs`), plus an embeddings syntax that routed to `/chat/completions`.
- **`envoy.md`, `f5.md`** — examples that couldn't authenticate, and an `apiBaseUrl` that builds `undefined/path-name` or a doubled path.
- **`xai.md`, `vertex.md`, `anthropic.md`, `deepseek.md`, `nvidia.md`, `localai.md`, `novita.md`** and others — corrected prices, defaults, env vars and model ids.

### Evergreen hardening

The point of this pass was to remove the things that rot, not just fix today's errors:

- **Duplicated price tables → mechanism + vendor link.** `hyperbolic.md` carried "rates as of January 2025" that were wrong by up to 20x against promptfoo's own table (Qwen3-235B listed at $0.40/M; actual $2.00 in / $8.00 out). Pages now explain that promptfoo ships a built-in table, link the vendor's pricing page as source of truth, and show the `inputCost`/`outputCost` override.
- **Lapsed future-tense deadlines** ("remain available until July 24, 2026" — that date has passed) rewritten as current-state. Past-tense history ("retired on X") was deliberately left alone.
- **Unverifiable vendor marketing** ("up to 80% savings", "zero rate limits") replaced with durable text.
- **Two `#L` GitHub permalinks unpinned** — both already pointed at the wrong lines.
- **20 dead external links repaired**, preferring stable URLs over deep slugs. Notably `f5.ai` was not F5 at all (it resolves to an Alibaba Cloud address in Shanghai); now points at F5's real product page.
- **Missing frontmatter titles** added.

### Removing retired content

Rather than preserving dead services behind historical notes, they're deleted:

| Page | Before | After | Why |
| --- | --- | --- | --- |
| `github.md` | 252 lines | 39 | GitHub retired the inference API on 2026-07-30; 233 lines were a collapsed "Historical configuration" block. |
| `llamaApi.md` | 393 lines | 38 | Meta retired the Llama API Public Preview on 2026-07-06. Setup, multimodal, tool calling, streaming, rate limits and best practices all described an endpoint that no longer answers. |
| `mistral.md` | ~30 ids listed | 2 lines | promptfoo keeps their pricing to cost-score cached results; that reason is now stated instead of enumerating ids that keep changing. |
| `deepseek.md` | 2 alias sections | removed | `deepseek-chat` / `deepseek-reasoner` are past their announced retirement. Examples now lead with `deepseek-v4-*`. |

No page was deleted, so existing URLs still resolve, and neither rewritten page had inbound anchor links into the removed sections. `index.md` and `meta.md` no longer advertise the Llama API as current.

One deliberate omission: the `llamaApi.md` migration example doesn't name a replacement model, because the obvious candidate (`llama-3.3-70b-versatile`) is itself deprecated on Groq's lower tiers — naming it would recreate the rot this pass removes.

## Verification

- All **849** YAML blocks across the provider docs parse.
- `SKIP_OG_GENERATION=true npm run build` — clean, no broken-link or MDX warnings.
- The rewritten `sequence.md` example **runs and passes** against the local build.
- Provider construction checked directly via `loadApiProvider` for the sagemaker, envoy and f5 changes.
- No heading text changed and no `highlight-*` directives removed (the two heading diffs are YAML comments inside replaced code blocks).
- `npm run l && npm run f` clean.

## Note

This is docs-only — no `src/` behavior changed. Several findings are *source* bugs the docs were papering over (`mcp` `defaultArgs` has never been wired up; `f5` concatenates `apiBaseUrl` naively; `litellm` fails auth silently). Those are documented accurately here rather than fixed, and are worth separate issues.


jameshiester-oai (commented):
Reviewed base `34f74d34e140b5e17d23770dfb2340057b1936b8` through head `7463c15cd84a5b5cf294467bffb6214dce8946d7` (fresh full review).

Verdict: one P2 finding (off-diff): `site/docs/providers/index.md:62` still recommends `deepseek:deepseek-r1`, while this PR rewrites `site/docs/providers/deepseek.md:45-63` to direct new configs to V4 Flash/Pro and removes DeepSeek-R1 from current pricing guidance. Update the index example to `deepseek:deepseek-v4-flash` or Pro so the landing page does not lead users to a retired/unpriced option.

Existing unresolved feedback remains valid (including the Mistral anchor and provider-specific docs corrections) and was revalidated without duplication.

CI: all listed checks passed.

QA: `SKIP_OG_GENERATION=true npm run build` was started locally; optional stats fetch hit GitHub API rate limiting and the command window ended during Docusaurus compilation, so no local build pass is claimed. Local Prettier and `git diff --check` passed; CI Docs build and site tests are green.

Limitations: requested toolkit launcher unavailable; three independent read-only reviews reconciled.

PR #10989, opened 2026-09-18
fix(providers): repair provider config bugs found in the docs audit

mldangelo:
Fixes the source bugs surfaced while auditing the provider docs in #10987. Each was verified by constructing the provider, not by reading.

> **Stacked on #10987.** Based on `docs/evergreen-provider-audit` so this diff shows only the source changes; GitHub will retarget to `main` when that merges.

### 1. Gateway providers discarded user config

`cloudera`, `jfrog` and `f5` assigned `apiBaseUrl` and `apiKeyEnvar` **after** spreading the user's config, so an explicit value in `config:` was silently dropped:

```ts
config: {
  ...providerOptions.config,   // user's apiBaseUrl lands here...
  apiKeyEnvar: 'CDP_TOKEN',
  apiBaseUrl: `https://${domain}/...`,  // ...then gets overwritten
}
```

That contradicts the documented precedence (*explicit options > environment > provider defaults*) and left no way to route these providers through a proxy. `atlascloud` already did this correctly; these now match it.

### 2. f5 built its URL by string concatenation

```ts
apiBaseUrl: providerOptions.config?.apiBaseUrl + '/' + endpoint
```

| Input | Produced |
|---|---|
| no `apiBaseUrl` | request to the literal string `undefined/path-name` |
| `https://gw.example.com/` | `https://gw.example.com//path-name` |

Now requires a gateway URL (clear error), accepts `F5_API_BASE_URL`, and trims trailing slashes.

### 3. MCP `defaultArgs` had never worked

Documented as a `config` key, but read from `options.defaultArgs` — a *constructor* option the registry never passes (`new MCPProvider({ config, id })`). The merge at call time always spread `{}`. Now read from config, with the constructor option kept for programmatic callers, and declared on the schema instead of being an undeclared passthrough key.

### 4. SageMaker overrode an explicit model type

An endpoint name containing `jumpstart` was coerced to the jumpstart type, so `sagemaker:huggingface:my-jumpstart-endpoint` resolved to `jumpstart` and requests were formatted for the wrong container. The coercion only ran on the three-segment form — where the type is always explicit and already validated — so it could never help, only override.

**`test/providers/registry.test.ts` encoded the bug as expected behavior** (`'sagemaker:custom:my-jumpstart-endpoint'` → `'jumpstart'`). Corrected, with cases added for the huggingface and jumpstart forms.

### 5. LiteLLM failed auth silently

`apiKeyEnvar` is pinned to `LITELLM_API_KEY` and `apiKeyRequired: false`, so a user who set only `OPENAI_API_KEY` got **no Authorization header and no error** — just a 401 from the proxy with nothing pointing at the cause. Now warns.

It deliberately does **not** fall back to `OPENAI_API_KEY`. `resolveProviderApiKey` namespaces credentials per provider on purpose (*"A named credential selects its own namespace, rather than falling back to another vendor"*), and falling back would send an OpenAI key to an arbitrary proxy URL. A test asserts that non-fallback.

### 6. Two smaller ones

- `envoy` read `process.env.ENVOY_API_BASE_URL` directly, skipping both the config-level and provider-level `env:` blocks. Now resolves config > provider env > context env > process env, like litellm.
- The `llama:` factory used `providerPath.split(':')[1]`, truncating model names containing colons (`llama:llama3:8b` → `llama3`). `localai` already had this fix.

## Tests

`cloudera`, `jfrog`, `f5` and `envoy` had **no test coverage at all**, which is the mechanical reason these went unnoticed. Adds `test/providers/gatewayConfig.test.ts` (15 cases) plus MCP and LiteLLM coverage.

- `npx vitest run test/providers/ test/sagemaker.test.ts` → **11,290 passed**, 0 failed
- `npx tsc --noEmit` clean; docs site builds clean
- The two `noExcessiveCognitiveComplexity` warnings on `registry.ts` and `families/aws.ts` are pre-existing — verified against `main`

Docs in #10987 that described the broken behavior are updated in the same commits.


jameshiester-oai (commented):
Reviewed stacked base `7463c15cd84a5b5cf294467bffb6214dce8946d7` through head `011e9ad8acf23054f511d9dbdefe375b57c1d241` (fresh after head advance).

Verdict: one P1 inline finding below. Two P2 off-diff findings: `site/docs/providers/f5.md:10-11` says `apiKeyEnvar` is fixed, but registry now honors an explicit override; and `architecture/edge-baseline.json` accepts two new providers-to-legacy-runtime imports without architectural rationale.

CI: current checks are pending; the prior head failed only the architecture gate.

QA: focused provider suites — 369 passed; `npm run architecture:check` and `git diff --check` pass at the new head.

Limitations: no live gateways were contacted. Toolkit launcher unavailable; independent read-only reports reconciled.

src/providers/registry.ts
```diff
@@ -671,12 +671,22 @@ export const providerMap: ProviderFactory[] = [
       if (endpoint.startsWith('/')) {
         endpoint = endpoint.slice(1);
       }
+      const configuredBaseUrl =
+        providerOptions.config?.apiBaseUrl ?? getEnvString('F5_API_BASE_URL');
```
jameshiester-oai, line 675: **[P1] Resolve the F5 base URL from scoped provider environment**

`loadApiProvider` merges suite and provider `env:` values onto `providerOptions.env`, but this factory ignores that map and checks only config or process environment. A standard F5 config with `env: { F5_API_BASE_URL: ... }` still throws. Use merged scoped env before the global fallback and add regression cases.

PR #10994, opened 2026-09-19
fix(eval): sanitize invalid XML characters in JUnit reports

emme1t:
JUnit exports can contain XML 1.0 forbidden characters when a provider label or test description includes control characters, U+FFFE, U+FFFF, or an unpaired UTF-16 surrogate. CI report readers then reject the XML document even when the evaluation succeeds.

Filter characters outside the XML 1.0 character range after serialization. This preserves the original grouping keys, ordinary XML escaping, and valid Unicode text. The regression cases parse the result with the existing `saxes` dependency and verify the exported names, including Chinese and emoji.

### Reproduction

Save this as `junit-repro.json`:

```json
{
  "providers": [{ "id": "echo", "label": "target\u0001 & <model>" }],
  "prompts": ["Hello {{name}}"],
  "tests": [
    {
      "description": "case\u0000 \ud83d\ude80 \u4e2d\u6587",
      "vars": { "name": "World" },
      "assert": [{ "type": "contains", "value": "World" }]
    },
    {
      "description": "case\ufffe \ud83d\ude80 \u4e2d\u6587",
      "vars": { "name": "World" },
      "assert": [{ "type": "contains", "value": "World" }]
    }
  ]
}
```

```sh
npm run local -- eval -c junit-repro.json --no-cache --no-write --no-table -o result.json result.junit.xml
python -c "import xml.etree.ElementTree as ET; print(ET.parse('result.junit.xml').getroot().tag)"
```

Before: both echo evaluations succeed with score 1, and the XML parser raises `ParseError: not well-formed (invalid token)`. After: both evaluations still succeed, and the parser prints `testsuites`. I also inspected both exported records and verified the report counts, names, XML escaping, and Unicode text.

### Validation

On Windows with the repository's Node 24.21.0 and npm 11.19.0:

- The new regression cases fail before the change: 10 failed, 67 passed.
- `npx vitest run test/util/output.test.ts test/util/outputFormats.test.ts --maxWorkers=1`: 80 passed.
- `npm run tsc`: passed.
- `npm run l` and `npx biome check src/util/junit.ts test/util/output.test.ts`: passed.
- `npm run f` formatted the two changed TypeScript files; its subsequent empty Prettier input reports `No parser and no file path given`. The explicit Biome check above passes.

- `npm test -- --maxWorkers=2 --bail=1`: 609 passed, 38 skipped, then stopped at `test/providers/openinterpreter.test.ts:356` because Windows denied `fs.symlinkSync` with `EPERM`. This occurs during that test's fixture setup.

- `npm run build` encounters the existing Windows `concurrently` child-shell error: `'NODE_OPTIONS' is not recognized`. With the same `NODE_OPTIONS=--max-old-space-size=8192` set in the process environment, the separate `npx tsdown` and `npm run postbuild` commands both pass.

- The generated CLI also passes the same reproduction: `node dist/src/entrypoint.js eval -c junit-repro.json --no-cache --no-write --no-table -o result.json result.junit.xml`, followed by the JSON/XML content checks above.

Developed with Codex collaboration as part of an Astra Ultra based automated PR project contributing to the open-source community.


emme1t:
Synced this branch with `main` at `a33f261` in merge commit `9beecd7`. The PR diff remains the JUnit sanitizer and its regression tests. I also corrected the reproduction JSON in the description to use valid JSON escapes for control characters.

Validation on Node 24.21.0 / npm 11.19.0 after a fresh `npm ci`:
- Output regression suites: **80 passed**.
- `npm run tsc`, changed-file lint, explicit Biome checks, and the commit hook: **passed**. `npm run f` formats the two TypeScript files, then reports its existing empty-input Prettier error.
- Real local echo evaluation with `--no-cache --no-write`: **2 passed**, both score 1, with successful JSON/XML content checks for report counts, names, XML escaping, Chinese, and emoji.
- Full-suite attempt: **3306 passed, 1 skipped**, then stopped on the 30s timeout in `test/evaluator/metrics.test.ts` (`should count rows in metric update order when persistence finishes out of order`). Rerunning that entire file independently passed **15/15**; the file and evaluator implementation match `main`.

The updated branch is ready for the refreshed CI and maintainer review. Follow-up completed with Codex collaboration.

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo-oai:
@codex review

mldangelo:
Thank you!

